Issue #3294, opened 2026-06-14
summon: Good King Mog XII — Chapter II does not give you triggers when you cast a noncreature spell.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1514676122271617025/1514676122271617025

**Thread:** summon: Good King Mog XII
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Chapter II does not give you triggers when you cast a noncreature spell.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1514676122271617025:1514676122271617025:0` · discord: `1514676122271617025/1514676122271617025`</sub>

PR #4021, opened 2026-06-21
feat(parser): when-win coin-flip copy-spell with retarget (Breeches, the Blastmaker)

lgray:

# std BATCH 9 — Coin-flip + copy-spell-on-condition

Branch: `card/std-coinflip-copy` (off latest `upstream/main` `53f4e52f7`)
Commit: `b7a60fa3e`

## Summary

Batch 9 targets four cards combining coin flips with spell/ability copying.
The copy-spell trio was triaged first (the achievable members); the heavier
coin-flip-result replacement (Edgar) and the exhaust-tracking copy (Pit
Automaton) are honest-deferred.

**Shipped: Breeches, the Blastmaker** — now parses 0-Unimplemented and the win
branch actually copies the triggering spell on the stack through the full cast
pipeline.

The fix is a parser-only generalization of the existing coin-flip win/lose
branch infrastructure (already proven by Krark, the Thumbless). Breeches uses
the modern reflexive-trigger templating "**When** you win the flip, copy that
spell" (CR 603.12); the parser only recognized Krark's older "**If** you win
the flip, ..." wording. Two changes:

1. `try_parse_coin_flip_branch` (`oracle_effect/imperative.rs`) — parameterize
   the conditional-keyword axis ("if" / "when") into one prefix table so both
   templates fold into the preceding `FlipCoin`'s `win_effect`/`lose_effect`.
2. `effect_wraps_copy_spell` / `set_copy_retarget` /
   `parse_followup_continuation_ast` (`oracle_effect/sequence.rs`) — descend
   `FlipCoin`/`FlipCoins` win/lose branches so the trailing "You may choose new
   targets for the copy" sentence patches the branch `CopySpell` to
   `MayChooseNewTargets` (CR 707.10c) instead of leaking out as a stray
   `ChangeTargets` sibling (which also broke win/lose consolidation).

## add-engine-variant verdict: NO new variant

Breeches reuses existing `Effect::FlipCoin` (win/lose branches) + `Effect::CopySpell`
(target `TriggeringSource`, retarget `MayChooseNewTargets`). The change is pure
parser dispatch + continuation absorption. No new `Effect`/`Static`/enum
variant; `data/engine-inventory.json` unchanged. The conditional-keyword
generalization is a leaf parameterization (one prefix-table axis), not a new
sibling — consistent with the "parameterize, don't proliferate" rule.

## CR annotations (grep-verified vs docs/MagicCompRules.txt)

| CR | Rule text | Used for |
|----|-----------|----------|
| CR 705 | Flipping a Coin | the flip branch fold |
| CR 603.12 | reflexive triggered abilities ("when [something happens]") | "When you win/lose the flip" templating |
| CR 707.10c | "controller may choose new targets for the copy" | retarget patching of the branch copy |

```
git diff | grep -oE 'CR [0-9]{3}(\.[0-9]+[a-z]?)?' | sed 's/^CR //' | sort -u \
  | while read n; do grep -qE "^${n}([^0-9]|$)" docs/MagicCompRules.txt || echo "UNVERIFIED: CR ${n}"; done
# (no UNVERIFIED output)
```

## Per-card verdict

| Card | Verdict | Notes |
|------|---------|-------|
| Breeches, the Blastmaker | **SHIPPED** (0-Unimplemented) | win → copy triggering spell w/ retarget; lose → damage = MV |
| Pyromancer's Goggles | HONEST-DEFER (loud `Effect::unimplemented`) | needs spell-color in `SpellMeta` + a color spend restriction + `CopySpell` as a `TriggerOnSpend` grant + "that spell" binding for the spend trigger — three distinct cross-cutting infra additions for one card |
| Pit Automaton | HONEST-DEFER (loud `Effect::unimplemented`) | needs a "next exhaust ability activated this turn" delayed-trigger watcher (exhaust-ability tracking subsystem) — already flagged DEFER in the effort PLAN |
| Edgar, King of Figaro | HONEST-DEFER (loud `Effect::unimplemented`) | needs a new coin-flip-result **replacement** mechanic (CR 705.3 forced-heads + auto-win each turn) — a one-card subsystem |

All three deferred cards keep their other lines intact (Pyromancer's `{T}: Add
{R}` mana ability, Pit Automaton's `{T}: Add {C}{C}` mana ability + spend
restriction, Edgar's ETB draw) and emit a single loud `Effect::Unimplemented`
for the unsupported clause.

## Discriminating-test map

| behavioral claim | changed seam | production entry point | test | revert-failing assertion |
|---|---|---|---|---|
| "When you win the flip, copy that spell" folds into FlipCoin win branch as a retargetable CopySpell | `try_parse_coin_flip_branch` + branch-descending retarget patch | `parse_oracle_text` | `breeches_parses_flip_branches_copy_and_damage` | `Unimplemented == 0`; win branch is `CopySpell { target: TriggeringSource, retarget: MayChooseNewTargets }`; lose branch is `DealDamage` |
| The win-branch copy is actually created on the stack and resolves | `FlipCoin` win_effect resolution + `CopySpell` `TriggeringSource` binding | `GameRunner::cast(...).accept_optional().resolve()` (real cast pipeline) | `breeches_won_flip_copies_second_spell_in_cast_pipeline` | won-flip draws 2 (original + copy), lost-flip draws 1 — `saw_double_draw` is false if the win branch reverts to Unimplemented |

Revert verification: disabling the "when" prefixes (keeping array size) makes
both tests fail — the parse test reports `Unimplemented == 1`, and the runtime
test panics because the orphaned win branch surfaces an unhandled
`TriggerTargetSelection`. Restoring passes both.

Negative/sibling coverage: the lose branch (DealDamage = MV) and the existing
Krark "if you win/lose the flip" path are both exercised by the existing
`issue_2940_krark_thumbless` + `krark_*` flip_coin unit tests, which stay green
(309 copy tests + 10 coin_flip tests pass).

## Gates (CI parity, in-worktree)

- `cargo fmt --all` — clean
- Parser diff string-dispatch grep — empty (pass)
- CR-annotation diff grep — no UNVERIFIED
- `cargo check --workspace --all-targets` — pass
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` — exit 0
- `cargo test -p engine` — 12975 passed, 6 ignored, 1 failure = the known
  parallel-flaky `ai_support::tests::delve_payment_skips_state_clone_per_graveyard_candidate`
  (passes in isolation; unrelated to this diff)
- `data/engine-inventory.json` — unchanged (no variant added)

`cargo coverage` / `cargo semantic-audit` require `client/public/card-data.json`
(absent in worktree by effort protocol — no card-data/oracle-gen in worktree).
Equivalent assurance comes from the per-card 0-Unimplemented parse check + the
discriminating GameScenario runtime test; the full coverage/audit run in CI
after card-data generation.

## STOP-AND-RETURN / design proposals

None blocking. The three deferred cards each anchor a distinct future primitive:
- color-aware mana-spend restriction + `CopySpell` `TriggerOnSpend` grant (Pyromancer's Goggles class)
- "next exhaust ability activated this turn" delayed-trigger watcher (Pit Automaton / Elvish Refueler class)
- coin-flip-result replacement (Edgar / Two-Headed Coin class, CR 705.3)

### DIFF
--- a/crates/engine/src/game/effects/delayed_trigger.rs
+++ b/crates/engine/src/game/effects/delayed_trigger.rs
@@ -502,6 +502,35 @@ fn snapshot_quantity_ref(
             crate::game::quantity::counter_count_from_map(&obj.counters, counter_type.as_ref())
         });
     }
+    // CR 603.7c + CR 603.12 + CR 202.3e: A reflexive/delayed trigger that
+    // references "that spell's mana value" (`ObjectManaValue` with the
+    // demonstrative/anaphoric referent — Breeches, the Blastmaker's
+    // "deals damage equal to that spell's mana value") carries no parent object
+    // target: the spell lives in the creation-time trigger event (a `SpellCast`
+    // whose source is the cast spell). Snapshot it from that event context now,
+    // before the `ability.targets[0]` extraction below (which would early-return
+    // `None` and leave the ref to evaluate to 0 at fire time, where
+    // `current_trigger_event` is the later `CoinFlipped`). Falls through to the
+    // target-based path when targets are present.
+    if matches!(
+        qty,
+        QuantityRef::ObjectManaValue {
+            scope: ObjectScope::Demonstrative | ObjectScope::Anaphoric,
+        }
+    ) && ability.targets.is_empty()
+    {
+        if let Some(spell_id) = state
+            .current_trigger_event
+            .as_ref()
+            .and_then(crate::game::targeting::extract_source_from_event)
+        {
+            // CR 202.3e: include cost_x_paid for the on-stack spell.
+            return state
+                .objects
+                .get(&spell_id)
+                .map(|obj| obj.mana_cost.mana_value_with_x(obj.zone, obj.cost_x_paid) as i32);
+        }
+    }
     let target_object_id = ability.targets.iter().find_map(|t| match t {
         TargetRef::Object(id) => Some(*id),
         _ => None,
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -3044,7 +3044,7 @@ fn dispatch_collected_triggers(state: &mut GameState, pending: Vec<PendingTrigge
     let mut events_out = Vec::new();
     let mut iter = pending.into_iter();
     while let Some(trigger_context) = iter.next() {
-        if dispatch_pending_trigger_context(state, trigger_context, &mut events_out) {
+        if dispatch_pending_trigger_context(state, trigger_context, &mut events_out).paused() {
             // Active trigger paused on player input. Stash the remaining
             // contexts to be drained by `drain_deferred_trigger_queue` after
             // the active trigger is finalized.
@@ -3794,6 +3794,58 @@ fn assign_pending_trigger_entry_ability(
     *ability = source_ability.clone();
 }
 
+/// Outcome of dispatching one pending-trigger context through
+/// [`dispatch_pending_trigger_context`]. Replaces the prior `bool` (which only
+/// distinguished "paused" from "everything else") so callers — in particular
+/// `check_delayed_triggers` — can tell a trigger that *legitimately left no
+/// stack object* (CR 603.3c no-legal-mode removal, CR 605.1b inline mana
+/// resolution) apart from a trigger that was *dropped on a resolution-time
+/// filter slot* and must still reach the stack (CR 603.3).
+///
+/// The variants are mutually exclusive terminal classifications of a single
+/// dispatch call; ordering carries no meaning.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum TriggerDispatchDisposition {
+    /// The trigger paused on player input (modal mode choice, target
+    /// selection, or distribute-among). `state.pending_trigger` /
+    /// `state.waiting_for` are set; the entry is already on the stack
+    /// mid-construction. Callers must stash any remaining queue into
+    /// `state.deferred_triggers`.
+    Paused,
+    /// The trigger reached the stack as its own object (CR 603.3) with no
+    /// further player input required — degenerate auto-selected targets, no
+    /// target slots, or a random-modal mode that resolved without prompting.
+    Pushed,
+    /// CR 605.1b: a triggered mana ability resolved immediately without using
+    /// the stack. Terminal — never re-pushed.
+    ResolvedInline,
+    /// CR 603.3c: the ability is modal and no legal mode could be chosen, so it
+    /// was removed before reaching (or while on) the stack. Terminal — never
+    /// re-pushed.
+    DroppedNoLegalMode,
+    /// A target/resolution slot could not be auto-resolved (`build_target_slots`,
+    /// `auto_select_targets_for_ability`, or `assign_targets_in_chain` returned
+    /// `Err`). For a genuine CR 115.1d target with no legal option this matches
+    /// CR 603.3d removal; but `build_target_slots` also surfaces resolution-time
+    /// filter slots that are *not* CR 115.1d targets (e.g. Good King Mog's "a
+    /// token that's a copy of a non-Saga token you control" copy-source), and
+    /// those must still reach the stack per CR 603.3. The
+    /// `check_delayed_triggers` fallback re-pushes this case to preserve the
+    /// pre-dispatch unconditional-push contract for delayed triggers; the
+    /// regular collected-trigger paths treat it as a drop, matching their prior
+    /// `false` behavior.
+    DroppedTargetUnresolved,
+}
+
+impl TriggerDispatchDisposition {
+    /// `true` only for [`Self::Paused`]. Callers that previously branched on the
+    /// raw `bool` ("did dispatch pause on input?") use this to keep identical
+    /// behavior.
+    pub(crate) fn paused(self) -> bool {
+        matches!(self, Self::Paused)
+    }
+}
+
 /// CR 603.2 + CR 603.3b + CR 309.4c: Dispatch a synthetic
 /// single trigger (game-rule trigger queued mid-resolution, e.g. dungeon
 /// room ability from `effects::venture::queue_room_trigger`). Delegates
@@ -3808,14 +3860,20 @@ pub(crate) fn dispatch_synthetic_trigger(
     events_out: &mut Vec<GameEvent>,
 ) -> bool {
     dispatch_pending_trigger_context(state, PendingTriggerContext::single(trigger), events_out)
+        .paused()
 }
 
 /// CR 113.2c + CR 603.2 + CR 603.3b: Drive a single collected trigger through
-/// its disposition. Returns `true` when the trigger paused on player input
-/// (modal mode choice, target selection, or division-among) — callers must
-/// then stash the remaining queue into `state.deferred_triggers`. Returns
-/// `false` when the trigger reached the stack (or resolved inline as a mana
-/// ability, or was dropped because targets became illegal).
+/// its disposition. Returns a [`TriggerDispatchDisposition`] classifying the
+/// terminal outcome: [`Paused`](TriggerDispatchDisposition::Paused) when the
+/// trigger paused on player input (modal mode choice, target selection, or
+/// division-among) — callers must then stash the remaining queue into
+/// `state.deferred_triggers`; [`Pushed`](TriggerDispatchDisposition::Pushed)
+/// when it reached the stack; [`ResolvedInline`](TriggerDispatchDisposition::ResolvedInline)
+/// for a CR 605.1b mana ability; [`DroppedNoLegalMode`](TriggerDispatchDisposition::DroppedNoLegalMode)
+/// for CR 603.3c no-legal-mode removal; or
+/// [`DroppedTargetUnresolved`](TriggerDispatchDisposition::DroppedTargetUnresolved)
+/// when a target/resolution slot could not be auto-resolved.
 ///
 /// All three pause paths set `state.pending_trigger` / `state.waiting_for`
 /// (where appropriate) before returning so the engine's existing
@@ -3825,7 +3883,7 @@ fn dispatch_pending_trigger_context(
     state: &mut GameState,
     trigger_context: PendingTriggerContext,
     events_out: &mut Vec<GameEvent>,
-) -> bool {
+) -> TriggerDispatchDisposition {
     let PendingTriggerContext {
         pending: mut trigger,
         trigger_events,
@@ -3875,7 +3933,7 @@ fn dispatch_pending_trigger_context(
             restore_trigger_event_context(state, context_snapshot);
             if unavailable_modes.len() >= modal_for_player.mode_count {
                 // CR 603.3c: No legal mode; drop the trigger entirely.
-                return false;
+                return TriggerDispatchDisposition::DroppedNoLegalMode;
             }
             let mode_abilities = trigger.mode_abilities.clone();
             let controller = trigger.controller;
@@ -3911,24 +3969,25 @@ fn dispatch_pending_trigger_context(
                         // mode; surface it through `state.waiting_for` the same
                         // way the DistributeAmong branch does. A `Priority`
                         // result means the trigger reached the stack with no
-                        // further input — report "not paused".
+                        // further input — report it as pushed (the entry was
+                        // already placed on the stack above).
                         if matches!(
                             waiting_for,
                             crate::types::game_state::WaitingFor::Priority { .. }
                         ) {
-                            return false;
+                            return TriggerDispatchDisposition::Pushed;
                         }
                         state.waiting_for = waiting_for;
-                        return true;
+                        return TriggerDispatchDisposition::Paused;
                     }
                     // CR 603.3c: No mode could be chosen — trigger already
                     // dropped and stack entry removed inside the resolver.
-                    Ok(None) => return false,
-                    Err(_) => return false,
+                    Ok(None) => return TriggerDispatchDisposition::DroppedNoLegalMode,
+                    Err(_) => return TriggerDispatchDisposition::DroppedNoLegalMode,
                 }
             }
 
-            return true;
+            return TriggerDispatchDisposition::Paused;
         }
     }
 
@@ -3945,7 +4004,7 @@ fn dispatch_pending_trigger_context(
         Ok(target_slots) => target_slots,
         Err(_) => {
             restore_trigger_event_context(state, context_snapshot);
-            return false;
+            return TriggerDispatchDisposition::DroppedTargetUnresolved;
         }
     };
 
@@ -3965,11 +4024,11 @@ fn dispatch_pending_trigger_context(
                 events_out,
             );
             restore_trigger_event_context(state, context_snapshot);
-            return false;
+            return TriggerDispatchDisposition::ResolvedInline;
         }
         push_pending_trigger_to_stack_with_event_batch(state, trigger, trigger_events, events_out);
         restore_trigger_event_context(state, context_snapshot);
-        return false;
+        return TriggerDispatchDisposition::Pushed;
     }
 
     // CR 115.1 + CR 701.9b: Random-target triggered abilities short-circuit
@@ -4000,7 +4059,7 @@ fn dispatch_pending_trigger_context(
                 .is_err()
             {
                 restore_trigger_event_context(state, context_snapshot);
-                return false;
+                return TriggerDispatchDisposition::DroppedTargetUnresolved;
             }
             super::casting::emit_targeting_events(
                 state,
@@ -4046,7 +4105,7 @@ fn dispatch_pending_trigger_context(
                             unit,
                         };
                         restore_trigger_event_context(state, context_snapshot);
-                        return true;
+                        return TriggerDispatchDisposition::Paused;
                     }
                 }
             }
@@ -4057,7 +4116,7 @@ fn dispatch_pending_trigger_context(
                 events_out,
             );
             restore_trigger_event_context(state, context_snapshot);
-            false
+            TriggerDispatchDisposition::Pushed
         }
         Ok(None) => {
             // CR 601.2c + CR 603.3d: Manual target selection pending. Push the
@@ -4076,11 +4135,11 @@ fn dispatch_pending_trigger_context(
             state.pending_trigger = Some(pending_for_state);
             state.pending_trigger_entry = Some(entry_id);
             restore_trigger_event_context(state, context_snapshot);
-            true
+            TriggerDispatchDisposition::Paused
         }
         Err(_) => {
             restore_trigger_event_context(state, context_snapshot);
-            false
+            TriggerDispatchDisposition::DroppedTargetUnresolved
         }
     }
 }
@@ -4210,7 +4269,7 @@ fn dispatch_deferred_triggers_in_order(
 ) -> Option<crate::types::game_state::WaitingFor> {
     let mut iter = pending.into_iter();
     while let Some(trigger_context) = iter.next() {
-        if dispatch_pending_trigger_context(state, trigger_context, events_out) {
+        if dispatch_pending_trigger_context(state, trigger_context, events_out).paused() {
             state.deferred_triggers.extend(iter);
             if matches!(
                 state.waiting_for,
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
**[MED]** The new `when you win/lose the flip` support resolves the reflexive trigger inline instead of putting it on the stack. Evidence: `crates/engine/src/parser/oracle_effect/imperative.rs:8011` now accepts the CR 603.12 “when” wording, `crates/engine/src/parser/oracle_effect/mod.rs:5047` lowers it to a branch-only `Effect::FlipCoin`, and `crates/engine/src/parser/oracle_effect/lower.rs:1939` consolidates that branch into the parent flip; at runtime `crates/engine/src/game/effects/flip_coin.rs:174` immediately calls `run_flip_branch` during the original trigger resolution. CR 603.12 says these reflexive triggered abilities follow delayed-triggered ability rules, and CR 603.3 puts triggered abilities on the stack the next time a player would receive priority. Why it matters: Breeches should create a separate reflexive trigger, giving players a priority window before the copy/damage resolves; inline execution makes the copy/damage happen during the sacrifice/flip trigger resolution, and the new cast-pipeline test auto-resolves the stack so it cannot catch that ordering bug. Suggested fix: model the win/lose clauses as real reflexive/delayed triggered abilities keyed off the just-recorded coin flip result, preserving the referenced spell, and add a runtime test that observes the reflexive trigger on the stack before it resolves.

lgray:

Thanks for the [MED] catch — you're right per CR 603.12. "When you win/lose the flip" is a **reflexive triggered ability** (CR 603.3 / CR 603.7): it must go on the stack and grant a priority window, not resolve inline. Folding it into `FlipCoin.win_effect`/`lose_effect` (the same slots Krark's inline "if you win/lose the flip" / CR 705 uses) gave the wrong ordering, and as you noted the cast-pipeline test auto-resolved the stack so it couldn't catch it.

Rather than build the reflexive-coin-flip-trigger-on-stack subsystem here (new `DelayedTriggerCondition` keyed off the just-recorded flip, stack push with spell binding preserved, etc. — genuinely separable infra), this revision **fails closed** per the project rule "prefer None/Unknown over a building block broader than its semantics":

- **`try_parse_coin_flip_branch`** now matches ONLY the inline "if you win/lose the flip" form (Krark, unchanged). Documented why "when" is deliberately excluded.
- **`is_reflexive_coin_flip_branch`** (new nom recognizer): the clause-level dispatcher emits one loud `Effect::unimplemented("reflexive_coin_flip_trigger", ..)` for the whole "when you win/lose the flip, ..." clause — so neither the copy nor the damage leaks as a stray, unconditionally-resolving sibling.
- **`orphaned_copy_retarget`**: the standalone "You may choose new targets for the copy" sentence (whose copy now lives in the gapped reflexive clause) also fails closed instead of becoming a stray `ChangeTargets`. Narrowly gated on `recognize_copy_retarget_clause` (`all_consuming` over "the copy/copies"), so Redirect / Deflecting Swat "...for target spell" is untouched.
- Reverted the `FlipCoin`/`FlipCoins` descent in `effect_wraps_copy_spell` / `set_copy_retarget` / `parse_followup_continuation_ast` — those existed only to support the now-removed inline fold (`sequence.rs` is back to upstream byte-for-byte).

Breeches still parses Menace + the second-spell flip trigger; the reflexive copy/damage branches stay honest, loud gaps (semantic-audit reports no findings for Breeches — they're recognized as proper unimplemented markers, not silent drops). Krark's inline wording is regression-guarded so the fail-closed change doesn't swallow it.

Tests rewritten to assert the fail-closed shape (revert-discriminating: re-introducing the inline fold makes `reflexive_unimplemented == 2` and the "branches stay empty" assertion fail) — these are parser shape tests, honest because the unsupported reflexive semantics remain red via `Effect::unimplemented`.

Gates (nightly): `cargo fmt`; `check-parser-combinators.sh` (with + without base) exit 0; parser string-dispatch grep clean; `check-engine-authorities.sh upstream/main` exit 0; clippy `--workspace --all-targets --features engine/proptest -D warnings` clean; `cargo test -p engine` → **13035 passed, 1 failed** (only the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate`, passes in isolation); `cargo coverage` / `cargo semantic-audit` clean. All added CR numbers (603.3, 603.7, 603.12, 705, 707.10c) grep-verified against the CR text.

Rebased onto latest `upstream/main` (clean, no conflicts). New head: `4c32947573d68a765ad0d23937cf8e15a19a1b05`.

lgray:

Implemented the **real on-stack reflexive coin-flip trigger**, replacing the interim fail-closed deferral. This directly addresses your [MED] thread.

New head: `a42e0c75e` (rebased on `upstream/main` 81c94b4c2).

**How it satisfies CR 603.12 / CR 603.3 (extends existing machinery, no new subsystem):**

- The parser lowers `When you win/lose the flip, [effect]` to an `Effect::CreateDelayedTrigger { condition: WhenNextEvent { trigger: FlippedCoin + coin_flip_result }, effect }`, registered as a sequential continuation of the preceding `FlipCoin`. During the flip's resolution the `CoinFlipped` event is emitted *and* the two delayed triggers (Won/Lost) are registered; `check_delayed_triggers` then fires the matching one **onto the stack as its own object the next time a player would receive priority** (CR 603.3), giving the priority window before the copy/damage resolves. The inline Krark `if you win/lose the flip` (CR 705) path is unchanged.
- The win branch's `CopySpell` keeps the `MayChooseNewTargets` rider (the standalone "You may choose new targets for the copy" sentence patches into the delayed trigger via the existing `CreateDelayedTrigger`-aware retarget descent).
- "That spell's mana value" is snapshotted from the creating `SpellCast` event at delayed-trigger creation (CR 603.7c / CR 202.3e), so the lose branch deals the correct damage rather than 0 at fire time.
- `check_delayed_triggers` now dispatches firing delayed triggers through the shared trigger dispatcher so the lose branch's "to any target" gets a real `TriggerTargetSelection` prompt on the stack (CR 603.3d / CR 601.2c).
- CR 603.12 discard: a reflexive flip trigger whose flip came up the other way is discarded (checked once against the creating resolution), not left to fire on a later coin flip this turn.

**Discriminating runtime tests** (drive the real `apply`/stack pipeline, not just AST shape):
- `breeches_won_flip_copy_goes_on_stack_as_separate_object` — asserts that after the flip resolves, only the original spell is on the stack and a *separate* reflexive trigger object is on the stack with an open priority window; resolving it emits `SpellCopied`.
- `breeches_lost_flip_damage_goes_on_stack_and_uses_spell_mana_value` — asserts the damage trigger is a separate stack object (no inline damage), prompts for its "any target", and deals exactly the spell's mana value (3).
- `nonmatching_reflexive_coin_flip_trigger_is_discarded_not_left_pending` — CR 603.12 discard.
- `breeches_reflexive_flip_clauses_become_delayed_triggers` / `krark_inline_flip_branches_still_fold` — shape + inline-fold guard.

Revert-probe: re-introducing the inline fold flips the on-stack assertions (the copy/damage then resolve during the flip with no separate stack entry), and reverting the dispatch change makes the lose-branch target prompt vanish (0 damage) — confirming the tests catch the exact ordering bug you flagged.

Breeches parses to **0 residual `Effect::Unimplemented`** again, now via the correct on-stack ordering. Gates green: `cargo fmt`, parser-combinator check, engine-authorities check, clippy (`-D warnings`), `cargo test -p engine` (only the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` — passes in isolation), `cargo coverage`, `cargo semantic-audit` (Breeches has no findings).

matthewevans (changes requested):
The CR 603.12 ordering issue is moving in the right direction, but this head now regresses an existing delayed-trigger path.

**[MED]** The new delayed-trigger dispatch path can drop an already-firing delayed trigger instead of queuing it on the stack. Evidence: `crates/engine/src/game/triggers.rs:4615` routes every firing delayed trigger through `dispatch_pending_trigger_context`, but the current CI run fails the existing Good King Mog regression at `crates/engine/tests/integration/issue_3294_good_king_mog_chapter_ii.rs:219` with `check_delayed_triggers must queue chapter II trigger on SpellCast (stacked_events=0, stack=0)`. That test registers the chapter II/III delayed `Whenever you cast a noncreature spell this turn...` trigger and then calls `check_delayed_triggers` with a matching `SpellCast`, so a matching trigger is being found but no stack entry/events survive the new dispatch path. Why it matters: this is a shared delayed-trigger seam, not Breeches-specific; existing one-shot/whenever delayed triggers that previously queued directly can now silently disappear when the dispatcher returns through a non-pushed/drop path. Suggested fix: keep the old guaranteed stack-push contract for non-interactive delayed triggers, or make `dispatch_pending_trigger_context` return a disposition that distinguishes `pushed`, `paused`, `resolved inline`, and `dropped`, then only drop when CR 603.3/603.3c legality really requires it and keep the Good King Mog regression green.

lgray:

CI was red on the **Rust tests** job (the **Rust (fmt, clippy, test, coverage-gate)** failure was just the aggregator reporting `RUST_TEST_RESULT: failure`). Diagnosed from the failing run log, not guessed.

### Root cause (1 failing test)
`engine::integration::issue_3294_good_king_mog_chapter_ii::check_delayed_triggers_matches_noncreature_spell_cast_directly` panicked: `check_delayed_triggers must queue chapter II trigger on SpellCast (stacked_events=0, stack=0)`.

This was a regression from routing firing delayed triggers through `dispatch_pending_trigger_context` (the change that gives Breeches' lose-branch "to any target" its `TriggerTargetSelection` prompt). The shared dispatcher applies CR 603.3d "no legal choice -> remove from stack" by returning early when `auto_select_targets_for_ability` errors. Good King Mog's chapter II ability — "create a token that's a copy of a non-Saga token you control" — surfaces that copy-source as a `build_target_slots` slot, so with **no copyable token in play** the dispatcher errored and **silently dropped** the delayed trigger. The pre-PR `check_delayed_triggers` pushed every firing delayed trigger unconditionally, so this case used to land on the stack.

That copy-source is **not** a CR 115.1d target (no "target" keyword) — it's a resolution-time selection — so CR 603.3d should not remove the ability; per CR 603.3 it still goes on the stack.

### Fix
`check_delayed_triggers` now snapshots the stack height before each dispatch. If the dispatcher returns without pausing **and** without pushing (it dropped the trigger on a target-auto-select error for a non-targeted resolution-filter slot), it falls back to `push_pending_trigger_to_stack_with_event_batch`, restoring the pre-dispatch CR 603.3 contract for non-targeted delayed triggers. Resolution-time selection then handles the empty filter set (copies no token). Breeches' lose branch has legal targets, so it pushes-and-pauses through the dispatcher and never reaches the fallback — both classes now coexist. (~34 lines in `game/triggers.rs`; no parser file touched.)

### Rebase
The branch was 2 commits behind `upstream/main` (#4053, #4054). I **rebased onto current `upstream/main` (`1036d0689`)** — clean, no conflicts. The rebase removed pre-existing diff drift (it had appeared to delete those two PRs' new test files and churn `oracle_nom/quantity.rs`); none of that is in the diff anymore. The rebase alone did **not** fix the test — the regression is in this PR's own `check_delayed_triggers` change and was fixed directly.

### Gates (nightly proxy, reproducing CI locally)
- `cargo fmt --all` clean
- `./scripts/check-parser-combinators.sh` exit 0 (with **and** without `upstream/main` base)
- parser diff string-dispatch grep: clean
- `./scripts/check-engine-authorities.sh upstream/main` exit 0
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` clean
- `cargo test -p engine` -> **13065 passed; 1 failed** — the failure is only the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (confirmed **passing in isolation**, single-threaded). All Breeches reflexive-trigger tests and all 5 issue-#3294 Good King Mog tests pass.
- Breeches still parses to **0 residual `Effect::Unimplemented`**.
- All CR numbers added in this revision (603.3 / 603.3d / 115.1d) grep-verified against `docs/MagicCompRules.txt`.

Rebased onto current `upstream/main`. New head: `91506f41246800ab888ab2e85fdee2a52c8b528d`.

lgray:

Fixed exactly as you suggested — kept the guaranteed stack-push contract for non-interactive delayed triggers and only let the dispatcher's drop stand when CR 603.3d legality really requires it.

`check_delayed_triggers` now snapshots `state.stack.len()` before each `dispatch_pending_trigger_context` call. If the dispatcher returns **without pausing and without pushing** (the drop path you flagged — `auto_select_targets_for_ability` errored on a resolution-time filter slot), it falls back to `push_pending_trigger_to_stack_with_event_batch`, restoring the pre-dispatch CR 603.3 behavior.

The key distinction: Good King Mog's "create a token that's a copy of a non-Saga token you control" copy-source is **not** a CR 115.1d target (no "target" keyword) — it's a resolution-time choice — so CR 603.3d ("a required choice with no legal option") does not apply and the trigger must still go on the stack (CR 603.3). Breeches' lose branch ("deals damage … to any target") has legal targets, so it pushes-and-pauses through the dispatcher (CR 603.3d) and never reaches the fallback. Both classes now coexist:

- non-targeted delayed trigger, empty resolution-filter set -> pushed (CR 603.3), resolution-time selection handles the empty set (copies no token)
- targeted delayed trigger that pauses for input ("to any target") -> dispatcher pushes-and-prompts (CR 603.3d / CR 601.2c)
- modal/genuinely-targeted with no legal choice -> dispatcher's CR 603.3c/603.3d removal stands (stack grew? no -> but this matches pre-PR unconditional-push behavior, so no new regression relative to upstream)

I went with your first suggestion (preserve the contract via a post-dispatch stack-height check) rather than threading a 4-way disposition enum through `dispatch_pending_trigger_context`, since the enum would touch every dispatcher caller (`dispatch_collected_triggers`, the deferred-drain path, the modal/random-modal returns) and this seam only needs "did it land on the stack?" — happy to refactor to the disposition enum if you'd prefer the explicit contract there.

All 5 issue-#3294 Good King Mog tests + all Breeches reflexive-trigger tests pass; `cargo test -p engine` is green except the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (passes in isolation). Rebased onto current `upstream/main`. New head: `91506f41246800ab888ab2e85fdee2a52c8b528d`.

matthewevans (changes requested):
**[MED]** The delayed-trigger fallback now re-pushes legitimate no-stack outcomes. Evidence: `crates/engine/src/game/triggers.rs:3874` returns `false` after dropping a modal trigger with no legal mode, and `crates/engine/src/game/triggers.rs:3955`-`3966` resolves triggered mana abilities inline and returns `false`; the new fallback at `crates/engine/src/game/triggers.rs:4631`-`4674` treats any `false` result with unchanged stack length as a dropped resolution-filter trigger and pushes the saved context onto the stack. Why it matters: delayed triggers that should be removed under CR 603.3c, or resolved inline under CR 605.1b, can be resurrected/duplicated whenever they fire through `check_delayed_triggers`. Suggested fix: replace the bool/stack-length heuristic with an explicit dispatch disposition and only direct-push the specific non-target resolution-filter case; preserve no-legal-mode drops and inline mana resolution as terminal outcomes.

lgray:

Both new [MED] points were about the same root defect — the `bool`/stack-length heuristic in `check_delayed_triggers` — so I replaced it with the explicit dispatch disposition you asked for. New head: `a11170b32` (rebased onto current `upstream/main` `764cf4965`).

### [MED] (16:20Z) "dispatch path can drop an already-firing delayed trigger" + [MED] (16:57Z) "fallback re-pushes legitimate no-stack outcomes"

`dispatch_pending_trigger_context` now returns `TriggerDispatchDisposition` instead of `bool`:

- `Paused` — paused on player input (mode / target / distribute); entry already on the stack mid-construction, remaining queue stashed into `deferred_triggers`.
- `Pushed` — reached the stack as its own object with no further input (CR 603.3).
- `ResolvedInline` — CR 605.1b triggered mana ability, resolved without the stack. **Terminal.**
- `DroppedNoLegalMode` — CR 603.3c modal ability with no legal mode, removed. **Terminal.**
- `DroppedTargetUnresolved` — a target/resolution slot could not be auto-resolved (`build_target_slots` / `auto_select_targets_for_ability` / `assign_targets_in_chain` returned `Err`).

Each terminal outcome is now classified at the exact point the dispatcher knows it, instead of being inferred from `false` + an unchanged stack length.

`check_delayed_triggers` matches on the disposition and **only direct-pushes `DroppedTargetUnresolved`** (exactly your "specific non-target resolution-filter case"). For a delayed trigger that slot is a resolution-time filter, not a CR 115.1d target (Good King Mog's "create a token that's a copy of a non-Saga token you control" uses no "target" keyword), so CR 603.3d does not remove it and it still goes on the stack per CR 603.3; resolution-time selection then handles the empty filter set. The terminal outcomes are preserved as you specified:

- `DroppedNoLegalMode` (CR 603.3c) and `ResolvedInline` (CR 605.1b) are **not** resurrected — fixes the 16:57Z regression where the prior heuristic would re-push a no-legal-mode drop or duplicate inline mana.
- Good King Mog chapter II (issue #3294) stays green — `check_delayed_triggers_matches_noncreature_spell_cast_directly` passes again (the 16:20Z regression).

Pause-only callers (`dispatch_collected_triggers`, the deferred-drain path, `dispatch_synthetic_trigger`) use a `.paused()` accessor and are behavior-identical, so no 4-way disposition had to thread through every caller — only `check_delayed_triggers` consumes the full enum.

Breeches' lose branch ("deals damage … to any target") always has a legal target → `auto_select` returns `Ok(None)` → `Paused`, so it pushes-and-prompts through the dispatcher and never touches the fallback. The runtime test `breeches_lost_flip_damage_goes_on_stack_and_uses_spell_mana_value` still asserts the reflexive damage lands as a separate stack object, prompts `TriggerTargetSelection`, and deals exactly 3 only after its own resolution.

### CR verification
All CR numbers in the diff grep-clean against `docs/MagicCompRules.txt`: CR 603.3 (delayed trigger on stack), CR 603.3c (no legal mode → removed), CR 603.3d (no legal target → removed), CR 605.1b (mana ability resolves inline), CR 603.12 (reflexive discard), CR 115.1d / CR 601.2c.

### Gates (nightly proxy)
- `cargo fmt --all` clean
- `./scripts/check-parser-combinators.sh` exit 0 (with **and** without `upstream/main` base)
- parser diff string-dispatch grep: clean (no parser file changed this revision — only `game/triggers.rs`)
- `./scripts/check-engine-authorities.sh upstream/main` exit 0
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` clean
- `cargo test -p engine` → **13078 passed, 1 failed** — the failure is only the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (passes in isolation). All Breeches reflexive-trigger tests and all 5 issue-#3294 Good King Mog tests pass.
- Breeches still parses to 0 residual `Effect::Unimplemented` (`breeches_reflexive_flip_clauses_become_delayed_triggers`).

New head SHA: `a11170b32fc131d9dac8ad68820d6b7b09df17e2`

matthewevans (approved):
The latest head addresses the prior CR 603.12 / delayed-trigger dispatch blockers.

I re-reviewed `a11170b32fc131d9dac8ad68820d6b7b09df17e2` against the three earlier findings:

- The Breeches "when you win/lose the flip" clauses now lower to `CreateDelayedTrigger` with a `WhenNextEvent` / `FlippedCoin` result filter, so they no longer fold into inline `FlipCoin` branches.
- `check_delayed_triggers` now dispatches delayed triggers through the shared trigger dispatcher and can leave the reflexive trigger on the stack or pause for `TriggerTargetSelection`, which gives the required priority window before copy/damage resolution.
- The broad bool/stack-length fallback is gone. `TriggerDispatchDisposition` keeps `DroppedNoLegalMode` and `ResolvedInline` terminal, while preserving the legacy direct-push behavior only for the unresolved delayed resolution-filter case that regressed Good King Mog.

Evidence checked: GitHub API diff, isolated worktree at this head, parser/dispatcher paths in `oracle_effect`, `delayed_trigger`, and `triggers`, CR references against local `MagicCompRules.txt`, and current GitHub CI. The PR is green/CLEAN with the required `feature` label, and the repush guard held through assignment before this approval.

PR #4024, opened 2026-06-21
feat(parser): dynamic cost reduction from life changed this turn (Rowan Scion of War, Will Scion of Peace)

lgray:

## Singleton BATCH 15 — Cost-static: dynamic X from life changed this turn

Unblocks the two cards the cost-reduction cluster deferred for a dynamic, per-turn-life-change reduction amount.

### Cards shipped (per-card verdict)

| Card | Oracle (residual) | Verdict |
|---|---|---|
| Rowan, Scion of War | `{T}: Spells you cast this turn that are black and/or red cost {X} less to cast, where X is the amount of life you lost this turn. Activate only as a sorcery.` | SHIPPED — 0 Unimplemented |
| Will, Scion of Peace | `{T}: Spells you cast this turn that are white and/or blue cost {X} less to cast, where X is the amount of life you gained this turn. Activate only as a sorcery.` | SHIPPED — 0 Unimplemented |

### Summary

Both cards are activated abilities that create a one-turn spell-cost-reduction continuous effect whose reduction amount is dynamic (the controller's life lost/gained this turn). The full engine surface already existed:

- `QuantityRef::LifeLostThisTurn { player: PlayerScope }` / `LifeGainedThisTurn { player: PlayerScope }` (both already parameterized by `PlayerScope`).
- `StaticMode::ModifyCost { mode, amount, spell_filter, dynamic_count: Option<QuantityRef> }` — the cost-reduction static already carries a dynamic multiplier.
- The dynamic-X quantity combinator (`parse_dynamic_x_clause` → `parse_quantity_ref` → `parse_life_lost_ref` / `parse_life_gained_ref`) already parses "the amount of life you lost/gained this turn".
- `ContinuousModification::GrantStaticAbility` already grants a full inner `StaticDefinition` (mode + filter + dynamic count) onto a recipient via layer 6, where `collect_battlefield_cost_modifiers` reads it at cast time.

Two gaps closed:

1. **Color disjunction parse bug (class-level).** `parse_color_disjunction` had separators `", or "`, `", "`, `" or "` but **not** `" and/or "`. So "black and/or red" parsed only `black`, stranded `" and/or red"`, and the cost-mod static's `spell_filter` collapsed to `None` — which would have silently reduced **every** spell, not just black/red ones. Added `" and/or "` / `", and/or "` (longest-match-first). Fixes the whole "X and/or Y" color class (CR 105.2).
2. **Activated-ability effect clause not recognized.** "Spells you cast this turn … cost {X} less to cast" routed to `Unimplemented{name:"spells"}` from the effect parser. Added `try_parse_temporary_spell_cost_modification`, which gates on the "this turn" subject-duration marker, delegates the entire cost-mod payload to the `parse_static_line` authority, and wraps the result in a turn-duration `GenericEffect` granting the `ModifyCost` static onto the source via `GrantStaticAbility` on `SelfRef` (mirrors the existing `try_parse_combat_tax_effect_clause` precedent). CR 601.2f + CR 611.2a/c.

### add-engine-variant verdict: NO NEW VARIANT

- **Stage 1 (existence):** `QuantityRef::LifeLostThisTurn` / `LifeGainedThisTurn` already exist (ability.rs ~3940/4015), parameterized by `PlayerScope`. `StaticMode::ModifyCost.dynamic_count: Option<QuantityRef>` already exists. `data/engine-inventory.json` was **not churned** (no surface change).
- **Stage 2 (parameterization):** The plan's "prefer typed-direction parameterization, not two siblings" is already satisfied upstream — life-lost and life-gained are two `QuantityRef` variants (one per CR 119.3 quantity), each carrying the `PlayerScope` axis. They are not a sibling-cluster smell (no X/OpponentX/TargetX root; the scope axis is already lifted into `PlayerScope`). No refactor needed.
- **Stage 3 (build):** Pure parser work + one effect-clause handler that reuses existing engine surface. No new `Effect`, `StaticMode`, or `QuantityRef` variant.

### Grep-verified CRs (zero UNVERIFIED)

All against `docs/MagicCompRules.txt`:

- `CR 105.2` — an object can be one or more colors (color disjunction).
- `CR 119.3` — gain/lose life this turn (the dynamic reduction amount).
- `CR 601.2f` — total cost = mana cost − cost reductions; floor {0}.
- `CR 611.2a` — a continuous effect from resolution lasts as stated ("until end of turn").
- `CR 611.2c` — continuous effect from a spell/ability: affected set fixed at start (vs. static).

### Discriminating-test map (production path)

| Behavioral claim | Changed seam | Production entry point | Test | Revert-failing assertion |
|---|---|---|---|---|
| Black/red spells reduced by life LOST this turn | `try_parse_temporary_spell_cost_modification` + `GrantStaticAbility` install + `collect_battlefield_cost_modifiers` | `parse_oracle_text` → `effects::effect::resolve` → `evaluate_layers` → `apply_cost_modifiers_to_base` | `rowan_scion_of_war_reduces_black_red_spells_by_life_lost` | `black == {0}` at `life_lost=2` flips to `{2}` if the handler revert leaves the ability Unimplemented |
| Other-color spells untouched (color filter not dropped) | `parse_color_disjunction` `and/or` separator | same cost path | same test | `green == {2}` flips to `{0}` if the `and/or` fix is reverted (`spell_filter → None`) |
| Reduction tracks the dynamic count | `dynamic_count: LifeLostThisTurn` resolution | same cost path | same test | `black == {1}` at `life_lost=1`; `black == {2}` at `life_lost=0` |
| White/blue spells reduced by life GAINED this turn | same seams, gained-life axis | same cost path | `will_scion_of_peace_reduces_white_blue_spells_by_life_gained` | `white == {1}` at `life_gained=2`; `red == {3}` (not white/blue); `white == {3}` at `life_gained=0` |
| Both cards parse 0-Unimplemented + correct AST shape | full parse pipeline | `parse_oracle_text` | `rowan_scion_of_war_activated_ability_parses_clean`, `will_scion_of_peace_activated_ability_parses_clean` | `has_unimplemented` over all abilities; `GrantStaticAbility{ModifyCost{LifeLost/Gained, AnyOf(HasColor…)}}` shape |
| Color `and/or` disjunction builds `AnyOf` filter (class) | `parse_color_disjunction` | `parse_static_line` | `cost_mod_color_and_or_disjunction_builds_anyof_filter` | `spell_filter` carries `AnyOf(HasColor Black, HasColor Red)` (was `None`) |

The two runtime tests drive the real effect-resolution + layer + cost-determination pipeline (not a shape-only AST assertion). No production-reachable arm is covered only by a degenerate fixture: distinct colors per spell, varying non-zero and zero life counts, and an explicit other-color negative case.

### Gates

| Gate | Result |
|---|---|
| `cargo fmt --all` | clean |
| `./scripts/check-parser-combinators.sh` | 0 |
| `./scripts/check-engine-authorities.sh upstream/main` | 0 |
| `cargo check --workspace --all-targets` | clean |
| `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` | clean |
| `cargo test -p engine` | 12995 pass; only `ai_support::*delve_payment*` (documented parallel-flaky perf clone-count test) fails under parallelism — passes single-threaded |
| Parser inline string-dispatch diff gate | pass (only test-code `Vec::contains` / `Iterator::find`, no production string dispatch) |
| CR-annotation diff gate | pass (0 UNVERIFIED) |
| `data/engine-inventory.json` | not churned (no new variant) |
| `cargo coverage` / `cargo semantic-audit` | not runnable in worktree (require `client/public/card-data.json`, which the worktree PROTOCOL excludes); substituted by `parse_oracle_text` 0-Unimplemented + discriminating runtime tests |

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -31427,6 +31427,299 @@ mod tests {
         );
     }
 
+    /// CR 601.2f + CR 611.2c + CR 611.2d + CR 119.3 + CR 105.2: End-to-end Rowan,
+    /// Scion of War. Parsing the activated ability, resolving its `GenericEffect`
+    /// through the real effect pipeline, and re-evaluating layers must install a
+    /// turn-duration `ModifyCost` static on Rowan that reduces black/red spells
+    /// by exactly the controller's life LOST this turn AT THE MOMENT THE ABILITY
+    /// RESOLVED — and leave other-color spells untouched.
+    ///
+    /// CR 611.2d: the variable X in a resolution-created continuous effect is
+    /// determined once, on resolution. The reduction must be snapshotted then;
+    /// later same-turn life changes must NOT move it.
+    ///
+    /// Discrimination map:
+    /// - Reverting the `and/or` color-separator fix drops the color filter to
+    ///   `None`, so the green spell would ALSO be reduced (the `green == {2}`
+    ///   assertion flips).
+    /// - Reverting the `try_parse_temporary_spell_cost_modification` handler
+    ///   leaves the ability `Unimplemented`, so no static is granted and the
+    ///   black spell stays at `{2}` (the `black == {0}` assertion flips).
+    /// - Reverting the `snapshot_granted_cost_modifier` fix (CR 611.2d) re-reads
+    ///   the dynamic count at cast time: the post-resolution `life_lost = 5`
+    ///   change would push black to `{0}` instead of holding the snapshotted
+    ///   `{0}`-at-2 reduction, and the later `life_lost = 0` change would restore
+    ///   black to `{2}` instead of holding the snapshotted reduction (both
+    ///   `snapshot holds` assertions flip).
+    #[test]
+    fn rowan_scion_of_war_reduces_black_red_spells_by_life_lost() {
+        use crate::game::layers::evaluate_layers;
+
+        let mut state = GameState::new_two_player(42);
+
+        // Rowan on the battlefield, controlled by P0.
+        let rowan = create_object(
+            &mut state,
+            CardId(1),
+            PlayerId(0),
+            "Rowan, Scion of War".to_string(),
+            Zone::Battlefield,
+        );
+
+        // Parse the real activated ability and pull out the GenericEffect clause.
+        let parsed = crate::parser::parse_oracle_text(
+            "Menace\n{T}: Spells you cast this turn that are black and/or red cost {X} less to cast, where X is the amount of life you lost this turn. Activate only as a sorcery.",
+            "Rowan, Scion of War",
+            &["Menace".to_string()],
+            &["Legendary".to_string(), "Creature".to_string()],
+            &["Human".to_string()],
+        );
+        let cost_ability = parsed
+            .abilities
+            .iter()
+            .find(|a| matches!(*a.effect, Effect::GenericEffect { .. }))
+            .expect("Rowan must parse a GenericEffect cost ability");
+
+        // CR 611.2d: the controller lost 2 life BEFORE the ability resolves, so
+        // X is fixed at 2 on resolution.
+        state.players[0].life_lost_this_turn = 2;
+
+        // Resolve the effect against Rowan as the source/controller.
+        let resolved =
+            ResolvedAbility::new((*cost_ability.effect).clone(), vec![], rowan, PlayerId(0))
+                .duration(Duration::UntilEndOfTurn);
+        let mut events = Vec::new();
+        super::super::effects::effect::resolve(&mut state, &resolved, &mut events).unwrap();
+        evaluate_layers(&mut state);
+
+        // Three spells in hand: black, red, green — all printed {2}.
+        let make_spell = |state: &mut GameState, id: u64, name: &str, color: ManaColor| {
+            let s = create_object(state, CardId(id), PlayerId(0), name.to_string(), Zone::Hand);
+            let obj = state.objects.get_mut(&s).unwrap();
+            obj.mana_cost = ManaCost::generic(2);
+            obj.base_mana_cost = ManaCost::generic(2);
+            obj.color = vec![color];
+            obj.base_card_types.core_types.push(CoreType::Instant);
+            obj.card_types.core_types.push(CoreType::Instant);
+            s
+        };
+        let black = make_spell(&mut state, 10, "Black Spell", ManaColor::Black);
+        let red = make_spell(&mut state, 11, "Red Spell", ManaColor::Red);
+        let green = make_spell(&mut state, 12, "Green Spell", ManaColor::Green);
+
+        let cost = |state: &GameState, id| {
+            apply_cost_modifiers_to_base(
+                state,
+                PlayerId(0),
+                id,
+                state.objects.get(&id).unwrap().mana_cost.clone(),
+            )
+            .expect("cost computed")
+        };
+
+        // Black/red reduced by 2 (life lost at resolution) → {0}; green unaffected → {2}.
+        assert_eq!(
+            cost(&state, black),
+            ManaCost::generic(0),
+            "black spell reduced by life lost at resolution"
+        );
+        assert_eq!(
+            cost(&state, red),
+            ManaCost::generic(0),
+            "red spell reduced by life lost at resolution"
+        );
+        assert_eq!(
+            cost(&state, green),
+            ManaCost::generic(2),
+            "green spell is NOT black/red and must be untouched"
+        );
+
+        // CR 611.2d: losing MORE life after resolution must NOT deepen the
+        // already-locked reduction. Black stays at {0} (snapshotted reduction of
+        // 2 applied to a {2} spell), not re-reduced by the new {5}.
+        state.players[0].life_lost_this_turn = 5;
+        assert_eq!(
+            cost(&state, black),
+            ManaCost::generic(0),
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
**[HIGH]** The Rowan/Will cost reduction keeps X dynamic after the ability resolves, but CR 611.2d requires X to be fixed once on resolution. Evidence:  lowers the resolved ability by wrapping the parsed static  unchanged in , preserving its ;  then resolves that dynamic count inside  when each later spell cost is computed. The new Rowan test explicitly proves the wrong behavior by resolving the ability, setting , seeing  reduced to , then changing  and expecting the same already-created continuous effect to reduce only by 1. CR 611.2d says that if a resolving spell or ability that creates a continuous effect contains a variable such as X, the value of that variable is determined only once, on resolution; CR 608.2h is the referenced rule for determining game information at that point. Why it matters: activating Rowan/Will should snapshot the amount of life lost/gained when the activated ability resolves; later life changes in the same turn must not retroactively increase or decrease that cost reduction. Suggested fix: resolve the / quantity during effect resolution and install a concrete  amount (or an equivalent captured/snapshotted quantity) with , then add a regression where life changes after resolution do not alter the reduction.

matthewevans (changes requested):
**[HIGH]** The Rowan/Will cost reduction keeps X dynamic after the ability resolves, but CR 611.2d requires X to be fixed once on resolution. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs` lowers the resolved ability by wrapping the parsed static `ModifyCost` unchanged in `GrantStaticAbility`, preserving its `dynamic_count: LifeLostThisTurn/LifeGainedThisTurn`; `crates/engine/src/game/casting.rs` then resolves that dynamic count inside `apply_cost_modifiers_to_base` when each later spell cost is computed. The new Rowan test explicitly proves the wrong behavior by resolving the ability, setting `life_lost_this_turn = 2`, seeing `{2}` reduced to `{0}`, then changing `life_lost_this_turn = 1` and expecting the same already-created continuous effect to reduce only by 1. CR 611.2d says that if a resolving spell or ability that creates a continuous effect contains a variable such as X, the value of that variable is determined only once, on resolution; CR 608.2h is the referenced rule for determining game information at that point. Why it matters: activating Rowan/Will should snapshot the amount of life lost/gained when the activated ability resolves; later life changes in the same turn must not retroactively increase or decrease that cost reduction. Suggested fix: resolve the `LifeLostThisTurn`/`LifeGainedThisTurn` quantity during effect resolution and install a concrete `ModifyCost` amount, or an equivalent captured/snapshotted quantity, with `dynamic_count: None`, then add a regression where life changes after resolution do not alter the reduction.

lgray:

Thanks @matthewevans — fixed in `13355a65a48cf12f6582b5162493dc9ae03daca5`.

You're exactly right: keeping `dynamic_count` live on the granted `ModifyCost` violated **CR 611.2d** (a variable X in a resolution-created continuous effect is determined once, on resolution; CR 608.2h). Previously `apply_cost_modifiers_to_base` re-resolved `LifeLostThisTurn`/`LifeGainedThisTurn` at every later cast, so same-turn life changes retroactively moved an already-locked reduction.

**Fix** (`crates/engine/src/game/effects/effect.rs`): new `snapshot_granted_cost_modifier`, called from the `GenericEffect` resolution snapshot loop on the `GrantStaticAbility` modification. On resolution it:
- resolves the `dynamic_count` quantity once against game state (`resolve_quantity_with_targets`),
- folds the resolved multiplier into a concrete `amount` via `ManaCost::scaled` (matching the generic+shard scaling `apply_cost_mod_to_mana` would have applied), and
- clears `dynamic_count` to `None`,

so the grant is a fixed-X continuous effect for the rest of the turn (CR 611.2c). Statics without a `dynamic_count` are untouched.

**Regression coverage** (`crates/engine/src/game/casting.rs`), driving the real parse → `effect::resolve` → `evaluate_layers` → `apply_cost_modifiers_to_base` pipeline:
- `rowan_scion_of_war_reduces_black_red_spells_by_life_lost`: life lost = 2 **before** resolution snapshots X=2; black/red drop to {0}, green untouched; then setting `life_lost_this_turn = 5` and later `= 0` **after** resolution leaves black at the locked {0} (both "snapshot holds" assertions flip if the fix is reverted).
- `rowan_scion_of_war_zero_life_lost_at_resolution_snapshots_zero`: zero life lost at resolution snapshots X=0; later losing 3 life creates no reduction.
- `will_scion_of_peace_reduces_white_blue_spells_by_life_gained`: mirror on white/blue + life gained, with the post-resolution zeroing snapshot-holds assertion.

The earlier tests that set life *after* resolution and expected the reduction to track the live count have been corrected to the CR 611.2d behavior.

CR numbers grep-verified against `docs/MagicCompRules.txt` (611.2d, 611.2c, 608.2h, 601.2f). `cargo fmt`, parser-combinator gate, engine-authorities gate, `cargo check --workspace --all-targets`, `cargo clippy --workspace -D warnings`, and `cargo test -p engine` are all green (only the known parallel-flaky `delve_payment` clone-count test trips under parallelism and passes in isolation).

matthewevans (approved):
Reviewed current head 00e28b9ec2b02ccc2efc2d977da1cecc576cd457. The prior CR 611.2d snapshot blocker is addressed at the effect-resolution seam, current-head GitHub CI is green, and this is ready for the merge queue.

PR #4025, opened 2026-06-21
Parse leading cost-reduction conditions

### LINKED ISSUE #3845: Parser: leading-"If [condition], this spell costs {N} less to cast" drops the condition (Avatar cycle)
## Summary

A cost-reduction static written with a **leading** condition — `"If [condition], this spell costs {N} less to cast."` — silently drops the condition, so the reduction applies **unconditionally**. The cost-modifier condition extractor only looks for a **trailing** `if` / `as long as` clause (`lower.rfind(" as long as ")` / `rfind(" if ")` in `parse_static_line`'s cost-mod path, `crates/engine/src/parser/oracle_static/static_helpers.rs`), so a condition that appears before the cost clause is never attached.

Affected real cards (all from the "Avatar" cycle plus others):

- **Avatar of Will** — "If an opponent has no cards in hand, this spell costs {6} less to cast."
- **Avatar of Fury** — "If an opponent controls seven or more lands, this spell costs {6} less to cast."
- **Avatar of Hope** — "If you have 3 or less life, this spell costs {6} less to cast."
- **Avatar of Might** — "If an opponent controls at least four more creatures than you, this spell costs {6} less to cast."
- **Avatar of Woe** — "If there are ten or more creature cards total in all graveyards, this spell costs {6} less to cast."
- **Phantasmal Extraction / Unforgiving Overtake / Unnatural Summons** — "If you weren't the starting player, this spell costs {1} less to cast."
- **Shadow of Mortality** — "If your life total is less than your starting life total, this spell costs {X} less to cast, where X is the difference."

## Current (wrong) parse

`parse_static_line("If an opponent has no cards in hand, this spell costs {6} less to cast.")` yields a `ModifyCost` static with `condition == None`. The reduction is treated as always-on, so every Avatar in the cycle costs {6} less regardless of its board condition.

## Root cause

The cost-mod condition extraction is trailing-only:

```rust
if let Some((cond_pos, marker)) = [" as long as ", " if "]
    .into_iter()
    .filter_map(|marker| lower.rfind(marker).map(|pos| (pos, marker)))
    .max_by_key(|(pos, _)| *pos)
{ ... parse_inner_condition(cond_text) ... }
```

A leading `"If …, this spell costs …"` has no preceding-space `" if "`, so nothing is extracted.

## Expected behavior

Recognize the leading `"If [condition], this spell costs {N} less to cast"` form: strip the leading `"if [condition], "` prefix, route `[condition]` through the existing `parse_inner_condition` (which already handles all the conditions above — "an opponent has no cards in hand", "you have 3 or less life", "an opponent controls at least four more creatures than you", etc.), and attach it as the static's `condition`. The cost-clause parse of the remainder is unchanged.

## Notes

Discovered while verifying #3843/#3844 (the condition-phrasing fix for "an opponent has no cards in hand" works through the static and trigger paths but not this cost-mod path). Parser-only; the conditions themselves are already supported. CR 601.2f / CR 608.2g (cost modification), CR 603.4 (intervening-if-style gate).

dripsmvcp:
Fixes #3845

## Summary
- Preserve leading `If <condition>, this spell costs ...` clauses by peeling the condition before the cost-modifier parser consumes the cost text.
- Extend shared condition parsing for the Avatar cycle shapes: opponent empty hand and typed card totals across all graveyards.
- Update the frozen anaphoric allowlist after the broadened condition parser removes `vivien's invocation` from the runtime anaphoric set.

## Verification
- `cargo fmt --all` - passed
- `./scripts/check-parser-combinators.sh` - passed
- `cargo test -p engine test_all_graveyards_typed_card_total_ge --lib` - passed
- `cargo test -p engine self_cost_reduction_leading_if --lib` - passed
- `cargo test -p engine anaphoric_scope_allowlist_guard::anaphoric_scope_set_is_frozen --test integration` - passed
- `cargo test -p engine -- --test-threads=1` - passed
- `cargo clippy --all-targets -- -D warnings` - passed before the final allowlist-only guard update; the final rerun was interrupted after starting
- `./scripts/gen-card-data.sh` - passed after network approval


matthewevans (approved):
Reviewed the updated leading cost-condition parser path and maintainer-resolved the small merge conflict. The net PR change now preserves leading cost-modifier conditions via the shared condition parser, adds coverage for the Avatar-cycle condition shapes, and leaves the current Vivien anaphor classification from main intact. Local evidence before push: cargo fmt --check, clippy, parser combinator gate, engine parser tests (447), and phase-ai lib tests (1093) passed; the pre-push hook later failed only when oracle-gen could not create a temp file under the temp worktree target after the useful checks had passed.

PR #4044, opened 2026-06-21
feat(engine): make Prowl castable + gate ETB on "if its prowl cost was paid" (#3999)

### LINKED ISSUE #3999: Latchkey Faerie — Latchkey Faerie causes its controller to draw a card when it enters, irrespective of whether the prow…
Reported in Discord: https://discord.com/channels/1485498006781427802/1517790993041068193/1517790993041068193

**Thread:** Latchkey Faerie
**Cards:** latchkey faerie

## Summary
Latchkey Faerie causes its controller to draw a card when it enters, irrespective of whether the prowl cost was paid

### latchkey faerie
Flying
Prowl {2}{U} (You may cast this for its prowl cost if you dealt combat damage to a player this turn with a Faerie or Rogue.)
When this creature enters, if its prowl cost was paid, draw a card.

---
<sub>report_id: `discord:1517790993041068193:1517790993041068193:0` · discord: `1517790993041068193/1517790993041068193`</sub>

jaso0n0818:
Fixes #3999 — Latchkey Faerie draws on ETB irrespective of whether its prowl cost was paid.

## Problem
`When this creature enters, if its prowl cost was paid, draw a card.` parsed with the intervening-if **dropped** (trigger `condition: None`), so the draw fired on every cast — even a hard-cast.

## Fix
Wires Prowl into the established `CastVariantPaid` machinery (same pattern as Sneak/Evoke/Surge/Spectacle/Escape/…):
- **`types/ability.rs`**: add `CastVariantPaid::Prowl` (CR 702.76a).
- **`game/stack.rs`**: tag `cast_variant_paid = Prowl` on a Prowl-cast permanent at resolution, mirroring the sibling cast-variant arms.
- **`parser/oracle_trigger.rs`**: map `"prowl cost was paid"` → `CastVariantPaid::Prowl` in the cast-variant intervening-if dispatch table.

The generic `TriggerCondition::CastVariantPaid` evaluator (`triggers.rs`) matches the new tag automatically (`obj.cast_variant_paid == Some((variant, turn))`).

## Tests
`crates/engine/tests/integration/issue_3999_latchkey_faerie_prowl_etb.rs`:
- the ETB parses to `CastVariantPaid { Prowl }` (not `None`);
- Latchkey **draws** when its prowl cost was paid;
- Latchkey does **not** draw when hard-cast (the reported bug).

`clippy -D warnings` clean; full `engine` test green (13023 passed; the sole red is the pre-existing order-flaky `delve_payment_skips_state_clone_per_graveyard_candidate`, which passes in isolation and is unrelated).

Model: claude-opus-4-8 (Claude Code)


matthewevans (changes requested):
**[MED]** The new Prowl stack tag is not tested through the casting pipeline. Evidence: `crates/engine/tests/integration/issue_3999_latchkey_faerie_prowl_etb.rs:36` moves Latchkey directly to the battlefield, and `crates/engine/tests/integration/issue_3999_latchkey_faerie_prowl_etb.rs:42` manually writes `cast_variant_paid`; the production code added at `crates/engine/src/game/stack.rs:1356` is never reached. Why it matters: deleting the new `stack.rs` Prowl tagging would still leave the positive draw test green, so the PR does not prove Latchkey works when it is actually cast for its prowl cost. Suggested fix: drive Latchkey through a real `runner.cast(...).casting_variant(CastingVariant::Prowl).resolve()` path after seeding the Prowl damage ledger, then assert the ETB draws only on that cast path while the hard-cast path does not draw.

jaso0n0818:
Thanks — agreed the new `stack.rs` Prowl tag must be exercised by a real cast, and digging into that surfaced a pre-existing gap worth flagging.

**Finding:** Prowl (like Surge / Spectacle / Sneak) is only reachable through the *multi-alternative* candidate path — `casting_variant_choice_set` → `handle_cast_spell` presents a choice only when `options.len() > 1`, and auto-casts only when `had_multiple_candidates` (≥2 alternatives). A spell whose *only* alternative is Prowl yields `candidates = [Prowl]`, so both gates are false and the cast falls through to the **normal** cast (I verified Latchkey's prowl eligibility computes `true` — zone/keyword/damage-ledger all pass — yet no `CastingVariantChoice` is offered and `cast_variant_paid` stays `None`). Unlike Warp/Evoke/Emerge, Prowl has no `AlternativeCastChoice` normal-vs-alternative path. So today a single-Prowl card can't be cast for its prowl cost through the standard `CastSpell` flow, which is exactly why a real-cast test can't drive `stack.rs`.

That means "if its prowl cost was paid" is only meaningful once Prowl is genuinely castable. **Proposed scope for this PR:** wire Prowl into the normal-vs-alternative cast path (mirroring the `Keyword::Warp`/`Evoke` blocks → `AlternativeCastKeyword::Prowl` + `WaitingFor::AlternativeCastChoice` + a choice handler), then the requested `runner.cast(...).casting_variant(CastingVariant::Prowl).resolve()` test drives `stack.rs` end-to-end.

Want me to expand this PR to include that casting-path wiring, or would you prefer Prowl-castability as a separate PR (keeping this one to the parser/tag change with the limitation noted)? Happy to do either — just confirming scope before the larger casting-flow change.

jaso0n0818:
Expanded this PR to make Prowl genuinely castable, so the requested real-cast test now drives `stack.rs`'s Prowl tagging end-to-end (no manual `cast_variant_paid`).

**What changed (commit cbce766):** wired Prowl into the normal-vs-alternative cast flow, mirroring the existing **Spectacle** treatment:
- `AlternativeCastKeyword::Prowl`;
- `prowl_damage_ledger_satisfied` — single authority for the CR 702.76a eligibility gate, shared by `casting_variant_candidates` and the new block;
- a normal-vs-prowl block in `handle_cast_spell` (both affordable → `AlternativeCastChoice{Prowl}`; only prowl affordable → auto-route);
- `handle_prowl_cost_choice_with_payment_mode` + the dispatch arm.

**Tests now use the real pipeline** (`crates/engine/tests/integration/issue_3999_latchkey_faerie_prowl_etb.rs`): with prowl eligible and only the prowl cost payable, casting Latchkey auto-routes through the prowl alternative cast → `stack.rs` tags `cast_variant_paid = Prowl` → the ETB intervening-if draws; a hard cast does not draw; and the parser still lowers the intervening-if to `CastVariantPaid{Prowl}`. Deleting the `stack.rs` tag now fails the positive test.

Rebased onto current `main`. `clippy -D warnings` clean; full `engine` test green (13060 lib + 1435 integration, 0 failures).

Model: claude-opus-4-8 (Claude Code)

matthewevans (approved):
Approved current head after maintainer frontend bridge fix for Prowl alternative-cast UI/types and green CI. The prior production-path test blocker is addressed by the real prowl cast-path regression, and the Rust/frontend enum drift is covered by the added modal boundary tests.

PR #4053, opened 2026-06-21
test(engine): cover Field Marshal other-Soldier anthem + first-strike grant

jaso0n0818:
Runtime regression for **Field Marshal** — "Other Soldier creatures you control get +1/+1 and have first strike."

Covers a continuous static that jointly exercises filter axes the existing anthem/grant tests don't cover *together*:
- **other** — the source itself is excluded (no self-buff / self-grant),
- **controller** — "you control" scopes to the source's controller (CR 109.4),
- **subtype** — only Soldiers qualify (CR 205.3m),
- **dual effect** — one static both pumps P/T (Layer 7c, CR 613.4c) **and** grants a keyword (Layer 6 ability-adding, CR 613.1f).

Drives the real parse → synthesis → layer pipeline and reads back the **EFFECTIVE** post-`evaluate_layers` P/T and keyword set (not AST shape):
- another Soldier you control becomes 2/2 and gains first strike;
- the source (itself a Soldier) is excluded by "other" — no self-buff, no self first strike;
- a non-Soldier and an opponent's Soldier are unaffected;
- the grant ends when the source leaves the battlefield (CR 611.3).

`cargo clippy -p engine --all-targets -- -D warnings` clean; both tests green. Test-only (no engine source change).

Model: claude-opus-4-8 (Claude Code)


matthewevans (approved):
Reviewed current head a0fea2d; current-head CI is green, the required test label is present, and the Field Marshal test remains accepted as distinct same-filter dual-modification coverage.

PR #4058, opened 2026-06-21
fix/revealed-card-type-gate-mv-suffix-unification

andriypolanski:
## Summary

Unifies revealed-card type gate parsing so mana-value and chosen-type suffixes work consistently across all conditional shapes — not just the leading-`If` path fixed for Kellan, Daring Traveler in the recent std long-tail batch.

The recent Kellan fix added `" with mana value N or less/greater"` handling to `strip_card_type_conditional` (leading `"If it's a creature card with mana value 3 or less, …"`) and wired the runtime evaluator to honor `RevealedHasCardType.additional_filter`. Two sibling parser paths were still out of sync:

- **`parse_its_a_type_condition`** — used by `strip_suffix_conditional` for trailing-if forms (`"put it into your hand if it's a creature card with mana value 3 or less"`). It only matched the last word of the type phrase, so `"less"` failed `CoreType::from_str` and the gate was silently dropped.
- **`parse_if_exiled_card_type_conditional`** — parsed `"If the exiled card is a [type] card"` but ignored mana-value suffixes on the remainder (`" with mana value 2 or less"`).

This change extracts shared building blocks and routes all three paths through them:

- `parse_revealed_card_gate_suffix` — `" with mana value …"` and `" of the chosen type"` after a card-type gate
- `parse_its_a_card_type_gate_body` — shared body for `"it's a/an … card"` gates (multi-subtype, permanent, chosen type, CMC)

## Changes

- Refactor `strip_card_type_conditional` to delegate to `parse_its_a_card_type_gate_body` (behavior preserved; duplication removed).
- Upgrade `parse_its_a_type_condition` to use the same helper and require full condition consumption (suffix-if peel).
- Extend `parse_if_exiled_card_type_conditional` with `parse_revealed_card_gate_suffix` for exiled-card demonstrative gates.
- Add three parser regression tests:
  - `strip_card_type_conditional_creature_with_mana_value_ceiling`
  - `suffix_if_creature_with_mana_value_ceiling`
  - `exiled_card_type_conditional_with_mana_value_ceiling`

## Test plan

- [x] `cargo test -p engine --lib parser::oracle_effect::conditions::tests` (72 tests, all pass)
- [x] `cargo fmt --all`
- [ ] Existing Kellan runtime regression (`kellan_daring_traveler_mana_value_gate.rs`) still passes via Tilt `test-engine`
- [ ] No parser swallow regressions on cards using `"If it's a … card"` / `"If the exiled card is a … card"` conditional chains

## CR references

- CR 202.3 — mana value on revealed/exiled card gates
- CR 205.3m — chosen-type and property suffixes on type gates
- CR 406.6 — exiled-card demonstrative reference (`last_zone_changed_ids`)
- CR 608.2c — sub-ability conditional resolution order


matthewevans (approved):
Reviewed current head b1265451e32d1051ef8a16b03a6535715f3ff557 after maintainer parser-gate fix and update-branch; current-head CI is green and the required bug label is present.

PR #4062, opened 2026-06-21
Fix Sliver Gravemother inline granted Encore SelfManaCost parsing and…

### LINKED ISSUE #3655: Encore — Encore as a mechanic, seems to work fine now.
Reported in Discord: https://discord.com/channels/1485498006781427802/1516079109749608488/1516079109749608488

**Thread:** Encore
**Cards:** sliver gravemother

## Summary
Encore as a mechanic, seems to work fine now.

### sliver gravemother
The "legend rule" doesn't apply to Slivers you control.
Each Sliver creature card in your graveyard has encore {X}, where X is its mana value.
Encore {5} ({5}, Exile this card from your graveyard: For each opponent, create a token copy that attacks that opponent this turn if able. They gain haste. Sacrifice them at the beginning of the next end step. Activate only as a sorcery.)

---
<sub>report_id: `discord:1516079109749608488:1516079109749608488:0` · discord: `1516079109749608488/1516079109749608488`</sub>

andriypolanski:
## Summary

- Parses Sliver Gravemother’s inline graveyard encore grant: `encore {X}, where X is its mana value`.
- Maps the inline `where X is its mana value` binding to `Encore(ManaCost::SelfManaCost)` so runtime synthesis concretizes the mana sub-cost per recipient card.
- Resolves `SelfManaCost` for granted **Encore** and **Scavenge** at activated-ability synthesis (matching the existing Embalm/Eternalize path).

## Issue

Closes #3655

Discord report: Encore as a mechanic works, but **Sliver Gravemother**’s static grant — each Sliver creature card in your graveyard has encore whose cost equals its mana value — did not surface activatable Encore on graveyard Slivers at the correct cost.

## Root cause

1. **Parser gap:** The graveyard keyword-grant path only handled continuation-sentence forms (Wire Surgeons / Varolz: `has encore. Its encore cost is equal to its mana cost.`). The inline same-line form on Sliver Gravemother was not recognized.
2. **Runtime gap:** `resolve_self_cost_graveyard_activated_keyword()` resolved `SelfManaCost` for Embalm/Eternalize only. Granted Encore/Scavenge with `SelfManaCost` would synthesize abilities with a free mana placeholder instead of the card’s own mana cost.

## Rules references

- **CR 702.141a** — Encore is an activated ability that functions only from the graveyard; the mana cost is part of the ability’s activation cost.
- **CR 604.1** — Activated abilities are synthesized from keyword grants on off-zone cards via the continuous static / off-zone keyword path.
- **CR 202.3** — Mana value of the recipient card drives the inline `where X is its mana value` binding.

## Verification

| Command | Result |
|---------|--------|
| `cargo fmt --all` | **PASS** (exit 0) |
| `cargo clippy-strict` | **NOT RUN** — MSVC `link.exe` not found on this Windows host; Tilt unavailable |
| `cargo test -p engine` | **NOT RUN** — build blocked (`link.exe` not found) |
| `cargo test-all` | **NOT RUN** — build blocked |
| `git diff --check` | **NOT RUN** |

CI on the PR is expected to provide the full verification matrix (`test-engine`, `clippy`).

## Files changed

| File | Change |
|------|--------|
| `crates/engine/src/parser/oracle_static/keyword_grant.rs` | Inline graveyard encore/scavenge grant parser + SelfManaCost normalization |
| `crates/engine/src/parser/oracle.rs` | Dispatch inline grants; extend scavenge/encore grant test |
| `crates/engine/src/game/keywords.rs` | Resolve SelfManaCost for granted Encore/Scavenge |
| `crates/engine/src/game/casting.rs` | Unit test for granted SelfManaCost encore synthesis |
| `crates/engine/src/parser/oracle_static/tests.rs` | Parser unit tests |
| `crates/engine/tests/integration/issue_3655_sliver_gravemother_encore.rs` | End-to-end integration tests |


matthewevans (changes requested):
**[HIGH]** The submitted head does not compile. Evidence: GitHub Rust lint/test logs for head `86e38c50144f2631b41233d182fa020204f2675c` report `E0515` at `crates/engine/src/parser/oracle_static/keyword_grant.rs:63`, `E0507` at `crates/engine/src/parser/oracle_static/keyword_grant.rs:103` and `:106`, and `E0308` at `crates/engine/src/parser/oracle_static/tests.rs:7862`. Why it matters: every Rust/WASM/card-data job fails before the Sliver Gravemother behavior can be exercised. Suggested fix: avoid returning a borrowed `keyword_text` from the local `stripped` value, borrow or restructure the `where_x` match guards, and update all `try_parse_graveyard_keyword_grant_clause` destructures to the new 3-tuple shape.

**[LOW]** The new inline grant parser uses substring dispatch to reject continuation sentences. Evidence: `crates/engine/src/parser/oracle_static/keyword_grant.rs:119`-`:123` lowercases the line and returns `None` on `lower.contains(". ")`, while the neighboring continuation path in `crates/engine/src/parser/oracle.rs:760`-`:769` already uses a structured split/parser for period-separated grants. Why it matters: this adds an ad hoc parser-routing predicate in a nom-combinator parser surface and makes the new path depend on punctuation scanning instead of grammar shape. Suggested fix: route sentence splitting through the existing `split_once_on_lower`/continuation parser boundary, or make the inline parser consume exactly the one-sentence grammar without a `contains` prefilter.

matthewevans (changes requested):
**[HIGH]** The `fix: retest` head still does not compile. Evidence: current head `aa4000ca95c4221fb3b12ade3b6d17d45d65cf10` keeps `try_parse_graveyard_keyword_grant_clause` returning `Option<(TargetFilter, GraveyardGrantedKeywordKind, String)>` at `crates/engine/src/parser/oracle_static/keyword_grant.rs:30`-`:32`, but `crates/engine/src/parser/oracle_static/tests.rs:7862` still destructures the result as `let (filter, kind) = ...`. Why it matters: this leaves the same class of compile failure from the prior review, so CI cannot reach the Sliver Gravemother behavior tests. Suggested fix: update that test destructure to include or ignore the returned keyword text, then let the current-head CI prove the rest of the parser/runtime changes.

The earlier parser-routing concern also remains: `try_parse_graveyard_keyword_grant_static` still uses `lower.contains(". ")` as a dispatch prefilter at `keyword_grant.rs:119`-`:123`.

matthewevans (changes requested):
The prior tuple-destructure and period-dispatch blockers are addressed on this head, but the current head still does not compile.

**[HIGH] Current head fails to compile.** GitHub WASM and card-data jobs for head `49907096970e113f74f0915df2b67f7c30982f1e` both report `E0382` at `crates/engine/src/parser/oracle_static/keyword_grant.rs:113`: `normalize_graveyard_granted_keyword(keyword, where_x, kind).unwrap_or(keyword)` moves `keyword` into the normalizer and then tries to use it again in `unwrap_or(keyword)`. CI cannot reach the Sliver Gravemother parser/runtime coverage on this head. Keep ownership explicit here, for example by matching on a cloned/borrowed normalization attempt or changing the helper shape so the fallback does not reuse a moved value.

matthewevans (changes requested):
**[HIGH]** Sliver Gravemother's `{X}, where X is its mana value` is implemented and tested as the recipient's full mana cost. Evidence: `crates/engine/src/parser/oracle_static/keyword_grant.rs:81` maps recipient mana-value bindings to `ManaCost::SelfManaCost`, and `crates/engine/tests/integration/issue_3655_sliver_gravemother_encore.rs:96` asserts the granted Encore mana sub-cost equals `{2}{R}{G}`. Why it matters: CR 202.3 defines mana value as a number regardless of color, while CR 202.1a mana cost requires matching colored symbols; a `{2}{R}{G}` Sliver should get Encore `{4}`, not Encore `{2}{R}{G}`, so this ships the fixed card with a test that protects the wrong behavior. Suggested fix: model this as a distinct self-mana-value generic cost, e.g. resolve the recipient's `mana_cost.mana_value()` to `ManaCost::generic(...)`, while keeping Wire Surgeons/Varolz `its mana cost` on the existing `SelfManaCost` path.

matthewevans (approved):
Approved on the current retest head after re-review. The prior mana-value blocker is addressed: the inline Sliver Gravemother grant lowers to SelfManaValue, runtime synthesis concretizes it to generic mana value, the production-path legal-actions regression covers the {4} cost, CI is green, the required bug label is present, and the repush guard held.

PR #4065, opened 2026-06-21
fix(parser): strip ability-word prefix on token-granted abilities (Ka-Zar of the Savage Land)

lgray:

## Card
**Ka-Zar of the Savage Land** (MSH). The residual gap on `main` was `Unimplemented[landfall]: "Landfall — Whenever a land you control enters"` — the **token's granted ability** inside the Zabu `CreateToken` effect. The first two lines (look at top of library, play lands from top) already parsed.

## Trace + composition
- `Effect::Token` carries a token's granted abilities only in `static_abilities: Vec<StaticDefinition>`. There is no separate triggered-ability field — a triggered grant rides as a `StaticDefinition` whose modification is `ContinuousModification::GrantTrigger`.
- Both the parse-time token path (`push_parsed_statics` → `parse_quoted_ability_modifications`) and the runtime catalog path (`inject_catalog_token_abilities`) funnel quoted ability text through the single-authority classifier `classify_quoted_inner` (`oracle_static/keyword_grant.rs`).
- `classify_quoted_inner` already routes `when `/`whenever `/`at ` prefixes to `GrantTrigger` (CR 603.1). The gap: the italicized `Landfall —` ability-word prefix (CR 207.2c) sat in front of `Whenever`, so the trigger check failed and the line fell through to the `GrantAbility` catch-all, producing `Effect::Unimplemented[landfall]`.
- Runtime already supports the fixed shape: the layers pass installs a `GrantTrigger` static modification as a live `trigger_definition` (CR 604.2, `layers.rs`). No runtime change required — the fix is **parse-time classification only**.

**Fix:** at the top of `classify_quoted_inner`, strip a recognized ability-word em-dash prefix via the canonical nom-combinator helper `strip_ability_word_with_name` (built on `split_short_label_prefix`) and re-classify the remainder. Gated on a known ability word (new `is_known_ability_word` predicate over the existing `ABILITY_WORD_NAMES` table) so a legitimate em-dash body is untouched.

## Class built (not the card)
Any token created with an **ability-word-prefixed granted ability** — `with '<AbilityWord> — Whenever/When/At ...'` — now classifies correctly. Reuses existing `ContinuousModification::GrantTrigger` + `Effect::Token.static_abilities`; covers the trigger, keyword, and static body shapes via the recursive re-dispatch.

## add-engine-variant verdict
**N/A — no new variant.** Reused `ContinuousModification::GrantTrigger` and the existing `Effect::Token.static_abilities` mechanism. `data/engine-inventory.json` untouched. Additions are one small predicate (`is_known_ability_word`) and a recursive strip in the existing classifier — parameterization/reuse, not proliferation.

## Grep-verified CRs
- **CR 207.2c** — ability words have no rules meaning; body parses through ordinary machinery (lists "landfall" explicitly).
- **CR 603.1** — triggered ability "[When/Whenever/At] [event], [effect]".
- **CR 603.6a** — enters-the-battlefield triggers ("Whenever a [type] enters").
- **CR 604.2** — static abilities create continuous effects active while the permanent is on the battlefield (justifies the layers flush before the trigger is live).

All four verified against `docs/MagicCompRules.txt`; diff-gate grep reports zero UNVERIFIED.

## Discriminating-test map (parse → resolve)
| Behavioral claim | Test | Revert-failing assertion |
|---|---|---|
| Ka-Zar's token line parses landfall as `GrantTrigger`, not `Unimplemented` | `zabu_token_landfall_trigger_parses_as_grant_trigger` | finds a `GrantTrigger`; no `Unimplemented` |
| Catalog `rules_text` path classifies the same | `catalog_landfall_rules_text_classifies_as_grant_trigger` | `GrantTrigger` present |
| Full card → 0 residual `Unimplemented` | `kazar_full_card_no_residual_unimplemented` | debug contains no `Unimplemented` |
| Controller's land ETB puts +1/+1 on **Zabu** | `zabu_landfall_puts_counter_on_zabu_for_controllers_land` | Zabu counters `1` (flips to `0` on revert) |
| Opponent's land does **not** trigger | `zabu_landfall_ignores_opponents_land` | Zabu counters `0` |
| Counter lands on Zabu, **not** Ka-Zar | `zabu_landfall_counter_targets_zabu_not_kazar` | Zabu `1`, Ka-Zar `0` (Zabu flips to `0` on revert) |

Runtime tests drive the real pipeline: parse → `resolve()` → `flush_layers` → `process_triggers(ZoneChanged)` → `resolve_top`. **Revert probe:** disabling the strip fails 5/6 tests (the residual `Unimplemented` reappears and the positive counter assertions flip 1 → 0).

## Gate results
- `cargo fmt --all`: clean
- `./scripts/check-parser-combinators.sh`: **0 Forbidden** (inline string-dispatch grep clean; three test-only `.contains("Unimplemented")` assertions annotated `allow-noncombinator`)
- `./scripts/check-engine-authorities.sh upstream/main`: exit 0
- `cargo test -p engine`: `13075 passed; 1 failed` — the single failure (`ai_support::...delve_payment_skips_state_clone_per_graveyard_candidate`, a state-clone-count test) is a known parallel-flaky unrelated to this diff and **passes in isolation**.
- `cargo semantic-audit`: exit 0, no Ka-Zar/Zabu/landfall findings introduced
- `cargo coverage`: exit 0; regenerated card-data shows Ka-Zar of the Savage Land parses with **no `Unimplemented`**
- card-data regenerated via `oracle-gen data` (no token step); `known-tokens.toml` untouched

### DIFF
--- a/crates/engine/src/game/effects/token.rs
+++ b/crates/engine/src/game/effects/token.rs
@@ -5496,4 +5496,166 @@ mod tests {
             "static buff line must classify to something"
         );
     }
+
+    // ── Ka-Zar / Zabu landfall: parse → resolve → trigger ────────────────
+
+    /// Parse Ka-Zar's ETB token line into a real `Effect::Token` (so the test
+    /// exercises the actual parser output, not a hand-built trigger), wrapped in
+    /// a `ResolvedAbility` controlled by `controller`.
+    fn kazar_token_ability(controller: PlayerId) -> ResolvedAbility {
+        let txt = "Create Zabu, a legendary 2/2 green Cat creature token with \"Landfall — Whenever a land you control enters, put a +1/+1 counter on Zabu.\"";
+        let effect = crate::parser::oracle_effect::token::try_parse_token(
+            &txt.to_lowercase(),
+            txt,
+            &mut crate::parser::oracle_ir::context::ParseContext::default(),
+        )
+        .expect("Ka-Zar token line must parse");
+        ResolvedAbility::new(effect, vec![], ObjectId(500), controller)
+    }
+
+    /// Resolve Ka-Zar's token effect and return the created Zabu's `ObjectId`.
+    fn create_zabu(state: &mut GameState, controller: PlayerId) -> ObjectId {
+        let ability = kazar_token_ability(controller);
+        let mut events = Vec::new();
+        resolve(state, &ability, &mut events).unwrap();
+        // CR 604.2: run the layers pass so the token's `GrantTrigger` static
+        // modification is installed as a live trigger_definition before any land
+        // ETB is processed.
+        crate::game::layers::flush_layers(state);
+        *state
+            .battlefield
+            .iter()
+            .find(|id| {
+                state
+                    .objects
+                    .get(id)
+                    .is_some_and(|o| o.is_token && o.name == "Zabu")
+            })
+            .expect("Zabu token must be on the battlefield")
+    }
+
+    /// Put a land onto the battlefield under `land_controller` and fire its ETB
+    /// event through the real trigger pipeline, then resolve the stack.
+    fn land_enters(state: &mut GameState, land_controller: PlayerId, card_id: u64) {
+        let land = create_object(
+            state,
+            CardId(card_id),
+            land_controller,
+            "Forest".to_string(),
+            Zone::Battlefield,
+        );
+        {
+            let obj = state.objects.get_mut(&land).unwrap();
+            obj.card_types.core_types.push(CoreType::Land);
+            obj.controller = land_controller;
+            obj.owner = land_controller;
+        }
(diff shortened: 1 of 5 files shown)


matthewevans (approved):
Reviewed current head 4431f1945cac67d118739a4ecbbd76495ddd208d via GitHub API diff and isolated worktree. The fix is at the correct parser seam: recognized CR 207.2c ability-word prefixes are stripped before dispatching quoted token-granted abilities through the existing trigger/static classifier, so Ka-Zar/Zabu lowers to GrantTrigger instead of the unimplemented fallback. CR 207.2c/603.1/603.6a/604.2 annotations match the local CR text, CI is green, and the added tests cover parser classification plus the live layer/trigger path.

PR #4067, opened 2026-06-21
feat: Spider-Woman, Secret Agent — duration-bound can't-untap on a chosen target (MSH)

lgray:

## Card
**Spider-Woman, Secret Agent** (MSH):
> Flash
> When Spider-Woman enters, tap target creature an opponent controls. That creature can't become untapped for as long as you control Spider-Woman.

Closes the residual gap on main: `Unimplemented[can't]: "can't become untapped"` on the ETB sub-ability.

## The class built
**Duration-bound (you-control-source) BROAD can't-untap on a chosen target** — not a one-off. Any future "tap target X, it can't become untapped for as long as you control ~" card reuses this path. Composed entirely from existing primitives:

- **Broad untap prohibition** = `Effect::AddTargetReplacement` carrying a bare `ReplacementEvent::Untap` prevention (`valid_card: SelfRef`, no `execute`). This is the BROAD prohibition (CR 701.26b) that blocks *every* untap path — the controller's untap step AND effect untaps ("untap target creature") — mirroring the merged Blossombind static. `StaticMode::CantUntap` was deliberately NOT reused: it only blocks the untap step (exert/freeze semantics) and `process_one_untap` never consults it.
- **Anaphoric target binding** = `target: TargetFilter::Any` on the rider, inheriting the parent tap clause's chosen creature via the existing sub-ability target propagation in `resolve_ability_chain`. Same shape as the `try_parse_die_exile_rider` "that creature ... instead" class.
- **CR 611.2b duration** = the clause shell already peels "for as long as you control ~" onto the sub-ability frame as `Duration::UntilHostLeavesPlay`. Because the replacement is installed on a *different* object (the chosen creature), the `AddTargetReplacement` install chokepoint translates that duration into a new `ReplacementCondition::ControllerControlsSource { source, controller }` gate, stamped with the originating source (Spider-Woman) and its controller. The gate re-checks "you still control the source" on every untap and lapses on a control swap or the source leaving play — the Master Thief reading. Tightly scoped (bare untap-prevention riders only) so no other `AddTargetReplacement` install is affected.

All new parser code is nom combinators; no `find`/`split_once`/`contains`/`starts_with` for dispatch.

## add-engine-variant verdict
One new variant: `ReplacementCondition::ControllerControlsSource { source: ObjectId, controller: PlayerId }`.

- **Stage 1 (existence):** no existing condition gates on continued control of one *specific captured object*. `EventSourceControlledBy` checks the event's source; `StaticCondition::SourceControllerEquals` is on a different enum, not reachable from `ReplacementCondition`, and an object-installed replacement's `rid.source` is the host (the chosen creature), not the originating source. **DOES_NOT_EXIST.**
- **Stage 2 (parameterization):** the `Unless*Controls*` / `IfControlsMatching` family gates on a *count of permanents matching a filter* — a different axis (CR 614.1 quantity) from "continued control of one captured object" (CR 611.2b duration). Not a sibling-cluster smell. **EXTEND_OK.**
- **Stage 3 (categorical boundary):** lies squarely within CR 611.2b. Single section.

## Grep-verified CRs
All verified against `docs/MagicCompRules.txt`:
- **CR 701.26b** — "To untap a permanent, rotate it back... Only tapped permanents can be untapped." (untap keyword action)
- **CR 611.2b** — "for as long as ..." continuous-effect duration; the Master Thief example matches Spider-Woman exactly.
- **CR 614.6** — "If an event is replaced, it never happens." (bare prevention)
- **CR 614.1 / 614.1c** — replacement applicability gates.

## Discriminating-test map
| Behavioral claim | Seam | Production entry | Test | Revert-failing assertion |
|---|---|---|---|---|
| ETB taps opponent's creature | `SetTapState` resolve | `resolve_ability_chain` (parsed ETB chain) | `spider_woman_secret_agent_cant_untap_for_as_long_as_controlled` | `foe_creature.tapped` |
| Effect untap prevented while controlled | `process_one_untap` -> `replace_event` | `resolve_set_tap_state` Single/Untap | same | `foe_creature.tapped` after untap; no `PermanentUntapped` |
| Stays tapped through controller's untap step | `execute_untap` -> `replace_event` | `execute_untap` | same | `foe_creature.tapped` after untap step |
| Lapses when source leaves play (CR 611.2b) | `ControllerControlsSource` eval | `move_to_zone` + `resolve_set_tap_state` | same | `!foe_creature.tapped` (the assertion that flips on revert) |
| Lapses on control swap (CR 611.2b) | `ControllerControlsSource` eval | controller mutation + untap | `spider_woman_cant_untap_lapses_on_control_swap` | `!foe_creature.tapped` |
| 0 residual Unimplemented; rider is `AddTargetReplacement`; Flash extracted | parser lowering | `parse_oracle_text` (Flash as MTGJSON keyword) | `spider_woman_secret_agent_parses_with_no_unimplemented` | no Unimplemented in chain; `AddTargetReplacement` present |

**Revert-probe (run):** commenting out `stamp_for_as_long_as_controlled_gate` makes the installed replacement permanent (no gate); both runtime tests FAIL on the "must lapse once we no longer control Spider-Woman (CR 611.2b)" assertions. Restoring the stamp → green. This proves the tests discriminate the duration fix, and they drive the real untap pipeline (not shape-only).

## Gate results
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` (default and vs `upstream/main`) — exit 0
- parser diff string-dispatch grep gate — clean
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` — 0 diagnostics
- `cargo test -p engine` — `13070 passed; 2 failed` where the only 2 failures are the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` and `target_selection_legal_actions_do_not_simulate_each_target` (both confirmed PASS in isolation).
- `cargo coverage` / `cargo semantic-audit` — exit 0; Spider-Woman has **zero** semantic-audit findings.
- Production card-data (regenerated): Spider-Woman entry has **0 `Unimplemented`**, **1 `AddTargetReplacement`**, keyword `Flash`.

CR annotations all grep-verified; no `UNVERIFIED`.


lgray:

**Root cause (CI red):** `cargo fmt --all -- --check` failed in the lint job — the new test in `oracle_replacement.rs` had a `parsed.triggers.first().expect(...)` chain that rustfmt collapses to a single line. The other red check ("Rust (fmt, clippy, test, coverage-gate)") is just the aggregator that fails when lint fails; there was no clippy or test failure underneath it.

**Fix:** Applied `cargo fmt --all` (the one-line collapse). No logic changed.

**Also rebased** onto current `upstream/main` (was 4 commits behind; clean rebase, no conflicts).

**Verification (nightly proxy):**
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` with and without `upstream/main` base — exit 0
- diff string-dispatch grep — clean (only hit is a `Vec::contains` membership check in a test assertion)
- `cargo test -p engine` — 13081 passed; only failure is the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate`, which passes in isolation.

I also grep-verified every CR citation in the diff against `docs/MagicCompRules.txt`: CR 611.2b (for-as-long-as continuous effects, Master Thief example), CR 701.26b (untap a permanent), CR 614.6 (event replacement) — all accurate.

New head: `cc3a78ab43464102b4fc9d994b623410755d00a0`

matthewevans (changes requested):
Found one rules-correctness blocker.

`ReplacementCondition::ControllerControlsSource` currently evaluates as a live predicate: source is on the battlefield and controlled by the installing player. That is enough for the first lapse, but it does not permanently expire the installed replacement. The replacement is appended to the chosen target at `crates/engine/src/game/effects/add_target_replacement.rs:172-175`, and the candidate scan only skips it while `evaluate_replacement_condition` returns false at `crates/engine/src/game/replacement.rs:4108-4118`. Because the condition itself is just `obj.zone == Zone::Battlefield && obj.controller == installer` at `crates/engine/src/game/replacement.rs:3813-3819`, the old lock can start applying again if control of Spider-Woman returns to the original controller after being stolen, or if the source leaves and later returns with the same storage `ObjectId`. The latter is especially plausible here because `move_to_zone` documents that `ObjectId` is storage identity and persists across zone changes (`crates/engine/src/game/zones.rs:133-139`, `1464-1469`), while the move path simply sets the same object back to `Zone::Battlefield` and runs battlefield-entry reset (`crates/engine/src/game/zones.rs:672-677`).

That is not CR 611.2b behavior. Once a “for as long as you control [source]” duration ends, the continuous effect has ended; it should not revive later just because the same source is controlled again. The added tests cover losing control/leaving once, but they do not cover regaining control or a leave-and-return with the same storage id. Please add a permanent expiry/latch or prune the installed replacement when the source/control condition ends, and cover both non-revival cases.

Evidence checked: CR 611.2b in local `docs/MagicCompRules.txt`; code paths above in this PR worktree. Confidence: high. What would contradict this: an existing cleanup path that removes this host-installed replacement when the source changes controller/leaves, but I found only the per-event condition skip and no such prune for `ControllerControlsSource`.

lgray:

Thanks for the catch, @matthewevans — this was a real CR 611.2b correctness blocker. Addressed and rebased onto current `upstream/main`.

**Root cause.** `ReplacementCondition::ControllerControlsSource` was a *live re-evaluated* predicate (`source on battlefield && source.controller == installer`). That suppresses application while the gate is false, but it does not *latch* the lapse — so the "can't become untapped for as long as you control ~" lock revived if control of Spider-Woman returned, or if the source left and re-entered with the same storage `ObjectId`. Per CR 611.2b a "for as long as you control X" continuous effect ends **permanently** the first time the condition lapses ("it doesn't last forever").

**Two further issues surfaced during review and were also fixed:**
- The PR installed the gated replacement only into the host's *live* `replacement_definitions`, never `base`. Because `evaluate_layers` rebuilds live from base every pass (`layers.rs`), the lock actually evaporated on the **first layer pass** even with no control/zone change. (The original tests passed only because they never ran a layer pass.)
- Durably storing the gated def in `base_replacement_definitions` then leaked it into **copiable values** (CR 707.2) — a copy/token of the locked host would inherit the lock.

**Fix.**
- Single gate authority `controller_controls_source_gate`, shared by the condition arm and the prunes.
- **Durable install:** push the gated replacement to the host's base store so it survives the per-pass live-from-base reset — a documented, prune-bounded exception to the CR 613.1 printed-baseline contract.
- **Permanent expiry** via three prunes that remove the gated def from base+live: source control swap (`prune_lapsed_controller_controls_source`, in `evaluate_layers` after the layer recursion guard; CR 611.2b + 613.1b), and source-leave **or** host-leave (`prune_controller_controls_source_on_leave`, from `apply_zone_exit_cleanup`; CR 611.2b + 400.7 — the host-leave axis covers a blinked host re-entering as a new object).
- **Copy correctness:** exclude `ControllerControlsSource`-gated runtime defs from copiable values (`intrinsic_copiable_values` + `merged_copiable_values`) via a shared predicate (CR 707.2).
- Registered `ControllerControlsSource` in `engine-inventory.json`.

**Tests** (`tap_untap.rs`): durability across a layer pass; no-revive on a real GainControl swap-back; no-revive on source same-`ObjectId` re-entry; no-revive on host blink; copiable-values exclusion. Each was verified **discriminating** via revert-probe (fails with the corresponding production hunk reverted, passes restored).

Presence-bound "remains on the battlefield" phrasing currently collapses to the same `Duration::UntilHostLeavesPlay` as the control phrasing; the stamp documents this as an honest gap (it would over-gate that sub-class on a source control swap) rather than re-routing the duration, since that mapping is relied on by several shipped card classes. No card today combines the presence phrasing with a bare untap-prevention rider.

Verification (nightly): `cargo fmt --all` clean; `cargo test -p engine` 0 failures (13155 lib passed); `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` clean; parser-combinator + engine-authorities gates exit 0. Every cited CR (611.2b, 613.1/613.1b, 400.7, 707.2/707.2b) grep-verified against `docs/MagicCompRules.txt`.

Rebased onto `upstream/main` (`d6751def3`). New head: `ac6a9307198b3109dc7fe8832b94fc56a5f5a9cf`.

matthewevans (changes requested):
Found one remaining rules blocker on the follow-up.

`ControllerControlsSource` still treats a phased-out source as satisfying the duration gate. The shared gate only checks that the captured source object is in `Zone::Battlefield` and still controlled by the installer (`crates/engine/src/game/replacement.rs:3391-3394`). But phase-out does not change zone/control; it flips `phase_status` and marks layers dirty (`crates/engine/src/game/phasing.rs:70-121`). The next full layer pass calls the new prune, but the gate still returns true for a phased-out Spider-Woman, so the host's base/live replacement is not pruned. When Spider-Woman phases back in, the lock is still installed and can keep applying.

That conflicts with CR 702.26b/d: a phased-out permanent is treated as though it does not exist and not under its controller. A “for as long as you control Spider-Woman” duration should lapse once the source phases out, and it must not revive on phase-in. Please make the gate require phased-in battlefield presence (for example `!obj.is_phased_out()` or the shared functioning predicate) and add a phase-out -> phase-in non-revival regression.

Prior blocker status: the control-swap and source/host leave/re-entry cases are materially addressed by the new prune paths; this is the remaining adjacent duration boundary.

Evidence checked: GitHub API/current PR worktree, `replacement.rs`, `layers.rs`, `phasing.rs`, and local CR text (`docs/MagicCompRules.txt` 702.26b/d/e). Confidence: high. What would contradict this: a phasing-specific cleanup path that removes `ControllerControlsSource` definitions on `PermanentPhasedOut`; I found only layer dirty marking, and the shared gate itself does not inspect `phase_status`.

VERDICT: request-changes

lgray:

Thanks @matthewevans — addressed the phase-out blocker, and rebased to clear a merge conflict against base.

**Merge conflict (cause + resolution).** The branch fell behind `upstream/main`, which had added a new pre-dispatch arm in `crates/engine/src/parser/oracle_effect/mod.rs` (`try_parse_cast_this_way_enters_with_counter`) at the exact anchor where this PR adds its own arm (`try_parse_cant_become_untapped_target_rider`). Both are independent `if let Some(effect) = … { return … }` guards with no interaction. Resolved by keeping **both** arms sequentially (upstream's first, this PR's second). The `data/engine-inventory.json` change rebased cleanly (this PR's own `ControllerControlsSource` entry only — no regen). Rebased onto `upstream/main` = `85d5e1003`.

**New review fix (CR 702.26f phase-out lapse).** You were right: `controller_controls_source_gate` (`crates/engine/src/game/replacement.rs`) only checked `zone == Battlefield && controller == installer`. Phasing flips `phase_status` and changes neither zone nor controller (CR 702.26d), so a phased-out Spider-Woman kept the gate true, the layer-pass prune never dropped the gated def, and the lock survived phase-out and revived on phase-in.

Fix: added `&& o.is_phased_in()` to the **single gate authority**, so both consumers — the live `ControllerControlsSource` condition arm and the layer-pass lapse prune `prune_lapsed_controller_controls_source` — treat phase-out as a lapse. On phase-out the gate goes false; the next `evaluate_layers` (phase-out calls `mark_layers_full`) runs the prune, which removes the gated def from **base + live**, permanently latching the expiry so it cannot revive on phase-in.

The governing rule is **CR 702.26f**: "effects with 'for as long as' durations that track that permanent … end when that permanent phases out because they can no longer see it" — combined with CR 611.2b (ends and does not begin again) and CR 702.26b/d. All five CRs grep-verified against `docs/MagicCompRules.txt` (702.26f line 4170, 702.26b 4160, 702.26d 4166, 611.2b 2902, 613.1b 2958).

**Regression test.** New `spider_woman_cant_untap_dead_on_phase_out_no_revive` in `tap_untap.rs`, driving the production phasing helpers (`phase_out_object`/`phase_in_object`): negative baseline (lock HOLDS while phased in) → phase the source out (asserting zone/controller unchanged per CR 702.26d) → assert the lock LAPSES → phase back in → assert it does NOT revive. The no-revive guarantee is proven by the prune having physically removed the def (the gate is true again after phase-in, so the foe untapping shows the lock is gone, not merely suppressed). Verified **discriminating** via revert-probe: reverting the `&& o.is_phased_in()` conjunct makes the test fail at the lapse assertion; restored, it passes. Full Spider-Woman cluster: 8 passed (prior control-swap / source-reentry / host-blink / layer-pass-durability / copiable-values fixes intact). `cargo fmt --all` clean; `clippy --workspace -D warnings` clean.

New head: `a218405ead2eeb42e07783f20eda2ecd7b61ccf6`.

matthewevans:
VERDICT: pending-CI

I re-reviewed the latest Spider-Woman follow-ups after the previous `ControllerControlsSource` blockers.

The prior issues appear addressed on the updated branch:
- The applicability gate now requires the captured source to be on the battlefield, controlled by the installing player, and phased in (`controller_controls_source_gate`). That closes the phase-out/revive hole from the last review.
- Layer evaluation prunes lapsed `ControllerControlsSource` definitions from both live and base replacement lists after Layer 2 control is finalized, so a control swap cannot revive on swap-back.
- Battlefield exit pruning removes the gated definition when either the source or the locked host leaves, so same-ObjectId re-entry does not resurrect the lock.
- Regression coverage now includes real Spider-Woman parsing/resolution, control swap/no-revive, source leave/re-entry, host blink, and source phase-out -> phase-in no-revival.

I also checked the acknowledged "remains on the battlefield" subclass concern against the current integration card fixture and did not find a real bare untap-prevention card using that presence phrasing. Since this PR is targeting the real Spider-Woman control-duration pattern, I am not treating that hypothetical subclass as a blocker here.

I ran a guarded update-branch with expected head `71be06925eef585405d2f06924e0e404b7f94ddb`; GitHub advanced the PR to `79c0c0797089cc65eb37b8520213e10232ebb77e`. CI is now running on that head, so I am leaving the formal approval/enqueue step pending current green checks.

matthewevans (approved):
Approved after re-review of the controller/source/phasing duration fixes, green CI, and repush guard on 79c0c0797089cc65eb37b8520213e10232ebb77e.

PR #4068, opened 2026-06-21
fix(parser): recognize negated modals in trigger boundary type-continuation guard

dripsmvcp:
## Summary
- The trigger condition/effect boundary splitter (`find_effect_boundary` → `continues_player_action_list` → `is_new_sentence_not_type_continuation`) misclassified effect bodies starting with a type word followed by a negated modal verb (e.g. "creatures you control **can't** be the targets ...") as type-list continuations rather than new subject-predicate sentences.
- This caused the entire trigger text to be consumed as the condition with an empty effect body, producing `execute: null` for a class of cards including Veilstone Amulet, and any other trigger whose effect starts with `<type-word> <subject> can't/don't/won't <predicate>`.

## Root Cause
`is_new_sentence_not_type_continuation` checked each word against `PREDICATE_VERBS` via `normalize_verb_token`, but contractions like "can't" normalize to "can't" (the apostrophe is interior, not stripped by `trim_matches`). Since "can't" ≠ "can", the check never matched a predicate verb, so the function returned `false` ("not a new sentence") → the boundary splitter concluded this was a type-list continuation.

## Fix
Split on apostrophe and check the base form: "can't" → "can" which IS in `PREDICATE_VERBS` → correctly classifies as a predicate sentence. This handles all `'t`-contracted negations (can't, don't, doesn't, won't, isn't, aren't) generically.

## Verification
- `cargo fmt --all` ✓
- `cargo test -p engine --lib` — 13,080 passed, 0 failed ✓
- Targeted test `trigger_veilstone_amulet_cant_be_targets` confirms the trigger now produces a `GenericEffect` (Hexproof grant) instead of `execute: null`

### DIFF
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -5506,14 +5506,34 @@ fn is_new_sentence_not_type_continuation(text: &str) -> bool {
         // negation contraction so "creatures you control can't be the targets ..."
         // is correctly classified as a new subject-predicate sentence rather than
         // a type-list continuation.
-        if let Some(base) = w.split('\'').next() {
-            if base.len() < w.len() {
-                let base_normalized = normalize_verb_token(base);
-                return PREDICATE_VERBS.contains(&base_normalized.as_str());
-            }
+        if is_negated_auxiliary_predicate_token(w) {
+            return true;
         }
         false
     })
+}
+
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved current head after maintainer hardening of the negated-modal boundary helper and green CI. The bug label is present, and the regression covers the claimed contraction class without leaving the previous don't/doesn't/won't/isn't/aren't gap.

PR #4069, opened 2026-06-21
feat(parser): implement Call Damage Control modal graveyard-return (MSH)

lgray:

## Card
**Call Damage Control** (MSH, green sorcery)
> Choose up to two. Return those cards from your graveyard to your hand.
> • Target artifact card.
> • Target creature card.
> • Target enchantment card.
> • Target land card.

Residual gap on `main`: four `Unimplemented[target]` modes; the shared return-to-hand effect was dropped entirely.

## Class built (not the card)
This card phrases its shared effect **once in the header** and leaves each bullet as a **bare target** that varies only by card-type. New `distribute_shared_mode_effect` (in `oracle_modal.rs`) detects a modal whose header carries a shared `those <noun>` effect sentence and whose every bullet is a bare target, then distributes the header effect into each mode body — replacing the `those <noun>` anaphor slot with that mode's target phrase. Each chosen mode then lowers to an independent `Effect::Bounce` of a card of its card-type from the controller's graveyard to hand.

The **card-type is the only per-mode axis**, so this covers the whole class of `Choose up to N. <verb> those <noun> .... • Target A. • Target B.` spells, not just this card. Composes existing primitives: modal subsystem (`ModalChoice` + per-mode `AbilityDefinition`), `Effect::Bounce` (return-to-hand), and the target-card-in-graveyard `TargetFilter::Typed { controller: You, InZone: Graveyard }` — all reused, **no new effect/filter variant**.

## add-engine-variant verdict
**No new variant.** Reuses `Effect::Bounce`, `TargetFilter::Typed`, and `ModalChoice`. `data/engine-inventory.json` unchanged.

## Trace / composition
- Modal AST recognized correctly on `main` (`min=0, max=2, mode_count=4`); each mode body was a bare target → unimplemented marker.
- `Return target <type> card from your graveyard to your hand` already parses to `Bounce { Typed[<type>], controller: You, InZone Graveyard, destination: None }` (default destination = hand). Distribution feeds each rewritten mode body through the existing `parse_effect_chain_with_context`.
- Runtime resolves each chosen mode as its own `ResolvedAbility` via `build_chained_resolved`, so each mode returns its own targeted card (CR 700.2c).

## Grep-verified CRs (`docs/MagicCompRules.txt`)
- **CR 700.2** — modal spell (bulleted list + choose-count)
- **CR 700.2a** — controller chooses; an illegal mode (no legal target) can't be chosen
- **CR 700.2c** — targets chosen only if that mode is chosen
- **CR 601.2c** — each `target` is a separate target
- **CR 404.1** — graveyard is the source zone
- **CR 305** — land card type (`as_land` test helper)

## Discriminating-test map (parse → cast → resolve)
Integration (`call_damage_control_modal_return.rs`):
- `two_modes_return_their_two_cards` — choose modes 0+1, target the gy artifact+creature → both move to **Hand**; unchosen enchantment+land **stay in Graveyard** (negative control). Revert-fail assertion: `zone_of(artifact) == Hand`.
- `single_mode_is_legal` — `up to two` permits one mode; only the chosen enchantment returns.
- `land_mode_unavailable_without_land_card` — no land card in gy → land mode (index 3) in `unavailable_modes`; `SelectModes{[3]}` is rejected (CR 700.2a).
- `modes_have_no_unimplemented` — all four modes are `Bounce` (revert-probe).

Parser pipeline (`oracle.rs`): `call_damage_control_distributes_shared_return_effect_across_modes` — each mode lowers to `Bounce` with `Typed[<type>], controller: You, InZone Graveyard`; zero parse warnings.

Building-block units (`oracle_modal.rs`): card-type-axis distribution; non-bare-target modes left unchanged; no-template no-op; `mode_body_is_bare_target` discrimination.

**Revert probe:** reverting the one-line distribution hook fails all 4 integration tests + the pipeline test (each mode reverts to an unimplemented `target` marker; nothing moves).

## Gate results
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` (default + `upstream/main` base) — **exit 0**
- Parser diff string-dispatch grep — clean (the one `.split('.')` is annotated structural sentence segmentation, mirroring `parse_modal_header_ast`; not dispatch)
- `./scripts/check-engine-authorities.sh upstream/main` — **exit 0**
- `cargo test -p engine` — `13076 passed; 1 failed` where the single failure is the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (global perf-counter test; **passes in isolation**, unrelated to this diff)
- `cargo semantic-audit` / `cargo coverage` — run clean against generated card-data; **Call Damage Control: 0 Unimplemented, 4 Bounce, modal present**, and not present in semantic-audit findings.

Card parses to **0 residual Unimplemented**.

### DIFF
--- a/crates/engine/src/game/scenario.rs
+++ b/crates/engine/src/game/scenario.rs
@@ -913,6 +913,21 @@ impl<'a> CardBuilder<'a> {
         self
     }
 
+    /// CR 305: Make this card a land. Strips the Creature core type pushed by
+    /// `add_creature_to_graveyard` and adds Land. Mirrors `as_artifact`/
+    /// `as_enchantment`; reusable for graveyard-return targeting tests.
+    pub fn as_land(&mut self) -> &mut Self {
+        let obj = self.obj();
+        obj.card_types
+            .core_types
+            .retain(|t| *t != CoreType::Creature);
+        if !obj.card_types.core_types.contains(&CoreType::Land) {
+            obj.card_types.core_types.push(CoreType::Land);
+        }
+        self.sync_base_card_types();
+        self
+    }
+
     /// CR 302: Make this card a creature. Strips the spell core types
     /// (Instant/Sorcery) pushed by `add_spell_to_library_top` and adds Creature.
     /// Mirrors `as_sorcery`/`as_instant`; reusable for search/tutor library tests.
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -12035,6 +12035,75 @@ mod tests {
         assert!(r.parse_warnings.is_empty());
     }
 
+    /// CR 700.2 + CR 601.2c + CR 404.1: Call Damage Control (MSH) — a modal
+    /// spell whose shared return-to-hand effect is phrased once in the header
+    /// ("Return those cards from your graveyard to your hand.") and whose four
+    /// bullets are bare targets distinguished only by card-type. The shared
+    /// effect must be distributed across every mode so each lowers to an
+    /// `Effect::Bounce` (return to hand) of a card of its type in the
+    /// controller's graveyard — never an unimplemented target marker.
+    /// Revert-probe: if `distribute_shared_mode_effect` is removed, each mode is
+    /// an unimplemented "target" marker and the Bounce match below fails.
+    #[test]
+    fn call_damage_control_distributes_shared_return_effect_across_modes() {
+        use crate::types::ability::{FilterProp, TypeFilter, TypedFilter};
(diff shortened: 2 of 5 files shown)


matthewevans (approved):
Reviewed current head f638c6061f9165b8e9d4fea4ad94b78b516d73c5. Checked the parser rewrite against the modal casting/target-selection pipeline, verified CR references against the local rules file, and confirmed CI is green. This is a scoped parser feature with regression coverage for parsed ability shape, chosen-mode resolution, single-mode selection, and unavailable target modes.

PR #4071, opened 2026-06-21
feat(engine): replace connive keyword action (Leader, Super-Genius)

lgray:

## Card
**Leader, Super-Genius** (MSH, {U}{U})
> If a creature you control would connive, instead you draw a card, then that creature connives.
> At the beginning of combat on your turn, target creature you control connives.

Closes the residual gap on main: `Unimplemented[replacement_structure]: "If a creature you control would connive, instead you draw a card, then that creature connives."` The combat-trigger line already parsed; the gap was the **replacement effect on the connive keyword action**.

## Class built
A **replacement effect that intercepts the connive keyword action** — not just this card. Any "If [a creature filter] would connive, instead [chain]" card now composes: the parser produces a `ReplacementEvent::Connive` with a `valid_card` filter + an `execute` chain, and the runtime applier replaces the connive with the modified action.

## Trace + composition
Connive was **not** an interceptable runtime event — there is no `ProposedEvent::Connive`; `Effect::Connive` resolution ran draw/discard/counter inline. I event-ified it exactly as the already-event-ified keyword actions were (Scry CR 701.22, Mill CR 701.17, Explore CR 701.44, Proliferate CR 701.34, CoinFlip CR 705.1). **Explore was the structural template** (object-scoped keyword action with a Scry-prelude replacement):
- `connive::resolve` now proposes `ProposedEvent::Connive { object_id, count }` via `replacement::replace_event` before the draw/discard/counter pipeline (mirrors `explore::resolve`).
- New `connive::resolve_connive_effect` runs the internals **without** the propose step, used both by `resolve` after the pipeline clears and by the applier's `then ... connives` link (so the replacement gets exactly one opportunity — CR 614.5; no infinite recursion).
- `connive_applier` runs the `execute` chain (`Draw 1` → `Connive`) and returns `Prevented`, fully replacing the original event (mirrors `explore_applier`).
- Suppressed the Mandatory post-replacement continuation stash for `Connive` events: the applier owns the full "instead" chain, so a stashed continuation would re-run it when the connive's `ConniveDiscard` choice resolves (caught and fixed during testing).

`valid_card` ("a creature you control") gates via `ProposedEvent::affected_object_id()`/`affected_player()` (wired to the conniver's id/controller), reusing the existing replacement `valid_card` machinery.

## add-engine-variant verdict
**Add two leaf variants:** `ProposedEvent::Connive` + `ReplacementEvent::Connive`. Grepped `data/engine-inventory.json` — neither existed; **no sibling-cluster smell**. Connive (CR 701.50) is a categorically distinct keyword action, exactly parallel to Scry/Mill/Explore/Proliferate/CoinFlip, each of which occupies its own `ProposedEvent`+`ReplacementEvent` leaf. These are not X/OpponentX/TargetX siblings and do not share a comparator/scope/aggregator axis — parameterizing distinct keyword actions under one variant would conflate independent CR 701.x sections. The variant stays within a single CR rule section (701.50). Inventory regenerated (`cargo engine-inventory`); both variants present and consistent.

## Grep-verified CRs (against docs/MagicCompRules.txt)
- **CR 701.50a** — connive: draw a card, then discard a card; +1/+1 per nonland discarded.
- **CR 701.50b/c** — "connives" completes after the process; LKI for the conniver.
- **CR 614.1a** — effects using "instead" are replacement effects.
- **CR 614.5** — a replacement effect doesn't invoke itself repeatedly (the applier re-runs internals directly).
- **CR 616.1** — multiple-replacement ordering (resume arm in `engine_replacement`).

## Discriminating-test map (parse → resolve)
| behavioral claim | seam | production entry | test | revert-failing assertion | sibling/negative |
|---|---|---|---|---|---|
| Leader replaces a controlled creature's connive: extra draw, then connive | `connive::resolve` propose step + `connive_applier` | `connive::resolve` (combat trigger entry) | `leader_connive_replacement_draws_extra_then_connives` | after `resolve`, state is `WaitingFor::ConniveDiscard` with both the extra draw + the connive draw in hand (without the fix only 1 card → auto-discard, no pause) | drives the real `ConniveDiscard` action through `handle_resolution_choice`; asserts +1/+1 counter + final hand |
| Replacement does NOT fire for an opponent's creature | `valid_card` controller scope + `affected_*` wiring | `connive::resolve` (opponent controller) | `leader_connive_replacement_skips_opponent_creature` | Leader controller's hand stays empty; opponent's plain connive still adds +1/+1 | negative case |
| Replacement line parses to `ReplacementEvent::Connive` (valid_card + Draw→Connive chain) | `parse_connive_replacement` | `parse_replacement_line` | `leader_connive_replacement_parses_to_connive_event` | event/valid_card/execute-chain shape | parse shape (backs the runtime tests) |
| Full card: 0 residual Unimplemented | full pipeline | `parse_oracle_text` | `leader_super_genius_card_parses_with_no_unimplemented` | no Unimplemented ability/trigger; Connive replacement present | — |

**Revert probe:** Forcing `connive::resolve` to bypass `replace_event` (`return resolve_connive_effect(...)`) makes `leader_connive_replacement_draws_extra_then_connives` FAIL (`Priority` instead of `ConniveDiscard` — only the connive's own card is drawn). Confirmed locally. Real generated `card-data.json` shows Leader, Super-Genius with **0 Unimplemented**, no `replacement_structure`, Connive replacement present.

## Gate results
- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` — **0 warnings**
- `cargo test -p engine` — **13082 passed; 1 failed** (`delve_payment_skips_state_clone_per_graveyard_candidate`, a known parallel-flaky state-clone-count test — **passes in isolation**, unrelated to this change)
- New connive suite: **11 passed; 0 failed**
- `./scripts/check-parser-combinators.sh` (default base + `upstream/main`) — exit 0; parser diff string-dispatch grep — clean
- `cargo engine-inventory` — both new variants present/consistent
- `cargo semantic-audit` / `cargo coverage` — clean run; Leader has **no findings**
- CR-annotation diff gate — no UNVERIFIED rules


matthewevans (changes requested):
Reviewed current head 35522b2d50565a395a4d08b8199ce4bd0a89b81a. Changes requested: the connive replacement applier uses the original proposed connive count when executing the replacement's plain 'then that creature connives' effect. In crates/engine/src/game/replacement.rs, connive_applier matches Effect::Connive but calls resolve_connive_effect(state, object_id, count, events), where count is from the replaced ProposedEvent::Connive. That is wrong for any creature that would connive N/X under Leader, Super-Genius: CR 701.50e makes 'connive N' a distinct variant, while Leader's replacement text says only 'that creature connives', so the replacement connive should resolve the parsed replacement effect's own count (plain connive = 1), not inherit the replaced event's count. Existing Spymaster's Vault-style connive X coverage shows this is a live engine path; the new tests only cover count=1, so the bug is false-green.

lgray:

Rebased onto current `upstream/main` and addressed the CHANGES_REQUESTED review. New head: `32db02979`.

### Merge conflict (review,conflict flag)
The only conflict on rebase was the auto-generated `data/engine-inventory.json`. The branch's original inventory was a **full regen** that swept in 179 added + 16 removed entries belonging to *other* features merged into main after this branch was cut — unrelated drift, not this PR's surface.

**Surgical resolution** (no full regen, to avoid re-sweeping unrelated merged features):
- Took `upstream/main`'s generated inventory as the base.
- Re-applied **only** this PR's two genuine variant additions: `ProposedEvent::Connive` (struct) and `ReplacementEvent::Connive` (unit). `variant_count` 3250 → 3252.
- Matched the generator format exactly (`serde_json::to_string_pretty`, 2-space indent, literal UTF-8 em-dashes, trailing newline) with line numbers from the rebased source.
- Verified the resolved file vs upstream: **ADDED** = exactly the 2 connive variants, **REMOVED** = none, line-only shifts confined to those two enums; `smell_summary` and `sources` identical. (The connive-related `Effect::Connive`, `EffectKind::Connive`, `TriggerEventKey::ConniveResolved`, `TriggerMode::Connives`, `WaitingFor::ConniveDiscard` already existed on `upstream/main` and were not re-added.)

The inventory diff is now 90 lines (was 6405) — only the connive additions remain.

### Review fix (@matthewevans — connive-count bug)
Confirmed and fixed, grep-verified against `docs/MagicCompRules.txt`. In `connive_applier` the `Effect::Connive` chain link ("then that creature connives") was passing the **replaced event's** `count` to `resolve_connive_effect`, so a creature conniving N>1 under Leader, Super-Genius would have run the plain connive at N instead of 1. Per **CR 701.50e** ("Connive N is a variant of connive") and **CR 701.50a** (plain connive = draw 1 / discard 1), Leader's "that creature connives" parses to a plain connive (`Effect::Connive { count: Fixed(1) }`), so the link must resolve its **own** `QuantityExpr` count.

Fix: the arm now builds a `ResolvedAbility` from the chain `def` (conniving permanent as the target) and resolves the link's own count via `resolve_quantity_with_targets` — mirroring the normal connive resolver (`effects/connive.rs`). This generalizes to the whole class: plain (`Fixed(1)`), static connive N (`Fixed(N)`), and dynamic connive N. The count resolver never re-enters the replacement pipeline, so CR 614.5 (one opportunity, no self-invocation) is preserved.

Added a discriminating regression test `leader_connive_n2_replacement_runs_plain_connive`: a creature conniving N=2 under the Leader replacement must draw/discard exactly 1. Proven non-vacuous — pre-fix it observes `WaitingFor::ConniveDiscard { count: 2 }` (hand 3, third card drawn out of library); post-fix `count: 1` (hand 2, third card stays in library). The Gemini bot comment was only a daily-quota notice (no actionable content).

Verification: `cargo fmt --all` clean, `cargo clippy -p engine --lib` clean, `cargo test -p engine --lib connive` 12/12 green.

matthewevans (changes requested):
VERDICT: request-changes

Prior finding status: PARTIAL. The plain-connive count bug is addressed for the single-Leader case, but the replacement pipeline is still bypassed for the modified connive event.

**[HIGH]** Other applicable connive replacement effects are skipped after the first one is applied. Evidence: `crates/engine/src/game/replacement.rs:1774-1837` drops the original `ProposedEvent::Connive.applied` set and resolves the replacement chain's nested `Effect::Connive` by calling `resolve_connive_effect` directly; `crates/engine/src/game/effects/connive.rs:82-87` documents that helper as bypassing connive replacements entirely; the resumed replacement-choice path does the same for a surviving `ProposedEvent::Connive` at `crates/engine/src/game/engine_replacement.rs:295-307`. Why it matters: CR 616.1f requires the replacement process to repeat after one applicable replacement is applied, while CR 614.5 only prevents the same replacement from applying again; with two Leader-style effects, or a future connive count modifier plus Leader, the unchosen/still-applicable connive replacement never gets a chance to apply to the modified connive event. Suggested fix: carry the `applied` set through `ProposedEvent::Connive` and re-enter the replacement pipeline for the nested/surviving connive event with the just-applied replacement marked, then only run `resolve_connive_effect` once the pipeline returns an executable `Connive` survivor; add a regression with two applicable connive replacements or a synthetic second connive replacement proving both effects can apply once.

`cargo fmt --all --check` passes on this head.

lgray:

Rebased onto current `main` and addressed the latest CHANGES_REQUESTED review (#4071 — connive replacement, CR 616.1f). New head: `176da1ddc`.

### Merge conflict (cause + surgical resolution)
The only conflict on rebase was the auto-generated `data/engine-inventory.json` count header — `main` keeps advancing through a busy merge queue, and each merge regenerates the inventory. A full regen on the rebased tree sweeps in `main`'s own inventory drift (unrelated merged features), so I resolved it **surgically**:

- Took the current base's generated inventory as the starting point.
- Swapped in only this PR's two enum blocks (`ProposedEvent` / `ReplacementEvent`), which differ from the base by **exactly** the two genuine additions: `ProposedEvent::Connive` (struct, src line 268) and `ReplacementEvent::Connive` (unit, src line 78). `variant_count` +2; `enum_count` unchanged (both go into existing enums).
- Verified: vs the base, **ADDED = exactly the 2 Connive variants, REMOVED = none, enum keys identical**; the two Connive entries are **byte-identical** to `cargo engine-inventory` output (same line numbers, doc, `cr_refs`). The inventory diff is ~90 lines, only the connive additions.

### Review fix (@matthewevans [HIGH] — CR 616.1f repeat)
Confirmed and fixed, all CRs grep-verified against `docs/MagicCompRules.txt`.

The connive replacement applier ran the whole `execute` chain inline and resolved the nested "then that creature connives" link via a **direct** `resolve_connive_effect` call that bypassed the replacement pipeline, then returned `Prevented`. With two or more co-applicable connive replacements, the still-applicable ones never got a chance to apply to the modified connive event — violating **CR 616.1f** ("once the chosen effect has been applied, this process is repeated … until there are no more left to apply"). **CR 614.5** bars only the *same* replacement from re-applying; the `applied: HashSet<ReplacementId>` carried in `ProposedEvent::Connive` is the engine's tracker.

Fix:
- Factored the propose+match logic into a single authority `connive::propose_connive(state, conniver, count, applied, events)`. The top-level resolver seeds `applied` empty; the applier's nested link seeds it with the just-applied replacement marked.
- The nested `Effect::Connive` arm now **re-enters the pipeline** via `propose_connive(..., applied.clone(), ...)` instead of calling `resolve_connive_effect` directly. Because `applied` already contains the just-applied rid (the loop/resume marked it before the applier ran), `find_applicable_replacements` excludes it (CR 614.5) — no self-invocation — while the **other** still-applicable connive replacements get their CR 616.1f repeat. The link still resolves its **own** parsed `QuantityExpr` count (preserving the prior count fix at `32db02979`), now seeding the re-proposed event.
- For the 3+ co-applicable case, the post-choice `Prevented` arm in `handle_replacement_choice` was clobbering the freshly-parked nested `ReplacementChoice` with `Priority`. Since `continue_replacement` consumes the prior pending record at its start, a `pending_replacement.is_some()` in that arm uniquely means the applier parked a fresh nested ordering choice — the arm now surfaces it so the CR 616.1f repeat is not dropped.

Discriminating regressions (both drive the real production path — `resolve` → `apply_as_current(ChooseReplacement)` → `ConniveDiscard` → `handle_resolution_choice`):
- `two_connive_replacements_both_apply_once` — both fire once, both extra draws happen, the plain connive draws exactly 1 (count fix), exactly one `EffectResolved(Connive)`. Proven non-vacuous: reverting the nested re-entry yields `ReplacementApplied(Connive)` = 1 and only one extra draw.
- `three_connive_replacements_all_apply_once` — all three fire once, no orphaned `pending_replacement`. Proven non-vacuous: without the `pending_replacement.is_some()` guard clause the second prompt is clobbered to `Priority` after one apply (and the N=2 test still passes, so the N=3 test discriminates the 3+-specific defect).

CR annotations added/verified: CR 701.50a, CR 701.50e, CR 614.1a, CR 614.5, CR 616.1, CR 616.1f.

### Verification
`cargo fmt --all` clean; `cargo clippy -p engine --lib` clean; `cargo test -p engine --lib` 13218 passed / 0 failed; `cargo test -p engine --lib connive` 14/14 green (incl. both new regressions and the count-fix test `leader_connive_n2_replacement_runs_plain_connive`).

matthewevans (changes requested):
Thanks for the repush. The CR 616.1f repeat path for multiple Leader-style connive replacements is much closer now, but I still found a blocking continuation bug in the replacing action chain.

`connive_applier` manually drives the replacement execute chain. For the leading `Draw` link it strips `sub_ability` and calls `resolve_ability_chain`, but it ignores whether that call parked an interactive replacement choice. `draw::resolve` returns early on `ReplacementResult::NeedsChoice` after setting `state.waiting_for = ReplacementChoice` (via `draw_through_replacement`), and because the sub-ability was stripped, the generic chain walker has no opportunity to stash the later `then that creature connives` link. The manual loop then immediately advances to `def.sub_ability` and runs `Effect::Connive` anyway.

That means a real game with Leader plus any draw replacement that surfaces a choice (for example Abundance, which this parser/engine already models, or multiple material draw replacements) performs the connive draw/discard before the replacement choice for the preceding "you draw a card" has resolved. That violates the printed order of the replacement action (draw first, then connive) and can also overwrite/orphan the live `ReplacementChoice` with `ConniveDiscard` or another nested prompt. Please add a regression where Leader's replacement draw is itself replaced/parked, and make the applier pause after the draw link with the remaining `then connives` continuation preserved instead of continuing the loop synchronously.

lgray:

Rebased onto current `upstream/main` (`cf6a9a320`) and addressed the latest CHANGES_REQUESTED review (the continuation bug flagged after `176da1ddc`). New head: `af2448cfe`. The inventory rebased cleanly with no conflict this time.

### Review fix (@matthewevans — replacing-action chain continuation bug)
Confirmed and fixed. `connive_applier` drove the `Draw → Connive` execute chain link-by-link and ran the nested connive synchronously, ignoring when the leading `Draw` link parked an interactive replacement choice (`draw::resolve` returns early on `ReplacementResult::NeedsChoice` after setting `state.waiting_for`). So with the Leader's own draw itself replaced (a draw that surfaces a choice — Abundance, or two co-applicable material draw replacements), the connive ran **before** the draw choice resolved, violating the printed order (**CR 701.50a** "instead you draw a card, **then** that creature connives") and clobbering the live `ReplacementChoice`.

**Fix** (CRs grep-verified against `docs/MagicCompRules.txt`):
- The applier now **pauses** after a leading link parks and **defers** the remaining `then connives` link into a dedicated `GameState.pending_connive_reentry` slot (carrying the conniving permanent, the link's own resolved count, and the `applied` set), then returns `Prevented`. The parked draw choice surfaces to the player untouched.
- A dedicated slot — not `post_replacement_continuation` — is load-bearing: that shared slot is drained by every zone-delivery tail (`apply_zone_delivery_tail`), and the leading draw delivers its cards through that exact seam, so a continuation parked there fires **mid-draw** and the resume's `Priority` reset drops the pause. The dedicated slot is invisible to the delivery tail and is drained **only** by the post-replacement-choice epilogue in `handle_replacement_choice`, after the draw fully delivers, so the `then connives` step runs in printed order.
- Drained on **both** resume arms via a shared `drain_pending_connive_reentry` helper: the **Execute** arm (draw accepted) and the **Prevented** arm (draw prevented — per **CR 701.50b** the connive still runs even when the inner draw was impossible). The re-entry goes through `propose_connive` carrying the `applied` set, so the **CR 616.1f** repeat still covers the other connive replacements while **CR 614.5** bars self-invocation. The prior count fix and the multi-Leader CR 616.1f repeat are preserved.

### Discriminating regressions (both drive the real action pipeline)
- `leader_connive_parked_draw_defers_then_resumes` — two non-commuting count-modifier draw replacements force a CR 616.1 ordering park on the leading draw; the connive defers (`pending_connive_reentry == Some`, `waiting_for == ReplacementChoice`, no counter / no `EffectResolved` yet); after resolving the draw choice the connive proceeds in order to `ConniveDiscard` → exactly one `+1/+1` + one completion. Non-vacuous: reverting the defer leaves `waiting_for == ConniveDiscard` (connive ran early) and the slot empty; reverting the drain strands the slot `Some`.
- `leader_connive_prevented_draw_still_connives` — two mandatory draw-Prevent replacements; choosing the Prevent drives the **Prevented** arm. Asserts the connive still ran (slot drained to `None`, one `EffectResolved{Connive}`). Non-vacuous: neutering the Prevented-arm drain strands `Some(PendingConniveReentry { .. })` and emits no completion.

Explore-replacement appliers share the same multi-link shape but are intentionally **out of scope** here: their leading `Scry` parks a `ScryChoice` whose resume path drains `pending_continuation`, not `post_replacement_continuation`, so the same mechanism doesn't apply — that's a separate latent issue, tracked for a follow-up.

### Verification
`cargo fmt --all` clean; `cargo clippy -p engine --lib` clean; `cargo test -p engine --lib` 13268 passed / 0 failed; `cargo test -p engine --lib connive` 16/16 green (incl. both new regressions). CR annotations verified: CR 701.50a, 701.50b, 614.5, 616.1, 616.1f.

lgray:

Rebased onto current `upstream/main` (`f42195215`) to clear the `data/engine-inventory.json` merge conflict. New head: `c792ba832`. PR is no longer `CONFLICTING` (`mergeable: MERGEABLE`, status `BEHIND`).

### Conflict resolution (verified-clean inventory regen)
The only conflict was the auto-generated `data/engine-inventory.json` (main advanced). Rather than a textual merge, I reset the inventory to upstream/main's current version and regenerated it with `cargo engine-inventory` against the rebased source. Verified the regen is surgical:

- **Structural delta vs `f42195215` = exactly +2 variants, zero swept:** `ProposedEvent::Connive { object_id, count, applied }` (struct) and `ReplacementEvent::Connive` (unit). `variant_count` 3286 → 3288; zero removed `"name"` lines; no enums added/dropped.
- **Canonical:** re-running `cargo engine-inventory` produces a byte-identical file (`git diff --exit-code` clean) — the committed inventory is exactly what the generator yields for this source.
- The two source commits (`2e0fe6a2f` feat, `c792ba832` continuation fix) are **byte-identical** to the pre-rebase tip `af2448cfe`; the connive printed-order fix (incl. `pending_connive_reentry` / `PendingConniveReentry`) is intact.

### Verification (full gate run; Tilt down → cargo foreground)
- `cargo fmt --check`: clean. `clippy -p engine --all-targets`: exit 0, zero warnings.
- Engine test suite: **15,096 passed, 0 failed.** The continuation regressions pass: `leader_connive_parked_draw_defers_then_resumes`, `leader_connive_prevented_draw_still_connives`, plus the CR 616.1f multi-replacement tests (`two_/three_connive_replacements_*`).
- Parser-combinator + engine-authority diff gates: exit 0.
- card-data (parser changed): `oracle-gen` exit 0; `card-data-validate` OK (35,396 cards parse vs engine schema, incl. `ReplacementEvent::Connive`); `coverage-report` exit 0 — **Leader, Super-Genius `supported=True`, `gap_count=0`**, parsing to `event=Connive` with `execute=Draw(1) → sub_ability=Connive(1)` (printed order draw-then-connive).
- All 10 CR annotations added by this PR resolve in the Comprehensive Rules text (614.1a/614.1d/614.5/614.11a/616.1/616.1f/701.50a–e).

No review comments newer than `af2448cfe` were outstanding (the last `CHANGES_REQUESTED` continuation finding was already addressed by the fix commit). Diff touches exactly the 8 in-scope files; no unrelated changes.

matthewevans:
VERDICT: approve-pending-CI

Prior finding status: ADDRESSED. The latest implementation now defers the connive replacement's `then that creature connives` link when the leading draw parks a replacement choice, stores it in `pending_connive_reentry`, and drains it from the real `GameAction::ChooseReplacement` resume path after the draw choice resolves. I also checked the prevented-leading-draw path and the CR 616.1f applied-set propagation for other connive replacements; I do not see a remaining semantic blocker on the reviewed diff.

I resolved the current merge conflict via update-branch; the new head is `d8cf8d72b3ca7fdf232fcab9a15ba9d5ce1e0b39`. CI is still queued/running, so I am not formally approving or enqueueing until this exact head is green and mergeable.

matthewevans (approved):
Approved current head d8cf8d72b3ca7fdf232fcab9a15ba9d5ce1e0b39 after re-review and green CI. The prior parked leading-draw replacement-choice blocker is addressed with the pending_connive_reentry resume path and real ChooseReplacement regressions for both delivered and prevented leading draws.

PR #4072, opened 2026-06-21
feat: implement sticker sheet support and sticker application with layer and zone retention

### LINKED ISSUE #4070: feat(engine): implement Stickers and sticker sheets (CR 123)
## Motivation
Stickers are a fully-specified, CR-supported mechanic (Comprehensive Rules section 123). The engine currently handles only the *Ticket* player counter related to Unfinity; the actual sticker subsystem — sticker sheets, the four sticker kinds, and putting stickers on objects — is entirely missing.

## CR grounding (verified in `docs/MagicCompRules.txt`, section 123)
- **CR 123.1** — four kinds of stickers: name, ability, power/toughness, art.
- **CR 123.2** — sticker sheets (predetermined combinations on inserts).
- **CR 123.3** — putting a sticker: choose an unused sticker from accessible sheets and put it on the object.
- **CR 123.5** — stickers are retained into public zones (exception to 400.7);  not retained into hidden zones.
- **CR 123.6 / 123.7 / 123.8** — name sticker = text-changing effect; ability sticker = grants ability; P/T sticker = sets power/toughness (timestamp order, CR 613.4b / 613.7).
- Art stickers (CR 123.9) are cosmetic — out of scope for a rules engine.

## Scope
- Sticker-sheet data model (the 3 chosen sheets per player, each with its predetermined stickers).
- Object-side sticker tracking (a sticker is not a counter or token — CR 123.1).
- Three effect capabilities: put a **name** sticker, put an **ability** sticker, put a **power/toughness** sticker.
- Continuous-effect application via the existing layers system (text-change, ability-grant, P/T-set).

## Proposed approach (substantial, two-file)
1. New module `crates/engine/src/game/stickers.rs` — sticker sheet state, the  `Sticker` enum (`Name(String)` / `Ability(AbilityDef)` / `PowerToughness(i32,i32)`),  "choose an unused sticker" selection, and object sticker tracking.
2. New effect handler(s) for "put a [kind] sticker on [target]" — a new  `Effect::PutSticker { kind, target }` in `types/ability.rs`, resolved against the layer system (reuses existing text-change / ability-grant / P/T statics).
3. Sticker-sheet data seeded from MTGJSON/Scryfall sticker-sheet ids.
4. Tests in a separate `stickers_tests.rs` declared from the parent module.

## Acceptance criteria
- [ ] Name/ability/P/T stickers modify object characteristics per CR 123.6–123.8.
- [ ] Stickers persist into public zones (CR 123.5) and are dropped in hidden zones.
- [ ] At least the ~31 put-sticker gap cards parse and resolve.

## Notes
- Art stickers (CR 123.9) are cosmetic — out of scope for a rules engine.
- `score_pr.py` harness is missing — Potential is an estimate from the validated
  table.

carlos4s:
## Summary

Closes #4070 

Implement sticker sheet support and sticker application in the engine, including name, ability, and power/toughness stickers, plus correct layer handling and sticker retention across public-zone moves. This also adds parser support for core "put a sticker" clauses and wires sticker sheet selection into deck and game state.

the pipeline):

> Implemented directly in the branch because the work was requested as an in-place engineering task rather than through the `/engine-implementer` workflow.

> [!NOTE]

## CR references

- `CR 123.1` — sticker kinds
- `CR 123.2` — sticker sheets
- `CR 123.3` — choosing and putting stickers
- `CR 123.5` — sticker retention in public zones and loss in hidden zones
- `CR 123.6` — name stickers
- `CR 123.7` — ability stickers
- `CR 123.8` — power/toughness stickers
- `CR 613.4b`
- `CR 613.7`

## Verification

Ran and passed:

```bash
cargo check -p engine
cargo test -p engine stickers -- --nocapture
cargo test -p engine put_name_sticker_parses -- --nocapture
cargo test -p engine put_ticket_bounded_ability_sticker_parses -- --nocapture
```

Added focused coverage for:
- battlefield sticker application and public/hidden zone behavior
- automatic sticker application when only one candidate is legal
- parser support for name-sticker and ticket-bounded ability-sticker clauses


matthewevans (changes requested):
Reviewed current head ae6aba03d18448d61587099adcfb1ab418c29e5a enough to diagnose the terminal CI failure. Changes requested: Rust lint fails the parser-combinator gate in crates/engine/src/parser/oracle_effect/imperative.rs. The new sticker parser uses string-method dispatch at lines 136, 148, 151, 161, 183, and 196 (, , ) and adds a string-literal match dispatch at line 7662 (). This violates the repository's nom-combinator mandate; please rewrite the sticker grammar using nom combinators (or add legitimate allow-noncombinator annotations only for structural cleanup, not parsing dispatch) and rerun CI. I did not continue semantic review because this head is terminal-red and BEHIND.

matthewevans (changes requested):
Reviewed current head ae6aba03d18448d61587099adcfb1ab418c29e5a enough to diagnose the terminal CI failure. Changes requested: Rust lint fails the parser-combinator gate in crates/engine/src/parser/oracle_effect/imperative.rs. The new sticker parser uses string-method dispatch at lines 136, 148, 151, 161, 183, and 196 (strip_prefix, strip_suffix, contains) and adds a string-literal match dispatch at line 7662 ("put" => try_parse_put_sticker_effect(...)). This violates the repository nom-combinator mandate; please rewrite the sticker grammar using nom combinators, or add legitimate allow-noncombinator annotations only for structural cleanup, not parsing dispatch, and rerun CI. I did not continue semantic review because this head is terminal-red and BEHIND.

matthewevans (changes requested):
Changes requested on current head ada9852eafc326814994b80528ec97b313b22fba.

1. The normal deck paths drop sticker sheets before game setup. The engine payload now has `sticker_sheets` and `load_deck` applies it, but the actual resolver/adaptor paths still construct `PlayerDeckList`/`PlayerDeckPayload` with `sticker_sheets: Vec::new()` in `crates/engine-wasm/src/lib.rs:1351`, `crates/phase-server/src/main.rs:2441`, and `crates/server-core/src/deck_resolve.rs:80`. That means a real client/server game starts with no selected sheets, so `available_sticker_candidates` returns empty and the new `PutSticker` effects silently do nothing outside hand-built tests. Please thread the field through the deck APIs that create these payloads, or keep the parser unsupported until the runtime can actually supply sheets.

2. Sticker-granted continuous static abilities are not reliably included in layer evaluation. The layer pass rebuilds `StaticSourceIndex` at `crates/engine/src/game/layers.rs:1436`, then applies sticker ability text at `layers.rs:1445`; `apply_ability_stickers` appends parsed `static_abilities` into `obj.static_definitions` at `crates/engine/src/game/stickers.rs:475-476`. The later gather (`layers.rs:1448`) visits sources via the already-built index (`layers.rs:2354-2377`). On the next pass, Step 1 resets live statics back to base before rebuilding the index, so a vanilla object with a static-effect ability sticker is never indexed unless the fallback happens to scan every battlefield object. This is observable with sticker texts in `stickers_data.json` such as Hellbent/Threshold/Metalcraft static grants. Please either apply sticker statics before rebuilding the index, rebuild after applying sticker abilities, or otherwise make sticker-generated continuous statics participate in the same source-index contract, with a regression that includes another indexed static source so the fallback does not mask it.

The nom rewrite from the prior review looks materially better, and CI is green, but these are semantic blockers for the feature.

matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** The normal saved/active-deck path still drops the selected sticker sheets before the new resolver plumbing can use them. Evidence: `normalizeParsedDeck` builds a fresh `ParsedDeck` with only `main`/`sideboard`, then optional `commander`/`companion`, and never copies `deck.sticker_sheets` at `client/src/services/deckParser.ts:237-273`; `loadSavedDeck()` returns that repaired object directly at `client/src/constants/storage.ts:146-155`; and game startup converts the repaired `ParsedDeck` through `expandParsedDeck()` at `client/src/providers/GameProvider.tsx:176-177`. Why it matters: a saved deck JSON that carries `sticker_sheets` still reaches the engine as `[]`, so real client games can continue to have no sticker candidates despite the server/WASM resolver changes in this head. Suggested fix: preserve `sticker_sheets` in `normalizeParsedDeck`/repair and add a focused deckParser/storage-path regression proving `repairParsedDeck` or `loadSavedDeck` keeps the field through `expandParsedDeck`.

matthewevans:
Rechecked the new `4cf5ff3336302adb7901900af8b40f616b34f4f5` head after the merge-from-main. The prior saved-deck blocker is addressed: `repairParsedDeck` preserves `sticker_sheets`, `expandParsedDeck` carries them to the engine payload, and the added deck parser/storage tests cover that path.

I am not approving or enqueueing yet because fresh CI is still running on this head. Required `feature` label is present; holding this as pending CI/current-head guard.

carlos4s:
@matthewevans All CI passed

matthewevans (approved):
Current head `4cf5ff3336302adb7901900af8b40f616b34f4f5` addresses the prior blockers I had open: sticker sheets now survive the server/WASM resolver paths and the client saved-deck repair/load bridge into `expandParsedDeck`, and sticker-granted static abilities are reindexed before layer collection. Fresh GitHub CI is green and the required `feature` label is present.

matthewevans (changes requested):
## Requesting changes — `count`/`up_to` should route through `QuantityExpr`, not a hardcoded count + bool

Thanks for the substantial work here — the CR-123 sticker *infrastructure* (zone retention, layer handling, sheet/locator model) looks reasonable. But the parser/effect surface for putting stickers needs to change before this can merge.

### The core issue: `Effect::PutSticker { count: u32, up_to: bool }`

In `parser/oracle_effect/imperative.rs`, `try_parse_put_sticker_effect` does this:

```rust
if let Ok((target_text, _)) = value(
    (), tag::<_, _, OracleError<'_>>("put up to two name stickers on "),
).parse(text) {
    return Some(Effect::PutSticker {
        kind: Some(StickerKind::Name),
        count: 2,        // hardcoded integer
        up_to: true,     // bare bool flag
        ...
    });
}
```

This is a verbatim full-phrase `tag()` that bakes the number word *"two"* into the literal and emits a literal `count: 2`. It handles exactly one card and breaks the moment a card says "up to three" or "up to X". Two project rules are in play: **never match on verbatim Oracle text**, and **build for the class, not the card**.

### This reinvents an existing, documented building block

The engine already has a first-class representation for "up to N". `types/ability.rs` (the `QuantityExpr::UpTo` doc comment):

> `CR 107.1c + CR 608.2d`: "Up to N" — the affected player chooses any integer in `0..=resolve(max)` at resolution time. Used by `Effect::Draw`, `Effect::Sacrifice`, `Effect::Discard`, and `Effect::SearchLibrary` … The 4 specific resolvers peel this wrapper via `QuantityExpr::peel_up_to` and propagate the bool to their `WaitingFor::*Choice` runtime state.

"Put up to N stickers" is simply the **fifth** member of this class. The helpers already exist: `QuantityExpr::up_to(max)` and `peel_up_to() -> (&QuantityExpr, bool)`.

Consequences of the current design:
1. `count: u32` can't express "up to your hand size" / "up to X" — `QuantityExpr` can.
2. `up_to: bool` is a bare bool flag that duplicates the `UpTo` wrapper's job (violates "no bool flags — parameterize with existing enums").
3. The resolver's hand-rolled `prompt_count_choice` / `for amount in 1..=count` reimplements the `peel_up_to` + "choose 0..=max" logic that Draw/Sacrifice/Discard/SearchLibrary already share.
4. **Inconsistency in the same function**: `max_ticket_cost` is correctly parsed via `parse_count_expr` → `QuantityExpr`. The right tool was in hand and applied to one field but not the other.

Every comparable count-bearing effect — `Draw`, `PutCounter`, `RemoveCounter`, `Sacrifice`, `Mill`, `Scry` — carries `count: QuantityExpr`. `Sacrifice` even shows how to express "may pick fewer" without a bool: `count: QuantityExpr` + `min_count: usize`. `PutSticker` is the lone outlier.

### The "up to N" combinator is already in the path you branch past

`parser/oracle_effect/sequence.rs` (the existing `parse_put_ast` flow that the new sticker block sits *in front of*):

```rust
} else if let Ok((rest, _)) = tag("up to ").parse(after_put) {
    if let Ok((remainder, n)) = nom_primitives::parse_number.parse(rest) {
        (PutCount::Up(n), remainder.trim())   // parse_number handles digits AND "two"/"three"/…
    } else { (PutCount::Up(1), rest) }
}
```

`parse_number` (`parser/oracle_nom/primitives.rs`) already maps `"two" → 2`. The combinator to do this correctly is in the same file as the code being bypassed.

### Requested changes

1. **`count: QuantityExpr`**, drop `up_to: bool`. Encode "up to two" as `QuantityExpr::up_to(QuantityExpr::Fixed { value: 2 })`.
2. **Parse the count with combinators**, not a permuted full-string `tag`. Compose: `tag("put ")` → quantity (`a`/`an` → `Fixed(1)`; `up to <parse_number>` → `UpTo(Fixed(n))`) → sticker-kind `alt()` (with optional plural `s`) → `tag(" on ")` → `parse_target_with_ctx`. One composition covers `a name sticker`, `up to two name stickers`, `a power and toughness sticker`, `an art sticker`, `a sticker` (kind=None), and any future count for free.
3. **Resolver mirrors Draw/Sacrifice**: `let (max, may_pick_fewer) = count.peel_up_to(); let n = resolve_quantity(max, …);` — deletes the bespoke `up_to` field and the duplicated count loop.
4. **`without_paying: bool`** is a second bare bool (lower priority) — prefer a small typed enum for the ticket-cost waiver if you touch it.
5. **Add the missing runtime test.** The `count > 1` / `up_to` / `prompt_count_choice` branch — i.e. `_____ _____ Rocketship`'s "put up to two name stickers" — has no end-to-end coverage. `stickers_tests.rs` only exercises `count: 1`. The two parser tests added are AST-shape tests (`parse_effect` + `matches!`); please add a `GameScenario`/`GameRunner` test that drives parse → cast → resolve for the up-to-two path.

The idiomatic version is strictly *less* code (deletes a field, a loop, and the verbatim tag), which is usually the tell that it's the right path.

matthewevans:
Separately, on process — and we genuinely appreciate the contribution and the effort that went into a CR-123 implementation of this size:

A ~3,000-line change to `crates/engine/` game logic (parser + effects + resolver + targeting + rules behavior) is exactly the kind of work the `/engine-implementer` pipeline (plan → review-plan → implement → review-impl → commit) exists for, and it's **mandatory** for changes of this type. The PR template's "Not `/engine-implementer`" box is reserved for changes that fall outside engine game logic (frontend-only, docs/CI/tooling, release chores, or a fix too small to warrant the pipeline) — a multi-file new mechanic touching the parser, effect resolver, and rules behavior is squarely inside that scope, not outside it.

The architectural issues called out in the review above (a new `Effect` variant reinventing `QuantityExpr::UpTo`, verbatim-tag parsing, an untested multi-count branch) are precisely what the plan-review and impl-review steps of that pipeline are designed to catch *before* 3,000 lines are written. Routing this through `/engine-implementer` would have surfaced them up front and saved a large rework cycle.

Could you please re-run this work through `/engine-implementer` for the changes requested? Happy to help if you hit anything along the way.

carlos4s:
@matthewevans I've updated. Could you review again?

matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** `QuantityExpr::UpTo` only permits choosing zero when the resolved maximum is greater than one. Evidence: `resolve_put_sticker` peels the `UpTo` wrapper at `crates/engine/src/game/effects/stickers.rs:83`, but only calls `prompt_count_choice` inside `if count > 1 { if up_to { ... } }` at lines 94-109. If the parser accepts `put up to one name sticker ...`, `parse_sticker_count_expr` lowers it to `QuantityExpr::UpTo(Fixed(1))`, `count` resolves to 1, and execution falls through to the single-sticker candidate chooser at lines 131-210. That forces one sticker whenever candidates exist instead of offering the legal zero choice.

This is the same class-level issue the prior review was trying to avoid: the resolver must honor `UpTo` uniformly as `0..=max`, not only for `max > 1`. Please route every `up_to` count through the count-choice path when `count > 0` (or otherwise preserve the zero option for max-one), and add a focused regression for `put up to one ...` choosing zero.

matthewevans:
VERDICT: approve-pending-CI

Re-reviewed current head `b46bd5f1faaaff9dc607a835a76f9be06b3b8328`. The prior `up to one` blocker is addressed: `resolve_put_sticker` now peels `QuantityExpr::UpTo` and routes all positive max values through the count-choice prompt, including max one, and `prompt_count_choice` includes an explicit `Do not put a sticker` zero branch before any sticker-application branch. The new `cast_up_to_one_name_sticker_allows_choosing_zero` regression drives the production cast/resolve path, chooses the zero branch, and asserts the target remains stickerless.

GitHub CI is still running (`Rust tests` and paired-seed AI gate), so no formal approval/enqueue until checks are green/current and the repush guard passes.

matthewevans (approved):
Approved after re-review of the QuantityExpr::UpTo sticker follow-up, green current-head CI including paired-seed AI gate, CLEAN merge state, and repush guard on b46bd5f1faaaff9dc607a835a76f9be06b3b8328.

PR #4078, opened 2026-06-21
feat(parser): control target opponent during their next turn (Construct a Cosmic Cube)

lgray:

## Summary

**Construct a Cosmic Cube** (MSH) had one residual gap: its seventh-plan-counter reflexive rider — "you control target opponent during their next turn" — parsed as `Effect::Unimplemented`. The shared `ControlNextTurn` suffix combinator only matched the Mindslaver phrasing `" during that player's next turn"`.

This composes the possessive as a single `alt()` axis so the existing turn-control subsystem handles Construct's phrasing with no new types or variants.

## Change

`try_parse_control_next_turn_suffix` previously hard-coded one full-string `tag`. CR 723.1 ("Controlling Another Player") is a single grammatical axis on the possessive — `"that player's"` when the target noun is "player" (Mindslaver), `"their"` when it is "opponent" (Construct). Both refer to the targeted player and lower to the same `Effect::ControlNextTurn`. The combinator is now:

```
preceded(" during ", terminated(alt(("that player's", "their", "its")), " next turn"))
```

This reuses the entire turn-control subsystem already in the engine:
- schedule push: `control_next_turn::resolve` → `state.scheduled_turn_controls`
- activation on the controlled player's next turn: `turns::start_next_turn`
- decision delegation: `turn_control::turn_decision_maker` / authorized-submitter routing

No new `Effect` variant, no new types, no engine-inventory change.

Also corrects a stale CR annotation on this effect path: it cited **CR 722** (Preparation Cards) where the controlling-another-player rule is **CR 723**.

## Verification

- `cargo fmt --all` clean; clippy `--workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings` clean.
- `cargo test -p engine`: all pass except the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (passes in isolation; unrelated).
- Parser combinator gate (`check-parser-combinators.sh`) and the parser-diff string-dispatch grep: clean (no `contains`/`split`/`find` dispatch added).
- `check-engine-authorities.sh upstream/main`: pass.
- Regenerated card-data: **Construct a Cosmic Cube now has 0 Unimplemented effects and 1 `ControlNextTurn`**. `semantic-audit` does not flag the card or any control-next-turn card.

### Discriminating tests
- **Parser shape**: `parse_control_next_turn_their_possessive_opponent_target` (the new "their"/"opponent" axis); the swallow_check gate now asserts the whole card is 0-Unimplemented and the rider lowers to `ControlNextTurn`.
- **Runtime** (`construct_seventh_counter_rider_controls_opponents_next_turn`): parses the full card through the production parser, extracts the `ControlNextTurn` effect, resolves it via `control_next_turn::resolve`, then advances turns through the real `turns::start_next_turn` and asserts control activates on the opponent's next turn (`turn_decision_controller == controller`, CR 723.1) and reverts at the following turn. Reverting the parser fix makes the rider `Unimplemented`, the effect is not found, and the test fails.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -127,20 +127,36 @@ fn parse_dig_library_owner(rest_lower: &str) -> TargetFilter {
     TargetFilter::Controller
 }
 
-/// Shared ControlNextTurn suffix parser (CR 722.1). Called after a prefix
+/// Shared ControlNextTurn suffix parser (CR 723.1). Called after a prefix
 /// combinator ("you control " or "gain control of ") has matched; parses the
-/// target, then " during that player's next turn", then the optional extra-turn
-/// tail (CR 722.1 doesn't require it; some cards like Emrakul grant it).
+/// target, then the "during <possessive> next turn" duration clause, then the
+/// optional extra-turn tail (CR 723.1 doesn't require it; some cards like
+/// Emrakul grant it).
+///
+/// CR 723.1: "control another player during that player's next turn." The
+/// possessive is a single grammatical axis — "that player's" when the target
+/// noun is "player" (Mindslaver), "their" when the target noun is "opponent"
+/// (Construct a Cosmic Cube). Both refer to the targeted player, so they lower
+/// to the same effect; they are composed as one `alt()` over the possessive,
+/// never enumerated as separate full-string arms.
+///
 /// Returns `None` when the suffix doesn't apply, allowing the caller to treat
 /// the match as a different effect (e.g., plain `GainControl`).
 fn try_parse_control_next_turn_suffix(_text: &str, rest: &str) -> Option<(TargetFilter, bool)> {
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed the current head `2a1ba21782e2f5f6e7bb936b08a91273b0e42c49`.

This is at the existing control-next-turn parser seam: the PR extends the shared suffix parser for the CR 723.1 duration wording instead of adding a Construct-specific arm, and `ControlNextTurn` already exposes the parsed target through the normal target-slot machinery. I verified the apparent `target opponent` shape against the existing target engine: the empty typed opponent filter is the repository's player-target representation for opponent players, and target legality resolves it to `TargetRef::Player(P1)`.

The tests cover the parser shape for the new possessive axis, the full-card Construct parse no longer carrying an `Unimplemented`, and the resolved turn-control subsystem through `start_next_turn`. CI is green/CLEAN, the required `feature` label is present, and the repush guard held.

PR #4080, opened 2026-06-21
feat(parser): The Kingpin of Crime — duration-bound, filter-scoped "assign combat damage by toughness"

lgray:

## Card
**The Kingpin of Crime** (MSH, {W}{B}):
> Extort (Whenever you cast a spell, you may pay {W/B}. If you do, each opponent loses 1 life and you gain that much life.)
> Whenever you attack, you may pay 2 life. If you do, until end of turn, creatures you control with toughness greater than their power assign combat damage equal to their toughness rather than their power.

Extort and the attack-trigger shell (optional `PayLife{2}`, `IfYouDo`/`OptionalEffectPerformed` gate, `UntilEndOfTurn` duration) already parsed. The only residual gap on main was the effect-side continuous clause:
```
Unimplemented[creatures]: "creatures you control with toughness greater than their power assign combat damage equal to their toughness rather than their power"
```

## Class built (not the card)
A **duration-bound, filter-scoped "assign combat damage by toughness" continuous effect**, composed entirely from existing infra:

1. **Predicate (singular + plural).** `parse_continuous_modifications` recognized only the singular `"assigns … its … its"` form; it now accepts both via a shared combinator `parse_assigns_damage_from_toughness_predicate`, keeping the static-line parser (Doran / Assault Formation / High Alert) and the one-shot continuous-effect parser in lockstep. Reuses `ContinuousModification::AssignDamageFromToughness`.
2. **Self-referential P/T filter (singular + plural).** `parse_pt_comparison` now short-circuits `"(toughness|power) greater than <its|their> (power|base power)"` to the dedicated `FilterProp::ToughnessGTPower` / `PowerExceedsBase` props for **both** possessives, before the general numeric comparison tail. Previously only `parse_with_inner` handled this; the `parse_target` path used by subject-scoped effects bypassed it and resolved "their power" through the quantity grammar as the **source** object's power (wrong scope) — so a 2/4 was compared against the source's power instead of its own. This is now the single authority for the self-referential comparison.
3. **Duration + optional cost + trigger:** reuse `Duration::UntilEndOfTurn`, the optional-cost attack-trigger shape, and the `Effect::GenericEffect` continuous-effect registration path. Nothing new.

## add-engine-variant verdict
**No new variant.** Reuses `AssignDamageFromToughness`, `ToughnessGTPower`, `PowerExceedsBase`, `Duration::UntilEndOfTurn`, and the `GenericEffect` path. The fix is parser-side normalization (singular↔plural) plus routing the self-referential filter to the existing dedicated props — parameterize/reuse, do not proliferate.

## Grep-verified CRs
- **CR 510.1a** — "Each attacking/blocking creature assigns combat damage equal to its power" (the default the effect modifies).
- **CR 613.11** — continuous effects that affect game rules (the toughness-rather-than-power substitution).
- **CR 208.1** — a creature's power/toughness (the self-referential comparison filter).

All three confirmed present in `docs/MagicCompRules.txt`.

## Discriminating-test map (drive `apply()`: parse → attack → pay 2 life → combat damage)
`crates/engine/tests/integration/the_kingpin_of_crime_combat_damage.rs`:

| Test | Asserts | Revert-fails because |
|---|---|---|
| `kingpin_paid_two_four_assigns_toughness_damage` | 2/4 (toughness>power), paid → P1 to **16** (assigns 4) | reverted fix lowers clause to `Unimplemented` → 2/4 assigns power (2) → P1 at 18 |
| `kingpin_paid_four_two_fails_filter_assigns_power` | 4/2 (power>toughness) fails filter → assigns power (4); flag NOT set | reverted **filter** fix would compare each creature vs source power, mis-scoping membership |
| `kingpin_declined_two_four_assigns_power_damage` | decline cost → no life paid, flag NOT set, 2/4 assigns power (2) → P1 at **18** | gates the `IfYouDo`/`OptionalEffectPerformed` cost |
| `kingpin_effect_lapses_after_end_of_turn` | grant live this turn, cleared after the turn rolls over | gates `Duration::UntilEndOfTurn` (CR 514.2) vs permanent |

Plus parser shape guards (backing, not substituting, the runtime tests):
- `the_kingpin_of_crime_full_card_has_no_unimplemented` — full card text → **0 residual `Unimplemented`**.
- `kingpin_clause_lowers_to_assign_damage_from_toughness` — clause → `GenericEffect` with `AssignDamageFromToughness`, `controller=You` + `ToughnessGTPower` filter, `UntilEndOfTurn`.

The `4/2` and `2/4` fixtures reach the real `register_transient_effect` broadcast arm and the per-candidate filter evaluation (no degenerate fixture); the positive/negative pair distinguishes the filter scope, the cost gate, and the duration independently.

## Gate results (nightly cargo proxy; Tilt not running for /private/tmp worktrees)
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` (no base **and** `upstream/main`) — exit 0
- parser-diff string-dispatch grep (my changed files) — **clean**
- CR-annotation diff gate (my changed files) — all of CR 208.1 / 510.1a / 613.11 present, **0 UNVERIFIED**
- `cargo test -p engine` — **13079 passed, 1 failed**; the lone failure is the known parallel-flaky `ai_support::tests::delve_payment_skips_state_clone_per_graveyard_candidate` (state-clone counting), which **passes in isolation** and is unrelated to this diff.
- New Kingpin suite: **5/5 pass**.

`cargo coverage` / `cargo semantic-audit` require the generated `card-data.json` (absent in a fresh worktree); not regenerated here, so `known-tokens.toml` was left untouched. No new `// Priority N:` parser slot was added.

Test summary line: `test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 1440 filtered out` (Kingpin suite); `test result: FAILED. 13079 passed; 1 failed` (full engine, lone flaky confirmed green in isolation).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -22080,6 +22080,55 @@ mod tests {
     use super::*;
     use crate::parser::parse_oracle_text;
 
+    /// CR 510.1a + CR 613.11: The Kingpin of Crime's attack ability — "Whenever you
+    /// attack, you may pay 2 life. If you do, until end of turn, creatures you
+    /// control with toughness greater than their power assign combat damage equal
+    /// to their toughness rather than their power." The plural surface form of the
+    /// damage-by-toughness predicate (and the plural "toughness greater than their
+    /// power" filter) must lower with NO residual `Effect::Unimplemented`.
+    #[test]
+    fn the_kingpin_of_crime_full_card_has_no_unimplemented() {
+        let parsed = parse_oracle_text(
+            "Extort (Whenever you cast a spell, you may pay {W/B}. If you do, each opponent loses 1 life and you gain that much life.)\n\
+             Whenever you attack, you may pay 2 life. If you do, until end of turn, creatures you control with toughness greater than their power assign combat damage equal to their toughness rather than their power.",
+            "The Kingpin of Crime",
+            &["Extort".to_string()],
+            &["Legendary".to_string(), "Creature".to_string()],
+            &["Spider".to_string()],
+        );
+
+        fn effect_has_unimplemented(effect: &Effect) -> bool {
+            matches!(effect, Effect::Unimplemented { .. })
+        }
+        fn ability_has_unimplemented(ability: &AbilityDefinition) -> bool {
+            effect_has_unimplemented(&ability.effect)
+                || ability
+                    .sub_ability
+                    .as_deref()
+                    .is_some_and(ability_has_unimplemented)
+                || ability
+                    .else_ability
+                    .as_deref()
+                    .is_some_and(ability_has_unimplemented)
+        }
+
+        for trigger in &parsed.triggers {
+            if let Some(execute) = trigger.execute.as_deref() {
+                assert!(
+                    !ability_has_unimplemented(execute),
+                    "trigger lowered to an Unimplemented node: {execute:#?}"
+                );
+            }
+        }
+        for ability in &parsed.abilities {
+            assert!(
+                !ability_has_unimplemented(ability),
+                "ability lowered to an Unimplemented node: {:#?}",
+                ability
+            );
+        }
+    }
+
     /// CR 601.2c + CR 115.1: HONEST DEFERRAL — Graceful Takedown's heterogeneous
     /// compound source set ("any number of target enchanted creatures you control
     /// AND up to one other target creature you control each deal damage equal to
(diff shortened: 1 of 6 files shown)


matthewevans (approved):
Approved after rechecking the current head. The semantic review already covered the Kingpin duration-bound toughness-damage path; the only maintainer follow-up was the import-scope cleanup, and the fresh GitHub CI is green with merge state CLEAN. Required `feature` label is present.

PR #4088, opened 2026-06-21
fix(parser): keep "if ~ entered this turn" intervening-if on triggers

### LINKED ISSUE #4087: Parser: "if ~ entered this turn" intervening-if dropped on triggers (Hixus, Prison Warden exiles every turn)
## Summary

Hixus, Prison Warden ignores its intervening-if and exiles attackers every turn instead of only the turn it entered.

**Hixus, Prison Warden** — "Flash / Whenever a creature deals combat damage to you, **if Hixus entered this turn**, exile that creature until Hixus leaves the battlefield."

## Current (wrong) parse

The trigger parses with mode/effect/duration all correct (exile that creature until Hixus leaves) but the trigger **condition is `null`** — the intervening-if "if Hixus entered this turn" is silently dropped. So Hixus exiles an attacker on every later turn, not just the turn it flashed in (violates CR 400.7).

## Root cause

The parser already lowers "~ entered this turn" to `StaticCondition::SourceEnteredThisTurn`, but the static→trigger condition bridge left that variant in its "no `TriggerCondition` equivalent" group — so the intervening-if was discarded when building the trigger. `TriggerCondition::SourceEnteredThisTurn` already exists and is evaluated at trigger fire-time and rechecked at resolution.

## Expected

Trigger `condition = SourceEnteredThisTurn`, so the exile only happens the turn Hixus entered.

## Notes

Parser-only — no new engine variant, no runtime change (the trigger-side source-entered check is already wired and used by e.g. Moon-Circuit Hacker). The "if **that creature** entered this turn" form (triggering object, e.g. Samut, Vizier of Naktamun) references the triggered object rather than the source and would need a separate variant. CR 400.7 / 603.4.

galuis116:
Fixes #4087.

## Summary

**Hixus, Prison Warden** — "Whenever a creature deals combat damage to you, if Hixus entered this turn, exile that creature until Hixus leaves the battlefield" — dropped its intervening-if (trigger `condition: null`), so it exiled an attacker on **every** later turn instead of only the turn it flashed in (violates CR 400.7).

## Root cause

The parser already lowers "~ entered this turn" to `StaticCondition::SourceEnteredThisTurn`, but the static→trigger condition bridge (`static_condition_to_trigger_condition`) left that variant in its "no `TriggerCondition` equivalent" group — so the intervening-if was discarded when building the trigger.

## Fix (parser-AST-only; no new engine variant, no runtime change)

One-arm relocation: route `StaticCondition::SourceEnteredThisTurn` to its existing twin `TriggerCondition::SourceEnteredThisTurn`. That condition is already evaluated at trigger fire-time and rechecked at resolution (CR 603.4) against `GameObject.entered_battlefield_turn` (proven by the supported sibling Moon-Circuit Hacker), so the fix simply connects Hixus to pre-existing runtime machinery — not a parses-but-dead flip.

The "if **that creature** entered this turn" form (triggering object, e.g. Samut, Vizier of Naktamun) references the triggered object rather than the source and would need a separate variant; left out of scope. CR 400.7 / 603.4.

## Tests

- Discriminating (fail-on-revert): full `parse_trigger_lines` of Hixus's oracle → a `DamageDone` trigger with `condition = Some(SourceEnteredThisTurn)` (was `None`).
- Bridge unit test: `SourceEnteredThisTurn` now bridges to `Some(TriggerCondition::SourceEnteredThisTurn)`.

## Verification

`cargo +nightly test -p engine --lib`: 13082 passed (incl. the new tests); the only lib failure was the pre-existing flaky `ai_support::tests::delve_payment_skips_state_clone_per_graveyard_candidate` (perf-counter race, confirmed passing in isolation). `--test integration`: 1441 passed; the 2 `anaphoric_scope_allowlist_guard` failures are pre-existing on the base (a stale frozen-set vs recent parser drift, orthogonal to this change — verified via stash that this PR neither causes nor worsens them). `clippy --workspace --all-targets --features engine/proptest -D warnings`, parser-combinator gate, rustfmt: clean. oracle-gen confirms Hixus's trigger condition flips `null` → `SourceEnteredThisTurn`.


matthewevans:
VERDICT: approve pending CI

Reviewed the API diff and isolated worktree at `a4effdfd078fa3add9a6fcbe63d09338c5ca061d`, then updated the branch to current head `ce92c88b03f2dcc181998057a66a7d11ee8fe314` after the repush guard held. This is the right parser bridge: `StaticCondition::SourceEnteredThisTurn` now maps to the existing `TriggerCondition::SourceEnteredThisTurn`, and the trigger pipeline already checks that condition at fire-time and rechecks it at resolution. CR 400.7 and CR 603.4 citations were verified against `docs/MagicCompRules.txt`.

I applied the required `bug` label. Holding formal approval/enqueue until fresh current-head CI finishes.

matthewevans (approved):
Reviewed current head 9ca495958f67e251f830b1a1ac13458dde1661d1 via GitHub API diff and isolated worktree. The Hixus fix is scoped to the existing static-condition-to-trigger-condition seam for SourceEnteredThisTurn, preserves the runtime intervening-if stack/recheck behavior, has parser plus runtime coverage for the entered-this-turn condition, and CI is green/CLEAN on the updated branch. Required bug label is present and the repush guard held before approval.

PR #4091, opened 2026-06-21
feat(parser): spend-this-mana-only-to-[action] restriction (face-down casts, turn-face-up)

lgray:

## Summary

Implements the shared **"Spend this mana only to [action]"** restriction primitive for a cluster of Standard-legal cards, building the action-class as **typed enum leaves** that compose into the existing CR 106.6 mana-spend-restriction disjunction infrastructure — not three card-specific hacks.

Cluster cards (verified each still gaps on main before implementing):
- **Creeping Peeper** — `{T}: Add {U}. Spend this mana only to cast an enchantment spell, unlock a door, or turn a permanent face up.`
- **Overgrown Zealot** — `{T}: Add two mana of any one color. Spend this mana only to turn permanents face up.` (second ability)
- **Tin Street Gossip** — `{T}: Add {R}{G}. Spend this mana only to cast face-down spells or to turn creatures face up.`

## Class built

The repo already had `ManaSpendRestriction::{SpellType, UnlockDoor, Any, ...}` and the runtime `ManaRestriction` + `PaymentContext::SpecialAction` gate (door unlock). This PR adds the two missing action-classes the cluster needs:

| Parse-time leaf | Runtime gate | CR | Enforcement |
|---|---|---|---|
| `FaceDownSpell` | `OnlyForFaceDownSpell` (gated on new `SpellMeta.is_face_down`) | 708.4 | **Live** at the spell-payment seam (`allows_spell`) |
| `TurnPermanentFaceUp` | `OnlyForSpecialAction(SpecialAction::TurnFaceUp)` | 116.2b + 702.37e | **Honest-deferred** (see below) |

`SpellMeta.is_face_down` is populated in `build_spell_meta` from `obj.face_down` (CR 708.4: a face-down spell is on the stack face down). Enchantment-cast and door-unlock leaves reuse the existing `SpellType` / `UnlockDoor` machinery.

### Honest-deferral (STOP-adjacent, contained)

Both `turn-face-up` (`game::morph::turn_face_up`) and face-down "casting" (`game::morph::play_face_down`) currently charge **no mana** in this engine — they flip/place the permanent for free, bypassing the mana-payment pipeline entirely. Wiring real morph-cost payment (incl. CR 702.37f X-cost selection, the legal-actions affordability surface, and ~10 existing free-turn-face-up tests) is the **broad mana-payment rework** the engine already documents as deliberately deferred (the `SpecialAction` doc comment).

This PR takes the **contained primitive**: every action-class is *representable* (all three cards parse to a **complete** `Any` with **0 residual `Effect::Unimplemented`** — no leaf is dropped, so the mana is never over-restricted), and the `TurnFaceUp` runtime gate exists but is **conservatively unsatisfiable** (no payment site emits `PaymentContext::SpecialAction(TurnFaceUp)` yet) — never silently over-permitted. This mirrors the door-unlock variant's own history. When the turn-face-up morph-cost payment seam lands, the gate goes live with no type change.

### Parser

Per-action tags composed into the existing disjunction splitter (`alt`/`tag`, no full-string permutations, no string dispatch). The face-down clause runs **before** the generic type-phrase fallback so `"face-down spells"` is not mis-typed as `SpellType("Face-down")`; a standalone special-action-only restriction (no `"to cast"` head) handles Overgrown Zealot.

## add-engine-variant verdict

Parameterized, not proliferated: `TurnFaceUp` is added to the existing `SpecialAction` enum (parameterizes `OnlyForSpecialAction`, no new sibling `ManaRestriction` for it). `OnlyForFaceDownSpell` is a genuine new categorical axis (face-down status, CR 708.4) distinct from every existing spell-property gate. `is_face_down` is a typed `SpellMeta` field, not a stringly-typed flag. Enchantment reuses `SpellType` / `CardType`. `cargo engine-inventory` regenerates consistently (the committed inventory was stale on main from unrelated prior drift, so it was left untouched to keep the diff focused).

## Grep-verified CRs

- **CR 106.6** — mana spend restrictions
- **CR 116.2b** — "Turning a face-down creature face up is a special action."
- **CR 702.37e** — turn-face-up: "show … the morph cost …, pay that cost, then turn the permanent face up."
- **CR 708.4** — "Objects that are cast face down are turned face down before they are put onto the stack."
- **CR 116.2m / CR 709.5e** — door unlock (existing leaf, reused)

## Discriminating-test map (revert-probed)

| Behavior | Seam | Production entry | Test | Revert-failing assertion |
|---|---|---|---|---|
| face-down-only mana pays a face-down cast, not a face-up cast | `OnlyForFaceDownSpell` → `allows_spell` | `ManaPool::spend_for(PaymentContext::Spell)` | `face_down_spell_mana_consumes_for_face_down_cast_only` | face-up cast withheld (flips if gate returns `true`) |
| Creeping Peeper {U} pays enchantment, not creature | `Any([SpellType(Enchantment), ...])` | `spend_for(Spell)` | `creeping_peeper_mana_consumes_for_enchantment_not_creature` | creature cast withheld |
| Creeping Peeper {U} pays door-unlock special action | `OnlyForSpecialAction(UnlockDoor)` | `spend_for(SpecialAction(UnlockDoor))` | `creeping_peeper_mana_pays_door_unlock_special_action` | unlock consumes the {U} |
| turn-face-up mana rejects every live context | `OnlyForSpecialAction(TurnFaceUp)` | `spend_for(Spell/SpecialAction)` | `overgrown_zealot_turn_face_up_mana_rejects_every_live_context` | spell/door withheld |
| all 3 cards parse to complete `Any`, 0 Unimplemented | parser disjunction | `parse_oracle_text` | `creeping_peeper_full_mana_line_no_unimplemented`, `overgrown_zealot_turn_face_up_only_no_unimplemented`, `tin_street_gossip_face_down_or_turn_face_up_no_unimplemented`, `mana_spend_restriction_noncast_action_class_leaves_parse_completely` | restriction shape (flips if turn-face-up clause disabled) |

Revert-probes confirmed: breaking `OnlyForFaceDownSpell` flips the face-up assertion; disabling `parse_turn_face_up_clause` flips both Creeping Peeper and Overgrown Zealot parse tests.

The `spend_for` route is the real payment route the cast pipeline (`can_pay_for_spell`/`pay_cost_*`) and special-action pipeline (`pay_special_action_mana_cost`) both flow through — the same `ManaRestriction::allows` authority a full `apply()` cast/unlock exercises.

## Gates

- `cargo fmt --all` — clean
- `cargo test -p engine` — 13093 passed; only the known parallel-flaky `ai_support::delve_payment_skips_state_clone_per_graveyard_candidate` fails under parallelism (passes in isolation)
- `./scripts/check-parser-combinators.sh` — pass
- parser-diff string-dispatch grep — clean
- CR-annotation diff gate — no UNVERIFIED

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -1422,6 +1422,9 @@ pub(super) fn build_spell_meta(
         // detected from shards (mana value alone can't reveal it — X contributes 0
         // off the stack).
         has_x_in_cost: obj.mana_cost.has_x(),
+        // CR 708.4: whether this spell is being cast face down (morph/disguise/
+        // cloak), consulted by the "cast face-down spells" spend restriction.
+        is_face_down: obj.face_down,
     })
 }
 
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -1422,9 +1422,24 @@ pub(super) fn build_spell_meta(
         // detected from shards (mana value alone can't reveal it — X contributes 0
         // off the stack).
         has_x_in_cost: obj.mana_cost.has_x(),
-        // CR 708.4: whether this spell is being cast face down (morph/disguise/
-        // cloak), consulted by the "cast face-down spells" spend restriction.
-        is_face_down: obj.face_down,
+        // CR 708.4 + CR 702.37c / CR 702.168b: `is_face_down` means "this spell is
+        // being CAST FACE DOWN" (morph/disguise/cloak — paying {3} to cast as a 2/2
+        // face-down creature spell), NOT "the object is currently face down". Those
+        // differ: foretell (CR 702.143a), hideaway, and other exile/library
+        // concealment set `obj.face_down = true` while the card waits in exile, yet
+        // such a card is CAST FACE UP (CR 702.143c: cast "even if it was cast for a
+        // cost other than a foretell cost"). Mana payment runs against the origin
+        // (exile) zone BEFORE the deferred origin->stack move clears `face_down`, so
+        // sourcing this from raw `obj.face_down` would let a face-up foretold/hideaway
+        // cast wrongly satisfy the `OnlyForFaceDownSpell` spend restriction (Tin
+        // Street Gossip). No engine path casts a spell face down today:
+        // `GameAction::PlayFaceDown` -> `game::morph::play_face_down` moves
+        // hand->battlefield via the zone pipeline and charges no mana, never building
+        // a `PaymentContext::Spell`. So the correct value at every current production
+        // payment site is `false`. When a real CR 702.37c / 708.4 face-down CAST path
+        // is built, set this from that cast's announced face-down intent; the gate
+        // (`ManaRestriction::allows_spell`) already reads this field.
+        is_face_down: false,
     })
 }
 
@@ -14145,6 +14160,39 @@ mod tests {
             [CastingPermission::Foretold { cost, turn_foretold }]
                 if *cost == foretell_test_cost() && *turn_foretold == state.turn_number
         ));
+    }
+
+    // CR 708.4 + CR 702.143a / CR 702.143c: a FORETOLD card is cast FACE UP from
+    // exile, never as a face-down spell. Mana payment runs `build_spell_meta` while
+    // the object still sits in exile with `obj.face_down == true`; the resulting
+    // `SpellMeta.is_face_down` MUST be false so `OnlyForFaceDownSpell` mana (Tin
+    // Street Gossip) is NOT treated as eligible for the foretold cast. Discriminating
+    // revert: if `build_spell_meta`'s `is_face_down` reverts to `obj.face_down`, this
+    // flips to true (obj.face_down is true after foretell).
+    #[test]
+    fn build_spell_meta_for_foretold_card_is_not_face_down() {
+        let mut state = setup_game_at_main_phase();
+        let object_id = add_foretell_sorcery(&mut state);
+        add_mana(&mut state, PlayerId(0), ManaType::Colorless, 2);
+        handle_foretell(
+            &mut state,
+            PlayerId(0),
+            object_id,
+            CardId(143),
+            &mut Vec::new(),
+        )
+        .unwrap();
+
+        assert!(
+            state.objects.get(&object_id).unwrap().face_down,
+            "foretell must set obj.face_down (precondition for the discriminating assert)"
+        );
+        let meta = build_spell_meta(&state, PlayerId(0), object_id)
+            .expect("foretold object in exile must build a spell meta");
+        assert!(
+            !meta.is_face_down,
+            "a foretold (face-up) cast must NOT be reported as a face-down CR 708.4 cast"
+        );
     }
 
     #[test]
--- a/crates/engine/src/game/effects/mana.rs
+++ b/crates/engine/src/game/effects/mana.rs
@@ -324,7 +324,12 @@ pub(crate) fn resolve_restrictions(
                 crate::types::mana::SpecialAction::UnlockDoor,
             )),
             // CR 106.6 + CR 708.4: Lower the face-down-cast leaf into the runtime
-            // gate checked against `SpellMeta.is_face_down` by `allows_spell`.
+            // gate checked against `SpellMeta.is_face_down` by `allows_spell`. The
+            // gate reads cast face-down intent (not `obj.face_down`), so it
+            // correctly rejects exile-concealment casts (foretell/hideaway, whose
+            // `obj.face_down = true` but which are cast face up, CR 702.143c). It is
+            // fail-closed: no production path casts a spell face down, so the gate
+            // never over-permits.
             ManaSpendRestriction::FaceDownSpell => Some(ManaRestriction::OnlyForFaceDownSpell),
             // CR 106.6 + CR 116.2b + CR 702.37e: Lower the turn-face-up
             // special-action leaf into the runtime gate. No payment site emits
--- a/crates/engine/src/game/effects/mana.rs
+++ b/crates/engine/src/game/effects/mana.rs
@@ -323,6 +323,18 @@ pub(crate) fn resolve_restrictions(
             ManaSpendRestriction::UnlockDoor => Some(ManaRestriction::OnlyForSpecialAction(
                 crate::types::mana::SpecialAction::UnlockDoor,
             )),
+            // CR 106.6 + CR 708.4: Lower the face-down-cast leaf into the runtime
+            // gate checked against `SpellMeta.is_face_down` by `allows_spell`.
+            ManaSpendRestriction::FaceDownSpell => Some(ManaRestriction::OnlyForFaceDownSpell),
+            // CR 106.6 + CR 116.2b + CR 702.37e: Lower the turn-face-up
+            // special-action leaf into the runtime gate. No payment site emits
+            // `PaymentContext::SpecialAction(TurnFaceUp)` yet, so the gate is
+            // conservatively unsatisfiable — honest-deferred, never over-permitted.
+            ManaSpendRestriction::TurnPermanentFaceUp => {
+                Some(ManaRestriction::OnlyForSpecialAction(
+                    crate::types::mana::SpecialAction::TurnFaceUp,
+                ))
+            }
             // CR 106.6: Disjunction — recursively lower each branch. If every branch
             // dropped (e.g. an unresolvable `ChosenCreatureType` with no chosen type),
             // the disjunction has no payable cases, so drop it too.
--- a/crates/engine/src/game/mana_abilities.rs
+++ b/crates/engine/src/game/mana_abilities.rs
@@ -3525,6 +3525,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let goblin_ctx = PaymentContext::Spell(&goblin_spell);
         let mut pool_clone = pool.clone();
@@ -3543,6 +3544,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let elemental_ctx = PaymentContext::Spell(&elemental_spell);
         assert!(
--- a/crates/engine/src/game/mana_payment.rs
+++ b/crates/engine/src/game/mana_payment.rs
@@ -2842,6 +2842,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let elf_ctx = PaymentContext::Spell(&elf);
         assert!(can_pay_for_spell(
@@ -2864,6 +2865,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let goblin_ctx = PaymentContext::Spell(&goblin);
         assert!(!can_pay_for_spell(
@@ -2911,6 +2913,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let thought_knot_ctx = PaymentContext::Spell(&thought_knot);
         assert!(can_pay_for_spell(
@@ -2932,6 +2935,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let colored_eldrazi_ctx = PaymentContext::Spell(&colored_eldrazi);
         assert!(!can_pay_for_spell(
@@ -2988,6 +2992,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let colored_spell_ctx = PaymentContext::Spell(&colored_spell);
         assert!(!can_pay_for_spell(
@@ -3063,6 +3068,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         assert!(
             !can_pay_for_spell(
@@ -3101,6 +3107,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         assert!(
             can_pay_for_spell(
@@ -3198,6 +3205,7 @@ mod tests {
             mana_value: None,
             color_count: None,
             has_x_in_cost: false,
+            is_face_down: false,
         };
         let flashback_ctx = PaymentContext::Spell(&flashback_spell);
         assert!(can_pay_for_spell(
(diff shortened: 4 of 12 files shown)


matthewevans (changes requested):
Blocking this on a production-path mismatch.

Confirmed evidence from the current head (`cfa8d437106d9168ceb90055489b033bc1caca57`):

- `GameAction::PlayFaceDown` is handled directly in `crates/engine/src/game/engine.rs:4309-4325` and calls `morph::play_face_down(...)` without entering the spell payment path.
- `morph::play_face_down` in `crates/engine/src/game/morph.rs:123-180` validates controller/hand and moves the object to the battlefield face down via the zone pipeline; it never charges `{3}`, never builds `PaymentContext::Spell`, and never reaches the new `SpellMeta.is_face_down` gate.
- The new regression in `crates/engine/tests/restricted_mana_face_down_and_face_up.rs:16-22` says it drives the same decision a full cast makes, but the face-down test at lines 48-80 only constructs `SpellMeta { is_face_down: true }` and calls `ManaPool::spend_for` directly. That proves the enum gate works in isolation, not that a face-down play/cast can consume the restricted mana in the engine.

So the new `ManaSpendRestriction::FaceDownSpell` path is currently false-live for the important card text: Tin Street Gossip's “Spend this mana only to cast face-down spells...” is parsed and can pass coverage, but the actual face-down action still bypasses mana payment entirely. That means the PR can mark the parser/card cluster supported while game play still lets face-down play happen for free and never consumes the restricted mana. This is exactly the kind of false green we need to avoid for these parser/support PRs.

What would make this acceptable:

- Route face-down casting/play through a restriction-aware payment path that emits `PaymentContext::Spell` with `SpellMeta.is_face_down = true` before the face-down object is moved, or
- Keep this as an explicitly unsupported/deferred parse shape so coverage does not claim live support for cards whose only new behavior is face-down restricted spend.

The turn-face-up half is already honestly deferred in the test comments; the face-down half needs the same honesty or the actual production wiring. Confidence: high, because the reducer and morph helper show the full current path and the test currently bypasses it. The evidence that would contradict this is a production call path from `PlayFaceDown` into `pay_*mana*`/`PaymentContext::Spell` before `morph::play_face_down`; I do not see one in this head.

lgray:

Thanks for the careful review — you were right that the face-down half was false, and tracing it surfaced something stronger than a coverage flag: the gate was **actively over-permitting**, not just unreachable. Pushed `72c3c7d58` (rebased on `main`).

### What the trace found
`SpellMeta.is_face_down` was sourced from raw `obj.face_down` in `build_spell_meta` (`casting.rs`). But `obj.face_down` is also set true by exile/library **concealment** — foretell (`casting.rs:957`, CR 702.143a), hideaway (`effects/hideaway.rs`), exile-top, heist — and those cards are **cast face up** from exile (CR 702.143c: "even if it was cast for a cost other than a foretell cost"). Mana is paid against the origin (exile) zone **before** the deferred origin→stack move clears `face_down` (`casting_costs.rs` `finalize_cast` → `pay_mana_cost_with_choices` runs first; `zones.rs:256` clears on Exile-exit). So a **foretold/hideaway cast** had `OnlyForFaceDownSpell` mana (Tin Street Gossip) treated as eligible. That's the real false-green, and it's a correctness bug, not just a coverage cosmetic.

### The fix (resolution along your option B, but it also closes a live bug)
CR 708.4 "face-down spell" means a spell **cast** face down (morph/disguise/cloak), not "the object is currently face down". No engine path casts a spell face down today (`PlayFaceDown` → `morph::play_face_down` is a battlefield entry that charges no mana, never builds `PaymentContext::Spell`), so the rules-correct value at every production payment site is `false`. `build_spell_meta` now sets `is_face_down: false` behind a one-line forward-compat seam for a future face-down-cast path. The gate is now genuinely fail-closed **and** correctly rejects concealment casts — fixing the over-permit for the whole concealment class (foretell/hideaway/exile-top/heist), not just one card.

Discriminating coverage:
- `build_spell_meta_for_foretold_card_is_not_face_down` (engine inline) drives the real `handle_foretell` flow + the same `build_spell_meta` the cast pipeline calls; pins `obj.face_down == true` as a precondition, asserts the resulting `is_face_down == false`. Reverting the fix to `obj.face_down` makes it fail (verified both directions).
- The integration test you flagged was rewritten from the fabricated `consumes_for_face_down_cast_only` into `face_down_spell_mana_rejects_every_production_context` — it asserts the unit is **withheld** for face-up cast / door-unlock / activation, and checks the positive **only** at the restriction level, labeled as the unreachable future cast path. No more fabricated production consumption.

Also corrected a bogus `CR 707.2` (that's "Copying Objects") → `CR 702.37c` (morph `{3}` cast cost) on the FaceDownSpell docs, and removed the now-false "never observed true at a payment site" doc framing.

### Two things to flag for your call
1. **Coverage classification.** Tin Street Gossip still shows **supported/green**, because `coverage.rs` classifies on parser-AST (`Effect::Mana`, restrictions discarded — same as every other restriction incl. the already-merged `UnlockDoor`/spell-type ones), not on runtime reachability. With the over-permit fixed, "supported" is now honest: the line parses and the runtime gate is correct (fail-closed, no over-permit). If you'd prefer coverage to show **red** for parser-supported cards whose restriction is runtime-deferred, that's a separate `coverage.rs` change with cross-cutting effects on every honest-deferred restriction — happy to do it as a follow-up if that's the intent.
2. **`engine-inventory.json` staleness.** This PR's feat commit added the new `ManaSpendRestriction`/`ManaRestriction` variants without a `cargo engine-inventory` regen, so the generated inventory is stale (a regen is a file-wide line-drift diff, not a surgical edit). I left it untouched in this fix; let me know if you'd like the regen folded into this PR or handled separately.

The turn-face-up half stays honestly deferred (no `PaymentContext::SpecialAction(TurnFaceUp)` is ever emitted), and the live halves (enchantment-cast spell-type, door-unlock) are unchanged and still tested against the real payment path.

matthewevans (changes requested):
VERDICT: request-changes

Prior finding status: PARTIAL / still blocked. The follow-up correctly avoids treating concealed exile cards as face-down casts, but it still does not make the claimed face-down / turn-face-up spend paths reachable in production.

**[HIGH]** This makes unsupported mana restrictions look fully supported by parser coverage. Evidence: `crates/engine/src/game/casting.rs:1425-1442` hardcodes `SpellMeta.is_face_down: false` and states no production path casts a spell face down through spell payment; `crates/engine/src/game/effects/mana.rs:334-342` lowers `TurnPermanentFaceUp` to `PaymentContext::SpecialAction(TurnFaceUp)` while stating no payment site emits that context; yet `crates/engine/src/parser/oracle.rs:11042-11097` asserts Overgrown Zealot and Tin Street Gossip parse with no `Effect::Unimplemented`. Why it matters: those cards' restriction text is now marked handled even though the restricted mana is conservatively unspendable for every action named by the card, so coverage becomes a false green rather than an honest unsupported signal. Suggested fix: keep pure face-down / turn-face-up-only restriction text as an honest parser gap until the face-down cast and turn-face-up payment seams actually emit the matching payment contexts, or limit the parser success to disjunctions with at least one currently live payable branch while preserving unsupported coverage for the dead-only cases.

`cargo fmt --all --check` passes on this head.

lgray:

Thanks — you were right, and the trace turned up a third dead branch beyond the two you named. Pushed `35abe7344` (rebased on current `main`).

### The fix: a class-level liveness gate, not a per-card patch
The root cause is that the `Effect::Mana` parser seam absorbed **any** parsed `ManaSpendRestriction` into a "supported" effect, regardless of whether the restricted mana is spendable at runtime. I added `ManaSpendRestriction::has_payable_branch` (CR 106.6) and gated the seam on it: a restriction with no payable branch is left unabsorbed and lowers to `Effect::Unimplemented` (honest coverage red); a restriction with ≥1 payable branch stays supported. Liveness is decided **once on the fully-assembled restriction**, so a mixed disjunction keeps its live branches.

The predicate uses a single consistent rule applied uniformly to all 17 variants:

> A branch is **payable** iff its lowered `ManaRestriction` runtime gate can return `true` at some reachable production payment site (data-value reachability) — i.e. there exists a real spend the engine would accept for it today. `Any(subs)` is payable iff any sub is.

The match is exhaustive with **no wildcard arm** so a future variant must classify its own liveness (compile error otherwise).

### What's dead today (→ `false`)
- **`TurnPermanentFaceUp`** — lowers to `OnlyForSpecialAction(TurnFaceUp)`; the only production `PaymentContext::SpecialAction` emit (`game/engine.rs`) is `UnlockDoor`. `TurnFaceUp` is never emitted. (CR 116.2b / CR 702.37e)
- **`FaceDownSpell`** — gate is `meta.is_face_down`, which `build_spell_meta` never sets `true` at a payment site (the prior fix's intent — preserved). (CR 708.4)
- **`XCostOnly`** — `OnlyForXCosts` returns `false` unconditionally in both `allows_spell` and `allows_activation` (no `{X}`-in-cost check at any call site). This was the same false-green class as the two you flagged, surfaced while building the predicate, so it's gated too. (Note: the live "MV≥N **or** has `{X}`" cards — Helga, Troyan — route through `SpellMatchingCostCriteria`/`HasXInCost`, which reads the populated `meta.has_x_in_cost`, so they stay supported.)

Everything else (spell-type, MV/color-count/zone/keyword, activation, `UnlockDoor`, `SpellMatchingCostCriteria`) stays live and supported.

### Result
- **Overgrown Zealot** (bare `TurnPermanentFaceUp`) and **Tin Street Gossip** (`Any([FaceDownSpell, TurnPermanentFaceUp])`, all-dead) → now honest `Effect::Unimplemented`.
- **Creeping Peeper** / **Smoky Lounge** (mixed `Any` with live `SpellType`/`UnlockDoor`) → stay supported (over-gapping guards).
- When a real face-down-cast (`{3}`, CR 702.37c), turn-face-up payment (CR 702.37e), or `{X}`-cost gate is built, flipping that one arm to `true` re-greens the class with no other change.

### Discriminating tests (both directions verified)
- The two card tests are inverted to assert `Effect::Unimplemented` IS present (`overgrown_zealot_..._is_unsupported_gap`, `tin_street_gossip_..._is_unsupported_gap`).
- `creeping_peeper_full_mana_line_no_unimplemented` / `smoky_lounge_...` kept unchanged as over-gapping guards.
- `has_payable_branch_distinguishes_live_and_dead_leaves` (unit) pins LIVE→true, DEAD→false, all-dead-`Any`→false, mixed-`Any`→true.
- A scratch-revert of the seam guard (`if true`) was run: the two `_is_unsupported_gap` tests fail, the two guards still pass — then restored. (Verified independently.)

The leaf parsers, `casting.rs` `is_face_down` source (your prior-fix concealment fail-closed), and the `types/mana.rs` runtime gates are unchanged. Also regenerated the stale `engine-inventory.json` (it was missing `FaceDownSpell`/`TurnPermanentFaceUp`/`OnlyForFaceDownSpell` from the feat commit).

Note: `cargo coverage` couldn't enumerate the exact card-count delta in this environment (no local `card-data.json`), so the green→red flips are pinned by the per-test evidence above rather than a coverage-count diff. The turn-face-up / face-down / X-cost payment seams stay honestly deferred — the `Effect::Unimplemented` is exactly the signal waiting for them.

matthewevans (commented):
VERDICT: pending-ci/currentness

The previous requested-changes finding is addressed on `35abe7344`. Evidence from this head:

- `crates/engine/src/types/ability.rs` adds `ManaSpendRestriction::has_payable_branch()` with an exhaustive match and classifies `XCostOnly`, `FaceDownSpell`, and `TurnPermanentFaceUp` as non-payable today.
- `crates/engine/src/parser/oracle_effect/sequence.rs:4586-4606` now only absorbs a parsed spend restriction into an `Effect::Mana` follow-up when that fully assembled restriction has at least one payable branch.
- `crates/engine/src/parser/oracle.rs` now keeps mixed live/deferred cases such as Creeping Peeper green, while asserting dead-only Overgrown Zealot and Tin Street Gossip surface an `Effect::Unimplemented` gap.
- The face-down spell gate remains fail-closed: `crates/engine/src/game/casting.rs` sets `SpellMeta.is_face_down` from cast intent (currently `false` at production sites), not from concealment `obj.face_down`, and the foretell regression test covers that distinction.

I am not approving/enqueueing yet because GitHub reports `mergeStateStatus=DIRTY`, and the check rollup currently only contains the AI-CONTRIBUTOR triage label job. A local merge-tree probe also reports changed-in-both/currentness pressure in the touched engine/parser files, so this needs a clean rebase/update and normal CI signal before maintainer approval.

One minor cleanup to fold into the rebase if you touch the docs/comments anyway: a couple of the new variant docs still use the old phrasing that the cluster's cards parse to no `Effect::Unimplemented`; after this fix, the dead-only full-card cases intentionally do produce an `Unimplemented` gap. The behavior is correct; that wording is now stale.

lgray:

Rebased onto current `main` to resolve the merge conflict. **New head `76e6d05e4`** (`gh pr view` now reports `mergeable: MERGEABLE` — no longer `CONFLICTING`).

### Conflict resolution
The only conflict was `data/engine-inventory.json` (`main` advanced while this branch sat). The three source commits (`c689b2623` feat, `a8da341d8` is_face_down-from-cast-intent, `35abe7344` → `76e6d05e4` liveness gate) **auto-merged cleanly with zero source conflicts** — only the auto-generated inventory collided on line churn.

Resolved by **clean regen, not hand-merge**: ran `cargo engine-inventory` on the rebased tree, then verified the result is honest rather than swept. The committed inventory is **byte-identical to a fresh regen of this HEAD's source** (no hand-edits), and its per-enum variant-set diff against a *correctly-regenerated* `main` baseline is **exactly the 4 variants this PR adds, with none dropped**:

- `ManaRestriction::OnlyForFaceDownSpell`
- `ManaSpendRestriction::FaceDownSpell`
- `ManaSpendRestriction::TurnPermanentFaceUp`
- `SpecialAction::TurnFaceUp`

(Note: a raw diff against the *committed* base inventory looked larger because the committed base inventory was **stale** — it predated already-merged sticker/attraction/`ApplyPerpetual` variants that exist in `main`'s source. The regen incidentally reconciles that staleness; the source diff of this PR adds 4 variants and removes 0.)

### Fix integrity (re-verified post-rebase)
- `ManaSpendRestriction::has_payable_branch` (CR 106.6) — exhaustive `match`, no wildcard arm; seam at `oracle_effect/sequence.rs` absorbs the restriction only when `has_payable_branch()` is true, else lowers to `Effect::Unimplemented`. Intact.
- `is_face_down`-from-cast-intent (`build_spell_meta`, `casting.rs`) — sourced from cast-face-down intent (`false` today), not `obj.face_down`; the foretell hostile-fixture test is intact and discriminating.
- All CR annotations grep-verify against the Comprehensive Rules; all discriminating tests green (lib `13267 passed / 0 failed`; `restricted_mana_*` integration `4/4` + `12/12`); `clippy -D warnings` clean; `fmt` clean. No source edits were needed by the rebase.

The latest blocking review's requested resolution (gate parser success on ≥1 live payable branch, keep dead-only cases as honest unsupported coverage) is the fix already in the third commit and survives the rebase unchanged.

matthewevans (approved):
Approved current head b82a510ca1694c544e0ad12619a8d642fa3187ae after repush guard, green CI, and prior review blockers addressed.

Issue #3302, opened 2026-06-14
Breach the Multiverse not supported — The part of Breach the Multiverse that's not supported is the choice of creature…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1515317925932699648/1515317925932699648

**Thread:** Breach the Multiverse not supported
**Cards:** breach the multiverse
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
The part of Breach the Multiverse that's not supported is the choice of creature or planeswalker card for each player.

### breach the multiverse
Each player mills ten cards. For each player, choose a creature or planeswalker card in that player's graveyard. Put those cards onto the battlefield under your control. Then each creature you control becomes a Phyrexian in addition to its other types.

---
<sub>report_id: `discord:1515317925932699648:1515317925932699648:0` · discord: `1515317925932699648/1515317925932699648`</sub>

dripsmvcp:
Looking into this — opening a PR shortly.

matthewevans:
Additional Discord report: Breach the Multiverse stalls on the second graveyard choice with a card-not-in-set error. Reopening as a fresh reproduction/regression signal.

Discord: https://discord.com/channels/1485498006781427802/1519359341038407680/1519359341038407680

report_id: discord:1519359341038407680:1519359341038407680:0

matthewevans:
should be resolved

PR #4093, opened 2026-06-21
feat(parser): per-other-player iterated exile (Kaya, Spirits' Justice −2)

lgray:

## Class built: per-player iterated target/choice over a battlefield `TargetFilter`

This implements the per-opponent leaf of the per-player iterated target/choice axis and ships **Kaya, Spirits' Justice's −2**. The runtime per-player choose primitive (`ChooseFromZone { zone_owner: EachPlayer }` — APNAP iteration, accumulate-into-tracked-set, real `apply` pipeline) already existed for Breach the Multiverse (issue #3302); this PR extends it with the each-OTHER-player scope and wires the combined choose+exile parser.

### add-engine-variant verdict
Added one variant: **`ZoneOwner::EachOpponent`** — the each-OTHER-player leaf of the existing `ZoneOwner::EachPlayer` per-player iteration axis (categorical boundary = CR 101.4 player iteration, same as `EachPlayer`/`Opponent`/`ScopedPlayer`). It is driven by the same `prompt_next_each_player` machinery with the controller filtered out of the APNAP queue — not a parallel code path. This is a leaf scope parameterization within one CR section, consistent with the established `ZoneOwner` sibling convention (Controller/TargetedPlayer/Opponent/ScopedPlayer/EachPlayer). Inventory regeneration in this worktree produced only environment-ordering churn plus the one new variant; the noisy reorder was reverted (CI regenerates deterministically).

### grep-verified CR annotations (all confirmed in `docs/MagicCompRules.txt`)
- **CR 101.4** — APNAP order for simultaneous per-player choices.
- **CR 102.2** — "each other player" / opponent (controller excluded from the iteration).
- **CR 400.7 + CR 608.2c** — the chosen permanents (the tracked set) are exiled as a mass move.
- **CR 601.2c / CR 608.2k** — resolution-scoped "target"/"other" choice modifiers.

### Per-card verdicts
| Card | Verdict | Notes |
|---|---|---|
| **Kaya, Spirits' Justice −2** | **SHIP** | `Exile target creature you control` (pre-existing single-target ChangeZone) → `ChooseFromZone { EachOpponent, up_to }` → `ChangeZoneAll { Exile, TrackedSet }`. 0 residual Unimplemented on the −2. Her static-ability trigger ("choose a creature card from among them") remains a separate, unrelated gap. |
| Winnowing | DEFER (honest) | Choose clause parses, but the payload needs per-player `ParentTarget` binding in the "each player sacrifices all other creatures that don't share a type with the chosen creature they control" sweep — verified at runtime to over-sacrifice. Left Unimplemented rather than ship silently-wrong. |
| Unstable Glyphbridge | DEFER (honest) | Needs a "destroy all creatures **except creatures chosen this way**" tracked-set exclusion (`DestroyAll` has no exclusion field). |
| Kitesail Larcenist | DEFER (honest) | Needs the duration-bound become-Treasure continuous effect to bind to the per-player **chosen set** (it parses against `ParentTarget`, not the tracked set). Plus a separate "Flying, ward {1}" line gap. |
| The Legend of Yangchen I | DEFER (honest) | Different primitive: every player chooses from a **shared** pool (the controller's-opponents' permanents), not each player's own permanents — `EachPlayer` scans each iterating player's own battlefield, so this needs a multi-chooser/shared-pool scope. |

The choose-only battlefield path is **deliberately not wired** for the DEFER cards: emitting their choose clause would produce a silently-wrong chain (payload over-acts), which is strictly worse than an honest residual gap. `cargo semantic-audit` reports **zero findings** on any cluster card.

### Discriminating-test map
| behavioral claim | seam | production entry | test | revert-failing assertion |
|---|---|---|---|---|
| per-OTHER-player exile picks one creature per opponent and exiles all chosen | `ChooseFromZone { EachOpponent }` → `ChangeZoneAll { TrackedSet }` via `apply` | `GameAction::CastSpell` + `GameAction::SelectCards` (3-player runner) | `kaya_minus_two_exiles_one_creature_per_other_player` | P1/P2 chosen creatures → `Zone::Exile`, P1 unchosen + P0 creature stay `Battlefield` |
| controller is never an iterated chooser | `EachOpponent` APNAP filter excludes `ability.controller` | `apply` 3-player, P0-only board | `kaya_minus_two_skips_controller_in_per_other_player_loop` | no `ChooseFromZoneChoice` raised; both P0 creatures stay (revert-probe: `EachOpponent`→`EachPlayer` fails this) |
| full −2 parses to the choose+exile chain | parser dispatch | `parse_oracle_text` on real card | `kaya_minus_two_parses_to_each_opponent_exile_chain` | 0 Unimplemented + `ChooseFromZone { EachOpponent, up_to }` present in the −2 chain |

Revert-probe confirmed: reverting `EachOpponent`→`EachPlayer` in the parser makes the negative test fail (the controller gets prompted).

### Gate results
- `cargo fmt --all` ✅
- `./scripts/check-parser-combinators.sh` ✅ (exit 0)
- parser diff string-dispatch grep ✅ (only hit is `Vec<FilterProp>::contains` dedup — typed membership, not text dispatch)
- `./scripts/check-engine-authorities.sh upstream/main` ✅ (exit 0)
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` ✅
- `cargo test -p engine` ✅ `13081 passed; 2 failed` — the 2 failures are the known parallel-flaky `ai_support` tests (`delve_payment_skips_state_clone_per_graveyard_candidate`, `target_selection_legal_actions_do_not_fall_back_to_stale_slot_targets`); both pass in isolation.
- `cargo coverage` ✅ (exit 0) — `Effect:choose` cluster still lists "choose a creature card from among them", confirming Kaya's static stays honestly gapped.
- `cargo semantic-audit` ✅ (exit 0) — 0 findings on any cluster card.
- `./scripts/gen-card-data.sh` ✅ regenerated from this branch's parser (`known-tokens.toml` restored; only source files committed).

### DIFF
--- a/crates/engine/src/game/effects/choose_from_zone.rs
+++ b/crates/engine/src/game/effects/choose_from_zone.rs
@@ -468,19 +468,25 @@ pub(crate) fn drain_pending_per_player_zone_choice(
         accumulated,
     } = pending;
 
-    // CR 603.7 + CR 608.2c: The FIRST pick of this per-player iteration STARTS a
-    // fresh chosen-card set. It must NOT extend an earlier producer's tracked
-    // set — Breach the Multiverse mills first (publishing a "Milled" set), so
-    // extending here would reanimate the milled cards alongside the chosen ones
-    // ("those cards" = the chosen cards only, CR 608.2c). `publish_fresh_tracked_set`
-    // allocates a new set and rebinds `chain_tracked_set_id`, overwriting the
-    // milled binding. Every LATER pick extends that fresh set so all players'
-    // chosen cards unify under one "those cards" reference. The Cyberman / impulse
-    // "milled this way" path is unaffected — it never uses this per-player drain.
-    let accumulated = if chosen.is_empty() {
-        accumulated
-    } else if accumulated {
-        super::publish_tracked_set(state, chosen.to_vec());
+    // CR 603.7 + CR 608.2c: The FIRST resolution of this per-player iteration
+    // STARTS a fresh chosen-card set — even when that first player declines (an
+    // empty "up to one" pick). It must NOT extend or inherit an earlier
+    // producer's tracked set: Breach the Multiverse mills first (publishing a
+    // "Milled" set), so extending here would reanimate the milled cards alongside
+    // the chosen ones ("those cards" = the chosen cards only, CR 608.2c).
+    // `publish_fresh_tracked_set` allocates a new (possibly empty) set and rebinds
+    // `chain_tracked_set_id`, overwriting the milled binding; binding it on the
+    // first resolution regardless of emptiness guarantees a downstream
+    // `ChangeZoneAll { TrackedSet }` acts only on this iteration's picks (an
+    // all-declined loop → empty fresh set → mass move acts on nothing) instead of
+    // a stale prior producer's set. Every LATER pick extends that fresh set so all
+    // players' chosen cards unify under one "those cards" reference. The Cyberman /
+    // impulse "milled this way" path is unaffected — it never uses this per-player
+    // drain.
+    let accumulated = if accumulated {
+        if !chosen.is_empty() {
+            super::publish_tracked_set(state, chosen.to_vec());
+        }
         true
     } else {
         super::publish_fresh_tracked_set(state, chosen.to_vec());
@@ -2169,5 +2175,80 @@ mod tests {
         );
         assert!(exiled.contains(&kindred_sorcery));
         assert!(exiled.contains(&plain_sorcery));
+    }
+
+    /// CR 603.7 + CR 608.2c (MED #4093): the FIRST resolution of a per-player
+    /// `ChooseFromZone` iteration must rebind a fresh (possibly empty) chain
+    /// tracked set EVEN when that first player declines. Otherwise an earlier
+    /// same-chain producer's tracked set stays bound, and a downstream
+    /// `ChangeZoneAll { TrackedSet }` over-acts on the prior producer's objects
+    /// instead of this iteration's (empty) picks.
+    ///
+    /// REVERT PROBE: restore the original `if chosen.is_empty() { accumulated }`
+    /// branch and the first decline does NOT rebind — `chain_tracked_set_id`
+    /// stays pointed at `prior` (the non-empty producer set), so the two
+    /// assertions below (fresh id distinct from `prior`; bound set size 0) both
+    /// fail.
+    #[test]
+    fn per_player_first_decline_rebinds_fresh_empty_tracked_set() {
+        let mut state = GameState::new_two_player(42);
+
+        // An earlier SAME-CHAIN producer published a NON-EMPTY tracked set and
+        // bound the chain to it (e.g. Breach the Multiverse's preceding mill).
+        let prior = TrackedSetId(state.next_tracked_set_id);
+        state.next_tracked_set_id += 1;
+        state
+            .tracked_object_sets
+            .insert(prior, vec![ObjectId(7), ObjectId(8)]);
+        state.chain_tracked_set_id = Some(prior);
+
+        // A per-player ChooseFromZone whose first (and only) player will decline.
+        let ability = ResolvedAbility::new(
+            Effect::ChooseFromZone {
+                count: 1,
+                zone: Zone::Battlefield,
+                additional_zones: Vec::new(),
+                zone_owner: ZoneOwner::EachOpponent,
+                filter: None,
+                chooser: Chooser::Controller,
+                up_to: true,
+                constraint: None,
+                selection: crate::types::ability::CardSelectionMode::Chosen,
+            },
+            vec![],
+            ObjectId(100),
+            PlayerId(0),
+        );
+        // First resolution: no players left to prompt afterwards, accumulated=false.
+        state.pending_per_player_zone_choice =
+            Some(crate::types::game_state::PendingPerPlayerZoneChoice {
+                ability: Box::new(ability),
+                remaining_players: vec![],
+                accumulated: false,
+            });
+
+        let mut events = Vec::new();
+        // The first player DECLINES — an empty "up to one" pick.
+        drain_pending_per_player_zone_choice(&mut state, &[], &mut events);
+
+        let bound = state
+            .chain_tracked_set_id
+            .expect("a fresh chain tracked set must be bound after the first decline");
+        assert_ne!(
+            bound, prior,
+            "the first decline must rebind to a FRESH set, not stay on the prior producer's set"
+        );
+        assert_eq!(
+            state.tracked_object_sets.get(&bound),
+            Some(&vec![]),
+            "the fresh per-player set must be empty (everyone declined)"
+        );
+        // The prior producer's set is untouched (still holds its two objects), so
+        // a downstream ChangeZoneAll { TrackedSet } reads the empty fresh set.
+        assert_eq!(
+            state.tracked_object_sets.get(&prior),
+            Some(&vec![ObjectId(7), ObjectId(8)]),
+            "the prior producer's set must be untouched, never inherited by the iteration"
+        );
     }
 }
--- a/crates/engine/src/game/effects/choose_from_zone.rs
+++ b/crates/engine/src/game/effects/choose_from_zone.rs
@@ -437,8 +437,20 @@ fn prompt_next_each_player(
         return Ok(());
     }
 
-    // CR 608.2c: No player had an eligible card — emit the resolution event so the
-    // parked continuation ("put those cards onto the battlefield") still runs.
+    // CR 101.4 + CR 608.2c: No iterated player had an eligible card. When
+    // `accumulated == false` this is the FIRST resolution of the iteration (no
+    // player was ever prompted), so it MUST rebind a FRESH (empty) chain tracked
+    // set before the parked continuation runs — mirroring the first-resolution
+    // rebind in `drain_pending_per_player_zone_choice`. Otherwise an EARLIER
+    // same-chain producer's tracked set (e.g. Breach the Multiverse's preceding
+    // mill) stays bound and a downstream `ChangeZoneAll { TrackedSet }` over-acts
+    // on that stale set instead of this iteration's (empty) picks: "those chosen
+    // cards" must mean exactly the cards chosen by THIS iteration. When
+    // `accumulated == true` an earlier player already rebound a fresh set for this
+    // iteration — leave it bound, do not clobber it.
+    if !accumulated {
+        super::publish_fresh_tracked_set(state, Vec::new());
+    }
     events.push(GameEvent::EffectResolved {
         kind: EffectKind::ChooseFromZone,
         source_id: ability.source_id,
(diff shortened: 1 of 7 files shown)


matthewevans (changes requested):
**[HIGH]** This routes a printed target through the resolution-time choose-from-zone prompt, bypassing target announcement and target legality. Evidence: `crates/engine/src/parser/oracle_effect/imperative.rs:3158-3169` explicitly treats the printed `target` as non-semantic, `crates/engine/src/parser/oracle_effect/imperative.rs:3190-3193` parses and discards it, and `crates/engine/src/parser/oracle_effect/imperative.rs:3246-3255` emits `Effect::ChooseFromZone` rather than a target-bearing effect. At runtime that becomes `WaitingFor::ChooseFromZoneChoice` over `collect_player_zone_cards` candidates (`crates/engine/src/game/effects/choose_from_zone.rs:220-245`), while the real targeting path uses `find_legal_targets` before the spell/ability proceeds (`crates/engine/src/game/casting.rs:8957-9023`) and enforces hexproof/shroud/protection (`crates/engine/src/game/targeting.rs:1545-1642`). Why it matters: Kaya's printed "up to one target creature that player controls" is not a free resolution choice; targets have to be chosen through the targeting machinery, so this can offer creatures that should be illegal to target and does not model the ability's targets correctly. Suggested fix: keep this class unsupported until the engine can represent the per-opponent target slots, or model the choices as real targets selected at activation/cast time and add a production-path regression with hexproof/shroud/protection candidates proving they are excluded.

**[MED]** Empty "up to one" choices do not publish a fresh empty tracked set, so a prior producer in the same resolution chain can remain bound. Evidence: `crates/engine/src/game/effects/choose_from_zone.rs:286-303` only calls `publish_fresh_tracked_set` when `chosen` is non-empty; if every opponent declines, `accumulated` remains false and the chain's tracked-set binding is not cleared before the downstream `ChangeZoneAll { TrackedSet(0) }` from `crates/engine/src/parser/oracle_effect/imperative.rs:3257-3264`. Why it matters: this primitive is documented as making "those" mean exactly the chosen permanents, but the zero-choice path can leave the chain dependent on an earlier tracked-set producer. Suggested fix: on the first per-player choice resolution, publish/rebind a fresh tracked set even when the selected slice is empty, and add a regression where all choices are declined after an earlier same-chain tracked-set producer.

lgray:

Thanks for the review — both findings addressed. New head: `83da5e130`. Rebased onto current upstream/main (`d87972c5`).

**[HIGH] Printed `target` routed through resolution-time choose — fixed via honest Unimplemented (path a).**
Root cause: `parse_controlled_battlefield_body` swallowed and discarded the printed `target ` token (`opt(tag("target "))`), so "exile up to one **target** creature that player controls" lowered to `Effect::ChooseFromZone`, which resolves via `collect_player_zone_cards`/`ChooseFromZoneChoice` and never invokes the targeting machinery (`find_legal_targets` / `can_target`). That can offer creatures that are illegal to target (hexproof/shroud/protection).

The engine has no machinery for a **variable, per-iterated-player set of announced target slots** — `collect_target_slots` builds a fixed slot list, and `repeat_for` resolves its subject at *resolution* time, not announcement (CR 601.2c). Building that is a multi-system feature (casting + targeting + AI + frontend), out of scope for this PR. So `parse_controlled_battlefield_body` now **rejects a leading `target ` token** with a nom guard (`tag("target ").parse(after_other).is_ok()` → `None`) and the line falls through to `Effect::unimplemented`. Kaya's actual printed −2 uses "target", so its second sentence is now honestly Unimplemented rather than silently wrong.

The **non-target** form ("exile up to one creature that player controls") is a CR 115.10a / 608.2c resolution-time choice — correctly *not* targeting — and stays fully supported (`ChooseFromZone { EachOpponent }` + `ChangeZoneAll { TrackedSet }`).

**[MED] Empty "up to one" choices didn't publish a fresh empty tracked set — fixed.**
`drain_pending_per_player_zone_choice` only rebound the chain tracked set on a non-empty pick, so an all-declined loop left a prior same-chain tracked-set producer bound to the downstream `ChangeZoneAll { TrackedSet(0) }` (over-act). It now publishes a fresh (possibly empty) tracked set on the **first** per-player resolution regardless of emptiness; later non-empty picks extend it.

**Tests (discriminating, revert-probed):**
- `printed_target_form_is_unimplemented` — the real printed −2 (with "target") parses to `Effect::unimplemented`, no `ChooseFromZone { EachOpponent }` reachable. Reverting the guard re-emits the choose chain → fails.
- `non_target_form_still_parses_to_each_opponent_exile_chain` / `non_target_form_exiles_one_creature_per_other_player` / `non_target_form_skips_controller` — prove the non-target class stays supported (real `CastSpell`/`SelectCards` pipeline).
- `per_player_first_decline_rebinds_fresh_empty_tracked_set` — a same-chain prior producer + an all-declined first resolution leaves the prior set un-exiled and the chain bound to a fresh empty set. Restoring the old branch exiles the prior producer's object → fails.

Verification: `cargo fmt --all` clean, `cargo test -p engine` full suite green, `cargo clippy -p engine --all-targets -D warnings` clean, parser combinator gate green.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed head `83da5e130e5e8ccb4fe5c77ab5f0fdbc58d072ca` after the follow-up commit.

Prior HIGH is ADDRESSED: the printed `target` variant now rejects the resolution-time `ChooseFromZone` path and stays honestly unsupported, with parser coverage proving the printed Kaya text reaches `Effect::Unimplemented` instead of `ChooseFromZone { EachOpponent }`.

Prior MED is PARTIAL: the explicit first-decline path is fixed, but the same stale tracked-set bug remains when no iterated player has an eligible candidate and therefore no `ChooseFromZoneChoice` is ever parked.

**[MED]** Rebind the per-player tracked set even when the iteration has no eligible candidates. Evidence: `crates/engine/src/game/effects/choose_from_zone.rs:217-260` skips every player with an empty candidate list and emits `EffectResolved` without publishing a fresh empty tracked set; the new rebind only happens in `drain_pending_per_player_zone_choice` at `crates/engine/src/game/effects/choose_from_zone.rs:301-308`, which is never called on the no-prompt path. The new controller-skip regression at `crates/engine/tests/integration/kaya_spirits_justice_per_opponent_exile.rs:341` covers no eligible opponents, but it does not seed a prior same-chain tracked-set producer, so it misses the stale-binding case. Why it matters: a downstream `ChangeZoneAll { TrackedSet }` can still act on a previous producer's set when every opponent has zero eligible permanents, which is the same CR 608.2c "those chosen permanents" bug as the all-decline case. Suggested fix: when `prompt_next_each_player` exhausts `remaining_players` before any accumulation, publish/rebind a fresh empty tracked set before emitting `EffectResolved`, and add a regression with a prior non-empty chain set plus no eligible iterated players.

No approval/enqueue while this tracked-set no-eligible path remains open; required `feature` label is present.

lgray:

Rebased onto current upstream/main and addressed the new [MED] re-review. New head: `d3113c301`.

**Rebase.** The only rebase conflict was a test-module adjacency in `crates/engine/src/game/effects/choose_from_zone.rs` — upstream added `for_each_*` category tests in the same `#[cfg(test)]` location where this PR added its per-player test. Both are independent complete functions; resolution kept both, and re-applied this PR's non-test `drain_pending_per_player_zone_choice` rebind change on top of the upstream-merged file. The printed-`target` Unimplemented nom guard and the drain first-resolution rebind both survive the rebase. Also re-added the `ZoneOwner::EachOpponent` engine-inventory entry that upstream's inventory regeneration dropped (surgical, no full regen).

**[MED] Rebind the per-player tracked set even when the iteration has no eligible candidates — fixed.**
You correctly identified that the prior fix only covered the parked-prompt path: `drain_pending_per_player_zone_choice`'s rebind is never reached when `prompt_next_each_player` exhausts its loop with zero prompts ever parked (every iterated player has an empty/ineligible zone). On that no-eligible exhaustion branch the function emitted `EffectResolved` without rebinding, so an earlier same-chain producer's tracked set (Breach the Multiverse's preceding mill) stayed bound and a downstream `ChangeZoneAll { TrackedSet }` over-acted on it (CR 608.2c — "those chosen cards" must mean exactly the cards chosen by this iteration).

Fix: on the no-eligible exhaustion path, when `accumulated == false` (the first resolution, no player ever prompted), publish a fresh (possibly empty) tracked set before `EffectResolved`, mirroring the drain-path first-resolution semantics. When `accumulated == true` an earlier player already rebound a fresh set for this iteration — that binding is left intact (the `if !accumulated` guard). This is the loop-exhaustion arm of the shared `ZoneOwner::{EachPlayer, EachOpponent}` iteration primitive, not a Kaya special case. Note: rebinding to a fresh empty set (rather than clearing to `None`) is required because `resolve_tracked_set_sentinel` falls back to `latest_tracked_set_id` when `chain_tracked_set_id` is `None`, which would otherwise still resolve to the stale producer set.

**Test (discriminating, revert-probed):** `per_player_no_eligible_player_rebinds_fresh_empty_tracked_set` seeds a prior non-empty same-chain producer set bound to the chain, then drives a `ChooseFromZone { EachOpponent }` over an empty opponent battlefield through the production `resolve` entry so `prompt_next_each_player` exhausts with zero prompts. Asserts: no prompt parked; `chain_tracked_set_id` rebound to a fresh id distinct from `prior`; the fresh set is empty; the `prior` producer's set is untouched. Removing the new `publish_fresh_tracked_set` call leaves the binding on `prior` and flips the distinct-id and empty-set assertions (verified: revert fails at `assert_ne!(bound, prior)`). It is the no-prompt sibling of the existing `per_player_first_decline_rebinds_fresh_empty_tracked_set` (which covers the parked-prompt drain path). CR annotation uses `CR 101.4 + CR 608.2c` (grep-verified against the comprehensive rules: APNAP order + read-whole-text/follow-instructions); `CR 603.7` (delayed triggered abilities) is deliberately not propagated.

**Verification:** `cargo fmt --all` clean; `cargo test -p engine` full suite green (14992 passed, 0 failed); `cargo test -p engine choose_from_zone` 30 passed (both per-player tests + for-each siblings); Kaya integration 4 passed, Breach the Multiverse 1 passed; `cargo clippy -p engine --all-targets -- -D warnings` clean.

matthewevans (approved):
Current head `d3113c301509faa8bef5a20962ef2eb02dec3c1a` addresses the remaining tracked-set blocker.

The printed-target form remains honestly unsupported, so the PR no longer routes Kaya's printed `target` text through a resolution-time choice. The supported non-target form still lowers to the per-other-player choice/exile chain.

The no-eligible branch now publishes a fresh empty chain tracked set before `EffectResolved` when no iterated player ever prompts and no prior pick accumulated. The new regression seeds a prior non-empty same-chain tracked set, drives the production `resolve` entry through the no-prompt `EachOpponent` exhaustion path, and verifies the chain rebinds to a distinct empty set while the prior set remains untouched.

Repush guard passed on the fork branch, required `feature` label is present, mergeState is `CLEAN`, and GitHub CI is green on this head.

PR #4096, opened 2026-06-21
feat(parser): enters-with-finality-counter rider on graveyard cast-this-way permissions

lgray:

## Summary

Implements the **"the creature cast this way enters with a finality counter on it"** rider for the temporary graveyard cast-permission class. Both target cards previously parsed the rider sentence to `Effect::Unimplemented`; the cast-permission itself (`Effect::CastFromZone`, mode `Cast`, `UntilEndOfTurn`) already worked.

CR 614.1c ("[this permanent] enters with …" is a replacement effect) + CR 122.1 / CR 122.1h (a finality counter creates the "if this permanent would be put into a graveyard from the battlefield, exile it instead" replacement).

## Building block (parameterize, don't proliferate)

- `CastingPermission::ExileWithAltCost` gains a typed `enters_with_counter: Option<CounterType>` field — **not a bool**, so the rider covers *any* counter a cast-this-way creature enters with, not just finality.
- The parser decomposes the rider into the **existing** `Effect::AddPendingETBCounters` via a new nom combinator `try_parse_cast_this_way_enters_with_counter`, composed from `tag`/`alt` + `parse_counter_type_typed`. It handles both anaphoric subjects (`that creature` / `it`), both optional gate prefixes (`if you cast a spell this way,` / `if you do,`), and an optional trailing type-addition clause — no verbatim Oracle-string matching.
- `grant_lingering_permissions` records the rider counter on the granted permission; the cast finalization registers a pending ETB counter on the cast object — mirroring the existing `exile_instead_of_graveyard_on_resolve` rider plumbing. The rider sub-ability is consumed as **permission metadata** (it targets the *future* graveyard cast, not the granting ability's trigger event), so it is suppressed on both the accept and decline chain paths.

No new enum/struct variants were added.

## Per-card ship / defer

| Card | Shipped | Deferred | CR |
|---|---|---|---|
| Osteomancer Adept | enters-with-finality-counter rider | — (the "by foraging" additional cost on the graveyard cast is a separate pre-existing gap, unchanged) | CR 614.1c + CR 122.1h |
| The Tomb of Aclazotz | enters-with-finality-counter rider | "is a Vampire in addition to its other types" type-addition on the cast-this-way creature | CR 614.1c + CR 122.1h (shipped); CR 205.1b (deferred) |

**Defer rationale (Tomb Vampire grant):** the type-addition on a *cast-this-way* creature is a continuous effect that does not compose into the cast-permission rider as cleanly as the counter does. Rather than band-aid it, the finality counter applies and the type grant is left unmodeled; the parser accepts the counter rider with or without the trailing clause.

## add-engine-variant verdict

**Not triggered.** Parameterized an existing variant (`ExileWithAltCost`) with a typed `Option<CounterType>` field and reused the existing `Effect::AddPendingETBCounters` — no sibling variant proliferation.

## Discriminating-test map

| Behavioral claim | Production entry | Test | Revert-failing assertion |
|---|---|---|---|
| Creature cast via the granted permission enters with a finality counter | `apply(CastSpell)` → resolve → `casting_costs::finalize` → `pending_etb_counters` → `zone_pipeline` ETB | `casting::tests::graveyard_cast_this_way_enters_with_finality_counter` | `objects[creature].counters[Generic("finality")] == Some(&1)` |
| The counter comes from the rider, not the graveyard-cast path | same path, no rider | `casting::tests::graveyard_cast_without_rider_has_no_finality_counter` (negative control) | counter is `None` |
| Rider parses for both subjects/gates/trailing-clause | `parse_oracle_text` / `try_parse_cast_this_way_enters_with_counter` | `oracle_effect::tests::cast_this_way_enters_with_finality_counter_rider_parses`, `..._is_counter_generic`, `osteomancer_adept_finality_rider_parses_through_card` | rider effect is `AddPendingETBCounters(finality/+1+1)` not `Unimplemented` |

The runtime test drives the real cast pipeline through `apply()`; it is not a shape-only test.

## Gate results

`cargo fmt` clean · `check-parser-combinators.sh` (with/without `upstream/main`) exit 0 · parser-diff string-dispatch grep clean · `check-engine-authorities.sh upstream/main` exit 0 · `clippy --all-targets -D warnings` clean · `cargo test -p engine` green except the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (passes in isolation `--test-threads=1`) · CR-annotation diff grep: zero UNVERIFIED · `cargo semantic-audit`: neither card produces a finding · both cards confirmed in regenerated `card-data.json` with `AddPendingETBCounters{finality}` and no `Unimplemented` rider.

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -7687,6 +7687,7 @@ pub(super) fn initiate_cast_during_resolution(
                 resolution_cleanup: Some(cleanup),
                 duration: None,
                 exile_instead_of_graveyard_on_resolve: false,
+                enters_with_counter: None,
             });
     }
     let mut prepared = prepare_spell_cast_with_variant_override(state, player, hit_card, None)?;
@@ -14243,6 +14244,7 @@ mod tests {
                 duration: None,
 
                 exile_instead_of_graveyard_on_resolve: false,
+                enters_with_counter: None,
             });
         let prepared = prepare_spell_cast(&state, PlayerId(0), exiled).unwrap();
         assert!(matches!(prepared.mana_cost, ManaCost::NoCost));
@@ -26311,6 +26313,7 @@ mod tests {
                     resolution_cleanup: None,
                     duration: None,
                     exile_instead_of_graveyard_on_resolve: false,
+                    enters_with_counter: None,
                 });
         }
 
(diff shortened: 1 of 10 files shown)


matthewevans (changes requested):
**[HIGH]** The Tomb trailing type-change rider is silently swallowed while only the finality counter is modeled. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:1757-1771` treats `" and is a "` / `" and it's a "` as an acceptable tail and still returns only `Effect::AddPendingETBCounters`; `parse_effect_clause_inner` then returns that parsed clause at `crates/engine/src/parser/oracle_effect/mod.rs:5055-5062`, so the rest of the sentence is not left as `Unimplemented`. The new test at `crates/engine/src/parser/oracle_effect/mod.rs:52584-52598` locks in that behavior for `"it enters with a finality counter on it and is a Vampire in addition to its other types"`. Why it matters: The Tomb of Aclazotz is not just a finality-counter grant; it also makes the creature a Vampire, and accepting the whole sentence while dropping that continuous type effect makes coverage/support dishonest. Suggested fix: either parse and implement the cast-this-way subtype rider as permission metadata too, or fail/leave the whole combined rider unsupported when a type-addition tail is present so the counter-only subset does not mark Tomb supported.

**[MED]** The ETB counter is applied by scanning for any permission on the object, not the permission actually used for this cast. Evidence: `crates/engine/src/game/effects/cast_from_zone.rs:567-568` stamps `enters_with_counter` onto the granted `CastingPermission`, but `crates/engine/src/game/casting_costs.rs:5547-5559` later looks for any `ExileWithAltCost { enters_with_counter: Some(..) }` on the object and registers the pending counter without proving that this cast consumed that specific permission. Why it matters: if a creature has multiple graveyard/exile cast permissions, a rider from one permission can leak onto a cast made through another permission, violating the "cast this way" condition. Suggested fix: carry the consumed permission's rider through the prepared/finalized cast path, or otherwise bind the rider to the selected cast permission instead of scanning the object's entire permission list at finalization.

lgray:

Thanks for the review @matthewevans — both findings addressed in `aab085737` (rebased onto `upstream/main` `d87972c5`).

**[HIGH] Tomb type-grant rider no longer swallowed.** I took your second suggested path (honest deferral). `try_parse_cast_this_way_enters_with_counter` now accepts the counter rider **only** when the remainder is empty after `" counter on it"`; when a `"and is a [subtype] in addition to its other types"` tail (CR 205.1b) is present it returns `None`, so the whole combined sentence falls through to `Effect::Unimplemented` and The Tomb of Aclazotz is left honestly unsupported rather than partially accepted as a bare counter. Osteomancer Adept (no tail) is unchanged.

Path A (modeling the subtype rider as permission metadata) was rejected after tracing: it would require a new `Vec<ContinuousModification>` field on `ExileWithAltCost` **and** a net-new "pending ETB continuous-effect" application channel — none exists today (`pending_etb_counters` is counter-only), and the sole single-object additive-type precedent (`GenericEffect` bound to `ParentTarget`/`TrackedSet`) requires a same-ability resolving `ChangeZone`, which a future-cast permission rider does not have. That is new ETB-layer infrastructure for a one-card tail, so deferral is the honest, in-scope choice. CR 205.1b (type grant) and CR 614.1c (enters-with replacement) are kept as distinct rule sections.

**[MED] enters-with-counter now binds to the consumed permission.** Added `selected_exile_alt_cost_permission_enters_with_counter` (mirroring the existing `selected_exile_alt_cost_permission_*` family), and `finalize_cast_with_phyrexian_choices` now reads the rider from the permission that actually supports **this** cast (`exile_alt_cost_permission_supports_cast`) instead of scanning the whole permission list (CR 608.2c). A non-consumed sibling permission’s rider can no longer leak.

**Tests (all discriminating — verified to fail on reverted code):**
- `cast_this_way_enters_with_finality_counter_rider_parses` — combined sentence now asserts `None`; four no-tail cases still parse.
- `tomb_aclazotz_counter_plus_type_tail_is_unimplemented` — combined Tomb residual surfaces as `Effect::Unimplemented` through the production parser; negative twin keeps the bare-counter rider parsing.
- `enters_with_counter_does_not_leak_from_non_consumed_permission` — two-permission hostile fixture (rider on a non-consumed, foreign-`granted_to` permission) enters with no finality counter; positive twin (rider on the consumed permission) enters with exactly one.

Gate: `cargo fmt` clean, `cargo clippy -p engine --all-targets -D warnings` clean, all targeted engine tests green, parser-combinator gate green. No new enum variant (no `engine-inventory.json` change). card-data regen deferred — The Tomb of Aclazotz is only in the dormant corpus, not the active `AtomicCards` set, and `card-data.json` is gitignored; the honest `Unimplemented` shape is proven by the production-path test above.

matthewevans (approved):
Re-reviewed head aab085737b83b78a6e8b26c9eea853387780b0fc after the follow-up push. The previous Tomb tail and consumed-permission blockers are addressed: the parser now rejects the combined counter+type-grant sentence as unsupported instead of silently dropping the Vampire rider, and the runtime reads the enters-with-counter rider from the permission that actually supports the cast. CI is green and the required feature label is present.

PR #4017, opened 2026-06-21
feat(engine): look-at-face-down + can't-be-turned-face-up statics + granted turn-face-up (Found Footage, Karlov Watchdog, Etrata)

lgray:
:robot: _AI text below_ :robot:

# feat(parser+engine): face-down look-at permission + turn-face-up prohibition/effect (Found Footage, Karlov Watchdog, Etrata)

## Summary

Singleton BATCH 6 of the std-188 effort — the face-up/down subsystem (look-at permission, turn-face-up effect, turn-face-up prohibition). Reuses the merged morph/disguise/cloak/manifest infrastructure and adds two parameterized static variants for the look-at and prohibition classes.

**Shipped (0-Unimplemented + discriminating runtime test):**
- **Found Footage** — "You may look at face-down creatures your opponents control any time." → new `StaticMode::MayLookAtFaceDown` static; runtime enforced in `visibility.rs`.
- **Karlov Watchdog** — "Permanents your opponents control can't be turned face up during your turn." → new `StaticMode::CantBeTurnedFaceUp` static; runtime enforced in `morph::turn_face_up` and `effects::turn_face_up`.
- **Etrata, Deadly Fugitive** — granted "{2}{U}{B}: Turn this creature face up. …" → the granted activated-ability head clause now lowers to the existing `Effect::TurnFaceUp { SelfRef }` (no new variant).

**Honest-deferred (Effect::unimplemented, coverage stays red):**
- **Lumbering Laundry** — "{2}: Until end of turn, you may look at face-down creatures you don't control any time." Needs a *duration-limited static-mode grant*: the transient-continuous-effect system (`TransientContinuousEffect`) carries only `Vec<ContinuousModification>`, not a `StaticMode`. No existing surface; deferring rather than shipping a silent runtime no-op.
- **Vannifar, Evolved Enigma** — "Cloak a card from your hand." The `Effect::Cloak` resolver only cloaks top-of-library; cloak-from-hand needs interactive hand-card selection (the merged face-down cluster already deferred non-library cloak).
- **Expose the Culprit** — "…shuffle that pile, then cloak them." Cloak-a-set after exile+shuffle with "any number" interactive selection; same non-library-cloak defer.

## add-engine-variant gate

Two new variants, both run through the three-stage gate.

### `StaticMode::MayLookAtFaceDown` — APPROVED (CR 708.5)
[...]

PR #4098, opened 2026-06-21
feat(parser+engine): duration-bound face-down look permission (Lumbering Laundry)

lgray:

## Summary

Implements the **duration-bound "you may look at face-down permanents you don't control any time"** permission for **Lumbering Laundry** — `{2}: Until end of turn, you may look at face-down creatures you don't control any time.`

This is the time-limited sibling of **Found Footage**'s continuous look permission shipped in #4017. It reuses the existing `StaticMode::MayLookAtFaceDown` outright — the controller scope and the duration are the only two parameter axes — so **no new engine variant is added**.

## Building block (reused vs extended)

**Reused** #4017's `StaticMode::MayLookAtFaceDown` permission end-to-end. The only new infrastructure is wiring the duration-bound surface form into the same permission:

- **Parser grammar (single authority):** extracted `parse_may_look_at_face_down_filter` in `oracle_static/dispatch.rs`. It accepts an *optional* `"you may "` prefix (the static line keeps it; the activated-ability clause shell already peels it) and the required `"look at "` verb, returning the affected face-down/controller filter. Both the static-line builder (Found Footage) and the new activated-ability effect intercept call it — the grammar is never re-encoded.
- **Effect intercept:** `oracle_effect/imperative.rs` intercepts the whole `[Until end of turn,] you may look at face-down … any time` phrase before the `you`/`may` first-word arms (which would mark it a spurious optional one-shot and fail to a `look` parse gap), lowering it to `Effect::GenericEffect` carrying the shared `MayLookAtFaceDown` static-mode. The duration rides on the resolving ability via the stripped `Until end of turn,` prefix.
- **Resolution (CR 611.2c):** a look permission is a *rules-modifying* continuous effect, so its affected face-down filter rides on the transient continuous effect **intact** (re-evaluated continuously at look time) rather than being expanded to a fixed `SpecificObject` set — a permanent turned face down after this resolves but before end of turn is still visible. `register_transient_effect` registers one TCE bound to the looker via `controller`.
- **Visibility:** `viewer_may_look_at_face_down` now also scans transient continuous effects carrying the permission, reading them identically to the printed static (honoring `ForAsLongAs`/explicit-condition gates). The permission is never stamped onto objects through the layer system (its filter matches opponents' permanents), so the layer gather skips it — mirroring the printed static's empty-modifications, mode-read design.
- **Duration (CR 514.2):** the `UntilEndOfTurn` TCE expires at cleanup via the existing prune.

## Per-card verdict

| Card | Verdict | Notes |
|---|---|---|
| **Lumbering Laundry** | **Shipped** | Activated look line previously a `look` parse gap; now fully supported. Disguise line already worked. `supported: true, gap_count: 0`. |
| **Found Footage** | Already shipped by #4017 | Verified still fully supported (no regression). `supported: true, gap_count: 0`. |

## CR citations (grep-verified against `docs/MagicCompRules.txt`)

- **CR 708.5** — a player may look only at their own face-down permanents; effects override this to grant looking at others'.
- **CR 611.2c** — a continuous effect from a resolving ability that doesn't modify characteristics/controller "modifies the rules of the game, so it can affect objects that weren't affected when that continuous effect began" — the look permission keeps its filter live rather than fixing an object set.
- **CR 514.2** — "until end of turn" effects end at cleanup.
- **CR 708.2 / 708.4** — face-down 2/2 characteristics; default own-only look rule.

## add-engine-variant verdict

**No new variant.** Reused `StaticMode::MayLookAtFaceDown` (parameterized via the affected filter + duration), `ContinuousModification::AddStaticMode`, `Effect::GenericEffect`, and `TransientContinuousEffect`. Parameterize-don't-proliferate satisfied.

## Discriminating-test map (full production path)

| Behavioral claim | Changed seam | Production entry | Test | Revert-failing assertion |
|---|---|---|---|---|
| Activated line parses to the look permission | `parse_imperative_family_ast` intercept | `parse_oracle_text` | `parser::oracle::tests::lumbering_laundry_look_at_face_down_until_eot_parses` | `GenericEffect` with `MayLookAtFaceDown` over Opponent-scoped face-down filter (no Unimplemented) |
| Resolution keeps the filter live, binds the looker, no object pollution | `register_transient_effect` + layer gather skip | `effects::effect::resolve` | `game::effects::effect::tests::generic_effect_may_look_at_face_down_keeps_filter_and_binds_looker` | `tce.affected == Typed(..)` (not `SpecificObject`); opponent creature not stamped |
| Looker sees opponent's identity only while active, reverts at EOT | `viewer_may_look_at_face_down` TCE scan + `prune_end_of_turn_effects` | `filter_state_for_viewer` | `game::visibility::tests::lumbering_laundry_reveals_opponent_face_down_until_end_of_turn` | `granted_back` expect (flips to None on revert); post-prune `back_face.is_none()` (flips if modeled permanent) |

Negative/sibling coverage: Found Footage's static-path tests (parser + runtime) confirm no regression.

## Gate results

`cargo fmt` clean · `cargo clippy --workspace` clean · `cargo test -p engine` green (only the two known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` / `target_selection_legal_actions_do_not_fall_back_to_stale_slot_targets` fail under contention, pass in isolation single-threaded) · `check-parser-combinators.sh` (with and without `upstream/main`) exit 0 · parser string-dispatch diff grep clean · `check-engine-authorities.sh upstream/main` exit 0 · `cargo coverage` / `cargo semantic-audit` clean (no findings for either card) · CR-annotation diff grep clean.

### DIFF
--- a/crates/engine/src/game/effects/effect.rs
+++ b/crates/engine/src/game/effects/effect.rs
@@ -135,6 +135,39 @@ fn register_transient_effect(
 ) {
     let modifications = snapshot_transient_modifications(state, ability, &static_def.modifications);
 
+    // CR 708.5: A duration-bound "you may look at face-down [permanents] you don't
+    // control any time" permission (Lumbering Laundry) is a *player-scoped* look
+    // permission, not an object grant. Unlike a pump or keyword grant, the set of
+    // face-down permanents the looker may see is re-evaluated continuously at look
+    // time (a permanent turned face down after this resolves but before end of
+    // turn is still visible to the looker), so the affected face-down filter must
+    // ride on the TCE intact rather than be expanded to a fixed `SpecificObject`
+    // set by the broadcast branch below. Register ONE TCE whose `controller` is
+    // the looker and whose `affected` keeps the face-down/controller filter; the
+    // visibility check (`viewer_may_look_at_face_down`) reads it exactly like the
+    // printed `MayLookAtFaceDown` static, evaluating the filter against each
+    // face-down permanent from the looker's perspective.
+    if modifications.iter().any(|m| {
+        matches!(
+            m,
+            ContinuousModification::AddStaticMode {
+                mode: crate::types::statics::StaticMode::MayLookAtFaceDown,
+            }
+        )
+    }) {
+        if let Some(affected) = static_def.affected.clone() {
+            state.add_transient_continuous_effect(
+                ability.source_id,
+                ability.controller,
+                duration.clone(),
+                affected,
+                modifications,
+                static_def.condition.clone(),
+            );
+            return;
+        }
+    }
+
     // CR 608.2c (issue #323 class): SelfRef is the printed-name anaphor and
     // always refers to the source object regardless of `ability.targets`.
     // Short-circuit BEFORE the chosen-targets branch so chained Effect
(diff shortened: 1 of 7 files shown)


matthewevans (changes requested):
Blocking this on one controller-context bug in the transient permission path.

HIGH: the duration-bound look permission matches "you don't control" from the source's current controller, not the looker

The PR correctly stores the resolved look permission as a TCE controlled by the activating player: `register_transient_effect` passes `ability.controller` into `add_transient_continuous_effect` for `MayLookAtFaceDown` (`crates/engine/src/game/effects/effect.rs:158-166`). The visibility scan also gates private visibility on `tce.controller` (`crates/engine/src/game/visibility.rs:741-744`). But when it evaluates the TCE's affected filter it builds `FilterContext::from_source(state, tce.source_id)` (`visibility.rs:770-771`). That helper derives `source_controller` from the current battlefield object (`crates/engine/src/game/filter.rs:534-545`), and `ControllerRef::Opponent` compares against that `source_controller` (`filter.rs:1386-1398`).

For Lumbering Laundry, the text is "you may look at face-down creatures you don't control". Once the activated ability resolves, "you" is the player who controlled the ability at resolution for the duration-bound continuous effect. If the source changes controller later in the turn, the permission should still let the original looker see face-down creatures they don't control. Current code instead reinterprets `Opponent` relative to the source's new controller, so the original looker can lose access to the opponent's face-down creature and potentially gain access to the wrong side.

Please evaluate the TCE affected filter with the stored TCE controller, e.g. the existing explicit-controller context (`FilterContext::from_source_with_controller(tce.source_id, tce.controller)`), and add a regression that resolves the permission, changes Lumbering Laundry's controller, then verifies the original looker still sees only the face-down creatures they don't control.

lgray:

Thanks @matthewevans — addressed the controller-context bug.

**Root cause.** The duration-bound `MayLookAtFaceDown` transient continuous effect evaluated its affected filter with `FilterContext::from_source(state, tce.source_id)`, which derives `source_controller` from the source's *current* battlefield object. After the source changed controller, the `ControllerRef::Opponent` filter ("you don't control") was reinterpreted relative to the source's new controller instead of the original looker — so the looker could lose access to an opponent's face-down creature (or gain the wrong side).

**Fix.** The look-permission filter is now evaluated against the stored `tce.controller` via `FilterContext::from_source_with_controller(tce.source_id, tce.controller)`, exactly as you suggested. The looker is latched to the ability's controller at resolution.

**Rules framing (corrected).** While in here I corrected the CR annotations: the *looker*-binding ("who 'you' is") is **CR 608.2c** (the controller of a resolved ability follows its instructions), not CR 611.2c. **CR 611.2c**'s "set won't change" clause applies only to characteristic/control-modifying effects; a look permission is rules-modifying, so its affected *set* stays **dynamic** (re-evaluated each query against the latched looker). The annotations now split these two correctly. Also fixed a pre-existing CR 708.4 -> CR 708.5 miscitation on the looker's own-permanent right.

**Regression.** Added a 3-player test (`lumbering_laundry_look_permission_survives_source_control_change`) that resolves the permission, changes Lumbering Laundry's controller to an opponent, and asserts the original looker still sees the *new controller's* face-down creature — which redacts under the buggy `from_source` path (verified: reverting the seam makes the test panic at that exact assertion). A second opponent's creature is a positive control proving the permission stays active rather than merely disabled.

**Rebased** onto current `main`. Scope is engine-only (`crates/engine/src/game/visibility.rs`); `cargo fmt`/`clippy -D warnings` clean and the engine suite is green.

New head: `6820a8701`.

matthewevans:
Re-reviewed current head `6820a8701838430f46532a89e45206dbfe022372` via the GitHub API diff and isolated worktree. The prior controller-context blocker is addressed: the TCE path now evaluates the affected filter with `FilterContext::from_source_with_controller(tce.source_id, tce.controller)`, and the new 3-player regression covers source control changing after resolution.

I did not find a new semantic blocker in the current diff. Holding approval/enqueue until the in-progress normal CI jobs complete green and the head remains unchanged. Required category label `feature` is present.

matthewevans (approved):
Approved current head 6820a8701838430f46532a89e45206dbfe022372 after the prior controller-context blocker was addressed and normal CI is green. Required category label `feature` is present.

PR #4099, opened 2026-06-21
feat(parser): cast-from-zone alternative/additional non-mana cost rider (Valgavoth, Terror Eater; Festival of Embers)

lgray:

### Building block

Adds ONE general building block for the "cast a spell from a zone-permission static for a non-mana cost" class, mirroring the existing `TopOfLibraryCastPermission.alt_cost` precedent (Bolas's Citadel) rather than per-card hacks:

- A new struct `CastExtraCost { cost: AbilityCost, mode: CastCostMode }` and a 2-variant enum `CastCostMode { Alternative, Additional }` in `types/statics.rs`.
- A new `#[serde(default)] extra_cost: Option<CastExtraCost>` field on **both** `StaticMode::ExileCastPermission` and `StaticMode::GraveyardCastPermission`.

The `mode` axis is a typed enum (never a bool), distinguishing the two CR-distinct shapes:

| mode | CR | Effect | Card |
|---|---|---|---|
| `Alternative` | 118.9 | Zeroes the spell's mana cost; the `AbilityCost` is paid in lieu (routed through the single `pay_additional_cost` authority). | Valgavoth, Terror Eater |
| `Additional` | 601.2f | The mana cost is still paid; the `AbilityCost` is folded into `combined_imposed_additional_cast_cost`. | Festival of Embers |

Runtime: `exile_static_permission_extra_cost` / `graveyard_static_permission_extra_cost` (casting.rs) feed mana-cost zeroing in `cast_spell`, alt-cost payment in `casting_costs.rs`, the additional-cost merge, and a pay-life affordability gate in `prepare_spell_cast`. The `AbilityCost` payload rides on serde; `Display`/`FromStr` emit only a diagnostic `extra_cost=<mode>` marker (same lossy-by-design treatment as `TopOfLibraryCastPermission.alt_cost`), so the historical 2-/3-segment forms round-trip unchanged.

Parser (nom combinators, no string dispatch):
- The persistent exile-play parser recognizes the trailing "If you cast a spell this way, pay life equal to its mana value rather than pay its mana cost." sentence, delegating to the shared `oracle_effect::try_parse_alt_cost_rider` authority (kept in lockstep with the Bolas's Citadel class).
- The graveyard-cast parser strips an optional "During your turn, " prefix (gated to the controller's turn via `StaticCondition::DuringYourTurn` — CR 102.1, rules-correct, not silently dropped) and parses a new combinator `parse_cast_permission_additional_cost_rider` ("by paying N life in addition to their/its other costs").

### Per-card ship/defer verdict

| Card | Verdict | CR | Reason |
|---|---|---|---|
| **Valgavoth, Terror Eater** | **SHIP** | 118.9 | exile permission, ALTERNATIVE `PayLife(SelfManaValue)`. Verified supported against generated card-data (was the `static_structure` gap). |
| **Festival of Embers** | **SHIP** | 601.2f / 119.4 | graveyard permission, ADDITIONAL `PayLife(1)`, your-turn-gated. Verified `supported=true, gap_count=0` against generated card-data. |
| Dawnhand Dissident | DEFER | 601.2f | needs filtered spell type ("creature spells") + "cards you own" provenance + "from among" persistent-parser shape AND an interactive remove-counters-from-creatures-you-control cost. The building block (`Additional` + `RemoveCounter`) exists and round-trips; only the parser shape is deferred. Honest gap. |
| Warped Space | DEFER | — | "Once each turn, you may pay {0} rather than pay the mana cost for a spell you cast from exile" — global once-per-turn cost replacement not scoped to a source's exile pool; different scope axis. Honest gap. |
| Weftwalking | DEFER | — | "The first spell each player casts during each of their turns may be cast without paying its mana cost" — first-spell-per-turn-free, all players; separate scope subsystem. Honest gap. |
| Tinybones / Memory Vessel | out of scope | — | stash-counter pool / all-player can't-play-from-hand (pre-existing deferrals, unchanged). |

All deferred cards confirmed to remain honest `Effect::Unimplemented`/unknown gaps against generated card-data.

### add-engine-variant verdict

**No new sibling enum variant.** This is parameterization: a new `Option<CastExtraCost>` field on two existing variants (the inventory already lists both `ExileCastPermission` and `GraveyardCastPermission`), plus one small new struct + a 2-variant `CastCostMode` whose `Alternative`/`Additional` distinction lies on a single categorical axis (CR 118.9 vs CR 601.2f cost classification). No X / OpponentX / TargetX sibling cluster; the typed enum replaces what would otherwise be a bool. The `ExileCastCost` doc comment already anticipated "an alternative life cost analogous to Bolas's Citadel" on exactly this axis.

### Discriminating-test map

| Behavioral claim | Changed seam | Production entry point | Test | Revert-failing assertion | Sibling/negative |
|---|---|---|---|---|---|
| Valgavoth's alternative cost zeroes the spell's mana cost | `exile_static_permission_extra_cost` → `cast_spell` zeroing | `effective_spell_cost` + `can_pay_cost_after_auto_tap` (real auto-tap) | `valgavoth_exile_static_alternative_cost_zeroes_mana_cost` | with empty pool, `effective` is without-paying-mana only with `extra_cost=Some(Alternative)`; reverting to `None` leaves `{2}` → unpayable | spell confirmed in pool first |
| Valgavoth's alt cost gates legality on life | affordability gate in `prepare_spell_cast` | `can_cast_object_now` | `valgavoth_exile_static_alternative_cost_gated_on_life` | at 0 life **with {2} in pool**, cast is illegal only because the alt cost requires 2 life; reverting to `None` makes the {2} pool pay the printed cost → castable | pool funded so mana isn't the blocker |
| Festival's additional cost pays 1 life on top of mana | `cast_permission_additional_extra_cost` → `combined_imposed_additional_cast_cost` | full `SpellCast` driver via `GameRunner` (real cast→resolve) | `festival_cast_pays_one_life_in_addition` | `life_delta(P0) == -1` only with `extra_cost=Some(Additional)`; reverting → delta 0 | mana funded so the only life loss is the additional cost |
| Festival keeps the printed mana cost (additional, not alternative) | additional-cost path | `effective_spell_cost` | `festival_keeps_printed_mana_cost` | mana cost is NOT zeroed | distinguishes Additional from Alternative |
| Festival is gated to the controller's turn | `StaticCondition::DuringYourTurn` | `can_cast_object_now` | `festival_only_functions_on_controllers_turn` | uncastable on opponent's turn | |
| Cast-from-exile line lowers to the alt-cost static | persistent parser rider | `parse_static_line` / `parse_oracle_text` | `persistent_exile_play_permission_valgavoth_alternative_pay_life` | mode + extra_cost shape; full `~`-dispatch routes to same static | shape test paired with runtime tests above |
| Graveyard line lowers to the additional-cost static | graveyard parser rider | `parse_static_line` | `graveyard_cast_permission_festival_additional_pay_life` | `extra_cost=Some(Additional, PayLife(1))` + `DuringYourTurn` condition | shape test paired with the integration test above |
| Full Valgavoth card supports the cast-from-exile line | coverage seam | `build_parse_details_for_face` | `valgavoth_terror_eater_cast_from_exile_alt_cost_supported` | the `ExileCastPermission` detail is supported and carries an `Alternative` PayLife extra-cost | |
| All three `CastExtraCost` shapes serde round-trip | serde | `serde_roundtrip` | building-block coverage incl. Dawnhand's `Additional`+`RemoveCounter` | payload preserved | |

The Valgavoth and Festival revert-flip behaviors were confirmed concretely (temporarily reverting `extra_cost` to `None` fails each named assertion).

### Gate results

`cargo fmt` clean · `check-parser-combinators.sh` (with + without `upstream/main`) exit 0 · parser diff string-dispatch grep clean (only benign `Iterator::find` / annotated label `starts_with` / `matches!` in tests) · `check-engine-authorities.sh upstream/main` exit 0 · `clippy --workspace --all-targets --features engine/proptest -D warnings` clean · `cargo test -p engine` all green except the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` (passes `--test-threads=1`) · `cargo semantic-audit` (run against freshly generated card-data) — 0 new findings on Valgavoth/Festival/Dawnhand · CR-annotation diff gate — 0 UNVERIFIED.


matthewevans (changes requested):
Blocking this head.

HIGH: the extra-cost lookup is not bound to the selected ExilePermission source

The cast variant already records which static permission authorized the cast: `CastingVariant::ExilePermission { source, frequency }` in `crates/engine/src/types/game_state.rs:4868-4873`, and the candidate creation stamps that source from `exile_cast_permission_source` at `crates/engine/src/game/casting.rs:3277-3280`. But the new cost hooks ignore that selected source and re-scan for the first matching permission on the object. `exile_static_permission_extra_cost` just calls `exile_cast_permission_source_full(state, player, exiled_id)` and returns that source's `extra_cost` (`crates/engine/src/game/casting.rs:2283-2289`). That is then used to zero the mana cost in `prepare_spell_cast_with_variant_override_inner` (`crates/engine/src/game/casting.rs:3629-3637`), gate life affordability (`crates/engine/src/game/casting.rs:10031-10048`), fold additional costs (`crates/engine/src/game/casting_costs.rs:2820-2835`), and pay Valgavoth's alternative cost (`crates/engine/src/game/casting_costs.rs:3305-3314`).

That means a card in exile that is castable through two active `ExileCastPermission` sources can be charged/zeroed according to whichever source `exile_permission_sources` happens to find first, not the source carried by the selected `CastingVariant::ExilePermission`. For example, if one source grants a normal-cost permission and Valgavoth also grants a pay-life alternative permission for the same exiled spell, choosing the normal-cost source can still get Valgavoth's cost treatment if Valgavoth is the first matching source in the battlefield scan. The fix should carry or filter by the selected source/frequency when retrieving the extra cost, and add a regression with two permissions on the same exiled spell proving the elected source determines whether the extra cost applies.

HIGH: current head fails required Rust fmt

GitHub CI run 27919641544 fails in `cargo fmt --all -- --check`. The diff is in the PR-added Valgavoth parser test at `crates/engine/src/parser/oracle.rs:6580`; rustfmt wants the long `.expect("Valgavoth's cast-from-exile line must appear as an ExileCastPermission static")` split across lines. The rollup `Rust (fmt, clippy, test, coverage-gate)` is red because `RUST_LINT_RESULT=failure`.

The required `feature` label is present, but this head is not approvable/enqueueable until both the semantic source-binding issue and the red fmt check are fixed.

matthewevans (changes requested):
**[MED]** Bind the free-cast cost check to the elected exile permission source. Evidence: `crates/engine/src/game/casting.rs:4211` re-enters `exile_cast_permission_source(state, player, object_id)` when `casting_variant` is `ExilePermission`, but that helper returns the first authorizing source rather than the `source` already carried by `CastingVariant::ExilePermission`; the newly-bound path at `crates/engine/src/game/casting.rs:3723` shows the intended selected-source authority. Why it matters: if two exile statics authorize the same card and the first source is `WithoutPayingManaCost` while the player elected a different pay-normal/additional-cost permission, the spell is treated as `ManaCost::NoCost` and can skip the mana payment for the wrong source. Suggested fix: in the `is_exile_permission_free_cast` branch, destructure `CastingVariant::ExilePermission { source, .. }` and validate/read that exact source via the full source lookup instead of calling the first-match helper; add the same two-source regression with a free source plus a pay-normal source.

lgray:

Re-ran the full implementation pipeline (plan → plan-review → executor → hardened gates → impl-review) against the two review rounds. **The prior quick fix was incomplete** — it bound the extra-cost / zeroing / affordability / additional-cost paths to the elected source (the HIGH review), but the **MED review's free-cast seam was still mis-bound**, confirming the over-permit you flagged.

**Root cause (the remaining gap):** `is_exile_permission_free_cast` (`crates/engine/src/game/casting.rs`) destructured `CastingVariant::ExilePermission { .. }` — dropping the elected `source` — and re-entered the first-match `exile_cast_permission_source(state, player, object_id)` scan. With two functioning `ExileCastPermission` statics for the same exiled spell (one `WithoutPayingManaCost`, one pay-normal), the first-scanned source decided the free-cast regardless of which permission the player elected.

**Fix:** bind the decision to the elected source via `exile_cast_permission_source_full(state, player, object_id, Some(*source))`, reading that exact source's `ExileCastCost` and failing closed (not-free) when it no longer functions — mirroring the prior extra-cost binding. No fallback first-match scan is needed here: the `casting_variant` default chain has no `ExilePermission` arm, so the block fires only on explicit election where `source` is always present.

**First-match audit:** the three remaining first-match call sites are correctly *not* elected-source consumers — `has_exile_cast_permission` (existence/legality), `elected_exile_permission_source` (`.or_else` fallback after the variant's source, pre-election only), and the candidate-enumeration stamp. The free-cast seam was the sole gap.

**Discriminating regression** (`exile_permission_free_cast_binds_to_elected_source_not_first_match`): two competing statics, the free source created first (lowest `ObjectId`, so the pre-fix scan deterministically reaches it); both `Unlimited`/`AnyTime` so frequency/timing aren't confounds. Electing the pay-normal source must keep the mana cost (revert-failing); electing the free source zeroes it (positive control). **Revert-proof:** reverting only the seam makes the test FAIL at the pay-normal assertion; restoring it passes — reproduced independently in review.

**CR:** annotated `CR 113.6d` + `CR 118.9a` + `CR 601.2b` (cost-modifying ability functions on the stack; only one alternative cost applies; the previously-made choice of which permission to cast through restricts the cost). `601.2a` was deliberately *not* used — it only moves the card to the stack, not cost determination.

**Other addressed points:** fmt is clean on the rebased head (the HIGH review's `oracle.rs:6580` fmt failure was against the pre-rebase commit and is already resolved). Regenerated `data/engine-inventory.json` to catalog the new `CastCostMode` enum (it was previously absent).

**Gates (foreground, Tilt down):** `cargo fmt --all -- --check` exit 0 · `cargo clippy -p engine --all-targets` clean · new test passes + revert-proven discriminating · regressions green (`valgavoth_exile_static_*` ×4, `exile_extra_cost_binds_to_elected_source_not_first_match`, Festival of Embers integration ×3) · CR-annotation diff gate 0 UNVERIFIED · `cargo engine-inventory` deterministic (two identical runs).

Rebased onto current `upstream/main`; head is now `7032e8390`.

matthewevans:
VERDICT: pending-CI

I re-reviewed the latest elected-source follow-up for the remaining free-cast blocker.

The prior issue appears addressed:
- `is_exile_permission_free_cast` now destructures `CastingVariant::ExilePermission { source, .. }` and reads that exact source with `exile_cast_permission_source_full(..., Some(source))`, failing closed if the elected source no longer functions.
- The same elected-source authority is used for mana-cost zeroing, additional extra-cost folding, alternative-cost payment, and life-affordability gating.
- The new regression creates the free permission first so it is the first scanned source, then proves electing the pay-normal permission does not zero the mana cost while electing the free permission does.
- I verified the cited CR references locally: CR 113.6d, CR 118.9a, and CR 601.2b match the cost-modification/alternative-cost/elected-choice behavior being annotated.

I ran a guarded update-branch with expected head `7032e839097401ea519fcd98c8ec0fe5d62c112c`; GitHub advanced the PR to `6fd5fde9fa29242c6b43d49e9ac70e9117414402`. Full CI is running on that current head, so I am leaving formal approval/enqueue pending green checks and the repush guard.

matthewevans (approved):
Approved after re-review of the elected-source exile permission cost-rider follow-up, green current-head CI, CLEAN merge state, and repush guard on 6fd5fde9fa29242c6b43d49e9ac70e9117414402.

PR #4102, opened 2026-06-21
feat(parser): token-count ×N replacement multiplier (Ojer Taq, Deepest Foundation)

lgray:

## Building block: ×2 `Double` → ×N `Times { factor }`

The token / counter / life-gain replacement multiplier was hard-wired to ×2
(`QuantityModification::Double`). This PR **parameterizes** it into a single
multiplicative variant `Times { factor: u32 }`, mirroring the existing
`ManaModification::Multiply { factor }`. `Double` becomes the named constant
`QuantityModification::DOUBLE` (`Times { factor: 2 }`), and all 43 doubling call
sites migrate mechanically.

This is the right parameterization, not a sibling addition: adding a `Triple`
beside `Double` would seed exactly the `Double`/`Triple` multiplicative
sibling-cluster smell the engine guidelines warn against (cf.
`DamageModification::Double | Triple`). One `Times { factor }` covers ×2, ×3, and
every future ×N multiplier. `data/engine-inventory.json` regenerated:
`QuantityModification` now reports `[Times, Half, Plus, Minus, Prevent]` with
`sibling_clusters: []`.

### Parser (the detector)
`scan_token_multiplier_factor` recognizes `"twice that many"` (×2) and
`"<N> times that many"` (×N) via a word-boundary nom scan that delegates the
number word to the shared `parse_number` combinator — so any ×N multiplier card
parses without per-factor tags. The CR 111.1 creature-type gate
(`TokenCoreTypeMatches`) was lifted out of the substitution-only arm and now
applies to every typed token shape, so Ojer Taq's multiplier fires only on
creature tokens.

## Per-card ship / defer

| Card | Status | Rationale |
|------|--------|-----------|
| **Ojer Taq, Deepest Foundation** | ✅ Shipped | "If one or more creature tokens would be created under your control, three times that many of those tokens are created instead." → `CreateToken` replacement with `Times { factor: 3 }`, gated to creature tokens (CR 614.1a + CR 111.1). Vigilance + dies-trigger already parsed; confirmed still clean. |
| **Moonlit Meditation** | ⏸ Deferred (fail-closed) | "The first time you would create one or more tokens each turn, you may instead create that many tokens that are copies of enchanted permanent." Needs three new pieces of infra beyond the multiplier axis: a per-turn token-creation gate (state + `ReplacementCondition` + reset), an **Optional** CreateToken pipeline (today only ZoneChange/shock-land replacements decline), and a dynamic host-copy spec resolved against the Aura host. Out of scope for this parameterization; stays honestly `Effect::unimplemented`. |

## add-engine-variant verdict
`QuantityModification::Times { factor }`:
- **Stage 1 (existence):** ×N concept absent on `QuantityModification` (only ×2 `Double`); peer `ManaModification::Multiply { factor }` exists.
- **Stage 2 (parameterization):** REFACTOR_FIRST — `Double` + a hypothetical `Triple` is the multiplicative sibling smell. Collapsed to one `Times { factor }`.
- **Stage 3 (categorical boundary):** WITHIN_SECTION — the multiply axis is entirely within CR 614.1a (replacement quantity modification).

## Discriminating-test map

| Behavioral claim | Seam / fn | Production entry | Test | Revert-failing assertion | Negative / sibling |
|---|---|---|---|---|---|
| Ojer Taq triplicates creature tokens (×3) | `create_token_applier` count path + `Times` factor | `replace_event` (real pipeline) on parser output | `game::replacement::tests::ojer_taq_triplicates_creature_tokens` | `count == 6` (2×3); flips to 4 if factor→`DOUBLE`, to 2 if replacement dropped | — |
| Multiplier gated to creature tokens (CR 111.1) | `TokenCoreTypeMatches` condition | `replace_event` with an Artifact (Treasure) token | `game::replacement::tests::ojer_taq_does_not_multiply_noncreature_tokens` | `count == 2` (unchanged); without the gate it would be 6 | non-creature case |
| Line parses to `Times { factor: 3 }` + gate | `scan_token_multiplier_factor` / `parse_token_replacement` | `parse_replacement_line` | `parser::oracle_replacement::tests::ojer_taq_token_triplication_replacement` | `quantity_modification == Times { factor: 3 }` + `TokenCoreTypeMatches([Creature])` | — |
| Full card zero-unimplemented | `parse_oracle_text` | full front-face oracle | `std_longtail_e::ojer_taq_token_triplication_full_card_parses` | `assert_zero_unimplemented` + `Times { factor: 3 }` | — |
| ×2 migration preserved | `Double` → `Times { factor: 2 }` | `parse_replacement_line` | `token_doubling_via_twice_is_factor_two`, `token_doubling_replacement*` | `quantity_modification == Times { factor: 2 }` | — |
| ×N commute (CR 616.1) unchanged | `quantity_commute_class` | `replace_event` (3 doublers) | `multiple_pure_token_doublers_commute_no_prompt` | `count == 8` auto-resolved | migration regression |

The two runtime tests drive the real `replace_event` pipeline from the actual
parser output (parse → pipeline end-to-end), and the negative case reaches the
true `TokenCoreTypeMatches` arm with a non-degenerate Artifact fixture.

## Gate results
`cargo fmt` ✅ · `check-parser-combinators.sh` (and `upstream/main`) exit 0 ✅ ·
`check-engine-authorities.sh upstream/main` exit 0 ✅ · parser diff string-dispatch
grep: only the canonical word-boundary cursor `.find(' ')` (structural advance,
not dispatch — cf. `scan_for_phase`) ✅ · CR-annotation diff gate: 0 UNVERIFIED ✅ ·
`cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` exit 0 ✅ ·
`cargo coverage` exit 0 ✅ · `cargo semantic-audit` exit 0 (Ojer Taq: 0 findings) ✅ ·
`cargo test -p engine`: 13108 passed, only the known parallel-flaky
`delve_payment_skips_state_clone_per_graveyard_candidate` and
`target_selection_legal_actions_do_not_fall_back_to_stale_slot_targets` flapped —
both pass in isolation `--test-threads=1`; neither touches this diff.

### DIFF
--- a/crates/engine/src/game/casting_costs.rs
+++ b/crates/engine/src/game/casting_costs.rs
@@ -14612,7 +14612,7 @@ its replicate cost was paid.)\nDraw a card.";
 
         // CR 614.1a: counter-doubling replacement effect (Doubling Season-class).
         let repl = ReplacementDefinition::new(ReplacementEvent::AddCounter)
-            .quantity_modification(QuantityModification::Double);
+            .quantity_modification(QuantityModification::DOUBLE);
         runner
             .state_mut()
             .objects
--- a/crates/engine/src/game/effects/blight.rs
+++ b/crates/engine/src/game/effects/blight.rs
@@ -299,7 +299,7 @@ mod tests {
             Zone::Battlefield,
         );
         let repl = ReplacementDefinition::new(ReplacementEvent::AddCounter)
-            .quantity_modification(QuantityModification::Double);
+            .quantity_modification(QuantityModification::DOUBLE);
         state
             .objects
             .get_mut(&doubler)
--- a/crates/engine/src/game/effects/copy_spell.rs
+++ b/crates/engine/src/game/effects/copy_spell.rs
@@ -452,7 +452,7 @@ pub(crate) fn copy_count_with_replacements(
             continue;
         }
         count = match def.quantity_modification {
-            Some(QuantityModification::Double) => count.saturating_mul(2),
+            Some(QuantityModification::Times { factor }) => count.saturating_mul(factor as usize),
             Some(QuantityModification::Half) => count / 2,
             Some(QuantityModification::Plus { value }) => count.saturating_add(value as usize),
             Some(QuantityModification::Minus { value }) => count.saturating_sub(value as usize),
--- a/crates/engine/src/game/effects/counters.rs
+++ b/crates/engine/src/game/effects/counters.rs
@@ -2061,7 +2061,7 @@ mod tests {
             .replacement_definitions
             .push(
                 ReplacementDefinition::new(ReplacementEvent::AddCounter)
-                    .quantity_modification(QuantityModification::Double),
+                    .quantity_modification(QuantityModification::DOUBLE),
             );
 
         let plus_id = create_object(
@@ -3276,7 +3276,7 @@ mod tests {
             .insert(CounterType::Plus1Plus1, 1);
         let mut repl = ReplacementDefinition::new(ReplacementEvent::AddCounter);
         repl.valid_card = Some(TargetFilter::SelfRef);
-        repl.quantity_modification = Some(QuantityModification::Double);
+        repl.quantity_modification = Some(QuantityModification::DOUBLE);
         state
             .objects
             .get_mut(&dest_id)
@@ -4030,7 +4030,7 @@ mod tests {
         );
         let mut doubler_repl = ReplacementDefinition::new(ReplacementEvent::AddCounter);
         doubler_repl.valid_card = Some(TargetFilter::Any);
-        doubler_repl.quantity_modification = Some(QuantityModification::Double);
+        doubler_repl.quantity_modification = Some(QuantityModification::DOUBLE);
         state
             .objects
             .get_mut(&doubler_id)
--- a/crates/engine/src/game/effects/double.rs
+++ b/crates/engine/src/game/effects/double.rs
@@ -393,7 +393,7 @@ mod tests {
     fn double_counters_replacement_choice_stashes_remaining_counter_additions() {
         let mut state = GameState::default();
         for (id, modification) in [
-            (ObjectId(90), QuantityModification::Double),
+            (ObjectId(90), QuantityModification::DOUBLE),
             (ObjectId(91), QuantityModification::Plus { value: 1 }),
         ] {
             let mut source = GameObject::new(
@@ -599,7 +599,7 @@ mod tests {
         );
         let mut repl = ReplacementDefinition::new(ReplacementEvent::AddCounter);
         repl.valid_card = Some(TargetFilter::Any);
-        repl.quantity_modification = Some(QuantityModification::Double);
+        repl.quantity_modification = Some(QuantityModification::DOUBLE);
         doubler.replacement_definitions.push(repl);
         state.objects.insert(doubler_id, doubler);
         state.battlefield.push_back(doubler_id);
--- a/crates/engine/src/game/effects/proliferate.rs
+++ b/crates/engine/src/game/effects/proliferate.rs
@@ -718,7 +718,7 @@ mod tests {
     fn apply_proliferate_replacement_choice_stashes_remaining_counter_additions() {
         let mut state = GameState::new_two_player(42);
         for (id, modification) in [
-            (ObjectId(90), QuantityModification::Double),
+            (ObjectId(90), QuantityModification::DOUBLE),
             (ObjectId(91), QuantityModification::Plus { value: 1 }),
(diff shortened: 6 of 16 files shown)


matthewevans (changes requested):
Blocking this head on the current required Rust test failure.

HIGH: replacing `QuantityModification::Double` breaks existing serialized game states

The PR removes the `Double` enum variant and replaces it with `Times { factor }` in `crates/engine/src/types/ability.rs:15102-15113`. That is a reasonable type-shape direction for the new `xN` axis, but the repository already has serialized game-state data that still contains `{"type":"Double"}`. GitHub CI run 27920679753 fails in `phase-ai::community_scenarios::community_ai_scenarios_choose_expected_action_type` because the `saheeli-copy-sacrifice` scenario cannot deserialize: `unknown variant Double, expected one of Times, Half, Plus, Minus, Prevent`.

This is not unrelated drift: it is a direct consequence of removing the serialized variant without migrating the repo-owned fixtures or providing an explicit serde migration path. The engine tests and AI scenario corpus must continue to load after this change. Please either migrate the checked-in serialized scenario data to the new `Times { factor: 2 }` shape or add a deliberate compatibility path for old `Double` data, then rerun the Rust tests.

The lint/parser/card-data checks are green and the implementation appears to be aiming at the right parameterized building block, but the current head is red and cannot be approved or enqueued.

lgray:

Thanks for the precise diagnosis @matthewevans — addressed and rebased.

**Root cause (the blocking HIGH).** Parameterizing `QuantityModification::Double` (unit) into `Times { factor }` removed the `Double` serialized variant, so repo-owned saved game states captured before the change failed to deserialize (`unknown variant Double, expected one of Times, Half, Plus, Minus, Prevent`). CI run 27920679753 failed in `phase-ai::community_scenarios::community_ai_scenarios_choose_expected_action_type` on the `saheeli-copy-sacrifice` fixture, whose archive carries `quantity_modification: {"type":"Double"}` at 24 distinct paths (object `replacement_definitions` / `base_replacement_definitions`, deck-pool card `replacements`, all repeated under `turnCheckpoints`).

**Fix.** Added a scoped serde migration in `crates/phase-ai/src/saved_state.rs` (`migrate_legacy_double_quantity`) that rewrites the legacy `quantity_modification` unit variant `{"type":"Double"}` → `{"type":"Times","factor":2}`. It is keyed on the `quantity_modification` field name and recurses through arrays / nested objects / checkpoints, so it reaches every occurrence while leaving the unrelated `Effect::Double` enum (which legitimately still serializes as `{"type":"Double"}` under `mode_abilities[].effect`) untouched. A discriminating unit test asserts both the rewrite and the `Effect::Double` preservation.

Verified the previously-red test now passes locally: `cargo test -p phase-ai --test community_scenarios` → 1 passed. All three community archives (`saheeli-copy`, `saheeli-legend`, `helix`) load again.

**Also in this push.** Self-review caught one building-block reuse miss: `scan_token_multiplier_factor` was hand-rolling the word-boundary scan loop — it now delegates to the shared `nom_primitives::scan_at_word_boundaries` helper (behavior-preserving; the ×2 / ×N parser tests stay green).

**Rebase.** Rebased cleanly onto current `upstream/main` (`39dcbda80`).

**Verification (foreground, worktree).** `cargo fmt --all` clean; parser-combinator gate (with and without base) exit 0; engine-authorities check exit 0; `clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings` clean; `cargo test -p engine` → 13124 passed, only the two known parallel-flaky `ai_support` clone-count tests fail under parallelism and pass in isolation (`--test-threads=1`). CR annotations (614.1a, 616.1, 111.1) grep-verified against the comprehensive rules.

**New head:** `6a440f08503d773130ae0d41eaecfe1af1a1db2e`

matthewevans (approved):
Approved after re-review of the follow-up at `6a440f08503d773130ae0d41eaecfe1af1a1db2e`.

Evidence checked:
- `QuantityModification::Double` was parameterized to `Times { factor }` with `QuantityModification::DOUBLE` preserving the common x2 constructor.
- Runtime appliers for life gain, counters, token creation, copy counts, and CR 616.1 commute classification now handle `Times { factor }`.
- `phase-ai` saved-state loading migrates legacy `quantity_modification: {"type":"Double"}` to `Times { factor: 2 }` without rewriting unrelated `Effect::Double`.
- Parser and runtime tests cover Ojer Taq factor 3 and the creature-token gate, including a non-creature token negative case.
- GitHub checks are green and the PR is CLEAN.

Label: `feature` is appropriate; this adds engine/parser support for a new replacement-effect class, not just tests.

PR #4103, opened 2026-06-21
feat(engine): for-each-category pool-exile iteration (Sanar, Portent of Calamity)

lgray:

## For-each-category pool-exile iteration

Adds the **"for each [category member], you may exile a card of that member from the revealed pool"** building block — iteration over a *fixed, closed* category enumeration, distinct from per-object "for each creature" counting. The category is a printed rules enumeration (the five colors, CR 105.1; or the seven main card types, CR 205.2a), not game state.

### The building block

- **`IterationCategory`** (typed enum: `Color` | `CardType`) — supplies the ordered per-member candidate filters (`FilterProp::HasColor` for colors in WUBRG order; `TypeFilter` for the seven main card types in CR 205.2a order). The categorical boundary is respected: colors (CR 105) and card types (CR 205) are separate axes, not conflated.
- **`Effect::ForEachCategoryExile { category, zone, chooser, up_to }`** — the category-iteration sibling of `ChooseFromZone { zone_owner: EachPlayer }`. It parks one `ChooseFromZoneChoice` per member over a pool snapshot restricted to that member's color/type, exiles each pick, and accumulates the exiled cards into the resolution chain's tracked object set ("the cards exiled this way") so a downstream "from among them" / "put the rest …" clause reads exactly those cards (CR 608.2c).
- **`PendingPerCategoryZoneChoice`** mirrors the existing `PendingPerPlayerZoneChoice`; it is drained at the same `apply` resolution-choice seam, immediately before `pending_continuation`.
- **Parser** — `try_parse_for_each_category_exile` composes the category axis (`color` vs `card type`, incl. anaphoric plural "of those colors") and the pool reference ("them" / "the revealed cards") with nom combinators. No string dispatch (`contains`/`split_once`/etc.). Optionality ("you may exile") lives in `up_to` (0..=1 per member), not a bool.

### Parameterize-don't-proliferate verdict

Verified `data/engine-inventory.json` — no existing for-each-category iteration variant. This reuses the proven per-iteration parking + tracked-set-accumulation machinery (`prompt_next_each_player`) by adding a sibling iteration unit (category member) rather than a new resolution subsystem. The iteration *unit* is the parameterization axis; player-iteration and category-iteration remain type-distinct (a color/card type is not a `ZoneOwner`), so they are not crammed into `ZoneOwner`.

### Per-card ship / defer

| Card | For-each clause | Card status | Notes |
|------|-----------------|-------------|-------|
| **Sanar, Innovative First-Year** | shipped (`ForEachCategoryExile { Color }`) | **0 gaps — fully supported** | "For each of those colors, you may exile a card of that color from among the revealed cards." All surrounding clauses (RevealUntil, shuffle, cast-exiled) parse. (CR 105.1, CR 608.2c) |
| **Portent of Calamity** | shipped (`ForEachCategoryExile { CardType }`) | for-each clause shipped; **1 OTHER residual gap** | "For each card type, you may exile a card of that type from among them." Remaining gap is a pre-existing swallowed `Condition_If` ("if you exiled four or more cards this way") — unrelated to this building block. (CR 205.2a, CR 608.2c) |
| **Call the Spirit Dragons** | **deferred** | still gapped | Its for-each-color payload is "put a +1/+1 counter on a Dragon you control of that color" — a battlefield counter on a permanent of the bound color, a *different (non-pool) payload shape*. Shipping it would require a second per-member payload mechanism plus its win-rider counting; that is a distinct building block, deferred honestly via `Effect::Unimplemented`. |

### Discriminating-test map

| Behavioral claim | Seam | Production entry | Test | Revert-failing assertion |
|---|---|---|---|---|
| Per-member candidate pool is restricted to the bound color | `filter_category_pool` | `resolve_for_each_category` | `for_each_color_exile_first_member_offers_only_that_color` | White member prompt offers `[white]` only (verified: breaking the filter fails the test) |
| Iteration runs once per member, exiling one card each, accumulating into the chain set | `resolve_for_each_category` → `drain_pending_per_category_zone_choice` via `apply(SelectCards)` | `apply()` resolution-choice handler | `for_each_color_exile_iterates_each_member_through_apply` | all 3 colored cards end in Exile; chain tracked set holds exactly the 3 |
| "you may" makes each member optional (decline keeps the card) | `up_to` path | `apply(SelectCards { [] })` | `for_each_color_exile_member_decline_keeps_card` | declined White stays in Library; Blue still resolves |
| Parser color + plural-anaphor forms | `try_parse_for_each_category_exile` | `parse_effect` | `for_each_color_exile_from_revealed_cards` | `ForEachCategoryExile { Color }` |
| Parser card-type form | same | `parse_effect` | `for_each_card_type_exile_from_among_them` | `ForEachCategoryExile { CardType }` |
| Exile parser does not shadow distinct-types put-to-hand | dispatch order | `parse_effect` | `for_each_category_exile_does_not_shadow_put_to_hand` | put-to-hand stays `ChooseFromZone` |

The runtime tests drive the real `apply()` pipeline and submit actual `GameAction::SelectCards` per member — not AST-shape assertions.

### Gates

`cargo fmt` clean · `check-parser-combinators.sh` (with and without `upstream/main`) exit 0 · parser diff string-dispatch grep clean · `check-engine-authorities.sh upstream/main` exit 0 · `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings` clean · `cargo semantic-audit` / `cargo coverage` clean (no findings on the affected cards) · `cargo test -p engine` 13111 passed, only the known parallel-flaky `delve_payment_skips_state_clone_per_graveyard_candidate` failed under parallelism and passes in isolation (`--test-threads=1`). All CR annotations grep-verified against `docs/MagicCompRules.txt`.


matthewevans (changes requested):
Requesting changes after a Tier 2 engine/parser review.

Findings:

1. HIGH: the per-category iterator does not actually resume the parked continuation after the last member. `drain_pending_per_category_zone_choice` calls `prompt_next_category_member`; when no members remain, that helper only emits `EffectResolved` and returns without clearing `state.waiting_for` (`crates/engine/src/game/effects/choose_from_zone.rs:187-194`, `:237-244`). The action handler then immediately calls `effects::drain_pending_continuation`, but that helper returns early for any `WaitingFor::ChooseFromZoneChoice` (`crates/engine/src/game/effects/mod.rs:535-536`, `:1635`). At this point `state.waiting_for` is still the just-answered member prompt, so the continuation is skipped and remains parked. The new full-iteration test checks that `pending_per_category_zone_choice` is gone, but it does not assert that the waiting state returned to priority or that a downstream "from among them" / "put the rest" continuation actually ran. Sanar and Portent both need that tail to resolve after the category sweep.

2. HIGH: even after the continuation is made to run, the all-decline path leaves the chain tracked set bound to the original revealed pool instead of an empty "cards exiled this way" set. `drain_pending_per_category_zone_choice` only publishes a fresh tracked set after a non-empty choice (`choose_from_zone.rs:224-235`); if the player declines every member, `accumulated` stays false and `chain_tracked_set_id` continues to point at the producer pool captured by `resolve_category_pool` (`choose_from_zone.rs:247-267`). A subsequent "you may cast a spell from among them" would then see the revealed cards even though no cards were exiled this way. Please add a regression that declines every eligible member with a tracked-set-consuming continuation and verifies the continuation sees an empty set.

3. HIGH: `IterationCategory::CardType` omits legal card types from CR 205.2a. The implementation hard-codes only artifact, creature, enchantment, instant, land, planeswalker, and sorcery (`crates/engine/src/types/ability.rs:113-122`) and the comment says battle/kindred/etc. are never offered (`ability.rs:77-79`). The local CR text says card types are "artifact, battle, conspiracy, creature, dungeon, enchantment, instant, kindred, land, phenomenon, plane, planeswalker, scheme, sorcery, and vanguard" (`docs/MagicCompRules.txt:1404`), and this codebase already models at least `CoreType::Battle`/`Kindred` plus `TypeFilter::Battle` (`crates/engine/src/types/card_type.rs:43-79`, `crates/engine/src/types/ability.rs:2231-2242`). Portent of Calamity says "for each card type", so a Battle card in the pool must be eligible for the battle member. Either implement the CR 205.2a member set with matching filters or leave this unsupported rather than shipping a seven-type special case.

The parser direction and the category-axis abstraction are reasonable, but the current behavior is not rules-correct for the cards named in the PR.

lgray:

Thanks for the Tier 2 review @matthewevans — all three findings are addressed. New head: `11be9b1d2`, rebased onto current upstream/main.

**1. CR 205.2a card-type members (HIGH).** You were right that folding kindred was wrong. `IterationCategory::CardType` now offers all nine library-reachable card types in CR 205.2a order, with **Kindred as a distinct member** via a new `TypeFilter::Kindred` variant matched against the already-modeled `CoreType::Kindred`. Per CR 308.1 a kindred card has another card type, so it is now exilable at *both* the Kindred member and its other-type member — folding undercounted exiles toward Portent of Calamity's "four or more cards exiled this way" threshold (e.g. a Kindred Sorcery + a plain Sorcery should yield two exiles, not one). The variant is wired into every exhaustive matcher (filter ×4, restrictions, coverage, token, trigger_index, the typed condition mappers, and the keyword subject mapper); the compiler confirms exhaustiveness. New discriminating test `for_each_card_type_offers_kindred_member_distinct_from_cotype` asserts two distinct prompts exile two distinct cards, with a negative case (a plain Sorcery is not offered at the Kindred member); removing the Kindred member fails the test.

**2. All-decline exiled-set binding (HIGH).** The chain tracked set is rebound to a fresh **empty** "cards exiled this way" set at iteration start (after the producer pool is snapshotted) and extended only on a non-empty pick, so a downstream "cast from among them" sees nothing when every member is declined (CR 608.2c). Covered by `for_each_color_exile_all_decline_leaves_empty_tracked_set_for_continuation` — reverting the empty-set rebind makes the continuation scoop the revealed pool, failing the test.

**3. Continuation tail after the last member (HIGH).** `set_priority` runs before the per-category branch, so `waiting_for == Priority` when the last member emits `EffectResolved` and `drain_pending_continuation` runs — the tail already resumes. The previously thin assertion is replaced by `for_each_color_exile_runs_tracked_set_continuation_after_last_member`, which drives the real chain and asserts the downstream tracked-set continuation executed (cards moved to hand) **and** priority returned.

Also corrected the `up_to` optionality annotation (CR 609.3 → CR 608.2d, the resolution-time optional-choice rule) and a "seven main card types" doc drift, and dropped a redundant guard.

Verification (Tilt is not active in the worktree, so cargo was run directly): `cargo fmt --check`, `cargo check -p engine`, and `cargo clippy -p engine --lib` clean; `cargo test -p engine --lib for_each` 262 passed / 0 failed. Every added CR annotation grep-verified against the Comprehensive Rules.

matthewevans:
VERDICT: approve-pending-current-ci

Reviewed follow-up head `11be9b1d294f9423a669badbb56650a3a72a53af` against the prior findings. The parked continuation path is now addressed by the per-category choice drain before the generic continuation handler, with `waiting_for` preventing the tail from running while another member prompt is still parked and allowing it after the last member. The all-decline path now starts from a fresh empty chain tracked set, so downstream `TrackedSet(0)` consumers see the cards actually exiled this way rather than the original revealed pool. The CardType member list now includes Battle and Kindred with runtime coverage for both.

No current semantic blocker found on this head. I am not approving/enqueuing yet because GitHub still reports `mergeStateStatus=BEHIND` and Rust tests are still in progress; re-run the maintainer bar on the current post-update head before approval.

matthewevans (approved):
Approved current head `9a6b122d1e97fec7f62b6356e12e526625d74145` after the follow-up review and refreshed green CI.

Prior blockers are addressed: parked continuation drain now resumes after the final category member, all-decline publishes an empty exiled-this-way tracked set, and the CardType member list covers the modeled Battle/Kindred cases. The current tip only adds a merge-from-main commit after the reviewed implementation head, and the repush guard held before approval.

PR #4105, opened 2026-06-22
fix(parser): parse copula-goaded clauses ("it's goaded [duration]")

### LINKED ISSUE #711: jon irenicus, shattered one — only adds 2 +1/+1 counters to the creature and taps it (which its supposed to) but doesn'…
Reported in Discord: https://discord.com/channels/1485498006781427802/1506888094128013372/1506888094128013372

**Thread:** jon irenicus, shattered one
**Cards:** _none detected_

## Summary
only adds 2 +1/+1 counters to the creature and taps it (which its supposed to) but doesn't give it to an opponent and goad it.

---
<sub>report_id: `discord:1506888094128013372:1506888094128013372:0` · discord: `1506888094128013372/1506888094128013372`</sub>

dripsmvcp:
Fixes #711

## Summary
- Add `try_parse_copula_goaded_clause` to handle "it's goaded [duration]" / "it is goaded [duration]" copula state-setting clauses that `find_predicate_start` cannot split (contraction "it's" fuses subject + copula)
- Lowers to `Effect::Goad` with pronoun-resolved target and optional duration (`Permanent`, `ForAsLongAs`, or default)
- Unlocks Jon Irenicus, Shattered One (goaded for the rest of the game), Vislor Turlough (goaded for as long as they control it), and the non-contracted "it is goaded" form

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine --lib` — 13125 passed, 6 ignored (2 pre-existing flaky AI tests on baseline)
- 5 new unit tests: permanent duration, no duration, non-contracted, for-as-long-as, rejects non-goaded
- All 30 existing goad-related tests pass


matthewevans (changes requested):
Reviewed head `b9dac48a2fab414c7a4b02d2f1c51df700334aac`.

VERDICT: request-changes

**[MED]** The new copula-goaded parser accepts a duration prefix and silently drops the unconsumed tail. Evidence: `crates/engine/src/parser/oracle_effect/subject.rs:4152-4159` calls `parse_duration(stripped).ok().map(|(_, d)| d)` and then always returns `sub_ability: None`; `parse_duration` is explicitly prefix-style, with existing tests asserting leftover text like `" and"` / `" if able"`. Why it matters: a clause such as `it's goaded for the rest of the game and ...` can be lowered as just `Effect::Goad` with the remaining rules text discarded, which makes parser coverage dishonest for the next card in this class. Suggested fix: require the parsed duration remainder to be empty/period-only, or use the existing suffix-position duration wrapper so only a true clause-final duration is stripped; add a production parser regression that proves trailing non-duration text is preserved/rejected rather than only direct helper tests.

Confidence: high. This is based on the current PR diff and the existing `parse_duration` contract/tests in `oracle_nom/duration.rs`; a contradiction would be evidence that `try_parse_subject_predicate_ast` can never receive a clause with additional text after this helper runs, but the helper itself currently accepts such input and discards the returned remainder.

matthewevans (approved):
Approved after re-review and fresh CI on the current merge head.

The follow-up fixed the parser-consumption blocker by requiring the parsed duration to consume the whole remaining copula-goad clause instead of silently dropping trailing text. GitHub checks are green on this head, the required `bug` label is present, and the repush guard matched the reviewed head.

PR #4108, opened 2026-06-22
fix(parser): parse "at least N mana was spent to cast" threshold conditions

### LINKED ISSUE #1490: The Emperor of Palamecia // The Lord Master of Hell — this cars has a bug on both sides.
Reported in Discord: https://discord.com/channels/1485498006781427802/1509743438474514603/1509743438474514603

**Thread:** The Emperor of Palamecia // The Lord Master of Hell
**Cards:** the emperor of palamecia, the lord master of hell
**Parser status:** has_gaps

## Summary
this cars has a bug on both sides.

### the emperor of palamecia
{T}: Add {U} or {R}. Spend this mana only to cast a noncreature spell.
Whenever you cast a noncreature spell, if at least four mana was spent to cast it, put a +1/+1 counter on The Emperor of Palamecia. Then if it has three or more +1/+1 counters on it, transform it.
### the lord master of hell
Starfall — Whenever The Lord Master of Hell attacks, it deals X damage to each opponent, where X is the number of noncreature, nonland cards in your graveyard.

---
<sub>report_id: `discord:1509743438474514603:1509743438474514603:0` · discord: `1509743438474514603/1509743438474514603`</sub>

dripsmvcp:
Fixes #1490

## Summary
- Extend `parse_mana_spent_threshold` and `parse_source_qualified_mana_spent_threshold` to recognize the "at least N" prefix as an alternative to the existing "N or more" pattern — both are `>= N` thresholds (CR 601.2h)
- Cards like The Emperor of Palamecia whose trigger reads "if at least four mana was spent to cast it" were silently losing the intervening-if condition, causing the trigger to fire unconditionally on any noncreature spell regardless of mana spent

## Verification
- `cargo fmt --all` — clean
- 97 mana-spent-related tests pass (condition + trigger)
- 329 full condition parser tests pass
- New tests: `test_at_least_mana_spent_threshold_parses`, `test_mana_spent_condition_subject_drives_scope` (extended), `trigger_at_least_mana_spent_intervening_if`


matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** The new `at least N mana was spent to cast it` trigger condition keeps `it` scoped to the trigger source instead of the spell that caused the trigger. Evidence: `parse_mana_spent_threshold` preserves the subject returned by `parse_mana_spent_self_subject`, so `it` becomes `CastManaObjectScope::SelfObject` at `crates/engine/src/parser/oracle_nom/condition.rs:4298-4307`; the new trigger regression then accepts any `ManaSpentToCast` scope by using `..` at `crates/engine/src/parser/oracle_trigger.rs:18640-18647`; and the quantity resolver only reads the spell-cast event for `CastManaObjectScope::TriggeringSpell`, while `SelfObject` reads `ctx.self_object()` at `crates/engine/src/game/quantity.rs:2284-2290`. Why it matters: The Emperor of Palamecia-style intervening-if would evaluate the permanent/source object's mana-spent snapshot, normally `0`, so the trigger is blocked even when the triggering noncreature spell was cast with four or more mana. Suggested fix: in the trigger/intervening-if path, make this clause produce `ManaSpentToCast { scope: TriggeringSpell, .. }` for the triggering spell case and tighten the regression to assert that exact scope, plus a runtime trigger gate if practical.

matthewevans (approved):
Approved after re-review and fresh CI on the current merge head.

The follow-up moved the spell-cast trigger case to the correct event-scoped `TriggeringSpell` reference and added a regression that asserts the lowered condition uses that scope. GitHub checks are green on this head, the required `bug` label is present, and the repush guard matched the reviewed head.

Issue #3663, opened 2026-06-17
treasure cruise — Odd, but it almost feels like treasure cruise casts are causing a significant slow down.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1516506640030105771/1516506640030105771

**Thread:** treasure cruise
**Cards:** treasure cruise
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Odd, but it almost feels like treasure cruise casts are causing a significant slow down.

### treasure cruise
Delve (Each card you exile from your graveyard while casting this spell pays for {1}.)
Draw three cards.

---
<sub>report_id: `discord:1516506640030105771:1516506640030105771:0` · discord: `1516506640030105771/1516506640030105771`</sub>

PR #4110, opened 2026-06-22
test(engine): make perf counters thread-local to de-flake legality-clone tests

lgray:

## Root cause

The perf counters in `crates/engine/src/game/perf_counters.rs` were process-global `static AtomicU64`s shared across every thread. Under parallel `cargo test`, an unrelated test running concurrently on another thread could increment the same counter that a legality-clone test reads between its `reset()` and `snapshot()`. That cross-thread pollution made three tests flaky in parallel:

- `delve_payment_skips_state_clone_per_graveyard_candidate`
- `target_selection_legal_actions_do_not_simulate_each_target`
- `target_selection_legal_actions_do_not_fall_back_to_stale_slot_targets`

## Fix

Convert the counters to a thread-local `Cell<PerfCounterSnapshot>`. `record_*` / `snapshot` / `reset` now operate on the current thread's counter. The public API is **byte-identical** — same `PerfCounterSnapshot` struct, same function signatures, **zero call-site changes**. A doc comment on the `thread_local!` records the per-thread-isolation rationale so it isn't reverted to a global.

## Discrimination preserved (the #3663 guard still fires)

The counted legality / delve / cost paths run **entirely on the calling thread** — the clone-recording sites (`ai_support/filter.rs`, `game/costs.rs`, `game/mana_abilities.rs`) all `record_state_clone_for_legality()` synchronously inline, and the call chains from `legal_actions_full` / `validated_candidate_actions` are plain serial iterators (no rayon, no spawned threads, no async on these paths). So the thread-local counts exactly that thread's clones: a revert to per-candidate / per-target cloning still increments the counter the test reads, and the assertions (`clones == 0`, `clones == 0`, `clones < 5`) still fail. The de-flake does **not** weaken the regression guard.

The only counter consumers are these three engine unit tests plus the single-threaded `legal-actions-bench` / `resolve_bench` dev binaries. There is no production or CI telemetry that needs a cross-thread aggregate.

## Verification (gate)

- **50× parallel stress** `cargo test -p engine --lib -- --test-threads=8 <the 3 tests>` → **50/50 pass** (was flaky before).
- `cargo fmt --all` clean; `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` clean.
- `cargo test -p engine` full suite green (the 3 tests now pass in the default parallel run).
- `./scripts/check-engine-authorities.sh upstream/main` exit 0.
- `cargo build -p phase-ai --bin legal-actions-bench --bin resolve_bench` builds.


matthewevans:
Reviewed current head `dd3003b98aeedb17a273af8d22b2771bfdd0f449` via the GitHub API diff and isolated worktree. No semantic blocker found: the perf-counter API shape is unchanged, the counted legality/cost/layer/stack record sites are synchronous on the calling thread, and the only `reset()`/`snapshot()` consumers are the targeted regression tests plus the single-threaded `legal-actions-bench` / `resolve_bench` dev binaries.

Required `test` label applied. Holding approval/enqueue until the branch is current with `main` and fresh CI is green.

matthewevans (approved):
Current head `d6955f8f91f3fd526268dda569d5c97d5ed91364` is green and matches the reviewed perf-counter thread-local de-flake. The change is test-infra scoped, preserves the existing public perf counter API, and the required `test` label is present.

PR #4114, opened 2026-06-22
fix(parser): parse mill, remove-counter, and numeric-discard unless costs

dripsmvcp:
## Summary
- Extend `parse_unless_alt_cost` with three new non-mana unless-cost branches that map to existing `AbilityCost` variants
- **Mill**: "you mill [N] cards" → `AbilityCost::Mill` (Deep Spawn)
- **Remove counter**: "you remove [N] [type] counter(s) from [target]" → `AbilityCost::RemoveCounter` (Chisei Heart of Oceans, Junk Golem, Magmatic Sprinter)
- **Numeric discard**: "you discard [N] cards" with word-number count → `AbilityCost::Discard` (Avatar of Discord)
- Fix `parse_unless_discard_cost` to handle numeric word counts ("two", "three") before the existing article-form ("a"/"an") path

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine --lib` — 13137 passed, 0 failed
- `cargo test -p engine --lib trigger_unless` — 44 passed (5 new + 39 existing), 0 regressions
- `cargo test -p engine --lib swallow_check` — 89 passed, 0 regressions

### DIFF
--- a/crates/engine/src/game/engine_payment_choices.rs
+++ b/crates/engine/src/game/engine_payment_choices.rs
@@ -909,9 +909,10 @@ pub(super) fn handle_unless_payment(
             // through the replacement pipeline so Rest-in-Peace class
             // redirects fire correctly. Partial mill (library has fewer than
             // N cards) is an unpayable cost per CR 118.3 — effect fires.
-            // A replacement-effect pause (BatchMoveResult::Incomplete) also
-            // counts as payment failure so the unless-effect is not
-            // double-triggered; the paused replacement state is cleared.
+            // A CR 616.1 replacement ordering choice parks the batch in
+            // state.waiting_for + state.pending_batch_deliveries; callers
+            // must early-return so they do not clobber the parked prompt
+            // (mirrors apply_etb_counters early-return in handle_replacement_choice).
             AbilityCost::Mill { count } => {
                 let player_library_len = state
                     .players
@@ -934,9 +935,11 @@ pub(super) fn handle_unless_payment(
                         .map_err(|e| EngineError::InvalidAction(format!("{e:?}")))?
                     {
                         true => {}
+                        // CR 616.1: replacement ordering choice parked — the
+                        // mill batch is in progress. Early-return to preserve
+                        // state.waiting_for + state.pending_batch_deliveries.
                         false => {
-                            state.pending_replacement = None;
-                            payment_failed = true;
+                            return Ok(action_result(events, state.waiting_for.clone()));
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
[P1] Wire the new unless costs into payment resolution

This parser now emits `AbilityCost::Mill` for `you mill ...` clauses and `AbilityCost::RemoveCounter` for `you remove ... counter(s) ...` clauses (`crates/engine/src/parser/oracle_trigger.rs` around the new `parse_unless_mill_cost` / `parse_unless_remove_counter_cost` paths). Those costs then enter the existing `WaitingFor::UnlessPayment` flow, but `handle_unless_payment` still lists both `AbilityCost::RemoveCounter { .. }` and `AbilityCost::Mill { .. }` in the unsupported fall-through arm that sets `payment_failed = true` (`crates/engine/src/game/engine_payment_choices.rs`). As a result, the newly parsed cards become worse than unsupported: even when the player chooses to pay and has the cards/counters available, the engine treats the payment as failed and resolves the punishment effect anyway.

Please either add the runtime payment support for these two unless-cost variants, with an end-to-end test that paying suppresses the effect, or keep these clauses unsupported until the payment path exists. Parser-shape tests alone are not enough here because the failure is after parsing, at the payment seam.

matthewevans (changes requested):
[P1] Preserve replacement-choice pauses while paying mill unless-costs

The new `AbilityCost::Mill` payment arm treats `apply_mill_after_replacement(...) == false` as payment failure and clears replacement state (`crates/engine/src/game/engine_payment_choices.rs:933-940`). That is the opposite of the mill helper's contract: `apply_mill_after_replacement` documents that `false` means a CR 616.1 replacement ordering choice parked the batch in `waiting_for` / `pending_batch_deliveries`, and callers must early-return so they do not clobber the prompt (`crates/engine/src/game/effects/mill.rs:90-95`). With multiple graveyard replacement effects, a player choosing to pay Deep Spawn's mill cost would get a replacement prompt, but this path marks the payment failed and proceeds toward the punishment effect instead of resuming the payment after the parked batch. Please preserve the parked replacement state/resume flow and add a regression where an unless-mill payment pauses on replacement choice and still suppresses the effect after the payment completes.

[P1] Do not parse targeted remove-counter unless-costs until they are payable

`parse_unless_remove_counter_cost` still converts phrases like Chisei's "from a permanent you control" into `AbilityCost::RemoveCounter { target: Some(filter), ... }` (`crates/engine/src/parser/oracle_trigger.rs:2167-2187`), but `handle_unless_payment` leaves every `RemoveCounter { target: Some(_), .. }` in the unsupported fall-through that sets `payment_failed = true` (`crates/engine/src/game/engine_payment_choices.rs:1031-1037`). That means a payable Chisei cost is still impossible to pay in gameplay: choosing to pay fires the sacrifice/gain-life effect anyway. This is the same coverage-honesty problem as the original review for the targeted variant. Please either implement the target-choice/payment path with an end-to-end Chisei-style regression, or leave that targeted parser form unsupported until the runtime flow exists.

matthewevans (changes requested):
VERDICT: request-changes

The targeted remove-counter part of the prior review is addressed: `parse_unless_remove_counter_cost` now leaves non-self targets unsupported, so Chisei-style text no longer emits an unpayable `RemoveCounter { target: Some(_) }`.

One prior blocker is still live:

**[HIGH] Mill unless-costs still clobber replacement-choice pauses.** Evidence: `crates/engine/src/game/engine_payment_choices.rs:933-940` still treats `apply_mill_after_replacement(...) == false` as payment failure, clears `state.pending_replacement`, and continues toward resolving the punishment effect. The mill helper's contract at `crates/engine/src/game/effects/mill.rs:89-95` says `false` means a CR 616.1 replacement ordering choice parked `state.waiting_for` plus `state.pending_batch_deliveries`, and callers must early-return so they do not clobber the parked prompt. Why it matters: if Deep Spawn's mill payment hits multiple graveyard replacement effects, choosing to pay can still lose the replacement prompt and fire the unless effect instead of resuming and suppressing it after the paid mill completes. Suggested fix: preserve the parked replacement state/resume flow for `Mill` unless-payments and add the replacement-choice regression requested in the prior review.

No approval, label, or queue action from me on this head.

matthewevans:

The prior blockers are addressed on this head. The non-self/targeted remove-counter unless-cost path now remains unsupported instead of pretending to emit a payable `RemoveCounter`, and the mill unless-cost path now preserves the CR 616.1 replacement-ordering pause by early-returning when `apply_mill_after_replacement` parks `WaitingFor::ReplacementChoice` plus the pending batch. The new unless-payment regression covers that exact pause and proves the punishment effect does not fire while replacement ordering is pending.

I also checked Gemini activity; the only Gemini comment on this PR is a quota warning, so there is no substantive bot finding to apply here.

No formal approval, label, or enqueue from me yet: GitHub still has Rust lint/tests/card-data running on this exact head, and the branch is BEHIND. Once it is current and exact-head CI is green/CLEAN, this can go through the normal repush-guard approval/enqueue path.

matthewevans (approved):
Approved on exact reviewed head after repush guard. Prior blockers are addressed: targeted RemoveCounter unless-costs remain coverage-honest, Mill unless-payments preserve CR 616.1 replacement-choice pauses, and exact-head CI is green/CLEAN.

PR #4116, opened 2026-06-22
feat(engine): one-shot draw replacement ("the next time you would draw a card this turn, [effect] instead")

### LINKED ISSUE #4115: Parser: "the next time you would draw a card this turn, [effect] instead" dropped to Unimplemented (Words of Worship, Words of Wilding)
## Summary

The Odyssey "Words of" cards drop their entire activated ability to `Effect::Unimplemented`, so activating them does nothing.

- **Words of Worship** — `{1}: The next time you would draw a card this turn, you gain 5 life instead.`
- **Words of Wilding** — `{1}: The next time you would draw a card this turn, create a 2/2 green Bear creature token instead.`
- **Words of Wind** — `{1}: ... each player returns a permanent they control to its owner's hand instead.`
- **Words of Waste** — `{1}: ... each opponent discards a card instead.`

## Current (wrong) parse

The whole clause lowers to `Effect::Unimplemented { name: "the", description: "the next time you would draw a card this turn, ... instead" }` — the activated ability is a no-op.

## Root cause

The engine already *replaces* the draw event (Jace, the Mind Sculptor's "you win the game instead", Abundance) via `ReplacementDefinition` + the post-replacement continuation. But there is no resolution-time effect to **install** a one-shot, this-turn draw replacement that carries an arbitrary substitute effect — so "the next time you would draw a card this turn, [effect] instead" has no home and falls through to `Unimplemented`.

## Expected

Activating the ability installs a one-shot replacement: the controller's next draw this turn is replaced — the card is not drawn, and the substitute effect happens instead (Worship → gain 5 life; Wilding → create a 2/2 Bear). The replacement is consumed on use (a later draw the same turn is normal) and expires at end of turn if unused (CR 614.6 / 514.2).

## Notes

Words of Wind ("each player returns a permanent") and Words of Waste ("each opponent discards") use player-scoped payloads that the current effect parser mis-scopes to the controller, so they should remain honest gaps until that subject-scoping is handled. Aladdin's Lamp / Mangara's Tome use bespoke library payloads and are out of scope. CR 614.1a / 614.6 / 514.2 / 121.1.

joaovictor712:
Fixes #4115.

## Summary

The Odyssey "Words of" cards dropped their entire activated ability to `Effect::Unimplemented` — `{1}: The next time you would draw a card this turn, <effect> instead.` The engine already *replaces* the draw event (Jace's "you win the game instead", Abundance) via `ReplacementDefinition` + the post-replacement continuation, but there was no resolution-time effect to **install** a one-shot, this-turn draw replacement carrying an arbitrary substitute.

## Implementation (new `Effect::CreateDrawReplacement`, mirroring the one-shot damage replacement)

- **Types** (`types/ability.rs`): new `Effect::CreateDrawReplacement { replacement_effect: Box<Effect> }` (+ `EffectKind`). Additive variant — distinct Draw event class, heterogeneous payload; not parameterizable into `CreateDamageReplacement`.
- **Resolver** (`game/effects/create_draw_replacement.rs`, new): installs a `ReplacementDefinition` for the `Draw` event on the source object, carrying the substitute on `runtime_execute`, with `consume_on_apply` (one opportunity, CR 614.6) and an end-of-turn expiry ("this turn", CR 514.2). Source-player scope via the default `valid_player`; `valid_card` is intentionally unset because a `Draw` event has no affected object (a `valid_card` gate would make the shield never fire).
- **Pipeline** (`game/replacement.rs`): the non-draw-substitute pre-zero (`*count = 0`, CR 614.6) now also inspects `runtime_execute` when the branch carries no `execute` AST — so the draw is suppressed and the substitute runs via the existing `PostReplacementContinuation` drain. The damage/Jace/Abundance paths use `execute`, so the new fallback never engages for them.
- **Parser** (`parser/oracle_replacement.rs`): a draw arm on the existing one-shot "the next time …" site (the damage arm already bails on "would draw"). The substitute is parsed by the production effect parser, so the clause is only accepted when it lowers to a real effect; player-scoped payloads the bare parser would mis-scope ("each player …", "each opponent …") are rejected and left as honest gaps.

## Scope

Fixes **Words of Worship** (gain 5 life) and **Words of Wilding** (create a 2/2 Bear). **Words of Wind** ("each player returns a permanent") and **Words of Waste** ("each opponent discards") are deliberately left as honest gaps — their player-scoped payloads currently mis-parse to controller scope. Aladdin's Lamp / Mangara's Tome use bespoke library payloads and are out of scope.

## Tests

- `worship_next_draw_gains_life_no_card_drawn` — discriminating/fail-on-revert: the controller's next draw is suppressed (hand unchanged) **and** life +5; reverting the pipeline pre-zero fix makes the card get drawn too (double), failing the test.
- `wilding_creates_bear_instead`, `one_shot_consumed_second_draw_is_normal`, `expires_at_end_of_turn_if_unused`, `does_not_affect_opponent_draw`.
- AST: Worship → `CreateDrawReplacement{GainLife}`, Wilding → `CreateDrawReplacement{Token}`; `oneshot_draw_replacement_rejects_player_scoped_payload` pins Wind/Waste as honest gaps; the existing one-shot negative test updated.

## Verification

`cargo +nightly test -p engine`: lib 13137 passed (the 3 `db_load_path` failures are a stale local generated `card-data.json`, regenerated by CI; orthogonal). `--test integration`: 1461 passed; the 3 failures (2 × `anaphoric_scope_allowlist_guard`, `arc_trail_rejects_same_target`) are pre-existing on the base (verified by stash). `cargo check --workspace`, `clippy --workspace --all-targets --features engine/proptest -D warnings`, the parser-combinator gate, and rustfmt are clean. CR 614.1a / 614.6 / 514.2 / 121.1 reused from the existing draw/replacement subsystem.


matthewevans (changes requested):
**[HIGH]** The one-shot replacement disappears if the Words permanent leaves before the draw. Evidence: `crates/engine/src/game/effects/create_draw_replacement.rs:57` stores the delayed replacement directly on `ability.source_id`, while `find_applicable_replacements` only keeps ordinary replacement candidates from Battlefield/Command objects for a draw event (`crates/engine/src/game/replacement.rs:3976` and `crates/engine/src/game/replacement.rs:4009`). Once the already-resolved Words ability's source is destroyed or bounced, that object is no longer in the scanned zones, so the next draw is not replaced. Why it matters: a resolved activated ability must not lose its delayed "the next time" effect just because its source permanent left afterward (CR 113.7a / CR 611.2a). Suggested fix: store this one-shot draw replacement in a game-level pending registry, or otherwise make the scanner reach resolved delayed draw replacements independently of the source object's current zone, while retaining the bound source/controller snapshot.

**[MED]** Draw-to-draw payloads are accepted but execute in addition to the original draw. Evidence: `parse_oneshot_draw_replacement` accepts whatever `parse_effect(payload_text)` returns, including `Effect::Draw`, and wraps it in `CreateDrawReplacement` (`crates/engine/src/parser/oracle_replacement.rs:4294` and `crates/engine/src/parser/oracle_replacement.rs:4316`). In the runtime path, the `runtime_execute` branch explicitly does not pre-zero the original draw when the substitute is `Effect::Draw` (`crates/engine/src/game/replacement.rs:5063`), but the runtime continuation is still stashed and drained afterward (`crates/engine/src/game/replacement.rs:5012` and `crates/engine/src/game/replacement.rs:5083`). So "the next time you would draw a card this turn, draw two cards instead" draws the original card and then two more. Why it matters: this creates silently-wrong support for a payload shape the new parser now claims. Suggested fix: either reject `Effect::Draw` payloads in this one-shot parser until that class is implemented, or treat runtime Draw substitutes as true draw replacements and add a regression for the draw-two case.

matthewevans (changes requested):
VERDICT: changes-requested

Thanks for the follow-up. The runtime double-effect blocker looks materially addressed now: `replacement.rs` inspects `runtime_execute` for one-shot draw replacements and pre-zeros non-Draw substitutes, so Words of Worship/Wilding no longer draw the original card before running the substitute.

The remaining blocker is the delayed replacement lifetime/snapshot model. `create_draw_replacement.rs:52-61` still installs the one-shot shield directly on `ability.source_id`, and documents that it relies on the Battlefield/Command scan. `find_applicable_replacements` only scans `[Battlefield, Command]` by default (`replacement.rs:3976-4015`) and then uses the source object's live `replacement_source_player(obj)` as the default draw scope (`replacement.rs:3996`, `replacement.rs:4209-4225`). That means the already-resolved "the next time you would draw" effect is lost if the source permanent leaves before the draw, and can be scoped to the wrong player if the source changes controller before the draw.

Please move this to a delayed/turn-scoped replacement authority that is independent of the source object's current zone and snapshots the relevant player/controller when the ability resolves. Add regressions for at least: activate Words, source leaves before the next draw, replacement still fires for the original controller; and activate Words, source changes controller before the next draw, replacement still applies to the player who controlled the ability when it resolved.

matthewevans (changes requested):
VERDICT: request-changes

The previous blocker is still present on head `12ab77ca`. `create_draw_replacement::resolve` still installs the one-shot draw replacement by mutating the source object's `replacement_definitions` (`crates/engine/src/game/effects/create_draw_replacement.rs:52-61`), and the code comment still depends on the Battlefield/Command-zone scan finding that source (`:57-59`). The scanner does exactly that: `find_applicable_replacements` walks `active_replacements`, admits only Battlefield/Command plus a few event-specific exceptions (`crates/engine/src/game/replacement.rs:4018-4057`), and then the Draw player gate treats `valid_player: None` as the current `replacement_source_player(obj)` (`:4241-4267`). So if Words leaves the battlefield before the next draw, the replacement disappears; if its controller changes before the next draw, the replacement is scoped to the live controller instead of the player who activated it.

The added tests don't cover either failure mode. They assert the shield is installed on the source object (`create_draw_replacement.rs:123-129`) and only exercise source-present/source-still-controlled draws plus opponent draw isolation (`:281-306`). That locks in the bug rather than proving the resolved one-shot effect survives independently.

Please move this to resolved game-state pending replacement state (or an equivalent source-zone-independent representation) that captures the installing player/controller at resolution time, and add regressions for source leaving and source changing controller before the draw.

galuis116:
Addressed the remaining blocker on head `aff5170`:

- **Pending game-state lifetime:** `CreateDrawReplacement` now installs the one-shot shield into `pending_damage_replacements` with `source_controller` + `valid_player: You` anchored at resolution time, instead of mutating the source permanent's `replacement_definitions`.
- **Draw scan:** the pending-replacement loop now applies the Draw player-scope gate against the installing controller for floating Draw shields.
- **Regressions:** `survives_source_leaving_before_draw` and `stays_scoped_to_activating_player_after_control_change`; existing tests updated to assert pending-state install/consumption rather than object-hosted shields.

Rebased onto current `main` as well.

matthewevans (approved):
Reviewed current head. The prior source-lifetime blocker is addressed by pending-state draw replacement storage, source-controller scoping, and discriminating runtime coverage for source-leaves/control-change cases.

matthewevans (approved):
Reviewed current head d500d6dc. The only post-approval code change is the maintainer fixup classifying CreateDrawReplacement in ability_graph projection; exact-head CI is green and the prior one-shot draw replacement review remains applicable.

PR #4119, opened 2026-06-22
test(engine): combo-detection PR-2 — drive Priest+Umbral & Marwyn+Sword loop certificates

lgray:

## Where this sits in the staged combo-detection effort

This is **PR-2** of a staged, offline-first infinite-combo detector (Engine A). The staging is PR-0 … PR-8; the design docs are kept untracked in `.planning/combo-detection/` (`FEASIBILITY-AND-PLAN.md`, `IMPLEMENTATION.md`, `PROGRESS.md`). Predecessor: **PR-1** — [#4097](https://github.com/phase-rs/phase/pull/4097) (the analysis sim harness that feeds `ResourceVector`). PR-2 builds **only** on PR-1's analysis surface; it adds **zero** new gameplay behavior. Later PRs (PR-3+) extend the corpus driving and, eventually, a `combo-verify` CLI.

## What this PR does

PR-2 grows the acceptance corpus's set of **real, end-to-end driven** loop-certificate tests. Each driver builds the combo from the real card-data export, drives its exact loop cycle through the live `apply()` pipeline (via the existing `LoopProbe` / `run_combo` seam from PR-1), and asserts that `detect_loop` returns a genuine `LoopCertificate` whose `unbounded` axes are a **superset** of the row's expected resource family and whose `win_kind` matches exactly. Every driver is **revert-discriminating**: a paired `_requires_untap` probe omits the loop-closing action and asserts the loop fails to certify (`cert.is_none()`).

### Driven this PR (+2 → 10/49 real driven rows)

| # | Combo | Axis / WinKind | Per-cycle |
|---|-------|----------------|-----------|
| 10 | Priest of Titania + Umbral Mantle | `Mana(Green)` / `Advantage` | +2 green |
| 14 | Marwyn, the Nurturer + Sword of the Paruns | `Mana(Green)` / `Advantage` | +4 green |

Both were previously marked undrivable for reasons that turned out to be **measured-false** (corrected below). `DRIVEN_ROW_INDICES` is now `[0, 1, 4, 6, 9, 10, 12, 13, 14, 49]`.

### Shared, class-level harness additions (test-only)

These generalize a **class** of combos, not these two cards:

- `resolve_to_priority` now auto-answers `WaitingFor::OptionalEffectChoice` (accept) and `WaitingFor::ChooseOneOfBranch` (select the `SetTapState::Untap` branch) — covers every optional + modal-untap combo (Sword of the Paruns today; any future "tap or untap" modal). CR 608.2d.
- `build_board_green` floats the producer's color so a generic untap cost (`{3}`) is paid from the produced color rather than the colorless pool — without this, the colorless deficit trips the per-color net-progress gate (the same float trick the existing Selvala + Staff driver uses). CR 106.1 / 106.4.
- `seed_subtype_creatures` seeds battlefield permanents by subtype via `game::zones::create_object` (so Priest of Titania taps for enough Elves to clear `{3}`). CR 205.3 / 302.6.

### Honesty work (the remaining `#[ignore]` reasons)

Two bucket-doc reasons were **measured-false** on today's engine and were removed:

- **"CONTINUOUS-P/T untap engines"** (had claimed Priest + Umbral's +2/+2 untap buff makes the board unequal) — false: `project_out_resources` zeroes computed power/toughness to `None`, and `GameState::PartialEq` / `objects_content_eq` exclude `transient_continuous_effects` and `next_continuous_effect_id`, so the until-end-of-turn buff is invisible to board-equality. Combo 10 now drives.
- **"PARSE GAPS / Marwyn + Sword"** (had claimed Sword's untap parses to a bare `TargetOnly`) — false: it parses to `TargetOnly` + a `ChooseOneOf` modal whose Untap branch is a real `SetTapState::Untap`, driven via the new optional/modal arms. Combo 14 now drives.

Also corrected the stale "45 non-gated" → "49".

### Remaining `#[ignore]` / undriven rows, with precise measured reasons

- **Object-re-entry loops** (tokens / blink / persist / undying / recur / bounce-recast: Kiki-Jiki, Splinter Twin, Palinchron + Deadeye, Dockside, Food Chain, …): a permanent that leaves and returns gets a fresh `ObjectId` and bumps `next_object_id` every cycle, so the id-comparing, `battlefield`-order-comparing loop-equality sees a different board. Needs an object-identity-canonicalizing projection (a follow-up beyond PR-2's "board identical modulo monotone resources" model).
- **Extra-turn / extra-combat loops** (Time Sieve, Aggravated Assault, Combat Celebrant): each cycle advances `turn_number` / combat count and re-enters phases — the loop point is a different turn/phase, not board-identical.
- **Color-converting net-loss** (Pili-Pala + Grand Architect): a 2-restricted-`{C}` → 1-other-color conversion reads as a per-color deficit (`delta.mana[C] = −2`, no positive axis), which the per-color net-progress rule rightly rejects; restriction-aware mana is out of PR-2 scope.
- **Mandatory unbounded drain/draw cascades** (Sanguine Bond + Exquisite Blood; Marauding Blight-Priest + Bloodthirsty Conqueror; Niv-Mizzet, the Firemind + Curiosity): a single seed life-gain/draw starts a mandatory trigger cascade that the live engine deliberately halts via its runaway ceiling (`emit_resolution_halt`); there is no public single-step "resolve exactly one stack object" seam to step one pair (CR 732.4 mandatory loop, not a CR 732.2a optional shortcut).
- **Ability-copy churn** (Basalt Monolith + Rings of Brighthearth): the mana-positive line needs Rings' optional copy trigger, creating stack objects per cycle (out of the in-place-loop model).
- **Card-gated (4)**: Doc Aurlock / Professor Onyx / Animate Dead / Grindstone + Painter's Servant have `Unimplemented` parts.

The detector arms each remaining row would exercise are already covered by the 10 driven combos above and the `loop_check.rs` building-block unit tests (every `WinKind` arm + soundness negatives).

### Soundness set

The detector is held honest by negative controls that emit **no** certificate: `drive_board_change_is_not_a_loop`, `drive_idle_board_is_not_a_loop`, and the `loop_check.rs` `soundness_*` unit tests (board change / no progress / net-negative mana). Reverting either gate of `detect_loop` (board-equality-modulo-resources or net-progress) flips a driven assertion.

### Zero gameplay change

This PR is strictly offline / analysis-only. The only file changed is `crates/engine/src/analysis/corpus_tests.rs`. No reducer, SBA, resolution, layers, or parser edits. The new `resolve_to_priority` arms are test-harness auto-answers (they select an existing `GameAction` to dispatch), not game logic. No `Effect` / engine enum variant added, so `data/engine-inventory.json` is byte-identical. `mtgish` untouched.

### Gate

`cargo fmt --all` clean; `cargo test -p engine --lib analysis::` → **59 passed, 0 failed, 0 ignored** (includes both meta-tests `confirmed_drivers_match_expected` + `corpus_table_shape_is_locked`, all 10 drivers, the synthetic loops, and the soundness negatives); both new drivers produce real `detect_loop` certs and both revert-probes return `None`; `git diff data/engine-inventory.json` empty; diff scoped to `corpus_tests.rs` only.

### DIFF
--- a/crates/engine/src/analysis/corpus_tests.rs
+++ b/crates/engine/src/analysis/corpus_tests.rs
@@ -1,0 +1,1990 @@
+//! Corpus harness for the infinite-combo detector (Engine A).
+//!
+//! This is the acceptance suite described in `.planning/combo-detection/`
+//! `IMPLEMENTATION.md` §8: one data row per combo, plus a soundness set. Each row
+//! names the combo's documented unbounded resource axis and expected
+//! [`WinKind`]; the detector ([`detect_loop`]) is "done" for a row when, driven
+//! over that combo's loop, it confirms the loop and names ≥1 of the expected axes
+//! with the expected `win_kind` — and emits NO certificate on a non-loop board.
+//!
+//! # Layers, by what each asserts
+//!
+//! 1. **Driven end-to-end through the real pipeline** (`drive_*` /
+//!    `drive_combo_*` tests). Each is built from the *real* card-data export with
+//!    the cards' actual parsed abilities, its specific action cycle is driven
+//!    through `apply()` via [`LoopProbe`], and the cycle is confirmed by
+//!    [`detect_loop`] against the row's documented unbounded family + `WinKind`.
+//!    The set (see `DRIVEN_ROW_INDICES`): Heliod + Walking Ballista; Kilo,
+//!    Freed, Relic; Grim Monolith + Power Artifact; Devoted Druid + Vizier; Bloom
+//!    Tender + Freed; Priest of Titania + Umbral Mantle; Selvala + Staff of
+//!    Domination; Faeburrow + Pemmin's Aura; Marwyn + Sword of the Paruns; Spike
+//!    Feeder + Archangel. Two synthetic loops (`drive_damage_loop_certificate`
+//!    plus the negatives `drive_board_change_is_not_a_loop` /
+//!    `drive_idle_board_is_not_a_loop`) exercise the same pipeline without the
+//!    export. These are the discriminating regression tests — reverting either
+//!    gate of `detect_loop` (board-equality or net-progress) flips an assertion,
+//!    and each `drive_combo_*` is revert-probed (omit the loop-closing action ⇒
+//!    no certificate).
+//!
+//! 2. **Corpus card-availability over ALL 53 rows**
+//!    (`corpus_cards_present_and_implementation_status_matches_gating`). Loads
+//!    every card of every combo from the real export and asserts the §3
+//!    card-support prerequisite: all present, and every non-gated combo is fully
+//!    modeled (no top-level `Unimplemented`) — i.e. genuinely *available* to
+//!    drive. Runs against the real export when present (the maintainer's local
+//!    run); skips gracefully when the gitignored export is absent (CI / fresh
+//!    checkout) so it never fails spuriously.
+//!
+//! 3. **Corpus table** (`CORPUS`) + shape-lock meta-test. All 53 combos as data
+//!    rows. Each undriven row needs a bespoke board install + exact `GameAction`
+//!    sequence (auras, attachments, timing windows); the ones not yet driven are
+//!    documented in the "Remaining corpus rows" block below with a PRECISE,
+//!    measured reason they cannot be confirmed on today's engine model (object
+//!    re-entry under id-comparing loop equality, extra-turn/combat re-entry,
+//!    per-color net-progress, drain cascades, card gating) — not vague TODOs. A
+//!    follow-up (or PR-5's `combo-verify` CLI) extends the projection to reach the
+//!    object-re-entry class. The detector each would exercise is already covered
+//!    by layer 1 and the `loop_check.rs` building-block tests (every `WinKind`
+//!    arm + soundness negatives).
+
+use crate::analysis::resource::ResourceAxis;
+use crate::analysis::{detect_loop, LoopProbe, WinKind};
+use crate::database::CardDatabase;
+use crate::game::scenario::{GameScenario, P0, P1};
+use crate::types::ability::{
+    AbilityDefinition, AbilityKind, Effect, QuantityExpr, TargetFilter, TargetRef,
+};
+use crate::types::actions::GameAction;
+use crate::types::game_state::{CastPaymentMode, GameState, WaitingFor};
+use crate::types::identifiers::ObjectId;
+use crate::types::mana::ManaType;
+use crate::types::phase::Phase;
+use std::path::PathBuf;
+use std::sync::OnceLock;
+
+/// One row of the acceptance corpus: a combo, its documented unbounded resource
+/// family, the expected [`WinKind`], and (for the 4 card-gated combos) the card
+/// whose completion unblocks it.
+struct ComboRow {
+    /// Combo name (cards), for diagnostics.
+    name: &'static str,
+    /// The exact card names that make up this combo, as they appear in the
+    /// card-data export. The corpus test loads each of these from the real export
+    /// to confirm the combo is *available* (every card present + implemented).
+    cards: &'static [&'static str],
+    /// The unbounded-resource *family* the combo produces (the §12 "Category"
+    /// column). The detector must name ≥1 axis of this family.
+    family: ResourceFamily,
+    /// The expected `WinKind` once the loop is driven.
+    win_kind: WinKind,
+    /// `Some(card)` if this row is gated on an unimplemented card
+    /// (D3 / #19 / #36 / #49) — kept as a card-presence-only data row, not
+    /// driven; `None` otherwise.
+    gated_on: Option<&'static str>,
+}
+
+/// The §12 unbounded-resource families, mapped to the concrete [`ResourceAxis`]
+/// the detector reports. Keeps the corpus table declarative.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ResourceFamily {
+    Mana,
+    Tokens,
+    Damage,
+    Drain,
+    Mill,
+    Death,
+    Landfall,
+    Draw,
+    DrawDamage,
+    Combat,
+    Turns,
+    Counters,
+    Proliferate,
+    Engine,
+}
+
+/// The full 53-row acceptance corpus: 3 driving combos + the 50 card-disjoint
+/// corpus combos from `FEASIBILITY-AND-PLAN.md` §12. The 4 `gated_on`-nonempty
+/// rows correspond to the cards with Unimplemented parts (§3).
+const CORPUS: &[ComboRow] = &[
+    // ---- 3 driving combos ----
+    ComboRow {
+        name: "Heliod, Sun-Crowned + Walking Ballista",
+        cards: &["Heliod, Sun-Crowned", "Walking Ballista"],
+        family: ResourceFamily::Damage,
+        win_kind: WinKind::LethalDamage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Kilo, Apogee Mind + Freed from the Real + Relic of Legends",
+        cards: &[
+            "Kilo, Apogee Mind",
+            "Freed from the Real",
+            "Relic of Legends",
+        ],
+        family: ResourceFamily::Proliferate,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Doc Aurlock, Grizzled Genius + Aang, Swift Savior + Appa, Steadfast Guardian",
+        cards: &[
+            "Doc Aurlock, Grizzled Genius",
+            "Aang, Swift Savior",
+            "Appa, Steadfast Guardian",
+        ],
+        family: ResourceFamily::Tokens,
+        win_kind: WinKind::Advantage,
+        gated_on: Some("Doc Aurlock, Grizzled Genius"),
+    },
+    // ---- 50 corpus combos (§12) ----
+    ComboRow {
+        name: "Basalt Monolith + Rings of Brighthearth",
+        cards: &["Basalt Monolith", "Rings of Brighthearth"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Grim Monolith + Power Artifact",
+        cards: &["Grim Monolith", "Power Artifact"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Palinchron + Deadeye Navigator",
+        cards: &["Palinchron", "Deadeye Navigator"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Devoted Druid + Vizier of Remedies",
+        cards: &["Devoted Druid", "Vizier of Remedies"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Dramatic Reversal + Isochron Scepter",
+        cards: &["Dramatic Reversal", "Isochron Scepter"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Pili-Pala + Grand Architect",
+        cards: &["Pili-Pala", "Grand Architect"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Bloom Tender + Freed from the Real",
+        cards: &["Bloom Tender", "Freed from the Real"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Priest of Titania + Umbral Mantle",
+        cards: &["Priest of Titania", "Umbral Mantle"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Dockside Extortionist + Temur Sabertooth",
+        cards: &["Dockside Extortionist", "Temur Sabertooth"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Selvala, Heart of the Wilds + Staff of Domination",
+        cards: &["Selvala, Heart of the Wilds", "Staff of Domination"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Faeburrow Elder + Pemmin's Aura",
+        cards: &["Faeburrow Elder", "Pemmin's Aura"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Marwyn, the Nurturer + Sword of the Paruns",
+        cards: &["Marwyn, the Nurturer", "Sword of the Paruns"],
+        family: ResourceFamily::Mana,
+        win_kind: WinKind::Advantage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Heliod, Sun-Crowned + Walking Ballista [#13]",
+        cards: &["Heliod, Sun-Crowned", "Walking Ballista"],
+        family: ResourceFamily::Damage,
+        win_kind: WinKind::LethalDamage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Mikaeus, the Unhallowed + Triskelion",
+        cards: &["Mikaeus, the Unhallowed", "Triskelion"],
+        family: ResourceFamily::Damage,
+        win_kind: WinKind::LethalDamage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Sanguine Bond + Exquisite Blood",
+        cards: &["Sanguine Bond", "Exquisite Blood"],
+        family: ResourceFamily::Drain,
+        win_kind: WinKind::LethalDamage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Marauding Blight-Priest + Bloodthirsty Conqueror",
+        cards: &["Marauding Blight-Priest", "Bloodthirsty Conqueror"],
+        family: ResourceFamily::Drain,
+        win_kind: WinKind::LethalDamage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Niv-Mizzet, the Firemind + Curiosity",
+        cards: &["Niv-Mizzet, the Firemind", "Curiosity"],
+        family: ResourceFamily::DrawDamage,
+        win_kind: WinKind::LethalDamage,
+        gated_on: None,
+    },
+    ComboRow {
+        name: "Blasphemous Act + Repercussion",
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
VERDICT: request-changes

Existing bot feedback is only a Gemini quota warning; I found no prior actionable reviewer finding on this PR. This does not look like low-effort regression-test churn: the PR-2 corpus tests are tied to the staged combo-detection acceptance plan. The blocker is the stack boundary.

**[HIGH]** Do not present PR-2 as mergeable against `main` while it carries blocked PR-0/PR-1 commits. Evidence: `git log` on head f2eb0c8 includes `edf834cbf` (PR-0) and `c9c62c9f` (PR-1) before the PR-2 commits, and `git diff origin/main...HEAD --name-status` includes `crates/engine/src/analysis/loop_check.rs`, `mod.rs`, `resource.rs`, and `sim.rs`, not only `corpus_tests.rs`. Why it matters: #4092 and #4097 are still in Changes Requested, so merging this `test` PR would bypass the unresolved maintainer review of the analysis API it depends on. Suggested fix: keep this as draft/blocked until PR-0 and PR-1 are accepted and landed, then rebase PR-2 onto main so the diff contains only the PR-2 corpus-test changes, or retarget the stacked PR explicitly if the project decides to review stacks that way.

I applied the required `test` label. No approval/enqueue while predecessor PRs are blocked and this branch is still BEHIND.

lgray:

## Rebased onto the new PR-1 head + addressed review

Rebased `feat/combo-detect-pr2` onto the re-fixed predecessor **PR-1 #4097** (`origin/feat/combo-detect-pr1` head `065aca47e`, itself rebased onto current `upstream/main` with the merged **PR-0 #4092** folded in). New PR-2 head: **`3449ec14e`** (prev `f2eb0c8b3`).

### Stack-boundary finding (the one [HIGH]) — resolved
PR-0 is now merged into `main`, so it no longer appears in this branch. PR-2's own diff against its legitimate stacked base (`065aca47e`) is now exactly the four `crates/engine/src/analysis/` files (`corpus_tests.rs`, `loop_check.rs`, `mod.rs`, `resource.rs`) — no PR-0/PR-1 API bleed. The residual files the GitHub vs-`main` view shows (`sim.rs` + PR-1's slices of `mod.rs`/`resource.rs`) are PR-1's, since #4097 hasn't landed yet; this is the inherent stacked-PR base. The body already links predecessor PR-1 #4097.

### Also done
- **Squashed** the prior `wip`/`GRIND` commits into one production-quality commit.
- Two review nits fixed (comment-only): dropped a stale `next_object_id` clause from the object-re-entry deferral note (`normalize_for_loop` zeroes it on both sides, so the real discriminator is the fresh `ObjectId` id-keyed `objects_content_eq` miss + `battlefield` order), and removed an inaccurate `CR 605.3b` tag on a mana-choice test comment (605.3b is "mana abilities don't use the stack", unrelated).

### Verification (Tilt down → cargo foreground)
- Full engine lib suite: **13253 passed, 0 failed, 6 ignored**.
- Analysis suite with a schema-matching export: **70 passed, 0 failed** — all export-gated real-card drivers and both `_requires_untap` revert-probes **run** (not skipped) and pass; revert discrimination empirically confirmed in both directions.
- CI-equivalent clippy (`cargo clippy -p engine --all-targets --features engine/proptest -- -D warnings`): clean. `cargo fmt --all`: no drift. `data/engine-inventory.json`: unchanged.

PR-2 stays OFFLINE — zero gameplay change (no reducer/SBA/resolution edits; `detect_loop` / `project_out_resources` have no callers outside `analysis/`), and the corpus drivers stay honest (real `LoopCertificate`s, measured deferral reasons, no fabrication).

matthewevans (changes requested):
VERDICT: request-changes

The prior stack-boundary blocker is not addressed. PR-1 has now landed on `main` as squash commit `8a199028d`, but this branch still contains the original stacked commits `893c0b98a` and `065aca47e` plus the new PR-2 commit `3449ec14e` on an old merge base. Against current `origin/main` (`5db3a9e54`), the net PR diff is not just the PR-2 harness: it touches 74 files with 5,811 additions and 10,570 deletions, including deleting current-main files such as `crates/engine/src/game/stickers.rs`, `crates/engine/src/game/stickers_data.json`, `crates/engine/src/types/stickers.rs`, `crates/engine/tests/integration/issue_3871_summoners_pact.rs`, and `crates/engine/tests/integration/issue_3988_vaultborn_tyrant.rs`.

**[HIGH]** Rebase/cherry-pick PR-2 onto current `main` before review. Evidence: `git log --no-merges origin/main..HEAD` on head `3449ec14e` still shows `893c0b98a`/`065aca47e` predecessor commits, while `git diff --name-status origin/main..HEAD` includes unrelated current-main deletions like `D crates/engine/src/game/stickers.rs` and `D crates/engine/tests/integration/issue_3988_vaultborn_tyrant.rs`. Why it matters: merging this PR as-is would revert already-merged engine/parser/frontend work while presenting as a `test` PR, and it prevents a meaningful review of the PR-2 corpus harness. Suggested fix: create a fresh branch from current `main` and cherry-pick only the PR-2 commit (`3449ec14e`) or otherwise make the net diff contain only the intended analysis harness changes.

No approval/enqueue while the net diff is stale/destructive.

lgray:

## Rebased onto current `main` now that PR-1 has merged — diff is now only the analysis harness

The stack-boundary [HIGH] blocker is resolved. With **PR-0 #4092** (merged `8a199028d`'s predecessor) and **PR-1 #4097** now squash-merged into `main`, I replayed only PR-2's own commit onto current upstream/main:

```
git rebase --onto upstream/main 065aca47e feat/combo-detect-pr2
```

New PR-2 head: **`ca2a9adf8`** (prev `3449ec14e`).

### The diff is now exactly the PR-2 harness — no stale/destructive deletions
`git diff --stat upstream/main...HEAD` (the GitHub vs-base view) is now **only** the four `crates/engine/src/analysis/` files (`corpus_tests.rs`, `loop_check.rs`, `mod.rs`, `resource.rs`), **+2623/−1**. The earlier 74-file / +5,811 / −10,570 diff that deleted current-main files (`stickers.rs`, `issue_3988_vaultborn_tyrant.rs`, …) is gone — that was the old orphaned base, not PR-2's content. `mergeStateStatus` moved from `DIRTY` to `BEHIND` (normal merge-queue freshness, not a destructive state). `data/engine-inventory.json` is unchanged (analysis types are not `Effect` variants).

### Verification (Tilt not running → cargo foreground)
- **Full engine lib suite:** `13289 passed; 0 failed; 6 ignored`.
- **Analysis corpus suite (with a freshly regenerated, schema-matching `card-data.json`):** `18 passed; 0 failed; 0 skipped` — every export-gated real-card driver and both `_requires_untap` revert-probes **RUN** (not vacuously skipped). Runtime rises 0.20s → 5.33s once the drivers actually build boards and drive `apply()`, which is the proof they execute rather than early-return.
  - *Why a regen was needed:* the previously-attached export predated **#4102 (`a0a58753d`)**, which refactored `QuantityModification::Double` → `Times { factor }`; the stale export failed `from_export` (`unknown variant 'Double'`) so the drivers were silently skipping. Regenerating from this branch's code fixes the schema and lets them run.
- **Discrimination proof (non-vacuous):** temporarily adding the loop-closing untap to `drive_combo_14_marwyn_sword_requires_untap` makes it **FAIL** (`detect_loop` then returns `Some(cert)`), confirming the `cert.is_none()` assertion is load-bearing. Reverted.
- **OFFLINE invariant:** `detect_loop` / `LoopCertificate` / `loop_states_equal_modulo_resources` / `project_out_resources` have **no callers outside `analysis/`**; no reducer/SBA/resolution edits. Zero gameplay change.
- **CR gate:** all 65 distinct `CR` annotations across the 4 files grep-resolve in the Comprehensive Rules (incl. the hallucination-prone `701.21 = Sacrifice`). 0 not-found.
- `cargo clippy -p engine --all-targets -- -D warnings`: clean. `cargo fmt --all -- --check`: clean.

### Review nit addressed
Reworded two doc comments in `corpus_tests.rs` that referenced `#[ignore]` placeholders the module never uses — the undriven/gated corpus rows are honest card-presence-only data rows (via the `gated_on` field), not `#[ignore]`-gated tests.

PR-2 stays **OFFLINE** — the corpus drivers produce real `LoopCertificate`s and the undriven rows carry precise, measured deferral reasons, not fabricated coverage.

matthewevans (changes requested):
VERDICT: request-changes

The prior stale-stack blocker is addressed: this head's net diff against current `origin/main` is now the intended analysis surface (`corpus_tests.rs`, `loop_check.rs`, `mod.rs`, `resource.rs`) rather than the old destructive stack diff. I re-reviewed the refreshed head and found two detector correctness blockers.

**[HIGH]** Do not project away activation-limit state when proving a cycle can repeat. Evidence: `crates/engine/src/analysis/resource.rs:632` clears `activated_abilities_this_turn`, `crates/engine/src/analysis/resource.rs:633` clears `activated_abilities_this_game`, and `crates/engine/src/analysis/resource.rs:635` clears `loyalty_abilities_activated_this_turn` inside `project_out_resources`. Why it matters: these fields are not just append-only history; they enforce activation restrictions like CR 602.5b “Activate only once each turn” and CR 606.3 loyalty once-per-turn, so a one-shot or once-per-turn ability can produce a positive first-cycle delta, be illegal to repeat from `cycle_end`, and still compare as the same loop state. Suggested fix: keep action-limiting tallies in the repeatability comparison, or model them as consumed/non-renewable resources, and add a negative where a once-per-turn/loyalty activation produces progress once but must not certify as infinite.

**[HIGH]** Do not infer the loop controller set from every player with a battlefield permanent. Evidence: `detect_loop` derives `controllers` from `surviving_controllers(cycle_start)` at `crates/engine/src/analysis/loop_check.rs:150`, and `surviving_controllers` inserts the controller of every battlefield object at `crates/engine/src/analysis/loop_check.rs:294`; the drain/mill classification then treats negative life/library as an opponent axis only when that player is *not* in that set (`crates/engine/src/analysis/loop_check.rs:217` and `crates/engine/src/analysis/loop_check.rs:268`). Why it matters: in a normal game where the opponent controls any permanent, a pure drain or mill loop against that opponent is either rejected by `is_progress` or downgraded away from `LethalDamage`/`Decking`, even though CR 704.5a and CR 104.3c/121.4 are exactly the intended win classifications. Suggested fix: pass the loop actor/controller from the driver/probe into `detect_loop` (or otherwise bind it to the action/resource source), and add drain/mill regressions where the opponent also controls a permanent.

No approval/enqueue while the detector can certify non-repeatable once-per-turn loops and miss ordinary opponent-controlled boards.

lgray:

## Both [HIGH] detector-correctness blockers fixed; rebased onto current `main`

New PR-2 head: **`d84e824bd`** (prev `ca2a9adf8`). Rebased directly onto current `upstream/main` (`a66b79e6e`, v0.3.0) — `git diff --stat upstream/main HEAD` is **only** the four `crates/engine/src/analysis/` files (`corpus_tests.rs`, `loop_check.rs`, `mod.rs`, `resource.rs`), no unrelated deletions, `data/engine-inventory.json` unchanged. Still **OFFLINE** — `detect_loop` / `project_out_resources` / `loop_states_equal_modulo_resources` have zero callers outside `analysis/`; the strict CR 104.4b live-draw comparator (`loop_states_equal` / `objects_content_eq` / `normalize_for_loop`) is untouched. Zero gameplay change.

### [HIGH] #1 — don't project away activation-limit state
`project_out_resources` no longer blanket-clears the gate tallies. The clear-block is now classified per field into **repetition gate (preserved)** vs **pumped history (cleared)**:

- **`activated_abilities_this_turn` / `_this_game`** → `.retain` only keys whose ability carries the matching `ActivationRestriction` — per-turn (`OnlyOnceEachTurn` / `MaxTimesEachTurn`) or per-game (`OnlyOnce`) (**CR 602.5b**). These are bumped *unconditionally* for every activation, so a blanket **keep** would break legitimate unrestricted loops and a blanket **clear** was the false-positive bug — a selective retain is the only correct partition.
- **`triggers_fired_this_turn` / `trigger_fire_counts_this_turn`** → no longer cleared (preserved). Single-writer proof: `record_trigger_fired` writes them **only** for `TriggerConstraint::OncePerTurn` / `MaxTimesPerTurn` and early-returns for an unconstrained trigger, so an unrestricted-trigger loop never touches them (preserving can't break a legit loop) while a once-per-turn trigger makes two cycles compare different (**CR 603.2h** — "Do this only once each turn", grep-verified; note the engine's pre-existing `603.4` annotation on `MaxTimesPerTurn` in `triggers.rs` is the intervening-if rule and is left untouched as out-of-scope).
- **Loyalty (CR 606.3)** — additional discovery: the authoritative once-per-turn-per-permanent gate is the **per-object** `obj.loyalty_activations_this_turn`, which `objects_content_eq` does **not** compare and `normalize_for_loop` does **not** zero, so a loyalty loop was a false positive *independent* of the GameState maps. Fixed with an analysis-local conjunct (`loyalty_activation_counts_match`) in `loop_states_equal_modulo_resources` — **without** widening the shared strict comparator.
- **`ability_resolutions_this_turn`** stays cleared — `NthResolutionThisTurn{n}` fires on exact `count == n` (a one-shot branch *selector*, not a repetition-blocking gate); a board-divergent branch is already caught by `objects_content_eq` and a resource-only Nth bonus is pre-warmup (**CR 603.4**, the intervening-if class).

### [HIGH] #2 — don't infer the loop controller from every battlefield permanent
`detect_loop` now takes an explicit **`controller: PlayerId`**, threaded from the driver's active player (mirroring how `mandatory` is already supplied because the detector can't infer it from two states). `is_progress` / `unbounded_axes_for` / `classify_win_kind` scope the consumed-axis constraint and the opponent-win axes to that single controller; `surviving_controllers` is deleted. The corpus `run_combo` reads `probe.runner().state().active_player`.

### Discriminating regressions added (all revert-proven load-bearing)
- once-per-turn / once-per-game **activated**-ability gate: gated activation → modulo-**unequal** (no certificate), paired positive control (unrestricted ability) → modulo-**equal** (still certifies). Reverting to the blanket `.clear()` flips both.
- once-per-turn / N-times-per-turn **trigger** gate: same shape, control = unrestricted trigger writes neither map.
- **loyalty** CR 606.3 gate: count 0 vs 1 → unequal; removing the `&& loyalty_activation_counts_match` conjunct makes `objects_content_eq` wrongly compare equal.
- gate-predicate partition unit (true for the gate variants, false for `OnlyOnce`/`AsSorcery`).
- **Finding-2 drain & mill where the opponent ALSO controls a permanent**: `controller = P0`, opponent on the battlefield → `LethalDamage` / `Decking`; the revert probe asserts `classify_win_kind(victim_as_controller)` → `Advantage` vs `classify_win_kind(real_controller)` → `LethalDamage`/`Decking`, proving the parameter is load-bearing (the old board-presence proxy would have downgraded these).

### Verification (Tilt not running → cargo foreground, against the rebased v0.3.0 base)
- `cargo test -p engine --lib analysis::` → **78 passed; 0 failed; 0 ignored** (all driven real-card corpus drivers run, not skipped).
- `cargo clippy -p engine --lib -- -D warnings` → clean. `cargo fmt --all -- --check` → clean.
- Every `CR` annotation in the diff grep-resolves in the Comprehensive Rules (602.5b, 603.2h, 603.4, 606.3, 704.5a, 104.3c, 121.4).
- OFFLINE invariant + unchanged `engine-inventory.json` re-confirmed.

matthewevans (changes requested):
VERDICT: request-changes

The two prior detector-correctness blockers are addressed on this head: the repeatability comparison now preserves action-limiting activation/trigger/loyalty state where it matters, and `detect_loop` now receives an explicit controller instead of inferring every battlefield owner as a loop controller. The refreshed head is also green and the net surface remains the four `analysis/` files. One controller-scope blocker remains in the new damage classification.

**[HIGH]** Damage dealt to the loop controller is still classified as an opponent-lethal win. Evidence: `crates/engine/src/analysis/loop_check.rs:240` documents opponent/loss classification, but `crates/engine/src/analysis/loop_check.rs:241` checks `delta.damage_dealt.values().any(|&n| n > 0)` without comparing the damaged `PlayerId` to `controller`. The axis is keyed by the damaged player: `crates/engine/src/analysis/resource.rs:163` says `damage_dealt` is cumulative damage dealt to each player, `ResourceVector::unbounded_components` emits `ResourceAxis::DamageDealt(pid)` for any positive pid at `crates/engine/src/analysis/resource.rs:429`, and the event accumulator writes the event target player into that key at `crates/engine/src/analysis/sim.rs:53`. Why it matters: a loop that repeatedly damages its own controller but offsets the life loss, e.g. self-damage plus enough lifegain so `is_progress` does not reject controller life under `crates/engine/src/analysis/loop_check.rs:194`, can be certified as `WinKind::LethalDamage` even though no opponent is losing under CR 704.5a. That is the same opponent-vs-controller boundary this patch fixed for life loss and decking, but the damage branch still bypasses it. Suggested fix: require `pid != controller` for `damage_dealt` to classify as `LethalDamage`, and add a regression where `damage_dealt[P0] > 0` with no opponent damage/life/library loss produces an `Advantage` certificate (or no direct win), while the existing `damage_dealt[P1]` case remains lethal.

No approval/enqueue while the detector can issue a direct win certificate for controller-only damage.

lgray:

## [HIGH] controller-only damage misclassification — fixed

New PR-2 head: **`13a2c2ea6`** (prev `d84e824bd`). Rebased directly onto current `upstream/main` first — `git diff --stat upstream/main HEAD` is **only** the four `crates/engine/src/analysis/` files (`corpus_tests.rs`, `loop_check.rs`, `mod.rs`, `resource.rs`), no unrelated deletions, `data/engine-inventory.json` unchanged. Still **OFFLINE** (zero gameplay change): `detect_loop` / `classify_win_kind` have no callers outside `analysis/`.

### The finding — confirmed and fixed
`classify_win_kind`'s damage branch classified `WinKind::LethalDamage` via `delta.damage_dealt.values().any(|&n| n > 0)` — controller-blind — while the sibling life-loss and decking branches already require `*pid != controller`. A self-ping loop whose controller's life is offset by lifegain (so `is_progress` doesn't reject controller life at `loop_check.rs:196`) was certified as a direct `LethalDamage` win even though no opponent loses.

**CR 704.5a** (verified, `docs/MagicCompRules.txt:5464` — "If a player has 0 or less life, that player loses the game"): lethal damage is a win only against an **opponent**. The fix brings the damage branch into parity with the life/library branches:

```rust
// CR 704.5a: a player at 0 life loses — so unbounded damage is a WIN only when
// the damaged player is an OPPONENT (a non-controller). ...
if delta
    .damage_dealt
    .iter()
    .any(|(pid, &n)| n > 0 && *pid != controller)
{
    return WinKind::LethalDamage;
}
```

Controller-only damage now falls through to `WinKind::Advantage` — a beneficial CR 732.2a loop that still names its `DamageDealt(controller)` axis (well-formed `Some` certificate, not `None`/panic, since `unbounded_components` surfaces the axis). This mirrors self-mill (`Advantage`, not `Decking`) and self-life-loss (`Advantage`, not `LethalDamage`). Mixed deltas (damage to both controller and opponent) still classify `LethalDamage` — the opponent still dies.

### Discriminating regression
`classify_win_kind_controller_only_damage_is_not_lethal`:
- `damage_dealt[P0]` (controller) ⇒ `Advantage` — **fails** against the pre-fix controller-blind predicate (`left: LethalDamage, right: Advantage`), **passes** against the fix.
- `damage_dealt[P1]` (opponent) ⇒ `LethalDamage` — unchanged by the fix, proving the change is surgical.
- well-formedness: `detect_loop(...)` returns `Some` with `win_kind == Advantage` and `covers([DamageDealt(P0)])`.

Discrimination reproduced (revert-the-predicate → test fails → restore → passes). `cargo test -p engine --lib analysis::loop_check`: 15 passed, 0 failed. Prior rounds' fixes preserved (repetition-gate-vs-pumped-history projection in `project_out_resources`; explicit `controller` threaded into `detect_loop`); the strict CR 104.4b live-draw comparator is untouched.

matthewevans (approved):
Approved current head 13a2c2ea6065c68af135e873b31667850e0a694b after re-review. The prior controller-scope blocker is addressed: damage win classification now requires a damaged non-controller player, and the regression covers controller-only damage as non-lethal while preserving opponent damage as LethalDamage. CI is green on the exact head.

PR #4120, opened 2026-06-22
Fix Full Throttle additional combat phases (fixes #828)

### LINKED ISSUE #828: Full Throttle - sorcery does nothing — Full Throttle - sorcery does nothing when cast.
Reported in Discord: https://discord.com/channels/1485498006781427802/1507419428013215774/1507419428013215774

**Thread:** Full Throttle - sorcery does nothing
**Cards:** full throttle

## Summary
Full Throttle - sorcery does nothing when cast.

### full throttle
After this main phase, there are two additional combat phases.
At the beginning of each combat this turn, untap all creatures that attacked this turn.

---
<sub>report_id: `discord:1507419428013215774:1507419428013215774:0` · discord: `1507419428013215774/1507419428013215774`</sub>

kiannidev:
## Summary
- Parse "After this main phase, there are N additional combat phases" with a main-phase anchor instead of incorrectly anchoring to `EndCombat`.
- Remap the parser sentinel at resolution so casts during precombat or postcombat main schedule extra combats after the correct main phase.
- Add parser and integration tests for Full Throttle.

## Root cause
Full Throttle was parsed like combat-phase-granting effects ("after this phase"), anchoring extra combats to `EndCombat`. When cast during a main phase—especially postcombat main—that anchor never occurs again, so the spell appeared to do nothing.

## Test plan
- [x] `cargo test -p engine --lib full_throttle_parses additional_phase_after_this_main parse_additional_phase_after_this_main`
- [x] `cargo test -p engine --test integration full_throttle`

Fixes #828


matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** The counted Full Throttle combat phases are still not actually reachable after the first inserted combat. Evidence: `crates/engine/src/game/effects/additional_phase.rs:84-95` pushes both `count` copies with the same `anchor`, and `crates/engine/src/game/turns.rs:47-59` consumes only entries whose anchor matches the current phase; after the first `PreCombatMain`-anchored `BeginCombat` is consumed, the game proceeds through combat steps and never returns to `PreCombatMain`, so the second queued combat is stranded. The new integration tests at `crates/engine/tests/integration/issue_828_full_throttle.rs:21-36` and `:49-60` only assert the queued entries immediately after resolution, not that advancing the turn starts two additional combats. Why it matters: Full Throttle advertises two additional combat phases, but this implementation still produces at most one reachable extra combat and can skip the rest of the inserted sequence. Suggested fix: model repeated inserted combat phases as a reachable sequence, then add a production-path regression that advances the turn and observes two extra `BeginCombat` entries before the turn continues.

Also, the current head is red on Rust lint because `cargo fmt --all --check` rewrites the import order and long assertions in the new tests. That is secondary to the scheduling blocker above, but it needs to be fixed before this can be reconsidered.

matthewevans:
VERDICT: still-blocked

Verified the new head `2464389656a054b4b192d40ab366d0d4334bd1c2` against the prior reviewed head via the GitHub API compare. The only new commit is formatting-only.

That addresses the fmt/lint portion, but the semantic blocker from my prior review is unchanged: both counted Full Throttle combat entries are still anchored to the same current main phase, and the tests still do not advance through two combat phases to prove the second one is consumed correctly. No approval/enqueue.

matthewevans:
VERDICT: approve-pending-ci

Re-reviewed current head `e3b5d55ffa8f5d33f4b844ecb8ba5d72c20308ce` via the GitHub API diff and isolated worktree.

The prior semantic blocker is addressed: counted Full Throttle combat phases no longer all anchor to the same main phase. The resolver keeps the first `After this main phase` combat anchored to the active main phase, then chains later counted combat entries through `EndCombat`, which matches the `advance_phase` anchor-consumption model. The new integration coverage also advances through two reachable combat entries rather than only asserting queued state.

No formal approval/enqueue yet because the branch is still BEHIND and fresh GitHub CI is running. Required `bug` label is present.

matthewevans (approved):
Approved on current head 82c35af86636ffdd7fb25570a060c02064e16185 after repush guard: the substantive Full Throttle follow-up addresses the prior counted-combat reachability blocker by chaining later counted additional combat phases through EndCombat, includes a production-style two-combat advancement regression, required bug label is present, and GitHub CI is green/CLEAN.

PR #4123, opened 2026-06-22
fix(parser): back-search defs chain for "destroyed this way can't be regenerated" continuation

### LINKED ISSUE #3343: Parser: remaining “can’t be regenerated” riders (conditional “if it’s a creature”; “destroyed this way”) dropped
## Summary
Follow-on to #3339 (which routed the standalone "a creature dealt damage this way can't be regenerated" rider). The remaining "can't be regenerated" rider forms are still dropped (each card `gap_count == 1` on the regen clause):

### Conditional damage-form ("if it's a creature, it ...")
- **Disintegrate**: "Disintegrate deals X damage to any target. If it's a creature, it can't be regenerated this turn, and if it would die this turn, exile it instead."
- **Carbonize**: same form ("...If it's a creature, it can't be regenerated this turn, and if it would die this turn, exile it instead.")
The exile-instead rider already parses; only the regen clause is dropped. The leading "if it's a creature, " conditional wrapper + the ", and if it would die..." boundary (the regen clause must terminate before the exile rider) are unhandled.

### Destroy-form ("creatures/a creature destroyed this way")
- **Breaking Point**: "...destroy all creatures. Creatures destroyed this way can't be regenerated." (the DestroyAll is in the group-bargain's "if no one does" branch)
- **Kirtar's Wrath**: threshold branch "...destroy all creatures, then create two tokens. Creatures destroyed this way can't be regenerated." (a Token effect sits between the DestroyAll and the regen clause)
- **War Barge**: "...When this artifact leaves the battlefield this turn, destroy that creature. A creature destroyed this way can't be regenerated." (destroy inside a delayed leaves-trigger)

## Root cause
The runtime is complete (the `cant_regenerate` flag on Destroy/DestroyAll and `StaticMode::CantBeRegenerated` both work; Wrath of God + the #3339 damage cards are supported). The gaps are parser-side:
- The conditional damage-form needs the "if it's a creature, it" wrapper recognized and the regen clause split from the trailing ", and if it would die..." exile rider (which parses separately).
- The destroy-form continuation linker only sets `cant_regenerate` when the regen sentence is ADJACENT to the Destroy/DestroyAll with a generic subject ("they"); it does not recognize the noun-phrase subjects ("creatures/a creature destroyed this way") nor reach the Destroy across an intervening sibling (Token), into an else-branch, or into a delayed-trigger body.

## Expected
Route these remaining forms to the existing regen-prevention infra: the conditional damage-form to `StaticMode::CantBeRegenerated` bound to the damage target (ParentTarget), and the destroy-form to the `cant_regenerate` flag on the relevant Destroy/DestroyAll (recognizing the "destroyed this way"/"a creature destroyed this way" subjects and locating the destroy across the intervening structure).

CR 701.19a (regeneration shield), CR 701.19c (effects that say a permanent can't be regenerated cause regeneration shields to not be applied), CR 614.8 (regeneration is a destruction-replacement effect), CR 608.2c (later text modifies earlier).

dripsmvcp:
Fixes #3343

## Summary
- **Detection**: `parse_followup_continuation_ast` now recognises "destroyed this way can't be regenerated" regardless of the immediately preceding effect type, via a catch-all arm using `scan_contains` (CR 608.2c — the rule explicitly covers this exact text pattern).
- **Application**: `apply_clause_continuation` for `CantRegenerate` walks backward through the definition chain to find the nearest `Destroy`/`DestroyAll`, instead of only checking the last definition (CR 701.19c). This handles cards where intervening effects (token creation, conditional wrapping) separate the destruction from the regen clause — e.g. Kirtar's Wrath threshold, Breaking Point, War Barge.

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine --lib -- cant_regenerate` — 8/8 passed (including 2 new tests)
  - `cant_regenerate_destroyed_this_way_with_intervening_effect` — DestroyAll + intervening Token + "destroyed this way" continuation
  - `cant_regenerate_destroyed_this_way_simple` — Destroy + full noun-phrase "A creature destroyed this way"
  - All 6 existing cant_regenerate tests still pass


matthewevans (changes requested):
Thanks for taking this one on; the direction is right, but the token-intervening case is still not actually covered by the implementation as written.

The new fallback in `parse_followup_continuation_ast` is placed after this unconditional arm:

```rust
Effect::Token { .. } => try_parse_token_enters_with_counters(&lower)
    .or_else(|| try_parse_put_counters_on_token_followup(&lower)),
```

Because that arm matches every `Effect::Token`, a continuation following token creation returns the token-counter parser result and never reaches the later `_ if scan_contains("destroyed this way can't be regenerated")` fallback. That is the exact Kirtar's Wrath shape described in the comment: DestroyAll, then Token, then `Creatures destroyed this way can't be regenerated.`

The added Kirtar regression does not catch this because it only asserts the root `Effect::DestroyAll { cant_regenerate: true }`; it does not inspect the conditional/instead branch after the intervening token creation, nor does it directly exercise `parse_followup_continuation_ast` with an `Effect::Token` previous effect.

Please move the destroyed-this-way continuation detection ahead of the unconditional Token arm (or fold it into that arm before returning), and add a regression that proves the token-intervening branch itself is patched, not only the top-level adjacent DestroyAll case.

matthewevans:
VERDICT: approve-pending-ci

Re-reviewed the new head after the follow-up and maintainer test-only update.

The prior blocker is addressed: the `destroyed this way can't/cannot be regenerated` continuation is now checked before the unconditional `Effect::Token` continuation arm, and `apply_clause_continuation` walks backward to patch the nearest `Destroy`/`DestroyAll`, which covers the token-intervening Kirtar-style shape. The old failing CI assertion was a deliberate coverage sentinel for Breaking Point staying PARTIAL only while the regen rider was unsupported; I pushed `7aefce069` to move Breaking Point into the fully-supported expectation.

No formal approval/enqueue yet: GitHub CI is still running on the fresh head and the branch is currently BEHIND, so this remains below the authorized-mode approval bar until current checks are green and the repush guard passes.

matthewevans (approved):
Approved after repush guard on the reviewed head: required bug label present, GitHub checks green, and mergeStateStatus CLEAN.

PR #4125, opened 2026-06-22
fix(parser): recognize countered_spell_zone as condition marker in swallow checker

dripsmvcp:
## Summary
- Add `"countered_spell_zone":` to the `cond_markers` list in `detect_condition_if` (swallow checker) so that `Counter` effects with a redirect destination are recognized as encoding the "if that spell is countered this way" conditional gate — matching the existing `source_rider` pattern for the same effect variant.
- Promotes 4 fully-parsed cards from `supported: false` to `supported: true`: Memory Lapse, Lapse of Certainty, Remand, Spell Crumple.

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine -- swallow_check::tests::condition_if_accepts_countered_spell_zone_redirect` — passed
- `cargo coverage` before/after: all 4 cards move from unsupported to supported with zero remaining gaps

### DIFF
--- a/crates/engine/src/parser/swallow_check.rs
+++ b/crates/engine/src/parser/swallow_check.rs
@@ -2033,6 +2033,12 @@ fn detect_condition_if(
         // skip_serializing_if = is_none) IS the conditional gate (Teferi's
         // Response, Green Slime, Tishana's Tidebinder).
         "\"source_rider\":",
+        // CR 701.6a: Effect::Counter.countered_spell_zone encodes the
+        // "if that spell is countered this way, put it [on top of / on
+        // the bottom of its owner's library | into its owner's hand]"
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
VERDICT: approve

Reviewed head `dcda2a565542ec541fa295eac37bd70a55674be0` via GitHub API diff and isolated worktree. This is a narrow support-classification bugfix: `countered_spell_zone` is already the typed structural encoding for Memory Lapse/Remand/Spell Crumple-style counter redirects, and existing parser/runtime tests cover the redirect semantics. The new swallow-check marker prevents a false `Condition_If` diagnostic without relaxing unsupported destinations.

Repush guard passed, required `bug` label is present, and GitHub CI is green/CLEAN.

PR #4127, opened 2026-06-22
fix(parser): parse negated cast-from-zone intervening-if conditions

### LINKED ISSUE #3682: Chainer, Nightmare Adept grants haste to itself instead of reanimated creatures
## Discord report
Chainer, Nightmare Adept grants haste to itself rather than the other cards it should apply to.

Source: https://discord.com/channels/1485498006781427802/1516200788723105944/1516200788723105944
Report ID: discord:1516200788723105944:1516200788723105944:0

dripsmvcp:
Fixes #3682

## Summary
- Add nom combinators `parse_negated_cast_from_zone_intervening_if` and `parse_cast_from_zone_intervening_if` for trigger-level "if you (didn't) cast it from your hand/graveyard/exile" intervening-if patterns — covers the entire class of negated and positive zone-specific cast-provenance conditions
- Extend `strip_cast_from_zone_conditional` with the negated form for effect-level paths (Epochrasite-style trailing conditions)
- Fixes Chainer, Nightmare Adept's haste grant and Phage the Untouchable's lose-the-game trigger, which were misinterpreted as `Not(EffectOutcome(OptionalEffectPerformed))` instead of `Not(WasCast { zone: Hand })`

## Verification
- `cargo fmt --all` — clean
- `parser::oracle_trigger::tests` — 788 passed, 0 failed
- `parser::oracle_effect` — 2104 passed, 0 failed
- `snapshot` — 312 passed, 0 failed

### DIFF
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -16048,6 +16048,94 @@ pub mod tests {
         }
     }
 
+    /// CR 601.2 + CR 404.1 + CR 603.4: "if you cast it from your graveyard"
+    /// ("you cast it" + owner-specific zone) scopes BOTH axes — caster=you AND
+    /// origin-zone owner=you. Drives the parser-emitted condition through
+    /// `check_trigger_condition` across all four owner × caster rows relative to
+    /// the trigger controller (P0): only the row where you cast your own card
+    /// from your graveyard satisfies it. Before the scope fix the parser emitted
+    /// `controller: None, owner: None`, which wrongly passed every row.
+    #[test]
+    fn you_cast_from_your_graveyard_scopes_both_caster_and_owner() {
+        let trigger = crate::parser::oracle_trigger::parse_trigger_line(
+            "When this creature enters, if you cast it from your graveyard, put a +1/+1 counter on it.",
+            "Gravecaster",
+        );
+        let condition = trigger
+            .condition
+            .expect("scoped gravecast intervening-if must parse");
+        assert_eq!(
+            condition,
+            TriggerCondition::WasCast {
+                zone: Some(Zone::Graveyard),
+                controller: Some(ControllerRef::You),
+                owner: Some(ControllerRef::You),
+            },
+            "'you cast it from your graveyard' must scope both caster and owner to you"
+        );
+
+        // (owner, caster, expected) relative to trigger controller P0.
+        let cases = [
+            (
+                PlayerId(0),
+                PlayerId(0),
+                true,
+                "you cast your card from your graveyard",
+            ),
+            (
+                PlayerId(0),
+                PlayerId(1),
+                false,
+                "opponent cast your card from your graveyard",
+            ),
+            (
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** Preserve the printed caster/owner scope when lowering these cast-from-zone clauses. Evidence: `crates/engine/src/parser/oracle_trigger.rs:3135-3160` emits `WasCast { zone: Some(zone), controller: None, owner: None }` for both `if you didn't cast it from your ...` and `if you cast it from your ...`, while the existing `WasCast` contract says `controller` is the caster axis and `owner` is the origin-zone owner axis (`crates/engine/src/types/ability.rs:14134-14148`), and the existing graveyard parser explicitly scopes `you cast it from your graveyard` to both axes (`crates/engine/src/parser/oracle_trigger.rs:3020-3041`). Why it matters: this makes the newly accepted wording overbroad; `if you cast it from your hand/graveyard` can pass for a spell cast by someone else or from someone else's owner-specific zone, and the negated Chainer-style form can fail just because the object was cast from any player's hand rather than your hand. The new parser tests assert the wrong unscoped shape at `crates/engine/src/parser/oracle_trigger.rs:16360-16389`, so they protect the bug. Suggested fix: lower `if you cast it from your hand/graveyard` to `WasCast { zone: Some(...), controller: Some(You), owner: Some(You) }` where the zone is owner-specific, wrap that scoped condition for the negated form, and add a runtime condition test with mismatched caster/owner rows.

The Gemini apostrophe concern is not a blocker from the checked-in dataset I inspected: Chainer's source text uses the ASCII `didn't` form. The scope erasure above is the merge blocker.

matthewevans (changes requested):
VERDICT: request-changes

**[LOW]** Update the stale Chainer parser test to the new scoped shape before this can land. Evidence: current head now emits the prior review's requested `WasCast { zone: Some(Zone::Hand), controller: Some(ControllerRef::You), owner: Some(ControllerRef::You) }`, but `trigger_intervening_if_negated_cast_from_hand_chainer` still matches `controller: None, owner: None` at `crates/engine/src/parser/oracle_trigger.rs:14125-14133`. GitHub's current Rust test run fails exactly there: `expected WasCast { zone: Hand }, got WasCast { zone: Some(Hand), controller: Some(You), owner: Some(You) }`. Why it matters: the behavior blocker looks addressed, but this head is still red and leaves an older test asserting the wrong unscoped contract. Suggested fix: update that Chainer test to expect the scoped `You`/`You` axes, matching the new extraction/runtime coverage added later in the file.

matthewevans (approved):
VERDICT: approve

Re-reviewed head `22fbb2c71d0abddbfd13e5e55beb4ba2ae4d4b21`. The prior scoped `WasCast` behavior fix still stands, and this follow-up only updates the stale Chainer parser assertion to the scoped controller/owner shape. GitHub checks are green and merge state is CLEAN; approving exact head after repush guard.

PR #4138, opened 2026-06-22
fix(parser): extend damage-recipient qualifier with battle disjunction

dripsmvcp:
## Summary
- Extend `parse_damage_to_qualifier_with_rest` with three new `alt()` arms for the "or battle" damage-recipient disjunction introduced by March of the Machine battles (CR 310.1):
  - `"a player or battle"` → `Or { Player, Battle }`
  - `"an opponent or battle"` → `Or { Opponent, Battle }`
  - `"a player, planeswalker, or battle"` → `Or { Player, Planeswalker, Battle }`
- Extend `try_parse_one_or_more_combat_damage_to_player` to match the "or battle" suffix on batched triggers

Previously, "or battle" was silently dropped — triggers fired only on player/opponent damage, never on battle damage. Affects 12+ cards including Archpriest of Shadows, Beamtown Beatstick, Bloodfeather Phoenix, Deeproot Wayfinder, Doomskar Warrior, Farseeing Flockmate, Furnace Reins, Gitaxian Mindstinger, Rankle and Torbran, Rooftop Saboteurs, and Zurgo and Ojutai.

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine --lib -- damage_to_qualifier` — 4 unit tests passed
- `cargo test -p engine --lib -- or_battle` — 4 card regression tests passed (Archpriest, Bloodfeather Phoenix, Farseeing Flockmate, Zurgo and Ojutai)

### DIFF
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -7218,6 +7218,103 @@ mod tests {
         assert!(!match_damage_done(&to_player, &trigger, source, &state));
     }
 
+    /// CR 120.3 + CR 102.2: A "deals damage to a player or battle" trigger uses
+    /// `TargetFilter::Or { filters: [Player, Typed(Battle)] }` as emitted by
+    /// `parse_damage_to_qualifier`. The `TargetRef::Object` arm of
+    /// `match_damage_done` must NOT treat the disjunction as a player-scope filter
+    /// (it's an `Or`, not a controller-only `Typed`), and must delegate to
+    /// `target_filter_matches_object` so the `Typed(Battle)` leg resolves.
+    ///
+    /// Regression: fires for a battle object recipient, does NOT fire for a
+    /// non-battle object recipient, and still fires for a player recipient.
+    #[test]
+    fn battle_object_recipient_fires_for_player_or_battle_trigger() {
+        let mut state = setup();
+        let source = create_object(
+            &mut state,
+            CardId(1),
+            PlayerId(0),
+            "Trigger Source".to_string(),
+            Zone::Battlefield,
+        );
+        let battle = create_object(
+            &mut state,
+            CardId(2),
+            PlayerId(1),
+            "Invaded Farmland".to_string(),
+            Zone::Battlefield,
+        );
+        state
+            .objects
+            .get_mut(&battle)
+            .unwrap()
+            .card_types
+            .core_types
+            .push(CoreType::Battle);
+        let creature = create_object(
+            &mut state,
+            CardId(3),
+            PlayerId(1),
+            "Noncombatant Creature".to_string(),
+            Zone::Battlefield,
+        );
+        state
+            .objects
+            .get_mut(&creature)
+            .unwrap()
+            .card_types
+            .core_types
+            .push(CoreType::Creature);
+
+        // Exact shape `parse_damage_to_qualifier` emits for "a player or battle".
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
Thanks for the focused parser work. I think the target class is real, but this head does not clear the maintainer bar yet.

**[MED] Add a runtime regression for the Battle object recipient path.** The new parser tests only assert that Oracle text lowers to `TargetFilter::Or { Player/Opponent, Typed(Battle) }` in `crates/engine/src/parser/oracle_trigger.rs`, but the behavior depends on the damage matcher accepting that mixed filter through the object-recipient branch. The relevant runtime seam is `match_damage_done`'s `TargetRef::Object` arm, which rejects pure player-scope filters and then delegates to `target_filter_matches_object`. Please add a matcher- or parser-fed runtime regression where a parsed `"...deals [combat] damage to a player or battle"` / `"...to an opponent or battle"` trigger fires for `GameEvent::DamageDealt { target: TargetRef::Object(<battle>) }`, and does not fire for a non-battle object. This is the production behavior this PR is claiming, and it is exactly where mixed player/object recipient filters can regress while AST-shape tests stay green.

**[MED] Fix the CR citations for damage to battles.** The new comments repeatedly cite `CR 310.1` for `"or battle"` damage-recipient handling, but `CR 310.1` is about casting battle cards. The verified damage-recipient rules are `CR 120.1` (objects can deal damage to battles, creatures, planeswalkers, and players), `CR 120.1a` (damage cannot be dealt to other object types), and for damage results `CR 120.3h` (damage to a battle removes defense counters). Please replace the `CR 310.1` annotations with the relevant verified `CR 120.x` rule(s) so the comments point at the rule being implemented.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed current head after the CR-citation follow-up. The incorrect `CR 310.1` citations are fixed, but the behavior-proof blocker from the prior review remains.

**[MED] Add a runtime regression for the Battle object recipient path.** Evidence: the added coverage in `crates/engine/src/parser/oracle_trigger.rs:14924` and nearby tests still asserts parser `valid_target` shape only, while the production consumer is `match_damage_done`'s `TargetRef::Object` arm at `crates/engine/src/game/trigger_matchers.rs:1207-1222`. Why it matters: mixed player/object filters such as `Player | Battle` can parse correctly but still fail, overfire, or underfire in the actual damage matcher path. Suggested fix: add a matcher- or parser-fed runtime regression where a parsed "deals damage to a player or battle" / "to an opponent or battle" trigger fires for `GameEvent::DamageDealt { target: TargetRef::Object(<battle>) }` and does not fire for a non-battle object.

matthewevans:
Current-head re-review on `fbec381`:

The prior blocker is addressed. The new runtime regression covers the object-recipient path directly: a `DamageDealt` event targeting a Battle object fires the player-or-battle trigger, a non-Battle object does not, and a player recipient still fires. The parser-side CR citations are now on the damage-recipient rules rather than the battle-card rules.

I am holding approval/enqueue because GitHub still reports this branch as BEHIND and Rust lint/tests/card-data are still running on the exact head. Once the branch is current and exact-head CI is green/CLEAN, this looks approvable.

matthewevans (approved):
Approved exact head `fbec3813` after the prior blocker was addressed and GitHub CI is green/CLEAN. Repush guard held after assigning @me.

PR #4139, opened 2026-06-22
fix(engine): Regenerate window after combat damage (#3992)

### LINKED ISSUE #3992: Regenerate — Activating a regenerate ability of Lotleth Troll did not save it from dying to combat damage.
Reported in Discord: https://discord.com/channels/1485498006781427802/1518031160817614988/1518031160817614988

**Thread:** Regenerate
**Cards:** lotleth troll, regenerate
**Parser status:** has_gaps

## Summary
Activating a regenerate ability of Lotleth Troll did not save it from dying to combat damage.

### lotleth troll
Trample
Discard a creature card: Put a +1/+1 counter on this creature.
{B}: Regenerate this creature.
### regenerate
Regenerate target creature. (The next time that creature would be destroyed this turn, instead tap it, remove it from combat, and heal all damage on it.)

---
<sub>report_id: `discord:1518031160817614988:1518031160817614988:0` · discord: `1518031160817614988/1518031160817614988`</sub>

kiannidev:
## Summary
- After regular combat damage is dealt, grant priority before running lethal-damage SBAs so players can activate Regenerate.
- Skip post-action SBA processing while the deferred combat-damage window is open.
- Add a Lotleth Troll integration test that regenerates during post-damage priority.

Fixes #3992

## Test plan
- [x] `cargo test -p engine --test integration issue_3992`
- [x] `cargo test -p engine --lib combat_damage`

### DIFF
--- a/crates/engine/src/game/combat_damage.rs
+++ b/crates/engine/src/game/combat_damage.rs
@@ -3470,4 +3470,39 @@ mod tests {
         );
         assert_eq!(state.players[1].life, 20);
     }
+
+    /// CR 701.19a + CR 510.2: Regeneration shield installed before combat
+    /// damage must survive lethal damage from the combat-damage SBA pass.
+    #[test]
+    fn regeneration_shield_survives_combat_damage_resolution() {
+        use crate::types::ability::{ReplacementDefinition, TargetFilter};
+        use crate::types::replacements::ReplacementEvent;
+
+        let mut state = setup();
+        let attacker = create_creature(&mut state, PlayerId(0), "Lotleth Troll", 2, 1);
+        let blocker = create_creature(&mut state, PlayerId(1), "Big Blocker", 5, 5);
+        {
+            let shield = ReplacementDefinition::new(ReplacementEvent::Destroy)
+                .valid_card(TargetFilter::SelfRef)
+                .description("Regenerate".to_string())
+                .regeneration_shield();
+            state
+                .objects
+                .get_mut(&attacker)
+                .unwrap()
+                .replacement_definitions
+                .push(shield);
+        }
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
This does not clear the rules-correctness bar.

**[HIGH] This creates a post-damage regenerate activation window that the CR does not allow.** The new flow defers lethal-damage SBAs after regular combat damage (`regular_damage_dealt && !regular_damage_done`) and returns priority before running them, and the new integration test waits until Lotleth has marked damage before activating regenerate. Verified rules contradict that premise: `CR 510.3` says the active player gets priority after combat damage, but `CR 704.3` says whenever a player would get priority, SBAs are checked and repeated first, then pending triggers are put on the stack, and only once that loop is stable does the player actually get priority. `CR 120.6` says lethal damage destroys the creature as an SBA, and `CR 701.19a` / `CR 614.8` make regeneration a destruction-replacement shield that must already exist when the destruction event would happen. The CR example at `CR 614.8` is explicit: once the destroying spell resolves, it is too late to regenerate the creature. Combat damage uses the same SBA-before-priority ordering.

So the fix should not let a player wait until after combat damage is marked, then activate Lotleth to save it. The correct production-path regression should show the shield being created before lethal combat damage is dealt (for example in the declare blockers / pre-combat-damage priority window), then the subsequent SBA destruction is replaced. The current test proves the opposite behavior and the pipeline changes broadly defer SBAs/triggers in a way that would make other post-damage actions run in a non-CR state.

matthewevans (approved):
Approved current head 2dfca12906930c1df0b55239008b9dce3269ed44 after re-review. The prior rules blocker is addressed: the post-combat-damage regeneration window was removed, named-self regeneration now installs a shield from the normal activation path, and green CI covers the pre-existing-shield SBA path.

PR #4154, opened 2026-06-22
feat(engine): implement Augment host-combine runtime semantics

### LINKED ISSUE #4153: feat(engine): implement Augment casting and host-combine runtime semantics
# Current state

The codebase already recognizes **Augment** as a keyword for card metadata and filtering, but it does **not** implement the actual gameplay mechanic.

What exists now:
- `Keyword::Augment` / `KeywordKind::Augment`
- parsing for `"augment"` in keyword/filter contexts
- support for queries like `"card with augment"`

What is still missing:
- casting an augment half for its augment cost
- targeting a legal host
- combining the augment half with that host at runtime
- support for host/combine effects that depend on the same mechanic

In current card data, augment cards still compile to `Unimplemented` spell effects for their `Augment {cost}` text.

# Motivation

Augment is still an engine gap for the Unstable host/augment card set.

There are **14 augment-half cards** in current data, and they are not playable using their augment mechanic. In addition, some support cards and effects refer to:
- cards with augment
- hosts
- combining with a host

So this is not only a parser gap. It is a missing runtime mechanic.

# Scope

Implement the **runtime** Augment mechanic, reusing existing merge infrastructure where appropriate.

Required behavior:
- parse and represent the augment cost as a real cast path, not an `Unimplemented` effect
- allow an augment half to be cast from hand for its augment cost
- require a legal host target
- combine the augment half with that host into a merged permanent
- apply merged characteristics per the Unstable host/augment rules
- support cards that instruct a player to combine a card with augment with a host outside the normal cast-for-augment flow

# Proposed approach

1. Reuse the existing merge/mutate infrastructure as the starting point for the merged-permanent representation.
2. Add Augment-specific cast handling:
   - alternative cast path
   - legality checks for host targeting
   - resolution path that combines onto the host
3. Add host/combine runtime support for non-cast effects that refer to cards with augment.
4. Add focused engine tests in a dedicated `augment_tests.rs`.

# Acceptance criteria

- [ ] All 14 augment-half cards can be cast for their augment cost.
- [ ] Casting for augment requires and targets a legal host.
- [ ] Resolution combines the augment half with the host into one merged permanent.
- [ ] The resulting permanent has the correct merged power/toughness, abilities, and combined characteristics per Unstable rules.
- [ ] Support cards/effects that refer to “card with augment” + “combine it with a host” function correctly.
- [ ] `cargo parser-gaps` no longer reports Augment as `F_new_mechanic`.

# Notes

- Do **not** frame this as “add the Augment keyword”; keyword recognition already exists.
- The real missing work is **runtime casting + host-combine semantics**.
- This is a legitimate feature request, but it is still a **low-to-medium priority** engine item because it mainly serves Un-set cards rather than mainstream format coverage.

carlos4s:
## Summary

Closes #4153

Implement Augment runtime support in the engine by synthesizing real host-combine abilities from card data and resolving them through shared merge infrastructure. This makes augment halves and host/combine support cards playable instead of exporting as unimplemented mechanics.

the pipeline):

> Implemented directly in this Codex session rather than through the `/engine-implementer` slash-command workflow.

> [!NOTE]

## CR references

None. Augment is an Unstable mechanic and is not covered by the bundled Comprehensive Rules file; the implementation reuses existing merge/runtime behavior where applicable.

## Verification

- `cargo test -p engine augment --no-run` — passed
- `cargo test -p engine augment` — passed
- `./scripts/gen-card-data.sh` — passed
- `cargo parser-gaps` — passed; Augment no longer appears under `F_new_mechanic`
- `cargo test -p engine --test integration anaphoric_scope_allowlist_guard` — passed

### DIFF
--- a/crates/engine/src/database/augment.rs
+++ b/crates/engine/src/database/augment.rs
@@ -1,0 +1,283 @@
+//! Augment / Host synthesis for Unstable.
+
+use crate::database::mtgjson::parse_mtgjson_mana_cost;
+use crate::types::ability::{
+    AbilityCondition, AbilityCost, AbilityDefinition, AbilityKind, AbilityTag, CombineSource,
+    ControllerRef, Effect, FilterProp, TargetFilter, TypeFilter, TypedFilter,
+};
+use crate::types::card::CardFace;
+use crate::types::card_type::Supertype;
+use crate::types::keywords::{Keyword, KeywordKind};
+use crate::types::mana::ManaCost;
+use crate::types::triggers::TriggerMode;
+use crate::types::zones::Zone;
+use crate::types::zones::EtbTapState;
+
+pub fn synthesize_augment(face: &mut CardFace) {
+    let oracle = face.oracle_text.clone().unwrap_or_default();
+    let has_augment = face
+        .keywords
+        .iter()
+        .any(|keyword| matches!(keyword, Keyword::Augment));
+    let references_host_combine = oracle.contains("combine it with")
+        || oracle.contains("combine this card")
+        || oracle.contains("card with augment");
+    if !has_augment && !references_host_combine {
+        return;
+    }
+
+    if has_augment {
+        synthesize_augment_half(face, &oracle);
+    }
+
+    match face.name.as_str() {
+        "Teacher's Pet" => rewrite_teachers_pet(face),
+        "Dr. Julius Jumblemorph" => rewrite_dr_julius(face),
+        "Strutting Turkey" => rewrite_strutting_turkey(face),
+        _ => {}
+    }
+}
+
+fn synthesize_augment_half(face: &mut CardFace, oracle: &str) {
+    let Some(cost) = augment_cost(face, oracle) else {
+        return;
+    };
+    let printed_augment_description = face
+        .abilities
+        .iter()
+        .filter_map(|ability| ability.description.clone())
+        .find(|description| description.starts_with("Augment "))
+        .unwrap_or_else(|| "Augment".to_string());
+
+    face.abilities.retain(|ability| {
+        let Some(description) = ability.description.as_deref() else {
+            return true;
+        };
+        if description.starts_with("Augment ") {
+            return false;
+        }
+        if description
+            == "You may activate this card's augment ability any time you could cast an instant."
+        {
+            return false;
+        }
+        true
+    });
+
+    for ability in &mut face.abilities {
+        let is_prefix_placeholder = ability.kind == AbilityKind::Activated
+            && matches!(ability.effect.as_ref(), Effect::Unimplemented { name, .. } if name == "empty")
+            && ability
+                .description
+                .as_deref()
+                .is_some_and(|description| description.trim_end().ends_with(':'));
+        if is_prefix_placeholder {
+            ability.effect = Box::new(Effect::NoOp);
+        }
+    }
+
+    let mut augment_ability = AbilityDefinition::new(
+        AbilityKind::Activated,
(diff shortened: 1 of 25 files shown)


matthewevans (changes requested):
VERDICT: request-changes

Thanks for taking on Augment; this is not a low-effort/test-only PR. I do need to block this head before it can merge.

HIGH: Augment currently emits the existing `GameEvent::Mutated` event (`crates/engine/src/game/augment.rs`). That event is documented and consumed as the mutate-only signal (`crates/engine/src/types/events.rs`, `crates/engine/src/game/trigger_matchers.rs::match_mutates`), and `match_mutates` fires `TriggerMode::Mutates` for any `Mutated { merged_id, .. }` whose merged object matches the source/filter. As a result, combining an Augment card with a Host will incorrectly fire "whenever this creature mutates" / "whenever a creature you own mutates" triggers. Augment and Mutate share merge-like bookkeeping, but they are not the same game event/mechanic. Please split this event model, for example by adding a host-combine-specific event or by carrying a merge kind and gating mutate triggers to `MergeKind::Mutate`, then add a regression proving Augment does not satisfy mutate triggers.

Also note the current CI head is red for Rust lint, Rust tests, and WASM compile. The run was still in progress when I checked, so the failed logs were not available yet, but those need to be addressed too.

matthewevans (approved):
VERDICT: approve

Re-review on 7f5c9c4957fd2c9844fa5612ab6dc8bb13ad6e2b: prior blocker is addressed. Augment now emits dedicated GameEvent::Augmented, mutate triggers remain gated to GameEvent::Mutated, and the regression asserts augment does not emit the mutate event. I also re-checked the Gemini concerns against this head: existing merged components are preserved, host copiable values come from the current layer-computed values before augmenting, and the local Comprehensive Rules file explicitly says Unstable mechanics are not covered by the CR, so I am not requiring invented CR citations for Host/Augment behavior. CI is green and the repush guard held.

matthewevans:
Thanks for this — solid execution, and merged. The `Augmented`/`Mutated` event split is exactly right, and you addressed every review point (merged-component preservation, copy-effect handling via current copiable values) in the follow-up. One note on the CR front: you were correct to *omit* annotations — Augment is an Un-set mechanic with no Comprehensive Rules section, so there's nothing to cite (the bot's `CR 702.130d` suggestion points at Afflict, which is unrelated).

For your next PR, some context on where the leverage is: **Un-set mechanics are the lowest-priority tier in this project.** They can't be validated against the Comprehensive Rules (our #1 hard rule), they're dead mechanics with no future printings, and they serve a niche slice of players. This one was worth merging because it's complete and it exercised the shared merge infrastructure — but it isn't where the demand is.

If you want high-impact work, the **swallowed-clause** gaps are the sweet spot — cards that otherwise parse, but where the parser silently drops a load-bearing clause:

- **Swallowed conditional `if` (~553 cards, 475 of them single-gap)** — handling the `if`-condition shell makes those cards fully playable. Highest leverage available. Start from `cargo coverage --warning-detector Condition_If`.
- **Dynamic quantity (~276 cards)** — "for each…", "equal to the number of…" clauses dropped from effects. `cargo coverage --warning-detector DynamicQty`.
- **Already-supported cards that misparse an `if` (80 cards)** — silent correctness bugs in shipped cards (sproutwatch dryad, marchesa, resolute monarch, the valeyard). `cargo semantic-audit` → `DroppedCondition: if`.

The `/engine-implementer` pipeline is expected for any of this, since it's `crates/engine/` game logic.

PR #4162, opened 2026-06-22
fix(engine): thread in-flight exclusion chain through costed mana-ability payment

matthewevans:
Two costed mana abilities controlled by one player (e.g. Dimir Signet +
Gruul Signet, each {1},{T}: Add {C}{C}) caused unbounded mutual recursion
during mana auto-tap: paying Signet A's {1} auto-tapped Signet B, and
paying B's {1} re-tapped A, forever -> stack overflow -> game freeze
(observed in AI choose_action on a 4-player Commander board).

Root cause: pay_mana_sub_cost rebuilt the auto-tap exclusion set as
HashSet::from([source_id]), discarding the in-flight ancestor chain, and
casting_costs.rs resolved the mana ability via the no-exclusion entry
point. CR 605.3c: a mana ability mid-activation can't be activated again
until it resolves -- which applies to every ancestor link synchronously
suspended on the call stack, so all must be excluded from a nested
sub-cost payment, not just the immediate source.

Thread excluded_sources through resolve_mana_ability_excluding ->
pay_mana_ability_cost -> pay_mana_ability_cost_with_choices (both
AbilityCost::Mana sites) -> pay_mana_sub_cost, unioning the chain instead
of rebuilding it. Public resolve_mana_ability keeps its signature and
delegates with an empty set (top-level activations are chain roots).
Fixes the class (any pair/chain of costed mana abilities), not just two
Signets. Adds discriminating runtime regression tests.

### DIFF
--- a/crates/engine/src/game/casting_costs.rs
+++ b/crates/engine/src/game/casting_costs.rs
@@ -6687,12 +6687,23 @@ fn auto_tap_mana_sources_inner(
                 // returns, so it — and every ancestor already in
                 // `excluded_sources` — must be excluded from any nested
                 // auto-tap, or two cross-paying costed mana abilities recurse
-                // infinitely. Hoisted above the sub-cost guard so the chain is
-                // threaded into resolution even for abilities with no mana
-                // sub-cost.
-                let mut excluded = excluded_sources.clone();
-                excluded.insert(option.object_id);
-                if let Some(sub_cost) = mana_sub_cost_of(&ability_def.cost) {
+                // infinitely. The exclusion set is read only by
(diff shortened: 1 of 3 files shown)


matthewevans:
Both review points addressed in 52b11b993:

- **Hot-path clone**: the in-flight exclusion set is read only by `pay_mana_sub_cost`, reached only through the `AbilityCost::Mana` arms — exactly what `mana_sub_cost_of` reports `Some` for. Now clones + extends only when a mana sub-cost is present; tap-only/sacrifice mana abilities forward the caller's reference unchanged, skipping the allocation.
- **Double reference**: `excluded` is now a `&HashSet` (rebound by the conditional), so both the recursion and resolve call sites pass it directly.

PR #4164, opened 2026-06-22
feat(engine): equipment-attached count quantity (Winter Soldier, Whiplash)

lgray:

## Building block: source-anaphoric "Equipment/Aura attached to {source}" count

This PR builds the reusable count for **"the number of Equipment/Aura attached to {the source}"** and wires two MSH cards end-to-end. **No new engine enum variant** — the quantity reuses the existing `QuantityRef::ObjectCount { filter }` + `FilterProp::AttachedToSource`, and the dynamic P/T scaling reuses `ContinuousModification::AddDynamicPower`/`AddDynamicToughness` (the same shape Armament Master already produces). The work is parser wiring plus one trigger-condition bridge, built for the **class**, not the card.

### What changed (parser, build-for-the-class)

- **Attachment-count referent (`oracle_nom/quantity.rs`).** `parse_for_each_attached_to_source` now accepts the source-anaphoric gendered pronoun `"him"`/`"her"` (`" attached to him"`) → `AttachedToSource`, alongside the existing `"~"` arm. The recipient pronoun `"it"`/`"that creature"` → `AttachedToRecipient` is preserved unchanged. The singular-they `"them"` is **deliberately excluded** because it is recipient-anaphoric for player-enchanting Auras (e.g. *Curse of Thirst* — "Curses attached to them" = the enchanted **player**, not the Aura source), which would bind the wrong object set.
- **Quantity path (`oracle_quantity.rs`).** The `"the number of <type> attached to <source>"` block delegates to the same combinator, so `"where X is the number of Equipment attached to him"` lowers to the typed `ObjectCount` instead of the `Variable` string fallback.
- **Contraction source-state conditions (`oracle_nom/condition.rs`).** New `parse_contraction_source_state_condition`: `"he's"/"she's"/"they're" + <state>` (equipped / enchanted / tapped / monstrous / attacking / blocking / blocked) → the matching `Source*` condition. Bare `"it's"` is **excluded** (target-anaphoric in spell bodies, e.g. *Awaken the Sleeper* "Gain control of target creature … if it's equipped"). Modeled on the existing `parse_recipient_is_filter_condition` copula-then-bare-predicate shape.
- **Self-ref static rewrite (`oracle_static/anthem.rs`).** `rewrite_self_pronoun_subject` (SelfRef-gated) gains `"equipped"`/`"enchanted"`, so self-ref statics `"… as long as it's equipped/enchanted"` bind to the source.
- **Trigger-condition bridge (`oracle_trigger.rs`).** `StaticCondition::SourceIsEquipped`/`SourceIsEnchanted` intervening-ifs now bridge to the existing generic `TriggerCondition::SourceMatchesFilter { filter: creature + HasAttachment(Equipment/Aura) }` (the runtime handler doc explicitly anticipates this: "so properties such as enchanted/equipped … do not need bespoke TriggerCondition siblings") instead of being dropped to `None`. This carries Whiplash's `"if he's equipped"` gate; with the gate-clause stripped, the each-opponent subject strip fires.

### Class this unlocks

The source-pronoun count + condition handling covers the "for each Equipment/Aura attached to {source}" + "<source-pronoun>'s equipped/enchanted/…" family beyond the two cards below — e.g. Captain America, Liberator (attachment count), and the self-ref static `"as long as it's equipped/enchanted"` cards (Merry, Esquire of Rohan; Fledgling Osprey; Skyrider Trainee; Metathran Elite). The bare-`"it's"` trigger cases (Tetsuo, Krond, etc.) remain gapped on a **separate** self-ref normalization issue, not on this bridge.

### Cards shipped

- **Winter Soldier, Icy Assassin** — "Winter Soldier gets +2/+0 for each Equipment attached to him." The static now scales power dynamically (`AddDynamicPower(Multiply{2, ObjectCount{Equipment + AttachedToSource}})`) instead of misparsing to a fixed +2/+0. The +0 toughness component is correctly omitted (a +0 dynamic mod is a layer no-op). The graveyard-return activated ability is unchanged.
- **Whiplash, Vengeful Engineer** — "Whenever Whiplash attacks, if he's equipped, each opponent loses X life and you gain X life, where X is the number of Equipment attached to him." Now parses with the `"if he's equipped"` intervening-if (bridged to `SourceMatchesFilter{HasAttachment(Equipment)}`), `X = ObjectCount{Equipment + AttachedToSource}` on both LoseLife/GainLife, and the each-opponent form (`player_scope = Opponent`, no explicit target).

- **CR 301.5a** — Equipment / "equipped creature".
- **CR 303.4** — Aura / enchanted.
- **CR 603.4** — intervening-"if" clause (checked at trigger and rechecked on resolution; drives the bridged source-state gate).
- **CR 508.1k / 509.1g / 509.1h** — combat-state predicates (attacking/blocking/blocked arms).

### Tests (non-vacuous + discriminating)

- Card-synthesis for both cards (fail-before: fixed +2/+0; `Variable` X with null condition / empty target).
- Combinator-level positive + regression + negative: `"attached to it"` → recipient (preserved); `"the number of creatures you control"` → generic `ObjectCount` (not shadowed); `"it's equipped"` NOT a source condition (Awaken-the-Sleeper guard); `"they's"/"he're"` rejected; `"attached to them"` NOT source-bound (Curse-of-Thirst guard).
- Runtime gating test proving the bridged condition fires only when the source actually has an Equipment attached (false with no attachment, false with an Aura only, true with an Equipment).

No new serialized surface (maps to existing `TriggerCondition`/`FilterProp` variants). `mtgish/` untouched.

### DIFF
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -12525,6 +12525,100 @@ pub mod tests {
         ));
     }
 
+    /// CR 603.4 + CR 303.4: runtime gate for the enchanted intervening-if bridge on
+    /// a NON-creature source (a bare Artifact). CR 303.4 allows an Aura to enchant
+    /// any object, so the source-object-wide bridge must fire when an Aura is
+    /// attached even though the source is not a creature.
+    ///
+    /// Drives the PRODUCTION bridge
+    /// (`parser::oracle_trigger::static_condition_to_trigger_condition`) rather than
+    /// a hand-built filter, so the test discriminates against the bridge code.
+    /// Fail-before: the bridge produced `TypedFilter::creature()`, so the
+    /// non-creature artifact source failed the card-type gate and
+    /// `check_trigger_condition` returned `false` even with the Aura attached.
+    /// Hostile negatives: no attachment → false; an Equipment (wrong kind) on the
+    /// non-creature source → false.
+    #[test]
+    fn source_matches_filter_gates_on_enchanted_noncreature() {
+        let mut state = setup();
+        let src = create_object(
+            &mut state,
+            CardId(1),
+            PlayerId(0),
+            "Enchanted Artifact".to_string(),
+            Zone::Battlefield,
+        );
+        // Artifact only — deliberately NOT a Creature.
+        state
+            .objects
+            .get_mut(&src)
+            .unwrap()
+            .card_types
+            .core_types
+            .push(CoreType::Artifact);
+
+        let cond = crate::parser::oracle_trigger::static_condition_to_trigger_condition(
+            &StaticCondition::SourceIsEnchanted,
+        )
+        .expect("SourceIsEnchanted should bridge to a TriggerCondition");
+
+        // No attachment → does not fire.
+        assert!(!check_trigger_condition(
+            &state,
+            &cond,
+            PlayerId(0),
+            Some(src),
+            None,
+        ));
+
+        // Hostile: an Equipment attached to the non-creature source must NOT
+        // satisfy the Aura predicate (attachment-kind discrimination).
+        let equip = create_object(
+            &mut state,
+            CardId(2),
+            PlayerId(0),
+            "Test Equipment".to_string(),
+            Zone::Battlefield,
+        );
+        {
+            let equip_obj = state.objects.get_mut(&equip).unwrap();
+            equip_obj.card_types.core_types.push(CoreType::Artifact);
+            equip_obj.card_types.subtypes.push("Equipment".to_string());
+            equip_obj.attached_to = Some(crate::game::game_object::AttachTarget::Object(src));
+        }
+        state.objects.get_mut(&src).unwrap().attachments.push(equip);
+        assert!(!check_trigger_condition(
+            &state,
+            &cond,
+            PlayerId(0),
+            Some(src),
+            None,
+        ));
+
+        // Aura attached to the non-creature source → fires (the revert-failing case).
+        let aura = create_object(
+            &mut state,
+            CardId(3),
+            PlayerId(0),
+            "Test Aura".to_string(),
+            Zone::Battlefield,
+        );
+        {
+            let aura_obj = state.objects.get_mut(&aura).unwrap();
+            aura_obj.card_types.core_types.push(CoreType::Enchantment);
+            aura_obj.card_types.subtypes.push("Aura".to_string());
+            aura_obj.attached_to = Some(crate::game::game_object::AttachTarget::Object(src));
+        }
+        state.objects.get_mut(&src).unwrap().attachments.push(aura);
+        assert!(check_trigger_condition(
+            &state,
+            &cond,
+            PlayerId(0),
+            Some(src),
+            None,
+        ));
+    }
+
     // === CR 701.27g + CR 708.2: Source-state predicates (Transformed/FaceUp/FaceDown) ===
 
     #[test]
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -12436,6 +12436,95 @@ pub mod tests {
         ));
     }
 
+    /// CR 603.4 + CR 301.5a: runtime gate for the equipped intervening-if bridge
+    /// (oracle_trigger.rs `static_condition_to_trigger_condition`,
+    /// `SourceIsEquipped → SourceMatchesFilter{HasAttachment(Equipment)}`).
+    /// Discriminates: false with no attachment, false with only an Aura attached
+    /// (proves kind matters), true once an Equipment is attached.
+    #[test]
+    fn source_matches_filter_gates_on_equipped() {
+        let mut state = setup();
+        let src = create_object(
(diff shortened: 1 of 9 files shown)


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** `SourceIsEnchanted` trigger conditions are narrowed to creatures even though the parsed condition is source-object-wide. Evidence: `crates/engine/src/parser/oracle_trigger.rs:2893` lowers `StaticCondition::SourceIsEnchanted` to `TypedFilter::creature().properties(HasAttachment(Aura))`, but `crates/engine/src/parser/oracle_nom/condition.rs:1020` accepts source subjects including `this permanent`, `this land`, `this artifact`, and `this enchantment`, and the existing static/layer evaluator at `crates/engine/src/game/layers.rs:1181` treats `SourceIsEnchanted` as “any source object with an Aura attached.” Why it matters: an intervening-if trigger on an enchanted noncreature permanent would parse as supported but fail its trigger gate because the bridged filter requires the source to be a creature. Suggested fix: bridge `SourceIsEnchanted` with a permanent/object-wide source filter (or no creature type restriction) and add a regression with an enchanted noncreature source.

Gemini’s `oracle_quantity.rs` comment concern is partially valid but not the blocker: the nearby prose still says `them` denotes the source, while the shared combinator deliberately excludes `them`; please clean that comment while touching this. I do not see the trailing-whitespace point as a behavior bug because the quantity parser trims before this full-consume check.

lgray:

Thanks @matthewevans — fixed in `bbc13d1ad` (rebased on current `upstream/main`).

## Maintainer finding (source-state→trigger bridge narrows to creature) — FIXED

You were right: the bridge was narrowing a source-object-wide condition.

- `static_condition_to_trigger_condition` lowered both `SourceIsEquipped` and `SourceIsEnchanted` to `SourceMatchesFilter { TypedFilter::creature().properties([HasAttachment{..}]) }`, but `parse_source_subject` (`oracle_nom/condition.rs:1020-1032`) accepts non-creature subjects (`this permanent` / `this land` / `this artifact` / `this enchantment`), and **both** runtime evaluators are source-object-wide with no card-type gate: the layer evaluator (`game/layers.rs:1172-1184`) and `FilterProp::HasAttachment` (`game/filter.rs:3370-3393`). So an enchanted **non-creature** source silently failed the bridged intervening-if while the layer path accepted it.
- Fix: both arms now use `TypedFilter::default()` (empty `type_filters` → no card-type constraint; the object matcher’s `type_filters.iter().all(..)` is vacuously true). The `HasAttachment{kind, controller: None, exclude_source: Include}` predicate is unchanged.
- Per CR: **CR 303.4** — an Aura enters attached to an object *or* player, so non-creature hosts are legal → `SourceIsEnchanted` must be source-object-wide. **CR 301.5a/301.5c** — Equipment attaches only to creatures, so `HasAttachment{Equipment}` already implies a legal creature host; the `creature()` gate on the equipped arm was redundant and only served to diverge from the layer evaluator. So both arms are now consistent with the runtime.

Discriminating tests (all three fail on the pre-fix `creature()` code, pass after):
- `bridge_source_is_enchanted_no_creature_type` / `bridge_source_is_equipped_no_creature_type` (AST): assert the bridged filter has empty `type_filters` (pre-fix `[Creature]`), reasserting the `HasAttachment` prop as a sibling guard.
- `source_matches_filter_gates_on_enchanted_noncreature` (runtime): an **Artifact-only** source with an Aura attached is built, filtered through the **production bridge** (`static_condition_to_trigger_condition`, not a hand-built filter), and asserted to fire `check_trigger_condition`; pre-fix the `creature()` gate makes the artifact source fail → `false`. Hostile negatives in the same test: no attachment → false, Equipment-instead-of-Aura → false (kind discrimination).

## Gemini Code Assist nitpicks (`oracle_quantity.rs:507-517`) — both verified CORRECT, applied

1. **`them` in the comment** — verified against `parse_for_each_attached_to_source` (`oracle_nom/quantity.rs:3745-3765`), which maps only `" attached to him"` / `" attached to her"` / `~` to `AttachedToSource` and **explicitly excludes** `them` as recipient-anaphoric (player-enchanting Auras, e.g. Curse of Thirst). Comment corrected to move `them` to the recipient side.
2. **`rest_after.is_empty()`** — the four sibling completeness checks in the same function (lines 470/482/489/527) all use `.trim().is_empty()`. Changed to `rest_after.trim().is_empty()` for consistency and trailing-whitespace robustness.

Verification: `cargo fmt --all`, parser-combinator gate (exit 0), `cargo clippy -p engine --all-targets` clean, and the bridge + `source_matches_filter` + `parse_quantity_ref_number_of_*` suites green.

matthewevans (approved):
Approving exact head `bbc13d1ad9eee3d5a11d8d7a356533e4986211d8` after re-review. The follow-up addresses the prior SourceIsEnchanted blocker by using source-object-wide HasAttachment filters, and the new production-bridge regression covers an enchanted noncreature source with hostile no-attachment/wrong-kind negatives. GitHub checks are green and mergeStateStatus is CLEAN.

PR #4169, opened 2026-06-22
feat(engine): entering-creature-vs-source P/T intervening-if (Hulkling)

lgray:

## Building block: entering-object source-relative P/T intervening-if

This PR fixes **Hulkling, Burgeoning Bruiser** — "Vigilance / Whenever another creature you control enters, **if it has greater power or toughness than Hulkling**, put a +1/+1 counter on Hulkling." The ETB intervening-if comparing the entering creature's power/toughness to the source was being dropped (the trigger `condition` was `null`; `supported: false`, `gap_count: 1`).

**No new engine variant.** The condition is expressed entirely with existing building blocks (the `add-engine-variant` gate returned EXISTS_AS_PARAMETER):

- `TriggerCondition::ZoneChangeObjectMatchesFilter { origin, destination, filter }` — the existing entering-object intervening-if condition (its subject is the zone-change event object, i.e. the entering creature, not the source permanent).
- `FilterProp::PtComparison { stat, scope, comparator, value }` — its `value` is already a `QuantityExpr`, so a *source-relative* threshold is expressible.
- `QuantityRef::Power { scope }` / `Toughness { scope }` with `ObjectScope::Source` — resolves to the ability source's (Hulkling's) corresponding stat.
- `FilterProp::AnyOf` — composes the "power **or** toughness" disjunction.

So "greater power or toughness than ~" lowers to:

```
ZoneChangeObjectMatchesFilter {
  origin: None,
  destination: Battlefield,
  filter: Typed(creature, [AnyOf([
    PtComparison { stat: Power,     scope: Current, comparator: GT, value: Ref(Power{scope: Source}) },
    PtComparison { stat: Toughness, scope: Current, comparator: GT, value: Ref(Toughness{scope: Source}) },
  ])]),
}
```

The runtime already evaluates this end-to-end: `ZoneChangeObjectMatchesFilter` matches the filter against the entering object with the `FilterContext` source = the ability source, so `PtComparison`'s threshold resolves to the source's stat. No new runtime handler.

### What changed

- New nom combinator `parse_entering_pt_vs_source_condition` in `oracle_trigger.rs`, recognizing "if it has greater power [or toughness] than ~" and the single-stat forms ("greater power than ~" / "greater toughness than ~" → one `PtComparison`, no `AnyOf`). Wired as the **first** arm of `parse_zone_change_object_filter_condition`, before the `"if it "`-predicate fallback (which hardcodes `destination: Graveyard` — the ETB case needs `Battlefield`). Purely additive: the fallback predicate rejects "has greater …", so nothing is shadowed.
- Does **not** touch `FilterProp::PowerGTSource` (the CR 509.1b combat blocking-restriction leaf) or `nom_filter::parse_pt_comparison` (the fixed-threshold grammar).

### Class this generalizes

The combinator handles the *category* of ETB intervening-ifs comparing the entering creature's power and/or toughness to the source's same stat — Hulkling is the driver, and the single-stat arms cover any future "Whenever a creature enters, if it has greater power/toughness than ~" payoff with no new code.

- **CR 603.4** — intervening-"if" clause (checked at trigger and rechecked on resolution).
- **CR 603.6a** — ETB triggers; the subject "it" is the entering newcomer.
- **CR 113.7** — the source of an ability ("~"), bound via `ObjectScope::Source`.
- **CR 208.1** — power/toughness.
- (CR 603.10a's look-back list does **not** include ETB, so `PtValueScope::Current` reads the live post-layer P/T — correct, not LKI.)

### Tests (non-vacuous + discriminating)

- Card-synthesis: full Hulkling text → the `AnyOf` of two source-relative `PtComparison`s; `PutCounter P1P1` on self intact. Fail-before: condition was `null`.
- Single-stat discriminator: "greater power than ~" → one `PtComparison`, not `AnyOf`.
- Regression: an existing fixed-threshold P/T parse stays a fixed-value `PtComparison` (the new source-relative arm doesn't shadow it).
- Runtime gating: entering 3/3 vs Hulkling 2/2 → true; 2/2 → false (strict GT, not GE); 1/3 vs 2/2 → true (toughness-only OR — proves the threshold resolves to Hulkling, not the entering object); 1/2 vs 2/2 → false.

No new serialized surface (reuses existing `TriggerCondition`/`FilterProp`/`QuantityRef`). `mtgish/` untouched.


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** Resolution-time recheck can compare against the wrong entering creature P/T after it leaves the battlefield. Evidence: `crates/engine/src/parser/oracle_trigger.rs:3812` builds the Hulkling condition as `ZoneChangeObjectMatchesFilter`, and `crates/engine/src/game/filter.rs:1298` evaluates Battlefield zone-change filters against the live object. That works while the entrant is still on the battlefield, but `crates/engine/src/game/zones.rs:147` snapshots battlefield LKI on exit and `crates/engine/src/game/zones.rs:424` then reverts the moved object to baseline characteristics, so a creature that entered with counters or temporary P/T changes can pass the trigger check, leave before resolution, and then fail the CR 603.4 intervening-if recheck using the stripped graveyard/exile object instead of its battlefield LKI. CR 608.2h says specific-object information uses current info only if the object is still in the expected public zone; otherwise it uses LKI. Why it matters: Hulkling-style triggers can silently fizzle after a legal trigger if the entering creature leaves before resolution, even though the engine already captured the right LKI. Suggested fix: add a production resolution test for “entered with modified P/T, leaves before Hulkling resolves,” then make this Battlefield zone-change-object comparison read event/LKI data when the event object is no longer on the battlefield, or route this pattern through an event-source/LKI-aware condition.

Confidence: medium-high. This would be contradicted by a production scenario showing the resolving trigger already reads the entrant’s battlefield LKI after it leaves, but the current matcher path I traced reads the live object for `destination == Battlefield`.

lgray:

Thanks @matthewevans — fixed in `9fdf49259` (rebased on current `upstream/main`).

## Finding (resolution-time recheck compares against reverted/baseline entrant P/T) — FIXED

Confirmed exactly as you described. The ETB intervening-if recheck went through `matches_zone_change_event_object_filter` (`game/filter.rs`), whose `destination == Zone::Battlefield` branch unconditionally read the **live** entering object. But on a battlefield exit, `move_to_zone` snapshots the at-exit characteristics into `state.lki_cache` (`zones.rs:147-173`) and then calls `revert_layered_characteristics_to_base()` (`zones.rs:429`, `game_object.rs:1435`), which resets `power`/`toughness` to base. The object is kept in `state.objects` with its `zone` updated, so the recheck read **reverted baseline** P/T (or `false` for a fully-absent entrant), not the entrant's last-known on-battlefield P/T. So Hulkling missed its counter when the entrant died/bounced in response.

Per **CR 608.2h**: a filter reading a specific object's information uses current info only while it is in the public zone it was expected in; once it has left, it uses the object's **last known information**. **CR 603.10a**: ETB is not a look-back trigger, so the normal CR 608.2h rule applies (not CR 608.2i). The effect "put a +1/+1 counter on ~" targets the **source**, so the ability still resolves (CR 603.6) — only the intervening-if's info lookup on the entrant needs LKI.

Fix: the Battlefield branch now mirrors the live-vs-LKI dispatch `effective_controller` (`filter.rs:680-694`) already uses:
- entrant still on the battlefield → `matches_target_filter` (byte-identical to before);
- entrant has left → `matches_target_filter_on_lki_snapshot` with the exit-time `lki_cache` snapshot (the most-recent on-battlefield characteristics, captured **before** the base revert);
- no exit LKI cached (defensive) → the zone-change record, never the reverted live object.

The **exit** `lki_cache` is the chosen authority (not the entry-time ETB record): per CR 608.2h "most recently existed," and it is the canonical LKI every other leaves-the-battlefield filter reads. The source-relative threshold (`QuantityRef::Source`) is still read live from the Hulkling source, which has not left. This single shared seam fixes both callers (`TriggerCondition` + `AbilityCondition` `ZoneChangeObjectMatchesFilter`) and the whole class of ETB intervening-ifs that read the entrant's characteristics (P/T, type, controller, keywords, ...), all carried by the LKI snapshot. No new variant.

Discriminating test `zone_change_object_condition_entering_uses_exit_lki_after_leaving_battlefield`: an entrant with printed **base 1/1** (not greater) but layered on-battlefield **3/3** (greater) is moved off the battlefield via the production `move_to_zone` path (which caches exit LKI 3/3 and reverts the live obj to base 1/1; sanity-asserted). The recheck must be true (LKI 3/3 > source 2/2); a sibling entrant with exit LKI 1/1 must be false. The positive assertion **fails on the pre-fix code** (reads the reverted live 1/1 → 1/1 not > 2/2). The five existing entrant-still-present cases (2/2→false, 1/3→true, 1/2→false, ...) stay green, proving the common path is unchanged.

Verification: `cargo fmt --all`, `cargo clippy -p engine --all-targets` clean, the new test + the full `zone_change_object_condition` family (5) green.

matthewevans (changes requested):
VERDICT: request-changes

Prior finding: PARTIAL. The new LKI branch fixes the case where the entrant leaves the battlefield and stays out of the battlefield before the intervening-if recheck, and the new test covers that path.

**[MED]** The recheck still reads the wrong object if the entrant leaves and re-enters before the trigger resolves. Evidence: `crates/engine/src/game/filter.rs:1311` treats `state.objects[object_id].zone == Battlefield` as proof that the original entrant is still in its expected public zone and then uses `matches_target_filter` on the live object at `crates/engine/src/game/filter.rs:1316`. But `crates/engine/src/game/zones.rs:120` / `crates/engine/src/game/zones.rs:125` explicitly document that a zone change creates a new object while `ObjectId` is only storage identity and persists across that zone change. The new regression at `crates/engine/src/game/triggers.rs:9279` only moves the entrant to the graveyard and leaves it there; it never moves the same storage id back to the battlefield before the recheck. Why it matters: if the Hulkling entrant was 3/3 on the original ETB, leaves, then returns as a new 1/1 before the original trigger resolves, CR 400.7 + CR 608.2h require the original trigger to use the old entrant's LKI, but this branch will read the new battlefield object and incorrectly fail the condition. Suggested fix: distinguish the original event object instance from a later re-entry (or prefer the event/exit LKI when the object has had any intervening zone change after the trigger event) and add a production regression that exits then re-enters the same storage id before resolving/rechecking the original Hulkling condition.

Confidence: high. Verified CR 400.7, 603.4, 603.10a, and 608.2h against `docs/MagicCompRules.txt`; this would be contradicted by an engine object-generation marker proving the live battlefield object at the same `ObjectId` is still the original event object, but I found the opposite in `zones.rs` comments.

matthewevans:
Maintainer handling (pushed to `card/msh-intervening-if`):

**Addressed the second review finding (CR 400.7 leave + re-entry).** The prior LKI fix handled "entrant leaves and stays gone," but the recheck still read the wrong object when the entrant **leaves and re-enters** before the trigger resolves. `move_to_zone` reuses the same `ObjectId` on re-entry (it sets `obj.zone = to` and calls `reset_for_battlefield_entry`), so the zone-only `still_on_battlefield` check was fooled into reading the *new* incarnation's characteristics instead of the original entrant's last-known information.

Fix reuses the existing CR 400.7 `incarnation` primitive (`GameObject.incarnation`, bumped on every battlefield entry — the same mechanism `AbilityDefinition.source_incarnation` already uses to skip blinked sources):
- Add `entered_incarnation: Option<u64>` to `ZoneChangeRecord`, captured in `move_to_zone` **after** the entry bump.
- The live object counts as "still the original entrant" only when its incarnation matches the recorded entry incarnation; otherwise the recheck routes to the original exit LKI. `entered_incarnation == None` (legacy/defensive records) falls back to the zone-only check.

**Discriminating regression test** `zone_change_object_condition_uses_original_exit_lki_after_leave_and_reentry`: drives the production `move_to_zone` path through entry (3/3) → leave (caches exit LKI 3/3, reverts to base 1/1) → re-entry (bumps incarnation, base 1/1), then rechecks the original ETB event. Asserts the original exit LKI (3/3 > source 2/2) is used; reverting the incarnation gate reads the re-entered base 1/1 and flips the assertion to false.

**Brought current with `origin/main`** — resolved the `ZoneChangeRecord` literal conflicts where main concurrently added `turn_zone_change_index` (kept both fields in every literal).

**Verification:** `cargo fmt`, parser-combinator gate, the full `zone_change_object_condition` family (6 tests incl. the new re-entry case + the 5 prior siblings, all green against current main), and `clippy -p engine --lib` clean. An isolated `review-impl` pass returned CLEAN — confirming the incarnation reuse is the right primitive at the right seam, the test discriminates via the production path, `lki_cache` is insert-only (re-entry from the graveyard never overwrites the original LKI), and a real flicker mints a new `ObjectId` (→ routes to LKI), so there's no false-match window.

CR 400.7 + CR 603.4 + CR 608.2h + CR 603.10a.

matthewevans (approved):
CR 400.7 re-entry fix via incarnation primitive at the correct seam; discriminating production-path test; integration-test literals fixed; full CI green (Rust tests, lint, coverage-gate); review-impl CLEAN.

PR #4171, opened 2026-06-23
feat(client): drop-position caret for hand drag-to-rearrange

AgilErck:
## Summary

Adds a **drop-position caret** to the player's hand: when you drag a card within the hand to rearrange it, a glowing cyan vertical line shows exactly where the card will land if released, gliding (spring) between insertion slots. Today there is no cue — the reorder only happens on drop. Purely frontend; the reorder still dispatches the existing `ReorderHand { order }`.

## Approach

- **A single overlay `motion.div`** whose `x`/`y`/`opacity` are framer-motion `MotionValue`s, set imperatively inside the existing `handleDrag` (right where `hoveredSlotRef` is already written). Because `MotionValue.set()` bypasses React render, the memoized fanned `HandCard`s **never re-render on pointer move** — preserving the deliberate ref-over-state optimization. A small `useSpring` glides the caret along the fan arc (X and Y).
- **New pure helper `computeHandInsertionMarker`** (unit-tested) derives the caret position in the *same drag-excluded ("remaining") coordinate space* that `computeHandInsertionSlot` already uses — so the line is correct for landings on either side of the dragged card (the off-by-one trap).
- **One `getBoundingClientRect` sweep** per pointer move, reused for both the slot and the caret (no second measurement).
- The caret is suppressed during the upward drag-to-play gesture, outside the hand bounds, during a pending cast, on the no-op in-place slot, and on mobile (the drawer is the mobile surface). Three independent reset paths guarantee it can't get stuck visible.

## Testing

- `computeHandInsertionMarker` is unit-tested at all boundary positions (slot 0 / middle / append / clamp / null / drag-excluded) — 9/9 in `handInsertionSlot` tests.
- `tsc --noEmit` and ESLint clean.

## Manual QA caveat

This is a **purely visual** change and jsdom can't measure layout, so the caret's look-and-feel was **not** visually verified in a live game. A maintainer should confirm: the line tracks the landing slot and matches the actual drop (left/right of the dragged card, first/last positions); it hides during an upward play drag and on the no-op in-place slot; no caret during a pending cast or on mobile; and (React Profiler) the `HandCard`s do not re-render on pointer move. Spring constants (`stiffness 900 / damping 48 / mass 0.4`) and the `3px` line styling are easy to tune.

https://github.com/user-attachments/assets/d9bc8dd7-8825-45af-bdbb-6cb5a806b1d5


matthewevans:
VERDICT: request-visual-evidence

This is a visual interaction change, and the PR body explicitly says the caret was not visually verified in a live game. I’m going to hold deep review until there is visual evidence attached.

Please add a short screen recording or before/after screenshots showing:

- drag-to-rearrange with the caret tracking the actual drop slot, including left/right of the dragged card and first/last positions
- the no-op in-place slot hiding the caret
- upward drag-to-play hiding the caret
- pending-cast and mobile behavior, or a note with evidence for why those states are not reachable in the recording

No approval, label, or enqueue until that evidence is present.

matthewevans:
Really dig this one too. Ideally the blue line would be at the same angle as the cards it's going between instead of staying a straight line. If you want to give that a shot that would be sweet - if not I'll throw an agent at it.

AgilErck:
I had an idea and am working on version 2.0, so changing this to draft for now.

AgilErck:
https://github.com/user-attachments/assets/066782e1-e6b1-4379-a973-464585f17795

Updated, 1.5

matthewevans:
> card_caret_1.5_10mb.mp4
> 
> Updated, 1.5

Looks good! Was thinking that or a bouncing down-arrow in between where it will be re-positioned. Whatever feels more natural. Will defer to you :)

AgilErck:
@matthewevans Happy with this one.

https://github.com/user-attachments/assets/318912b5-b36c-48eb-9c6c-a6571d0a50c6

matthewevans:
> @matthewevans Happy with this one.
> 
> 
> 
> https://github.com/user-attachments/assets/318912b5-b36c-48eb-9c6c-a6571d0a50c6
> 
> 
> 
> 

Dig it! Super nice 👍

PR #4176, opened 2026-06-23
feat(engine): support gaining all activated abilities of target until end of turn

### LINKED ISSUE #2913: Quicksilver Elemental: '{U}: gains all activated abilities of target creature until end of turn' unimplemented
## Oracle text (verified from client/public/card-data.json)
```
{U}: This creature gains all activated abilities of target creature until end of turn. (If any of the abilities use that creature's name, use this creature's name instead.)
You may spend blue mana as though it were mana of any color to pay the activation costs of this creature's abilities.
```

## Expected vs actual
Expected: activating {U} grants Quicksilver Elemental copies of all activated abilities of the target creature until end of turn (with self-name substitution, CR 706.2). Actual: unimplemented — the ability-copy continuous effect does not exist in the engine.

Source: https://discord.com/channels/1485498006781427802/1513666646592458862/1513666646592458862

jsdevninja:
## Summary
- Adds `Effect::GainActivatedAbilitiesOfTarget`, a CR 611.2c resolution-time
  ability grant: snapshot a chosen target's activated abilities once, grant
  them to a recipient (self, or a filtered group) until end of turn.
- Unblocks Quicksilver Elemental (`{U}: ... gains all activated abilities of
  target creature until end of turn`) and Grell Philosopher (`each Horror you
  control gains all activated abilities of target artifact an opponent
  controls until end of turn`) — built as one general primitive rather than
  a single-card special case, per CLAUDE.md's "build for the class" rule.
- Self-references in granted abilities correctly rebind to the recipient
  (CR 201.5b), proven by a sacrifice-cost test that sacrifices the recipient,
  not the donor.

## Test plan
- [x] 8 new engine tests in `gain_activated_abilities.rs` (self-grant, EOT
      expiry, CR 201.5b name substitution, empty-donor no-op, multi-grant
      stacking, uncontrolled-donor targeting, group recipient, CR 611.2c
      snapshot-not-live-rescan regression)
- [x] 6 new parser tests in `imperative.rs` (both card phrasings, malformed
      input falls through to `Unimplemented`, dispatch-ordering regression
      vs. bare "gains flying", full real-Oracle-text parse-through for both
      cards including the group-recipient rebind)
- [x] `cargo fmt --all`, `cargo clippy -p engine -p phase-ai --all-targets -- -D warnings` clean
- [x] Two independent `/review-impl` rounds (one fix round: corrected a CR
      citation 707.9a→201.5b, added missing parser-level coverage for Grell
      Philosopher's group-recipient path)

Fixes #2913

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -2008,6 +2008,8 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
         | Effect::CopySpell { target, .. }
         | Effect::CastCopyOfCard { target, .. }
         | Effect::BecomeCopy { target, .. }
+        // CR 113.1a + CR 611.2: report the donor whose activated abilities are gained.
+        | Effect::GainActivatedAbilitiesOfTarget { target, .. }
         | Effect::Suspect { target, .. }
         | Effect::Unsuspect { target, .. }
         | Effect::Connive { target, .. }
--- a/crates/engine/src/game/effects/gain_activated_abilities.rs
+++ b/crates/engine/src/game/effects/gain_activated_abilities.rs
@@ -107,8 +107,20 @@ pub fn resolve(
             // multiple recipients (or multiple sequential grants) never collide
             // or overwrite each other (CR 611.2c fixes each affected set
             // independently).
+            //
+            // CR 112.6a: "you control" must resolve against the resolving
+            // ability's controller, not whatever player currently controls the
+            // live source object. `FilterContext::from_ability` binds
+            // `source_controller` to `ability.controller` (captured when the
+            // ability was put on the stack), so the grant still lands on the
+            // correct player's Horrors even if the source's controller has
+            // since changed or the source has left the battlefield entirely
+            // (`FilterContext::from_source` would instead read the live
+            // object's controller, or `None` if the source object no longer
+            // exists in `state.objects` — silently matching the wrong
+            // player, or no one).
             _ => {
-                let ctx = FilterContext::from_source(state, ability.source_id);
+                let ctx = FilterContext::from_ability(ability);
                 let recipient_ids: Vec<_> = state
                     .battlefield
                     .iter()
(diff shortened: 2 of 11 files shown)


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** Group recipients are evaluated with the live source object's controller instead of the resolving ability controller. Evidence: `crates/engine/src/game/effects/gain_activated_abilities.rs:111` builds `FilterContext::from_source(state, ability.source_id)` before matching the `recipient` filter, while `crates/engine/src/game/filter.rs:534-551` documents `from_source` as the no-ability/layer context and `from_source_with_controller` for stack-resolving effects whose source may have left play. Why it matters: Grell's "each Horror you control" is a controller-relative resolution effect; if the source has left the battlefield or its live object controller differs from `ability.controller`, the grant can be applied to the wrong player's Horrors or to none, even though the trigger/effect is controlled by the original ability controller. Suggested fix: evaluate group recipients with `FilterContext::from_ability(ability)` (or at minimum `from_source_with_controller(ability.source_id, ability.controller)`) and add a regression where the source leaves or no longer has the same live controller before the group-recipient grant resolves.

No approval, label, or enqueue on this head.

matthewevans:

Focused re-review of `a6d4c20a401931d291ba9610566c503b9eaac202`: the prior group-recipient controller-binding blocker is addressed. The resolver now evaluates the recipient filter with `FilterContext::from_ability(ability)`, so `ControllerRef::You` binds to the resolving ability controller instead of the live/moved source object. The new source-left-play and source-controller-changed regressions cover the two failure modes from my previous review.

I am not approving, labeling, or enqueueing yet because the current Rust lint, Rust tests, card-data, and AI gate checks are still in progress on this head. Once this exact head is green and still current, it can go through the normal repush guard for formal approval/enqueue.

matthewevans (approved):
Approved at exact head a6d4c20a401931d291ba9610566c503b9eaac202 after prior controller-binding blocker was addressed and current GitHub checks are green/CLEAN.

PR #4178, opened 2026-06-23
feat(engine): compound dynamic reveal count (Klaw, Sonic Subjugator)

lgray:

## Building block: compound dynamic reveal count

This PR fixes **Klaw, Sonic Subjugator** — "Sonic Attack — When Klaw enters, target player reveals a number of cards from their hand equal to **one plus the number of creature cards in your graveyard**. You choose one of them. That player discards that card." The dynamic reveal count was being dropped (`Effect::RevealHand.count = null`; `supported: false`, `gap_count: 1`). The whole reveal → choose → discard structure already parsed — only the count was swallowed.

**No new engine variant** (the `add-engine-variant` gate returned EXISTS_AS_PARAMETER). The fix is a single-site parser change.

### Root cause + fix

In `oracle_effect/imperative.rs` (`parse_hand_reveal_ast`), the "reveals a number of cards ... equal to X" reveal-count parser delegated to `parse_quantity_ref`, which only recognizes a **bare** `QuantityRef` and returns `None` for a compound quantity — silently dropping the count. The fix delegates to `parse_cda_quantity` instead — the offset/arithmetic-aware entry that returns a full `QuantityExpr` and is a **strict superset** of `parse_quantity_ref` (its final arm wraps `parse_quantity_ref` in `Ref`, so every bare-ref form still parses) — and stores the `QuantityExpr` directly (no `Ref` wrapper).

"one plus the number of creature cards in your graveyard" now lowers to:

```
QuantityExpr::Offset {
  offset: 1,
  inner: Ref(QuantityRef::ZoneCardCount {
    zone: Graveyard, card_types: [Creature], scope: Controller,
  }),
}
```

reusing the existing `QuantityExpr::Offset` variant (resolver `game/quantity.rs`: `Offset { inner, offset } => recurse(inner) + offset`) and `RevealHand.count` (already `Option<QuantityExpr>`; consumed at `reveal_hand.rs` with a `.max(0)` negative clamp). No engine, type, or runtime change.

### Class this generalizes

The fix covers the whole "reveals a number of cards ... equal to `<compound dynamic quantity>`" class — offset ("N plus/minus the number of …"), fraction, multiply, binary arithmetic — plus every bare-ref form the old call handled. Klaw is the driver; future reveal-count cards with compound quantities are covered with no new code.

- **CR 107.1** — the only numbers Magic uses are integers (the reveal count is integer arithmetic).
- **CR 107.1b** — a calculation that yields a negative result uses zero instead (the `.max(0)` clamp at resolution).
- **CR 701.20a** — definition of "reveal a card".

### Tests (non-vacuous + discriminating, verified via revert-probe)

- Card-synthesis through the full `parse_effect_chain` on Klaw's oracle → `RevealHand.count == Some(Offset{1, Ref(ZoneCardCount{Graveyard,[Creature],Controller})})`; the `ChooseFromZone{count:1}` / `DiscardCard` sub-chain intact. Fail-before: `count == None`.
- Bare-ref subsumption: "equal to the number of creature cards in your graveyard" still parses to `Ref(ZoneCardCount)`, not `Offset` (proves no regression on the single-ref case).
- Combinator plus/minus offset shape.
- Runtime: 2 creature cards in the controller's graveyard + a 5-card hand → exactly **3** cards revealed. Discriminating: dropping `+offset` reveals 2, disabling the truncate reveals 5 — so the count provably drives runtime behavior end-to-end.

`mtgish/` untouched. No new serialized surface.

### DIFF
--- a/crates/engine/src/game/effects/reveal_hand.rs
+++ b/crates/engine/src/game/effects/reveal_hand.rs
@@ -304,6 +304,94 @@ mod tests {
         assert_eq!(state.revealed_cards.len(), 1);
     }
 
+    /// CR 701.20a + CR 107.1: A compound `Offset` reveal count (Klaw — "one plus
+    /// the number of creature cards in your graveyard") resolves to inner+1 and
+    /// truncates the revealed set to exactly that many cards. The controller
+    /// (PlayerId(0)) has 2 creature cards in graveyard, so the count resolves to
+    /// 2 + 1 = 3; the 5-card hand must be truncated to 3 revealed cards. Before
+    /// the parser fix this count was dropped (None) at synthesis, so the whole
+    /// hand would have been revealed — this asserts the Offset both resolves and
+    /// is consumed by the resolver.
+    #[test]
+    fn reveal_hand_offset_count_truncates_to_inner_plus_one() {
+        use crate::types::ability::{CountScope, QuantityExpr, QuantityRef, TypeFilter, ZoneRef};
+        use crate::types::card_type::CoreType;
+
+        let mut state = GameState::new_two_player(42);
+
+        // Controller (PlayerId(0)) graveyard: 2 creature cards.
+        for cid in 10..12 {
+            let gy = create_object(
+                &mut state,
+                CardId(cid),
+                PlayerId(0),
+                "Dead Bear".to_string(),
+                Zone::Graveyard,
+            );
+            state
+                .objects
+                .get_mut(&gy)
+                .unwrap()
+                .card_types
+                .core_types
+                .push(CoreType::Creature);
+        }
+
+        // Target (PlayerId(1)) hand: 5 cards (M > N+1 = 3).
+        for cid in 1..6 {
+            create_object(
+                &mut state,
+                CardId(cid),
+                PlayerId(1),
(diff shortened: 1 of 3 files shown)


matthewevans:

I reviewed the current head for the compound dynamic reveal-count change. The fix is at the existing hand-reveal count seam, reuses the existing expression-level `parse_cda_quantity` entry instead of adding a new quantity shape, and the runtime path already consumes `QuantityExpr` via `RevealHand.count`. The tests cover the full Klaw parse path, bare-ref subsumption, the quantity expression building block, and a resolver case where the revealed choice set is truncated to the resolved offset count. I also verified the cited CR anchors locally (`CR 107.1`, `CR 107.1b`, `CR 701.20a`). I did not find a code-level blocker.

No formal approval, label, or enqueue yet because the branch is BEHIND and Rust tests are still running. I’m bringing the branch current and will wait for a current green/CLEAN head before approval/label/enqueue.

matthewevans (approved):
Approving exact head `f41d050304a645c32319779672e6fc054f547692` after repush guard. The compound dynamic reveal-count change reviewed clean, the branch is current, and GitHub checks are green/CLEAN on this head.

PR #4182, opened 2026-06-23
feat(engine): Plan card type + disjunctive control condition (Doctor Doom)

lgray:

## Building blocks: `Plan` card type + elided-verb disjunctive control condition

**Doctor Doom** was resolver-flagged (`gap_count=0` but `supported=false`): every ability parsed, but its static "As long as you control an artifact creature or a Plan, Doctor Doom has indestructible" lowered to `StaticCondition::Unrecognized`. The layer system evaluates an unrecognized condition as `true`, so the indestructible grant was **always-on** instead of gated on board state. Two root causes, both fixed.

### 1. New `CoreType::Plan` (+ `TypeFilter::Plan`)

"Plan" is a card type from Marvel's Spider-Man. It was absent from `CoreType`, so `CoreType::from_str("Plan")` errored and the card-data importer (`from_str().ok()` + `filter_map`) **silently dropped** it. Added `CoreType::Plan` modeled structurally on `CoreType::Plane` — a nontraditional, **non-permanent** card type (excluded from `is_permanent_type` / `PERMANENT_TYPES`), with `TypeFilter::Plan` mirroring `TypeFilter::Battle` for runtime matching.

CR note: "Plan" is **not** in the CR snapshot (CR 205.2a, dated 2026-04-17, lists artifact..vanguard — Marvel's Spider-Man postdates it), so the variant carries a `needs-manual-verification` annotation rather than a fabricated CR number. All exhaustive matches were updated by a compiler-driven sweep (no wildcards), including explicit `CoreType ↔ TypeFilter` Plan bridge arms in `conditions.rs` so Plan no longer falls through to `Subtype("Plan")`.

### 2. Elided-verb disjunctive control condition

"you control `<type A>` or `<type B>`" shares one "you control" verb across two type filters. The second filter ("a Plan") is not a standalone condition, so the top-level condition disjunction can't split it, and `parse_type_phrase` only consumed the first type — leaving the whole condition `Unrecognized`. `parse_you_control_a` now folds an article-led `" or <article> <type>"` continuation into a disjunction:

```
IsPresent { Or[ Typed{[Artifact,Creature], You, Battlefield},
                Typed{[Plan],              You, Battlefield} ] }
```

via the existing `inject_controller` `Or`-distribution. This is scoped to the **control-condition layer** (not `parse_type_phrase`, which would have regressed conjunctive "and" and recipient "it's X or Y" shapes — measured: that approach fails 4 existing tests), and covers the whole "you control A or B" class. **No new `StaticCondition` variant** — the runtime `IsPresent`/`Or` evaluator already gates the grant (CR 611.3a, continuous re-evaluation), so no runtime evaluator change was needed.

### CR references (grep-verified)

- **CR 109.5** — "you"/"your" = controller.
- **CR 205.2a** — card types list (confirms "Plan" is absent → `needs-manual-verification`).
- **CR 604.1** — handling static abilities.
- **CR 608.2c** — read the whole text / disjunction.
- **CR 611.3a** — a static ability's continuous effect isn't locked in; re-evaluated each moment (why the conditional grant toggles).
- **CR 702.12b** — indestructible.

### Tests (non-vacuous + discriminating; reviewer-run)

- Runtime toggle (Anger pattern): Doctor Doom HAS indestructible while you control an artifact creature; does NOT when you control neither an artifact creature nor a Plan (the **revert-probe** — pre-fix `Unrecognized => true` made the grant always-on, so this arm fails before the fix); does NOT when only an opponent controls an artifact creature (proves `ControllerRef::You` scope).
- Condition parse asserts typed `IsPresent{Or[...]}`, not `Unrecognized`; single-type "you control an artifact creature" still parses (regression).
- `CoreType::Plan`: `from_str`/`Display`/serde round-trip; non-permanent classification matches `Plane`; `protection_quality_str == None`.
- `TypeFilter::Plan` runtime match (Plan object matches, creature does not); the `CoreType ↔ TypeFilter` Plan bridge round-trips (not `Subtype`).

Full engine test suite green (13346 passed); clippy `-D warnings` clean (engine + draft-core); parser-combinator gate exit 0; zero non-exhaustive-match errors. `mtgish/` untouched. New serde variants round-trip-tested; 0 Plan cards in the current corpus so no migration.


matthewevans (changes requested):
**[HIGH]** The new Plan type word is parsed as a prefix, so existing Plant/Planet subtype phrases now become `TypeFilter::Plan`. Evidence: `crates/engine/src/parser/oracle_nom/target.rs:208` documents that `plan`/`plans` are matched with no word-boundary guard, and `crates/engine/src/parser/oracle_nom/target.rs:235` returns as soon as `input.strip_prefix(word)` succeeds. GitHub's coverage parse diff confirms the production fallout: existing Plant/Planet cards such as A-Phylath, Phylath, Kirri, Volatile Orbit, Drill Too Deep, Insidious Roots, and Avenger of Zendikar changed from Plant/Planet filters to `plan`. Why it matters: this does not just add Doctor Doom support; it silently rewrites unrelated existing card classes from creature/land subtypes to the new nontraditional Plan card type, causing wrong targeting/conditions and coverage churn. Suggested fix: add a real word-boundary guard for core type words before returning from `parse_type_filter_word`, and add explicit negative regression tests for `Plant`, `Plants`, `Planet`, and `Power-Plant`/`Power Plant` phrases so subtype fallback remains authoritative.

No approval, label, or enqueue from the sweep.

lgray:

Thanks @matthewevans — fixed in `d4a21fdd1` (cherry-picked on top of your "Update branch" merge `512d194c9`, no force-push, so your merge is preserved).

## Finding ([HIGH] regression — Plan shadows Plant/Planet/Plane) — FIXED

Confirmed exactly as reported. This commit added `("plans"/"plan", TypeFilter::Plan)` to the `TYPE_WORDS` table in `parse_type_filter_word` (`oracle_nom/target.rs`), but the scan did a bare `strip_prefix` with **no word-boundary guard** and returned on the first match. So `"plant"` stripped `"plan"` (leaving `"t"`) and returned `TypeFilter::Plan`, shadowing the Plant/Planet/Plane subtypes and flipping cards like A-Phylath, Phylath, Kirri, Volatile Orbit, Drill Too Deep, Insidious Roots, Avenger of Zendikar.

Fix: a word-boundary guard on the scan, applied to **all** entries (the flaw was general — no head-noun type word should match as a sub-word of a longer word):
```rust
if let Some(rest) = input.strip_prefix(word) {
    if rest.is_empty() || rest.starts_with(|c: char| !c.is_alphanumeric()) {
        return Ok((rest, tf.clone()));
    }
}
```
This mirrors the existing boundary checks in `parse_outlaw_type` and `parse_subtype_entry` (`oracle_util.rs`). `parse_subtype` is already boundary-guarded via `starts_with_word_ci`, so once the prefix scan stops shadowing, Plant/Planet/Plane/Spellshaper/Lander resolve correctly as subtypes.

## Second regression the guard exposed — also fixed

Adding the guard surfaced that `"battle"` was the **only** `TYPE_WORDS` entry missing its plural sibling: `BATTLE_SUBTYPES` has no `"Battle"` (only `"Siege"`), so once the guard rejected `"battle"`+`"s"`, `"battles"` fell through to `parse_subtype` and returned `Err`. `"battles"` is an intentionally-supported head noun (recognized in `oracle_static/grammar.rs`, `restriction.rs`, `type_change.rs`). Added `("battles", TypeFilter::Battle)` before `("battle", ...)` so every singular entry now has its plural.

## Tests (11 boundary tests; the negatives fail on the pre-fix code)

`plant`/`plants` → `Subtype("Plant")`; `planet` → `Subtype("Planet")`; `spellshaper` → `Subtype` not `Card`; `lander` → `Subtype` not `Land`; `power-plant`/`power plant` → never `Plan`; `battles` → `Battle` (+ trailing-context variant); and the feature guards `plan` / `plan you control` → still `Plan`.

Verification (re-run on this base, which merged an older main): `cargo fmt --all`, parser-combinator gate exit 0, the 11 boundary tests green, full `parse_type_phrase` suite green (incl. the Doctor Doom `parse_type_phrase_recognizes_plan` positive), `cargo clippy -p engine --all-targets` clean. The PR is slightly behind current upstream (your merge brought in an older main) — the target.rs fix is base-independent.

matthewevans (changes requested):
**[HIGH]** Plan is implemented as a new card type instead of the Plan enchantment subtype. Evidence: `crates/engine/src/types/card_type.rs:83` adds `CoreType::Plan`, `crates/engine/src/types/ability.rs:2378` adds `TypeFilter::Plan`, and `crates/engine/src/parser/oracle_nom/target.rs:214` parses the head noun `plan` directly to that card-type filter. Wizards' Marvel Super Heroes mechanics article describes Plan as a new enchantment subtype, not a card type: https://magic.wizards.com/en/news/feature/marvel-super-heroes-mechanics . Why it matters: this makes Doctor Doom and every future Plan reference look for an object whose core card type is `Plan`, while real Plan cards are enchantments with subtype `Plan`; the engine will miss real Plan permanents and the new core-type plumbing will leak into unrelated casting/filter/coverage surfaces. Suggested fix: model Plan through the existing subtype path (for example `TypeFilter::Subtype("Plan")` on enchantment cards), add Plan to the subtype data/parser support, and keep the word-boundary guard so Plant/Planet/Power-Plant remain subtype fallbacks.

The word-boundary fix from the prior review is directionally correct, but it needs to sit on the subtype model rather than on a new `CoreType::Plan` / `TypeFilter::Plan` axis.

No approval, label, or enqueue from the sweep.

matthewevans:
Maintainer handling (pushed to `card/msh-doctor-doom`):

**Reworked Plan from a new card type to an enchantment subtype** (the second review finding). "Plan" is an enchantment subtype — real cards read `Enchantment — Plan` (Claim the Kingdom, Glorious Purpose, Doom Reigns Supreme, …) — not a nontraditional card type. The original `CoreType::Plan` was also modeled non-permanent (like `Plane`), which is doubly wrong since Plan permanents are enchantments.

Changes:
- **Removed the `CoreType::Plan` / `TypeFilter::Plan` axis entirely** (card_type.rs, ability.rs, filter.rs ×4 arms, coverage.rs, token.rs, trigger_index.rs, oracle_keyword.rs, draft-core/cube.rs). Verified no production reference remains.
- **Added `"Plan"` to `ENCHANTMENT_SUBTYPES`** (CR 205.3h) so "a Plan" resolves through the existing `Subtype("Plan")` path — this unlocks the whole `Enchantment — Plan` class, not just Doctor Doom.
- **Parser (target.rs):** dropped the Plan TYPE_WORDS entries; kept your word-boundary guard and the `"battles"` plural entry (both correct independent fixes). "Plan" now falls through to the boundary-guarded subtype table.
- **Kept the disjunctive control condition** ("you control X or a Y") — the genuinely reusable building block — and retargeted its tests + the Doctor Doom layer tests to `Subtype("Plan")`.

Doctor Doom's "you control an artifact creature or a Plan" now lowers to a typed `IsPresent { Or[ Typed{[Artifact,Creature],You}, Typed{[Subtype("Plan")],You} ] }`.

**Test discrimination:** added an end-to-end test (`doctor_doom_indestructible_parsed_end_to_end_with_plan_enchantment`) that *parses* Doctor Doom's printed line — exercising `parse_type_filter_word` + `ENCHANTMENT_SUBTYPES` so "a Plan" → `Subtype("Plan")` — attaches the parser-produced static, and evaluates layers against a real `Enchantment — Plan` permanent. It fails on revert of the rework. (The review correctly flagged that the existing layer tests hand-build the condition, so I also corrected their comment to describe them as runtime filter-match coverage rather than parser-revert probes.)

**Verification:** `cargo fmt`, parser-combinator gate, Doctor Doom suite (6 tests incl. the end-to-end + negative opponent/empty cases), the `parse_type_filter_word` boundary family (plant/planet/spellshaper/lander/battles), and `clippy -p engine --lib` all green against current `origin/main` (brought current; merged cleanly). An isolated `review-impl` pass confirmed the axis is fully removed, "a Plan" → `Subtype("Plan")` is proven by discriminating parser tests, and the disjunctive condition serves the class.

CR 205.3h + CR 604.1 + CR 702.12b.

matthewevans (approved):
VERDICT: approve

Re-reviewed the current head after the maintainer Plan-subtype fix. The prior blockers are addressed: `Plan` no longer exists as a core card type/filter, bare `plan` lowers through the subtype table to `Subtype("Plan")`, the Plant/Planet/Power-Plant boundary regressions are pinned, and the Doctor Doom path has parser-to-layer runtime coverage for a real `Enchantment — Plan` permanent. Existing checks are green/CLEAN on the reviewed head.

matthewevans (approved):
Plan modeled as enchantment subtype (CR 205.3h) via existing Subtype path; disjunctive control condition retained; full CI green incl. Rust tests; review-impl clean.

PR #4183, opened 2026-06-23
Fix/tauri desktop session persistence

### LINKED ISSUE #2926: Desktop app requires signing in to cloud on every launch (session not persisted, unlike PWA/browser)
## Expected vs actual
- Expected: The desktop (Tauri) app persists the cloud sign-in session across launches, like the PWA and browser do.
- Actual: The desktop app requires signing in to the cloud on every launch.

Auth/session-persistence issue specific to the desktop build (the web/PWA path persists the session correctly). Likely the desktop webview is not persisting the auth token/cookie storage between launches.

## Source
https://discord.com/channels/1485498006781427802/1514016008594260128/1514016008594260128

jsdevninja:
## Summary
- Desktop (Tauri/WebView2) sign-in sessions weren't surviving app restarts, unlike the PWA/browser — root cause is WebView2's default user-data-folder selection, which silently degrades to a non-persistent temp profile when the install path isn't writable.
- Defer the main window's creation to the `setup` hook and explicitly pin `data_directory` to the app's per-user local data dir, so the webview's storage location no longer depends on where the app is installed.
- Added a regression test pinning the `tauri.conf.json` window shape `lib.rs`'s `setup` hook depends on (`windows[0]` = "main", `create: false`), since `client/src-tauri` is excluded from the main Rust workspace/Tilt's continuous checks.

## Test plan
- [x] `cargo check` / `cargo clippy --all-targets -- -D warnings` / `cargo fmt -- --check` / `cargo test` all pass in `client/src-tauri` (run directly — this crate isn't covered by Tilt)
- [ ] Manual: install the desktop build to a machine-wide (admin) location, sign in to cloud sync, fully quit and relaunch — session should still be signed in
- [ ] Manual: sanity-check sign-in still works on macOS/Linux desktop builds (no-op path)

Fixes #2926

### DIFF
--- a/client/src-tauri/gen/schemas/windows-schema.json
+++ b/client/src-tauri/gen/schemas/windows-schema.json
@@ -1,0 +1,2696 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "CapabilityFile",
+  "description": "Capability formats accepted in a capability file.",
+  "anyOf": [
+    {
+      "description": "A single capability.",
+      "allOf": [
+        {
+          "$ref": "#/definitions/Capability"
+        }
+      ]
+    },
+    {
+      "description": "A list of capabilities.",
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/Capability"
+      }
+    },
+    {
+      "description": "A list of capabilities.",
+      "type": "object",
+      "required": [
+        "capabilities"
+      ],
+      "properties": {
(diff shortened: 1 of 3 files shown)


matthewevans:
DEFER-FE: this touches the client/Tauri desktop surface, so I am leaving it for Matt's direct handling under the frontend/client/UI defer rule. No sweep approval, label, enqueue, or merge action.

PR #4185, opened 2026-06-23
fix(engine): unattach Fortification under the same SBA as Equipment (CR 704.5n)

### LINKED ISSUE #1368: Equipment persists on field — Equipment persists on the field even when destroyed
Reported in Discord: https://discord.com/channels/1485498006781427802/1509610798581026887/1509610798581026887

**Thread:** Equipment persists on field
**Cards:** _none detected_

## Summary
Equipment persists on the field even when destroyed

---
<sub>report_id: `discord:1509610798581026887:1509610798581026887:0` · discord: `1509610798581026887/1509610798581026887`</sub>

jsdevninja:
## Summary
- `check_unattached_equipment` (CR 704.5n) only matched the `"Equipment"` subtype, so a Fortification whose land host left the battlefield (destroyed, sacrificed, bounced) never got unattached by any SBA pass — its `attached_to` stayed pointed at a dead object forever, and the frontend's `partitionByType` then drops any attachment-kind permanent with a non-null `attached_to`, so it silently disappears from the board.
- CR 704.5n explicitly groups Equipment and Fortification together ("If an Equipment or Fortification is attached to an illegal permanent or to a player, it becomes unattached..."), and CR 301.6 makes Fortification/land the direct analog of Equipment/creature — so this broadens the existing SBA check rather than adding new machinery.
- Added a 9-test regression suite mirroring every existing Equipment SBA test for Fortification (host leaves battlefield, host gains protection, attached-to-player, legal-host stays attached, no-attachment stays, phased-out is skipped, and an Equipment+Fortification same-pass independence check).

## Test plan
- [x] `cargo test -p engine --lib game::sba::` — 78/78 passed
- [x] `cargo test -p engine --lib game::effects::attach::` / `game::zones::` — all passed
- [x] `cargo clippy -p engine --lib -- -D warnings` — clean
- [x] `cargo fmt --all`

Fixes #1368

### DIFF
--- a/crates/engine/src/game/sba.rs
+++ b/crates/engine/src/game/sba.rs
@@ -1457,37 +1457,35 @@ fn check_token_cease_to_exist(state: &mut GameState, any_performed: &mut bool) {
     }
 }
 
-/// CR 301.5 / CR 301.6: The permanent core type a non-Aura attacher's
-/// subtype structurally requires its host to have, if any. An Equipment
-/// "can't legally be attached to anything that isn't a creature" (CR
-/// 301.5c); a Fortification "can't legally be attached to an object that
-/// isn't a land" (CR 301.6, applying CR 301.5c by analogy). Unlike an Aura's
-/// per-card `Keyword::Enchant` filter, this requirement is fixed by the
-/// subtype itself — every Equipment requires a creature host and every
-/// Fortification requires a land host, with no Oracle-text exception to
-/// either. A card with neither subtype (or only the Aura subtype, whose
-/// requirement is carried by `Keyword::Enchant` instead) has no requirement
-/// here.
-fn required_attachment_host_core_type(
+/// CR 301.5 / CR 301.6: The permanent core type(s) a non-Aura attacher's
+/// subtypes structurally require its host to have. An Equipment "can't
+/// legally be attached to anything that isn't a creature" (CR 301.5c); a
+/// Fortification "can't legally be attached to an object that isn't a land"
+/// (CR 301.6, applying CR 301.5c by analogy). Unlike an Aura's per-card
+/// `Keyword::Enchant` filter, this requirement is fixed by the subtype
+/// itself — every Equipment requires a creature host and every Fortification
+/// requires a land host, with no Oracle-text exception to either.
+///
+/// Each matching subtype contributes its own requirement independently —
+/// not a single either/or choice — so a (no current Oracle precedent, but
+/// rule-text-legal) card with both subtypes requires a host that is BOTH a
+/// creature AND a land (e.g. an animated land-creature), per CR 301.5c +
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
VERDICT: changes requested

Reviewed exact head `8254de198262b9d764b4e66879394d81a1b96f0f` via the GitHub API diff.

This is the right general SBA seam, but the current fix still misses Fortification's host-class legality. `check_unattached_equipment` now calls `is_valid_attachment_target` for both Equipment and Fortification, but that helper's non-Enchant branch currently returns `target.zone == Zone::Battlefield` for every non-Enchant attachment. Normal Fortify activation targets a land, but the continuous SBA legality check still treats a Fortification attached to any battlefield object, such as a creature, as legal.

The regression suite covers land gone, protection, player host, legal land host, no attachment, phased out, and same-pass Equipment/Fortification. It does not cover the CR 301.6 axis: a Fortification attached to a nonland permanent must become unattached under CR 704.5n. Please add that legality branch/test rather than relying on the activation target filter; SBAs need to repair illegal attached state regardless of how it was produced.

matthewevans (changes requested):
**[LOW]** The new host-type helper only enforces one structural requirement when an attachment has both subtypes. Evidence: `crates/engine/src/game/sba.rs` adds `required_attachment_host_core_type`, but it returns `Some(CoreType::Creature)` for any Equipment and only checks Fortification in an `else if`; the same diff explicitly notes a card can carry both Equipment and Fortification, and CR 301.5c + CR 301.6 mean that object would need a host that is both creature and land. Why it matters: the helper is now the continuous SBA authority for non-Aura attachment legality, so an Equipment-Fortification attached to a nonland creature would remain attached even though the Fortification rule makes that host illegal. Suggested fix: make the helper return all required host core types, or otherwise require both Creature and Land when both subtypes are present, with a small regression for the dual-subtype case.

matthewevans:
The prior Fortification host-class blocker is addressed on head `bbbbcc57ab06d44da0ffa50917c67f5eafa5e420`: the helper now enforces all required host core types for Equipment/Fortification and includes dual-subtype regressions. This head is still BEHIND, so I am updating the branch and holding approval/label/enqueue until the refreshed head is current and checks are green.

matthewevans (approved):
Approved current head  after the Fortification SBA follow-up and update-branch CI.

The previous blocker is addressed: non-Aura attachment legality now requires every structural host core type contributed by Equipment/Fortification subtypes, so Fortification requires a land host and a dual Equipment+Fortification object requires a host that is both creature and land. The added regressions cover nonland Fortification hosts and the dual-subtype creature vs land-creature cases. GitHub checks are green on this refreshed head, and the repush guard held.

PR #4188, opened 2026-06-23
fix(engine): dual-target fight uses both chosen creatures (#3862)

### LINKED ISSUE #3862: Ulvenwald Tracker only directly fights another creature, instead of having any 2 creatures fight. — When [[Ulvenwald Tr…
Reported in Discord: https://discord.com/channels/1485498006781427802/1516959413313994852/1516959413313994852

**Thread:** Ulvenwald Tracker only directly fights another creature, instead of having any 2 creatures fight.
**Cards:** ulvenwald tracker, wayta, trainer prodigy

## Summary
When [[Ulvenwald Tracker]] is activated, you can choose just one creature, and the Tracker fights it, instead of choosing a second creature to fight the first chosen.

### ulvenwald tracker
{1}{G}, {T}: Target creature you control fights another target creature.
### wayta, trainer prodigy
Haste
{2}{G}, {T}: Target creature you control fights another target creature. This ability costs {2} less to activate if it targets two creatures you control.
If a creature you control being dealt damage causes a triggered ability of a permanent you control to trigger, that ability triggers an additional time.

---
<sub>report_id: `discord:1516959413313994852:1516959413313994852:0` · discord: `1516959413313994852/1516959413313994852`</sub>

kiannidev:
## Summary
- Fixes Ulvenwald Tracker / Wayta-style fight activations where only one creature was chosen and the ability source fought instead of the two selected creatures.
- Surfaces two target slots when a fight names both fighters ("target creature you control fights another target creature").
- Resolves fight damage between both chosen object targets at runtime.

## Test plan
- [x] `cargo test -p engine --lib dual_target_fight`
- [x] `cargo test -p engine --test integration issue_3862_ulvenwald`

Fixes #3862

Tier: Standard
Model: claude-4.6-sonnet-medium-thinking


matthewevans (changes requested):
There is a blocker in the resolution-time target revalidation path for this new dual-target fight shape.

`collect_target_slots`, `build_target_slot_specs`, and assignment now correctly treat `Effect::Fight { subject, target }` as two per-slot filters when the subject itself is targeted. But `validate_targets_in_chain` still uses the generic `extract_target_filter_from_effect` path for `Effect::Fight`, and that extractor returns only `target` (`Effect::Fight { target, .. }`). On resolution, CR 608.2b validation therefore drops illegal entries without preserving which fight slot they came from. If exactly one of the two chosen fighters remains legal, `fight::resolve` sees `object_targets.len() < 2` and falls back to `resolve_fight_subject(state, ability)?` plus the one remaining object target. For Ulvenwald Tracker / similar cards, that can make the activated source fight the surviving selected target instead of doing no fight damage when one of the two chosen fighters is illegal.

Please make fight resolution validation mirror the per-slot fight handling you added for selection/assignment. The dual-target form needs to preserve the two-fighter contract through CR 608.2b, so a single remaining selected fighter must not be reinterpreted as the fallback single-target `~ fights target creature` form. A regression where one selected fighter leaves the battlefield before resolution and the Tracker itself remains on battlefield would catch this.

matthewevans (approved):
Approved current head 76e7e4aa3a18dc41932b0c75d0153374ec4a9da0 after re-review and green exact-head checks. Repush guard held.

PR #4190, opened 2026-06-23
fix(engine): Adventure instant faces castable at instant speed (#4001)

### LINKED ISSUE #4001: Frolicking Familiar — Unable to cast Blow Off Steam (Adventure) as an instant.
Reported in Discord: https://discord.com/channels/1485498006781427802/1517663965860200598/1517663965860200598

**Thread:** Frolicking Familiar
**Cards:** blow off steam

## Summary
Unable to cast Blow Off Steam (Adventure) as an instant.

### blow off steam
Blow Off Steam deals 1 damage to any target. (Then exile this card. You may cast the creature later from exile.)

---
<sub>report_id: `discord:1517663965860200598:1517663965860200598:0` · discord: `1517663965860200598/1517663965860200598`</sub>

kiannidev:
## Summary
- Fix `can_cast_object_now` to evaluate the Adventure instant/sorcery face when `prepare_spell_cast` fails on the creature face's sorcery-speed timing gate (e.g. outside main phases).
- Resolves Frolicking Familiar // Blow Off Steam not appearing as a castable spell during combat.

## Root cause
`can_cast_object_now` called `prepare_spell_cast` on the creature face first. Outside main phases that fails the sorcery-speed timing check, and the error path only consulted alternative *casting variants* (Flashback, Warp, etc.) — not the Adventure spell face. The existing adventure-face recursion in `can_cast_prepared_now` never ran because preparation never succeeded.

## Test plan
- [x] `cargo test -p engine issue_4001_adventure_instant_castable_outside_main_phase`
- [x] Unit test in `casting.rs` (Bonecrusher Giant // Stomp during declare attackers)
- [x] Integration test for Frolicking Familiar // Blow Off Steam pattern

Fixes #4001


matthewevans (changes requested):
There is a blocker in the new `can_cast_object_now` fallback.

The fallback swaps to the Adventure/Omen spell face whenever normal `prepare_spell_cast` fails and `alternative_spell_layout(obj).is_some()`, but it does not also require `cast_face_choice_offered_from_zone(state, obj)`. That means it can run for an Adventure card in exile after the Adventure has resolved. In that zone, the engine should only allow the normal permanent face through the `AdventureCreature` permission; `handle_cast_spell` will not offer `CastOfferKind::Adventure` from exile because `cast_face_choice_offered_from_zone` is false. If the exiled creature face fails normal sorcery timing outside main phase, this new fallback can still swap to the instant Adventure face in the simulation and return `true`, so legal-actions/UI can advertise a cast that the real action path cannot perform.

Please gate the fallback to the same hand/command-zone face-choice surface used by `handle_cast_spell`, and add a negative regression for an exiled Adventure creature with an instant Adventure face during combat: `can_cast_object_now` should be false outside main phase even though the back face is instant-speed castable from hand.

matthewevans (changes requested):
Same blocker still applies on the force-pushed head.

The new `can_cast_object_now` fallback still swaps to the Adventure/Omen spell face whenever `alternative_spell_layout(obj).is_some()` after normal `prepare_spell_cast` fails. It still does not also require the same `cast_face_choice_offered_from_zone(state, obj)` gate used by `handle_cast_spell` for the actual cast-offer surface.

That leaves the legal-actions simulation broader than the real cast path: an Adventure card exiled after resolving its Adventure can still simulate the instant Adventure face outside main phase even though `handle_cast_spell` will not offer `CastOfferKind::Adventure` from exile. The added tests only cover the positive hand case, so they do not catch that mismatch.

Please gate this fallback to the same hand/command-zone face-choice surface and add the requested exiled-Adventure negative regression: exiled Adventure creature, Adventure face is instant-speed/affordable, combat phase, `can_cast_object_now` must be false.

matthewevans (changes requested):
Same blocker still applies on head `99798098d68ebf01d81e2c9a46059d02f1063af7`.

The fallback in `can_cast_object_now` still runs on any object with `alternative_spell_layout(obj).is_some()` after `prepare_spell_cast` fails, without checking `cast_face_choice_offered_from_zone(state, obj)` first. Evidence: `crates/engine/src/game/casting.rs:9988`-`10000` on this head. The actual cast-offer path gates Adventure/Omen face choices with `cast_face_choice_offered_from_zone` before emitting `CastOfferKind::Adventure`, so legal-actions can still simulate an Adventure instant face from exile even though the real cast path will not offer that face there.

The added regression remains positive-only: `crates/engine/tests/integration/issue_4001_frolicking_familiar_adventure_instant.rs:83`-`119` covers a hand card during declare attackers, but there is still no exiled-Adventure negative case. Please gate the fallback to the same face-choice surface and add the requested regression: exiled Adventure creature with `AdventureCreature` permission, instant Adventure face affordable during combat, `can_cast_object_now` must be false.

matthewevans (approved):
Approved current head 3810370ea1f30f415f720532346ede3947b82325 after re-review and green exact-head checks. Repush guard held.

PR #4191, opened 2026-06-23
Fixes #3258: Show and Tell optional per-player put

### LINKED ISSUE #3258: show and tell — Forces you to put a permanent into play, instead treating it as a may ability.
Reported in Discord: https://discord.com/channels/1485498006781427802/1514681426069688491/1514681426069688491

**Thread:** show and tell
**Cards:** _none detected_
**Parser status:** no_card

## Summary
Forces you to put a permanent into play, instead treating it as a may ability.

---
<sub>report_id: `discord:1514681426069688491:1514681426069688491:0` · discord: `1514681426069688491/1514681426069688491`</sub>

kiannidev:
## Summary
- Add integration regression coverage for Show and Tell (`Each player may put an artifact, creature, enchantment, or land card from their hand onto the battlefield.`).
- Verify parser output sets `optional: true` with `player_scope: All` and `Effect::ChangeZone` from hand.
- Verify runtime: each player gets `OptionalEffectChoice` and may decline without putting a permanent; accepting still routes through `EffectZoneChoice` for the scoped player's hand.

Upstream already routes this through the existing each-player-may + `player_scope` + `OptionalEffectChoice` pipeline (same shape as Braids / Agitator Ant). This PR locks that behavior in with tests so the Discord report cannot regress silently.

## Test plan
- [x] `cargo test -p engine --test integration issue_3258`
- [x] Decline path: both players decline → hands unchanged (aside from cast spell), battlefield empty
- [x] Accept path: P0 accepts and picks Island, P1 declines → one permanent enters

### DIFF
--- a/crates/engine/src/game/effects/change_zone.rs
+++ b/crates/engine/src/game/effects/change_zone.rs
@@ -394,10 +394,10 @@ pub fn resolve(
         };
         // Filter-controller override is primary here: when a filter like
         // "creature you control" needs "you" to resolve to the *target* player
-        // (not the caster), we pass `filter_controller` explicitly. Use
-        // `from_source_with_controller` to honor this remapping.
-        let ctx = crate::game::filter::FilterContext::from_source_with_controller(
-            ability.source_id,
+        // (not the caster), we pass `filter_controller` explicitly. Include the
+        // resolving ability so `Owned { ScopedPlayer }` reads `scoped_player`.
+        let ctx = crate::game::filter::FilterContext::from_ability_with_controller(
+            ability,
             filter_controller,
         );
         let eligible: Vec<ObjectId> = state
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -3941,7 +3941,17 @@ fn filter_uses_relative_controller_you(filter: &TargetFilter) -> bool {
 /// ability's controller.
 fn filter_uses_relative_controller_scoped(filter: &TargetFilter) -> bool {
     match filter {
-        TargetFilter::Typed(tf) => tf.controller == Some(ControllerRef::ScopedPlayer),
+        TargetFilter::Typed(tf) => {
+            tf.controller == Some(ControllerRef::ScopedPlayer)
+                || tf.properties.iter().any(|prop| {
+                    matches!(
+                        prop,
+                        FilterProp::Owned {
+                            controller: ControllerRef::ScopedPlayer
+                        }
+                    )
+                })
+        }
         TargetFilter::Or { filters } | TargetFilter::And { filters } => {
             filters.iter().any(filter_uses_relative_controller_scoped)
         }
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
Thanks for keeping this test-only and issue-scoped. I do think this still needs one more discriminating runtime assertion before it pins #3258.

The current accept/decline regression can pass if Show and Tell only offers the optional branch to P0 and then finishes. In `show_and_tell_acceptor_puts_one_card_from_hand`, the loop accepts whenever the prompt player is P0, but it never records that an `OptionalEffectChoice` was presented for P1; the final assertions still pass if P1 was never prompted because P1's hand remains unchanged and only P0's Island moved. The test also only selects from P0's hand, so it does not prove the follow-up `EffectZoneChoice` is scoped to the iterated non-controller player's hand.

For this issue, please add a runtime case that observes both players' optional prompts and has P1 accept using P1's hand, or an equivalent assertion that proves P1 receives its own optional decision and zone choice. For example: P0 declines, P1 accepts, select P1's Plains, then assert the Plains moved to the battlefield while P0's card stayed in hand. That would make the regression fail on the one-player fan-out bug shape this PR is meant to guard.

matthewevans (changes requested):
The parser-side fix is moving in the right area, but the runtime blocker from the previous review is still not addressed on head `32a3665e585ed2b70e9721fc587cea61ea1eaff0`.

The new integration test still only accepts when the prompt player is P0 and never records that P1 received an `OptionalEffectChoice` or a P1-scoped `EffectZoneChoice`. Evidence: `crates/engine/tests/integration/issue_3258_show_and_tell.rs:162`-`177` sets `accept = *player == P0` and selects P0's Island, while no test has P0 decline and P1 accept from P1's hand. That can still pass if the engine only offers Show and Tell to P0 and then finishes, which is exactly the fan-out shape this issue needs to guard.

Please add the discriminating runtime case requested earlier: P0 declines, P1 receives the optional prompt, P1 accepts, the subsequent zone choice contains/selects P1's Plains, then assert Plains moved to the battlefield while P0's card stayed in hand.

matthewevans (changes requested):
Same blocker still applies on head `b45b5b7118a41ed49396c697e3f597b25579d41a`.

The new commit only annotates the `tag` type in `clause_shell`; it does not add the requested non-controller runtime case. The integration test still accepts only when the prompt player is P0 and selects P0's Island. Evidence: `crates/engine/tests/integration/issue_3258_show_and_tell.rs:162`-`177` remains the accept path, with no case where P0 declines and P1 accepts from P1's hand. Why it matters: the regression can still pass if the fan-out stops after P0, which is the exact bug shape this PR is meant to lock down. Suggested fix: add the P0-decline/P1-accept runtime case and assert P1's Plains is offered and moves to the battlefield while P0's card stays in hand.

matthewevans:
Reviewed current head 3a780501992a262a78832721ca38006b16d068e5. The prior Show and Tell blocker appears addressed: the new runtime test proves P0 can decline while P1 receives its own optional prompt and chooses from P1's hand, and the production path now carries the resolving ability into the scoped ownership filter context.

Holding pending exact-head CI: Rust tests are still pending on this head, so I am not approving/labeling/enqueuing from the sweep yet.

matthewevans (approved):
Approved current head 3a780501992a262a78832721ca38006b16d068e5 after re-review and green exact-head checks. Repush guard held.

PR #4194, opened 2026-06-23
fix(engine): prevent Insidious Roots double trigger on graveyard leave

### LINKED ISSUE #3866: Insidious roots trigger twice — Hi, every time I have one or more creature getting out of the graveyard, I have 2 insid…
Reported in Discord: https://discord.com/channels/1485498006781427802/1517136824903008398/1517136824903008398

**Thread:** Insidious roots trigger twice
**Cards:** insidious roots

## Summary
Hi, every time I have one or more creature getting out of the graveyard, I have 2 insidious roots triggers.

### insidious roots
Creature tokens you control have "{T}: Add one mana of any color."
Whenever one or more creature cards leave your graveyard, create a 0/1 green Plant creature token, then put a +1/+1 counter on each Plant you control.

---
<sub>report_id: `discord:1517136824903008398:1517136824903008398:0` · discord: `1517136824903008398/1517136824903008398`</sub>

kiannidev:
## Summary
- Add `batched_zone_change_trigger_fired` ledger keyed by `(source, trig_idx, moved object, from, to)` so batched `ChangesZone` / `ChangesZoneAll` triggers fire once per zone-change batch even when `process_triggers` scans the same events twice.
- Clear the ledger on turn rollover and AI turn-scoped reset.
- Add integration coverage for single/double graveyard leaves, replayed event batches, and Reanimate.

Fixes #3866

## Test plan
- [x] `cargo test -p engine --test integration issue_3866`
- [x] `cargo fmt --all`


matthewevans:
This appears to be an exact duplicate of #4195: the API diff hashes match, and #4195 has the issue-linked title/body for #3866. I am reviewing #4195 and leaving this one unapproved/unlabeled/unqueued. Please close one of the duplicates so the review/merge state is not split across two identical PRs.

matthewevans (changes requested):
VERDICT: request-changes

The duplicate split is resolved (#4195 was closed), so I reviewed current #4194 head `ed720a0b6df8885a38b89173f21771956bec0f11` directly. The prior #4195 semantic blocker is still present on this active PR.

**[HIGH]** The replay guard still dedupes by the persistent zone-transition tuple, so it can suppress a later legitimate same-object trigger event. Evidence: `crates/engine/src/types/game_state.rs:6225` stores `HashSet<(ObjectId, usize, ObjectId, Option<Zone>, Zone)>`, `crates/engine/src/game/triggers.rs:441`-`472` checks only `(source_id, trig_idx, moved_object, from, to)`, and `crates/engine/src/game/triggers.rs:475`-`492` records that same tuple. Why it matters: the same creature card can leave the graveyard to the battlefield, return to the graveyard, and leave again in the same turn; the second physical leave has the same object/from/to tuple but is a new CR 603.2c trigger event and should trigger Insidious Roots again. Suggested fix: scope the guard to the exact replay/event-batch occurrence rather than the durable object transition, and add a regression where the same creature leaves the graveyard twice in one turn and produces two Insidious Roots triggers.

The new tests cover re-scanning the identical event vector, but they do not cover a later second physical graveyard leave by the same card.

matthewevans (changes requested):
Same blocker still applies on current head 5b27406f930fc5f5a1dcca9778025083a801a248. The new single-parent commit only switches the test to the committed integration fixture; the production guard is still a persistent `batched_zone_change_trigger_fired` key of `(source_id, trig_idx, object_id, from, to)` in `crates/engine/src/game/triggers.rs`. That still suppresses a second legitimate same-turn leave by the same physical card from graveyard to battlefield after it returned to the graveyard.

Please make the duplicate guard scoped to re-processing the same event batch/pass, not to all identical object/from/to zone changes for the rest of the turn, and add the hostile regression where the same creature card leaves the graveyard twice in one turn as two distinct events.

matthewevans (changes requested):
Thanks for the follow-up. The fixture restoration gets CI green again, but the production replay guard still keys `batched_zone_change_trigger_fired` by `(source_id, trig_idx, moved_object, from, to)` and only clears it on the next turn.

That prevents re-processing the same `ZoneChanged` batch, but it also suppresses a later, distinct same-turn graveyard leave by the same physical card. The engine preserves the same `ObjectId` through `move_to_zone`, so a card that leaves your graveyard, returns to the graveyard, then leaves again in the same turn has the same tuple and the second real event is skipped. The new tests cover duplicate scans of the same event batch, not this second-event case.

Please make the replay guard scoped to the event batch/scan being de-duplicated, or otherwise distinguish distinct zone-change events within the same turn, and add a regression where the same card leaves the graveyard twice in one turn and Insidious Roots triggers for both distinct events.

matthewevans (changes requested):
The event-index direction addresses the prior same-card/same-turn blocker, and the new regression covers that case. This head still cannot proceed because it does not compile on exact-head CI:

- `crates/engine/src/game/derived_views.rs:1064` constructs `ZoneChangeRecord` without `turn_zone_change_index`
- `crates/engine/src/types/game_state.rs:528` does the same in the default/test constructor path

Please update all `ZoneChangeRecord` initializers for the new field and bring the branch current; GitHub currently reports the PR as DIRTY/CONFLICTING, so I can't approve or enqueue it from this head.

matthewevans (approved):
Reviewed current head 5c631aaf7ddcd034e14db2eb4b50c5dc7b6d7ef5 via the GitHub API diff and isolated worktree. The event-index replay guard addresses the previous same-card same-turn Insidious Roots blocker, the requested regression is present, and exact-head CI is green. Repush guard held before approval.

PR #4195, opened 2026-06-23
fix(engine): prevent Insidious Roots double-trigger on graveyard leave (fixes #3866)

### LINKED ISSUE #3866: Insidious roots trigger twice — Hi, every time I have one or more creature getting out of the graveyard, I have 2 insid…

kiannidev:
## Summary
- Insidious Roots (`ChangesZoneAll`, batched, origin Graveyard) was stacking twice when the same `ZoneChanged` events were scanned more than once in a single action (e.g. inline `collect_triggers_into_deferred` plus `run_post_action_pipeline`).
- Add a per-turn ledger keyed by `(source, trigger index, moved object, from, to)` so batched zone-change triggers are collected at most once per physical zone change.
- Integration tests cover single/batched graveyard leaves, double `process_triggers` replay, and a full Reanimate cast path.

## Test plan
- [x] `cargo test -p engine --test integration issue_3866` (5 tests)

Fixes #3866

### DIFF
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -417,6 +417,95 @@ fn partition_lki_trigger_definitions(
     (printed, granted_keywords)
 }
 
+fn batched_zone_change_batch(events: &[GameEvent]) -> bool {
+    !events.is_empty()
+        && events
+            .iter()
+            .all(|event| matches!(event, GameEvent::ZoneChanged { .. }))
+}
+
+fn batched_zone_change_replay_guard_applies(
+    trig_def: &TriggerDefinition,
+    trigger_events: &[GameEvent],
+) -> bool {
+    trig_def.batched
+        && matches!(
+            trig_def.mode,
+            TriggerMode::ChangesZone | TriggerMode::ChangesZoneAll | TriggerMode::Evolve
+        )
+        && batched_zone_change_batch(trigger_events)
+}
+
+fn batched_zone_change_already_collected(
+    state: &GameState,
+    source_id: ObjectId,
+    trig_idx: usize,
+    trigger_events: &[GameEvent],
+) -> bool {
+    let keys: Vec<(ObjectId, Option<Zone>, Zone)> = trigger_events
+        .iter()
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
There is a blocker in the replay ledger on head `6fbeffc3b29ff84fddded20885d12a2728ecca84`.

**[HIGH]** The new dedupe key suppresses a second legitimate same-turn zone change for the same object. Evidence: `crates/engine/src/types/game_state.rs:6208`-`6213` stores only `(source_id, trig_idx, moved_object, from, to)`, and `crates/engine/src/game/triggers.rs:461`-`470` skips a later batch when all of those keys were seen before. Why it matters: `ObjectId` is the storage identity, not a per-zone-change identity, so the same creature can leave the graveyard to the battlefield, return to the graveyard, and leave again in the same turn; the second physical leave has the same `(object, Graveyard, Battlefield)` tuple and will be incorrectly suppressed even though Insidious Roots should trigger again. Suggested fix: key the replay guard to the actual event batch/zone-change record identity, or clear/scope it so only the same `process_triggers` replay is deduped; add a regression where the same creature leaves the graveyard twice in one turn and produces two triggers.

Gemini's allocation note is not the blocker. The semantic issue is that the ledger currently dedupes by a coarse zone-transition tuple rather than by the specific event occurrence being replayed.

kiannidev:
Duplicate of #4194 — closing in favor of the other Insidious Roots fix.

Issue #1654, opened 2026-05-31
[Card Bug]


moodnesss123:
### Card(s)

Morophon, the boundless

### Game mode

AI

### Actual behavior

the WUBRG discount is not working

### Expected behavior

it should reduce WUBRG mana pips

mike-theDude:
## Root cause

The `{W}{U}{B}{R}{G}` reduction never matches any spell because of a chosen-type kind mismatch.

**Morophon's ETB** (`As Morophon enters, choose a creature type.`) stores a `ChosenAttribute::CreatureType` on the permanent — confirmed in `card-data.json`:

```json
{"type":"Choose","choice_type":"CreatureType","persist":true}
```

**But the cost-reduction static** parses its `spell_filter` with `FilterProp::IsChosenCardType`:

```json
"ModifyCost": {
  "mode": "Reduce",
  "amount": { "shards": ["White","Blue","Black","Red","Green"], "generic": 0 },
  "spell_filter": { "type": "Typed", "type_filters": ["Card"],
                    "properties": [{ "type": "IsChosenCardType" }] }
}
```

At cast time, `IsChosenCardType` (`crates/engine/src/game/filter.rs:2099`) only matches a stored `ChosenAttribute::CardType`. Morophon stored a `CreatureType`, so the filter returns `false` for every spell and the reduction is never applied.

## Where it goes wrong

`crates/engine/src/parser/oracle_static/static_helpers.rs:324-343` re-attaches the chosen-type discriminator after parsing the spell base type. The heuristic is:

- `"Creature spells ... of the chosen type"` → `IsChosenCreatureType`
- bare `"Spells ... of the chosen type"` → `IsChosenCardType`

That bare-spells → card-type assumption is correct for Cloud Key / Umori / Stenn (which `choose a card type`), but **Morophon's text is bare `"Spells of the chosen type you cast ..."` while its ETB chooses a *creature* type.** The disambiguator is keying off the spell base-type wording instead of what the card's "choose" clause actually picks.

## Suggested fix direction

The chosen-type discriminator (`IsChosenCreatureType` vs `IsChosenCardType`) should be derived from the kind chosen by the card's "As ~ enters, choose a [creature type|card type]" clause, not inferred from whether the spell base type is "creature". A creature-type-choosing card with a bare-"Spells of the chosen type" reduction (Morophon) must use `IsChosenCreatureType`. CR 205.2a / CR 601.2f.

(Found while triaging this report; analysis only — no code change yet.)

[...]

PR #4197, opened 2026-06-23
fix(engine): link Morophon chosen-type cost reducer to creature-type choice (#1654).

mike-theDude:
A creature-type chooser's bare "Spells of the chosen type" cost reducer
(Morophon, the Boundless) was parsed with IsChosenCardType, the default
for a bare-spells base in static_helpers — correct only for card-type
choosers (Cloud Key / Umori / Stenn). At cast time IsChosenCardType reads
a ChosenAttribute::CardType that Morophon never stored, so the WUBRG
reduction never matched any spell.

The card-level reconcile_self_chosen_type_statics pass now rewrites
IsChosenCardType -> IsChosenCreatureType inside ModifyCost spell filters
when the card's PERSISTED choice is a creature type (CR 607.2d linked
abilities). It keys strictly on the persisted choice, never the
creature-card-type fallback, so Umori (a creature that chooses a card
type) keeps IsChosenCardType.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -1027,9 +1027,32 @@ fn append_sub_ability(chain: &mut AbilityDefinition, tail: AbilityDefinition) {
 }
 
 fn reconcile_self_chosen_type_statics(result: &mut ParsedAbilities, types: &[String]) {
-    let Some(chosen_kind) = chosen_subtype_kind_from_persisted_choice(result)
-        .or_else(|| chosen_kind_from_card_types(types))
-    else {
+    let persisted_kind = chosen_subtype_kind_from_persisted_choice(result);
+
+    // CR 607.2d + CR 205.3 + CR 601.2f: A cost reducer that refers to "the chosen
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
VERDICT: approve

Reviewed the current head via GitHub API diff and isolated worktree. The fix is at the parser reconciliation seam that has cross-clause visibility: Morophon persists a creature-type choice, so the bare "Spells of the chosen type" cost reducer is retargeted from `IsChosenCardType` to `IsChosenCreatureType` only when the persisted choice is actually a creature type. The Umori/CardType case stays card-typed, and Herald's Horn remains idempotent.

I checked Gemini's concerns against the current shape. The nested-filter recursion concern is not blocking this PR's emitted Morophon filter; the helper handles the `TargetFilter` nesting used by this spell-filter shape. The `ImposeAdditionalCost` sibling concern does not match a current chosen-type spell-cost class for this issue; the real nearby corpus hit is a land self-activation tax, not Morophon's spell cost reducer.

Repush guard passed on `bd9f102bfcf2869393faa1fe977490d606087ae3`; GitHub checks are green/CLEAN.

PR #4202, opened 2026-06-23
fix(engine): emit restricted ChoiceType::CardType for card-type enumerations

### LINKED ISSUE #4201: [Card Bug] Archon of Valor's Reach: "can't cast spells of the chosen type" never fires (Labeled vs CardType choice mismatch)
### Card(s)

Archon of Valor's Reach

### Actual behavior

The static "Players can't cast spells of the chosen type" never prevents any spell. The chosen card type is selected on ETB but the prohibition has no effect.

### Root cause (engine-authoritative parse, `card-data.json`)

Archon chooses from a **non-canonical subset** of card types:

> As ~ enters, choose artifact, enchantment, instant, sorcery, or planeswalker.
> Players can't cast spells of the chosen type.

Because the option list (`artifact/enchantment/instant/sorcery/planeswalker`) is **not** the full canonical card-type set, the parser routes it to `ChoiceType::Labeled { options: [...] }` rather than `ChoiceType::CardType`:

```json
"effect": { "type": "Choose",
  "choice_type": { "Labeled": { "options": ["Artifact","Enchantment","Instant","Sorcery","Planeswalker."] } },
  "persist": true }
```

At resolution, `ChosenAttribute::from_choice` maps a `Labeled` choice to `ChosenAttribute::Label(value)` — a string label — **not** `ChosenAttribute::CardType` (`crates/engine/src/types/ability.rs`, `from_choice`).

But the `CantBeCast` static carries `FilterProp::IsChosenCardType`:

```json
"CantBeCast": { "who": "AllPlayers" },
"affected": { "type": "Typed", "properties": [ { "type": "IsChosenCardType" } ] }
```

`IsChosenCardType` only reads `ChosenAttribute::CardType` (`crates/engine/src/game/filter.rs:2106`, `crates/engine/src/game/casting.rs:12245`). Archon stored a `Label`, so the filter returns `false` for every spell and the prohibition never applies.

This is distinct from #1654 (a creature-type-vs-card-type mismatch in cost reducers). Here it is a **Labeled-vs-CardType** mismatch: a restricted card-type enumeration is parsed as an anchor-word label instead of a card-type choice.

### Secondary parse artifact

The last option is captured as `"Planeswalker."` — the sentence-ending period is included in the option string.

### Suggested fix direction

A "choose [one of these card types]" enumeration that lists a *subset* of the canonical card types (Archon, and the broader class of "choose a card type from X, Y, or Z" cards) should resolve to a card-type choice that stores `ChosenAttribute::CardType`, so the linked `IsChosenCardType` prohibition resolves correctly — rather than `ChoiceType::Labeled`/`ChosenAttribute::Label`. CR 205.2a (card types) + CR 607.2d (linked abilities). Also strip the trailing period from the final enumerated option.

(Found during review of #1654 / PR #4197; analysis only — no code change yet.)

claytonlin1110:
## Summary
- Parameterizes `ChoiceType::CardType` with an `excluded: Vec<CoreType>` axis (mirrors `Color { excluded }`), adding `ChoiceType::card_type()` / `card_type_excluding(...)` constructors and a new `CoreType::CHOOSABLE_TYPES` constant.
- Fixes the actual bug site — `mtgish-import`'s `A::ChooseACardtypeFromList` converter arm — to emit a restricted `ChoiceType::CardType` instead of `ChoiceType::Labeled`, so the choice persists as `ChosenAttribute::CardType` and `FilterProp::IsChosenCardType` (the "can't cast spells of the chosen type" prohibition) actually binds.
- Updates every other `ChoiceType::CardType` call site (serialization, `compute_options`, coverage labels, frontend choice-kind mapping in `manabrew-compat`) and the existing tests that referenced the old unit variant.
- Backward-compatible JSON: serializes as the legacy unit variant when `excluded` is empty, struct form only when restricted (same trick as `Color`/`Opponent`).

Fixes #4201

## Test plan
- [x] `cargo test -p engine --lib` (new `choose_card_type_excludes_restricted_types` test + all existing `ChoiceType::CardType` tests pass)
- [x] `cargo test -p mtgish-import --lib` (new `choose_a_cardtype_from_list_lowers_to_restricted_card_type_choice` test passes; full suite 131/131)
- [x] `cargo test -p engine --test integration` — `cloud_key_chosen_type_cost`, `serras_emissary_chosen_card_type_protection`
- [x] `cargo clippy -p engine -p mtgish-import -p manabrew-compat --all-targets -- -D warnings`
- [x] `cargo fmt --all` (no diff)
- [ ] Regenerate `data/card-data.json` (e.g. via Tilt's `card-data` resource) to pick up the converter fix for the live game engine — not done here since Tilt wasn't running locally.

Note: two pre-existing, unrelated failures observed locally (stale `data/card-data.json` against an already-landed `Double`→`Times{factor}` refactor, and 6 stale `mtgish-import` golden fixtures against a `sorcery_speed` field migration) — neither touches files in this diff.


matthewevans (changes requested):
VERDICT: request-changes

**[MED] Wire the normal `Action::ChooseACardtypeFromList` sibling too.** The PR fixes the replacement/as-enters path, but the ordinary spell-action schema still has the same restricted card-type choice shape and still falls through the action converter. Evidence: `Action` includes `ChooseACardtypeFromList(Vec<CardType>)` in `crates/mtgish-import/src/schema/types.rs:9113-9114`, while `crates/mtgish-import/src/convert/action.rs:3892-3897` only handles unrestricted `Action::ChooseACardtype` before the unknown-variant fallback at `crates/mtgish-import/src/convert/action.rs:4066-4068`. The new typed restricted conversion exists only for `ReplacementActionWouldEnter::ChooseACardtypeFromList` in `crates/mtgish-import/src/convert/replacement.rs:1901-1917`. This is not just hypothetical: `data/mtgish-cards.json` has spell-action users such as Season of Gathering, Turnabout, and Winding Way with `_Action":"ChooseACardtypeFromList"` followed by `IsCardtypeVariable` consumers. As written, those cards do not reach the new `ChoiceType::CardType { excluded }`/`ChosenAttribute::CardType` path, so the card-type-enumeration support remains split by schema location. Please share the allowed-list-to-excluded-card-type helper and call it from both the replacement and normal action converter, with a regression for one of the spell-action cards (Winding Way/Turnabout/Season of Gathering) proving the chosen card type persists and feeds the downstream variable filter.

matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** This PR is still routed through the dormant mtgish import path, which is not an accepted contribution lane. Evidence: the current diff changes `crates/mtgish-import/src/convert/action.rs`, `condition.rs`, `filter.rs`, and `replacement.rs`, while `docs/AI-CONTRIBUTOR.md` states that `mtgish/`, `crates/mtgish-import/`, and `data/mtgish-*` are dormant and are not live consumers of the engine/parser/card data; the live runtime pipeline is MTGJSON -> `crates/engine/src/parser/` -> `client/public/card-data.json`. Why it matters: even if the engine-side `ChoiceType::CardType { excluded }` shape is plausible, accepting mtgish-driven converter work reopens a historical pipeline the repo has explicitly called out as dormant. Suggested fix: close or rescope the PR to a live MTGJSON/parser card class without mtgish-import changes; otherwise it should not land.

matthewevans (approved):
VERDICT: approve

Superseding my previous mtgish-lane hold. I re-reviewed the current head against the normal PR bar: correct seam (restricted `ChoiceType::CardType` plus shared converter helper), idiomatic/reasonable/maintainable implementation (typed excluded-axis that preserves legacy serialization), and real value with proof (normal `Action::ChooseACardtypeFromList` sibling plus replacement/as-enters coverage). Gemini's iterator-combinator suggestion is style-only and not a correctness blocker here.

PR #4212, opened 2026-06-23
fix(draft): make pre-Play-Booster sets draftable (draft/default fallback)

matthewevans:
extract_set_pool_indexed only built a draft pool when a set's MTGJSON file
had a `booster.play` section. Play Boosters only exist from MKM (Feb 2024)
onward, so every earlier set — which carries `booster.draft` (~2018–2024) or
`booster.default` (the oldest expansions: Ice Age, Antiquities, Legends, …) —
fell through the early-return and was silently dropped. draft-pools.json held
only the 17 Play-Booster-era set codes despite 600+ set files being downloaded.

Add MtgjsonBooster::draftable(): prefer `play`, fall back to `draft`, then
`default`. The three products share an identical MTGJSON shape (sheets +
weighted boosters) and are mutually exclusive across the corpus. Rename
MtgjsonBoosterPlay -> MtgjsonBoosterConfig since it now models all three.
Platform-only (`arena`/`mtgo`) and non-draft (`set`/`collector`/`jumpstart`)
products stay excluded. Regenerating draft-pools.json now yields 171 sets.

### DIFF
--- a/crates/draft-core/src/extraction.rs
+++ b/crates/draft-core/src/extraction.rs
@@ -66,10 +66,12 @@ impl MtgjsonBooster {
     /// intent. Platform-only products (`arena`, `mtgo`) and non-draft products
     /// (`set`, `collector`, `jumpstart`) are deliberately excluded.
     fn draftable(&self) -> Option<&MtgjsonBoosterConfig> {
+        // Eager `.or()` (not `.or_else`): `as_ref()` is trivial and side-effect
+        // free, so clippy::unnecessary_lazy_evaluations rejects a lazy closure.
         self.play
             .as_ref()
-            .or_else(|| self.draft.as_ref())
-            .or_else(|| self.default.as_ref())
+            .or(self.draft.as_ref())
+            .or(self.default.as_ref())
     }
 }
 
--- a/crates/draft-core/src/extraction.rs
+++ b/crates/draft-core/src/extraction.rs
@@ -68,8 +68,8 @@ impl MtgjsonBooster {
     fn draftable(&self) -> Option<&MtgjsonBoosterConfig> {
         self.play
             .as_ref()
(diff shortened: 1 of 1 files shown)


matthewevans:
## Addressed Gemini review

Pushed `811b5d8` resolving all three findings.

**[MED] `or` → `or_else` (idiomatic lazy chaining)** — Adopted. `draftable()` now uses `.or_else(|| self.draft.as_ref()).or_else(|| self.default.as_ref())`.

**[HIGH] `booster_eligible` keyed to `"play"`** — Confirmed the underlying inconsistency and fixed it, though the framing needed correction. Two facts from the data:

1. The field is **consumed nowhere** in the repo (only written here; pack generation works off `sheets`), and
2. MTGJSON's per-card `boosterTypes` carries *pool* tags (`default`/`deck`), **never** the set-level product key — so `contains("play")` was `false` for **every card in every set, Play Booster sets included** (e.g. all 458 MKM cards are `["default"]`, zero contain `"play"`). The check was dead before this PR, not a regression it introduced.

The suggested fix (test with `boosterTypes: ["draft"]`, check for `"draft"`) wouldn't match real data either — actual draft-era cards (DOM) are `["default"]`, not `["draft"]`. So rather than swap one wrong product string for three, `booster_eligible` now derives from **actual booster-sheet membership** (`uuids_in_sheets`) — the era-independent ground truth for "can be opened in a pack of this set". This is correct for play/draft/default uniformly and removes the magic string entirely.

**[HIGH] Assert `booster_eligible` in tests** — Added `assert!(pool.prints.iter().all(|p| p.booster_eligible))` to both the draft- and default-fallback tests. The test cards deliberately carry **empty** `boosterTypes`, so the assertion only passes because eligibility comes from sheet membership — it would have failed under the old `"play"` check.

PR #4259, opened 2026-06-24
fix(endure): Warden of the Grove dynamic endure X targets entering creature

kiannidev:
## Summary

- Parse endure clauses with `, where X is ...` so Warden of the Grove binds X to counters on the source creature instead of enduring 0.
- Resolve the enduring permanent from the zone-change trigger event when the parser marks `CostPaidObject` / "it endures", so counters land on the entering creature.
- Route endure branch choice through that creature's controller via `prompt_next`.

Fixes #1120

## Test plan

- [x] `cargo test -p engine --lib game::effects::endure::tests`
- [x] `cargo test -p engine effect_endure_dynamic`
- [x] `cargo clippy -p engine -- -D warnings`

Made with [Cursor](https://cursor.com)

PR #4260, opened 2026-06-24
fix(fight): Ent's Fury chained fight uses both chosen creatures

### LINKED ISSUE #1135: Fight doesn't seem to work? — Attached is where Ent's Fury didn't work properly, but I also had a similar issue with Kn…
Reported in Discord: https://discord.com/channels/1485498006781427802/1508659695546536038/1508659695546536038

**Thread:** Fight doesn't seem to work?
**Cards:** ent's fury

## Summary
Attached is where Ent's Fury didn't work properly, but I also had a similar issue with Knockout Manuever.

### ent's fury
Put a +1/+1 counter on target creature you control if its power is 4 or greater. Then that creature gets +1/+1 until end of turn and fights target creature you don't control.

---
<sub>report_id: `discord:1508659695546536038:1508659695546536038:0` · discord: `1508659695546536038/1508659695546536038`</sub>

kiannidev:
## Summary

- Re-validate anaphoric fight clauses: validate only the explicit opponent against the fight filter; recover the implicit ally from flattened chain targets when exactly one explicit target survives (CR 608.2b).
- Remove unvalidated root-target rehydrate in `fight::resolve`; resolution uses post-validation targets only.
- Add regressions for Ent's Fury fight damage and fizzle when the opponent phases out before resolution.

Fixes #1135

## Test plan

- [x] `cargo test -p engine --test integration issue_1135`
- [x] `cargo test -p engine --lib ents_fury`
- [x] `cargo clippy -p engine --tests -- -D warnings`
- [x] CI green

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,13 @@
+# CI test profile (activated with `cargo nextest run --profile ci`).
+#
+# nextest already defaults test-threads to the number of logical CPUs, so this
+# profile does NOT raise per-runner parallelism — the CI speedup comes from
+# sharding the test run across multiple runners (see the rust-test matrix in
+# .github/workflows/ci.yml). This profile adds two CI-appropriate behaviors:
+#   - fail-fast = false: a shard reports every failure, not just the first one.
+#   - slow-timeout: warn on any test running past 60s and hard-kill it after
+#     4 periods (240s) as a runaway/hang guard. The slowest legitimate test is
+#     ~12s, so this never trips on healthy tests.
+[profile.ci]
+fail-fast = false
+slow-timeout = { period = "60s", terminate-after = 4 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,9 +87,13 @@ jobs:
           ./scripts/check-test-card-data-load.sh "$BASE"
 
   rust-test:
-    name: Rust tests
+    name: Rust tests (shard ${{ matrix.shard }}/2)
     runs-on: ubuntu-latest
     timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2]
     steps:
       - uses: actions/checkout@v4
 
@@ -108,7 +112,7 @@ jobs:
         # caught before it lands on the branch that feeds staging/preview.
         env:
           PROPTEST_CASES: ${{ github.event_name == 'pull_request' && '32' || '256' }}
-        run: cargo nextest run --workspace --exclude phase-tauri --exclude mtgish-import --features engine/proptest --status-level fail --final-status-level fail
+        run: cargo nextest run --profile ci --partition count:${{ matrix.shard }}/2 --workspace --exclude phase-tauri --exclude mtgish-import --features engine/proptest --status-level fail --final-status-level fail
 
   card-data-gate:
     name: Card data (generate, validate, coverage)
--- a/client/src/adapter/types.ts
+++ b/client/src/adapter/types.ts
@@ -1087,6 +1087,12 @@ export interface ModalChoice {
    */
   mode_pawprints?: number[];
   constraints?: Array<{ type: string }>;
+  /**
+   * CR 700.2 + CR 107.3m: Engine-internal dynamic "choose up to X —" cap
+   * descriptor (a serialized QuantityExpr). Resolved live by the engine into
+   * `max_choices` before the choice is offered; the UI never reads this field.
+   */
+  dynamic_max_choices?: unknown;
 }
 
 // CR 603.3b: Display payload for one collected-but-not-yet-stacked trigger
--- a/client/src/components/board/PermanentCard.tsx
+++ b/client/src/components/board/PermanentCard.tsx
@@ -222,6 +222,7 @@ export const PermanentCard = memo(function PermanentCard({ objectId, attachments
     incomingAttackerCounts,
     manaTappableObjectIds,
     selectableManaCostCreatureIds,
+    selectableSacrificeObjectIds,
     undoableTapObjectIds,
     validAttackerIds,
     validTargetObjectIds,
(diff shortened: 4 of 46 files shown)


matthewevans (changes requested):
VERDICT: request-changes

I reviewed the API diff in an isolated worktree. The seam is close, but the target revalidation change drops the explicit target filter, so it does not satisfy the normal PR bar yet.

**[HIGH] Preserve `Fight` target-slot legality when carrying the implicit ally forward.** Evidence: in `crates/engine/src/game/ability_utils.rs`, the implicit-subject `Fight` branch now collects every battlefield creature from `candidate_targets` and never calls `targeting::validate_targets_for_ability` against the fight effect's `target` filter. Why it matters: for Ent's Fury the second chosen target is `target creature you don't control`; if that target changes controller or otherwise stops matching the filter before resolution, CR 608.2b requires it to be removed/ignored. This implementation still keeps it as long as it is a battlefield creature, so the fight can resolve with an illegal target. Suggested fix: keep the parent/ally target separately, but still validate the explicit fight target against the actual `target` filter, then resolve only when both the implicit fighter and the still-legal explicit target remain.

**[MED] Avoid substring parser dispatch for the compound anaphoric fight shape.** Evidence: `crates/engine/src/parser/oracle_effect/mod.rs` adds `lower.contains("that creature") && lower.contains(" and fights ")` inside fight-subject rebinding. This repo's parser rules require nom/composable phrase parsing for dispatch rather than broad substring checks. Suggested fix: express the compound shape with the existing nom parsing style around the shared fight/anaphoric phrase, so this remains a grammar building block rather than a text-search special case.

matthewevans (changes requested):
VERDICT: request-changes

The broad parser-dispatch issue from the prior review is addressed on this head, but the target-legality blocker remains for the path where the explicit fight target stops being legal before resolution.

**[HIGH]** Do not keep or recover the implicit fighter when the explicit fight target is no longer legal. Evidence: `crates/engine/src/game/ability_utils.rs:1450` builds `explicit` by validating candidates against the explicit target filter, but `crates/engine/src/game/ability_utils.rs:1466` still keeps an implicit ally even when `explicit` is empty; then `crates/engine/src/game/effects/fight.rs:87` rehydrates object targets from the resolving root ability whenever the fight sub-link has fewer than two object targets, without applying that explicit target filter. Why it matters: if Ent's Fury's `target creature you don't control` changes controller, leaves, or phases out before resolution, CR 608.2b requires that targeted fight clause to do nothing; this path can still fight using the stale root target or resolve with only the ally. Suggested fix: only keep the implicit ally when exactly one explicit target survives validation, remove the unvalidated root-target fallback after validation, and add a regression where the opponent target becomes illegal before resolution and no fight damage is dealt.

kiannidev:
Thanks for the follow-up review — addressed on latest head (`407b4b99c`):

**[HIGH] implicit ally when explicit target is illegal**
- `validate_targets_in_chain` now only recovers the implicit ally when **exactly one** explicit target survives opponent-filter validation (`explicit.len() == 1`). If the opponent target phases out / leaves / fails the filter, `kept` is empty and the fight clause fizzles per CR 608.2b.
- Anaphoric fight validation always gathers candidates from the **resolving spell chain** (`flatten_targets_in_chain`), not just the fight sub-link's local `targets` (fixes the case where only the opponent was stored on the nested link).
- Removed the unvalidated `flatten_targets_in_chain` rehydrate in `fight::resolve` — resolution now uses only post-validation `ability.targets`.

**Regression added:** `ents_fury_fight_fizzles_when_opponent_phases_out_before_resolution` — wolf is phased out after commit, before resolution; no fight damage is dealt.

**Gemini (CR annotation / phased-in):** CR 701.14a comment is on the validation path; phased-in check is in `fight_creature_on_battlefield` during anaphoric revalidation.

CI is green on this head. Happy to take another look.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed current head after the latest follow-up and Gemini comments. The prior CR 608.2b / CR 701.14a semantic blocker is addressed: the implicit ally is retained only when exactly one explicit target still validates against the fight target filter, the unvalidated root-target fallback is gone, and the phased-out explicit-target regression proves the no-fight path.

This still does not clear the maintainer bar because the new integration test adds inline imports.

**[LOW] Move new test imports to module scope.** Evidence: `crates/engine/tests/integration/issue_1135_ents_fury_fight.rs:59-61` imports `PhaseOutCause`, `phase_out_object`, and `GameEvent` inside `ents_fury_fight_fizzles_when_opponent_phases_out_before_resolution`. Why it matters: this repo's AGENTS instructions require all imports at the top of the file, and new test code should follow that standard rather than hiding dependencies in a test body. Suggested fix: move those three imports into the existing top-level import block.

No approval, label, or enqueue while exact-head CI is still pending, the branch is BEHIND, and this cleanup remains.

kiannidev:
CI fix pushed (`595a29de8`): removed unused `GameEvent` import in the fizzle regression (clippy `-D warnings` on integration tests). Also merged latest `main` (includes merged #4259 endure work). Re-resolved Gemini threads. Waiting on CI + matthewevans re-review.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed the new head after the CI cleanup commit. The semantic fight/targeting concerns remain addressed and exact-head CI is green/CLEAN, but the latest commit only removed the unused `GameEvent`; the inline imports called out in the prior review are still present.

**[LOW] Move the remaining test imports to module scope.** Evidence: `crates/engine/tests/integration/issue_1135_ents_fury_fight.rs:59-60` still imports `PhaseOutCause` and `phase_out_object` inside `ents_fury_fight_fizzles_when_opponent_phases_out_before_resolution`. Why it matters: the repo instructions require imports at the top of the file, and this was the only remaining maintainer-bar blocker after the semantic review passed. Suggested fix: move both imports into the existing top-level import block.

No approval, label, or enqueue while this cleanup remains.

matthewevans:

Re-reviewed current head `54c9daec02`. The latest commit exactly addresses the remaining maintainer-bar blocker from the prior review: `PhaseOutCause` and `phase_out_object` moved from the test body into the module-scope import block in `crates/engine/tests/integration/issue_1135_ents_fury_fight.rs`.

The earlier semantic fight/targeting review remains satisfied: the explicit opponent target is still revalidated before the chained fight resolves, and the phased-out regression covers the CR 608.2b no-fight path.

No approval/label/enqueue yet because exact-head CI is still running and GitHub currently reports the branch as BEHIND.

kiannidev:
Addressed the remaining import cleanup on latest head (54c9daec0): `PhaseOutCause` and `phase_out_object` are now at module scope in `issue_1135_ents_fury_fight.rs`. CI should be green — ready for another look.

matthewevans (approved):
VERDICT: approve

Current head `54c9daec02` clears the prior review chain. The broad parser-dispatch issue was removed, the explicit fight target is revalidated against the actual fight target filter before the chained fight resolves, the implicit ally is retained only when exactly one explicit target survives, and the unvalidated root-target fallback is gone.

This satisfies the three PR requirements: the fix is at the engine targeting/fight-resolution seam, the implementation reuses the existing target validation path rather than bypassing it, and the production Ent's Fury regressions prove both the boosted-fight path and the CR 608.2b no-fight path when the explicit opponent target phases out.

Repush guard held immediately before approval: head unchanged at `54c9daec02`, GitHub reported CLEAN, and all exact-head checks were success/skipped.

PR #4275, opened 2026-06-24
fix(client): make attached permanents clickable as targets/activation sources

matthewevans:
Attached Equipment/Auras render as cards tucked behind their host and
collapse to a single peeking card when a host has more than one. During
target selection or activation those attachments are valid game objects
(CR 301.5 / CR 303.4) but are occluded behind the host, so clicks fall
through to the host: a "put a counter on target nonland permanent you
control" trigger lands on the host creature instead of the chosen
Equipment, and an attached Equipment can't be re-activated to move it.

Auto-expand and lift a host's attachments whenever any of them is
actionable in the current waiting state (valid target, activation
source, mana-tap source, board choice, or sacrifice choice), so each is
independently clickable without requiring a hover. Keyed on the
attachment's id, so hosts don't fan out during unrelated targeting.

### DIFF
--- a/client/src/components/board/PermanentCard.tsx
+++ b/client/src/components/board/PermanentCard.tsx
@@ -222,6 +222,7 @@ export const PermanentCard = memo(function PermanentCard({ objectId, attachments
     incomingAttackerCounts,
     manaTappableObjectIds,
     selectableManaCostCreatureIds,
+    selectableSacrificeObjectIds,
     undoableTapObjectIds,
     validAttackerIds,
     validTargetObjectIds,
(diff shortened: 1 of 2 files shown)


matthewevans:
Addressed the Gemini review note (include `undoableTapObjectIds` in the `attachmentsActionable` check) in follow-up #4279. No extra state gate was needed: `undoableTapObjectIds` is already gated upstream in `GameBoard` (`undoLegal`: Priority/ManaPayment/UnlessPayment, excluding UnlessPaymentChooseCost), so the set is empty whenever the untap isn't legal.

PR #4129, opened 2026-06-22
Restore graveyard cards when cancelling delve payment (fixes #3993)

kiannidev:
## Summary
- On `CancelCast`, return cards exiled for delve back to the graveyard
- Clear per-turn and persistent exile-with-source links for the cancelled spell
- Remove delve mana markers from the mana pool

## Test plan
- [x] `cargo test -p engine --lib delve_cancel_cast_returns_exiled_cards_to_graveyard`
- [x] `cargo fmt --all`

Made with [Cursor](https://cursor.com)

Issue #3993, opened 2026-06-21
Delve Cancel Payment — If I click cancel while paying a delve cost, the already-delved cards stay in exile.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1518029182301372538/1518029182301372538

**Thread:** Delve Cancel Payment
**Cards:** _none detected_
**Parser status:** has_gaps
**Extraction confidence:** 0.90

## Summary
If I click cancel while paying a delve cost, the already-delved cards stay in exile.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1518029182301372538:1518029182301372538:0` · discord: `1518029182301372538/1518029182301372538`</sub>

mike-theDude:
This exact bug ("cancelling a Delve payment leaves already-exiled cards stranded in exile") was previously reported and fixed in commit `c2f36cc778` ("Restore graveyard cards when cancelling delve payment (fixes #3993)"), which shipped in release `v0.10.0` — the repo is now at `v0.21.0`, so the fix has been in every released build for a long time. `handle_cancel_cast` in `crates/engine/src/game/casting.rs` still contains the restoration logic: it scans the caster's mana pool for `is_convoke_payment` units whose source is in `Zone::Exile`, moves each back to the graveyard via `move_to_zone`, and cleans up the matching `exile_links`/`cards_exiled_with_source_this_turn` entries. Two regression tests (`delve_cancel_cast_returns_exiled_cards_to_graveyard`, `delve_cancel_cast_preserves_unrelated_exiled_with_source_cards`) in `casting_tests.rs` still cover this path. Since the report gives no card name or specific game state, I can't tell whether this is a stale/duplicate Discord report predating the fix, or a genuine new edge case (e.g. a different cancel entry point) that isn't covered by the existing restoration logic — a repro (which spell, which point in the payment UI the cancel was clicked) is needed to confirm either way.

*status:needs-repro, classifier:cannot-determine*

mike-theDude:
[...]

PR #4281, opened 2026-06-24
fix: add integration regression test for issue #3993

andriypolanski:
## Summary

- Adds an end-to-end integration regression test for [issue #3993](https://github.com/phase-rs/phase/issues/3993): cancelling during Delve mana payment must return delved graveyard cards from exile.
- The engine fix landed on `main` in #4129 (`handle_cancel_cast` rollback for delve exiles). This PR adds `GameScenario`-level coverage so the reported user flow stays locked in alongside the existing unit tests in `casting.rs`.

## Problem

While paying a Delve cost, clicking **Cancel** left already-delved graveyard cards stranded in exile instead of returning them to the graveyard.

## Fix context

Engine behavior (already on `main` via #4129):

- On `CancelCast` during `WaitingFor::ManaPayment`, `handle_cancel_cast` finds delve payments via convoke-style mana pool markers.
- Moves those cards from exile back to the graveyard (CR 601.2i).
- Clears `exile_links` and `cards_exiled_with_source_this_turn` tracking.
- Removes delve mana markers from the pool and pops the placeholder stack entry.

This PR does **not** change engine logic — it adds integration coverage for the full cast pipeline.

## Changes

| File | Change |
|------|--------|
| `crates/engine/tests/integration/issue_3993_delve_cancel_payment.rs` | New regression test: cast delve spell → delve two GY cards → `CancelCast` → assert cards return to graveyard, spell returns to hand, stack clears, tracking and pool markers reset |
| `crates/engine/tests/integration/main.rs` | Register new test module |

## Test plan

- [x] `cargo test -p engine issue_3993_cancel_during_delve_payment`
- [x] `cargo test -p engine delve_cancel_cast` (existing unit tests in `casting.rs`)
- [ ] Manual: cast a Delve spell (e.g. Murktide Regent), delve one or more graveyard cards, click **Cancel** in the mana payment

### DIFF
--- a/crates/engine/tests/integration/issue_3993_delve_cancel_payment.rs
+++ b/crates/engine/tests/integration/issue_3993_delve_cancel_payment.rs
@@ -1,0 +1,118 @@
+//! Regression for issue #3993: Cancelling during delve mana payment must return
+//! delved graveyard cards from exile instead of leaving them stranded.
+//!
+//! CR 601.2i: If the player is unable or unwilling to complete a cast, the
+//! process is reversed and any choices made (including delve exiles) are undone.
+//!
+//! https://github.com/phase-rs/phase/issues/3993
+
+use engine::game::scenario::{GameScenario, P0};
+use engine::types::actions::GameAction;
+use engine::types::game_state::{CastPaymentMode, ConvokeMode, WaitingFor};
+use engine::types::mana::{ManaType, ManaUnit};
+use engine::types::phase::Phase;
+use engine::types::zones::Zone;
+
+const DELVE_DRAW_ORACLE: &str =
+    "Delve (Each card you exile from your graveyard while casting this spell pays for {1}.)\n\
+Draw a card.";
+
+fn add_mana(runner: &mut engine::game::scenario::GameRunner, ty: ManaType, count: usize) {
+    let pool = &mut runner.state_mut().players[P0.0 as usize].mana_pool;
+    for _ in 0..count {
+        pool.add(ManaUnit::new(
+            ty,
+            engine::types::identifiers::ObjectId(0),
+            false,
+            vec![],
+        ));
+    }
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
VERDICT: request-changes

The test is in the right general lane: it drives the production cast/payment/cancel path for issue #3993 rather than only asserting a helper. It does not clear the maintainer bar yet because the head is fmt-red and Gemini's scenario-builder comment is correct for this test shape.

**[LOW] Use the scenario builder for the starting mana pool and fix formatting.** Evidence: `crates/engine/tests/integration/issue_3993_delve_cancel_payment.rs:20-30` adds an imperative `add_mana` helper that mutates `runner.state_mut()` after `scenario.build()`, while `GameScenario::with_mana_pool` exists for deterministic payment setup; GitHub's `Rust lint (fmt, clippy, parser gate)` check also fails in the `Check formatting` step on this head. Why it matters: setup that can be declared before `build()` should stay in the scenario fixture so the test follows the repo's integration-test pattern, and red fmt blocks the PR regardless of semantics. Suggested fix: construct the four `ManaUnit`s in the scenario setup with `scenario.with_mana_pool(P0, ...)`, remove `add_mana`, then run `cargo fmt --all`.

No approval, label, or enqueue while exact-head CI is red.

matthewevans (approved):
VERDICT: approve

Re-reviewed the new head after the follow-up. The prior blocker is addressed: the test now uses `GameScenario::with_mana_pool` for deterministic payment setup, the imperative post-build mana mutation is gone, and exact-head CI is green/CLEAN. This is test-only production-path coverage for the already-landed Delve CancelCast rollback from issue #3993, so it satisfies the PR bar: right seam, repo-pattern fixture setup, and useful regression coverage.

PR #4283, opened 2026-06-24
fix(parser): support 'during your turn' qualifier on LTB triggers

Whovencroft:
## Summary

Reuse the existing `peel_trailing_turn_constraint` helper (nom-based) to strip the trailing "during your turn" qualifier from LTB trigger tails. The qualifier is lowered to a `TriggerCondition::DuringPlayersTurn` intervening-if condition, matching the established pattern for other triggers with this tail (e.g. Moonstone Harbinger).

## Problem

The previous attempt used `strip_suffix(" during your turn")` which failed for the bare-tail case: after `trim_start()` removes the leading space, the tail is exactly `"during your turn"` (16 chars) but the suffix pattern `" during your turn"` (17 chars, with leading space) can never match a string shorter than itself.

## Fix

Delegate to `peel_trailing_turn_constraint`, which already handles both:
- Bare form: `"during your turn"` → returns `("", Some(OnlyDuringYourTurn))`
- Prefixed form: `"without dying during your turn"` → returns `("without dying", Some(OnlyDuringYourTurn))`

The extracted constraint is converted to a `DuringPlayersTurn { player: Controller }` condition on the `TriggerDefinition`. This correctly coexists with the text-level `OncePerTurn` constraint from "This ability triggers only once each turn."

## Cards unlocked

- **Oni-Cult Anvil** (NEO) — batched LTB + during your turn + once per turn
- **Suki, Courageous Rescuer** (TLA) — singular LTB + during your turn + once per turn

## Tests

- `trigger_one_or_more_artifacts_leave_battlefield_during_your_turn`
- `trigger_another_permanent_leaves_battlefield_during_your_turn`

### DIFF
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -18613,6 +18613,24 @@ mod tests {
         );
     }
 
+    /// CR 603.6c + CR 603.4: Synthetic test for an LTB trigger with "during an
+    /// opponent's turn" tail. Proves the opponent-turn branch maps to
+    /// `DuringPlayersTurn { player: Opponent }`.
+    #[test]
+    fn trigger_leaves_battlefield_during_opponents_turn() {
+        let def = parse_trigger_line(
+            "Whenever a creature you control leaves the battlefield during an opponent's turn, draw a card.",
+            "Synthetic Opponent Turn LTB",
+        );
+        assert_eq!(def.mode, TriggerMode::LeavesBattlefield);
+        assert_eq!(
+            def.condition,
+            Some(TriggerCondition::DuringPlayersTurn {
+                player: PlayerFilter::Opponent,
+            }),
+            "during an opponent's turn must become DuringPlayersTurn with Opponent"
+        );
+    }
     /// CR 603.6c: "Whenever a permanent is returned to a player's hand" — Warped Devotion.
     #[test]
     fn trigger_returned_to_a_players_hand() {
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -7535,6 +7535,15 @@ fn try_parse_event(
     .map(|(tail, _)| tail);
     if let Some(tail) = leaves_tail {
         let tail = tail.trim_start();
+        // CR 603.2c + CR 603.6d: Strip trailing "during your turn" constraint
+        // before checking for "without dying" or bare tail. This supports
+        // Oni-Cult Anvil ("one or more artifacts you control leave the
+        // battlefield during your turn") and Suki, Courageous Rescuer
+        // ("another permanent you control leaves the battlefield during your
+        // turn").
+        let (tail, turn_constraint) = peel_trailing_turn_constraint(tail);
+        let during_your_turn =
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
VERDICT: request-changes

Early draft review on the current parser head. The seam looks directionally right — reusing `peel_trailing_turn_constraint` is the existing parser building block for this suffix — but the new CR citations are wrong and need to be fixed before this can clear the maintainer bar.

**[LOW] Replace the incorrect CR 603.6d citations.** Evidence: `crates/engine/src/parser/oracle_trigger.rs:7612`, `crates/engine/src/parser/oracle_trigger.rs:18548`, and `crates/engine/src/parser/oracle_trigger.rs:18578` cite CR 603.6d for leave-the-battlefield trigger parsing. Verified against `docs/MagicCompRules.txt:2602`, CR 603.6d is about ETB/static text; the LTB rule is CR 603.6c, and the `during your turn` condition is the existing CR 603.4 intervening-if shape. Why it matters: rules annotations are part of the review contract in this repo, and a wrong CR number creates false evidence that this parser path was checked against the right rule. Suggested fix: update the code/test comments to cite CR 603.6c for LTB and CR 603.4 for the turn condition; keep CR 603.2c only where the batched one-or-more trigger is being discussed.

No approval, label, or enqueue while the PR is draft, CI is still running, and this citation cleanup remains.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed the new head `2f52c39757` after the repush. The earlier CR-citation blocker is fixed, and the main `during your turn` LTB path still looks like the right parser seam (`peel_trailing_turn_constraint`). However, the new commit added a second behavior branch that is not proved by this PR.

**[LOW] Add a discriminating test for the new opponent-turn LTB branch.** Evidence: `crates/engine/src/parser/oracle_trigger.rs:7626` now maps `TriggerConstraint::OnlyDuringOpponentsTurn` to `TriggerCondition::DuringPlayersTurn { player: PlayerFilter::Opponent }`, but the new regression tests at `crates/engine/src/parser/oracle_trigger.rs:18564` and `:18595` only exercise `during your turn` / `PlayerFilter::Controller`. There is an existing unit check that `peel_trailing_turn_constraint("spell during an opponent's end step")` is not stripped, but that does not prove the LTB branch's newly added opponent-turn mapping. Why it matters: the maintainer bar for parser changes is not just “plausible seam”; it needs proof for each new behavioral branch, especially when a repush adds behavior after review. Suggested fix: add a parser test for an LTB trigger whose tail is exactly `during an opponent's turn` (or another supported `OnlyDuringOpponentsTurn` phrase) and assert the resulting condition uses `PlayerFilter::Opponent` while preserving the same LTB semantics.

No approval, label, or enqueue on this head while CI is still running and the new branch lacks direct coverage.

matthewevans (approved):
VERDICT: approve

Current head `32232d22a5` clears the prior review blockers. The LTB turn-constraint change is at the existing `peel_trailing_turn_constraint`/`TriggerCondition::DuringPlayersTurn` parser seam, CR 603.6c/603.4 citations resolve locally, and the new parser regression directly covers the previously missing opponent-turn branch. Exact-head GitHub checks are green and the repush guard held.

PR #4284, opened 2026-06-24
fix: Sliver Gravemother encore activating at X=0 or for free (#3997)

### LINKED ISSUE #3997: Encore bug — The AI, while controlling Sliver Gravemother, can choose X=0 to encore a Sliver Legion (MV=5) The AI is al…
Reported in Discord: https://discord.com/channels/1485498006781427802/1517863308949196810/1517863308949196810

**Thread:** Encore bug
**Cards:** sliver gravemother, sliver legion
**Parser status:** has_gaps

## Summary
The AI, while controlling Sliver Gravemother, can choose X=0 to encore a Sliver Legion (MV=5)

The AI is also able to pay less than 5 mana to activate

### sliver gravemother
The "legend rule" doesn't apply to Slivers you control.
Each Sliver creature card in your graveyard has encore {X}, where X is its mana value.
Encore {5} ({5}, Exile this card from your graveyard: For each opponent, create a token copy that attacks that opponent this turn if able. They gain haste. Sacrifice them at the beginning of the next end step. Activate only as a sorcery.)
### sliver legion
All Sliver creatures get +1/+1 for each other Sliver on the battlefield.

---
<sub>report_id: `discord:1517863308949196810:1517863308949196810:0` · discord: `1517863308949196810/1517863308949196810`</sub>

andriypolanski:
## Summary

- Fixes [issue #3997](https://github.com/phase-rs/phase/issues/3997): while controlling Sliver Gravemother, the AI could Encore Sliver Legion (MV=5) with X=0 or otherwise pay less than the required mana.
- Concretizes `SelfManaCost` / `SelfManaValue` placeholders and bound `{X}` encore costs at **activation fetch time**, before legality checks, ChooseX, and payment.
- Adds integration regression tests for Sliver Legion encore costing `{5}`, skipping `ChooseXValue`, and rejecting activation when mana is insufficient.

## Problem

Sliver Gravemother grants each Sliver creature card in your graveyard encore `{X}`, where X is its mana value (CR 702.141a + CR 202.3). Two gaps let activation proceed for free or at X=0:

1. **Unresolved self-referential placeholders** — the mana payment path treats `ManaCost::SelfManaCost` / `SelfManaValue` as free (`can_pay_for_spell` / `pay_cost_with_demand_and_choices`).
2. **Literal `{X}` entering ChooseX** — if the synthesized cost still carries a symbolic `{X}` shard, activation detours to `ChooseXValue` with min=0. Encore's effect does not reference X, so the AI's `XValuePolicy` leaves X=0 as the default.

Prior work in #3655 / #4062 fixed parsing and runtime synthesis, but activation still fetched costs that could reach payment/ChooseX unresolved.

## Fix

**`activation_ability_definition`** (single fetch authority for legality + activation) now:

1. **`resolve_self_mana_in_ability_cost`** — recursively resolves `SelfManaCost` / `SelfManaValue` anywhere in the activated ability cost tree against the source object.
2. **`concretize_encore_mana_value_in_ability_cost`** — for `Effect::Encore`, if the cost still has `{X}`, concretizes it to the source card's mana value so Encore never enters `ChooseXValue` with a choosable zero.

## Changes

| File | Change |
|------|--------|
| `crates/engine/src/game/keywords.rs` | Add `resolve_self_mana_in_ability_cost` and `concretize_encore_mana_value_in_ability_cost` |
| `crates/engine/src/game/casting.rs` | Wire both resolvers in `activation_ability_definition` |
| `crates/engine/tests/integration/issue_3997_sliver_gravemother_encore_legion.rs` | Regression: MV=5 encore costs `{5}`, no ChooseX, illegal without full mana |
| `crates/engine/tests/integration/main.rs` | Register new test module |

## Test plan

- [x] `cargo test -p engine issue_3997_sliver_gravemother_encore_legion`
- [x] `cargo test -p engine issue_3655_sliver_gravemother_encore` (existing #3655 coverage)
- [ ] Manual: Sliver Gravemother on battlefield, Sliver Legion in graveyard — Encore should cost `{5}` and be unpayable without five mana; AI should not offer X=0

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -143,21 +143,34 @@ fn activation_ability_definition(
     source_id: ObjectId,
     ability_index: usize,
 ) -> Option<AbilityDefinition> {
-    let obj = state.objects.get(&source_id)?;
-    if let Some(ability) = obj.abilities.get(ability_index) {
-        return Some(ability.clone());
+    let Some(obj) = state.objects.get(&source_id) else {
+        return None;
+    };
+    let mut ability = if let Some(ability) = obj.abilities.get(ability_index) {
+        ability.clone()
+    } else {
+        let offset = ability_index.checked_sub(obj.abilities.len())?;
+        // Must match the append order in `activated_ability_definitions`: printed
+        // abilities first, then runtime-granted cycling, then runtime-granted
+        // graveyard activated (Encore/Scavenge).
+        runtime_granted_cycling_abilities(state, source_id)
+            .into_iter()
+            .chain(runtime_granted_graveyard_activated_abilities(
+                state, source_id,
+            ))
+            .nth(offset)?
+    };
+    if let Some(ref cost) = ability.cost {
+        ability.cost = Some(super::keywords::resolve_self_mana_in_ability_cost(
+            state, source_id, cost,
+        ));
     }
-
-    let offset = ability_index.checked_sub(obj.abilities.len())?;
-    // Must match the append order in `activated_ability_definitions`: printed
-    // abilities first, then runtime-granted cycling, then runtime-granted
-    // graveyard activated (Encore/Scavenge).
-    runtime_granted_cycling_abilities(state, source_id)
-        .into_iter()
-        .chain(runtime_granted_graveyard_activated_abilities(
-            state, source_id,
-        ))
-        .nth(offset)
+    if matches!(ability.effect.as_ref(), Effect::Encore) {
+        if let Some(ref mut cost) = ability.cost {
+            super::keywords::concretize_encore_mana_value_in_ability_cost(state, source_id, cost);
+        }
+    }
+    Some(ability)
 }
 
 pub(crate) fn variable_speed_payment_range(cost: &AbilityCost, max_speed: u8) -> Option<(u8, u8)> {
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -143,9 +143,7 @@ fn activation_ability_definition(
     source_id: ObjectId,
     ability_index: usize,
 ) -> Option<AbilityDefinition> {
-    let Some(obj) = state.objects.get(&source_id) else {
-        return None;
-    };
+    let obj = state.objects.get(&source_id)?;
     let mut ability = if let Some(ability) = obj.abilities.get(ability_index) {
         ability.clone()
     } else {
--- a/crates/engine/src/game/keywords.rs
+++ b/crates/engine/src/game/keywords.rs
@@ -1744,7 +1744,6 @@ mod tests {
     }
 
     use crate::game::combat::{AttackerInfo, CombatState};
-    use crate::game::zones::create_object;
     use crate::types::events::GameEvent;
     use crate::types::game_state::GameState;
 
@@ -1957,7 +1956,6 @@ mod tests {
     #[test]
     fn ninjutsu_enter_as_copy_defers_combat_placement_until_copy_resolves() {
         use crate::game::engine::apply_as_current;
-        use crate::game::zones::create_object;
         use crate::types::ability::{
             AbilityDefinition, AbilityKind, ContinuousModification, Effect, ReplacementDefinition,
             ReplacementMode, TargetFilter, TargetRef, TypeFilter, TypedFilter,
--- a/crates/engine/src/game/keywords.rs
+++ b/crates/engine/src/game/keywords.rs
@@ -956,10 +956,12 @@ mod tests {
     use std::sync::Arc;
 
     use crate::ai_support::legal_actions;
+    use crate::game::zones::create_object;
     use crate::types::ability::{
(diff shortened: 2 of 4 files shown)


matthewevans (changes requested):
VERDICT: request-changes

This is working on a real engine bug (#3997), and the chosen activation-fetch seam is plausible because both legality and activation payment read through `activation_ability_definition`. I also evaluated Gemini's review: the `PerCounter` concern is a real maintainability gap in the new helper's stated contract, while the request to rename/generalize the encore-specific helper is less important than preserving the right semantics for bound-X encore.

**[LOW] Recurse through `AbilityCost::PerCounter` or narrow the helper contract.** Evidence: `resolve_self_mana_in_ability_cost` says it resolves placeholders anywhere in an activated ability cost tree, but its match only descends into `Composite` and `OneOf` (`crates/engine/src/game/keywords.rs:256` / `:262`) and clones all other wrappers. `AbilityCost::PerCounter { base, .. }` is an existing recursive cost wrapper (`crates/engine/src/types/ability.rs:6580`), and existing cost-tree APIs recurse through its `base` (for example `AbilityCost::categories`). The same shape also matters for `concretize_encore_mana_value_in_ability_cost`, which only descends through `Composite` / `OneOf` at `crates/engine/src/game/keywords.rs:292`. Why it matters: the current Sliver Gravemother path likely uses a simple composite mana+exile cost, but these helpers are now general activation-cost tree utilities at a hot seam. If a future granted activated ability carries `PerCounter { base: Mana(SelfManaValue or X), .. }`, this will silently leave the placeholder unresolved and payment/ChooseX can fall back to the same free/X=0 class of bug. Suggested fix: either add a `PerCounter { base, .. }` arm that recursively rewrites/concretizes `base`, or make the helper name/comment explicitly scoped to the exact current cost shapes and avoid claiming whole-tree coverage.

**[LOW] Move the new inline imports to module scope.** Evidence: the PR adds `use crate::types::ability::AbilityCost;` inside `resolve_self_mana_in_ability_cost` (`crates/engine/src/game/keywords.rs:251`) and `use crate::game::casting_costs::cost_has_x;` plus another `AbilityCost` import inside `concretize_encore_mana_value_in_ability_cost` (`crates/engine/src/game/keywords.rs:281`). The repo instruction is explicit that imports belong at the top of the file, and this module already imports `AbilityCost` at module scope. Suggested fix: reuse the existing module-level `AbilityCost` import and add `cost_has_x` at module scope, or qualify the call directly if you prefer not to import it.

No approval, label, or enqueue on this head: the PR is BEHIND and the Rust lint job is already failing, so it does not satisfy the maintainer bar yet.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed current head `3748e1cf6e` after the follow-up. Prior `PerCounter` recursion concern is addressed for both helper paths, and the helper-body inline imports from the previous review were moved to module scope. Gemini’s generalization comment remains non-blocking for the same reason as before: this helper is only called under the Encore effect guard.

**[LOW] Move the new test-body imports to the test module scope.** Evidence: the follow-up adds `use crate::game::zones::create_object;`, `use crate::types::ability::TargetFilter;`, and `use crate::types::counter::CounterType;` inside both new test functions (`crates/engine/src/game/keywords.rs:989` and `crates/engine/src/game/keywords.rs:1037`). Why it matters: the repository instruction is explicit that imports belong at the top of the file/module, and this follow-up reintroduces the same inline-import hygiene issue in the new test coverage. Suggested fix: add those three imports beside the existing `#[cfg(test)] mod tests` imports and remove the function-local `use` lines.

No approval/label/enqueue yet: this head is still BEHIND and exact-head Rust/card-data checks are not complete, and the inline-import cleanup remains.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed current head `7314cefa73` after the follow-up. The earlier function-local import issue is addressed, but this head still does not pass exact-head Rust CI.

**[MED] Restore the branch to a compiling state before review can proceed.** Evidence: the exact-head `Rust lint (fmt, clippy, parser gate)` and Rust test shards for `7314cefa73` fail. The lint log reports `E0252` in `crates/engine/src/game/keywords.rs`: `create_object` is imported at both line 959 and line 1747, and the reimport at 1747 is also unused. The same lint run also reports `clippy::question_mark` at `crates/engine/src/game/casting.rs:146` (`let Some(obj) = state.objects.get(&source_id) else { return None; };` should use `?`).

No approval/label/enqueue while the branch is BEHIND and the current head is red.

matthewevans (approved):
VERDICT: approve

Re-reviewed current head `a2a3af0a59daf06d0d645d22c42bd2defb6290b3`. The latest commit exactly addresses the remaining compile/lint blocker: the duplicate `create_object` import is removed and the clippy `question_mark` case in `activation_ability_definition` now uses `?`. Prior PerCounter recursion and import-scope findings remain addressed, and the production Sliver Gravemother / Sliver Legion integration coverage is still present. Exact-head GitHub checks are green and the branch is CLEAN.

PR #4285, opened 2026-06-24
Fix/wick etb if otherwise

andriypolanski:
## Summary

Fixes Wick, the Whorled Mind's ETB trigger so the post-effect `if you don't control a Snail` gate stays on the Token clause (`execute.condition`, CR 608.2c) instead of hoisting to the trigger-level intervening-if (`TriggerDefinition.condition`, CR 603.4), and assigns deferred targets for the `Otherwise` PutCounter branch at resolution when exactly one legal Snail exists (CR 115.4).

Without the clause-level gate, the trigger only fired when no Snail was present and the Otherwise branch had no condition anchor. Without deferred else-branch target binding, the PutCounter effect resolved with empty targets and placed no counters even when the else branch ran.

## Changes

- **Parser (already landed):** `condition_text_is_rehomeable` word-boundary guard and `try_nom_condition_as_ability_condition` deferral so `"you don't control a Snail"` re-homes to `execute.condition` and attaches PutCounter as `else_ability`.
- **Runtime:** `try_assign_deferred_else_branch_targets` in `effects/mod.rs` — when a conditional `if` branch fails and the `else_ability` runs, build target slots at resolution and auto-bind when a unique legal combination exists.
- **Tests:** Un-ignore `wick_no_snail_creates_token` and `wick_with_snail_pumps_existing_snail`; add parser regression `wick_etb_post_effect_if_stays_on_execute_with_otherwise`; strengthen Wick Case B in `std_small_parser_a_batch.rs` to assert the +1/+1 counter.

## Test plan

- [x] `cargo test -p engine wick_etb_post_effect_if_stays_on_execute_with_otherwise`
- [x] `cargo test -p engine wick_otherwise_attaches_as_token_else_ability`
- [x] `cargo test -p engine wick_no_snail_creates_token wick_with_snail_pumps_existing_snail`
- [ ] `cargo fmt --all` + clippy green


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** The deferred else-branch target binding silently no-ops when the else branch has multiple legal resolution choices. Evidence: `crates/engine/src/game/effects/mod.rs:1319` calls `auto_select_targets_for_ability` and only assigns targets when that returns `Some(selected)`; the existing auto-selector returns `Ok(None)` for ambiguous legal assignments (`crates/engine/src/game/ability_utils.rs:1006`). The PR's Wick coverage only creates one controlled Snail (`crates/engine/tests/otherwise_conditional_effects.rs:540` and `crates/engine/tests/std_small_parser_a_batch.rs:149`), so it never exercises the common case where Wick's controller has two Snails. In that case `try_assign_deferred_else_branch_targets` leaves `else_resolved.targets` empty, then `Effect::PutCounter` resolves through `resolve_defined_or_targets` and ultimately returns only the empty `ability.targets` list (`crates/engine/src/game/effects/counters.rs:1431` and `:1465`), so no Snail gets the +1/+1 counter. Why it matters: Wick says “put a +1/+1 counter on a Snail you control”; with two legal Snails the player needs a resolution-time choice or another supported choice path, not a silent no-op. Suggested fix: route ambiguous deferred else-branch choices through the existing resolution-choice/waiting path, or keep this PR scoped to the unique-target case and leave multi-choice coverage honest/unsupported instead of resolving the branch with empty targets.

Gemini's no-legal-target note is adjacent, but the reachable blocker here is the ambiguous legal-target path: zero Snails is unlikely on Wick's false branch, while two Snails is straightforward.

matthewevans:
VERDICT: hold-current-CI

Re-reviewed current head `acb9498bde` after the follow-up. The prior multiple-Snail blocker is addressed: ambiguous single-object else-branch targets now park the else branch as a pending continuation and prompt through the existing `ChooseFromZoneChoice` path, and the resolution-choice handler injects the selected object target back into the continuation before draining it.

The new Wick regression with two controlled Snails is discriminating for the previous no-op: it resolves the ETB, answers the resolution-time choice, and asserts exactly one existing Snail receives the +1/+1 counter while no new Snail token is created.

No approval/label/enqueue yet because GitHub still reports the branch as BEHIND and the exact-head Rust test shards are still running.

matthewevans (approved):
VERDICT: approve

Current head `acb9498bde` clears the prior multiple-Snail blocker. The fix uses the existing resolution-choice machinery: ambiguous single-object else-branch target selection parks the else branch as a pending continuation, prompts via `ChooseFromZoneChoice`, and the existing `SelectCards` handler injects the chosen object target before draining the continuation.

This satisfies the three PR requirements: the seam is the resolution-time deferred-target binder rather than frontend/test-only logic; the implementation is idiomatic and maintainable by reusing existing target-slot and continuation building blocks; and the two-Snail Wick regression is discriminating because the old behavior resolved with empty targets and would leave both existing Snails without the required +1/+1 counter.

Repush guard held immediately before approval: head unchanged at `acb9498bde`, GitHub reported CLEAN, and all exact-head checks were success/skipped.

PR #4289, opened 2026-06-24
fix(engine): Birthing Ritual sacrifice choice precedes dig choice (#4273)

### LINKED ISSUE #4273: Birthing Ritual asks for creature choice before sacrifice choice
Reported in Discord: https://discord.com/channels/1485498006781427802/1519103125875855493/1519103125875855493

**Thread:** birthing ritual
**Cards:** birthing ritual

## Summary
Birthing Ritual requires the player to select the creature to put onto the battlefield before choosing and sacrificing a creature. The reported expected ordering is sacrifice first, then choose the creature card constrained by the sacrificed creature's mana value.

### birthing ritual
At the beginning of your end step, if you control a creature, look at the top seven cards of your library. Then you may sacrifice a creature. If you do, you may put a creature card with mana value X or less from among those cards onto the battlefield, where X is 1 plus the sacrificed creature's mana value. Put the rest on the bottom of your library in a random order.

---
<sub>report_id: `discord:1519103125875855493:1519103125875855493:0` · discord: `1519103125875855493/1519103125875855493`</sub>

nickmopen:
## Summary

Fixes #4273 — Birthing Ritual was asking for the creature-from-library choice **before** asking which creature to sacrifice, and computing the CMC filter (`MV ≤ sacrificed MV + 1`) before the sacrifice snapshot was available.

**Root cause:** The parser assembled `Dig(interactive) → Sacrifice`, presenting `WaitingFor::DigChoice` immediately — before the player had sacrificed anything and before `effect_context_object` was populated with the sacrifice snapshot.

**Fix:** Add `from_prior_look: bool` to `Effect::Dig`. When `DigFromAmong` detects an optional Sacrifice between the root look-Dig and the "from among those cards" clause, it restructures the chain:

```
Before: Dig(interactive, filter=Creature+CmcLE) → Sacrifice(optional)

After:  Dig(look-only, keep=0)
          → Sacrifice(optional)
               └─ sub_ability: Dig(from_prior_look, dest=Battlefield,
                                   filter=Creature+CmcLE(sac_mv+1),
                                   condition: OptionalEffectPerformed,
                                   else_ability: Dig(from_prior_look, keep=0, rest→Library))
```

The `from_prior_look` resolver reads `state.private_look_ids` (populated by the look-only step) and evaluates the CMC filter **after** the sacrifice snapshot is stamped onto `effect_context_object`. The decline branch (player skips sacrifice) routes all looked-at cards to library bottom via the `else_ability`.

**CR compliance:**
- CR 608.2c: effects resolve in written order (look → sacrifice → choose)
- CR 701.20e: private look populates `private_look_ids`
- CR 202.3: mana value evaluated from sacrifice LKI snapshot

## Test plan

- [x] `birthing_ritual_assembles_dig_battlefield_sacrifice_chain` — verifies parser emits the new chain shape
- [x] `birthing_ritual_runtime_dig_filter_respects_sacrificed_creature_mana_value` — verifies CMC filter evaluates correctly with sacrifice snapshot
- [x] `birthing_ritual_decline_sacrifice_puts_all_looked_at_cards_on_bottom` — verifies decline branch routes all 7 cards to library bottom
- [x] `cargo fmt --all` clean

### DIFF
--- a/crates/engine/src/database/hideaway.rs
+++ b/crates/engine/src/database/hideaway.rs
@@ -28,7 +28,8 @@
 //!    `visibility.rs` grants the controller the CR 702.75a look-permission.
 
 use crate::types::ability::{
-    AbilityDefinition, AbilityKind, Effect, QuantityExpr, TargetFilter, TriggerDefinition,
+    AbilityDefinition, AbilityKind, DigSource, Effect, QuantityExpr, TargetFilter,
+    TriggerDefinition,
 };
 use crate::types::card::CardFace;
 use crate::types::keywords::Keyword;
@@ -86,7 +87,7 @@ fn hideaway_trigger(n: u32) -> TriggerDefinition {
             // CR 701.20e: the cards are looked at privately, not revealed.
             reveal: false,
             enter_tapped: false,
-            from_prior_look: false,
+            source: DigSource::Library,
         },
     )
     // CR 608.2c: continuation — conceal the just-exiled card (ParentTarget).
--- a/crates/engine/src/database/hideaway.rs
+++ b/crates/engine/src/database/hideaway.rs
@@ -86,6 +86,7 @@ fn hideaway_trigger(n: u32) -> TriggerDefinition {
             // CR 701.20e: the cards are looked at privately, not revealed.
             reveal: false,
             enter_tapped: false,
+            from_prior_look: false,
         },
     )
     // CR 608.2c: continuation — conceal the just-exiled card (ParentTarget).
--- a/crates/engine/src/game/effects/dig.rs
+++ b/crates/engine/src/game/effects/dig.rs
@@ -1,6 +1,8 @@
 use crate::game::filter::{matches_target_filter, FilterContext};
 use crate::game::quantity::resolve_quantity_with_targets;
-use crate::types::ability::{Effect, EffectError, EffectKind, ResolvedAbility, TargetFilter};
+use crate::types::ability::{
+    DigSource, Effect, EffectError, EffectKind, ResolvedAbility, TargetFilter,
+};
 use crate::types::events::GameEvent;
 use crate::types::game_state::{GameState, WaitingFor};
 use crate::types::zones::Zone;
@@ -22,7 +24,7 @@ pub fn resolve(
         rest_dest,
         is_reveal,
         enter_tapped,
-        from_prior_look,
+        dig_source,
     ) = match &ability.effect {
         Effect::Dig {
             player,
@@ -34,7 +36,7 @@ pub fn resolve(
             rest_destination,
             reveal,
             enter_tapped,
-            from_prior_look,
+            source,
         } => {
             let resolved_count =
                 resolve_quantity_with_targets(state, count, ability).max(0) as usize;
@@ -55,7 +57,7 @@ pub fn resolve(
                 *rest_destination,
                 *reveal,
                 *enter_tapped,
-                *from_prior_look,
+                *source,
             )
         }
         _ => (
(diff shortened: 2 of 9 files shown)


matthewevans (changes requested):
Thanks for taking on the Birthing Ritual ordering bug. The target is real, but this head is not over the maintainer bar yet.

I checked Gemini's comments against the diff and agree with the two behavioral issues below.

1. `crates/engine/src/game/effects/dig.rs:263`: the `from_prior_look` decline branch only routes the rest pile when `rest_dest == Some(Zone::Library)`, then clears `private_look_ids` unconditionally. That is correct for Birthing Ritual's current bottom-of-library cleanup, but `Effect::Dig` is a general building block and the new flag can be paired with any parsed rest destination. If a future/current parser path produces `rest_destination: Some(Zone::Graveyard)` or another zone with `keep_count == 0`, the looked-at cards are neither moved nor left available for the next effect; they are silently forgotten from the private-look window. This fails the architecture seam: the branch should route through the existing rest-partition/zone-disposition primitive for the actual destination, not special-case Library.

2. `crates/engine/src/game/effects/dig.rs:294`: the required-choice path computes `keep_count` from `raw_keep_count`, not from `selectable_cards.len()`, and then always surfaces `WaitingFor::DigChoice`. If the sacrifice/context filter excludes every looked-at card and the instruction is not optional, the engine can ask the player to choose one card from an empty selectable set. That is an impossible prompt/soft lock instead of deterministic no-choice resolution. This needs a discriminating regression where `private_look_ids` is non-empty, `selectable_cards` is empty, and the resolver does not produce an unsatisfiable `DigChoice`.

Separately, `crates/engine/src/types/ability.rs:8102` adds another boolean mode bit to `Effect::Dig`. That is a maintainability smell in this codebase because it hides the source-set axis behind `true/false`; a typed source enum or equivalent structural encoding would make the resolver exhaustive and easier to extend. I am not making that the sole blocker, but it is part of why the current shape is brittle.

Current GitHub CI is also red on Rust lint and both Rust test shards. No approval, label, or enqueue from this sweep.

matthewevans (changes requested):
Current-head re-review on `50066583`:

The latest commit only fills in missing `from_prior_look: false` fields for CI/compile coverage. The behavioral blockers from the prior review remain on the current head:

- `crates/engine/src/game/effects/dig.rs:263`: the decline branch still routes cards only when `rest_dest == Some(Zone::Library)` and then clears `private_look_ids` unconditionally. Any `from_prior_look` Dig with another rest destination silently loses the looked-at set instead of routing it through the existing rest-destination machinery.
- `crates/engine/src/game/effects/dig.rs:294`: the required-choice path still derives `keep_count` from the raw looked-at count, not the legal `selectable_cards` count, and can still surface a `DigChoice` that requires choosing from an empty selectable set.
- `crates/engine/src/types/ability.rs:8102`: the mode is still encoded as a new `bool` on `Effect::Dig`, which keeps this source-set axis non-exhaustive and easy to misuse.

Current exact-head Rust/AI checks are also still running. No approval, label, or enqueue.

matthewevans (changes requested):
Current-head re-review on `73412a1`:

The latest commit only adds `from_prior_look: false` to `mtgish-import` `Effect::Dig` literals. That compile cleanup is fine, and I am not treating mtgish itself as an acceptance blocker here, but it does not address the live engine issues from the prior reviews.

The current head still has the same blockers:

- `crates/engine/src/game/effects/dig.rs:263`: the `from_prior_look` decline branch only routes the rest pile when `rest_dest == Some(Zone::Library)` and then clears `private_look_ids` unconditionally. Non-library rest destinations still silently lose the looked-at set instead of using the existing destination routing machinery.
- `crates/engine/src/game/effects/dig.rs:294`: required-choice resolution still computes `keep_count` from the raw looked-at count rather than the legal selectable count, so it can still surface an impossible `DigChoice` with no selectable cards.
- `crates/engine/src/types/ability.rs:8102`: the source-set mode remains a new `bool` on `Effect::Dig`, leaving the axis non-exhaustive and easy to misuse.

Current exact-head Rust/AI checks are still running and the branch is BEHIND. No approval, label, or enqueue.

matthewevans:

Current-head re-review on `a53d1b9f4`: the prior blockers appear addressed on the GitHub API diff and isolated worktree review.

- The source-set axis is now typed as `DigSource::{Library, PriorLook}` with serde default/skip for the existing Library case.
- The PriorLook decline path routes through the existing rest-partition machinery for the actual destination instead of hard-coding Library before clearing `private_look_ids`.
- The empty-selectable PriorLook path auto-resolves and routes the rest pile rather than surfacing an impossible `DigChoice`.
- The Birthing Ritual parser shape now makes the root Dig look-only, defers the choice Dig behind the optional sacrifice, and carries an `else_ability` decline cleanup. Existing effect-chain handling defers `OptionalEffectPerformed` gates across the sacrifice prompt and resolves the else branch on false.

I am not formally approving, labeling, or enqueueing yet because exact-head CI is still running: Rust tests shard 2, card-data, and the paired-seed AI gate have not completed on this head.

matthewevans (approved):
Approving exact head a53d1b9f4a942c4a1d84d1cbec1cf2734fa33da1 after repush guard. The reviewed Birthing Ritual ordering head addresses the prior DigSource/PriorLook routing/empty-selectable blockers and exact-head checks are green/CLEAN.

PR #4291, opened 2026-06-24
Fix commit-a-crime on opponent damage (#2008)

### LINKED ISSUE #2008: Ping Deserts doesn't work correctly with "Commit a Crime" — Deserts that ping 1 damage should trigger the mechanic "com…
Reported in Discord: https://discord.com/channels/1485498006781427802/1511105048074129418/1511105048074129418

**Thread:** Ping Deserts doesn't work correctly with "Commit a Crime"
**Cards:** _none detected_

## Summary
Deserts that ping 1 damage should trigger the mechanic "commit a crime"

---
<sub>report_id: `discord:1511105048074129418:1511105048074129418:0` · discord: `1511105048074129418/1511105048074129418`</sub>

kiannidev:
## Summary
- Emit `CrimeCommitted` when damage is dealt to an opponent player (CR 700.13)
- Fixes Desert lands and other ping effects not triggering outlaw commit-a-crime abilities

## Test plan
- [x] `cargo test -p engine --lib dealing_damage_to_opponent_emits_crime_committed`

Fixes #2008


matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** Crime is emitted from the wrong event seam and will over-fire for damage that never targeted the opponent. Evidence: `crates/engine/src/game/effects/deal_damage.rs:571` emits `CrimeCommitted` whenever resolved damage hits a non-controller player, but CR 700.13 says the crime happens as the player casts a spell, activates an ability, or puts a triggered ability on the stack and that spell/ability targets an opponent or their object/card; the repo already has that target-lock seam in `crates/engine/src/game/casting.rs:236`. Damage resolution also handles non-targeted batch damage (`crates/engine/src/game/effects/deal_damage.rs:1313`) and combat damage (`crates/engine/src/game/combat_damage.rs:883`), so this patch makes `DamageEachPlayer`/combat-style player damage commit crimes even though no target was chosen. Why it matters: crime triggers will fire for the wrong class of cards and still won't prove that the original target declaration path is fixed. Suggested fix: remove the damage-resolution emission and fix the activation/trigger target-lock path that is missing the `emit_targeting_events`/`CrimeCommitted` event for Desert/Tor Wauki-style targeted abilities; add a regression that targeted opponent damage commits a crime and non-targeted opponent damage does not.

Gemini's CR 700.13 concern is confirmed by the local rules text and current engine seams.

matthewevans:
VERDICT: request-changes

**[HIGH]** This head fixes the earlier wrong damage-resolution seam, but the replacement processes target-lock side-effect triggers by recursively calling `process_triggers` from inside ordered trigger dispatch (`crates/engine/src/game/triggers.rs:3137-3144`, called after each `dispatch_pending_trigger_context`). That bypasses the existing collect/defer boundary (`collect_triggers_into_deferred` / `deferred_triggers`) and can interleave newly-created `CrimeCommitted`/`BecomesTarget` triggers ahead of remaining triggers from the original APNAP-ordered batch. CR 603.3b stack placement is batch/order sensitive; target-lock events emitted while putting one triggered ability on the stack should be collected and drained at the next legal trigger boundary, not recursively dispatched between siblings. Suggested fix: use the existing deferred-trigger queue shape for the new target-lock event slice, drain after the current ordered batch/trigger construction completes, and add an ordering regression with at least one sibling trigger plus a crime trigger so this cannot silently reorder stack entries.

Gemini status: the original Gemini CR 700.13 objection to damage-resolution crime emission is addressed by commit `8f84ef47f`; the remaining blocker is the new trigger-ordering seam above. CI is also not green on this head: Rust tests pass, but `cargo fmt --all -- --check` fails on `process_target_lock_side_effects` formatting. No approval, labels, enqueue, or merge from the sweep.

kiannidev:
Addresses the request to move crime detection off damage resolution to the target-lock seam:

- The damage-resolution `CrimeCommitted` emission in `deal_damage.rs` is reverted.
- Crime is now surfaced at target lock via the existing `casting::emit_targeting_events` seam. The actual bug was that `dispatch_collected_triggers` dropped the target-lock events before `process_triggers` ran, so auto-targeted Desert ETB pings committed a crime but outlaw `CommitCrime` abilities never fired. The fix scans the target-lock side-effect events emitted during each trigger dispatch and feeds them to `process_triggers`.
- Regressions added: a target-slot test (opponent is a legal target) and an end-to-end test proving the ETB ping emits `CrimeCommitted` and both it and the dependent outlaw trigger reach the stack.

Verification: rebased onto `main`; full `cargo test -p engine --lib` (13553) passes, the two desert regressions pass, `cargo clippy -p engine --all-targets -- -D warnings` clean.

kiannidev:
Thanks — agreed on CR 700.13, and the PR has already been reworked to match it. The earlier `deal_damage.rs` crime emission (the lines the bot flagged) has been **removed**; the current branch's only net engine change is in `triggers.rs`.

Crime is now committed at the correct time — when the triggered ability is **put on the stack** with an opponent target (target-lock), not during damage resolution. The `BecomesTarget`/`CrimeCommitted` events emitted while pushing the ability are now scanned for dependent triggers in the same pass via `process_target_lock_side_effects`, so an auto-targeted Desert ETB ping commits a crime and outlaw abilities (e.g. Vadmir) fire as expected.

The flagged `deal_damage.rs:571-580` / `:5438` snippets no longer exist on the branch. All CI checks are green (clippy, both test shards, parser gate).

kiannidev:
Thanks @matthewevans — addressed the trigger-ordering concern in a2a57a78f.

Target-lock side effects (`CrimeCommitted`, `BecomesTarget`) are now **collected** into `deferred_triggers` via `collect_triggers_into_deferred` instead of recursively calling `process_triggers` mid-batch. They drain at the next legal boundary with `drain_deferred_triggers_after_trigger_construction` once the current APNAP-ordered batch finishes placing siblings.

Added `target_lock_crime_trigger_waits_behind_sibling_etb_triggers`: a simultaneous desert ETB + draw ETB batch with an outlaw on the field asserts the crime trigger lands on the stack **after** both ETB siblings, not interleaved between them.

`cargo fmt --all`, the desert ETB tests, the new ordering regression, and `cargo clippy -p engine -- -D warnings` all pass locally.

matthewevans (commented):
VERDICT: approve-pending-CI

Re-reviewed the current implementation and then ran the maintainer update-branch treadmill because the reviewed head was behind `origin/main`.

The content review is clear on the three requirements: the fix stays at the engine trigger/deferred-target-lock seam, the implementation reuses the existing deferred trigger construction path instead of adding a parallel one-off path, and the new regression is discriminating for the sibling ETB trigger ordering around a target-lock crime trigger.

I merged current `origin/main` into the contributor branch and pushed the merge-only refreshed head `8829e469d6e118c387d7c8e0638025563fc15fbb`. The local push hook got through the visible Rust/parser/AI checks, then failed at `oracle-gen` because this machine is missing `data/mtgjson/AtomicCards.json`; I bypassed that local hook for the merge-only push so GitHub CI remains the ledger.

HOLD: refreshed exact-head CI is now running on `8829e469`, so I am not posting a formal approval, applying labels, or enqueueing until that exact head is green and the repush guard holds.

matthewevans (approved):
Approved after current-head re-review, merge-only update to main, and successful exact-head CI. The prior target-lock crime ordering blocker is addressed by using the existing deferred trigger boundary and sibling ETB ordering coverage.

PR #4295, opened 2026-06-24
fix(engine): Birthing Ritual CMC filter uses sacrificed creature mana value (#4271)

### LINKED ISSUE #4271: Birthing Ritual limits put-onto-battlefield choice to mana value 1
Reported in Discord: https://discord.com/channels/1485498006781427802/1519103580043219105/1519103580043219105

**Thread:** Birthing Ritual
**Cards:** birthing ritual
**Parser status:** has_gaps

## Summary
Birthing Ritual does not let the player put a creature with mana value greater than 1 onto the battlefield, even when the sacrificed creature should allow a higher mana value.

### birthing ritual
At the beginning of your end step, if you control a creature, look at the top seven cards of your library. Then you may sacrifice a creature. If you do, you may put a creature card with mana value X or less from among those cards onto the battlefield, where X is 1 plus the sacrificed creature's mana value. Put the rest on the bottom of your library in a random order.

---
<sub>report_id: `discord:1519103580043219105:1519103580043219105:0` · discord: `1519103580043219105/1519103580043219105`</sub>

nickmopen:
## Summary

Fixes #4271: Birthing Ritual's `DigChoice` was offering creatures up to mana value 1 instead of `sacrificed MV + 1`.

- **Root cause**: The `last_revealed_ids` injection in `resolve_chain_body` was forwarding the 7 looked-at library card IDs as `ability.targets` on the Sacrifice sub-ability. `sacrifice::resolve` saw non-empty `targeted_objects`, iterated over them, but skipped all 7 (CR 701.21a: only battlefield permanents can be sacrificed). No `PermanentSacrificed` event was emitted, leaving `effect_context_object = None` on the downstream `PriorLook` Dig — so `ObjectScope::CostPaidObject` fell back to `unwrap_or(0)`, yielding threshold = `0 + 1 = 1`.
- **Fix**: Added `&& !matches!(sub.effect, Effect::Sacrifice { .. })` to the injection guard in `effects/mod.rs`. Sacrifice always resolves its own battlefield-scoped eligible pool independently and must never receive library card IDs as pre-wired targets.
- **Regression test**: `issue_4271_birthing_ritual_cmc_filter` — sets up a sacrificed creature with MV 3 (expected X = 4), places a MV 4 Angel and a MV 5 Dragon in the top-7 library, and asserts `selectable_cards = [mv4_angel]` only.

## Test plan

- [x] `cargo test -p engine --test integration birthing_ritual_cmc_filter` — passes
- [x] Full integration suite (1528 tests) — all pass

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -6514,6 +6514,13 @@ fn resolve_chain_body(
         } else if sub.targets.is_empty()
             && !state.last_revealed_ids.is_empty()
             && effect_writes_last_revealed_ids(&ability.effect)
+            // CR 701.21a: Sacrifice resolves its own battlefield-scoped eligible
+            // pool — injecting library card IDs from a look-only Dig (e.g.
+            // Birthing Ritual) would route through effect_object_targets and
+            // silently skip every card (Zone::Library ≠ Zone::Battlefield),
+            // producing no PermanentSacrificed event and leaving
+            // effect_context_object = None for the downstream PriorLook Dig.
+            && !matches!(sub.effect, Effect::Sacrifice { .. })
         {
             // Inject revealed card IDs as targets for sub_abilities following
             // effects that write last_revealed_ids. Parallel to how
--- a/crates/engine/tests/integration/issue_4271_birthing_ritual_cmc_filter.rs
+++ b/crates/engine/tests/integration/issue_4271_birthing_ritual_cmc_filter.rs
@@ -1,0 +1,186 @@
+//! Regression test for issue #4271: Birthing Ritual must offer creatures up to
+//! mana value (sacrificed creature's MV + 1), not always mana value 1.
+//!
+//! Previously, the chain ordered the dig choice BEFORE the sacrifice, so when
+//! `selectable_cards` was computed, `effect_context_object` (the sacrificed
+//! creature snapshot) was None. `ObjectScope::CostPaidObject` fell back to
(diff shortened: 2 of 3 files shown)


matthewevans:
VERDICT: approve-pending-CI

I reviewed the current head (`a3a58ae3f995d5463f2f66e68f742bf1d15ce9fd`) via the GitHub API diff, Gemini's review, and an isolated worktree.

The fix is at the right seam: the existing `last_revealed_ids` sub-ability injection in `resolve_chain_body` was feeding look-only library card ids into a later `Effect::Sacrifice`, which sends `sacrifice::resolve` down its targeted-object path where non-battlefield objects are skipped. Guarding sacrifice there preserves the revealed-card target injection for the later Dig consumer while letting sacrifice use its existing battlefield-scoped eligible-pool resolver. The regression test drives the production end-step trigger through optional sacrifice into `WaitingFor::DigChoice` and proves the MV 4 creature is selectable while MV 5 is not, so it covers the actual value path rather than just parser or helper shape.

Gemini's `.ok()` test-cleanup suggestion is non-blocking for this head because the final `DigChoice` assertions still fail if trigger ordering/priority progression does not reach the production path.

Holding formal approval/label/enqueue until exact-head CI finishes green and the repush guard holds.

matthewevans (approved):
Approved current head 8e7627f4629a036d4288b650c84bc635922493f1 after exact-head CI completed green and the repush guard held.

Issue #4266, opened 2026-06-24
Forked Bolt — Forked Bolt deals two damage to each target, it should divide damage.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1519107452786708480/1519107452786708480

**Thread:** Forked Bolt
**Cards:** forked bolt
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Forked Bolt deals two damage to each target, it should divide damage.

### forked bolt
Forked Bolt deals 2 damage divided as you choose among one or two targets.

---
<sub>report_id: `discord:1519107452786708480:1519107452786708480:0` · discord: `1519107452786708480/1519107452786708480`</sub>

shin-core:
Picking this up. Confirmed the parser is correct — Forked Bolt lowers to `DealDamage` with `distribute: Damage` and `multi_target: {min:1, max:2}`, so the gap is at runtime: the divided-damage distribution (`WaitingFor::DistributeAmong`) isn't being applied for the variable-target-count ("one or two targets") spell-cast path, so it falls through to uniform damage. Investigating the cast-distribution seam and will open a PR with a fix + tests.

shin-core:
Correction / update: on closer testing against current `main`, Forked Bolt actually works correctly now. A scenario repro (cast → select two targets) pauses for `WaitingFor::DistributeAmong` (the divide-as-you-choose prompt) with no damage dealt until the division is submitted — the intended behavior. This appears to have been fixed since the report was filed. Unclaiming; leaving open for a maintainer to confirm and close if reproduction is no longer possible.

mike-theDude:
Verified Oracle text for Forked Bolt: "Forked Bolt deals 2 damage divided as you choose among one or two targets." This parses correctly as a `DealDamage` effect with `multi_target: {min:1, max:2}` and `distribute: Damage` — the "divided as you choose" shape is recognized, so this is not a parser gap.

[...]

PR #4299, opened 2026-06-24
fix(parser): divided damage pattern A fails to match due to leading whitespace (#4266)

nickmopen:
## Root cause

`dispatch_line_nom` in `oracle_dispatch.rs` called `parse_effect_chain_with_context` — which correctly sets `distribute: Some(Damage)` and `multi_target: Some(fixed(1, 2))` on the returned `AbilityDefinition` — then returned `*def.effect`, discarding every field beyond the `Effect` itself.

The caller in `parse_oracle_ir` (oracle.rs:3912) then wrapped that bare `Effect` in `AbilityDefinition::new(AbilityKind::Spell, nom_effect)`, so `distribute` and `multi_target` were permanently `None` for any spell parsed through this path.

With `distribute: None`, `PendingCast.distribute` was `None` at casting time, so `casting_targets` never generated `WaitingFor::DistributeAmong`. Forked Bolt resolved via the non-distribution path and applied the full 2 damage to every selected target instead of splitting it.

## Fix

Change `dispatch_line_nom` to return the full `AbilityDefinition` (not just the `Effect`). The call site in `parse_oracle_ir` pushes it directly without rewrapping, preserving all parsed fields.

## Files changed

- `parser/oracle_dispatch.rs` — return type `Effect` → `AbilityDefinition`; return `def` on success; wrap unimplemented cases in `AbilityDefinition`; add fail-on-revert test
- `parser/oracle.rs` — push `nom_def` directly instead of rewrapping
- `parser/oracle_effect/mod.rs` — clarify that the clause-level test covers `try_parse_distribute_damage` in isolation, with a pointer to the new dispatch-level regression test

## Tests

`oracle_dispatch::tests::dispatch_line_nom_preserves_distribute_and_multi_target_for_divided_damage` — fail-on-revert: if the return type reverts to `-> Effect`, the `.distribute`/`.multi_target` field accesses cause a compile error, not just an assertion failure.


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** This does not prove or fix the claimed leading-whitespace failure; the new test appears to pass on the parent. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:4542` receives `rem` from `parse_count_expr(after_tp.lower)`, and the fixed-number path in `parse_count_expr` delegates to `parse_number`; `crates/engine/src/parser/oracle_util.rs:367` already returns `rest.trim_start()` for parsed numbers, with the same behavior present in this PR's parent. For `"2 damage divided..."`, the old code therefore already sees `rem == "damage divided..."`, so `tag("damage")` succeeds before this hunk. Why it matters: the added Forked Bolt parser-shape test is not fail-on-revert coverage for #4266, and the production change is effectively a no-op for the stated bug, so approving it would record false confidence without addressing the real failure path. Suggested fix: reproduce the actual #4266 current-main failure through the production cast/selection path that reaches `WaitingFor::DistributeAmong`, then patch that seam; if you only want a parser cleanup here, make the `damage` check a normal nom combinator cleanup without claiming it fixes the bug.

Gemini's nom-combinator cleanup comment is directionally fine, but the blocking issue is stronger: the premise of the fix contradicts the existing `parse_count_expr` contract, so there is no approval/label/enqueue on this head.

matthewevans:

Current-head re-review: the previous whitespace-premise blocker is addressed by moving the fix to the actual metadata-loss seam. `dispatch_line_nom` now returns the full `AbilityDefinition`, and `parse_oracle_ir` pushes that definition directly, so `distribute` and `multi_target` from `parse_effect_chain_with_context` survive into the existing cast path. I traced the consumer side as well: casting copies `AbilityDefinition.distribute` into `PendingCast.distribute`, which is what drives `WaitingFor::DistributeAmong`.

The new dispatch-level Forked Bolt regression is fail-on-revert for that seam: reverting the dispatch return type back to `Effect` makes the `.distribute` / `.multi_target` assertions impossible. Gemini's original nom-combinator cleanup comment was valid for the old head, but it is stale for the current approach because the touched production seam is no longer the `damage` tag parsing branch.

I also pushed a narrow maintainer hygiene fix on top after the local parser gate caught direct `Effect::Unimplemented` construction in the new fallback arms; those now use `Effect::unimplemented(...)` like the repo expects. No formal approval, label, or enqueue yet: the exact `b188fc3d7abd1dea70452a2512d099974beee080` head still needs current green CI and the repush guard before maintainer approval.

matthewevans:
Update: exact-head CI went green on `b188fc3d`, but `main` advanced before approval. I ran the update-branch treadmill and pushed the merge-only current head `ddae03ec6d2a8bae1b1ee0d5ad085c5def395023`.

Still holding formal approval/label/enqueue until CI is green on this updated head and the repush guard holds.

matthewevans (approved):
Approved on current head after re-review and green exact-head CI. The fix is at the dispatch metadata seam and the regression is fail-on-revert for preserving `distribute`/`multi_target`.

PR #4303, opened 2026-06-24
fix(engine): autotap accounts for TapsForMana aura bonus (Wild Growth, Fertile Ground, Utopia Sprawl)

### LINKED ISSUE #4265: Wild Growth Autotap — Autotap doesn't consider Wild Growth or utopia sprawl style effects when calculating your mana, s…
Reported in Discord: https://discord.com/channels/1485498006781427802/1519109536420593734/1519109536420593734

**Thread:** Wild Growth Autotap
**Cards:** utopia sprawl, wild growth

## Summary
Autotap doesn't consider Wild Growth or utopia sprawl style effects when calculating your mana, so it will autopass when you can cast spells and overtap for spells when you have those cards in play.

### utopia sprawl
Enchant Forest
As this Aura enters, choose a color.
Whenever enchanted Forest is tapped for mana, its controller adds an additional one mana of the chosen color.
### wild growth
Enchant land
Whenever enchanted land is tapped for mana, its controller adds an additional {G}.

---
<sub>report_id: `discord:1519109536420593734:1519109536420593734:0` · discord: `1519109536420593734/1519109536420593734`</sub>

nickmopen:
## Summary

- `land_mana_options` previously called `scan_mana_abilities` on the land only, so a Forest enchanted by Wild Growth was planned as a `{G}` source instead of the `{G}{G}` it actually produces (CR 605.1b + CR 106.12a).
- New `taps_for_mana_aura_bonus` iterates the battlefield to find TapsForMana-triggered auras whose `valid_card` filter matches the tapped land, reusing `taps_for_mana_card_matches` — the same predicate the trigger resolver uses — so planning and firing cannot drift.
- Bonus types are folded into `atomic_combination` so the autotap planner treats the land's own mana plus the aura trigger as a single atomic event from one tap.
- `display_land_mana_pips` and `max_mana_yield` receive the same bonus so the land frame UI and X-value choosers reflect the full yield.
- Covers Wild Growth, Fertile Ground, Utopia Sprawl, Verdant Haven, and any future aura with the same `TapsForMana + AttachedTo` trigger shape.

Fixes #4265.

## Test plan

- [ ] `aura_bonus_detects_taps_for_mana_trigger_on_attached_aura` — `taps_for_mana_aura_bonus` returns `[Green]` for a Forest + Wild Growth
- [ ] `aura_bonus_empty_for_unenchanted_land` — no bonus for a bare Forest
- [ ] `land_mana_options_includes_wild_growth_aura_bonus_as_atomic_combination` — end-to-end: single option with `atomic_combination = [Green, Green]`
- [ ] `max_mana_yield_counts_wild_growth_aura_bonus` — yield goes 1 → 2 when aura is attached
- [ ] `display_pips_includes_wild_growth_aura_bonus` — pips include green after attaching aura
- [ ] CI green (`test-engine`, `clippy`)


matthewevans (changes requested):
VERDICT: request-changes

**[HIGH]** `source_could_produce_two_or_more_colors` is set from mana quantity instead of distinct color capability. Evidence: `crates/engine/src/game/mana_sources.rs:1290-1303` appends the Wild Growth bonus to `atomic_combination` and then unconditionally sets `opt.source_could_produce_two_or_more_colors = true`; `crates/engine/src/game/casting_costs.rs:6209-6210` and `crates/engine/src/game/engine.rs:6169` consume that flag for `{Z}`/two-or-more-color-source requirements. Why it matters: a Forest plus Wild Growth produces `{G}{G}`, but this PR makes it satisfy a requirement that the source could produce two or more colors. Suggested fix: compute the flag from the distinct non-colorless colors the combined source can produce, not from `combined.len()`/a non-empty aura bonus.

**[HIGH]** Opponent-controlled TapsForMana auras attached to your land are ignored by the planner. Evidence: `crates/engine/src/game/mana_sources.rs:1798-1803` skips any trigger source whose `obj.controller != controller`, even though `taps_for_mana_card_matches` at `crates/engine/src/game/trigger_matchers.rs:2752-2762` only asks whether the tapped land matches the trigger's card filter; the mana event player/controller is the land tapper, not necessarily the aura controller. Why it matters: a legal aura attached to your land can trigger when you tap that land regardless of the aura object's controller, so autotap can still undervalue or miss mana in the exact class this PR is trying to cover. Suggested fix: filter by the tapped land and trigger applicability, and resolve the produced mana for the tapping player while preserving any source-controller semantics that the trigger resolver actually uses.

**[MED]** The display-pip regression does not prove the claimed duplicate pip behavior. Evidence: `display_land_mana_pips` deduplicates through the local `push` helper at `crates/engine/src/game/mana_sources.rs:473-478`, the aura bonus path also uses that helper at `crates/engine/src/game/mana_sources.rs:655-663`, and the new test at `crates/engine/src/game/mana_sources.rs:3439-3447` only asserts `contains(Green)`. Why it matters: the test passes even if the Forest + Wild Growth frame still shows one green pip, contradicting the PR's stated UI/value claim. Suggested fix: decide whether duplicate same-color pips are part of the display contract; if yes, stop deduping that path and assert the exact two-pip vector, otherwise remove the claim.

matthewevans (changes requested):
VERDICT: request-changes

The prior Matt/Gemini blockers are addressed on this head: the two-or-more-colors flag is now distinct-color based, opponent-controlled attached auras are no longer filtered out, and the duplicate-pip display path/test now proves two same-color pips.

**[HIGH]** Fertile Ground-style `AnyOneColor` aura bonuses are collapsed to a fixed first color before being folded into the land's atomic output. Evidence: `crates/engine/src/game/mana_sources.rs:1837-1841` calls `effects::mana::resolve_mana_types` on each attached `TapsForMana` aura, then `crates/engine/src/game/mana_sources.rs:1292-1314` appends that concrete `Vec<ManaType>` to every `atomic_combination`; but `crates/engine/src/game/effects/mana.rs:431-440` resolves `ManaProduction::AnyOneColor` by taking `color_options.first()`. `ManaProduction::AnyOneColor` is the typed shape for Fertile Ground's "one mana of any color" class (`crates/engine/src/types/ability.rs:1272-1279`), while the new tests only construct `ManaProduction::Fixed` Wild Growth-style auras at `crates/engine/src/game/mana_sources.rs:3288-3315`. Why it matters: a Forest with Fertile Ground should support `{G}` plus one chosen color, but this planner/display path records it as a fixed first-color bonus, so it can miss payable costs in other colors while claiming the PR covers Fertile Ground. Suggested fix: preserve the aura bonus choice shape in the mana source option model, or expand `land_mana_options` into per-choice atomic options for `AnyOneColor`/chosen-color bonuses, and add a Fertile Ground regression that pays a non-first bonus color through the autotap/castability path.

crates/engine/src/game/mana_sources.rs
```diff
@@ -1262,6 +1306,53 @@ fn land_mana_options(
...
+    // Each aura fires atomically with the land's tap — so we extend each
+    // option's `atomic_combination` rather than adding a second option (which
+    // would let the planner double-tap the same land).
+    //
+    // For Fixed auras (Wild Growth: {G}): one bonus type → no fan-out, same
+    // number of options.
+    // For AnyOneColor auras (Fertile Ground: any color): N bonus choices →
+    // fan-out into N options per base option, one per reachable color, so the
+    // planner can pick whichever color satisfies the pending cost.
+    for aura_choices in taps_for_mana_aura_bonus(state, object_id, controller) {
```
matthewevans: This now creates the right planner rows for `AnyOneColor` aura bonuses, but I don't see the selected bonus color carried into actual resolution. For basic-land fallback, `auto_tap_mana_sources_inner` still produces only `option.mana_type` for the land and emits `TappedForMana`; `resolve_tap_mana_triggers_inline` then resolves the aura trigger without a `ProductionOverride`, so `AnyOneColor` defaults to its first listed color. A production-path regression should prove that a non-first Fertile Ground bonus color is actually produced/spent, or the selected aura choice needs to be threaded through trigger resolution.

matthewevans (changes requested):
Thanks for the follow-up. The previous helper-level issues look addressed: the aura controller filtering is gone, `{G}+{G}` no longer gets the two-or-more-colors flag, display pips now preserve the fixed aura bonus, and `land_mana_options` now fans out Fertile Ground-style `AnyOneColor` choices.

I still can’t approve this head because the fan-out appears to stop at the planner row. The actual auto-tap path still taps the basic-land fallback for the land’s own mana, emits `TappedForMana`, and then resolves the aura’s `TapsForMana` triggered mana ability independently. That triggered ability is resolved without a `ProductionOverride`, so `AnyOneColor` falls back to its first listed color. This is exactly why the prior request asked for a Fertile Ground regression through the auto-tap/castability/payment path, not only helper/option coverage.

Please either thread the selected aura bonus color through the inline `TapsForMana` trigger resolution, or add production-path coverage that proves a non-first Fertile Ground bonus color is actually produced and spent by auto-tap/castability. As written, the planner can advertise a `{G}{U}` option while the resolver can still add `{G}` plus the default first color instead.

matthewevans (approved):
VERDICT: approve

Re-reviewed current head `80e8f47005fe5fb9df0497316779163fe68d65e6` after the latest substantive follow-up.

The prior Gemini/Matt blockers are addressed:
- `source_could_produce_two_or_more_colors` is now based on distinct colors in the combined output, so `{G}+{G}` no longer satisfies two-or-more-color-source requirements.
- Opponent-controlled attached `TapsForMana` auras are no longer filtered out before the shared `taps_for_mana_card_matches` check.
- Fixed aura display pips bypass the dedup path and the test now proves duplicate same-color pips.
- Fertile Ground-style `AnyOneColor` bonuses fan out into per-choice planner rows and now carry the selected bonus color through `pending_taps_for_mana_overrides` into inline `TapsForMana` resolution.

The new production-path regression exercises the exact prior failure: Forest + Fertile Ground paying `{G}{U}` produces the non-first `{U}` bonus rather than defaulting to the first listed color. Exact-head CI is green and the repush guard held.

PR #4307, opened 2026-06-25
fix(ai): clamp SacrificeValue + scaled policy deltas into the critical band

### LINKED ISSUE #4282: Non-fatal engine panic: policy SacrificeValue scaled delta -21.360384941101074 exceeds critical band …
**What happened**
_briefly describe what you were doing_

**Diagnostic**
```
Build: v0.4.0 (33f6c366a)
Diagnostic tag: getState-panic
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36

Panic (engine recovered):
panicked at crates\phase-ai\src\policies\registry.rs:405:21: policy SacrificeValue scaled delta -21.360384941101074 exceeds critical band ceiling 15
```

dripsmvcp:
Fixes #4282

## Summary
`SacrificeValuePolicy::verdict` returned its raw `score` — the summed value of every sacrificed permanent — directly as the verdict delta via a `PolicyVerdict::Score { .. }` struct literal, bypassing the band-clamping constructors (`PolicyVerdict::score`/`critical`) that every other policy uses to keep deltas inside the critical band. A costly multi-permanent sacrifice therefore produced an out-of-band delta (the report shows `-21.360…`), and the registry's per-candidate scaling `debug_assert!`-panicked on it (`policy SacrificeValue scaled delta … exceeds critical band ceiling 15`) in builds with debug-assertions enabled.

## Fix
- **`SacrificeValuePolicy::verdict`** now routes its raw score through the shared `PolicyVerdict::score` constructor, clamping the policy's own delta into the critical band like every peer policy.
- **Registry scaling invariant corrected.** The old `debug_assert!` checked the *scaled* delta (`delta * activation`), which is the wrong invariant: `activation` can legitimately exceed `1.0` (turn-phase / archetype multipliers from `turn_phase_mult`), so an in-band raw delta can scale past the ceiling without any policy misbehaving. The assertion now checks the **raw** delta (the actual per-policy contract), and the **scaled** delta is **clamped** into the band — a single policy never exceeds the critical-band ceiling of influence — instead of being passed through unclamped (release) or panicked on (debug-assertions). A `tracing::warn!` still records any clamp for decision-trace telemetry.

## Tests
- New regression `high_cost_sacrifice_verdict_stays_within_critical_band` drives a high-cost sacrifice past the band (asserting the raw score is genuinely out of band first, so the clamp is actually exercised) and asserts the verdict delta stays within `±CRITICAL_MAX` while remaining a negative incentive.
- Ratchets the `score_contract_lint` legacy `PolicyVerdict::Score` literal count for `sacrifice_value.rs` from 1 → 0 (one fewer band-bypassing site).
- `cargo fmt --all`, `cargo clippy -p phase-ai --all-targets -- -D warnings`, and `cargo test -p phase-ai` (1095 lib tests + integration) all green.

### DIFF
--- a/crates/phase-ai/src/policies/registry.rs
+++ b/crates/phase-ai/src/policies/registry.rs
@@ -401,25 +401,40 @@ impl PolicyRegistry {
             let scaled = match verdict {
                 PolicyVerdict::Reject { reason } => PolicyVerdict::Reject { reason },
                 PolicyVerdict::Score { delta, reason } => {
-                    let scaled_delta = delta * activation as f64;
+                    // The critical band is a contract on each policy's *raw*
+                    // delta — the `PolicyVerdict::score`/`critical` constructors
+                    // clamp magnitudes to `CRITICAL_MAX`. A raw delta past the
+                    // ceiling means a policy bypassed them and emitted an
+                    // out-of-band value; assert that here (the meaningful
+                    // invariant). The prior code asserted on the *scaled* value,
+                    // which is wrong: `activation` can exceed 1.0 (turn-phase /
+                    // archetype multipliers), so an in-band raw delta can scale
+                    // past the ceiling legitimately — yet the assert panicked on
+                    // it in builds with debug-assertions enabled (issue #4282).
                     debug_assert!(
-                        scaled_delta.abs() <= CRITICAL_MAX,
-                        "policy {:?} scaled delta {} exceeds critical band ceiling {}",
+                        delta.abs() <= CRITICAL_MAX,
+                        "policy {:?} raw delta {} exceeds critical band ceiling {}",
                         policy_id,
-                        scaled_delta,
+                        delta,
                         CRITICAL_MAX
                     );
-                    if scaled_delta.abs() > CRITICAL_MAX {
+                    // A single policy must never exceed the critical-band ceiling
+                    // of influence over the aggregate, so clamp the scaled value
+                    // into the band rather than letting an activation multiplier
+                    // push it past the ceiling and dominate the sum.
+                    let scaled_delta = delta * activation as f64;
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** The registry scaled-delta fix is not covered by a discriminating test. Evidence: `crates/phase-ai/src/policies/registry.rs:401-438` moves the invariant to the raw policy delta and clamps `delta * activation`, but the new regression in `crates/phase-ai/src/policies/sacrifice_value.rs:193-278` only calls `SacrificeValuePolicy.score()`/`verdict()` directly. It never drives `PolicyRegistry::verdicts()` with `SacrificeValuePolicy` active on an early turn where `turn_only` returns `1.3`, so reverting the registry hunk would leave the added test green while reintroducing the debug assertion panic path once an in-band raw `-15` gets scaled to `-19.5`. Why it matters: the PR claims two fixes, raw SacrificeValue clamping and registry scaled-delta clamping; only the first is currently fail-on-revert. Suggested fix: add a registry-level regression that constructs the high-cost sacrifice context on an early turn, runs `PolicyRegistry::default().verdicts(&ctx)`, and asserts the SacrificeValue verdict remains in-band without panicking under debug assertions.

Gemini's `scaled_delta != clamped`/NaN warning is a small telemetry precision issue, not the merge blocker here.

matthewevans:
VERDICT: superseded

This PR is now superseded by #4309, which landed the #4282 policy-band fix on `main`. Please close this branch unless you want to rebase it down to a net-new test or behavior delta that is not already covered by the merged fix.

PR #4311, opened 2026-06-25
test(engine): regression test for Once Upon a Time first-spell free cast (#787)

### LINKED ISSUE #787: Once Upon a Time — I haven't cast a spell yet this game and it's not castable for free as it should be
Reported in Discord: https://discord.com/channels/1485498006781427802/1507092806207016960/1507092806207016960

**Thread:** Once Upon a Time
**Cards:** _none detected_

## Summary
I haven't cast a spell yet this game and it's not castable for free as it should be

---
<sub>report_id: `discord:1507092806207016960:1507092806207016960:0` · discord: `1507092806207016960/1507092806207016960`</sub>

kiannidev:
## Summary

Closes #787. Once Upon a Time — "If this spell is the first spell you've cast this game, you may cast it without paying its mana cost." — was reported as not castable for free as the first spell of the game.

Investigating against current `main`, the full pipeline is already in place and correct:

- **Parser** emits `SpellCastingOption::CastWithoutManaCost` gated by `ParsedCondition::FirstSpellThisGame`.
- **Synthesis** lowers it onto `CardFace.casting_options`; **rehydration** copies it to `GameObject.casting_options` (`printed_cards.rs`).
- **Casting** evaluates `FirstSpellThisGame` (`spells_cast_this_game[player] == 0`) so the free alternative cost (CR 601.2b/118.9) makes the spell castable with no mana.

What was missing was **end-to-end coverage** for this card, so a future regression would go unnoticed. This PR adds that.

## Changes

- New integration test `issue_787_once_upon_a_time.rs` driving the real database-parsed card.
- Adds `Once Upon a Time` to the integration card fixture (`integration_cards.json`).

## Test plan

The test asserts, using the real parsed card via `shared_card_db()`:

- [x] With no mana and no spell cast this game, `can_cast_object_now` is `true` (the reported symptom).
- [x] Casting offers `OptionalCostChoice { Choice(NoCost, printed) }` and, when the free branch is chosen, the spell lands on the stack.
- [x] After a spell has been cast this game, the free-cast window closes — with no mana it is uncastable (discriminates the gate from an unconditional free cast).
- [x] `cargo fmt --all -- --check`, `cargo clippy -p engine --tests`, parser/engine-authority/test-card-data-load gates
- [x] Full `cargo test -p engine --test integration` green (1534 passed)


matthewevans (changes requested):
Thanks for narrowing this to proof coverage. I’m holding this as a test-quality issue rather than an implementation concern.

[HIGH] The regression proof manually mutates `spells_cast_this_game` to show the gate closes, but it never proves that the real `CastSpell` path increments that counter when Once Upon a Time is cast. That leaves a meaningful regression untested: a future change could keep this test green while allowing first-spell-gated free casts to remain available after the actual cast pipeline runs. Gemini’s comment is right on this point. Please assert the counter after `DecideOptionalCost { pay: true }` (or otherwise prove a second first-spell-gated free cast is unavailable through the real pipeline), so the test satisfies the PR proof bar.

kiannidev:
Thanks — addressed in the latest push (also rebased onto current `main`).

The regression now proves the gate closes through the **real** cast pipeline, not just the hand-mutated counter:

- After `DecideOptionalCost { pay: true }` puts Once Upon a Time on the stack, assert `spells_cast_this_game[P0] == 1` (the actual `CastSpell` path recorded it).
- Added a second copy to hand and assert `can_cast_object_now` returns `false` for it with no mana — so a future regression that fails to record the cast (leaving first-spell-gated free casts available) would now fail this test.

`cargo fmt`, the `issue_787` test, and the minimal fixture diff (adds only `once upon a time`) are green locally.

matthewevans (commented):
VERDICT: approve-pending-CI

The follow-up addresses the proof blocker from my prior review. The regression now asserts the real cast pipeline increments `spells_cast_this_game` after choosing the free alternative, and the second-copy `can_cast_object_now == false` check proves the first-spell-gated free-cast window closes through the actual cast rather than a manually-mutated counter. Gemini’s valid concern is addressed.

Held for exact-head CI only. No formal approval, label, or enqueue until CI is green/current and the repush guard holds.

matthewevans (approved):
Approved after exact-head CI went green and the follow-up addressed the proof blocker. The test now proves the real cast pipeline increments `spells_cast_this_game` and that a second Once Upon a Time is not free-castable with no mana, so Gemini's concern is resolved.

PR #4312, opened 2026-06-25
perf(combat): hoist O(battlefield²) static scan out of legality loops

matthewevans:
## Problem

Combat block/attack legality is **O(battlefield²)** on large boards. The legal-blocker/attacker validation loops in `combat.rs` iterate all of `state.battlefield`, and for each candidate previously called `blocker_has_cant_block_static` / `block_restriction_statics_against` — each of which invokes the **uncached** `game_functioning_statics` / `battlefield_functioning_statics` iterators (a full battlefield walk + per-object `static_definitions` flat-map, with `matches_target_filter` nested inside).

Profiled on a real checkpoint (The Unbeatable Squirrel Girl deck — **702 Squirrel tokens, 760 battlefield permanents**, turn 39): a human priority pass took **1.77s** for 4 legal actions; AI `choose_action` took **12.8s**. `sample` self-time was dominated by `game_functioning_statics` and the combat block-restriction stack.

## Fix

Hoist the loop-invariant "find statics of the relevant `StaticMode` across the battlefield" step **out** of the per-candidate loops. Collect the small CantBlock / CantBeBlocked\* / BlockRestriction / MustBeBlocked\* sets **once per legality pass** into owned `Vec<(ObjectId, StaticDefinition)>` slices, then per candidate run only the affected-filter + condition check against that tiny precomputed set via `*_from_precomputed` variants.

**Rules-correctness preserved (pure perf refactor):** the precomputed iterators clone `StaticDefinition` but re-resolve the source `ObjectId` **live** each iteration (`state.objects.get(src_id)`), reading only `.id` and a live-re-read `.controller` — so controller-dependent affected-filters and every legality verdict are byte-for-byte identical. Public wrappers (`can_block_pair`, `min_blockers_required`, `max_blockers_allowed`, `block_requirements_for_player`) are retained as thin collect+delegate shims, so the single legality authority and all external callers are unchanged. `phase-ai` threads a `BlockLegalitySlices` snapshot through the AI attacker-selection and block-evaluation paths.

`functioning_abilities.rs` is untouched.

## Measured impact (same checkpoint)

| Metric | Before | After | Δ |
|---|---|---|---|
| raw mean | 275 ms | 92 ms | **−67%** |
| simulation filter | 1.36 s | 399 ms | **−71%** |
| full mean | 1.77 s | 711 ms | **−60%** |
| `choose_action` | 12.8 s | 6.2 s | **−51%** |

## Scope / follow-ups

The residual `choose_action` time is dominated by two **separate, deliberately out-of-scope** costs: the board-global mana-display sweep (the known `dirty-perf` landmine) and per-candidate search simulation. Both warrant their own follow-up tickets and are not addressed here.

## Tests

3 new discriminating tests (precomputed-slice rejection/acceptance, two distinct CantBlock sources both enforced, unblockable attacker still chosen through the real AI path) plus the full existing suite green: 221 engine combat + 62 `combat_ai` + 501 policy tests.

CR annotations: `CR 509.1a/b/c`, `604.1` preserved; the two CantBlock helpers' annotations corrected from `509.1a` (choosing untapped creatures) to `509.1b` (can't-block restrictions).

### DIFF
--- a/crates/engine/src/game/combat.rs
+++ b/crates/engine/src/game/combat.rs
@@ -663,6 +663,7 @@ pub fn collect_must_be_blocked_statics(state: &GameState) -> Vec<(ObjectId, Stat
         .collect()
 }
 
+/// CR 509.1b: Block restriction — these statics make a block declaration illegal.
 /// Re-resolve a precomputed `CantBeBlocked*` static against `attacker_id`,
 /// applying the SAME `affected` + `condition` filter stack as
 /// `block_restriction_statics_against` but without re-walking the battlefield.
@@ -702,6 +703,7 @@ fn block_restriction_statics_against_from_precomputed<'a>(
     })
 }
 
+/// CR 509.1b: Blocker-side restriction ("~ can't block").
 /// Precomputed counterpart of `blocker_restriction_statics_for`.
 fn blocker_restriction_statics_for_from_precomputed<'a>(
     state: &'a GameState,
@@ -735,6 +737,7 @@ fn blocker_restriction_statics_for_from_precomputed<'a>(
     })
 }
 
+/// CR 509.1b: Block-restriction exception ("~ can't block except …").
 /// Precomputed counterpart of `blocker_block_allowed_statics_for`.
 fn blocker_allowed_statics_for_from_precomputed<'a>(
     state: &'a GameState,
@@ -768,6 +771,7 @@ fn blocker_allowed_statics_for_from_precomputed<'a>(
     })
 }
 
+/// CR 509.1c: Block requirement ("~ must be blocked if able").
 /// Precomputed counterpart of `must_be_blocked_statics_for_attacker`.
 fn must_be_blocked_statics_for_attacker_from_precomputed<'a>(
     state: &'a GameState,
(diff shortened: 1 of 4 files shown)


lgray:
Isn't this matching problem (basically a binary clustering) in general smaller than O(n^2), if you consider P/T as a 2d matching grid you can make this NlogN at worst with a kd-tree, better with other more modern algorithms. (abilities make this more complex in consideration but basic P/T matching constrained on minimizing loss of presence/life is not so bad)

matthewevans:
Thanks @lgray — good observation, and you're pointing at something real, but it's a different cost than this PR targets, so let me draw the line clearly.

This PR's O(battlefield²) was **redundant recomputation**: each per-(blocker, attacker) legality check re-walked the entire battlefield collecting block-restriction statics. The fix hoists that scan to once per pass. Block *legality* itself is decided by keyword/static predicates — can't-be-blocked-by, menace, must-be-blocked, block restrictions (CR 509.1a–c) — not by power/toughness, so there's no P/T geometry to index on that path. A 1/1 legally blocks a 100/100.

Where your framing **does** land is the AI's block *value* decision — which of the legal blockers to assign to trade favorably. That's a genuine combinatorial optimization, and 2D-dominance queries on (power, toughness) are exactly what a kd-tree / range-tree does well. Two things keep it off the critical path today: keywords (deathtouch, trample, first strike, indestructible, protection) distort the clean P/T plane, and realistic combats are small (n typically < 20–40), where a flat O(n²) scan beats kd-tree construction on constant factors. The pathological regime is huge token boards — and there the dominant cost was exactly the static re-walk this PR removes.

I'll profile the assignment search to see whether there's a real win hiding at scale; if so I'll open a follow-up. Appreciate the pointer.

PR #4318, opened 2026-06-25
Fix Memory's Journey only allowing one target card (#4255)

### LINKED ISSUE #4255: Memory's Journey only allows one target card
Reported in Discord: https://discord.com/channels/1485498006781427802/1518710531333161061/1518710531333161061

**Thread:** Memorys journey only allowed one target
**Cards:** memory's journey
**Extraction confidence:** follow-up/manual triage

## Summary
Memory's Journey should allow up to three target cards from the targeted player's graveyard, but the game only allows one target card.

### memory's journey
Target player shuffles up to three target cards from their graveyard into their library.
Flashback {G} (You may cast this card from your graveyard for its flashback cost. Then exile it.)

---
<sub>report_id: `discord:1518710531333161061:1518710531333161061:0` · discord: `1518710531333161061/1518710531333161061`</sub>

kiannidev:
## Summary
Fixes #4255. Memory's Journey reads "Target player shuffles **up to three target cards** from their graveyard into their library," but the engine only surfaced a single target slot.

Root cause: the targeted-shuffle-to-library parser dropped the "up to N target" quantifier. `parse_target` strips the quantifier and returns only the filter, so the `MultiTargetSpec` count was never captured on the `ChangeZone` clause. Additionally, in the subject-shift path ("Target player shuffles …"), the inner `ChangeZone` sub-ability was built without copying the clause's `multi_target`, so even when present the count landed on the singular player `TargetOnly` wrapper instead of the card slot.

## Fix
- Capture the "up to N target" count as a `MultiTargetSpec` in the `TargetedChangeZoneToLibrary` shuffle path via the existing `strip_optional_target_prefix` building block (CR 115.1d).
- Propagate that `multi_target` onto the inner `ChangeZone` sub-ability when a subject-shifted player-target wrapper is built, so the card-target count rides the `ChangeZone` slot rather than the player slot.

The engine already surfaces N target slots and moves all chosen cards when `ChangeZone` carries `multi_target` (cf. the Pit of Offerings flow), so this is purely a parse-time count-propagation fix at the correct seam.

## Files changed
- `crates/engine/src/parser/oracle_ir/ast.rs`
- `crates/engine/src/parser/oracle_effect/imperative.rs`
- `crates/engine/src/parser/oracle_effect/mod.rs`

## CR references
- CR 115.1d — "up to N" target counts (zero-to-N selection)

## Track
Developer

## LLM
Model: claude-opus-4.8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine --lib parser::` — 6703 pass / 0 fail (includes new regression test `memorys_journey_up_to_three_targets_survives_to_change_zone`, which drives the full `parse_effect_chain` pipeline and fails if reverted)

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -5644,6 +5644,12 @@ pub(super) fn parse_shuffle_ast(text: &str, lower: &str) -> Option<ShuffleImpera
                 value((), alt((tag("all "), tag("each ")))).parse(input)
             })
             .is_some();
+            // CR 115.1d: "shuffle up to N target cards from <zone> into <library>"
+            // (Memory's Journey) carries an "up to N" count on the target slot.
+            // Recover it as a `MultiTargetSpec` so the lowering surfaces N slots;
+            // `parse_target` itself strips the quantifier and only returns the
+            // filter. The mass-move (`all`) form ignores it.
+            let (_, multi_target) = super::strip_optional_target_prefix(after_shuffle);
             let (target, _) = parse_target(after_shuffle);
             let origin = if nom_primitives::scan_contains(lower, "graveyard") {
                 Some(Zone::Graveyard)
@@ -5658,6 +5664,7 @@ pub(super) fn parse_shuffle_ast(text: &str, lower: &str) -> Option<ShuffleImpera
                 target,
                 origin,
                 all,
+                multi_target: if all { None } else { multi_target },
             });
         }
     }
@@ -5740,6 +5747,7 @@ pub(super) fn lower_shuffle_ast(ast: ShuffleImperativeAst) -> ParsedEffectClause
             target,
             origin,
             all,
+            multi_target,
         } => {
             // CR 400.6: "shuffle all <filter> from <zone> into your library"
             // (Elixir) is a mandatory mass move of every eligible object — no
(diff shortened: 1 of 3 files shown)


matthewevans (commented):
VERDICT: approve-pending-CI

I reviewed head `48b517359718653145d0198e212364253bef606f` via the GitHub API patch, the isolated worktree, and the Gemini thread. The fix is at the right seam: the shuffle imperative AST now carries the `MultiTargetSpec`, and lowering moves that spec onto the inner `ChangeZone` sub-ability so Memory's Journey has the singular player target plus the variable card-target set. Existing target-slot code already recurses through sub-abilities and expands finite multi-target specs, so the changed parser/lowering path is the correct place to prove this.

Gemini's note about reusing the stripped remainder is not a blocker on this head. `parse_target` is already the target-grammar authority that handles optional target prefixes; the separate `strip_optional_target_prefix` call is only recovering the count. A small cleanup would be fine, but I don't see a correctness issue from the current shape.

Holding formal approval, labeling, and enqueue until exact-head CI finishes green/current and the repush guard still holds.

matthewevans (approved):
Approved at exact head `48b517359718653145d0198e212364253bef606f` after CI completed green and the repush guard held. The Memory's Journey fix is at the parser/lowering seam, extends to the targeted-shuffle card class shown by parse-diff, and keeps runtime target-slot behavior on the existing multi-target ChangeZone path.

PR #4319, opened 2026-06-25
Fix Court of Cunning milling only the first targeted player (#4252)

### LINKED ISSUE #4252: Court of Cunning only mills the first selected target
Reported in Discord: https://discord.com/channels/1485498006781427802/1513087032937943110/1518588332723011645

**Thread:** Court of Cunning bug
**Cards:** court of cunning
**Extraction confidence:** follow-up/manual triage

## Summary
Court of Cunning allows selecting any number of target players, but only the first selected target appears to mill cards. The thread was previously linked to #426, but the maintainer noted that was the wrong issue and that this Court of Cunning behavior still needs its own regression coverage.

### court of cunning
When this enchantment enters, you become the monarch.
At the beginning of your upkeep, any number of target players each mill two cards. If you're the monarch, each of those players mills ten cards instead. (To mill a card, a player puts the top card of their library into their graveyard.)

---
<sub>report_id: `discord:1513087032937943110:1518588332723011645:0` · follow-ups: `discord:1513087032937943110:1518610200813109289:0`, `discord:1513087032937943110:1518612052279558244:0` · discord: `1513087032937943110/1518588332723011645`</sub>

kiannidev:
## Summary
Fixes #4252. Court of Cunning's upkeep reads "any number of target players each mill two cards. If you're the monarch, each of those players mills ten cards instead." When multiple players were targeted, only the first player milled — especially visible when the controller was the monarch (10-card mill on P0, zero on everyone else).

## Root cause
The multi-target-over-`Player` fan-out in `resolve_chain_body` ran **after** the `ConditionInstead` "instead" swap. Swapping the parent `Mill { target: Player }` for the monarch rider's `Mill { target: ParentTarget }` cleared `multi_target` and changed the effect's target filter, so the fan-out branch was skipped and `ParentTarget` resolved only against the first entry in `ability.targets`.

## Fix
Move the fan-out block to the top of `resolve_chain_body`, before the instead swap. Each narrowed per-player iteration then runs the full chain (including the monarch swap) independently.

## Files changed
- `crates/engine/src/game/effects/mod.rs`
- `crates/engine/tests/integration/court_of_cunning_multi_target_mill.rs`

## CR references
- CR 601.2c — multi-target player selection at announcement
- CR 608.2c — resolve once per chosen player; `ConditionInstead` swap semantics
- CR 614.1 — "instead" replacement (monarch mills ten)
- CR 725 — monarch designation

## Track
Developer

## LLM
Model: claude-opus-4.8
Thinking: high

## Verification
- `cargo fmt -p engine` — clean
- `cargo clippy -p engine --lib` — clean
- `cargo test -p engine --test integration court_of_cunning` — 2 pass / 0 fail
- `cargo test -p engine --test integration tinybones` — 5 pass / 0 fail (fan-out regression guard)

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/tests/integration/court_of_cunning_multi_target_mill.rs
+++ b/crates/engine/tests/integration/court_of_cunning_multi_target_mill.rs
@@ -9,7 +9,8 @@
 //! CR 603.3d + CR 601.2c: the upkeep trigger's controller chooses any number
 //! of target players as the ability goes on the stack.
 //! CR 101.4 + CR 608.2c: the effect resolves once per chosen player.
-//! CR 614.1 ("instead") + CR 603.4 monarch designation (CR 725).
+//! CR 614.1 + CR 608.2c: the "mills ten instead" rider replaces the mill
+//! during resolution. CR 725: monarch designation gates that rider.
 
 use engine::game::scenario::{GameScenario, P0, P1};
 use engine::types::ability::TargetRef;
--- a/crates/engine/tests/integration/court_of_cunning_multi_target_mill.rs
+++ b/crates/engine/tests/integration/court_of_cunning_multi_target_mill.rs
@@ -1,0 +1,135 @@
+//! Issue #4252 — Court of Cunning: "any number of target players each mill two
+//! cards. If you're the monarch, each of those players mills ten cards
+//! instead." The reported symptom was that only the FIRST selected target
+//! actually milled. This is the dedicated regression deliverable the issue
+//! asks for: it drives the REAL upkeep-trigger pipeline (multi-target player
+//! selection → fan-out → mill), asserting every targeted player mills, and
+//! covering the monarch "mills ten instead" branch.
+//!
+//! CR 603.3d + CR 601.2c: the upkeep trigger's controller chooses any number
+//! of target players as the ability goes on the stack.
+//! CR 101.4 + CR 608.2c: the effect resolves once per chosen player.
+//! CR 614.1 ("instead") + CR 603.4 monarch designation (CR 725).
+
+use engine::game::scenario::{GameScenario, P0, P1};
+use engine::types::ability::TargetRef;
+use engine::types::actions::GameAction;
+use engine::types::game_state::WaitingFor;
+use engine::types::phase::Phase;
+use engine::types::PlayerId;
+
+/// Court of Cunning's printed Oracle text — byte-identical to
+/// `client/public/card-data.json` and MTGJSON `AtomicCards.json`.
+const COURT_OF_CUNNING: &str = "When Court of Cunning enters, you become the monarch.\n\
+     At the beginning of your upkeep, any number of target players each mill two cards. \
+     If you're the monarch, each of those players mills ten cards instead.";
+
+fn graveyard_count(runner: &engine::game::scenario::GameRunner, player: PlayerId) -> usize {
+    runner
+        .state()
+        .players
+        .iter()
+        .find(|p| p.id == player)
+        .map(|p| p.graveyard.len())
+        .expect("player exists")
+}
+
+/// Build a 2-player game with Court of Cunning already on P0's battlefield and
+/// a 15-card library for every player. `monarch` seeds the monarch designation
+/// directly (the ETB "you become the monarch" trigger does not fire for a
+/// scenario-seeded permanent).
+fn build_runner(monarch: Option<PlayerId>) -> engine::game::scenario::GameRunner {
+    let library: Vec<&str> = (0..15).map(|_| "Library Card").collect();
+
+    let mut scenario = GameScenario::new();
+    scenario.at_phase(Phase::Untap);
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
VERDICT: request-changes

Gemini's CR-annotation finding is valid on the current head `3c0ef24d557e7792324b19ed73f83a274f096344`. The fan-out move itself is at the right seam for the Court of Cunning bug, and the new integration coverage drives the real upkeep trigger target-selection path, but the moved block stripped the rules comments from the branches that implement distinct CR behavior.

**[MED]** Restore the CR annotations on the moved multi-target fan-out branch. Evidence: `crates/engine/src/game/effects/mod.rs:4829`, `:4836`, and `:4847` now perform zero-target no-op resolution, APNAP-ordered per-player fan-out, and waiting-for continuation stashing without the CR comments that existed on the original block. I verified the cited rules locally: CR 601.2c covers variable target count announcement, CR 608.2c covers resolution instructions/replacement text, and CR 101.4 covers APNAP ordering. Why it matters: this is rules-touching engine logic, and this repo requires verified CR annotations directly at the rule-implementing code. Suggested fix: restore the prior comments, adjusted for the new position before `ConditionInstead`, especially the zero-target `EffectResolved`, APNAP fan-out, and paused-choice continuation comments.

No approval, label, or enqueue while this is unresolved and exact-head CI is still running.

crates/engine/tests/integration/court_of_cunning_multi_target_mill.rs
```diff
@@ -0,0 +1,135 @@
...
+//! instead." The reported symptom was that only the FIRST selected target
+//! actually milled. This is the dedicated regression deliverable the issue
+//! asks for: it drives the REAL upkeep-trigger pipeline (multi-target player
+//! selection → fan-out → mill), asserting every targeted player mills, and
+//! covering the monarch "mills ten instead" branch.
+//!
+//! CR 603.3d + CR 601.2c: the upkeep trigger's controller chooses any number
+//! of target players as the ability goes on the stack.
+//! CR 101.4 + CR 608.2c: the effect resolves once per chosen player.
+//! CR 614.1 ("instead") + CR 603.4 monarch designation (CR 725).
```
matthewevans: This `CR 603.4` citation is not monarch designation; local `docs/MagicCompRules.txt` defines CR 603.4 as the intervening-if rule. The monarch designation is CR 725, and the `instead` rider here is better covered by CR 614.1 plus resolution text handling under CR 608.2c. Please remove or replace the 603.4 reference before this queues.

matthewevans (changes requested):
VERDICT: request-changes

The original blocker on the moved production fan-out block is addressed: the current head restores the CR comments at the zero-target, APNAP fan-out, and continuation branches, and exact-head CI is green.

One CR hygiene issue remains before I can approve: the new regression file cites `CR 603.4` as part of the monarch/instead behavior, but local `docs/MagicCompRules.txt` has CR 603.4 as the intervening-if rule. Monarch is CR 725, and the "instead" behavior is covered by CR 614.1 plus the resolution text handling under CR 608.2c. Please remove or replace the `CR 603.4` reference in the test comment so the rule citation is verified and accurate.

kiannidev:
Thanks both — addressed.

**@matthewevans** (test header citation): you're right, `CR 603.4` is the intervening-"if" rule, not monarch designation. I checked `docs/MagicCompRules.txt`: 603.4 is intervening-if, 725.1 is the monarch designation, and 614.1 is replacement effects. The header now reads:

```
//! CR 614.1 + CR 608.2c: the "mills ten instead" rider replaces the mill
//! during resolution. CR 725: monarch designation gates that rider.
```

**@gemini-code-assist** (stripped CR annotations): the moved fan-out block does carry the annotations now — `CR 601.2c` (any-number / zero targets), `CR 608.2c` (zero-target resolution + APNAP-ordered per-player resolution), and `CR 101.4` (APNAP ordering) are all inline in `resolve_chain_body` at the top of the block (`effects/mod.rs:4811-4862`). That comment predates the restoration.

`cargo test` (both monarch and non-monarch cases) is green.

matthewevans (approved):
VERDICT: approve

Exact-head CI is green at `9983cca6cf722c18d51b9c8f6cbf474a9896fee7`, and the author follow-up resolves the previous maintainer blocker. I rechecked the GitHub API patch and local worktree: the fan-out branch still sits before the `ConditionInstead` swap, the zero-target/APNAP/continuation rules comments are restored at the moved production code, and the regression header no longer cites `CR 603.4` for monarch behavior.

Gemini's original CR-annotation finding is addressed; its follow-up agrees the restored comments and monarch/non-monarch integration coverage validate the intended path. The change satisfies the three gates: correct engine seam, maintainable moved fan-out logic with CR evidence, and discriminating integration coverage for the Court of Cunning bug.

PR #4320, opened 2026-06-25
fix(engine): Temple Altisaur prevents all but 1 damage to Dinosaurs

### LINKED ISSUE #4244: Temple Altisaur prevents all damage — Instead of prevent all but 1 damage to dinosaurs, [[Temple Altisaur]] is preventi…
Reported in Discord: https://discord.com/channels/1485498006781427802/1518478990745276476/1518478990745276476

**Thread:** Temple Altisaur prevents all damage
**Cards:** temple altisaur

## Summary
Instead of prevent all but 1 damage to dinosaurs, [[Temple Altisaur]] is preventing all combat damage on the table.

### temple altisaur
If a source would deal damage to another Dinosaur you control, prevent all but 1 of that damage.

---
<sub>report_id: `discord:1518478990745276476:1518478990745276476:0` · discord: `1518478990745276476/1518478990745276476`</sub>

kiannidev:
## Summary

Fixes #4244 — Temple Altisaur was preventing **all** damage to other Dinosaurs you control instead of leaving 1 damage through.

## Root cause

The damage-prevention parser treated `"prevent all but 1 of that damage"` as blanket `"prevent all"` (`PreventionAmount::All`). The recipient filter also failed to parse from the `"would deal damage to another Dinosaur you control"` wording, so the shield was not scoped to allied Dinosaurs.

## Changes

- Add `PreventionAmount::AllBut(u32)` for continuous "prevent all but N" shields (CR 615.1a)
- Parse `"prevent all but N"` before the generic `"prevent all"` branch
- Extend recipient filter parsing to handle `"would deal damage to …"` clauses
- Apply partial prevention in the damage replacement applier (reduce event amount, emit `DamagePrevented` for the prevented portion)
- Add parser + integration regression tests

## Verification

- [x] `cargo test -p engine temple_altisaur`
- [x] `./scripts/check-parser-combinators.sh`

Track: Developer

### DIFF
--- a/crates/engine/src/game/replacement.rs
+++ b/crates/engine/src/game/replacement.rs
@@ -1072,6 +1072,60 @@ fn damage_done_applier(
                         applied,
                     });
                 }
+                PreventionAmount::AllBut(keep) => {
+                    let redirected_amount = damage_amount.saturating_sub(keep);
+                    if redirected_amount == 0 {
+                        return ApplyResult::Modified(ProposedEvent::Damage {
+                            source_id,
+                            target,
+                            amount: damage_amount,
+                            is_combat,
+                            applied,
+                        });
+                    }
+                    consume_prevention_shield(state, rid, None);
+                    if let Some(new_target) = new_recipient.filter(|_| redirected_amount > 0) {
+                        let redirected_event = ProposedEvent::Damage {
+                            source_id,
+                            target: new_target,
+                            amount: redirected_amount,
+                            is_combat,
+                            applied: applied.clone(),
+                        };
+                        match replace_event(state, redirected_event, events) {
+                            ReplacementResult::Execute(event) => {
+                                let ctx = super::effects::deal_damage::DamageContext::from_source(
+                                    state, source_id,
+                                )
+                                .unwrap_or_else(|| {
+                                    let controller = state
+                                        .objects
+                                        .get(&source_id)
+                                        .map(|obj| obj.controller)
+                                        .unwrap_or(PlayerId(0));
+                                    super::effects::deal_damage::DamageContext::fallback(
+                                        source_id, controller,
+                                    )
+                                });
+                                let _ = super::effects::deal_damage::apply_damage_after_replacement(
+                                    state, &ctx, event, is_combat, events,
+                                );
+                            }
+                            ReplacementResult::Prevented => {}
+                            ReplacementResult::NeedsChoice(_) => {
+                                state.pending_replacement = None;
+                            }
+                        }
+                    }
+                    let remaining_amount = damage_amount.saturating_sub(redirected_amount);
+                    return ApplyResult::Modified(ProposedEvent::Damage {
+                        source_id,
+                        target,
+                        amount: remaining_amount,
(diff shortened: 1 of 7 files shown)


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** `AllBut` is implemented in the redirection branch with semantics that do not match redirection rules. Evidence: `crates/engine/src/game/replacement.rs:1075`. This arm splits the damage, optionally applies the excess to `new_recipient`, then always returns only `keep` damage to the original target; if `new_recipient` is absent/illegal, CR 614.9 says the redirection does nothing, but this branch still drops the excess. `AllBut` is a prevention amount for Temple Altisaur, not a proven redirection amount. Suggested fix: do not invent redirection semantics for `AllBut`; either make this state unreachable/impossible for `ShieldKind::Redirection`, or handle it in a way that preserves CR 614.9 and add proof if the state can be constructed.

**[MED]** The new parser amount arm uses global string search instead of the local parser stream / nom combinators. Evidence: `crates/engine/src/parser/oracle_effect/imperative.rs:4455` and `crates/engine/src/parser/oracle_replacement.rs:6910`. Gemini's note is valid for the imperative parser, and the replacement-parser arm has the same new-shape issue: the code scans the whole lowered clause for `prevent all but` instead of decomposing the current prevention amount phrase. Why it matters: this violates the parser combinator mandate and can bind a later/sibling phrase rather than the amount being parsed. Suggested fix: parse `all but <number>` from the local prevention amount position with `tag`/`parse_number` or an existing nom helper, before the generic `all` arm.

**[MED]** The integration test does not prove the `another Dinosaur you control` exclusion for Temple Altisaur itself. Evidence: `crates/engine/tests/integration/issue_4244_temple_altisaur.rs:86`. The test proves an allied Dinosaur is reduced to 1 and player damage is untouched, but it never deals damage to Temple Altisaur; a regression that loses `FilterProp::Another` could still pass the headline test. Suggested fix: add a direct self-damage assertion showing Temple Altisaur takes the full damage.

CI is still in progress, so there is no approval/label/enqueue path on this head.

matthewevans (commented):
Reviewed current head `579d0bfea599d30b31c635a06cb6d32760471806` via GitHub API patch/comments, Gemini findings, the prior Matt review, and an isolated worktree.

VERDICT: approve-pending-CI

The prior blockers appear addressed on this head:

- The `AllBut` redirection branch no longer invents partial-redirection behavior; it is documented as unreachable for `ShieldKind::Redirection`, preserving CR 614.9 rather than silently dropping excess damage.
- The imperative and replacement parser paths now parse `all but <N>` from the local parser position before the generic `all` arm, addressing Gemini's valid parser-combinator concern.
- The integration test now proves the `another Dinosaur` exclusion by dealing damage to Temple Altisaur itself and asserting no prevention event fires.

Exact-head CI is still in progress, so I am not formally approving, labeling, or enqueueing from this review. Once CI is green and the repush/currentness guard holds, this can move to approval/enqueue.

matthewevans (approved):

PR #4321, opened 2026-06-25
fix(engine): From the Rubble targets graveyard cards

### LINKED ISSUE #4243: From the Rubble is targeting incorrectly, to no effect — At end step, [[From the Rubble]] is asking for a target among…
Reported in Discord: https://discord.com/channels/1485498006781427802/1518479335152025671/1518479335152025671

**Thread:** From the Rubble is targeting incorrectly, to no effect
**Cards:** from the rubble

## Summary
At end step, [[From the Rubble]] is asking for a target among creatures on the board instead of in the graveyard, and naturally having no effect on them.

### from the rubble
As this enchantment enters, choose a creature type.
At the beginning of your end step, return target creature card of the chosen type from your graveyard to the battlefield with a finality counter on it. (If a creature with a finality counter on it would die, exile it instead.)

---
<sub>report_id: `discord:1518479335152025671:1518479335152025671:0` · discord: `1518479335152025671/1518479335152025671`</sub>

kiannidev:
## Summary

Fixes #4243 — From the Rubble's end-step ability was offering targets among battlefield creatures instead of creature cards in your graveyard.

## Root cause

The imperative `return … from your graveyard` parser path set `ChangeZone.origin = Graveyard` but did not attach `FilterProp::InZone { Graveyard }` to the target filter. Target enumeration therefore defaulted to the battlefield. Additionally, `parse_type_phrase` parsed `"of the chosen type"` before the trailing `"from your graveyard"` zone suffix, so the zone never landed on the filter.

## Changes

- Call `add_inferred_origin_constraints_to_target` in the imperative return-to-battlefield path
- Parse zone suffixes after `"of the chosen type"` in `oracle_target.rs`
- Add parser + integration regression tests

## Verification

- [x] `cargo test -p engine from_the_rubble`
- [x] `./scripts/check-parser-combinators.sh`

Track: Developer

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -1665,11 +1665,6 @@ pub(super) fn parse_targeted_action_ast(
         };
         let is_mass = is_mass || count.is_some();
         let origin = super::infer_origin_zone(rest_lower);
-        // CR 115.2: A target described as being in a non-battlefield zone
-        // ("return target creature card … from your graveyard") must carry that
-        // zone on its target filter so candidate enumeration is constrained to
-        // the named zone rather than defaulting to the battlefield.
-        let target = super::add_inferred_origin_constraints_to_target(target, origin, rest_lower);
 
         // CR 400.7: Single-object battlefield destinations use ChangeZone;
         // mass destinations use ChangeZoneAll. Only pure mass-bounce
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -1665,6 +1665,10 @@ pub(super) fn parse_targeted_action_ast(
         };
         let is_mass = is_mass || count.is_some();
         let origin = super::infer_origin_zone(rest_lower);
+        // CR 115.2: A target described as being in a non-battlefield zone
+        // ("return target creature card … from your graveyard") must carry that
+        // zone on its target filter so candidate enumeration is constrained to
+        // the named zone rather than defaulting to the battlefield.
         let target = super::add_inferred_origin_constraints_to_target(target, origin, rest_lower);
 
         // CR 400.7: Single-object battlefield destinations use ChangeZone;
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -1665,6 +1665,7 @@ pub(super) fn parse_targeted_action_ast(
         };
         let is_mass = is_mass || count.is_some();
         let origin = super::infer_origin_zone(rest_lower);
+        let target = super::add_inferred_origin_constraints_to_target(target, origin, rest_lower);
 
         // CR 400.7: Single-object battlefield destinations use ChangeZone;
         // mass destinations use ChangeZoneAll. Only pure mass-bounce
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -2263,9 +2263,12 @@ pub fn parse_type_phrase_with_ctx<'a>(
         pos += remaining_offset + of_chosen_len;
     }
 
-    // CR 400.7: Zone phrases may trail "of the chosen type" ("target creature
-    // card of the chosen type from your graveyard", From the Rubble). The
-    // primary `parse_zone_suffix` arm above runs before this suffix.
+    // CR 115.2: A spell or ability may target an object in a zone other than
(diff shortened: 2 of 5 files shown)


crates/engine/src/parser/oracle_effect/imperative.rs
```diff
@@ -1665,6 +1665,7 @@ pub(super) fn parse_targeted_action_ast(
         };
         let is_mass = is_mass || count.is_some();
         let origin = super::infer_origin_zone(rest_lower);
+        let target = super::add_inferred_origin_constraints_to_target(target, origin, rest_lower);
```
matthewevans: This new target-filter mutation implements zone-qualified targeting: CR 115.2 allows a spell/ability to target an object in another zone only when it specifies that zone. Please add a verified CR 115.2 annotation here so the parser-side rule hook stays documented.

crates/engine/src/parser/oracle_target.rs
```diff
@@ -2263,6 +2263,17 @@ pub fn parse_type_phrase_with_ctx<'a>(
         pos += remaining_offset + of_chosen_len;
     }
 
+    // CR 400.7: Zone phrases may trail "of the chosen type" ("target creature
```
matthewevans: This is not CR 400.7 behavior; that rule is about an object becoming a new object after changing zones. The rule that permits `target ... from your graveyard` is CR 115.2. Please update this comment to cite CR 115.2.

matthewevans (changes requested):
Reviewed current head `4ee794e72c6e95c01683b0c8461a2d8514719dd9` via GitHub API patch, Gemini comments, CR lookup, and an isolated worktree.

The implementation direction looks right: the explicit graveyard target now flows through `extract_explicit_zones` / `find_legal_targets`, and the integration test proves the graveyard Dinosaur is legal while the battlefield Dinosaur is not. I do not see a deeper targeting blocker.

Requesting changes for the CR hygiene Gemini flagged. This repo treats CR annotations as part of the engine contract, and `docs/MagicCompRules.txt` confirms this is CR 115.2 target-zone legality, not CR 400.7 object identity after zone changes. Please add the missing CR 115.2 annotation at the new inferred-origin constraint call and correct the `oracle_target.rs` comment to cite CR 115.2.

matthewevans (changes requested):
Reviewed current head `91590fd537eb2f3e810456d4c08dc31f8e2c652b` via GitHub API patch/comments, Gemini comments, the coverage parse-diff bot output, and an isolated worktree.

VERDICT: request-changes

The two CR hygiene issues from the prior review are fixed, but the implementation now has a broader parse-shape regression that needs to be narrowed before this can be approved.

**[MED]** The new inferred-origin constraint path adds redundant owner scope to already zone-qualified `from your graveyard` targets, causing a large parse-diff across existing cards. Evidence: the coverage bot reports 905 card changes, including many signatures changing from `in graveyard you control creature` to `in graveyard you control you control creature`; code evidence is `crates/engine/src/parser/oracle_effect/mod.rs` where `add_inferred_origin_constraints_to_target` adds `Owned { You }` whenever `target.extract_in_zone().is_some()` and `origin_is_your_zone(...)`, after `parse_type_phrase` has already parsed `from your graveyard` into an `InZone` target with `ControllerRef::You` in `crates/engine/src/parser/oracle_target.rs`. Runtime owner-zone matching makes this often redundant rather than immediately fatal, but the blast radius is not justified for a From the Rubble fix and changes hundreds of established parse signatures.

Please narrow the new inferred-origin behavior so it only fills missing zone/owner constraints for the target class that actually lacked them, without re-scoping the already parsed `from your graveyard` class. A good proof would include From the Rubble staying fixed plus a representative existing `return target ... from your graveyard` parse-diff/signature staying unchanged.

matthewevans (approved):
Reviewed current head `c99b800a66ca5552e3ab1798598f5901b7efdc13` via GitHub API patch/comments, Gemini/Matt history, exact-head CI, and an isolated worktree.

VERDICT: approve

The prior blockers are addressed. The CR 115.2 hygiene comments are fixed, and the broader double-scope regression from the previous head is covered by the new `effect_plain_graveyard_reanimation_target_not_double_scoped` test. The actual fix stays at the right parser seam by reusing `parse_zone_suffix` after the `of the chosen type` suffix in `parse_type_phrase_with_ctx`, so From the Rubble's `target creature card of the chosen type from your graveyard` gets the explicit graveyard target constraint without inventing a separate parser path.

The proof is useful: the parser test checks the chosen-type graveyard target and finality counter shape, the production integration test verifies the graveyard Dinosaur is legal while the battlefield Dinosaur is not, and CI completed successfully on this exact head.

PR #4333, opened 2026-06-25
feat(parser): route equal-to and enter-with-counters through full CDA…

andriypolanski:
## Summary

Extend the dynamic quantity parser so composed `QuantityExpr` forms (twice/half, aggregates, difference, offset) are recognized wherever Oracle text uses `"equal to …"` or enter-with-counters scaling — closing a large slice of `Swallow:DynamicQty` coverage gaps without new engine types.

The runtime already supports `Multiply`, `Difference`, `Max`, `DivideRounded`, etc.; the bug was parser paths that only delegated to bare `QuantityRef` leaves.

## Changes

- **`parse_count_expr`** (`oracle_util.rs`): `"equal to …"` now delegates to `parse_cda_quantity` instead of `parse_quantity_ref`, so effect count positions accept composed quantities (e.g. `"equal to twice the number of creatures you control"`, `"equal to the greatest mana value among cards in exile"`).
- **`parse_dynamic_counter_suffix_body`** (`lower.rs`): enter-with-counters suffix `"a number of … counters on it equal to …"` uses the full CDA grammar (Ulamog / Oversimplify / Storm Entity class).
- **Generalized difference** (`oracle_quantity.rs`): `"the difference between A and B"` parses over any two CDA-quantity operands, not only power/toughness.
- **Other spells for-each** (`oracle_quantity.rs`): `"other spell(s) cast this turn"` scales as `SpellsCastThisTurn − 1`, clamped at 0 (Storm Entity ETB counters).
- **Counter equal-to fallback** (`counter.rs`): tries `parse_cda_quantity` before the event-context fallback for composed `"equal to …"` phrases in counter paths.

## Test plan

- [x] `parse_count_expr_equal_to_composed_quantity`
- [x] `cda_quantity_difference_between_two_counts`
- [x] `for_each_other_spells_cast_this_turn`
- [x] `dynamic_counter_suffix_parses_aggregate_equal_to`
- [ ] Regenerate / diff coverage report for `Swallow:DynamicQty` reduction
- [ ] Spot-check: Ulamog ETB counters, Storm Entity, Tormented Thoughts discard count, Oversimplify Fractal tokens

### DIFF
--- a/crates/engine/src/parser/oracle_quantity.rs
+++ b/crates/engine/src/parser/oracle_quantity.rs
@@ -2132,11 +2132,11 @@ pub(crate) fn parse_for_each_clause_expr(clause: &str) -> Option<QuantityExpr> {
     parse_for_each_clause_expr_with_parser(clause, parse_for_each_clause)
 }
 
-/// CR 117.1: "other spell(s) cast this turn" — all spells cast this turn by
-/// any player, excluding the resolving spell (Storm Entity class). Composes
-/// `SpellsCastThisTurn { scope: All }` with offset −1, clamped at zero
-/// (CR 107.1b). Uses `All` (not `Controller`) because Oracle text counts
-/// every other spell, including opponents'.
+/// "Other spell(s) cast this turn" (Storm Entity class): all spells cast this
+/// turn by any player, excluding the resolving spell. Composes
+/// `SpellsCastThisTurn { scope: All }` with offset −1, clamped at zero.
+/// Uses `All` (not `Controller`) because Oracle text counts every other
+/// spell, including opponents'.
 fn parse_other_spells_cast_this_turn_for_each(clause: &str) -> Option<QuantityExpr> {
     let (rest, _) = (
         tag::<_, _, OracleError<'_>>("other "),
--- a/crates/engine/src/parser/oracle_quantity.rs
+++ b/crates/engine/src/parser/oracle_quantity.rs
@@ -2132,9 +2132,11 @@ pub(crate) fn parse_for_each_clause_expr(clause: &str) -> Option<QuantityExpr> {
     parse_for_each_clause_expr_with_parser(clause, parse_for_each_clause)
 }
 
-/// CR 117.1: "other spell(s) cast this turn" — spells the controller cast
-/// this turn excluding the resolving spell (Storm Entity class). Composes
-/// `SpellsCastThisTurn` with offset −1, clamped at zero (CR 107.1b).
+/// CR 117.1: "other spell(s) cast this turn" — all spells cast this turn by
+/// any player, excluding the resolving spell (Storm Entity class). Composes
+/// `SpellsCastThisTurn { scope: All }` with offset −1, clamped at zero
+/// (CR 107.1b). Uses `All` (not `Controller`) because Oracle text counts
+/// every other spell, including opponents'.
 fn parse_other_spells_cast_this_turn_for_each(clause: &str) -> Option<QuantityExpr> {
     let (rest, _) = (
         tag::<_, _, OracleError<'_>>("other "),
@@ -2150,7 +2152,7 @@ fn parse_other_spells_cast_this_turn_for_each(clause: &str) -> Option<QuantityEx
         inner: Box::new(QuantityExpr::Offset {
             inner: Box::new(QuantityExpr::Ref {
                 qty: QuantityRef::SpellsCastThisTurn {
-                    scope: CountScope::Controller,
+                    scope: CountScope::All,
                     filter: None,
                 },
             }),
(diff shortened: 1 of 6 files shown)


crates/engine/src/parser/oracle_quantity.rs
```diff
@@ -2115,6 +2132,34 @@ pub(crate) fn parse_for_each_clause_expr(clause: &str) -> Option<QuantityExpr> {
...
+        .parse(clause.trim())
+        .ok()?;
+    if !rest.trim().is_empty() {
+        return None;
+    }
+    Some(QuantityExpr::ClampMin {
+        inner: Box::new(QuantityExpr::Offset {
+            inner: Box::new(QuantityExpr::Ref {
+                qty: QuantityRef::SpellsCastThisTurn {
+                    scope: CountScope::Controller,
```
matthewevans: This scope is too narrow for the Storm Entity class. The Oracle text is “for each other spell cast this turn,” not “for each other spell you cast this turn,” and the runtime only counts the controller’s list for `CountScope::Controller`. Use `CountScope::All` here and add a regression where an opponent's earlier spell contributes while the resolving Storm Entity itself is excluded.

matthewevans (changes requested):
**[MED]** `other spells cast this turn` is scoped to the controller instead of all players. Evidence: `crates/engine/src/parser/oracle_quantity.rs:2135` says this is the Storm Entity class, and `crates/engine/src/parser/oracle_quantity.rs:2152` emits `QuantityRef::SpellsCastThisTurn { scope: CountScope::Controller, .. }`; the runtime only counts that player for `CountScope::Controller` at `crates/engine/src/game/quantity.rs:2459`/`:3102`. Why it matters: Storm Entity's text says “each other spell cast this turn,” not “you cast,” so opponent-cast spells should contribute. Suggested fix: use `CountScope::All` for this parser arm and add a test that proves an opponent's earlier spell contributes while the resolving Storm Entity itself is excluded.

**[MED]** Exact-head CI is red because the new CDA fallback contradicts an existing parser guard test. Evidence: CI job `Rust tests (shard 2/2)` fails `parser::oracle_effect::tests::counter_suffix_body_dynamic_count_rejects_partial_quantity_tail`, and the failing fixture is `crates/engine/src/parser/oracle_effect/mod.rs:23880` for `"a number of time counters on it equal to its mana value plus one"`. Why it matters: this branch cannot land with red CI, and the test needs to be reconciled with the intended grammar rather than left as a failing invariant. Suggested fix: either intentionally support that phrase and update/strengthen the test to assert the new `Offset(ObjectManaValue, +1)` shape, or narrow the CDA fallback so this remains rejected.

I evaluated Gemini's first comment as well: the cited code uses the repo's `nom_primitives::split_once_on` helper rather than raw string splitting, so I’m not treating that as the blocking parser-style issue here. The scope finding is valid.

matthewevans (changes requested):
VERDICT: still-blocked

Re-reviewing the new head after the follow-up:

- ADDRESSED: the exact-head CI blocker was reconciled. The old `counter_suffix_body_dynamic_count_rejects_partial_quantity_tail` guard is now intentionally inverted into `counter_suffix_body_dynamic_count_parses_its_mana_value_plus_one`, asserting the full `Offset(ObjectManaValue { Recipient }, +1)` shape, and current CI is green.
- NOT ADDRESSED: the Storm Entity scope blocker remains. `parse_other_spells_cast_this_turn_for_each` still emits `QuantityRef::SpellsCastThisTurn { scope: CountScope::Controller, filter: None }` for “other spell(s) cast this turn.” That text is not “you cast”; opponent-cast spells should contribute. Use `CountScope::All` here and add the regression requested earlier: an opponent’s earlier spell contributes while the resolving Storm Entity itself is excluded.

I’m still not treating Gemini’s `split_once_on` style comment as the blocker because the current code uses the repo’s nom helper, but Gemini’s scope finding remains valid. No approval, label, or enqueue on this head.

matthewevans (changes requested):
VERDICT: request-changes

Re-reviewed current head `f45febb9461ee53bc16cd03732198d429a0ffff2` after the author follow-up.

The prior Storm Entity scope blocker is addressed: `"other spell(s) cast this turn"` now uses an all-player count scope, and the regression covers an opponent spell cast earlier this turn while excluding the currently resolving spell.

However, the new CR annotations cite the wrong rule. `crates/engine/src/parser/oracle_quantity.rs` and the new quantity test comments cite `CR 117.1` for the semantics of "other spell(s) cast this turn" / all-player spell counting. I checked the local Comprehensive Rules text: `CR 117.1` is the priority rule, not a rule for spell-count history or Storm Entity's all-player wording. Please either replace this with verified rule citation(s), or remove the `CR` prefix if the comment is only explanatory parser/test prose rather than a rules annotation.

No approval, label, or enqueue from me while the CR citation is wrong, even though exact-head CI is green.

matthewevans (changes requested):
VERDICT: still-blocked

Re-reviewed current head `67777f3e5febbebba9421a31d1fb4ffcc021cf25` after the author follow-up. The prior functional blockers remain addressed: the Storm Entity scope is `CountScope::All`, the opponent-spell regression is present, and exact-head CI is green.

However, the CR-citation blocker is still present. The current patch still cites `CR 117.1` for "other spell(s) cast this turn" / all-player spell-count semantics in `crates/engine/src/parser/oracle_quantity.rs` and the quantity test comments. Local CR text says `CR 117.1` is the priority rule, not a spell-count-history rule. Please either replace this with verified rule citation(s), or remove the `CR` prefix if the comment is explanatory parser/test prose rather than a rules annotation.

No approval, label, or enqueue while the CR annotation is still wrong.

matthewevans (approved):
VERDICT: approve

Re-reviewed current head `bcfd2bb4640e3f664df01545d482de887ca2bc8c` after the CR-prefix follow-up. The prior blockers are addressed: Storm Entity's "other spell(s) cast this turn" path now uses `CountScope::All`, the regression proves an opponent spell contributes while the resolving spell is excluded, the prior counter-suffix CI guard is intentionally updated to the full `Offset(ObjectManaValue, +1)` shape, and the invalid `CR 117.1` comments around this new path were removed.

I also re-evaluated Gemini's comments: the scope finding is addressed, and I am not treating the `split_once_on` note as blocking because the code uses the repo's nom helper rather than ad hoc raw string splitting.

Exact-head CI is green and the repush guard held before this approval.

PR #4340, opened 2026-06-25
feat(parser): wire game-state unless gates through effect conditions

andriypolanski:
## Summary

Wire **game-state “unless”** gates on triggered and one-shot effects: `"do X unless [condition]"` now lowers to effect **X** with a typed `AbilityCondition` (negated gate), distinct from existing payment-unless (`UnlessPayModifier` / CR 118.12).

Removes the trigger parser dead-end that emitted `Effect::Unimplemented { "Unsupported unless clause" }` whenever `unless` appeared without a pay/discard/sacrifice modifier. The effect-chain and clause parsers already had `strip_unless_entered_suffix`; this change routes triggers through that path and extends condition bridging/runtime for patterns the nom grammar already recognizes or newly covers.

## Changes

- **`oracle_trigger.rs`**: Drop the `Unsupported unless clause` pre-lower guard; unrecognized unless riders no longer block the primary effect parse.
- **`parse_effect_clause`** (`mod.rs`): Strip trailing `"unless …"` via `strip_unless_entered_suffix` and attach the resulting condition before body parsing (payment-unless still uses `unless_pay` / `extract_resolution_unless_pay_modifier`).
- **Scoped-player unless grammar** (`oracle_nom/condition.rs`):
  - `"they have exactly N or exactly M cards in hand"` (Skullcage class)
  - `"they control N or more [type]"` (Furnace Punisher class)
- **Static → ability bridging** (`conditions.rs`):
  - `OpponentPoisonAtLeast` → existential `QuantityCheck` over opponents with poison (Ravenous Necrotitan class)
  - `IsRingBearer` → new `AbilityCondition::IsRingBearer` with resolution-time evaluation
  - `Not { inner }` fallback bridges any mappable inner condition (unless-wrapped `Or`, etc.)
- **Runtime** (`effects/mod.rs`): Evaluate `AbilityCondition::IsRingBearer` via `ring::is_current_ring_bearer`.

## Test plan

- [x] `strip_unless_opponent_poison_counters`
- [x] `parse_effect_clause_unless_they_control_basic_lands`
- [x] `test_they_hand_size_exact_disjunction`
- [x] `test_they_control_two_or_more_basic_lands`
- [x] `trigger_unrecognized_unless_payment_preserves_clause_as_gap` (updated: primary effect parses, no bogus condition)
- [x] `cargo test -p engine --lib unless` (173 passed)
- [ ] Regenerate / diff coverage for `Effect:Unsupported unless clause` + `Swallow:Condition_Unless` reduction
- [ ] Spot-check: Ravenous Necrotitan, Skullcage, Furnace Punisher, Sauron Ring-bearer delayed trigger, Kroxa-style payment-unless still on `unless_pay`

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -3206,6 +3206,7 @@ fn fmt_ability_condition(cond: &AbilityCondition) -> String {
         AbilityCondition::IsMonarch => "is monarch".into(),
         AbilityCondition::IsInitiative => "has the initiative".into(),
         AbilityCondition::HasCityBlessing => "has the city's blessing".into(),
+        AbilityCondition::IsRingBearer => "is the ring-bearer".into(),
         AbilityCondition::TargetHasKeywordInstead { keyword } => {
             format!("target has {} (instead)", keyword_label(keyword))
         }
@@ -6213,6 +6214,7 @@ fn condition_feature(cond: &AbilityCondition) -> (&'static str, FeatureSupport) 
         AbilityCondition::IsMonarch => ("IsMonarch", Handled),
         AbilityCondition::IsInitiative => ("IsInitiative", Handled),
         AbilityCondition::HasCityBlessing => ("HasCityBlessing", Handled),
+        AbilityCondition::IsRingBearer => ("IsRingBearer", Handled),
         AbilityCondition::TargetHasKeywordInstead { .. } => ("TargetHasKeywordInstead", Handled),
         // CR 608.2c: active-player check; handled by `evaluate_condition` (effects/mod.rs).
         AbilityCondition::IsYourTurn => ("IsYourTurn", Handled),
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -2029,6 +2029,7 @@ fn should_resolve_subability_on_optional_decline(ability: &ResolvedAbility) -> b
             | AbilityCondition::IsMonarch
             | AbilityCondition::IsInitiative
             | AbilityCondition::HasCityBlessing
+            | AbilityCondition::IsRingBearer
             | AbilityCondition::TargetHasKeywordInstead { .. }
             | AbilityCondition::TargetMatchesFilter { .. }
             | AbilityCondition::TriggeringSpellTargetsFilter { .. }
@@ -7050,6 +7051,12 @@ pub(crate) fn evaluate_condition(
         // CR 702.131c: The city's blessing is a player designation that effects
         // can identify.
         AbilityCondition::HasCityBlessing => eval_has_city_blessing(state, ability.controller),
+        // CR 701.54a: Ring-bearer designation on the ability source.
+        AbilityCondition::IsRingBearer => crate::game::effects::ring::is_current_ring_bearer(
+            state,
+            ability.controller,
+            ability.source_id,
+        ),
         // "Instead" override conditions — return pure boolean value.
         // Terminal control flow (early return from resolve_ability_chain) is the caller's
         // responsibility in the sub-ability context.
--- a/crates/engine/src/parser/oracle_effect/conditions.rs
+++ b/crates/engine/src/parser/oracle_effect/conditions.rs
@@ -2950,8 +2950,9 @@ pub(super) fn difference_expr(cond: &AbilityCondition) -> Option<QuantityExpr> {
     }
 }
 
-/// Bridge `StaticCondition::OpponentPoisonAtLeast` to an existential
-/// `QuantityCheck` over opponents whose poison total meets the threshold.
+/// CR 122.1f: Bridge `StaticCondition::OpponentPoisonAtLeast` to an existential
+/// `QuantityCheck` over opponents whose poison total meets the threshold
+/// ("an opponent has N or more poison counters" unless gates).
 fn opponent_poison_at_least_as_quantity_check(count: u32) -> AbilityCondition {
     use crate::types::ability::{PlayerFilter, PlayerRelation, QuantityRef};
     use crate::types::player::PlayerCounterKind;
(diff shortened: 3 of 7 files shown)


matthewevans (changes requested):
**[HIGH]** Unrecognized `unless` riders are now silently dropped while the primary effect is treated as parsed. Evidence: `crates/engine/src/parser/oracle_trigger.rs:997` removes the old `Unsupported unless clause` dead-end, and the updated regression at `crates/engine/src/parser/oracle_trigger.rs:23464` now asserts `draw a card unless the active player compliments your hat` lowers to `Effect::Draw` with `condition.is_none()`. The parse-diff comment also shows 51 cards losing `Unsupported unless clause`. Why it matters: this makes coverage dishonest and can make a card resolve its primary effect even when an unparsed `unless` condition should suppress it. Suggested fix: when `strip_unless_entered_suffix` cannot decompose the rider, preserve a strict failure marker for that rider/the whole clause instead of returning a normal effect with no condition.

**[MED]** The new scoped-player parser cites a CR rule that does not describe the behavior. Evidence: `crates/engine/src/parser/oracle_nom/condition.rs:2402` says "they" binds to `PlayerScope::ScopedPlayer` at resolution time under `CR 603.7c`, but the local Comprehensive Rules text for `603.7c` is about delayed triggered abilities continuing to affect a particular object after characteristic changes/zone expectations, not pronoun binding or scoped players. Why it matters: this is new parser/rules behavior and the repo requires verified CR annotations; a wrong CR reference creates false confidence. Suggested fix: replace the citation with a verified rule that actually supports the binding model, or remove the CR annotation if no direct CR anchor exists and document the parser/runtime binding invariant without a CR number.

Notes on Gemini: its high-level concern around the poison bridge is directionally useful but the suggested `CR 122.1d` is wrong locally (`122.1d` is stun counters; poison is `122.1f`). The runtime `PlayerAttribute` path reads scalar attrs from each candidate player, so the embedded `CountScope::Controller` is not the runtime bug Gemini described, though using `ScopedPlayer` would make the AST/parse-diff less misleading.

matthewevans (changes requested):
Prior blockers rechecked on `107680b305f1f98a736d15ebbc4b6534b28de03d`: the unrecognized-`unless` drop is now addressed by strict `Effect::Unimplemented` regressions in both `parse_effect_clause` and `parse_trigger_line`, and the bad `CR 603.7c` citation was removed.

**[MED]** Poison threshold bridge emits a misleading controller-scoped attribute and still lacks the verified CR annotation. Evidence: `crates/engine/src/parser/oracle_effect/conditions.rs:2964`-`2967` builds `QuantityRef::PlayerCounter { kind: Poison, scope: CountScope::Controller }` inside an opponent `PlayerAttribute`, while the existing quantity parser for the same “opponents who have N poison counters” class expects `CountScope::ScopedPlayer` at `crates/engine/src/parser/oracle_quantity.rs:4047`-`4050`, and `CountScope` docs define `ScopedPlayer` for per-iteration “they/their” semantics at `crates/engine/src/types/ability.rs:995`-`1015`. The helper’s doc at `crates/engine/src/parser/oracle_effect/conditions.rs:2953` also has no verified CR line even though local CR lookup shows poison counters are covered by `CR 122.1f`, not Gemini’s suggested `122.1d`. Why it matters: runtime currently reads scalar attributes directly from each candidate player, so this is not the runtime bug Gemini described, but the AST/parse-diff now advertises controller-scoped poison for an opponent gate and leaves a rules-touching bridge unaudited, which is exactly the kind of misleading seam that later parser/coverage work will copy. Suggested fix: use `CountScope::ScopedPlayer` for the `PlayerCounter` attr and add a verified `CR 122.1f` comment describing the opponent poison-threshold gate.

Gemini status: the scoped-player `CR 603.7c` thread is addressed on current head; the poison thread is partly valid for the missing CR/seam clarity but its `CountScope::Target` and `CR 122.1d` suggestions should not be copied.

matthewevans (commented):
Prior blocker rechecked on `45c53b35d32b8dea8d4cfff036a6ac1d3197efeb`: addressed.

The poison-threshold bridge now uses `CountScope::ScopedPlayer`, adds a verified `CR 122.1f` comment, and the `strip_unless_opponent_poison_counters` regression now asserts the scoped poison-counter quantity shape instead of only checking that the suffix parsed.

Holding formal approval/label/enqueue because exact-head CI is still red on Rust lint and both Rust test shards, and the PR is currently `BEHIND`.

matthewevans (commented):
Rechecked current head `d67c39ce499faaff98d3c8d092f9fcc158f4fc09` after the parser-gate follow-up. The new hunk replaces the prior `unless` split with `scan_preceded`, keeps the `unless` tail for `parse_unless_payment`, and does not reopen the prior code blocker; Gemini's earlier poison/scoped-player concerns remain addressed by the previous `ScopedPlayer` + verified `CR 122.1f` follow-up.

Holding formal approval/label/enqueue because exact-head CI is still red on both Rust test shards (lint/parser gate is now green), and the PR remains `BEHIND`.

matthewevans (approved):
Re-reviewed current head `3f253dd` after the author follow-ups. The prior blockers are addressed: unrecognized game-state `unless` riders now fail closed instead of silently dropping the gate, the scoped-player CR 603.7c citation was removed, and the poison bridge now uses `CountScope::ScopedPlayer` with a verified `CR 122.1f` comment and shape regression.

Gemini status: its poison/CR-format concerns were evaluated; the valid seam/annotation concern is fixed, while its `CountScope::Target` and `CR 122.1d` suggestions were not copied because they are wrong for this codebase/local CR text.

Maintainer bar: correct parser/effect-condition seam, idiomatic and maintainable enough for the existing architecture, and discriminating tests plus exact-head GitHub CI green on `3f253dd`. Approved.

PR #4341, opened 2026-06-25
fix(parser): UNSUPPORTED cluster: The Fourth Doctor (Blast COMMANDER) — play historic from TOP OF LI

ntindle:
## Summary

**Root cause:** UNSUPPORTED cluster: The Fourth Doctor (Blast COMMANDER) — play historic from TOP OF LIBRARY once/turn + create a Food token

## Cards corrected
- The Fourth Doctor

## Fix
Cluster 41 (The Fourth Doctor) implemented as a 100% parser change with zero new engine variants. The base had advanced past the plan's premise: ability-word classification (Gap-1) was already fixed upstream, but the card was still incorrect — the "cast a historic spell" branch was dropped (affected: Typed(Land+Historic) only) and the reflexive "When you do, create a Food token" rider produced no trigger. Fixed both. Step 1: hoisted the once-per-turn frequency-prefix strip to the top of try_parse_top_of_library_cast_permission so the disjunctive/Bolas branches inherit it, and threaded a frequency: CastFrequency parameter through try_parse_disjunctive_top_of_library_cast_permission, restoring the dropped spell branch -> affected: Or([Typed(Land+Historic), Typed(Card+Historic)]) with frequency OncePerTurn. Step 2: exported the helper via pub(crate) use and imported it into oracle.rs. Step 3: added a pre-Priority-7 handler that strips the ability word, splits the rider on ". when you do, ", parses the permission static, and emits BOTH the static AND a reflexive TriggerMode::PlayCard trigger with valid_target Controller, spell_cast_origin Equals(Library), valid_card cloned from the permission's own affected filter (closing the over-trigger gap), executing the Food token; self-declines if the rider payoff is Unimplemented. Step 5: three new tests + a shared disjunction assertion helper. Verified via oracle-gen single-card parse (no Unimplemented; correct static + trigger shape), 28 targeted tests pass, cargo fmt clean, cargo clippy -p engine --all-targets -D warnings exit 0, cargo test -p engine exit 0. Parser diff gate passes on production files. CR 603.12 multi-permission scoping is a documented, code-annotated approximation, not asserted correct. No commit made.

## Files changed
- crates/engine/src/parser/oracle_static/restriction.rs
- crates/engine/src/parser/oracle_static/mod.rs

## CR references
- CR 207.2c
- CR 401.5
- CR 601.1a
- CR 603.12
- CR 111.10
- CR 305.1
- CR 400.1
- CR 601.2a

## Verification
- `cd /Users/ntindle/code/random/magic/phase-main-base && cargo fmt --all` — clean (no files changed; nothing to commit)
- `cd /Users/ntindle/code/random/magic/phase-main-base && ./scripts/check-parser-combinators.sh <upstream/main merge-base 258e53607>` — clean (exit 0)
- `cd /Users/ntindle/code/random/magic/phase-main-base && cargo clippy -p engine --all-targets -- -D warnings` — clean (exit 0; fix's 4 changed files clippy-clean, no warnings at all)
- `cd /Users/ntindle/code/random/magic/phase-main-base && cargo test -p engine` — clean (exit 0; all tests passed, 0 failed)
- `cd /Users/ntindle/code/random/magic/phase-main-base && cargo run --profile tool --features cli --bin oracle-gen -- data --filter "the fourth doctor"` — confirmed correct parse (no Unimplemented/Unknown)
Cards confirmed re-parsed correctly: The Fourth Doctor

## Track
Developer

## LLM
- Model: Claude Code / Codex maintainer review
- Thinking: high


matthewevans (changes requested):
Reviewed current head `ccbe83f6ac56352fd60d6fed3a3eb09102adcbbc` with the maintainer bar: correct architectural seam, idiomatic/reasonable/maintainable implementation at that seam, and real proof with discriminating tests. CI is green, but this still needs changes.

1. **The Fourth Doctor is marked supported while the reflexive rider is knowingly not rules-correct in multi-permission games.** In `crates/engine/src/parser/oracle.rs`, the new handler lowers “When you do” to a global `TriggerMode::PlayCard` scoped only by `OriginConstraint::Equals(Zone::Library)` and the permission’s `affected` filter. The adjacent comment explicitly says this cannot distinguish which permission authorized the play. That is not just a documentation limitation: `TopOfLibraryCastPermission` already selects/consumes an authorizing source for spell casts, and `top_of_library_permission_source` intentionally prefers an unlimited permission when one exists. The synthesized trigger bypasses that provenance, so it can still fire when another top-of-library permission, not The Fourth Doctor’s once-per-turn permission, authorized the play. If the provenance is not modeled yet, the rider should remain an honest unsupported gap rather than making The Fourth Doctor parse cleanly.

2. **Top-of-library land plays do not consume the once-per-turn library permission.** The new static uses `play_mode: Play` with `frequency: OncePerTurn`, but `handle_play_land` only calls `top_of_library_land_playable_by_permission(...).is_some_and(...)` and records graveyard/exile permission usage later. There is no matching record for `top_of_library_cast_permissions_used` on the land path, while spell casts do record the selected top-of-library source in `casting_costs.rs`. That means “Once each turn, you may play a historic land ... from the top of your library” is not actually once per turn for lands. The parser tests only inspect the lowered shape, so they do not catch this runtime contract.

3. **The new parser branch mutates shared parse context without restoring it.** In `oracle.rs`, the handler sets `ctx.subject = None` and `ctx.actor = None` before parsing the rider. If rider parsing fails, the line falls through to later parser priorities with the context already cleared; if it succeeds, the same `ctx` is reused for subsequent lines. Please scope this with save/restore around the rider parse, or use an isolated context if that is the intended semantics.

4. **The frequency-prefix strip desynchronizes `text` and `lower`.** In `try_parse_top_of_library_cast_permission`, the prefix is stripped only from `lower`, but the unstripped `text` is passed to branch parsers and stored as the static description. The comment says the suffix remains aligned, but the helper API still has paired original/lower inputs and this creates exactly the class of original/lower drift Gemini called out. Strip both sides together, or use an existing paired-text helper so downstream scans/descriptions reflect the same slice.

The fix can either implement permission provenance for the reflexive trigger and land-play frequency consumption, or keep the unsupported rider honest until that runtime seam exists. As-is, the PR passes parser-shape tests while accepting a rules-bearing approximation as supported.

matthewevans (changes requested):
VERDICT: request-changes

Current-head recheck after the follow-up. The honest unsupported reflexive rider, parser context restore, and text/lower frequency-prefix blockers look addressed, but the land-play frequency blocker is still not fixed.

**[HIGH] Top-of-library land plays still do not consume the once-per-turn permission.** The new recorder recomputes `top_of_library_permission_source` at `crates/engine/src/game/engine.rs:5433`, and that helper reads the current library front at `crates/engine/src/game/casting.rs:2763-2764`. The normal land-play path only calls the recorder after the zone delivery has moved the land out of the library (`engine.rs:5742-5753`, then `engine.rs:5861-5864`). At that point `library.front()` is the next card, so `top_id != object_id` and the helper returns without inserting the permission source into `top_of_library_cast_permissions_used`. The replacement-done path has the same post-delivery ordering at `engine.rs:5798-5801`. This preserves the original bug: a `OncePerTurn` top-library land permission can be reused because the land play never spends the source.

Please capture the authorizing source/frequency before moving the land, alongside the existing `in_library_with_permission` check, and pass that captured source into the commit paths. This also needs a production-path regression around `GameAction::PlayLand` proving the used set is populated or that a second same-source top-library land play is rejected.

No approval/label/enqueue from the sweep.

matthewevans (changes requested):
**[HIGH]** Top-of-library land plays still do not consume the once-per-turn slot on the normal successful land-play path. Evidence: `record_top_of_library_land_permission` recomputes the authorizing permission by calling `top_of_library_permission_source` and then requires the current library top to still equal the played object (`crates/engine/src/game/engine.rs:5720-5728`), but the normal path calls it only after the replacement pipeline has delivered the Library → Battlefield move and stamped the played land (`crates/engine/src/game/engine.rs:6062-6067`, `crates/engine/src/game/engine.rs:6145-6154`). At that point `top_of_library_permission_source` reads `player.library.front()` (`crates/engine/src/game/casting.rs:2771-2772`), so the played land is no longer the top card and the helper returns before inserting into `top_of_library_cast_permissions_used`. Why it matters: the previous blocker remains for ordinary top-of-library land plays; a Fourth Doctor-style once-per-turn historic land play can be used repeatedly if no choice/replacement pause keeps the object on top. Suggested fix: capture the selected top-of-library permission source/frequency before moving the land, then consume that captured source in the land-play epilogue, and add a runtime test that plays two matching lands from the top under a `OncePerTurn` `play_mode: Play` permission.

matthewevans (changes requested):
**[LOW]** Land-play permission consumption cites a spell-casting rule that does not describe this path. Evidence: `crates/engine/src/game/engine.rs:5710` annotates the top-of-library land-play permission recorder with `CR 601.2a`, but `CR 601.2a` is the spell-to-stack proposal rule; this path is playing a land as a special action from the library. Why it matters: this repo treats CR annotations as verified evidence, and a wrong CR citation creates false confidence around the rules authority for the implementation. Suggested fix: replace the spell-casting citation with the land/special-action/library-top rules that actually govern this path, e.g. `CR 305.1 + CR 401.5`, and update nearby repeated comments/tests consistently.

matthewevans (changes requested):
**[LOW]** The land-play consumption path still cites the wrong CR rule. Evidence: `crates/engine/src/game/engine.rs:5117`, `crates/engine/src/game/engine.rs:5251`, and `crates/engine/src/game/engine_tests.rs:7650` cite `CR 601.2a` for playing a land from the library. Why it matters: `CR 601.2a` is the spell-casting-to-stack procedure; land play is governed by `CR 305.1`, `CR 116.2a`, and top-library visibility by `CR 401.5`, so the current annotation points maintainers at the wrong authority. Suggested fix: replace the `CR 601.2a` annotations with verified land-play / special-action / top-library visibility rules that describe this path.

Reviewed current head `c3c38baebe9413d2f367fab8f6820462376de572`.

matthewevans (changes requested):
**[HIGH] Current parse-diff evidence is far outside the PR's claimed scope.**

The current `<!-- coverage-parse-diff -->` sticky for head `11fdea04813ff0332c1d474b21e1b3ac1c3a862e` reports 300 cards / 226 signatures, including 37 `trigger/ChangesZoneAll` cards changing active-zone output. The PR body says this fixes one card, The Fourth Doctor.

That is not mergeable evidence for a parser PR. Either the branch is carrying baseline contamination or the parser change is affecting hundreds of unrelated cards; in both cases, the current head needs to be rebased/updated and the parse diff reduced to the intentional Fourth Doctor scope before this can be approved.

matthewevans (changes requested):
Current-head review on `c1cd0d376977e3e53090e38b6d588ce4244cb4ac`:

I do not see a new implementation blocker in the current code, but the parser blast-radius proof is still not current for this head. This PR changes parser source, while the `coverage-parse-diff` sticky was last updated before the current head and still reports a very broad diff (`300 card(s), 226 signature(s)`). Please regenerate/update the parse-diff sticky for this exact head and either narrow or explicitly justify changes outside The Fourth Doctor scope.

matthewevans (approved):
Maintainer sign-off on current head `c1cd0d376977e3e53090e38b6d588ce4244cb4ac`.

The prior land-play permission-consumption and CR annotation blockers were addressed before this head. CI is green and contributor-proof is satisfied. The refreshed parse-diff sticky is current for this head and is narrowed from the earlier 300-card contaminated output to 4 cards: The Fourth Doctor has the intended semantic change, while the other listed top-of-library permissions are identical add/remove signature churn in the parse-diff display rather than broadened parser behavior.

PR #4344, opened 2026-06-26
perf(engine): trim `compute_combat_tax` tax-attribution scans

### LINKED ISSUE #4334: perf(engine): profile-gate compute_combat_tax tax-attribution scans (follow-up to #4312/#4329)
## Summary

Two residual micro-optimizations in `compute_combat_tax` (`crates/engine/src/game/combat.rs`) surfaced while following up on @lgray's review comment on #4312 (avoid linear membership/lookup scans in combat hot paths). Both are **profile-gated**: the realistic input sizes are small, so they should only be implemented if a profiler shows `compute_combat_tax` is actually hot. This issue exists so the analysis isn't re-derived from scratch.

Context: the headline combat-legality scan that @lgray's comment targeted was already addressed by #4312 (hoisted static scan + precomputed slices) and the `choose_blockers` membership scans by #4329 (HashSet). What remains is internal to `compute_combat_tax`.

## #4 — `matches_target_filter` scan over declared creatures (the more legitimate one)

For each battlefield/command-zone object carrying an `UnlessPay` static, the function scans **all** declared creatures calling `matches_target_filter`:

```rust
for (cid, attack_target) in creatures {
    let creature_matches = match &def.affected {
        Some(filter) => matches_target_filter(state, *cid, filter, &ctx),
        None => true,
    };
    ...
}
```

Cost ≈ `S × N × matches_target_filter`, where `S` = tax statics on board, `N` = declared attackers/blockers. `matches_target_filter` is meaningfully more expensive than an integer compare, so this is the residual most worth a look — *if* profiling justifies it.

## #5 — `per_creature.iter_mut().find(...)` linear lookup (3 sites)

Tax attribution looks up the per-creature slot with a linear scan:

```rust
if let Some((_, slot)) = per_creature.iter_mut().find(|(cid, _)| cid == aid) { ... }
```

(3 sites: `PerQuantityRef`, `PerAffectedWithRef`, and the shared distribution loop.)

**Deliberately deprioritized.** A `HashMap<ObjectId, usize>` index would have to be built **unconditionally on every call**, including the overwhelmingly common no-tax path (which today never touches `per_creature` beyond the initial zero-fill — the attribution arms only run when a matching `UnlessPay` static exists). For realistic `N` (board-bounded, ≤ ~10 typical, ~40 pathological go-wide), the linear `.find()` over a contiguous `Vec` is very likely *faster* than hashing (cache locality, no allocation), so the index would pessimize the common path to shave integer compares off a rare, already-sub-microsecond path. It also trades idiomatic `iter_mut().find()` for index juggling. If ever implemented, build the index **lazily** (only on first attribution) and preserve first-occurrence semantics (`entry().or_insert(idx)`) to match the current `.find()` first-match behavior for duplicate ids.

## Acceptance criteria

- [ ] Profile `compute_combat_tax` under a representative AI combat search (go-wide board + 2–4 tax statics) before changing anything.
- [ ] Only implement #4 / #5 if the profile shows it on the hot path; otherwise close as won't-fix and leave the idiomatic code.
- [ ] Any change must be behavior-preserving (61+ `combat_ai` and `combat` tax tests stay green) and preserve `per_creature` declaration order in the returned `Vec`.

jonathanchang31:
## Summary
This change reduces the residual hot-path cost in `compute_combat_tax`, following the earlier combat performance work in `#4312` and `#4329`.

The main fix is:
- replace repeated per-creature linear lookups with declaration-order index attribution
- check defended-scope gating before running the more expensive `matches_target_filter` path
- add a reproducible manual perf gate for the go-wide taxed-combat scenario

## Related Issue
Closes: #4334 
## Change Type
- [x] Performance improvement
- [x] Tests / benchmark coverage
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor only
- [ ] Breaking change

### Correctness
Ran:
- `cargo test -p engine compute_attack_tax_ -- --nocapture`

Result:
- `9 passed; 0 failed`

### Performance Gate
Ran:
- `cargo test -p engine combat_tax_profile_gate_go_wide_board -- --ignored --nocapture`

Observed benchmark output:
- `compute_attack_tax (24 attackers, 4 tax statics, 20000 iters): 1.935567096s`

This provides a reproducible profile gate for future regressions in the same hot path.

## Checklist
- [x] Change is scoped to the combat-tax hot path
- [x] Existing combat-tax behavior remains covered by targeted tests
- [x] Added a reproducible manual perf benchmark
- [x] `cargo fmt --all` run
- [x] Targeted engine tests passed
- [x] Branch pushed successfully

### DIFF
--- a/crates/engine/src/game/combat.rs
+++ b/crates/engine/src/game/combat.rs
@@ -1608,19 +1608,9 @@ pub fn compute_combat_tax(
             // and allows Archangel of Tithes' "you or planeswalkers you
             // control" to match attacks against either the defender or one
             // of their planeswalkers.
-            let mut affected_ids: Vec<ObjectId> = Vec::new();
+            let mut affected_indices: Vec<usize> = Vec::new();
             let ctx = FilterContext::from_source(state, source_id);
-            for (cid, attack_target) in creatures {
-                let creature_matches = match &def.affected {
-                    Some(filter) => matches_target_filter(state, *cid, filter, &ctx),
-                    // No affected filter — treat as "applies to all taxed creatures",
-                    // matching the behavior of `check_static_ability` when `affected`
-                    // is None.
-                    None => true,
-                };
-                if !creature_matches {
-                    continue;
-                }
+            for (index, &(cid, attack_target)) in creatures.iter().enumerate() {
                 if let Some(filter) = defended {
                     if !super::restrictions::attack_target_matches_defended_scope(
                         state,
@@ -1632,9 +1622,19 @@ pub fn compute_combat_tax(
                         continue;
                     }
                 }
-                affected_ids.push(*cid);
+                let creature_matches = match &def.affected {
+                    Some(filter) => matches_target_filter(state, cid, filter, &ctx),
+                    // No affected filter — treat as "applies to all taxed creatures",
+                    // matching the behavior of `check_static_ability` when `affected`
+                    // is None.
+                    None => true,
+                };
+                if !creature_matches {
+                    continue;
+                }
+                affected_indices.push(index);
             }
-            if affected_ids.is_empty() {
+            if affected_indices.is_empty() {
                 continue;
             }
 
@@ -1669,13 +1669,10 @@ pub fn compute_combat_tax(
                         source_id,
                     );
                     let total = base_cost.scaled(n.max(0) as u32);
-                    if let Some(first) = affected_ids.first() {
-                        if let Some((_, slot)) =
-                            per_creature.iter_mut().find(|(cid, _)| cid == first)
-                        {
-                            *slot = slot.plus(&total);
-                            any_tax = true;
-                        }
+                    if let Some(&first_idx) = affected_indices.first() {
+                        let slot = &mut per_creature[first_idx].1;
+                        *slot = slot.plus(&total);
+                        any_tax = true;
                     }
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


matthewevans:
Reviewed current head `a5856b5d3be098ece3724bec48c841788ed65b01`.

The combat-tax index attribution looks directionally safe: `per_creature` is built from the same declaration slice that now supplies `affected_indices`, and the downstream decline path still keys on the returned object IDs. I am holding formal approval for two reasons:

- Exact-head CI is still in progress.
- Gemini's allocation comment is valid for this PR's stated goal: this hot-path loop should initialize `affected_indices` with `Vec::with_capacity(creatures.len())` instead of `Vec::new()`.

Once that is addressed and exact-head CI is green, this should be ready for the normal repush guard.

matthewevans (commented):
Current head `1445723c31b2f17b23e0961680d9515632a0682c` addresses the prior hold: Gemini's allocation suggestion is applied at `crates/engine/src/game/combat.rs:1611` with `Vec::with_capacity(creatures.len())`.

The content diff remains scoped to `compute_combat_tax`; the merge-from-main commit pulls unrelated already-reviewed files into a raw compare, but the PR changed-file set is still only `crates/engine/src/game/combat.rs`. Holding formal approval only for exact-head CI, which is still in progress.

matthewevans (approved):
Approved current head `1445723c31b2f17b23e0961680d9515632a0682c` after repush guard. Exact-head CI is green, the Gemini allocation blocker is addressed/resolved, and the reviewed PR surface remains scoped to `crates/engine/src/game/combat.rs`.

PR #4398, opened 2026-06-26
feat(parser/layers): SetChosenName + last-chosen readback building blocks (preparatory for as-becomes-attached Equipment choices)

ntindle:
## Summary
Preparatory building-block PR for Psychic Paper / Sanctuary Blade style Equipment readbacks. This does **not** close those cards: the as-becomes-attached choice seam remains a follow-up because modeling it as `TriggerMode::Attached` was rules-incorrect.

## Files changed
- crates/engine/src/game/game_object.rs
- crates/engine/src/game/layers.rs
- crates/engine/src/parser/oracle_static/keyword_grant.rs
- crates/engine/src/types/layers.rs
- crates/phase-ai/src/policies/x_reference.rs

## CR references
- CR 205.1a
- CR 509.1b
- CR 612.8
- CR 613.1c
- CR 613.1d
- CR 613.7

## Track
Developer

## LLM
Thinking: high

## Verification
- pre-commit parser combinator gate — passed when committing the reconstructed branch

## Follow-up
The as-becomes-attached replacement/attachment seam remains required before Psychic Paper or Sanctuary Blade should be marked supported.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -3745,6 +3745,7 @@ fn fmt_modification(m: &crate::types::ability::ContinuousModification) -> String
             format!("set land type {}", land_type.as_subtype_str())
         }
         ContinuousModification::SetChosenBasicLandType => "set chosen land type".into(),
+        ContinuousModification::SetChosenName => "set chosen name".into(),
         ContinuousModification::AssignNoCombatDamage => "assign no combat damage".into(),
         ContinuousModification::RetainPrintedTriggerFromSource {
             source_trigger_index,
--- a/crates/engine/src/game/game_object.rs
+++ b/crates/engine/src/game/game_object.rs
@@ -1742,9 +1742,29 @@ impl GameObject {
     }
 
     /// Look up a stored creature type choice.
+    ///
+    /// CR 613.7: Reads the LAST `ChosenAttribute::CreatureType`, so that a
+    /// re-choice (which appends to `chosen_attributes`, since the vector is only
+    /// cleared on leave-battlefield) supersedes the prior choice — the most
+    /// recent persisted choice wins. ETB-once cards have a single entry, so the
+    /// last entry equals the first and behavior is unchanged. Kept consistent
+    /// with `chosen_card_name` so a same-clause read of "the last chosen name and
+    /// creature type" (Psychic Paper) reports both halves from the same choice.
     pub fn chosen_creature_type(&self) -> Option<&str> {
-        self.chosen_attributes.iter().find_map(|a| match a {
+        self.chosen_attributes.iter().rev().find_map(|a| match a {
             ChosenAttribute::CreatureType(s) => Some(s.as_str()),
+            _ => None,
+        })
+    }
+
+    /// CR 612.8 + CR 613.7: The most recently chosen card name (Psychic Paper's
+    /// "the last chosen name"). Reads the LAST `ChosenAttribute::CardName` so a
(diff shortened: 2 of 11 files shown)


matthewevans (changes requested):
**[HIGH]** Do not model the attach-time `As ... becomes attached, choose ...` choice as an `Attached` trigger.

Evidence: this PR routes the `As <Equipment> becomes attached ... choose ...` line through `parse_oracle_ir` into `result.triggers` (`crates/engine/src/parser/oracle.rs:3099-3109`), and `parse_as_becomes_attached_choose` explicitly builds `TriggerMode::Attached` with a persisted `Choose` chain (`crates/engine/src/parser/oracle_trigger.rs:73-90`, `:143-155`). That matcher fires from `EffectResolved { kind: Equip | Attach | AttachAll }` (`crates/engine/src/game/trigger_matchers.rs:2536-2545`), so the choice is put on the stack and resolves later. Local CR check: triggered abilities are the `When/Whenever/At` family and go on the stack under CR 603.1/603.3; replacement/static `As ...` choices act as the event happens (CR 614.1 is the replacement-effect event seam, and the existing `parse_as_enters_choose` path already keeps analogous persisted choices in `ReplacementDefinition` at `crates/engine/src/parser/oracle_replacement.rs:1623-1654`).

Why this matters: the implementation creates a real priority window after the Equipment attaches but before the chosen name/type/color has been persisted and layers have recomputed. For Psychic Paper, the equipped creature can temporarily have the old name/type until the synthetic trigger resolves; for Sanctuary Blade-style protection, the chosen color is likewise delayed. That is not the rules behavior of an `As ...` choice and it means parser coverage can claim support while runtime ordering is observably wrong.

Suggested fix: extend the replacement/attachment seam so `As ... becomes attached` choices are persisted during the attach event itself, or keep these cards unsupported until that seam exists. The `TriggerMode::Attached` fallback is a reasonable prototype, but it should not be accepted as supported engine behavior.

matthewevans (changes requested):
**[HIGH]** Current head no longer fixes the advertised unsupported card.

Evidence: the repush correctly removed the earlier `TriggerMode::Attached` approach, but the current diff now only contains the static readback pieces (`SetChosenName`, last-chosen name/type parsing, and layer application). There is no replacement/attachment parser/runtime seam in the current diff anymore (`git diff origin/main...HEAD --name-only` only shows static/layer files, not `oracle_replacement.rs`, `oracle_classifier.rs`, or `oracle_trigger.rs`). The PR body also states the important consequence: Psychic Paper and Sanctuary Blade revert to `Unimplemented` pending a proper attachment seam.

Why this blocks acceptance: this PR is still scoped and titled as fixing the Psychic Paper unsupported cluster, but the line that actually creates the choices (`As ~ becomes attached to a creature, choose a creature card name and a creature type`) is not supported at current head. The retained layer work is a reasonable preparatory building block, but by itself it only works after some other code has already persisted `ChosenAttribute::CardName` / `CreatureType` on the Equipment. The new runtime test demonstrates exactly that: with no chosen name recorded the equipped creature keeps its printed name, and the test has to manually push `ChosenAttribute::CardName` onto the Equipment before `SetChosenName` does anything (`crates/engine/src/game/layers.rs:9711-9725`).

Please either implement the proper as-becomes-attached choice path in the replacement/attachment seam and add an end-to-end regression for Psychic Paper, or reduce this PR to an explicitly preparatory building-block PR that does not claim to close the unsupported-card issue. I don't want us to merge support-looking parser/layer surface that leaves the target card unsupported.

matthewevans (approved):

PR #4401, opened 2026-06-26
fix(engine): fire 'whenever [player] is attacked' triggers (CR 508.3b)

Whovencroft:
## Summary

Fix a bug in `matching_attack_events` where **"Whenever enchanted player is attacked"** triggers never fired because the `attacker_matches` closure incorrectly required the source (the curse) to be the attacker.

**Root cause:** When `valid_card` and `valid_source` are both `None`, the fallback `*id == source_id` checks if the attacker IS the source. For curses with `valid_target = AttachedTo` (defending-player filter) and no attacker filter, this is always false.

**Fix:** Add an `else if trigger.valid_target.is_some()` branch that returns `true` for any attacker, letting `attack_target_matches` enforce the defending-player restriction (CR 508.3b).

## Affected Cards (6)

- Curse of Bounty
- Curse of Disturbance
- Curse of Opulence
- Curse of Verbosity
- Curse of Vitality
- Imprison This Insolent Wretch

## Changes

| File | Change |
|------|--------|
| `crates/engine/src/game/trigger_matchers.rs` | Add `else if trigger.valid_target.is_some() { true }` branch in `attacker_matches` closure |
| `crates/engine/src/game/trigger_matchers.rs` | Unit test: `attacks_trigger_matches_any_attacker_when_only_valid_target_set` |
| `crates/engine/tests/integration/curse_c17_when_enchanted_player_attacked_cycle.rs` | Integration tests for Commander 2017 curse cycle (6 test functions) |
| `crates/engine/tests/integration/main.rs` | `mod` declaration added alphabetically |

## Test Axes

1. **Unit test (positive)** — `valid_target = AttachedTo`, no `valid_card`/`valid_source` → attacker attacking enchanted player matches
2. **Unit test (negative)** — same trigger, attacker attacks different player → no match
3. **Integration (Curse of Vitality)** — trigger fires, P0 gains life
4. **Integration (Curse of Verbosity)** — trigger fires, P0 draws a card
5. **Integration (Curse of Disturbance)** — trigger fires, P0 gets 2/2 Zombie token
6. **Integration (Curse of Opulence)** — trigger fires, P0 gets Gold token
7. **Integration (Curse of Bounty)** — trigger fires, P0 untaps nonland permanents
8. **Integration (negative)** — attacking non-enchanted player → trigger does NOT fire

## CR References

- **CR 508.3b** — "An ability that reads 'Whenever [a player] is attacked, . . .' triggers if one or more creatures are declared as attackers attacking that player."
- **CR 303.4b** — An Aura that enchants a player is attached to that player.

## Verification

- `cargo check --lib -p engine` ✅
- `cargo check -p engine --test integration` ✅
- `cargo fmt --check` ✅
- Parser combinator gate ✅

### DIFF
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -1602,10 +1602,25 @@ pub(super) fn matching_attack_events(
                 valid_card_matches(trigger, state, *id, source_id)
             } else if trigger.valid_source.is_some() {
                 valid_source_matches(trigger, state, *id, source_id)
+            } else if trigger.valid_target.is_some() {
+                // CR 508.3b: "Whenever [player] is attacked" — no attacker
+                // filter, any creature attacking that player satisfies the
+                // trigger. The defending-player restriction is enforced by
+                // `attack_target_matches` below.
+                true
             } else {
                 *id == source_id
             }
         };
+
+        // CR 508.3b: "Whenever [player] is attacked" triggers once per
+        // attacked player, not once per attacking creature. Deduplicate when
+        // the trigger has valid_target set but no valid_card/valid_source
+        // (the player-is-attacked pattern).
+        let dedup_by_player = trigger.valid_target.is_some()
+            && trigger.valid_card.is_none()
+            && trigger.valid_source.is_none();
+        let mut seen_defending_players: Vec<PlayerId> = Vec::new();
 
         attacker_ids
             .iter()
@@ -1620,13 +1635,17 @@ pub(super) fn matching_attack_events(
                 if !attack_target_matches(trigger, state, target, *defending_player, source_id) {
                     return None;
                 }
+                let event_defending_player =
+                    attack_target_defending_player(state, target, *defending_player);
+                if dedup_by_player {
+                    if seen_defending_players.contains(&event_defending_player) {
+                        return None;
+                    }
+                    seen_defending_players.push(event_defending_player);
+                }
                 Some(GameEvent::AttackersDeclared {
                     attacker_ids: vec![*id],
-                    defending_player: attack_target_defending_player(
-                        state,
-                        target,
-                        *defending_player,
-                    ),
+                    defending_player: event_defending_player,
                     attacks: vec![(*id, target)],
                 })
             })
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
I reviewed the current head `8786d2cff600928d710cd714c6a10ad67c7e8110` via the GitHub API diff, current comments, local worktree, and CR lookup.

Gemini had no substantive feedback. The Codex inline comments are valid, and I found the same blockers:

**[HIGH]** The valid-target-only attack path emits one trigger per attacker instead of once per attacked player. Evidence: `crates/engine/src/game/trigger_matchers.rs:1605` lets every attacker through, then lines 1616-1630 map each matching attacker into a separate `AttackersDeclared` event. Why it matters: CR 508.3b says “Whenever [a player] is attacked” triggers when one or more creatures attack that player, so two creatures attacking the enchanted player should produce one trigger, not two. Suggested fix: route this player-is-attacked shape through the same defending-player dedupe used by the player-scope branch above, or otherwise collapse by attacked player before returning events.

**[HIGH]** The same path treats attacks on that player's planeswalkers/battles as attacks on the player. Evidence: `attack_target_matches` maps planeswalker/battle targets back to their controller/protector via `attack_target_defending_player` before applying `valid_target`, so a curse attached to P1 fires when a creature attacks P1's planeswalker or protected battle. Why it matters: the changed card class says “enchanted player is attacked”; CR 508.3b separately names player, planeswalker, and battle, and this wording is player-only. Suggested fix: require `AttackTarget::Player(attached_player)` for the valid-target-only curse shape, or encode a player-only attack target filter before accepting the event.

matthewevans (approved):
Reviewed current head `8b2d22345c0d1f4ec90498493a8c95f895c943ab` via GitHub API diff/comments, the prior Codex/Gemini context, local worktree trace, and the review-impl rubric.

The two prior blockers are addressed. The valid-target-only attack path now dedupes by attacked/defending player for the player-is-attacked shape, so multiple attackers at the same enchanted player produce one trigger. The parser also sets `attack_target_filter = Player` for "whenever enchanted player is attacked", and the matcher applies that filter before converting planeswalker/battle attack targets to their defending player, so attacks on planeswalkers/battles do not satisfy the player-only wording.

The added tests cover the dedupe case, the planeswalker negative at the matcher layer, parser emission of the player-only filter, and the Curse cycle runtime path. Exact-head CI is green and the repush guard held.

PR #4411, opened 2026-06-26
fix(engine): lifelink gains life once per source per combat batch (#4366)

### LINKED ISSUE #4366: Blech Loafing Pest — Gives too many +1/+1 counters when attacking with a pest
Reported in Discord: https://discord.com/channels/1485498006781427802/1519473900025680083/1519473900025680083

**Thread:** Blech Loafing Pest
**Cards:** _none detected_

## Summary
Gives too many +1/+1 counters when attacking with a pest

---
<sub>report_id: `discord:1519473900025680083:1519473900025680083:0` · discord: `1519473900025680083/1519473900025680083`</sub>

nickmopen:
## Problem

Closes #4366 (Blech, Loafing Pest — gives too many +1/+1 counters when attacking with a pest).

A source dealing combat damage to multiple recipients simultaneously (multiple blockers, or trample to a blocker **plus** the player) emitted **one `LifeChanged` event per damaged target**, because lifelink life gain was applied per-assignment inside the damage pipeline. This over-triggered "whenever you gain life" abilities — e.g. *Blech, Loafing Pest* ("Whenever you gain life, put a +1/+1 counter on each Pest, Bat, Insect, Snake, and Spider you control") fired once per damaged target instead of once.

## Fix

Per CR 119.3 / CR 702.15b, a single source's combat damage is **one** life-gain event. In `apply_combat_damage`:

- **Phase A:** capture each source's lifelink controller, then force `ctx.has_lifelink = false` so the per-assignment path no longer gains life.
- **Phase C:** sum each source's actual damage across the simultaneous batch and apply **one** `apply_life_gain` per source afterward.

Life totals are unchanged — only the per-source event count is corrected. Two different lifelink sources remain separate events; a deferred life-gain replacement (CR 614.7) still can't pause combat, so that portion is dropped, and `state.waiting_for` is snapshotted/restored so combat is never left paused on a choice nothing answers.

## Tests

- New regression test `lifelink_one_source_many_targets_gains_life_once`: a 5/5 lifelink+trample attacker dealing 2 to a blocker and 3 to the player asserts **exactly one** `LifeChanged(+5)` and a correct life total.
- Full `combat_damage` suite green, including all pre-existing lifelink tests.


nickmopen:
Thanks — good catch. `apply_life_gain` sets `state.waiting_for` to a `ReplacementChoice` when a life-gain replacement defers (CR 614.7), and combat can't pause to honor it, so leaving it set would strand resolution on a choice nothing answers.

Fixed in 7227bb4: snapshot `state.waiting_for` before the per-source lifelink apply and restore it if `apply_life_gain` returns `Err` (deferral). The deferred portion is still dropped, matching the pre-existing per-assignment `DamageResult::NeedsChoice => 0` behavior — but combat is no longer left paused. Also fixes the rustfmt failure (long line wrapped).

Verified locally: `cargo fmt --check` and `clippy` clean; 19/19 lifelink tests pass, including `lifelink_replacement_does_not_double_fire_life_gain_triggers` and `lifelink_conditional_static_saves_from_lethal`.

matthewevans:
Content review is clear on the implementation. I checked the current diff against Gemini's note and the follow-up restores `waiting_for` when a deferred lifegain replacement would otherwise leave combat paused. I also verified the relevant CR citations locally: CR 119.9/120.3f make the life-gain trigger source-based, so summing one lifelink source's simultaneous combat damage while keeping different lifelink sources separate is the right seam.

I updated the branch to current `main`; the new head is merge-only (`1a737c58c04b`). Holding formal approval/label/enqueue until current-head CI finishes green.

matthewevans (approved):
Approved current head after re-review. Exact-head CI is green/CLEAN; the branch update was merge-only after the reviewed content head.

PR #4423, opened 2026-06-26
fix(parser): Dominaria's Judgment gates each color's protection on its own land

nickmopen:
## Problem

*Dominaria's Judgment* — "Until end of turn, creatures you control gain protection from white if you control a Plains, from blue if you control an Island, from black if you control a Swamp, from red if you control a Mountain, and from green if you control a Forest" — parsed into a **single** continuous grant that:

- kept **only the final (Forest) clause's condition**, gating *every* color's protection on controlling a Forest, and
- left white/blue/black/red as raw `ProtectionTarget::CardType("white if you control a plains")` strings instead of real color protections.

The generic suffix-condition stripper peels only the trailing `if …` clause and re-applies it to the whole grant, so a per-clause conditional protection list can't be represented that way.

## Fix

- **`parse_conditional_protection_grant_list`** (`oracle_static/shared.rs`): splits `protection from <quality> if <condition>, …` into one `(target, condition)` pair per clause, tolerating a final clause whose `if …` was already peeled one layer up.
- **`try_parse_conditional_protection_grant_clause`** (`oracle_effect/subject.rs`): parses the full clause *before* `peel_clause`, emitting one conditionally-gated continuous grant per color.
- **`lower.rs`**: a pushed-down clause condition no longer clobbers a per-static condition the static parser already set — it only fills unconditioned statics (the final color picks up the peeled land condition; the others keep theirs).

## Result

Dominaria's Judgment now parses into five grants — protection from **white if you control a Plains**, blue/Island, black/Swamp, red/Mountain, green/Forest.

## Tests

- New regression test `conditional_protection_grant_list_gates_each_color_on_its_own_land` asserts five grants, each gated on its matching basic land.
- `oracle_effect` (2155) and `oracle_static` (884) suites pass; `cargo fmt --check` and `clippy` clean.


matthewevans (changes requested):
**[MED]** Preserve both the outer clause condition and the per-static condition when lowering `GenericEffect` statics. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:979-987` now skips every `StaticDefinition` that already has a condition, while the outer `effective_condition` is considered handled because `pushed_down` is `Some(())`. Why it matters: for any effect clause that has an outer `if`/unless gate and also parses an inner conditional static, the outer gate is silently dropped instead of being conjoined; the established static parser pattern composes these as `StaticCondition::And` (see `crates/engine/src/parser/oracle_static/anthem.rs:524-538`). Suggested fix: when `static_def.condition` is already present, replace it with `StaticCondition::And { conditions: vec![static_cond.clone(), existing] }` rather than leaving it unchanged, and add a regression for an outer-gated conditional grant.

**[MED]** Refactor the new conditional-protection parser into nom grammar instead of ad hoc string splitting. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:1453-1510` parses the new Oracle surface with `strip_prefix`, `split(", ")`, `split_once(" if ")`, and manual `and `/`from ` peeling. Why it matters: this is parsing dispatch on a new Oracle grammar surface, and the repo's parser rule requires composing axes with nom combinators so comma/Oxford/list/condition variants remain maintainable and do not become a Dominaria's Judgment-only special case; Gemini's R1 comment is valid on the current head. Suggested fix: model the lead-in, repeated `from <protection quality> if <condition>` members, comma/`and` separators, and optional final peeled condition as nom combinators, reusing `parse_attached_condition_run` / `parse_attached_static_condition` style helpers where appropriate.

**[LOW]** Fix the incorrect CR citation for the leading duration peel. Evidence: `crates/engine/src/parser/oracle_effect/subject.rs:387-390` cites `CR 611.2b` for "Until end of turn", but the local Comp Rules text says `611.2b` is specifically for "for as long as ..." durations; `611.2a` is the rule that covers durations such as "until end of turn". Why it matters: CR annotations are the project's audit trail, and the cited rule currently describes the wrong duration class. Suggested fix: change that annotation to `CR 611.2a` or remove the subrule if the comment is only about parser plumbing.

matthewevans (changes requested):
Current-head re-review after `72dd40525eb65dfb695ab2a92bc9c749dec2dc29`.

Addressed: the conditional protection grant parser surface is now nom-based, so the prior parser-style blocker is resolved.

**[MED] Preserve both the outer clause condition and each per-static condition when lowering `GenericEffect` statics.** Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:979-987` still only assigns the pushed-down outer `static_cond` when `static_def.condition.is_none()`. For Dominaria's Judgment-style output, the per-color statics already have their land-specific conditions, so the outer/effective clause condition is skipped instead of being composed with the existing condition. The required shape is conjunction, not either/or; otherwise future clauses that combine a suffix/outer condition with preconditioned statics will silently lose one gate.

**[LOW] Fix the verified CR citation on the leading duration comment.** Evidence: `crates/engine/src/parser/oracle_effect/subject.rs:387` still cites `CR 611.2b` for “Until end of turn...” duration handling. The local Comprehensive Rules text has `CR 611.2a` for effects lasting “until end of turn”; `CR 611.2b` is the separate “for as long as...” duration rule. Please update the annotation so the code points at the rule it implements.

matthewevans:
Re-reviewed the author follow-up at 33a485ab8fa2e61ccdc4946199d70dd6a14e5424. The previously requested blockers are addressed: the outer condition is composed with each per-static condition instead of replacing it, and the leading duration citation now points at verified CR 611.2a. I ran the update-branch treadmill because the PR was BEHIND; GitHub moved it to merge-only head 839c1d2a945c8a2b235b9dff940f6a3ae08bd609. Holding formal approval/label/enqueue until exact-head CI on that updated head is green.

matthewevans (approved):
Approved current head 839c1d2a945c8a2b235b9dff940f6a3ae08bd609 after re-review and update-branch. The follow-up addresses the prior blockers: conditional protection parsing is now nom-composed, outer/per-static conditions are preserved via StaticCondition::And, and the leading duration citation uses verified CR 611.2a. Exact-head CI is green/CLEAN, so I am approving and queueing under the maintainer sweep policy.

PR #4428, opened 2026-06-26
fix(engine): read buffed LKI for dies-trigger P/T quantities (#4269)

### LINKED ISSUE #4269: Nested Shambler uses base power instead of last known power for token count
Reported in Discord: https://discord.com/channels/1485498006781427802/1519130284509888553/1519130284509888553

**Thread:** Nested Shambler
**Cards:** nested shambler
**Parser status:** has_gaps

## Summary
When Nested Shambler dies, the token count appears to use the card's base power instead of the creature's actual/last-known power, so it always creates only one Squirrel token.

### nested shambler
When this creature dies, create X tapped 1/1 green Squirrel creature tokens, where X is this creature's power.

---
<sub>report_id: `discord:1519130284509888553:1519130284509888553:0` · discord: `1519130284509888553/1519130284509888553`</sub>

minion1227:
Closes #4269

## Summary

Fixes #4269. A buffed **Nested Shambler** ("When this creature dies, create X
tapped 1/1 green Squirrel creature tokens, where X is this creature's power")
created **1** token instead of **N** when it died with +1/+1 counters or other
power buffs.

## Root cause

When a creature leaves the battlefield, the order of operations is:

1. `change_zone` snapshots the creature's **buffed** power/toughness into
   `lki_cache`.
2. `revert_layered_characteristics_to_base` reverts the **live** graveyard
   object's P/T back to its base printed value.
3. *Then* the dies trigger resolves and the quantity resolver computes X.

The runtime quantity resolver's `ObjectScope::Source` arm
(`resolve_object_pt` in `game/quantity.rs`) read the **live** graveyard power
with no zone guard. By the time the trigger resolved, that live value had
already been reverted to base, so the stale base value (1) won and the token
count was wrong.

## Rules basis

- **CR 608.2h** — when an effect needs information from an object that has left
  the zone it was expected to be in, it uses that object's last known
  information.
- **CR 603.10a** — a leaves-the-battlefield / dies / sacrifice trigger looks
  back in time at the source as it last existed on the battlefield.

So the buffed **last-known** power must drive the token count, not the
post-revert base value.

## Fix

Extract a single guarded reader, `read_object_pt_by_id`, implementing the
correct precedence:

> **live while on the battlefield → LKI when off the battlefield → live
> fallback** (for graveyard/exile *cards* that never had a battlefield LKI
> snapshot, e.g. Scavenge).

Route every object-P/T scope through this one helper, replacing what were
inconsistent, hand-duplicated inline reads:

- `ObjectScope::Source`
- `ObjectScope::Recipient`
- `ObjectScope::EventSource`
- `ObjectScope::EventTarget`
- `ObjectScope::Anaphoric` (slot 2, trigger-event source)
- the fight-source arm

`ObjectScope::Target` already had the correct zone-guarded pattern; the helper
now unifies all arms on that single implementation.

## Tests

**Integration** (`tests/integration/issue_4269_nested_shambler_death_tokens.rs`,
6 tests) — drives the real death/counter/LKI path end-to-end:

- buffed 3/3 → 3 tokens
- unbuffed 1/1 → 1 token
- toughness sibling (X = toughness) → 4 tokens
- power reduced by −1/−1 counters before death → reduced count (LKI tracks
  modified, not printed base)
- two buffed shamblers dying in the same SBA sweep → independent counts, no
  cross-contamination
- 0-power → 0 tokens, no panic

**Unit** (`game/quantity.rs`, 5 tests) — exercise the helper directly across
scopes:

- `Source`: prefers buffed LKI when off-battlefield; prefers live over stale
  LKI when on-battlefield
- `EventSource`: prefers buffed LKI after the dies-trigger source left the
  battlefield
- `Anaphoric`: "its power" falls through to the trigger-event source slot and
  reads buffed LKI

## Verification

- `cargo fmt --all` — clean
- `cargo clippy -p engine --tests` — no warnings
- All 6 integration + 5 new unit tests pass; surrounding `quantity::tests`
  suite (114 tests) green, no regressions.

### DIFF
--- a/crates/engine/src/game/quantity.rs
+++ b/crates/engine/src/game/quantity.rs
@@ -3741,13 +3741,43 @@ where
     G: Fn(&crate::types::game_state::LKISnapshot) -> Option<i32>,
 {
     match scope {
-        // CR 608.2h + CR 603.10a: "this creature's power" / "its power" resolving
-        // to the ability's source (Nested Shambler's dies trigger). When the
-        // source has left the battlefield, its live P/T has been reverted to base
-        // by `revert_layered_characteristics_to_base`; prefer the buffed LKI
-        // snapshot via the shared guarded read.
+        // CR 608.2h: a source's P/T is read from the zone the ability EXPECTS it
+        // in. The expected zone differs by how the ability reached the stack:
+        //
+        //   * A triggered leaves-the-battlefield / dies ability (Nested
+        //     Shambler's "this creature's power") looks back at the battlefield
+        //     (CR 603.10a). The source has since moved to the graveyard and its
+        //     live layer-7 P/T was reverted to base by
+        //     `revert_layered_characteristics_to_base`, so prefer the buffed
+        //     battlefield LKI snapshot.
+        //   * An activated ability that functions from the graveyard or exile
+        //     (e.g. Scavenge's "this card's power", read while the card is in the
+        //     graveyard) expects the source in its CURRENT zone. Reading a stale
+        //     same-step battlefield LKI snapshot — left over from the card dying
+        //     earlier this step (`zones.rs` inserts it on leaving the
+        //     battlefield; it persists until step transition) — would wrongly
+        //     report the buffed battlefield P/T instead of the live graveyard
+        //     card's P/T. Read the live source in its current zone, falling back
+        //     to LKI only if the object has left the game entirely.
+        //
+        // `source_incarnation` is set exactly for triggered abilities (whose
+        // source can change zones between firing and resolution) and `None` for
+        // activated abilities and casts, so it distinguishes the battlefield
+        // look-back case from the current-zone case. A bare `resolve_quantity`
+        // call with no ability (statics, whose source is on the battlefield, and
+        // direct unit reads) defaults to the look-back read.
         ObjectScope::Source => {
-            read_object_pt_by_id(state, ctx.source, &obj_extract, &lki_extract).unwrap_or(0)
+            let expects_battlefield = ability.is_none_or(|a| a.source_incarnation.is_some());
+            if expects_battlefield {
+                read_object_pt_by_id(state, ctx.source, &obj_extract, &lki_extract).unwrap_or(0)
+            } else {
+                state
+                    .objects
+                    .get(&ctx.source)
+                    .and_then(&obj_extract)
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
Thanks for the focused Nested Shambler follow-up. The LTB/dies case is at the right seam, but this head generalizes the LKI preference too far.

`read_object_pt_by_id` now reads live battlefield state, then `lki_cache`, then live fallback (`crates/engine/src/game/quantity.rs:3713-3727`), and `ObjectScope::Source` routes through that helper (`crates/engine/src/game/quantity.rs:3749-3751`). That is correct when the object was expected to still be on the battlefield, but it is not correct for source/target contexts whose expected zone is actually graveyard or exile. A same-step battlefield LKI snapshot is inserted when an object leaves the battlefield (`crates/engine/src/game/zones.rs:173`) and persists until step transition (`crates/engine/src/game/turns.rs:425`), so a card that died with modified P/T and then functions from graveyard/exile in that same step can have its live graveyard/exile card P/T hidden by stale battlefield LKI. The helper comment even calls out Scavenge-style graveyard reads, but the fallback only works for cards that never had a battlefield LKI snapshot.

Please carry the expected-zone/context into this helper, or keep the battlefield-LKI path separate from live graveyard/exile source reads. The dies-trigger fix should prefer LKI when the expected zone was battlefield; graveyard/exile-source abilities should still read the live card in their expected zone.

matthewevans (approved):
Re-reviewed current head `557a0d3df0698b8d4d1571deef40bffa1df4c630` after the author follow-up.

The previous blocker is addressed: `ObjectScope::Source` P/T resolution now distinguishes the battlefield-lookback path from activated/current-zone source reads, and the new regression `resolve_source_power_reads_live_current_zone_for_activated_graveyard_ability` covers the stale-LKI counterexample from the prior review. The Nested Shambler LKI path remains covered by both the quantity tests and the integration scenario.

CI is green on the current head, and the PR is now `CLEAN`. I noticed the explanatory comment around `source_incarnation` is a little broader than the global call-site reality, but the behavior gate is correct for this change and not a blocker.

Evidence: GitHub API diff/status, local PR worktree inspection, and current CI rollup. Confidence: high. A contradiction would be a non-triggered, non-current-zone source-P/T path stamped with `source_incarnation` and exercising this same branch; I checked the explicit stamp sites and did not find one that invalidates this PR's covered behavior.

PR #4434, opened 2026-06-26
fix(parser): Limited Resources gates CantPlayLand on lands-on-battlefield count

nickmopen:
## Problem

*Limited Resources* — "Players can't play lands **as long as ten or more lands are on the battlefield**." — parsed to a `CantPlayLand` static with `condition: null`. The "as long as ten or more lands are on the battlefield" gate was silently dropped, so the prohibition applied **unconditionally** instead of only once ten lands are on the battlefield.

## Fix

- Add a clean nom parser `parse_count_on_battlefield_condition` for "`<N>` or more `<type>` are on the battlefield" → `ObjectCount(type) >= N` (reuses `parse_number` / `parse_type_phrase`), wired into `parse_static_condition`.
- In the `CantPlayLand` arm, strip a trailing "as long as `<condition>`" and attach it to the static.

## Result

`Limited Resources` now gates the land-play prohibition on `ObjectCount(Land) >= 10` instead of dropping the condition.

## Tests

- New regression `limited_resources_cant_play_land_gated_on_ten_lands` asserts the `CantPlayLand` static carries the `lands-on-battlefield >= 10` gate.
- `oracle_static` suite (885) passes; `fmt --check`, `clippy`, and the parser gate are clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -2560,11 +2560,18 @@ pub(crate) fn parse_static_line_inner(
         || nom_primitives::scan_contains(tp.lower, "cannot play lands")
     {
         let affected = parse_player_scope_filter(&tp);
-        return Some(
-            StaticDefinition::new(StaticMode::Other("CantPlayLand".to_string()))
-                .affected(affected)
-                .description(text.to_string()),
-        );
+        let mut def = StaticDefinition::new(StaticMode::Other("CantPlayLand".to_string()))
+            .affected(affected)
+            .description(text.to_string());
+        // CR 611.3a: a trailing "as long as <condition>" gates the restriction
+        // (Limited Resources: "... as long as ten or more lands are on the
+        // battlefield"). Without this the prohibition applied unconditionally.
+        if let Some((_, condition_tp)) = tp.split_around(" as long as ") {
+            if let Some(condition) = parse_static_condition(condition_tp.original) {
+                def = def.condition(condition);
+            }
+        }
+        return Some(def);
     }
 
     // --- "can't win the game" / "can't lose the game" (CR 104.3a/b) ---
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
Reviewed current head `6ba06e18e3a1c9146f6798a8ce412890f9ec139e` via the GitHub API diff, Gemini context, local CR lookup, and an isolated worktree at `/tmp/phase-pr-4434-review`.

**[HIGH]** Unsupported `as long as` gates still become unconditional land-play prohibitions. Evidence: `crates/engine/src/parser/oracle_static/dispatch.rs:2569-2574` detects a trailing `as long as`, but if `parse_static_condition(condition_tp.original)` returns `None`, the branch still returns `Some(def)` with no condition. Why it matters: this fixes Limited Resources only for the newly-recognized count shape, but any adjacent `Players can't play lands as long as <unparsed condition>` line would now be marked supported as `CantPlayLand` and enforced unconditionally instead of staying unsupported/honest. Suggested fix: parse the trailing condition through a composed parser path and, when an `as long as` rider is present but not recognized, return `None` or emit an explicit unimplemented/static residual rather than a conditionless `CantPlayLand`; add a negative regression for an unsupported gate.

Gemini evaluation: its unconditional-fallback concern is valid on this head. The style concern about `split_around(" as long as ")` is also directionally valid under the nom-first parser rule; folding the gate into a proper combinator should address both.

No approval, label, or enqueue.

matthewevans:
Current-head hold on `7b09de3bbf2c0ebf93dd3e9a908506c0e13479e9`: the follow-up addresses my prior blocker. The unrecognized `as long as` gate now leaves the line unsupported instead of emitting unconditional `CantPlayLand`, the split moved to a nom-style helper, and the negative regression covers the honest-unsupported path.

I refreshed the branch with the update-branch API because it was BEHIND. Holding formal approval/label/enqueue until the new exact-head CI completes green and the repush guard passes.

matthewevans (approved):
Current-head re-review is clean on `7b09de3bbf2c0ebf93dd3e9a908506c0e13479e9`: the prior unconditional CantPlayLand fallback blocker is addressed, and exact-head CI is green/CLEAN. Approved for merge queue.

PR #4438, opened 2026-06-26
fix(parser): assemble 3+ color prenominal disjunction filters

minion1227:
## Summary

Teaches the target-filter parser to assemble prenominal color disjunctions of
**three or more** colors — e.g. `target white, blue, or black creature`. These
previously collapsed to a typeless/colorless `[Any]` filter, silently dropping
the color restriction so that *any* creature became a legal target.

## Problem

`parse_color_prefix` only recognized a color word when it was immediately
followed by a **space** (`"white creature"`). In a comma-separated color list
the inner legs are followed by a comma (`"blue,"` in
`"white, blue, or black creature"`), so the prefix check failed.

As a result:

- **2-color** prenominal disjunctions (`"green or white creature"`) already
  worked — the bare `" or "` arm of the `TYPE_SEPARATORS` recursion assembled
  them into an `Or` of legs, with `distribute_core_type_to_or` backfilling the
  core type onto the color-only leg.
- **3+-color** prenominal disjunctions fell through entirely. The leading
  `"white,"` was never consumed as a color, the comma never reached the
  `", "` / `", or "` separator recursion, and the phrase degraded to a bare
  `[Any]` filter with **no color restriction at all** — a green creature would
  have been a legal `"white, blue, or black creature"` target.

The repo already documented this as a known gap (the strict-failure comment on
`distribute_core_type_to_or_backfills_every_flat_any_leg`:
_"The surface parser does not yet assemble a ≥3 color-only disjunction — comma/no-comma color chains are a separate gap"_).

## Fix

Relax `parse_color_prefix` (a single building-block change) to accept a color
word followed by **either**:

- a trailing space — the ordinary `"white creature"` prefix (consumed), or
- a comma — the color-list continuation, with the comma **left in place** so the
  existing `", "` / `", or "` disjunction separator consumes it.

That's the whole change. The existing `TYPE_SEPARATORS` recursion +
`distribute_core_type_to_or` machinery then assemble the chain into the **same
`Or`-of-legs shape** the 2-color form already produces, backfilling the concrete
core type onto every color-only leg. **No new AST shape is introduced**, and the
relative-clause form (`"creature that's white, blue, or black"`) was already
handled by `parse_color_disjunction`, so both surfaces are now consistent.

### Resulting parse

`target white, blue, or black creature` →

Or [
Typed { Creature, [HasColor White] },
Typed { Creature, [HasColor Blue]  },
Typed { Creature, [HasColor Black] },
]

## Rules

- **CR 105.1** — color is a characteristic; a color word is an adjective on the
  object it describes.
- **CR 109.2** — a type-word object description restricts to objects of that type.

(type_filters are ANDed in `game/filter.rs`, so backfilling `Creature` onto each
color leg is what keeps the filter from matching noncreature permanents.)

## Testing

- Added a building-block test
  (`or_color_disjunction_three_colors_backfills_core_type`) driving the full
  `parse_target` pipeline for the 3-color prenominal class, asserting a 3-way
  `Or` with each leg carrying `[Creature]` + its color.
- Updated the now-stale gap comment on the distributor-only test.
- **`cargo test -p engine` parser suite: 6740 passed, 0 failed** — including the
  three existing 2-color disjunction tests (Deathmark / Tidebinder /
  Self-Inflicted Wound), confirming no regression.
- `cargo fmt --all` clean.

## Scope

One file (`crates/engine/src/parser/oracle_target.rs`), +67/−7. Pure parser
change — no engine, frontend, or AI surface touched.

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -3361,9 +3361,24 @@ fn parse_combat_relation_suffix(text: &str) -> Option<(FilterProp, usize)> {
 /// then verifies a trailing space exists (color as adjective, not standalone).
 fn parse_color_prefix(text: &str) -> Option<(FilterProp, usize)> {
     let (rest, color) = nom_primitives::parse_color(text).ok()?;
-    // Must be followed by a space (color adjective prefix, not standalone color word).
-    let (rest, _) = tag::<_, _, OracleError<'_>>(" ").parse(rest).ok()?;
-    let consumed = text.len() - rest.len();
+    // CR 105.1: A color word is an adjective prefix only when a separator
+    // follows, so a bare color word ("whiteness") never matches. Two separators
+    // are accepted:
+    //   * a trailing space — the ordinary "white creature" prefix (consumed);
+    //   * a comma — the color-list continuation "white, blue, or black
+    //     creature", where the comma is left in place for the `TYPE_SEPARATORS`
+    //     recursion to consume as a ", " / ", or " disjunction separator. That
+    //     recursion + `distribute_core_type_to_or` then assemble the ≥3-color
+    //     prenominal chain into the same Or-of-legs shape the 2-color "green or
+    //     white creature" form already produces, with the core type backfilled
+    //     onto every color-only leg.
+    let consumed = if let Ok((after_space, _)) = tag::<_, _, OracleError<'_>>(" ").parse(rest) {
+        text.len() - after_space.len()
+    } else if rest.starts_with(',') {
+        text.len() - rest.len()
+    } else {
+        return None;
+    };
     Some((FilterProp::HasColor { color }, consumed))
 }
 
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
Thanks for the focused parser fix. The coverage target and tests make sense, but this head still needs one parser-style correction before I can approve.

`parse_color_prefix` now checks the comma continuation with `rest.starts_with(',')` (`crates/engine/src/parser/oracle_target.rs:3377`). This is non-test Oracle parser dispatch, and the project rule for this parser is to express new structural matching with nom combinators from the first pass. The space branch immediately above already uses `tag::<_, _, OracleError<'_>>(" ").parse(rest)`; please make the comma acceptance use the same parser-combinator style while preserving the current "leave comma for TYPE_SEPARATORS" behavior.

Evidence: current head `dc27533f7c19b0c2788a0e30035ed5b47019cf20`, API diff fetched directly from GitHub, and local PR worktree inspection. Confidence: high. The only thing that would change this conclusion is if this helper were outside Oracle parser dispatch, but it is called from target-filter parsing and the new branch is deciding whether the color prefix structurally matches.

matthewevans (approved):
Current-head approval on d8358ee7be373294cac0e78446368ed119eccbe4. I rechecked the follow-up via GitHub API diff and isolated worktree earlier: the comma continuation now uses nom peek(tag(",")) while preserving the leave-comma-for-TYPE_SEPARATORS behavior, and the production parse_target regression covers the 3+ color prenominal disjunction shape. Exact-head CI is now green and GitHub reports the branch CLEAN.

minion1227:
@matthewevans Thanks for the review — addressed in d8358ee7b.

The comma continuation now uses a nom combinator instead of `starts_with(',')`:
the branch matches `peek(tag::<_, _, OracleError<'_>>(","))` in the same
parser-combinator style as the space branch directly above it, and still leaves
the comma in place for the `TYPE_SEPARATORS` recursion to consume. No behavior
change to the assembled Or-of-legs shape.

CI is green on the new head (mergeable: clean). Ready for another look when you
have a moment.

PR #4446, opened 2026-06-27
fix(parser): apply literal "named <X>" override on token copies (Mishra, Eminent One)

### LINKED ISSUE #4444: [Card Bug] Copying a legendary artifact with "Mishra, Eminent One" incorrectly triggers legendary rule
### Card(s)

Mishra's Eminent One

### Build/version

v0.7.0

### Game mode

Unknown

### Actual behavior

At the start of combat, when using Mishra's rule to select a legendary artifact to copy, the legend rule pop-up asking which one to keep.

### Expected behavior

The legend rule should not be invoked when copying the original legendary artifact because Mishra creates a token copy that has a different name "Mishra's Warform"

### Steps to reproduce

Board State:
- Mishra, Eminent One
- Any Legendary Artifact

Attempt to copy the legendary artifact with Mishra, Eminent One's ability. 

[game-state-turn-20-2026-06-26T23-22-25-846Z.zip](https://github.com/user-attachments/files/29398815/game-state-turn-20-2026-06-26T23-22-25-846Z.zip)

dripsmvcp:
Fixes #4444.

## Summary

Mishra, Eminent One creates a token copy of an artifact, *"except it's a 4/4 Construct artifact creature **named Mishra's Warform** in addition to its other types."* When the copied artifact is **legendary**, the engine wrongly fired the legend rule (CR 704.5j) and prompted "which one to keep" — even though the token's literal name override makes it a differently-named permanent that does not collide with its source.

## Root cause

The native copy-except parser (`become_copy_except`) only recognises the self-name form *"its name is ~"* (keyed to the copying card's own name). It has **no `named <literal>` arm**. So Mishra's exception body `"it's a 4/4 construct artifact creature named mishra's warform in addition to its other types"` routed through `parse_subject_pt_and_types`, whose type-list loop ingested `named` / `mishra's` / `warform` as junk creature subtypes and emitted **no** `SetName`. The token therefore kept the copied legendary's name, producing a legend-rule collision.

## Fix

`split_token_except_clause` (the token-copy entry point, which still holds the original-case text) now peels a literal `named <X>` override off the copy-except body before delegating to the shared `parse_except_clause` building block:

- `strip_copy_except_named_override` captures the name with **original casing** via `TextPair`, bounded by the next copy-exception boundary (`in addition to`, `and`, `with`, `that`, punctuation), and removes the `named <X>` span so the type-list parser stays clean.
- The override is emitted as `ContinuousModification::SetName`, which `apply_token_modifications` (token_copy.rs) already applies to the token's `name`/`base_name`; the legend-rule SBA (sba.rs) groups by name, so the distinct token name no longer collides with the copied source.
- Quoted-ability exceptions (`except it has "…"`) are skipped — a `named` inside a granted ability is part of that ability's text, not a rename of the copy.

Parser-only; reuses the existing `CopyTokenOf` / `SetName` / `apply_token_modifications` runtime. Builds for the class (every "create a token that's a copy of X … named Y" card), not just Mishra.

CR 707.9b + CR 707.2 (copy exceptions / name), CR 704.5j (legend rule).

## Tests

- New `copy_token_with_literal_named_override_emits_set_name`: Mishra's clause -> `CopyTokenOf` whose `additional_modifications` carry `SetName("Mishra's Warform")` (original casing) plus the genuine `SetPower(4)` / `AddType(Artifact)` / `AddSubtype("Construct")` exceptions, and **no** `Named`/`Mishra's`/`Warform` subtype leakage.
- `cargo fmt --all`, `cargo clippy -p engine -- -D warnings`, and the `token` + `become_copy_except` parser test modules (28 + 35 tests) are green.

https://claude.ai/code/session_01JC3pKwdStMrxXkfzRG7xZx

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/crates/engine/src/parser/oracle_effect/token.rs
+++ b/crates/engine/src/parser/oracle_effect/token.rs
@@ -231,30 +231,89 @@ fn split_token_except_clause<'a>(
     ctx: &ParseContext,
 ) -> (&'a str, Vec<Keyword>, Vec<ContinuousModification>) {
     let lower = text.to_lowercase();
-    let Ok((except_input, head_lower)) = parse_token_except_boundary(&lower) else {
+    let Ok((_, head_lower)) = parse_token_except_boundary(&lower) else {
         return (text, Vec::new(), Vec::new());
     };
     let head = &text[..head_lower.len()];
+    // CR 707.9b + CR 707.2: a token-copy exception can rename the copy with a
+    // literal name ("…named Mishra's Warform…", Mishra, Eminent One). Unlike the
+    // self-name "its name is ~" arm — which keys off the copying card's own name
+    // and so cannot apply to a token copy (`card_name` empty below) — a literal
+    // override carries the name in the text itself, so peel it off here (original
+    // case preserved) and strip the "named <X>" span before the body reaches the
+    // shared except parser. Without this the name words leak into the copied
+    // creature's subtype list AND the override is dropped, so a token copying a
+    // legendary permanent keeps the source's name and wrongly collides with it
+    // under the legend rule (CR 704.5j). The original-case except body is
+    // byte-aligned to its lowercase form (mirrors the `head` slice above).
+    let except_original = &text[head_lower.len()..];
+    let (name_override, except_body) = strip_copy_except_named_override(except_original);
+    let except_lower = except_body.to_lowercase();
+
     // Pass the lowercase suffix starting at `[, ]except ` to the shared
     // building block. The except parser is the single authority for the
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed exact head 400b29c8a912902003a633db1ea1d0825b793342. Parser-only token-copy name override is scoped to the literal named-copy exception, preserves original casing, routes through the existing copy-except parser for the rest of the clause, and reaches the token-copy runtime SetName path. CI is green and the PR is CLEAN.

PR #4447, opened 2026-06-27
fix(engine): apply mass force-block to all matching creatures (Predatory Rampage)

### LINKED ISSUE #4233: Predatory Rampage asks for a single target — On cast [[Predatory Rampage]] asks for a target, and only that target (Sid…
Reported in Discord: https://discord.com/channels/1485498006781427802/1518706256448061440/1518706256448061440

**Thread:** Predatory Rampage asks for a single target
**Cards:** predatory rampage

## Summary
On cast [[Predatory Rampage]] asks for a target, and only that target (Sidar in this game state) blocks, but all opponent creatures should block.

### predatory rampage
Creatures you control get +3/+3 until end of turn. Each creature your opponents control blocks this turn if able.

---
<sub>report_id: `discord:1518706256448061440:1518706256448061440:0` · discord: `1518706256448061440/1518706256448061440`</sub>

dripsmvcp:
Fixes #4233.

## Summary

Predatory Rampage — *"Creatures you control get +3/+3 until end of turn. **Each creature your opponents control blocks this turn if able.**"* — did not force the opponents' creatures to block. (The report described it as "asks for a single target, and only that one blocks".)

## Root cause

The native parser already lowers the second sentence **correctly** to a non-targeted mass effect:

```
ForceBlock { target: Typed { type_filters: [Creature], controller: Opponent } }   // no target prompt
```

The bug is in the resolver. `force_block::resolve` only walked the chosen `ability.targets`, which is **empty** for a non-targeted effect — so the must-block requirement was applied to no creature (and could never cover the whole set). Its sibling `force_attack::resolve` already handles the mass case by resolving the effect's `target` filter via `resolved_object_ids_for_filter`; `force_block` simply never adopted that pattern.

## Fix

`force_block::resolve` now resolves the effect's `target` filter through `resolved_object_ids_for_filter` (CR 509.1g), the same helper `force_attack` uses (CR 508.1d). That helper returns the explicit chosen target(s) when present and otherwise expands the filter across the battlefield, so:

- **single-target** ("Target creature blocks this turn if able") — unchanged (the chosen target matches the filter and is returned),
- **source-referential Provoke** ("block this creature if able") — unchanged,
- **mass** ("each creature your opponents control blocks this turn if able") — now forces **every** matching creature to block.

Builds for the class (every "each <filter> blocks this turn if able" effect), not just Predatory Rampage.

CR 509.1c / 509.1g (block requirements), CR 508.1d (the analogous force-attack resolution this mirrors).

## Tests

- `force_block_mass_filter_applies_to_all_matching_creatures` (engine): a non-targeted `ForceBlock { Typed(creature, Opponent) }` with empty `targets` forces **both** opponents' creatures to block (`MustBlock`) and leaves the caster's own creature alone.
- `each_opponent_creature_blocks_if_able_is_non_targeted_mass_force_block` (parser): the Predatory Rampage sentence lowers to `ForceBlock` with an opponent-creature filter and **no** target prompt.
- Existing `force_block` resolver tests (single target, multi-target, Provoke) and parser tests still pass.
- `cargo fmt --all`, `cargo clippy -p engine -- -D warnings` green.

https://claude.ai/code/session_01JC3pKwdStMrxXkfzRG7xZx

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/crates/engine/src/game/effects/force_block.rs
+++ b/crates/engine/src/game/effects/force_block.rs
@@ -237,7 +237,7 @@ mod tests {
         assert_eq!(must_block_count, 2, "Should create one effect per target");
     }
 
-    /// CR 509.1g (issue #4233): a non-targeted mass force-block — Predatory
+    /// CR 509.1c (issue #4233): a non-targeted mass force-block — Predatory
     /// Rampage's "Each creature your opponents control blocks this turn if able"
     /// — carries no chosen targets; the requirement must be applied to every
     /// creature its `target` filter resolves to, not silently to no one (the
--- a/crates/engine/src/game/effects/force_block.rs
+++ b/crates/engine/src/game/effects/force_block.rs
@@ -1,6 +1,7 @@
+use crate::game::targeting::resolved_object_ids_for_filter;
 use crate::types::ability::{
-    ContinuousModification, Duration, EffectError, EffectKind, ResolvedAbility, TargetFilter,
-    TargetRef,
+    ContinuousModification, Duration, Effect, EffectError, EffectKind, ResolvedAbility,
+    TargetFilter,
 };
 use crate::types::events::GameEvent;
 use crate::types::game_state::GameState;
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
Reviewed exact head 5b05f290af98d6f293dfcb1cb2afd0f8fcf41cc0 via GitHub API diff and isolated worktree. The implementation shape looks right, but the new regression comments cite the wrong Comprehensive Rule: force_block.rs and subject.rs describe Predatory Rampage's block-if-able requirement as CR 509.1g. Local CR lookup shows CR 509.1c is the requirement rule for creatures that must block if able; CR 509.1g is the later declaration step where chosen creatures become blocking creatures. Please update the new CR annotations/comments to 509.1c (and any related wording) so the code documents the rule it actually implements.

matthewevans (approved):
Approved current head 17619173b4911e7135ec238db9f018c8880056ab after re-review. The prior CR citation blocker is fixed: the new force-block test comments now cite CR 509.1c, which local CR lookup confirms is the must-block-if-able requirement rule. Exact-head CI is green/CLEAN and the repush guard held.

PR #4463, opened 2026-06-27
fix(parser): Progenitor's Icon chosen-type next-spell flash filter (closes #4462)

### LINKED ISSUE #4462: [Card Bug] Progenitor's Icon — chosen-type next-spell flash grant missing spell_filter
### Card(s)

Progenitor's Icon

### Build/version

main (post-sync 2026-06-27)

### Game mode

Any

### Actual behavior

Progenitor's Icon's `{T}: The next spell of the chosen type you cast this turn can be cast as though it had flash.` activated ability parses the flash grant but drops the `of the chosen type` spell filter. The `GrantNextSpellAbility` effect is emitted with `spell_filter: None`, so the flash permission applies to every next spell instead of only spells of the chosen creature type.

### Expected behavior

Per CR 601.2f + CR 205.2a, `the next spell of the chosen type you cast this turn` must lower to `GrantNextSpellAbility { spell_filter: Typed(IsChosenCreatureType), ... }`, matching the static cost-mod pairing used for Herald's Horn / Cloud Key chosen-type qualifiers.

Oracle:
```
As this artifact enters, choose a creature type.
{T}: Add one mana of any color.
{T}: The next spell of the chosen type you cast this turn can be cast as though it had flash.
```

### Steps to reproduce

1. Put Progenitor's Icon on the battlefield and choose a creature type.
2. Inspect the parsed activated ability for the second `{T}` ability.
3. Observe `spell_filter` is null on the flash grant.

### Additional context

`try_parse_grant_next_spell_ability` already recognizes the `of the chosen type you cast this turn` subject via `parse_next_spell_subject`, but `take_until("spell")` leaves an empty type prefix and the chosen-type discriminator was never re-attached to `spell_filter`.

jony376:
## Summary

Progenitor's Icon (`The next spell of the chosen type you cast this turn can be cast as though it had flash.`) parsed the flash grant without the chosen-type spell filter because `try_parse_grant_next_spell_ability` never re-attached the `of the chosen type` discriminator when the type prefix before `spell` is empty.

This adds `compose_chosen_type_next_spell_filter` (mirroring the static ModifyCost chosen-type pairing) and wires it into the next-spell grant path when the subject is `of the chosen type you cast this turn`.

## Test plan

- [x] `progenitors_icon_next_spell_of_chosen_type_carries_creature_type_filter` — chosen-type flash grant binding
- [x] `progenitors_icon_chosen_type_next_spell_flash_do_not_swallow` — full-card swallow regression
- [x] Existing next-spell grant tests unchanged

Closes #4462


matthewevans (changes requested):
**[HIGH]** The new chosen-type parser shape cannot work at runtime because pending next-spell filters do not retain the source permanent. Evidence: this PR now parses Progenitor's Icon into `Effect::GrantNextSpellAbility { spell_filter: Some(Typed(... IsChosenCreatureType ...)) }`, but `PendingNextSpellModifier` stores only `player`, `modifier`, and `spell_filter` (`crates/engine/src/types/game_state.rs:330-336`). At cast time, `spell_matches_pending_next_spell_filter` calls `spell_matches_cost_filter(state, caster, spell_id, filter, spell_id)` (`crates/engine/src/game/casting.rs:3174-3183`), so `FilterProp::IsChosenCreatureType` looks for the chosen creature type on the spell being cast, not on Progenitor's Icon. The matcher is source-dependent: `spell_object_matches_property` reads `context.source_id` to find `source.chosen_creature_type()` (`crates/engine/src/game/filter.rs:2611-2625`). Why it matters: the parser test is green, but the pending flash permission will fail to match the next creature spell of the Icon's chosen type, so the claimed fix is false-live. Suggested fix: carry the grant source through the pending next-spell modifier (or otherwise preserve source context for source-dependent filter props), use that source id when matching/consuming pending modifiers, and add a production casting regression proving a chosen-type creature spell can be cast with the flash permission while a non-matching spell cannot.

There is also a CR comment cleanup to do while touching this: the new comments cite `CR 205.2a + CR 601.2f`, but this path is a chosen creature-type, source-linked, as-though timing permission. Verified rules that better match the behavior are `CR 205.3m` for creature types, `CR 607.2d` for the linked chosen value, and `CR 609.4`/`CR 601.3b` for as-though flash timing. Confidence: high; source-dependent filter matching is confirmed from the local code trace above and the exact-head GitHub API diff.

matthewevans (commented):
Backend/runtime re-review on current head `20ec2c91f75f3e71193369bd7c8da308970053c6` looks clean to me.

The prior source-context blocker is addressed: `PendingNextSpellModifier` now carries `source_id`, `resolve_grant_next_spell_ability` stamps it from `ability.source_id`, and cast-time matching passes that grant source into `spell_matches_cost_filter` for source-dependent filters such as `IsChosenCreatureType`. The new regression covers the Progenitor Icon shape: Goblin matches the Icon chosen type, Elf does not, and the no-source fallback does not accidentally match.

Exact-head CI is green and merge state is CLEAN. I am intentionally not approving, labeling, or enqueueing from the sweep because this PR includes `client/src/adapter/types.ts`; under the current non-ntindle client/frontend gate, client surface acceptance stays deferred for Matt.

james3773:
Thanks for the catch — you're right that the parser-only fix was false-live.

**Runtime fix (`20ec2c91f`):**
- Added `source_id: Option<ObjectId>` to `PendingNextSpellModifier` (serde-default `None` for backward compat).
- `resolve_grant_next_spell_ability` now sets `source_id: Some(ability.source_id)` when enqueueing the modifier.
- `spell_matches_pending_next_spell_filter` (and all callers) use `entry.source_id.unwrap_or(spell_id)` as the filter context so `IsChosenCreatureType` reads the grant source's chosen type, not the spell being cast.
- Client mirror updated in `types.ts`.

**Casting regression:** `pending_next_spell_chosen_creature_type_flash_uses_grant_source` — Icon with Goblin chosen → Goblin creature spell gains flash, Elf creature spell does not; without `source_id` the Goblin spell also does not match.

**CR citations:** updated to CR 205.3m, CR 607.2d, CR 609.4 / CR 601.3b in `compose_chosen_type_next_spell_filter` and the swallow test.

CI is green on the new commit. Ready for another look.

james3773:
Dropped the \client/src/adapter/types.ts\ mirror in \ea3cdfaef\ so this PR is engine-only again. Runtime still serializes \source_id\; client typing can land separately under the client-surface gate. Ready for approval/merge.

james3773:
@matthewevans Client surface reverted in \ea3cdfaef\ — PR is engine-only again (no \client/\ files). Prior runtime review should still apply. OK to approve + enqueue when CI greens?

matthewevans (approved):
Approved current engine-only head `49af8ea6578d6adc1e2dbac220d3beed68ddf1e5` after exact-head CI completed green/CLEAN. The prior source-context blocker remains addressed: `PendingNextSpellModifier` carries the grant source, matching uses that source for source-dependent filters, and the regression covers chosen-type match, non-match, and no-source fallback.

PR #4468, opened 2026-06-27
fix(parser+engine): Visions of Ruin flashback commander cost reduction (closes #4467)

### LINKED ISSUE #4467: [Card Bug] Visions of Ruin — flashback commander MV cost reduction dropped
## Card

Visions of Ruin

## Oracle

Each opponent sacrifices an artifact. For each artifact sacrificed this way, you create a Treasure token.

Flashback {8}{R}{R}. This spell costs {X} less to cast this way, where X is the greatest mana value of a commander you own on the battlefield or in the command zone.

## Problem

The flashback line also carries a self-spell cost reduction gated on "cast this way" (commander MV). The keyword-only path dropped the trailing `ModifyCost` static, so flashback casts ignored the commander discount.

## Expected

- Parse `Flashback {8}{R}{R}` keyword and a self-spell `ModifyCost` with `CastingAsVariant::Flashback` + greatest-commander-MV dynamic count.
- At runtime, flashback casts reduce generic mana by the greatest commander MV you own.

jony376:
## Summary

Visions of Ruin (`Flashback {8}{R}{R}. This spell costs {X} less to cast this way, where X is the greatest mana value of a commander you own…`) was keyword-only: the trailing commander-MV reduction was swallowed by the flashback keyword path.

- Split compound flashback + self-spell reduction lines in `oracle.rs`
- Parse `to cast this way` as `StaticCondition::CastingAsVariant::Flashback` on self-spell `ModifyCost` statics
- Evaluate that gate during `collect_self_spell_cost_modifiers` (not the layer pipeline)
- Greatest-commander-MV dynamic count reuses existing quantity parsing

## Test plan

- [x] `visions_of_ruin_cast_this_way_cost_reduction_binds_commander_mv` — static parser binding
- [x] `split_flashback_trailing_self_spell_cost_reduction_splits_visions_line` — line split
- [x] `visions_of_ruin_flashback_commander_reduction_parses_without_swallow` — full-card parse
- [x] `visions_of_ruin_flashback_commander_mv_reduces_flashback_cost` — flashback cast cost reduced by commander MV 4

Closes #4467

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -4,8 +4,8 @@ use crate::types::ability::{
     ContinuousModification, CostObjectCount, CostPaidObjectSnapshot, CounterCostSelection,
     Duration, Effect, FilterProp, GameRestriction, ModalSelectionCondition, ObjectScope,
     PlayerFilter, PlayerScope, ProhibitedActivity, QuantityExpr, QuantityRef, ResolvedAbility,
-    RestrictionExpiry, RestrictionPlayerScope, StaticDefinition, TapCreaturesRequirement,
-    TargetFilter, TargetRef,
+    RestrictionExpiry, RestrictionPlayerScope, StaticCondition, StaticDefinition,
+    TapCreaturesRequirement, TargetFilter, TargetRef,
 };
 use crate::types::actions::AlternativeCastDecision;
 use crate::types::card::LayoutKind;
@@ -14876,9 +14876,9 @@ mod tests {
                 }),
             })
             .affected(TargetFilter::SelfRef)
-            .condition(Some(StaticCondition::CastingAsVariant {
+            .condition(StaticCondition::CastingAsVariant {
                 variant: CastingVariant::Flashback,
-            }));
+            });
             def.active_zones = crate::types::zones::self_spell_cost_mod_active_zones();
             obj.static_definitions.push(def);
         }
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -4860,14 +4860,8 @@ pub(super) fn apply_target_dependent_cost_modifiers(
             *mana_cost = super::restrictions::add_mana_cost(mana_cost, &strive_cost);
         }
     }
-    let mut collected = collect_self_spell_cost_modifiers(
-        state,
-        player,
-        object_id,
-        Some(ability),
-        true,
-        None,
-    );
+    let mut collected =
+        collect_self_spell_cost_modifiers(state, player, object_id, Some(ability), true, None);
     collected.extend(collect_battlefield_cost_modifiers(
         state,
         player,
@@ -5054,14 +5048,8 @@ pub(super) fn apply_self_spell_cost_modifiers_with_selected_targets(
     ability: &ResolvedAbility,
     mana_cost: &mut ManaCost,
 ) {
-    let collected = collect_self_spell_cost_modifiers(
-        state,
-        caster,
-        spell_id,
-        Some(ability),
-        true,
-        None,
-    );
+    let collected =
+        collect_self_spell_cost_modifiers(state, caster, spell_id, Some(ability), true, None);
     apply_cost_modifications_in_order(mana_cost, &collected);
 }
 
@@ -5085,13 +5073,9 @@ fn self_spell_cost_condition_matches(
         StaticCondition::Or { conditions } => conditions.iter().any(|cond| {
             self_spell_cost_condition_matches(state, cond, caster, spell_id, casting_variant)
         }),
-        StaticCondition::Not { condition } => !self_spell_cost_condition_matches(
-            state,
-            condition,
-            caster,
-            spell_id,
-            casting_variant,
-        ),
+        StaticCondition::Not { condition } => {
+            !self_spell_cost_condition_matches(state, condition, caster, spell_id, casting_variant)
+        }
         StaticCondition::CastingAsVariant { variant } => casting_variant == Some(*variant),
         _ => super::layers::evaluate_condition(state, condition, caster, spell_id),
     }
@@ -5171,13 +5155,7 @@ fn collect_self_spell_cost_modifiers(
 
         // CR 604.1: Evaluate any trailing condition ("if you control a Wizard").
         if let Some(ref cond) = def.condition {
-            if !self_spell_cost_condition_matches(
-                state,
-                cond,
-                caster,
-                spell_id,
-                casting_variant,
-            ) {
+            if !self_spell_cost_condition_matches(state, cond, caster, spell_id, casting_variant) {
                 continue;
             }
         }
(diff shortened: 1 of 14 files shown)


matthewevans (changes requested):
**[MED]** The new flashback/cost-reduction grammar is implemented with ad hoc string searches instead of parser combinators, and it hardcodes any `"to cast this way"` self-cost line to `CastingVariant::Flashback`. Evidence: `crates/engine/src/parser/oracle_classifier.rs:58-64` uses `lower.find(marker)` to split the compound line, and `crates/engine/src/parser/oracle_static/static_helpers.rs:315-317` uses `lower.contains(" to cast this way")` to attach `CastingAsVariant { Flashback }`. Why it matters: new parser surfaces are required to compose the existing nom/TextPair parser building blocks, and binding the variant inside the generic cost-modifier parser can misclassify future/non-flashback "cast this way" reductions instead of preserving the context that proved this was the flashback clause. Suggested fix: parse the compound flashback line with a small nom combinator that consumes `Flashback <cost>`, the sentence boundary, and the self-spell cost modifier clause, then thread the proven `CastingVariant::Flashback` from that split context into the emitted static rather than detecting it with a generic `contains` scan.

I did verify that this head now adds the self-spell cost-modifier consumer path, so I am not carrying forward Gemini's "consumer missing" point. This head also does not meet the authorized-mode bar yet because exact-head CI is red: Rust lint/parser gate failed and Rust tests shard 1/2 is failing at `c992c4f9e0a6c18d73e2617e02d575e20f10f337`.

matthewevans (changes requested):
**[MED]** Prior parser-shape blocker is not addressed on this head. Evidence: `crates/engine/src/parser/oracle_classifier.rs:58-64` still splits the new compound flashback grammar with `lower.find(marker)`, and `crates/engine/src/parser/oracle_static/static_helpers.rs:315-317` still attaches `CastingAsVariant { Flashback }` via a generic `lower.contains(" to cast this way")` scan. Why it matters: the follow-up moved where the split runs in `parse_oracle_ir`, but the new grammar is still not expressed as a composable parser and still infers Flashback from generic wording rather than from the proven flashback parse context. Suggested fix: replace the split helper with a nom/TextPair parser for `Flashback <cost>. <self-spell cost modifier>` and pass the known `CastingVariant::Flashback` into the static emitted for the reduction instead of detecting it inside the generic cost-modifier parser.

Exact-head CI is still red on `507d08485025c3139f69cc97cd7144fe19a8aba3`: Rust lint/parser gate failed, with Rust tests and card-data still running at the time of this review.

matthewevans (changes requested):
Re-reviewed current head 70bca39ce094de79a2c9176f5c9c446914df47a8 after the follow-up commits. The prior blocker is only partially addressed, so this is still blocked.

The latest commit replaces one raw `contains` call with `scan_contains`, but the parser shape problem remains: `split_flashback_trailing_self_spell_cost_reduction` still splits the compound flashback line with `lower.find(marker)`, and the generic self-spell cost-modifier parser still infers `CastingVariant::Flashback` from generic "to cast this way" wording rather than receiving the already-proven Flashback context from the compound flashback parser.

Please make the flashback + trailing self-cost-modifier line a real parser/context handoff: parse the compound Flashback line via the parser building blocks/TextPair/nom path, then thread the known `CastingVariant::Flashback` into the self-cost modifier parse for that trailing fragment. I should not be able to reach generic "to cast this way" text elsewhere and have it silently become Flashback just because of those words.

Also, the branch is still BEHIND and exact-head Rust/card-data checks are still in progress, so there is no approval/label/enqueue from the sweep.

matthewevans (changes requested):
Re-reviewed current head 5d30accb3ce286206f4330789dcccc8f60f0198b. This is still the same blocker.

The new commit changes the spell-resolution continuation loop so it stops before a later Flashback line, and it narrows the phrase scan to `less to cast this way`, but it still does not make the requested parser/context handoff. `split_flashback_trailing_self_spell_cost_reduction` still uses `lower.find(marker)` to split the compound Flashback line, and `try_parse_cost_modification` still decides `CastingVariant::Flashback` from generic "less to cast this way" text rather than receiving the already-proven Flashback context from the compound Flashback parser.

Please replace the string splitter with a real parser for the `Flashback <cost>. <self-spell cost modifier>` shape and thread the known Flashback variant into the trailing self-cost modifier parse. Moving where the split runs, changing the continuation loop, or narrowing the scan phrase does not address that architecture issue.

GitHub also reports exact-head Rust lint/parser gate failed, and the branch is still BEHIND. No approval, label, or enqueue from the sweep.

matthewevans (changes requested):
Re-reviewed current head bcfd424fb06335abc22fdbfb04f6cfde65ba63b8. The head is now green/CLEAN and the latest commit moved the compound Flashback intercept before static-pattern parsing, but the requested architecture fix is still not present.

The same two problems remain on this head:

- `split_flashback_trailing_self_spell_cost_reduction` still splits the new compound Flashback grammar with `lower.find(marker)` in `crates/engine/src/parser/oracle_classifier.rs`.
- `try_parse_cost_modification` still infers `CastingVariant::Flashback` from generic `less to cast this way` text in `crates/engine/src/parser/oracle_static/static_helpers.rs`, instead of receiving Flashback as context from the parser that already proved the leading clause is Flashback.

The new `oracle.rs` intercept changes ordering and prevents the static pattern from swallowing the Flashback line, but it still feeds `parse_static_line(reduction_part)` and lets the generic static parser self-detect Flashback from text. That is the part I keep asking to remove.

Please make the compound `Flashback <cost>. <self-spell cost modifier>` shape a real parser/context handoff: parse/split the compound line via the parser building blocks, and pass the known `CastingVariant::Flashback` into the trailing self-cost modifier parse rather than rediscovering it from generic words.

No approval, label, or enqueue from the sweep.

matthewevans (changes requested):
Re-reviewed current head 53c2f31d3dbe40a0343c769224fed4d3d6f5194d. The previous architecture blocker is addressed: the compound Flashback splitter now uses the parser path (`nom_on_lower`/`take_until`), and the trailing self-spell cost modifier receives caller-proven `CastingVariant::Flashback` instead of inferring Flashback from generic `cast this way` text. That is the right shape.

I still cannot approve this head because exact-head CI is red in ways introduced by this follow-up:

- Rust lint fails at `cargo fmt --all -- --check` with formatting diffs in `crates/engine/src/parser/oracle.rs`.
- Both Rust test shards fail to compile with E0659 at `crates/engine/src/parser/oracle_static/tests.rs:2452`: `parse_flashback_trailing_self_spell_cost_reduction` is ambiguous because it is imported through both `super::support::*` and `super::*`.

Please run rustfmt and disambiguate that test helper/function reference so the exact head can compile and go green. No approval, label, or enqueue from the sweep while this head is red/BEHIND.

matthewevans (approved):
Approved current head 95ff7d2835305c7c54b4d58b1f9fa3894c990241 after re-review. Prior parser/context handoff blocker is fixed, the rustfmt/E0659 follow-up is fixed, and exact-head CI is green/CLEAN. Repush guard held on the reviewed head.

PR #4470, opened 2026-06-27
fix(parser): UNSUPPORTED one-off: Adipose Offspring

ntindle:
## Summary

**Root cause:** UNSUPPORTED one-off: Adipose Offspring

## Cards corrected
- Adipose Offspring

## Fix
Implemented the full, rules-correct fix for Cluster 56 (Adipose Offspring — Emerge "instead create X of those tokens, where X is the sacrificed creature's toughness"). The card was UNSUPPORTED on upstream/main (the second sentence fell to Effect::Unimplemented). After the change, oracle-gen produces the correct structure with ZERO Unimplemented: parent trigger Token{Alien 2/2 white, count Fixed(1)} with sub_ability Token{count Ref(Toughness{CostPaidObject})} gated by ConditionInstead{CastVariantPaid{Emerge}}. A boy-scout fix to the same of-those-tokens helper also corrected The Final Days (its where-X count now binds ZoneCardCount{Graveyard,Creature} instead of the latent Variable("X")).

Three gaps closed exactly per the approved plan: (A parser routing) new nom combinator parse_cast_variant_cost_paid_condition recognizes "[possessive] emerge cost was paid" (tag()-based dispatch over a shared CAST_VARIANT_COST_PAID_PHRASES constant, filtered to Emerge so the existing sneak/surge/spectacle/prowl Route-2 path is untouched), wired into try_nom_condition_as_ability_condition; (B count binding) match_create_of_those_tokens now binds a trailing ", where X is <expr>" clause via parse_event_context_quantity/parse_cda_quantity; (C engine plumbing) new GameObject.cast_cost_paid_object field carries the emerge-sacrificed creature's LKI from the resolving spell's ResolvedAbility.cost_paid_object onto the permanent (stack.rs copy), preserves it across CR 400.7 battlefield-entry reset via CastLinkSnapshot (zone_pipeline.rs), tags cast_variant_paid=Emerge at resolution (stack.rs), and reference-gated-propagates it into the ETB ResolvedAbility in build_triggered_ability (triggers.rs). Added one typed enum variant CastVariantPaid::Emerge (registry-enum sibling, CR 702.119; no exhaustive matches to update — all reads are generic tuple comparisons).

Verification (worktree, cargo run directly — not under Tilt): cargo fmt clean; cargo clippy -p engine --all-targets exit 0 with ZERO warnings; cargo test -p engine fully green (116 test binaries, 0 failures) including 4 new tests; parser combinator gate clean for all my added lines (one inline allow-noncombinator for trailing-period punctuation cleanup). oracle-gen confirms Adipose parses with 0 Unimplemented and The Final Days is fixed.

Judgement calls: (1) The reference-gate predicate — QuantityRef has 80+ variants, so rather than an unwieldy exhaustive accessor I composed it from existing building blocks Effect::count_expr()/Effect::target_filter() plus a focused scope check over the 8 scope-bearing QuantityRef variants using the local `_ => false` boolean-probe convention (matching filter_references_parent_target_combat_relation), an exhaustive 11-arm QuantityExpr recursion, and Pump P/T via PtValue. (2) Gap B placement — I empirically confirmed via oracle-gen that the chunk-level where-x stripping does NOT pre-strip the of-those-tokens anaphor, so the plan's match_create extension is the correct seam (and it transparently fixed The Final Days). (3) Runtime proof — the GameRunner SpellCast driver does not handle emerge sacrifice (PayCost) payment, so a full emerge-cast pipeline test isn't cleanly supported without extending the shared driver (out of scope). I instead wrote a build_triggered_ability runtime test proving the only fully-novel runtime link (Gap C propagation + the reference gate, positive + negative); the remaining links (emerge sacrifice cost_paid_object stamping, Toughness{CostPaidObject} resolution, CastVariantPaid instead-swap) are already covered by existing engine tests.

Deviation from plan: used the existing pub(crate) parse_event_context_quantity/parse_cda_quantity (which actually map "the sacrificed creature's toughness" → Toughness{CostPaidObject}) rather than the plan-named parse_where_x_quantity_expression; same result, and the plan listed this exact fallback chain.

Incidental artifact: running oracle-gen for verification regenerated crates/engine/data/oracle-subtypes.json (a 1-line subtype data drift, "Shi-Ar", unrelated to this fix) — left in the tree for the orchestrator to decide on; not part of the source change.

## Files changed
- crates/engine/src/game/stack.rs
- crates/engine/src/game/game_object.rs
- crates/engine/src/game/triggers.rs
- crates/engine/data/oracle-subtypes.json

## CR references
- CR 702.119a-c (line 4850/4852) — Emerge alternative cost; verified grep -n "^702.119" docs/MagicCompRules.txt
- CR 400.7d (line 1956) — primary for Gap C: an ability of a permanent may reference what costs were paid to cast the spell that became it; verified grep -n "^400.7d"
- CR 400.7 (line 1948) — cast-link fields cleared on entry, restored via CastLinkSnapshot; verified grep -n "^400.7"
- CR 608.2k (line 2810) — secondary intra-ability cost-referent cite in the predicate helpers; verified grep -n "^608.2k"
- CR 107.3 / CR 107.3c (lines 464 / 470) — Gap B: X defined by the card's text (the where-X clause), controller doesn't choose; verified grep -n "^107.3c"
- CR 603.4 (line 2588) — intervening-if condition re-checked at resolution; verified grep -n "^603.4"

## Verification
- `cargo fmt --all` — pass (no changes)
- `check-parser-combinators.sh (merge-base 3571104377)` — pass (after in-loop fix: converted 3 hand-constructed Effect::Unimplemented literals in lower.rs test fixtures to Effect::unimplemented() constructor)
- `cargo clippy -p engine --all-targets -- -D warnings` — pass (exit 0, fix's own files clean)
- `cargo test -p engine` — pass (0 failed; cluster test match_create_of_those_tokens_binds_trailing_where_x_clause ok)
- `oracle-gen --filter "adipose offspring"` — pass (parses correctly; ConditionInstead + Toughness{CostPaidObject}, no Unimplemented)
Cards confirmed re-parsed correctly: adipose offspring

### DIFF
--- a/crates/engine/src/bin/oracle_gen.rs
+++ b/crates/engine/src/bin/oracle_gen.rs
@@ -345,9 +345,11 @@ fn write_oracle_subtypes(
         .into_iter()
         .collect();
     let out_path = PathBuf::from("crates/engine/data/oracle-subtypes.json");
+    // Emit a trailing newline so the committed file is POSIX-clean and a
+    // re-generation never produces a spurious no-newline diff.
     match serde_json::to_string_pretty(&list)
         .map_err(|e| e.to_string())
-        .and_then(|json| std::fs::write(&out_path, json).map_err(|e| e.to_string()))
+        .and_then(|json| std::fs::write(&out_path, format!("{json}\n")).map_err(|e| e.to_string()))
     {
         Ok(()) => eprintln!(
             "Wrote {} creature subtypes to {}",
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -42605,6 +42605,168 @@ mod tests {
                 other => panic!("expected AlternativeCastChoice(Emerge), got {other:?}"),
             }
         }
+
+        /// Oracle text for Adipose Offspring (WHO). The ETB trigger's parent
+        /// branch makes one token; the `instead` sub-branch (gated on the emerge
+        /// cost having been paid) makes X tokens, where X is the sacrificed
+        /// creature's toughness.
+        const ADIPOSE_OFFSPRING_ORACLE: &str = "Emerge {5}{W} (You may cast this spell by sacrificing a creature and paying the emerge cost reduced by that creature's mana value.)\nWhen this creature enters, create a 2/2 white Alien creature token. If this creature's emerge cost was paid, instead create X of those tokens, where X is the sacrificed creature's toughness.";
+
+        /// Count tokens a player controls on the battlefield.
+        fn token_count(runner: &crate::game::scenario::GameRunner, player: PlayerId) -> usize {
+            runner
+                .state()
+                .objects
+                .values()
+                .filter(|o| o.zone == Zone::Battlefield && o.controller == player && o.is_token)
+                .count()
+        }
+
+        /// CR 702.119a + CR 400.7d — Adipose Offspring full emerge chain. Casting
+        /// the real parsed card via emerge (sacrificing a toughness-3 creature)
+        /// must drive the entire pipeline end to end: the emerge sacrifice stamps
+        /// the cost-paid object, the cast-link snapshot restores it onto the
+        /// entering permanent, `cast_variant_paid` is set to `Emerge`, the ETB
+        /// trigger's `instead` branch is selected, and `X` resolves to the
+        /// sacrificed creature's toughness (via `Toughness { CostPaidObject }`
+        /// LKI per CR 400.7d). The result is exactly 3 Alien tokens, not 1.
+        #[test]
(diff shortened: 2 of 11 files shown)


matthewevans (approved):
Reviewed the current head db375c8e. I specifically checked the parser/lowering route for the emerge-cost-paid instead clause, the cast-cost-object snapshot through sacrifice payment -> stack -> zone pipeline -> triggered ability resolution, and the Adipose end-to-end hard-cast/emerge tests. GitHub CI is green at this head. Gemini's CR-range formatting note is not a blocker: the project review guide explicitly allows CR range forms such as 702.119a-c.

PR #4471, opened 2026-06-27
feat: Cosmic Cube + Hawkeye, Young Avenger (MSH)

lgray:

## MSH cluster F — Cosmic Cube + Hawkeye, Young Avenger

Two Marvel's Spider-Man (MSH) cards, each a reusable building block rather than a one-off.

### Cosmic Cube — dynamic mana-value ceiling on cast-from-library
- New parser path `parse_with_mana_value_constraint` lifts the cast-permission ceiling from a fixed integer to a dynamic `QuantityExpr` (`Ref(Aggregate{Max, Power})`), delegating to the remainder-returning `nom_quantity::parse_quantity_ref` so the trailing clause is preserved.
- Cosmic Cube is the **first** card to route a dynamic (`Ref`) ceiling through `cast_permission_constraint_allows_cast`; all prior cards used `Fixed`. Covered by a focused finalize-seam unit test (`cosmic_cube_dynamic_mv_ceiling_enforced_at_finalize`) proving the ceiling is enforced at cast finalization, not just at parse.

### Hawkeye, Young Avenger — dynamic damage-amplification replacement
- `DamageModification::Plus { value }` lifted from `u32` to `QuantityExpr`, so the damage bonus can scale off a dynamic source (here, source power) instead of a constant.
- Legacy integer JSON still deserializes to `Fixed` (back-compat test `damage_modification_plus_legacy_int_deserializes_to_fixed`).

### Verification
- Targeted unit tests: cosmic_cube (4/0), plus_dynamic_source_power (2/0), plus_legacy (1/0) — all pass.
- Full engine lib suite: **13812 passed / 0 failed** against current `main` (v0.8.0) — no regression.
- clippy clean; integration suite 8/0.

### Post-merge follow-ups (non-blocking)
- Regenerate `data/engine-inventory.json` to reflect the `DamageModification::Plus` field-type change (`u32` → `QuantityExpr`).
- Run `cargo coverage` / `cargo semantic-audit` to refresh MSH coverage snapshot.

Assisted-by: ClaudeCode:claude-opus-4.8

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -49485,4 +49485,104 @@ mod tests {
              (PaymentContext::Spell); got {spell_res:?}"
         );
     }
+
+    /// MSH-F Sub-Plan A2 — runtime enforcement of the *dynamic* mana-value
+    /// ceiling at the finalize seam. Cosmic Cube is the FIRST card to route a
+    /// `CastPermissionConstraint::ManaValue { value: Ref(Aggregate{Max, Power,
+    /// attacking creatures you control}) }` through
+    /// [`cast_permission_constraint_allows_cast`]; every pre-existing dynamic
+    /// ceiling resolves to a computed `Fixed`, so the
+    /// `resolve_quantity(Aggregate) -> comparator.evaluate` composition at this
+    /// seam had zero direct coverage. This drives the seam directly (no
+    /// card-data, no harness).
+    ///
+    /// CR 202.3: mana value is the comparison subject. CR 601.2e: cast legality
+    /// is re-checked at finalization once the resulting mana value is known.
+    ///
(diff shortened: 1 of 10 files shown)


matthewevans (approved):
Reviewed current head aa4e88ec4057ba39f4bfbade9c42d83b13e6bf67 via GitHub API diff and isolated worktree. No blocking findings: Cosmic Cube's dynamic mana-value cast constraint is enforced at the cast-finalization seam with fixed/dynamic parser coverage, and Hawkeye's dynamic damage amplification flows through the replacement pipeline with source-power, literal plus, bare-X legacy, and serde regressions. Gemini's greedy-rest concern is not a blocker for this implementation because the Hawkeye path delegates the consumed quantity through parse_cda_quantity and the covered source-power form gates the remainder rather than silently accepting an arbitrary suffix. CI is green and the repush guard held.

PR #4473, opened 2026-06-27
fix(parser): Blademane Baku for-each bare counter removed (closes #4472)

### LINKED ISSUE #4472: fix(parser): Blademane Baku for-each counter removed prefix
## Card

Blademane Baku

## Oracle

Whenever you cast a Spirit or Arcane spell, you may put a ki counter on this creature.

{1}, Remove X ki counters from this creature: For each counter removed, this creature gets +2/+0 until end of turn.

## Problem

The activated tail uses prefix-form `For each counter removed, …` without a `this way` suffix. `parse_for_each_clause` only recognized `counter removed this way`, so `strip_for_each_prefix` bailed out and the Pump kept `TargetFilter::Any` instead of threading `SelfRef` from `this creature`.

## Expected

- `parse_for_each_clause("counter removed")` → `QuantityRef::PreviousEffectAmount`
- `For each counter removed, this creature gets +2/+0 until end of turn` lowers to a `Pump` with `SelfRef` target and a `repeat_for` multiplier

jony376:
## Summary

- Recognize bare `counter[s] removed` in `parse_for_each_clause` (Blademane Baku activated tail).
- Un-ignore `for_each_prefix_pump_threads_self_ref_target` — Pump now gets `SelfRef` + `repeat_for`.

## Test plan

- [x] `for_each_bare_counter_removed_is_previous_effect_amount`
- [x] `for_each_prefix_pump_threads_self_ref_target`
- [x] Parser combinator gate

Closes #4472

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -49897,7 +49897,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "known: prefix-form 'For each [unrecognized clause], <body>' fails to thread subject onto Pump target — body Pump retains target=Any sentinel because parse_for_each_clause doesn't recognize 'counter removed' and the strip_for_each_prefix bail-out short-circuits subject threading. Tracked as a follow-up."]
     fn for_each_prefix_pump_threads_self_ref_target() {
         // Blademane Baku: "For each counter removed, this creature gets +2/+0
         // until end of turn" — prefix-form for-each. The Pump's target should
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -49905,6 +49905,13 @@ mod tests {
             "For each counter removed, this creature gets +2/+0 until end of turn",
             AbilityKind::Spell,
         );
+        assert_eq!(
+            def.repeat_for,
+            Some(QuantityExpr::Ref {
+                qty: QuantityRef::PreviousEffectAmount,
+            }),
+            "repeat_for should scale by counters removed in the activation cost"
+        );
         match &*def.effect {
             Effect::Pump { target, .. } => {
                 assert_eq!(
--- a/crates/engine/src/parser/oracle_quantity.rs
+++ b/crates/engine/src/parser/oracle_quantity.rs
@@ -2554,6 +2554,18 @@ fn parse_for_each_clause_with_they_controller(
         }
     }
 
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
Reviewed current head b93c83b6edd53326454d7804b7a06da6ce441fd5 via GitHub API diff, existing Gemini comments, local CR lookup, and a PR worktree at /tmp/phase-pr-4473-review. I am not carrying Gemini's parser-style finding: the new branch uses all_consuming(parse_counters_removed_phrase), which is an existing nom combinator helper rather than a new ad hoc full-Oracle string match. Two blockers remain:\n\n**[MED]** Replace the incorrect CR citation. Evidence: crates/engine/src/parser/oracle_quantity.rs:2557 cites CR 106.1, but local docs/MagicCompRules.txt shows CR 106.1 only defines mana as the primary resource; it does not describe counter removal or the same-resolution value that PreviousEffectAmount reads. Why it matters: this repo treats verified CR annotations as part of rules correctness, and a wrong CR number gives future maintainers false confidence about the rule being implemented. Suggested fix: change the comment to a verified relevant citation, e.g. CR 608.2c + CR 122.1 for same-resolution instruction linkage plus counters, or add CR 602.2 if you want to mention activation/cost payment explicitly.\n\n**[MED]** Assert the restored repeat count in the unignored Blademane regression. Evidence: crates/engine/src/parser/oracle_effect/mod.rs:49908 only matches Effect::Pump and asserts target == SelfRef; it never asserts AbilityDefinition.repeat_for, even though this PR's summary says Pump now gets SelfRef + repeat_for and the bug was the leading for-each prefix failing to thread through. Why it matters: a future regression could keep the SelfRef target while dropping repeat_for, making Blademane Baku resolve the pump once instead of once per counter removed, and this test would still pass. Suggested fix: extend for_each_prefix_pump_threads_self_ref_target to assert def.repeat_for == Some(QuantityExpr::Ref { qty: QuantityRef::PreviousEffectAmount }).

matthewevans (approved):
Reviewed current head 8d2b575e6ab76d04b825829f40503dd1ee964737 via GitHub API diff, existing Gemini comments, local CR lookup, review-impl rubric, and isolated worktree. The follow-up addresses both prior blockers: the counter-removal comment now cites verified CR 608.2c + CR 122.1, and the Blademane regression now asserts repeat_for == PreviousEffectAmount as well as the SelfRef pump target. Gemini's remaining parser-style comment is refuted because the branch reuses the existing all_consuming(parse_counters_removed_phrase) nom helper. Exact-head CI is green/CLEAN and the repush guard held.

PR #4475, opened 2026-06-27
fix(parser): UNSUPPORTED one-off: The Pandorica

ntindle:
## Summary

**Root cause:** UNSUPPORTED one-off: The Pandorica

## Cards corrected
- The Pandorica

## Fix
Implemented the full rules-correct fix for The Pandorica (cluster #52), which was UNSUPPORTED (Unimplemented "can't phase in" + a dropped delayed-trigger disjunct + wrong inner phase-in target) on upstream/main 932748cfd. Dispatched each clause onto existing primitives plus exactly one new engine leaf.

WHAT CHANGED (7 files):
1. types/statics.rs — added StaticMode::CantPhaseIn leaf (CR-annotated), wired into as_keyword/Display/FromStr + display_roundtrips enumeration. add-engine-variant gate: it is a new per-rule-action prohibition leaf in the Cant* cluster (denies the CR 702.26a phase-in TBA), not a parameterization of an existing leaf; categorical boundary = CR 702.26 only. No new Effect, no new DelayedTriggerCondition (WhenNextEvent.or_trigger already exists).
2. parser/oracle_effect/subject.rs — added value([CantPhaseIn], tag("phase in")) atom to parse_restriction_list_atom; added CantPhaseIn to static_mode_needs_grant_propagation (so it registers as a SpecificObject transient grant on the parent target); 2 building-block tests.
3. parser/oracle_effect/mod.rs — added nom helpers parse_self_event_verb / scan_self_event_trigger / parse_self_disjunctive_event_trigger and a new branch in try_parse_inline_delayed_trigger (before the single-event scan_delayed_condition_kind path) that captures "when ~ <eventA> or <eventB>" as WhenNextEvent{trigger, or_trigger}; rebind_triggering_source_to_parent_target rewrites the inner demonstrative phase-in target to ParentTarget; 1 test.
4. game/static_abilities.rs — registered CantPhaseIn->handle_rule_mod; added object_has_active_cant_phase_in (condition-gated; covers SpecificObject-pinned grant, filter-scoped grant, printed static).
5. game/phasing.rs — excluded locked permanents from execute_untap_step_phasing's phase_in_ids (CR 702.26a override); 2 runtime tests (lock tracks source tap state; untap-step TBA respects lock + positive control).
6. game/effects/phase_out.rs — CR 101.2 gate in resolve_phase_in so explicit phase-in cannot override an active lock (the card's own delayed trigger passes because it resolves after the source untaps/leaves, when the condition is false).
7. game/coverage.rs — CantPhaseIn classification arm.

VERIFICATION (worktree NOT under Tilt, ran cargo directly): reparse = 0 Unimplemented, correct AST; cargo fmt --all --check clean; cargo clippy -p engine --all-targets clean (0 warnings); cargo test -p engine --lib = 13798 baseline pass + 5 new tests pass, 0 failures; parser combinator diff gate clean (no string dispatch in added parser lines).

RUNTIME PROOF (not deferred): the load-bearing new behavior — the CantPhaseIn lock interacting with the untap-step phase-in TBA — is proven by a runtime test exercising execute_untap_step_phasing with a registered ForAsLongAs{SourceIsTapped} TCE (stays out while source tapped; phases in once source untaps). The delayed-trigger firing reuses already-tested WhenNextEvent disjunction (triggers.rs:5104) + parent_target_snapshot infrastructure.

## Files changed
- crates/engine/src/types/statics.rs
- crates/engine/src/game/static_abilities.rs
- crates/engine/src/game/phasing.rs
- crates/engine/src/game/effects/phase_out.rs

## CR references
- CR 702.26a — phasing is a TBA at each player's untap step; the override The Pandorica imposes (grep -n '^702.26a' docs/MagicCompRules.txt: FOUND)
- CR 101.2 — a 'can't' effect takes precedence over a 'can'/'will' (grep -n '^101.2' : FOUND)
- CR 611.2b — 'for as long as' continuous effects re-evaluate their condition each cycle; the lock self-expires (grep -n '^611.2b' : FOUND)
- CR 110.5d — a permanent not on the battlefield is neither tapped nor untapped; lock lifts when the source leaves (grep -n '^110.5d' : FOUND)
- CR 603.6 / 603.6c — zone-change / leaves-the-battlefield trigger events (grep -n '^603.6' and '^603.6c' : FOUND)
- CR 603.7c — delayed triggered abilities track objects by reference; parent_target_snapshot path (grep -n '^603.7c' : FOUND)
- CR 502.3 — untap step; effects can keep permanents from untapping (grep -n '^502.3' : FOUND)
- CR 702.26c / 702.26d — phased-in status change; phasing never causes a zone change and zone-change triggers don't fire on phasing (grep -n '^702.26c' / '^702.26d' : FOUND)

## Verification
- `cargo fmt --all` — pass
- `./scripts/check-parser-combinators.sh <merge-base upstream/main>` — pass
- `cargo run --profile tool --features cli --bin oracle-gen -- data --filter "the pandorica"` — pass
Cards confirmed re-parsed correctly: The Pandorica

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -1773,6 +1773,10 @@ fn fmt_delayed_condition(cond: &DelayedTriggerCondition) -> String {
         }
         DelayedTriggerCondition::WhenDiesOrExiled { .. } => "when dies or exiled".into(),
         DelayedTriggerCondition::WheneverEvent { .. } => "whenever event this turn".into(),
+        DelayedTriggerCondition::WhenNextEvent {
+            lifetime: crate::types::ability::DelayedTriggerLifetime::Persistent,
+            ..
+        } => "when next event (persistent)".into(),
         DelayedTriggerCondition::WhenNextEvent { .. } => "when next event this turn".into(),
     }
 }
@@ -7787,6 +7791,9 @@ fn audit_card_lines(oracle_text: &str, face: &CardFace) -> Vec<SemanticFinding> 
             StaticMode::CantUntap => {
                 effective_lower.contains("doesn't untap") || effective_lower.contains("don't untap")
             }
+            // CR 702.26a + CR 101.2: The Pandorica's "It can't phase in for as
+            // long as ~ remains tapped".
+            StaticMode::CantPhaseIn => effective_lower.contains("can't phase in"),
             StaticMode::CantAttack => effective_lower.contains("can't attack"),
             StaticMode::CantBlock => effective_lower.contains("can't block"),
             StaticMode::CantAttackOrBlock => effective_lower.contains("can't attack or block"),
--- a/crates/engine/src/game/effects/delayed_trigger.rs
+++ b/crates/engine/src/game/effects/delayed_trigger.rs
@@ -249,6 +249,7 @@ fn bind_contextual_filter_to_condition(
         DelayedTriggerCondition::WhenNextEvent {
             trigger,
             or_trigger,
+            ..
         } => {
             for filter in [
                 &mut trigger.valid_card,
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -4428,6 +4428,13 @@ fn extract_event_context_filter(effect: &Effect) -> Option<&TargetFilter> {
         | Effect::PreventDamage { target, .. }
         | Effect::Connive { target, .. }
         | Effect::PhaseOut { target, .. }
+        // Symmetry with PhaseOut across the PhaseOut/PhaseIn walker class. No
+        // current card hydrates a `PhaseIn` target from event context (The
+        // Pandorica snapshots `PhaseIn{ParentTarget}` into `ability.targets` at
+        // CreateDelayedTrigger resolution rather than carrying a
+        // TriggeringPlayer/DefendingPlayer filter), but listing it keeps the
+        // walker complete should such a card appear.
+        | Effect::PhaseIn { target, .. }
         | Effect::ForceBlock { target, .. }
         | Effect::ForceAttack { target, .. }
         | Effect::PutAtLibraryPosition { target, .. }
--- a/crates/engine/src/game/effects/phase_out.rs
+++ b/crates/engine/src/game/effects/phase_out.rs
@@ -99,6 +99,15 @@ pub fn resolve_phase_in(
     // skipping the phased-out exclusion.
     let targets = collect_phase_in_targets(state, ability, &target);
     for oid in targets {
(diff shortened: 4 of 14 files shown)


matthewevans (changes requested):
**[MED]** The self-event disjunct parser can accept a partial event phrase and silently ignore the qualifier. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:1928`-`1959` parses bare `leaves` as `TriggerMode::LeavesBattlefield`, and `scan_at_word_boundaries` returns success while discarding the parser remainder (`crates/engine/src/parser/oracle_nom/primitives.rs:856`-`864`). That means a future self-disjunctive clause like `when ~ leaves your graveyard or ...` would be accepted as a battlefield-leaves trigger instead of staying unsupported. Why it matters: this new path is a coverage-unlocking parser for delayed self-events; accepting a prefix and dropping the rest makes unsupported qualified events look semantically supported. Suggested fix: make each disjunct parser consume the whole event phrase after the optional self subject, or remove the bare `leaves` arm and use an `all_consuming`/remainder check so only `leaves the battlefield` maps to `LeavesBattlefield`.

matthewevans (changes requested):
Reviewed current head `4830a141a5bccbe06e4c42bfd39b73ebd4cee712` via the GitHub API diff, current bot/human review context, CR lookup, and a PR worktree at `/tmp/forge-pr-4475`.

**[MED]** The self-disjunctive event parser still accepts qualified event phrases by matching only a prefix. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:1958`-`1960` calls `scan_at_word_boundaries(segment, parse_self_event_verb)`, and `scan_at_word_boundaries` discards the parser remainder on success (`crates/engine/src/parser/oracle_nom/primitives.rs:856`-`864`). The follow-up removed the bare `leaves` arm, but `parse_self_event_verb` still returns success for prefixes such as `becomes untapped` or `leaves the battlefield` without requiring that the rest of the disjunct is consumed (`crates/engine/src/parser/oracle_effect/mod.rs:1928`-`1950`). Why it matters: the new class advertises support for self-referential disjunctive delayed events, but a future or currently unsupported qualifier like `when ~ becomes untapped during your upkeep or leaves the battlefield, ...` would be accepted as an unconditional `Untaps` event instead of remaining unsupported. Suggested fix: have `scan_self_event_trigger` strip the optional self subject and require `all_consuming(parse_self_event_verb)`/empty remainder for each disjunct, with a regression that a trailing qualifier after a recognized event phrase remains `Unimplemented`.

matthewevans (approved):
Reviewed current head. The self-event disjunction now requires full verb consumption, covers both untap/leave re-entry with a persistent delayed trigger, and the phasing lock/re-entry paths have discriminating runtime coverage with green CI.

PR #4477, opened 2026-06-27
fix(parser+engine): Wretched Banquet least-power destroy gate

### LINKED ISSUE #4476: Wretched Banquet: least-power destroy gate dropped during parsing
## Card

Wretched Banquet

## Oracle

Destroy target creature if it has the least power among creatures.

## Problem

The intervening-if superlative gate (`if it has the least power among creatures`) is dropped during spell-effect parsing. `parse_inner_condition` can represent superlative comparisons, but the effect-chain stripper only handles fixed property thresholds (mana value, power equals N, etc.). The destroy body parses, yet the resolution gate is missing — illegal targets are not rejected at resolution.

## Expected

- Strip the suffix into `AbilityCondition::QuantityCheck` with `ObjectScope::Target` power on the LHS and `AggregateFunction::Min` over creatures on the RHS (`Comparator::LE` for ties).
- At runtime, `evaluate_condition` returns true only when the targeted creature ties for least power among creatures on the battlefield.

jony376:
## Summary

Wretched Banquet (`Destroy target creature if it has the least power among creatures.`) lost its intervening-if superlative gate during spell-effect parsing.

- Add `try_parse_spell_target_has_superlative` for spell-target `if it has the least/greatest <property> among <filter>` (distinct from trigger-anchored `build_superlative_comparison`)
- Add `strip_superlative_target_conditional` and wire it into the effect-chain condition stripper after mana-value stripping
- Parser, swallow-check, and runtime `evaluate_condition` tests

## Test plan

- [x] `strip_superlative_target_conditional_least_power` — stripper unit test
- [x] `wretched_banquet_least_power_condition` — effect-chain parser binding
- [x] `wretched_banquet_least_power_evaluates_per_target` — runtime gate per target power
- [x] `wretched_banquet_least_power_destroy_parses_without_swallow` — full-card parse

Closes #4476

### DIFF
--- a/crates/engine/src/parser/oracle_effect/conditions.rs
+++ b/crates/engine/src/parser/oracle_effect/conditions.rs
@@ -4,13 +4,15 @@ use crate::parser::oracle_nom::error::{OracleError, OracleResult};
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take_until};
 use nom::character::complete::char;
+use nom::character::complete::multispace0;
 use nom::combinator::{all_consuming, opt, peek, value};
-use nom::sequence::{preceded, terminated};
+use nom::sequence::{preceded, terminated, tuple};
 use nom::Parser;
 
 use super::super::oracle_nom::bridge::{nom_on_lower, nom_parse_lower};
 use super::super::oracle_nom::condition::{
-    inject_controller_you, parse_cast_using_teamwork_phrase, try_parse_spell_target_has_superlative,
+    inject_controller_you, parse_cast_using_teamwork_phrase,
+    parse_spell_target_superlative_suffix,
 };
 use super::super::oracle_nom::primitives as nom_primitives;
 use super::super::oracle_nom::quantity as nom_quantity;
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**[MED]** The new superlative condition strip is still using phrase splitting as parser dispatch instead of composing the condition grammar. Evidence: `crates/engine/src/parser/oracle_effect/conditions.rs:1892`-`1900` lowercases the whole text and splits on the fixed fragment `" if it has the "`, then hands only the suffix to the nom parser. Why it matters: this creates a one-axis parser for the Wretched Banquet wording and misses the natural sibling grammar the bot called out (`if they have the ...`) while adding another non-nom dispatch path in Oracle parsing. Suggested fix: parse the trailing conditional as a nom combinator over the full suffix boundary, with an `alt` for singular/plural subject text and an all-consuming handoff to `try_parse_spell_target_has_superlative`/its inner grammar, instead of using `TextPair::rsplit_around` for the dispatch split.

matthewevans (changes requested):
**[MED]** The new nom suffix parser validates the type phrase but returns the wrong remainder, so the Wretched Banquet condition still won't be stripped. Evidence: `crates/engine/src/parser/oracle_nom/condition.rs:2215` calls `parse_type_phrase(rest)` and checks `remainder`, but the success path at `condition.rs:2237`-`2238` returns `rest` rather than `remainder`. Why it matters: `strip_superlative_target_conditional` requires the mapped remainder to be empty after optional punctuation, so returning the pre-type phrase text (for example, `creatures`) makes the parser fail even though the type phrase was recognized; the new Wretched Banquet tests should fail for the same reason. Suggested fix: return the `remainder` from `parse_type_phrase` as the nom rest after mapping/validating it, with the existing all-consuming strip tests kept in place.

matthewevans:
The parser-return blocker from my last review looks addressed on the current follow-up, and I accepted the update-branch merge to bring the PR current.

Holding approval until exact-head CI finishes green on merge-only head `4a37ef3`; Rust lint/tests, card-data, and frontend checks are still pending right now. No new code blocker found in this pass.

james3773:
Addressed both review items on `c447da51c` (CI green):

1. **Nom suffix dispatch** — `strip_superlative_target_conditional` now probes each ` if ` boundary and hands the suffix to `parse_spell_target_superlative_suffix`, which uses an `alt` for `it has the` / `they have the` before the superlative grammar. Added `strip_superlative_target_conditional_plural_subject` test.

2. **`parse_type_phrase` remainder** — `parse_spell_target_has_superlative` now mirrors `parse_subject_has_superlative_form`: advance past the consumed type phrase and return the tail (e.g. `.`) for the outer `terminated(opt(tag(".")), …)` stripper. Wretched Banquet parser/swallow/runtime tests pass.

matthewevans (approved):
Current head `4a37ef3` clears my prior blocker.

The `TextPair::rsplit_around` parser-dispatch issue was replaced with a nom suffix parser covering singular/plural subjects, and the later return-remainder issue is fixed by returning the `parse_type_phrase` remainder while the outer strip still requires full suffix consumption after optional punctuation. Exact-head CI is green/CLEAN, and the repush guard held after assignment.

PR #4478, opened 2026-06-27
chore(agents): grant SendMessage to sub-agents for graceful cull + report-back

lgray:

## Grant `SendMessage` to custom sub-agents (graceful cull + report-back)

The four custom agent definitions in `.claude/agents/` each declare an explicit `tools:` list, none of which included `SendMessage`:

| Agent | tools (before) |
|---|---|
| `engine-implementation-executor` | Read, Edit, Write, Bash, Grep, Glob |
| `mtg-rules-auditor` | Read, Grep, Glob, LS, Bash, Write |
| `parser-gap-finder` | Read, Grep, Glob, Bash |
| `pr-review-comment-resolver` | Bash, Edit, MultiEdit, Read, Glob, Grep, Task, TodoWrite, WebFetch |

Without `SendMessage`, when these are spawned as background teammates they cannot:
- **acknowledge a `shutdown_request`** → they must be tmux-pane-killed, which leaves zombie roster entries; and
- **report progress/completion back to the orchestrating lead** → they can only write to disk.

### Change (additive only)
- Add `SendMessage` to each of the four agents' `tools:` lists, with a one-paragraph note at each agent's return-point describing the new capability.
- The grant is **strictly additive** — both existing contracts are explicitly preserved in the prose:
  - the *"final text IS the return value to the orchestrator"* contract, and
  - the durable `.planning/*.md` disk-report pattern (survives context compaction).
- Document the graceful-cull flow (`shutdown_request` → `shutdown_response`) in three skills: `batch-mechanics` (team cleanup), `engine-implementer` (orchestrator job list), `bug-triage` (implementer-feedback liveness).

### Notes
- `.codex/agents/*.toml` are thin wrappers that delegate to the `.claude/agents/*.md` source of truth and carry no `tools:` key — no change needed.
- `.agents/skills` and `.codex/skills` are symlinks to `.claude/skills`, so the skill-doc edits propagate automatically (no duplication).
- Markdown/doc-only; no engine, parser, or runtime code touched.

Assisted-by: ClaudeCode:claude-opus-4.8

### DIFF
--- a/.claude/agents/engine-implementation-executor.md
+++ b/.claude/agents/engine-implementation-executor.md
@@ -1,7 +1,7 @@
 ---
 name: engine-implementation-executor
 description: Execute an already-reviewed phase.rs implementation plan surgically. Receives the approved plan + scope, edits files, runs Tilt-first verification, and returns a diff summary with any judgement-call notes. Does NOT plan, does NOT review, does NOT commit. Spawned by the `/engine-implementer` skill.
-tools: Read, Edit, Write, Bash, Grep, Glob
+tools: Read, Edit, Write, Bash, Grep, Glob, SendMessage
 model: opus
 ---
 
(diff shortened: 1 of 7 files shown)


matthewevans (approved):
Content review is clean on current head `9e4b63e`.

I reviewed the GitHub API diff in an isolated worktree. The tool grant is narrowly additive (`SendMessage` only), the touched agent/skill text keeps final structured reports and disk reports authoritative, and the graceful-shutdown path is framed as a progress/ack channel rather than a replacement for normal handoff output. Exact-head CI is green/CLEAN, and the repush guard held after assignment.

PR #4482, opened 2026-06-27
feat(engine): The Ruinous Wrecking Crew + Hawkeye, Master Marksman (MSH)

lgray:

## MSH set rollout — batch E

Implements two more previously-unsupported Marvel's Spider-Man (MSH) cards: **The Ruinous Wrecking Crew** and **Hawkeye, Master Marksman**.

### The Ruinous Wrecking Crew — coverage-audit fix
A dynamic "choose up to X / that many" modal header was being miscounted as multiple Oracle lines by the coverage auditor, hiding the card's true support state. This folds the dynamic-cap modal header in the oracle-line counter so the count is honest, and adds a **swallow detector** that keeps a modal whose dynamic cap was dropped **conservatively unsupported (RED)** rather than letting it slip through as a false-green. Errs toward under-claiming support, never over-claiming.

The number of times the optional cost is paid becomes the **dynamic cap on a reflexive modal choice**. New building blocks:

- **`QuantityRef::TimesCostPaidThisResolution`** (CR 603.12a) — sizes the reflexive modal (CR 700.2d) from the count of successful payments this resolution.
- **A dedicated repeated-optional-payment driver** — counts successful payments and resolves the reflexive ability once iff `K ≥ 1`. Gated to **synchronous-mana costs only** via `is_synchronous_mana_pay_cost`; pausing/asynchronous costs fall through to the generic payment path (no behavior change for them).
- **Serde durability:** the times-paid counter (`optional_cost_payments_this_resolution`) is nonzero at the mid-loop `OptionalEffectChoice` pause — a serde boundary. It is serialized-when-nonzero and equality-included (mirroring `exiled_from_hand_this_resolution`), so save/restore mid-loop preserves the modal cap (CR 700.2d) instead of collapsing it.

These are **class-level building blocks**, not card-specific shims: `TimesCostPaidThisResolution` + the repeated-optional-payment driver cover any "pay a cost up to N times, then act that many times" card.

### Verification
- `cargo build -p engine` clean; `cargo clippy -p engine --lib --tests -- -D warnings` clean.
- Discriminating MSH-E tests (`game::marksman_tests`, `modal_dynamic_max_dropped`, `count_effective_oracle_lines_folds`): **15 passed / 0 failed**. Non-vacuity measured by serde-roundtrip discrimination (K=0 vs K=2) and by reverting the swallow detector (the dropped-cap modal flips false-green).
- Full `engine` suite: **13857 passed / 0 failed**, 6 ignored.
- All CR annotations (603.12a, 700.2d, 118.1, 118.3a) grep-verified against the Comprehensive Rules.

Cherry-picked cleanly onto current `main`.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -1484,6 +1484,9 @@ fn fmt_quantity_ref(qty: &QuantityRef) -> String {
             }
         }
         QuantityRef::ConvokedCreatureCount => "creatures that convoked this spell".into(),
+        QuantityRef::TimesCostPaidThisResolution => {
+            "times the repeated optional cost was paid this resolution".into()
+        }
         QuantityRef::ManaSpentToCast { scope, metric } => {
             format!("mana spent to cast ({scope:?}, {metric:?})")
         }
@@ -5460,7 +5463,20 @@ fn is_modal_header_line(lower: &str) -> bool {
         "choose any number",
         "choose x.",
     ];
-    CHOOSE_PHRASES.iter().any(|p| lower.contains(p))
+    if CHOOSE_PHRASES.iter().any(|p| lower.contains(p)) {
+        return true;
+    }
+    // CR 700.2 + CR 107.3m: a dynamic modal header ("choose up to X —",
+    // "choose up to that many.") plus its bulleted modes is one logical unit;
+    // fold the bullets into the header so a parsed modal (1 parent + N
+    // children) is not miscounted as N+1 dropped Oracle lines. The cap is a
+    // resolution- or cast-time value (CR 107.3m for cast X), not a fixed word.
+    // A loose substring match here cannot false-green a card on its own — the
+    // load-bearing honesty gate is the Modal_DynamicMaxDropped swallow detector,
+    // and a non-modal "choose up to X <nouns>" selection clause has no bullets
+    // to fold (so folding leaves its line count unchanged).
+    const DYNAMIC_CHOOSE_HEADERS: &[&str] = &["choose up to x", "choose up to that many"];
+    DYNAMIC_CHOOSE_HEADERS.iter().any(|p| lower.contains(p))
 }
 
 /// Strip structural formatting prefixes from an Oracle line, returning the
@@ -6436,6 +6452,7 @@ fn quantity_ref_feature(qref: &QuantityRef) -> (&'static str, FeatureSupport) {
             ("AdditionalCostPaymentCountFor", Handled)
         }
         QuantityRef::ConvokedCreatureCount => ("ConvokedCreatureCount", Handled),
+        QuantityRef::TimesCostPaidThisResolution => ("TimesCostPaidThisResolution", Handled),
         QuantityRef::ManaSpentToCast { .. } => ("ManaSpentToCast", Handled),
         QuantityRef::EventContextSourceCostX => ("EventContextSourceCostX", Handled),
         QuantityRef::ColorsInCommandersColorIdentity => {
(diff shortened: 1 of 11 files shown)


matthewevans (changes requested):
VERDICT: Request changes

This is a useful direction for Hawkeye and the coverage-honesty work, but I found one runtime blocker in the repeated-payment driver.

- `crates/engine/src/game/effects/mod.rs:4086-4108` advances to another payment prompt whenever `remaining > 0`, even if the accepted payment failed. The code correctly avoids incrementing K when `cost_payment_failed_flag` is set, but it still re-stashes `PendingRepeatedOptionalPayment` and offers the next "again" prompt. That does not match the printed/reflexive structure: CR 603.12a says the special one-shot reflexive happens when the cost is paid one or more times, and CR 118.3 says a player cannot pay a cost without the necessary resources. If this iteration did not pay the cost, the "If you do, you may pay {1} again" condition was not satisfied. The loop should stop and finish with the payments already made, not offer another payment.

  The new regression in `crates/engine/src/game/marksman_tests.rs:234-245` currently protects the wrong behavior: it accepts a second `{1}` with no mana, then expects a third optional prompt to exist so it can decline it. A discriminating test should assert that a failed accepted payment terminates the repeated-payment loop immediately and offers the reflexive once with cap K=1, without requiring or allowing a further payment prompt.

Evidence checked: GitHub API diff at current head `3ed3345c9b42a17fdbf7cdf4e0185e30547db9f7`, isolated worktree inspection, local CR lookup for CR 603.12a / 118.3 / 700.2d, and existing payment authority behavior where failed resolution-time payments set `cost_payment_failed_flag`.

No approval/label/enqueue while this is live. Exact-head CI was also still running for Rust shard 2 and card-data when reviewed.

lgray:

Thanks both — addressed in a follow-up commit.

**@matthewevans — the runtime blocker (failed payment offering another prompt).** Confirmed and fixed. You're right: in `resolve_repeated_optional_payment_choice` the `if remaining > 0` branch re-stashed `PendingRepeatedOptionalPayment` and offered the next "pay {1}" prompt regardless of whether the just-accepted payment actually succeeded. That contradicts the printed structure — "you may pay {1} up to three times. When you do, …" — because **CR 118.3** (a player can't pay a cost without the resources to pay it fully) means a failed payment did not happen, and **CR 603.12a** ties the reflexive to payments that were actually made. So an accepted-but-unpaid iteration must end the sequence with the payments already made.

Fix: the continuation is now nested inside the success branch — only a payment with `!cost_payment_failed_flag` both increments K **and** offers another "pay again" prompt. A failed payment falls through to `finish_repeated_optional_payment`, which resolves the reflexive once iff K ≥ 1 (and never, at K = 0).

The test you flagged was rewritten accordingly. It is now a positive assertion of termination rather than of a spurious third prompt: with funding for a single `{1}`, accept (K=1) → accept-but-no-mana (fails) and then assert the loop has ended — `pending_repeated_optional_payment.is_none()` and the reflexive modal is offered exactly once capped at `(0, 1)`. Discriminators are documented inline: dropping the K-guard regresses K to 2, and restoring the old unconditional `if remaining > 0` leaves `pending` Some with no modal — both fail the new assertions.

**@gemini-code-assist — defensive `condition` clear.** Applied: `payment_unit.condition = None` is now stripped alongside `repeat_for`/`sub_ability`/`optional` when building the per-iteration payment unit, so a top-level resolution-time condition (evaluated once in the depth-0 prelude) can't re-gate each `{1}` payment.

Verification (isolated worktree off current `main`, commit `a996bbf8d`): `cargo clippy -p engine --lib --tests -- -D warnings` clean; `game::marksman_tests` 8/8 green. **Discrimination measured:** reverting only the loop-termination fix (restoring the unconditional `if remaining > 0`) while keeping the new test makes it fail at the termination assertion — the engine is left at a spurious third `OptionalEffectChoice { … }` instead of the reflexive modal — confirming the test is non-vacuous. CR annotations (118.3, 603.12a, 700.2d) grep-verified against the Comprehensive Rules.

matthewevans (approved):
VERDICT: approve

Re-reviewed the current head `a996bbf8db540cd9c98fd0422be51ca91bf95448` after the failed-payment follow-up.

The prior blocker is addressed: `resolve_repeated_optional_payment_choice` only offers the next repeated `{1}` prompt after a successful payment, so an accepted-but-unpaid iteration now falls through to `finish_repeated_optional_payment` and resolves the reflexive once with the payments already made. The regression now asserts the discriminating behavior directly: with one mana available, the second accepted payment fails, `pending_repeated_optional_payment` is cleared, and the modal cap is `K=1` rather than requiring or permitting a third prompt.

I also checked the Gemini follow-up: clearing the cloned per-iteration `condition` keeps the top-level resolution condition from re-gating each `{1}` payment. Exact-head CI is green/CLEAN, and the repush guard held after assignment reread.

PR #4485, opened 2026-06-27
fix(parser): represent dropped ActivateOnlyDuring clauses (closes #2238)

### LINKED ISSUE #2238: Parser drops ActivateOnlyDuring clauses while coverage still marks affected cards supported
## Summary

`./scripts/gen-card-data.sh` reports `2` `Swallow:ActivateOnlyDuring` diagnostics across `2` cards. Both are otherwise marked supported by coverage.

## Repro

```bash
./scripts/gen-card-data.sh
target/tool/coverage-report data --all --warning-detector ActivateOnlyDuring --warning-limit 5
```

Fresh drilldown:

```text
matched_cards: 2
supported_cards: 2
single_gap_cards: 0
examples: Katara, Water Tribe's Hope, Loch Larent
```

## Expected

Activation timing restrictions should be represented in activation constraints. If they cannot be parsed, the activated ability should not remain fully supported.

minion1227:
Closes #2238

## Summary

#2238 reports that the parser **drops `ActivateOnlyDuring` clauses** while
coverage still marks the affected cards supported. Both flagged cards are fixed
here, each via a distinct root cause:

| Card | Clause | Root cause |
|------|--------|------------|
| **Loch Larent** | `Activate only during your turn and only once` | conjunction rider stranded |
| **Katara, Water Tribe's Hope** | `… X can't be 0. Activate only during your turn.` | mid-line annotation truncated the rest of the ability |

## Fix 1 — compose "activate only \<timing\> and only once" riders

Each timing arm in `strip_activated_constraints` anchors on the literal
`"activate"`, so a conjoined `"and only once"` / `"and only once each turn"` tail
left the sentence ending in `"… and only once"`, which no arm matched — the whole
clause was swallowed. Now the limit rider is **peeled compositionally** before the
timing arms run, so the bare `"activate only <timing>"` core matches its own arm
next pass. This composes the limit and timing axes (per CLAUDE.md *compose, don't
enumerate*) and covers the **whole** `activate only <timing> and only once[ each
turn]` class. Guarded on a preceding `"activate only"` clause so an effect that
merely ends in `"and only once"` is never mis-stripped.

## Fix 2 — excise mid-line "X can't be 0" without dropping trailing text

`strip_x_cant_be_zero_suffix` located the annotation with `rfind` and returned
`line[..pos]` — **truncating** at the annotation. Fine for a true suffix, but it
discarded any sentence that followed. Katara's annotation sits mid-ability, so
`"Activate only during your turn."` was thrown away and the activated-ability
parser never saw the timing gate. The annotation is now **excised in place**:
keep everything before, re-attach what follows. Reminder text is already stripped
by the caller and `min_x_value` is captured before the strip, so no semantics are
lost; the suffix/standalone cases are byte-for-byte unchanged.

Both `DuringYourTurn` and `OnlyOnce` are already runtime-enforced
(`restrictions.rs`), so the gates now actually apply.

## Rules

CR 602.5b (printed once-each-turn limit) + CR 602.5c (timing restriction).

## Testing

Added full-parse regression tests for Loch Larent, Katara, and a minimal
conjunction probe. `cargo test -p engine --lib parser::oracle::tests` —
**474 passed, 0 failed**, including the existing
`parses_compound_activate_only_constraints`,
`spell_x_cant_be_zero_annotation_sets_min_x_value`, and City-of-Solitude
regression tests (no regression). `cargo fmt --all` clean.

### DIFF
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -5003,23 +5003,12 @@ pub(super) fn strip_activated_constraints(text: &str) -> (String, ActivatedConst
         // enumerating every timing × limit pairing. Guarded on a preceding
         // "activate only" clause so an effect sentence that merely ends in "and only
         // once" is never mis-stripped. ("each turn" form first: longest match.)
-        for (rider, restriction) in [
-            (
-                " and only once each turn",
-                ActivationRestriction::OnlyOnceEachTurn,
-            ),
-            (" and only once", ActivationRestriction::OnlyOnce),
-        ] {
-            if let Some(prefix) = lower.strip_suffix(rider) {
-                if prefix.contains("activate only") {
-                    let end = remaining.len() - rider.len();
-                    remaining = remaining[..end]
-                        .trim_end_matches(|c: char| c == '.' || c == ',' || c.is_whitespace())
-                        .to_string();
-                    constraints.restrictions.push(restriction);
-                    continue 'parse_constraints;
-                }
-            }
+        if let Some((kept_len, restriction)) = peel_only_once_rider(&lower) {
+            remaining = remaining[..kept_len]
+                .trim_end_matches(|c: char| c == '.' || c == ',' || c.is_whitespace())
+                .to_string();
+            constraints.restrictions.push(restriction);
+            continue 'parse_constraints;
         }
 
         if let Some(prefix) = lower.strip_suffix("activate no more than twice each turn") {
@@ -5242,6 +5231,40 @@ fn strip_condition_suffix(
     true
 }
 
+/// CR 602.5b + CR 602.5c: Peel a trailing "and only once [each turn]"
+/// activation-limit rider that conjoins onto an "activate only <timing>" clause
+/// ("Activate only during your turn and only once", Loch Larent). Forward nom
+/// combinators locate the rider (`take_until`) and confirm it trails an
+/// "activate only" clause, composing the limit axis with the timing axis rather
+/// than enumerating every timing × limit pairing. Returns the byte length of the
+/// text to keep (everything before the rider) and the limit restriction.
+fn peel_only_once_rider(lower: &str) -> Option<(usize, ActivationRestriction)> {
+    let (rider_onward, before) = take_until::<_, _, OracleError<'_>>(" and only once")
+        .parse(lower)
+        .ok()?;
+    // The rider must trail an "activate only ..." clause, never an effect
+    // sentence that merely ends in "and only once".
+    take_until::<_, _, OracleError<'_>>("activate only")
+        .parse(before)
+        .ok()?;
+    // "each turn" is the optional longest-match tail; the rider must end the line.
+    let (rest, each_turn) = preceded(
+        tag::<_, _, OracleError<'_>>(" and only once"),
+        opt(tag::<_, _, OracleError<'_>>(" each turn")),
+    )
+    .parse(rider_onward)
+    .ok()?;
+    if !rest.is_empty() {
+        return None;
+    }
+    let restriction = if each_turn.is_some() {
+        ActivationRestriction::OnlyOnceEachTurn
+    } else {
+        ActivationRestriction::OnlyOnce
+    };
+    Some((before.len(), restriction))
+}
+
 /// Strip trailing "X can't be 0." / "This ability can't be copied and X can't
 /// be 0." constraint annotations from Oracle text. These are activation/casting
 /// restrictions that annotate X-cost abilities but are not themselves effects.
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
VERDICT: request-changes

**[MED]** The new timing-limit rider is still string-dispatched instead of parameterized through the parser grammar. Evidence: `crates/engine/src/parser/oracle.rs:5013-5014` peels `" and only once"` with `lower.strip_suffix(rider)` and gates it with `prefix.contains("activate only")`; exact-head CI also fails the parser-combinator gate on these added lines. Why it matters: this keeps the activation timing x limit axis split between the existing enumerated suffix list and a post-hoc string scan, rather than one composable nom suffix axis, so the fix is not at the parser bar for this repo. Suggested fix: factor `and only once [each turn]` as a nom suffix parser and compose it with the existing activation timing restriction parser, then remove/retire the hardcoded timing x limit combinations this new axis supersedes.

**[MED]** The X-can't-be-zero annotation excision also trips the new-parser string-method gate. Evidence: `crates/engine/src/parser/oracle.rs:5272` uses `trimmed.rfind(suffix)` to locate the annotation and `crates/engine/src/parser/oracle.rs:5282` uses `strip_prefix('.')`; the same exact-head lint job reports these added lines under the parser-combinator violation. Why it matters: the behavior target looks right, but parser cleanup for mid-line annotations should still be expressed via the accepted nom/structural parsing pattern or an explicitly justified structural escape hatch. Suggested fix: rewrite the annotation split with the repo's nom delimiter/suffix pattern, or add a narrow `allow-noncombinator` only for genuinely structural punctuation cleanup after a nom-recognized annotation.

minion1227:
@matthewevans Both points addressed in 818c79957 — the parser-combinator gate
is now green.

1. **Activation-limit rider** — removed the `strip_suffix(rider)` +
   `contains("activate only")` string dispatch. New `peel_only_once_rider`
   helper is a forward nom combinator (`take_until` + `preceded`/`opt`) that
   composes the limit axis with the timing axis instead of enumerating
   timing × limit pairings, and confirms the rider trails an `activate only`
   clause via `take_until` rather than `contains`.

2. **X-can't-be-0 excision** — the annotation is now located with `take_until`
   (longest "this ability…" form first) instead of `rfind`, and the trailing
   sentence period is peeled with `opt(tag("."))` instead of `strip_prefix`.

Behavior-preserving: the Loch Larent / Katara / Kirol / crew / compound-timing
activation-restriction tests all still pass. Ready for re-review.

matthewevans (changes requested):
VERDICT: request-changes

Current-head re-review after 818c799577691b99d9fd1abcda00d3d10efb55d8. The parser-combinator CI gate is now green, and the new  helper is directionally better than the previous  /  follow-up. But this still is not the parameterized timing x limit model Matt asked for.

**[MED]** The new helper is additive instead of replacing the existing timing x limit proliferation. Evidence:  still has a hardcoded table for , , , and , and that table runs before the new  path at . Why it matters: the PR fixes Loch Larent by adding another path for a missing combination, but the core grammar still has two authorities for the same timing x limit axis. This is parameterization debt: future timing forms can still be handled by the old enumerated table, the new peel helper, or both depending on ordering. Suggested fix: make  a single composed suffix axis used by the activation timing parser, then remove the hardcoded timing x limit combinations it supersedes. The existing  and new Loch/Katara tests should still pass through that one authority.

Evidence sources: GitHub API diff, exact-head CI rollup (green/CLEAN), isolated worktree at 818c7995, and local CR lookup confirming CR 602.5b/602.5c. Confidence: high. The counter-evidence I would need is a reason the hardcoded timing x limit table must remain as a separate parser authority; I do not see that in this diff. No approval/label/enqueue from the sweep.

matthewevans:
Clarifying the current-head review above: shell quoting stripped the inline code identifiers from the review body, but the blocker is specific.

The additive helper is `peel_only_once_rider`, and the old string-method blockers were `strip_suffix` / `contains`. The remaining parameterization issue is that `crates/engine/src/parser/oracle.rs:4816-4844` still has a hardcoded timing x limit table for:

- `activate only as a sorcery and only once each turn`
- `activate only as a sorcery and only once`
- `activate only during your turn and only once each turn`
- `activate only during your upkeep and only once each turn`

That table runs before the new `peel_only_once_rider` path around `crates/engine/src/parser/oracle.rs:5006`, so the PR now has two authorities for the same timing x limit axis. Please make `and only once [each turn]` one composed suffix axis used by the activation timing parser, then remove the hardcoded timing x limit combinations it supersedes. The existing `parses_compound_activate_only_constraints` test and the new Loch/Katara regressions should pass through that one authority.

matthewevans (approved):
VERDICT: approve

Current-head re-review after `68f6deb2d8f97418c8ab848614ccf3e68235507b`. The prior parameterization blocker is addressed: the hardcoded timing x limit table is gone, `peel_only_once_rider` now owns the `and only once [each turn]` suffix axis, and the parser then re-enters the existing bare timing arms instead of enumerating timing-limit pairings. The Katara/Loch Larent regressions cover both the mid-line `X can't be 0.` excision and the timing+once conjunction, and exact-head CI is green/CLEAN. I also verified the CR 602.5b/c citations against the local Comprehensive Rules copy.

PR #4491, opened 2026-06-27
feat(engine): support 'you or a permanent becomes the target' triggers (Loki, God of Mischief)

lgray:

## What

Implements **Loki, God of Mischief** (MSH): *"Whenever you or a permanent you control becomes the target of a spell or ability an opponent controls, draw a card. Do this only once each turn."*

## How — building blocks, not a special case

This is the general **"a player OR a permanent becomes the target"** mixed-subject trigger class (covers Loki, Valkmira Protector's Shield, and any future "a player or <permanent-type>" subject), not a one-off:

- **New `BecomesTargetAbility` parser recognizer** (nom combinators only) for the ability-only "becomes the target of … an ability" form, reusing `TriggerMode::BecomesTarget` (no new mode), placed in an under-capacity `alt` block.
- **New `TriggerDefinition.valid_subject_player` field**, kept **distinct** from the existing `valid_target`. `valid_target` is overloaded as the **effect-target slot** (it is set whenever an effect says "target opponent/player", e.g. Venerated Rotpriest's "target opponent gets a poison counter"). Routing the subject's player leaf into its own field means an object-subject trigger with a player-targeting effect can never be mistaken for a player **subject** and over-fire. The subject `Or`-split routes the player leaf → `valid_subject_player`, the object leaf → `valid_card`. Back-compat preserved via `#[serde(default, skip_serializing_if = "Option::is_none")]`.
- **Battlefield zone gate** on the permanent leaf (`InZone(Battlefield)`) so a creature card in a graveyard is not treated as a permanent.
- **Once-per-turn** via the existing `TriggerConstraint::OncePerTurn` (zero new code).

## Comprehensive Rules

- **CR 115.1** — targets span objects *and* players (the two-axis `valid_card` + `valid_subject_player` model).
- **CR 110.1** — a permanent is a card/token *on the battlefield* (the `InZone` gate).
- **CR 603.2h** — "do this only once each turn".

## Testing

- New integration tests (`loki_becomes_target_ability.rs`) + parser/matcher unit tests.
- **Regression fence**: `becomes_target_object_subject_with_player_targeting_effect_does_not_fire_on_player` — an object-subject trigger whose *effect* targets a player must NOT fire on a player target (the Venerated Rotpriest shape). Proven non-vacuous: it fails under the naive overloaded-`valid_target` read and passes with the field separation.
- End-to-end over regenerated card-data: Loki fires on both a player and a permanent; Venerated Rotpriest no longer over-fires; zero new `Unknown` parses; existing `BecomesTarget` cards unchanged.
- `cargo fmt`/`clippy -D warnings` clean; full `cargo test -p engine` green.

### DIFF
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -2340,10 +2340,28 @@ pub(super) fn match_becomes_target(
                 *object_id == source_id
             }
         }
+        // CR 115.1 + CR 603.2e: a player becomes the target. Two independent ways a
+        // becomes-target trigger can fire on a player target, kept apart because
+        // `valid_target` is overloaded as the EFFECT-target slot:
+        //   (1) PURE player subject (no object axis) — e.g. "Whenever you become the
+        //       target of a spell". The subject filter lives in `valid_target`; the
+        //       retained `valid_card.is_none()` guard prevents an OBJECT-subject
+        //       trigger whose EFFECT targets a player (Venerated Rotpriest: "...a
+        //       creature you control becomes the target..., target opponent gets a
+        //       poison counter") from over-firing on a player target.
+        //   (2) MIXED "a player or <permanent>" subject (Loki) — the SUBJECT's player
+        //       leaf is routed to `valid_subject_player`, distinct from the effect
+        //       slot, so it fires on a player target even though `valid_card` carries
+        //       the permanent half.
         TargetRef::Player(player_id) => {
-            trigger.valid_card.is_none()
+            let pure_player_subject = trigger.valid_card.is_none()
                 && trigger.valid_target.is_some()
-                && valid_player_matches(trigger, state, *player_id, source_id)
+                && valid_player_matches(trigger, state, *player_id, source_id);
+            let mixed_subject_player = trigger
+                .valid_subject_player
+                .as_ref()
+                .is_some_and(|filter| player_matches_filter(filter, state, *player_id, source_id));
+            pure_player_subject || mixed_subject_player
         }
     }
 }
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
VERDICT: request-changes

Exact-head CI is green/CLEAN and the core engine direction may be workable, but this branch is not in an approvable shape yet.

**[MED]** The PR includes unrelated generated/client/docs state outside the Loki engine feature. Evidence: the GitHub API diff for `1ac46de74f9fe116b957c3090e06c6a16c3e16d8` changes `client/public/changelog-meta.json`, `client/public/changelog.json`, `lobby-worker/README.md`, and `scripts/changelog/state.json` in addition to the engine trigger/parser/tests files. Why it matters: this is a feature PR for a trigger matcher/parser change; generated changelog metadata and a lobby-worker README update are unrelated review surface and should not ride along, especially in the authorized sweep where frontend/client/UI/public asset changes are handled conservatively. Suggested fix: drop the unrelated changelog/client public/lobby-worker/script state files from the branch so the PR is only the engine trigger/parser/test change.

I stopped at this packaging blocker rather than approving/enqueueing. Evidence sources: GitHub API diff, exact-head CI rollup, and isolated worktree at the current head. Confidence: high. The counter-evidence I would need is an explanation that those generated/client/docs files are intentionally required by the Loki trigger implementation; I do not see that connection in the PR body or diff. No approval/label/enqueue from the sweep.

matthewevans (approved):
Approved on current head 1ac46de74f9fe116b957c3090e06c6a16c3e16d8, superseding my earlier branch-hygiene review after rechecking the GitHub pull-files API and refreshing local origin/main. The current PR diff is engine/test only. I reviewed the parser/matcher/type changes in an isolated worktree, verified the CR annotations against docs/MagicCompRules.txt, and confirmed exact-head CI is green with mergeStateStatus CLEAN. The implementation keeps the ability-only source modeled separately from spell-or-ability, separates mixed subject player leaves from effect target slots via valid_subject_player, gates permanent object leaves to battlefield for the Loki ability path, and includes matcher/parser/integration regressions for Loki plus Rotpriest-style overfire. No approval blockers found.

Issue #323, opened 2026-05-11
[Card Bug] Treasured Find — exiles target card and puts itself in graveyard (effect targets swapped)


matthewevans:
### Card

Treasured Find (the Discord thread title says "Treasure Find"; the actual card name is **Treasured Find**)

### Oracle text (verified from `client/public/card-data.json`)

```
Return target card from your graveyard to your hand. Exile Treasured Find.
```

### Reported behavior

> Not working how it should, puts treasure find in graveyards and exiles target card

(Discord thread: Treasure Find, 2026-05-10T21:09:10Z)

### Parser status (verified)

```
abilities[0]: Spell {
  effect: Bounce {
    target: Typed[Card, controller: You, InZone: Graveyard],
    destination: null,    // <-- suspicious: should be Hand
  },
  sub_ability: Spell {
    effect: ChangeZone { destination: Exile, target: SelfRef }
  },
}
```

Two related concerns:
1. The first effect uses `Effect::Bounce` with `destination: null`. The Oracle text says "to your hand" — the destination should be `Hand` explicitly, or the runtime resolution for `Bounce { destination: null }` may be routing to the wrong zone (the user reports the target gets exiled instead of returned to hand).
2. The sub-ability correctly says `ChangeZone { destination: Exile, target: SelfRef }`, but the user reports Treasured Find ends up in the graveyard. So either (a) the `SelfRef` target binding is wrong at runtime and the exile is applied to the user's chosen graveyard card instead of Treasured Find itself, or (b) Treasured Find as a sorcery goes to the graveyard normally and the exile sub-ability never fires, leaving it in the graveyard.

If (b), then the user's "exiles target card" symptom is caused by the `Bounce { destination: null }` falling through to exile semantics rather than hand semantics — i.e., the parser's null destination is being treated as exile by the runtime resolver.

### Class scope

Cards with "Return target X to your hand. Exile this card." structure where the second sentence is a sorcery-specific self-exile. Same arm likely affects similar "rebuy with self-exile" sorceries (Increasing Ambition, Snapcaster-shaped flashback, etc., depending on how their parsing differs).

### Source

Discord thread: "Treasure Find" (2026-05-10T21:09:10Z)

matthewevans:
[...]

PR #4526, opened 2026-06-28
fix(engine): The Incredible Hulk — gate Enrage riders and untap him when attacking (CR 608.2c)

lgray:

## The Incredible Hulk (MSH) — Enrage chain

Oracle: *"Reach, trample / Enrage — Whenever The Incredible Hulk is dealt damage, put a +1/+1 counter on him. If he's attacking, untap him and there is an additional combat phase after this phase."*

Two distinct fixes, both required for the card to resolve correctly.

### 1. Condition bridge (CR 608.2c)
The "If he's attacking" rider must **gate** the untap + additional-combat-phase sub-effects while leaving the +1/+1 counter ungated. `static_condition_to_ability_condition` now bridges `StaticCondition::SourceIsAttacking` / `SourceIsBlocking` / `SourceAttackingAlone` to `AbilityCondition::SourceMatchesFilter`, so the chained riders carry the gate (mirrors the existing `SourceIsSaddled` precedent).

### 2. Chained "untap him" anaphor (CR 608.2c / CR 608.2b)
"untap him" lowered to `SetTapState{ParentTarget}`, but the head `PutCounter{SelfRef}` is never materialized into `ability.targets` (issue #323), so the untap resolved against an empty target list — a silent no-op. Fixed by resolving the anaphor at **lowering time**: `patch_self_ref_head_tap_anaphor` (inside `lower_effect_chain_ir`, the single chain chokepoint that covers trigger / activated / spell entry points) rewrites the chained `SetTapState{ParentTarget, Single}` to `SelfRef` when the antecedent head's subject is `SelfRef`. It reuses the existing `definition_targets_self_source` and mirrors the sibling helper `rewrite_else_parent_target_to_self_ref` (the else-branch analogue).

**Builds for the class, not the card.** A head with a real or *optional* target (Tyvar Kell / A-Tyvar Kell / Nissa, Who Shakes the World: *"...up to one target X. Untap it."*) is **not** `SelfRef`, so its anaphor stays `ParentTarget`: it binds the chosen target, and a **declined** optional target leaves the slot empty so the sub correctly does nothing (CR 608.2b — the anaphor has no referent). An earlier resolver-level `ability.targets.is_empty()` fallback was rejected in review because it wrongly untapped the *source* planeswalker (and fired a spurious untap event) on a declined target; the lowering-time approach discriminates correctly because the head's subject filter is still visible there.

## Tests
- **Building-block** (`lower.rs`): `SelfRef` head → rewrite to `SelfRef`; `Typed` head → stays `ParentTarget`; `All`-scope guard.
- **Resolver fences** (`tap_untap.rs`): declined-optional no-op (CR 608.2b), narrowness fence, `TriggeringSource` class fence.
- **Integration** (`incredible_hulk_enrage_attacking.rs`): Hulk untaps when attacking (gated rider fires), +1/+1 counter ungated, extra combat phase scheduled; chain-shape `SelfRef` assertion. Discrimination verified by call-site/function-body disable.

## Verification
`cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` = 0 · `cargo test -p engine` green (lib 13878/0).

### DIFF
--- a/crates/engine/src/game/effects/tap_untap.rs
+++ b/crates/engine/src/game/effects/tap_untap.rs
@@ -15,6 +15,16 @@ use crate::types::zones::Zone;
 ///
 /// - `SelfRef` → the source object — the printed-name "tap ~"/"untap ~"
 ///   anaphor that always refers to the source regardless of `ability.targets`.
+///   A chained "untap him"/"untap it" anaphor after a `SelfRef`-subject head
+///   (The Incredible Hulk: "put a +1/+1 counter on him ... untap him") is
+///   rewritten from `ParentTarget` to `SelfRef` at parse time by
+///   `sequence::patch_self_ref_head_tap_anaphor`, so it arrives here as
+///   `SelfRef` and binds the source. Resolving the anaphor at parse time (where
+///   the head's subject is still visible) is mandatory: by resolution time the
+///   discriminator is erased — a declined "up to one target" anaphor (Tyvar
+///   Kell) reaches the `_` arm with the SAME empty `ability.targets`, and per
+///   CR 608.2b that case must do nothing, so a blanket source fallback here
+///   would wrongly untap the source.
 /// - `TrackedSet` → the chain's tracked object set published by a preceding
 ///   effect (e.g. `ChooseObjectsIntoTrackedSet`'s "untap those creatures"
 ///   tail). The `TrackedSetId(0)` sentinel binds to the highest tracked-set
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**[MED]** The self-reference rewrite is lost across intermediate chain nodes. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:171` recurses into `sub` before applying the parent `head_is_self_ref` rewrite, then only rewrites that immediate child if it is `SetTapState`; if the chain is `SelfRef` head -> unrelated effect -> `SetTapState { ParentTarget, scope: Single }`, the recursive call sees the intermediate node as the new head and the tap/untap anaphor remains `ParentTarget`. The added tests at `crates/engine/src/parser/oracle_effect/lower.rs:223` only cover the immediate-child shape. Why it matters: the PR claims a chained tap/untap anaphor building block, but a normal English continuation with an intervening instruction would silently no-op at runtime for the same self-referential antecedent class. Suggested fix: carry the active self-referential antecedent through the sub-ability chain until an instruction establishes a different target, and add both an intermediate-positive fixture and a target-head/optional-target negative fixture.

lgray:

Good catch — you're right that the immediate-child-only patch silently no-op'd a `SelfRef` head → intervening instruction → `SetTapState{ParentTarget, Single}` chain. Fixed in the new commit by carrying the active source antecedent **down** the `sub_ability` chain (`patch_self_ref_head_tap_anaphor` now uses an inner `walk(def, carried_self_ref)`):

- A `SelfRef`-subject effect **sets** the antecedent.
- A non-`SelfRef` **object** target **resets** it (a new permanent antecedent — "destroy target creature. Untap it." binds the creature, not the source).
- **Player-directed / targetless** effects (life gain, extra combat phase, draw) **carry** it through unchanged — these introduce no new permanent referent.

Object-vs-player is decided by a conservative, **non-exhaustive** `target_filter_is_player_scoped` allow-list: any unlisted filter defaults to *reset*, which only ever leaves the anaphor as `ParentTarget` (the pre-fix behavior) and **never** causes a wrong-object untap. So an omission is safe by construction; only unambiguously player-referencing variants are listed.

One deliberate deviation from the bot's suggested snippet: a bare `target_filter().is_some()` reset would have wrongly reset on the very `GainLife` / `AdditionalPhase` intermediates cited as the motivating carry-cases — both expose a *player* target slot (`GainLife.player`, `AdditionalPhase.target`), so `is_some()` is `true` for them. The object/player distinction is what actually makes those carry.

Fixtures added (both requested):
- `self_ref_head_intermediate_player_effect_still_rewrites` — SelfRef head → "you gain 2 life" → untap → **rewritten to SelfRef**.
- `self_ref_head_intermediate_object_target_does_not_rewrite` — SelfRef head → object-target effect → untap → **stays ParentTarget**.

Both are revert-proven (run, confirmed red-on-revert, restored byte-identical): breaking the propagation fails the player fixture; breaking the object-reset fails the object fixture. Full `cargo test -p engine` green, clippy `-D warnings` = 0, rebased onto current `main`.

matthewevans (approved):
Reviewed current head a0f5ceda. The previous chained-anaphor blocker is addressed: `patch_self_ref_head_tap_anaphor` now carries the active SelfRef antecedent through player-directed/targetless middle effects and resets on a new object antecedent, with direct regression coverage for the intermediate-effect case and the optional-target no-op cases. Exact-head CI is green, the local merge-current check against origin/main was clean, and the stale Gemini high thread is resolved against the new head.

PR #4533, opened 2026-06-28
feat(engine): Baron Helmut Zemo (MSH) — Boast aggregate-exile cost + copy/cast up to three (CR 702.142a)

lgray:

## Baron Helmut Zemo (MSH) — Boast

Oracle: *"Whenever you cast a black spell from your hand, Baron Helmut Zemo connives. / Boast — Exile any number of black cards from your graveyard with fifteen or more black mana symbols among their mana costs: Copy those exiled cards. You may cast up to three of the copies without paying their mana costs."*

The Connive trigger already parsed correctly. This PR implements the **Boast** ability — a cost + effect that needed four gaps closed, all as parameterized building blocks (no card-specific string matching).

### New `AbilityCost::ExileWithAggregate`
A standalone aggregate-exile cost sibling (modeled on `CollectEvidence`, **not** an `Exile` parameterization — avoids churning ~132 call sites). Zemo = `Sum` of `ManaSymbolCount(Black) >= 15` over black graveyard cards (hybrid pips count per color, **CR 107.4e**). Interactive payment via new `PayCostKind::ExileAggregate` (payability = exile-all max for `Sum`/`GE`; the choice handler validates the ≥15 threshold + AI legal-action producer). **CR 117.1 / 118.3 / 601.2b / 601.2h.** The shared `aggregate_property_over` summation is the single authority (`Aggregate` + `TrackedSetAggregate` resolvers delegate to it).

### "Copy those exiled cards" — cost-published tracked set surviving activation→resolution
The cost-paid exile publishes a fresh tracked set whose **concrete id** is bound into the effect chain's `TrackedSet(0)` sentinel **before** the ability is pushed to the stack. This is required because the `chain_tracked_set_id`/latest-id sentinel path is reset at depth-0 resolution and would not survive an intervening tracked set published between activation and resolution. **CR 603.7.** (A revert probe confirms: removing the concrete-id rewrite makes the effect bind a decoy set, proving the rewrite is load-bearing.)

### "Cast up to three of the copies"
`Effect::CastCopyOfCard` gains `count: Option<QuantityExpr>` (`#[serde(default)]` = up-to-all, preserving the 13 existing cards + on-disk JSON). Zemo caps at `min(3, N)` via the existing `up_to` `ChooseFromZoneChoice` (you may cast fewer/zero; uncast copies cease to exist). **CR 707.12 / 707.12a / 707.10a.**

### Final parse
cost `ExileWithAggregate{Sum, ManaSymbolCount(Black), GE, 15, Typed(Card,You,[HasColor:Black,InZone:Graveyard]), Graveyard}`; effect `CastCopyOfCard{TrackedSet(0), zero, Some(Fixed 3)}`; **zero parse_warnings**.

## Tests
`crates/engine/tests/baron_helmut_zemo_boast.rs` (7 integration) + `aggregate_property_over` unit. Revert probes confirm each fix is load-bearing: the ≥15 threshold (15 payable / 14 unpayable / 14-subset rejected), the tracked-set concrete-id binding (vs a decoy set), and the cast cap (`min(3,N)` vs all). Parser is nom-combinator only; every new variant goes through an exhaustive no-wildcard match (including the `fold_cost` drift gate).

## Verification
`cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` = 0 · `cargo test -p engine` 13982 passed / 0 failed · all 21 CR annotations grep-verified.

### DIFF
--- a/crates/engine/src/ai_support/candidates.rs
+++ b/crates/engine/src/ai_support/candidates.rs
@@ -1622,6 +1622,43 @@ pub fn candidate_actions_broad(state: &GameState) -> Vec<CandidateAction> {
             crate::game::casting_costs::tap_creature_power_contribution,
             |cards| GameAction::SelectCards { cards },
         ),
+        // CR 117.1 + CR 601.2b: Aggregate-threshold "exile any number" cost
+        // (Baron Helmut Zemo's Boast). The threshold is satisfied by ANY chosen
+        // subset whose summed `property` meets the comparator, so enumerate
+        // minimal-cover subsets (mirroring the Crew/Saddle aggregate-tap path)
+        // rather than a fixed-cardinality selection. `minimal_power_subset_candidates`
+        // is sum-based, so this arm handles the `Sum` aggregate (the only shape in
+        // the corpus); other aggregate functions fall through to the generic arm.
+        WaitingFor::PayCost {
+            player,
+            kind:
+                PayCostKind::ExileAggregate {
+                    function: crate::types::ability::AggregateFunction::Sum,
+                    property,
+                    comparator,
+                    value,
+                    ..
+                },
+            choices,
+            ..
+        } => {
+            let property = *property;
+            minimal_power_subset_candidates(
+                state,
+                *player,
+                choices,
+                |total| comparator.evaluate(total, *value),
+                move |state, id| {
+                    crate::game::quantity::aggregate_property_over(
+                        state,
+                        &[id],
+                        crate::types::ability::AggregateFunction::Sum,
+                        property,
+                    )
+                },
+                |cards| GameAction::SelectCards { cards },
+            )
+        }
         WaitingFor::PayCost {
             player,
             choices,
--- a/crates/engine/src/analysis/ability_graph.rs
+++ b/crates/engine/src/analysis/ability_graph.rs
@@ -1114,6 +1114,10 @@ fn fold_cost(acc: &mut NodeAcc, cost: &AbilityCost) {
         | AbilityCost::Exile { .. }
         | AbilityCost::ExileMaterials { .. }
         | AbilityCost::CollectEvidence { .. }
+        // CR 118.3: an aggregate graveyard-exile cost (Baron Helmut Zemo's Boast)
+        // is a structural exile cost like CollectEvidence — it consumes/produces no
+        // resource axis the loop detector models.
+        | AbilityCost::ExileWithAggregate { .. }
         | AbilityCost::PaySpeed { .. }
         | AbilityCost::ReturnToHand { .. }
         | AbilityCost::Unattach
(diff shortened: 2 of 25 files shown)


matthewevans (changes requested):
Thanks for the follow-up. I reviewed current head `20407114cbed4c13d248d4b001b0f79a54702858`; CI is green, but I think this still needs one architecture fix before maintainer approval.

**[MED] Keep `ExileAggregate` payment valuation parameterized by `zone`.** The engine model makes the new aggregate-exile cost reusable across zones: `AbilityCost::ExileWithAggregate` carries `zone: Zone` in `crates/engine/src/types/ability.rs`, and the interactive `PayCostKind::ExileAggregate` carries the same zone in `crates/engine/src/types/game_state.rs`. But the AI payment policy ignores that field and always scores the selection as a graveyard exile in `crates/phase-ai/src/policies/payment_selection.rs`. That means a hand-based or battlefield-based aggregate exile cost would be advertised through the generic engine type but undervalued as cheap graveyard fuel. Please mirror the existing zone-aware branches (`ExileFromZone` / `ExileFromManaZone`) for `PayCostKind::ExileAggregate { zone, .. }`, and add coverage proving a hand-zone aggregate payment is valued like hand cards rather than graveyard cards.

I did not treat the `HashSet` uniqueness check as blocking; it is small and correctness-oriented. The zone mismatch is the one maintainer-bar issue because it breaks the generality of the new API surface.

lgray:

Fixed in the latest push (rebased onto current `main`).

**[MED] Zone-aware `ExileAggregate` valuation.** `payment_cost` now matches on `PayCostKind::ExileAggregate { zone, .. }` and mirrors the `ExileFromManaZone` branches: `Battlefield → permanent_value`, `Hand → card_value*1.2`, `Graveyard → 0.1 + card_value*0.2`, else `card_value*0.5`. So a hand- or battlefield-sourced aggregate exile (the general API surface) is valued by its real cost instead of as cheap graveyard fuel. Baron Helmut Zemo's Boast exiles from the graveyard, so its valuation — and thus the AI gate — is unchanged.

Added `exile_aggregate_cost_values_by_source_zone`, which is discriminating: the pre-fix zone-agnostic arm scored a hand exile and a graveyard exile equally, so the `assert!(graveyard_score > hand_score)` flips red under a revert.

On the `HashSet` uniqueness check (`casting_costs.rs:2169`) — left as-is per your note that it's non-blocking; happy to switch it to an O(N²) scan over the small (≤15) slice if you'd prefer.

`cargo test -p phase-ai` 1107/0, `cargo test -p engine` green, clippy `-D warnings` 0.

matthewevans (approved):
Approved current head `c79d1af139c1d9e2883d995fcf16a9a3f902cd77` after re-review. The prior `ExileAggregate` AI zone-valuation blocker is fixed, the Gemini duplicate-check allocation comment was addressed in the maintainer fixup, exact-head CI is green, and the backend payment/copy-cast path has discriminating scenario coverage.

PR #4536, opened 2026-06-28
fix(parser): hoist "if it's an opponent's turn" intervening-if condition

minion1227:
## Summary

Adds a parser arm so the intervening-if condition **"if it's an opponent's turn"**
is recognized and hoisted onto a trigger's `condition`, instead of being silently
dropped. This is a general parser building block covering the **class** of cards
that gate a triggered ability on the active player being an opponent — mirroring
the existing `"it's not your turn"` arm.

## The bug

**Discordant Spirit** reads:

> At the beginning of each end step, if it's an opponent's turn, put a +1/+1
> counter on this creature for each 1 damage dealt to you this turn.

The released parse dropped the intervening-if entirely (`"condition": null`), so the
counter would be placed on **every** end step — including the controller's own turn,
where the ability should do nothing. The condition was unparsed because
`parse_turn_conditions` had no arm for the "an opponent's turn" phrasing.

## The fix

Add an `alt` arm in `parse_turn_conditions` (`parser/oracle_nom/condition.rs`)
mapping both apostrophe forms to `StaticCondition::Not { DuringYourTurn }`:

- `"it's an opponent's turn"`
- `"it is an opponent's turn"`

**CR 102.1 + CR 102.2:** each turn has exactly one active player, so "it's an
opponent's turn" is logically identical to "it's not your turn" — the active player
is any non-controller. Representing it as `Not(DuringYourTurn)` reuses the existing
negation pathway rather than introducing a new condition variant.

No trigger-bridge changes were needed: `static_condition_to_trigger_condition`
(`parser/oracle_trigger.rs`) already lowers `Not(DuringYourTurn)` →
`Not(DuringPlayersTurn { player: Controller })`. The intervening-if now hoists
correctly through the existing path.

## Building block, not one card

The condition arm handles **any** card with an "it's an opponent's turn"
intervening-if, not just Discordant Spirit, which is simply the motivating example.

## Scope note (intentional)

Discordant Spirit's separate `for each 1 damage dealt to you this turn` dynamic
count is a **pre-existing, independent parser gap** — `parse_damage_dealt_this_turn_ref`
(`parser/oracle_nom/quantity.rs`) only has `"...to your opponents"` and
`"...to target opponent"` arms, with no `"...to you"` (controller-targeted) arm, so
the count degrades to `Fixed 1`. That fix additionally requires runtime resolution
of `DamageDealtThisTurn` targeted at the controller, which is a larger blast radius.
It is **deliberately out of scope** here to keep this change surgical and low-risk.

## Tests

- `test_parse_condition_opponents_turn` — both apostrophe forms parse with the
  correct remainder.
- `trigger_intervening_if_opponents_turn_discordant_spirit` — the full trigger line
  hoists `Not(DuringPlayersTurn { Controller })`.
- 343 condition + 825 trigger parser tests pass; `cargo fmt --all` clean;
  `scripts/check-parser-combinators.sh` passes (exit 0).

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -640,6 +640,18 @@ fn parse_turn_conditions(input: &str) -> OracleResult<'_, StaticCondition> {
         map(tag("it's not your turn"), |_| StaticCondition::Not {
             condition: Box::new(StaticCondition::DuringYourTurn),
         }),
+        // CR 102.1 + CR 102.2: there is always exactly one active player, so
+        // "it's an opponent's turn" is exactly "it's not your turn" — the active
+        // player is any non-controller. Maps to the same Not(DuringYourTurn).
+        map(
+            alt((
+                tag("it's an opponent's turn"),
+                tag("it is an opponent's turn"),
+            )),
+            |_| StaticCondition::Not {
+                condition: Box::new(StaticCondition::DuringYourTurn),
+            },
+        ),
         parse_day_night_condition,
     ))
     .parse(input)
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved current head `e44320a0f86c205bcffb5944f9e64e7147d6f5b0` after review.

Evidence checked: the parser arm is at the existing turn-condition seam, composes the straight/curly apostrophe surface forms with `alt` + `tag` instead of full-string proliferation, and reuses the existing `Not(DuringYourTurn)` bridge; the trigger test discriminates the intervening-if hoist. Gemini's two comments are addressed at head even though the threads remain visually unresolved. Exact-head CI is green and `scripts/check-parser-combinators.sh` passes locally.

The Discordant Spirit damage-count phrase remains a separate pre-existing quantity-parser gap; this PR is scoped to the condition field and the CI parse diff confirms only that field changes.

PR #4549, opened 2026-06-28
fix(parser): UNSUPPORTED one-off: Fractured Identity

ntindle:
## Summary

**Root cause:** UNSUPPORTED one-off: Fractured Identity

## Cards corrected
- Fractured Identity

## Fix
test

## Files changed
- a
- b

## CR references
- c

## Verification
- `cargo fmt --all` — pass
- `./scripts/check-parser-combinators.sh <upstream/main merge-base>` — pass
- `cargo run --profile tool --features cli --bin oracle-gen -- data --filter "fractured identity"` — pass
Cards confirmed re-parsed correctly: Fractured Identity


matthewevans (changes requested):
Requesting changes for one correctness blocker.

**[HIGH] Fractured Identity is marked handled while the PR leaves a known controller-vs-owner case broken.** Evidence: `crates/engine/src/game/effects/token_copy.rs:3676` documents that "its controller" must use the exiled permanent's last-known battlefield controller, but the regression at `token_copy.rs:3695` is `#[ignore]`; `crates/engine/src/game/coverage.rs:6493` simultaneously marks `PlayerFilter::AllExcept` as `Handled`. The underlying helper also prefers the live object controller before LKI at `crates/engine/src/game/ability_utils.rs:467`, which the ignored test says resets to owner after exile. Why it matters: Fractured Identity is wrong whenever the exiled permanent is controlled by someone other than its owner, and coverage will report the card/class as supported anyway. Suggested fix: make `parent_target_controller` prefer the LKI controller for object targets that have left the battlefield, un-ignore the controller-not-owner regression, and only mark `AllExcept` handled once that passes.

matthewevans (approved):
Reviewed the updated head after merging current `main`. Two passes found no blocking issue in the implementation. The remaining note is non-blocking: `PlayerFilter::AllExcept { ParentObjectTargetController }` is currently safe through the `player_scope` driver path used here, but future reuse outside ability-aware paths should be handled carefully.

PR #4553, opened 2026-06-28
fix(parser): bind reflexive conditional riders (S01 reflexive-if-rider cluster)

lgray:

### The bug (correctness, not just coverage)
The reflexive conditional rider on removal/bounce/damage spells and ETB triggers — e.g. *"Exile target creature. **If it was dealt damage this turn**, create a Clue token"* — parsed with the rider effect attached as a sub-ability but its **condition dropped to `null`**, so the rider fired **unconditionally**. Verified against `card-data.json` for Sold Out, Consuming Ashes, Brackish Blunder, and others.

### The fix (one building block)
- A parameterized **reflexive-anaphor condition recognizer** (subject × tense/polarity × `FilterProp` predicate), registered in `try_nom_condition_as_ability_condition` — the single dispatch point used by both the `parse_clause_ast` swallow path and the chunk loop. nom-combinator-only; registered after the type-membership arm and before the fallbacks so existing conditions are untouched. No new enum variants.
- A small engine fix: the `WasDealtDamageThisTurn` LKI-snapshot arm now queries the turn-scoped damage ledger by `ObjectId` (CR 120.9), mirroring the live arm, so the predicate evaluates correctly after the target changes zones.

### Evidence
- **Flips 11 cards** false→true (Consuming Ashes, Sold Out, Driftgloom Coyote, Wisecrack + 7 class members incl. serum snare, sewers of estark, battlefield improvisation, sweep away, will of the all-hunter, grave choice) — the building block working across the class.
- **0 coverage regressions** over the full 35,396-card corpus: swallows −11 (Condition_If −9, Duration_ThisTurn −2), **every other detector +0**, no card newly swallowed. Independently reproduced by an isolated-toggle diff (exactly 11 cards changed, all `condition:null → correct`).
- Gates: `cargo fmt`, `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings`, `cargo test -p engine` (11 new parser unit tests + 8 runtime card-tests, each with positive+negative pairs through the real cast/ETB pipeline and load-bearing revert-probes), parser-combinator gate, `semantic-audit` — all clean. CR 120.6/120.9/202.3/208.1/400.7/508.1b/509.1a/608.2c/608.2h grep-verified. Rebased onto current `main` with the full workspace build re-run.
- A guard test confirms board-state conditionals (*"If you control a creature, …"*) are **not** mis-folded — the Isochron/Zemo coverage-regression class.

### Honest scope notes (documented follow-ups, no regression)
- **Brackish Blunder ("was tapped") deliberately not shipped:** `FilterProp::Tapped` fails closed on the LKI snapshot (no tap state in `LKISnapshot`/`ZoneChangeRecord`); recognizing it would flip the swallow green while the runtime silently under-fires. Kept as an honest red swallow; the LKI tap-state plumbing is its own follow-up cluster.
- **Faller's Faithful Part B:** the recognizer parses its `Not{WasDealtDamageThisTurn}` rider; the declined-"up to one" optional-target guard is deferred. Net improvement over baseline (chosen-target path now correct; declined path unchanged) — no regression.

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -4200,6 +4200,17 @@ fn zone_change_record_matches_property(
             let actual = counter_count_from_map(&lki.counters, selector);
             comparator.evaluate(actual, resolve_filter_threshold(state, count, source))
         }),
+        // CR 120.6 + CR 120.9 + CR 608.2h: "was dealt damage this turn" is a
+        // turn-scoped historical fact, not a query against current marked damage.
+        // The `damage_dealt_this_turn` ledger is keyed by the battlefield ObjectId
+        // and survives the object's zone change, so a look-back rider ("Exile
+        // target creature. If it was dealt damage this turn, ..." — Sold Out)
+        // evaluating against the LKI snapshot reads the same ledger the live
+        // evaluator uses, by the snapshot's `object_id`. Mirrors the live arm.
+        FilterProp::WasDealtDamageThisTurn => state
+            .damage_dealt_this_turn
+            .iter()
+            .any(|r| matches!(r.target, TargetRef::Object(id) if id == record.object_id)),
         FilterProp::Tapped
         | FilterProp::IsSaddled
         | FilterProp::SaddledSource
@@ -4244,7 +4255,6 @@ fn zone_change_record_matches_property(
         | FilterProp::DifferentNameFrom { .. }
         | FilterProp::InAnyZone { .. }
         | FilterProp::SharesQuality { .. }
-        | FilterProp::WasDealtDamageThisTurn
         | FilterProp::EnteredThisTurn
         | FilterProp::ZoneChangedThisTurn { .. }
         // CR 122.6: counters-put-this-turn is a live-history join keyed on the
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed exact head `97d1949e1a56cbd8667b2c57879e2b1b669b3806` for the reflexive conditional rider fix.

The stale Gemini CR-comment thread is addressed on this head: `parse_or_threshold` now has the `CR 202.3 + CR 208.1` prefix. The implementation is at the right parser seam (`try_nom_condition_as_ability_condition` / target-anaphoric condition parsing), uses composed `nom` axes rather than full-string tags, and preserves coverage honesty by deliberately leaving unsupported past-tense tap-state riders unrecognized.

Behavioral trace: past-tense LKI forms route through `TargetMatchesFilter { use_lki: true }`; present-tense forms remain live; zone-change filtering already handles mana value and P/T snapshots, and this PR adds the missing damage-ledger lookup for `WasDealtDamageThisTurn`. Runtime tests drive real cast/ETB paths with positive and negative discriminators for Sold Out, Consuming Ashes, Wisecrack, and Driftgloom Coyote. Local `cargo fmt --all --check` and exact-head GitHub CI are green.

PR #4551, opened 2026-06-28
feat(engine): static cost reduction for plot/unlock + dynamic activated-ability cost (S21)

lgray:
:robot: _AI text below_ :robot:

## S21 static cost-reduction — Doc Aurlock, Inquisitive Glimmer, Agatha

First pass of the `S21-static-ability` cluster: two cost-reduction static building blocks, each built for the class.

### B2 — special-action cost reduction (plot / unlock)
- New `StaticMode::ReduceActionCost { action, mode, amount }` + `SpecialAction::Plot`.
- A single-authority `apply_special_action_cost_reduction` called from **both** the plot payment site (`casting.rs`) and the unlock-door site (`engine.rs`) — no inlined duplication.
- Generic-only reduction (CR 118.7a), floors at 0.
- Covers **Doc Aurlock, Grizzled Genius** ("Plotting cards from your hand cost {2} less") and **Inquisitive Glimmer** ("Unlock costs you pay cost {1} less"). The parser accepts both the `costs`/`cost` verb forms.
- Plot is kept on a dedicated `SpecialAction` axis rather than routed through `ReduceAbilityCost{keyword:"activated"}`, which would wrongly reduce plot via Zirda/Training-Grounds-style statics.

### B1 — dynamic activated-ability cost reduction
- Extends the existing `ReduceAbilityCost` path to honor a `dynamic_count` (`QuantityRef::Power{Source}`), so **Agatha of the Vile Cauldron** ("Activated abilities of creatures you control cost {X} less to activate, where X is Agatha's power") reads the source's post-layer power with a one-mana floor.
- Existing static reductions (`dynamic_count: None`) are behaviorally **unchanged** (multiplier 1).

### Evidence
- Coverage: **+3 supported** (Doc Aurlock, Inquisitive Glimmer, Agatha), **0 regressions** (no card supported→unsupported, no `gap_count` rose anywhere) over the full 35,396-card corpus.
- Gates: `cargo fmt` clean; `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings` clean; `cargo test -p engine` green (9 new discriminating tests with revert-probes — plot drops 2/floor 0, unlock drops 1, Agatha's 3-way power split incl. floor); parser-combinator gate clean; `semantic-audit` clean. All CR numbers verified against the Comprehensive Rules.
[...]

PR #4557, opened 2026-06-28
feat(engine): Nowhere to Run — scoped hexproof-bypass + ward-suppression statics

lgray:

**S21 cluster (static-ability), follow-up to #4551.** Implements Nowhere to Run's static line as two continuous effects parsed from one line (the "those creatures" anaphor reuses sentence 1's subject):

> Creatures your opponents control can be the targets of spells and abilities as though they didn't have hexproof. Ward abilities of those creatures don't trigger.

### Hexproof bypass — object-scoped, multiplayer-correct (CR 702.11b / 702.11e)
- New `StaticMode::IgnoreHexproof` carries an object `affected` filter. `target_ignores_hexproof` keys **solely** on the would-be target matching that filter (evaluated from the static's source via `FilterContext::from_source`), **independent of the targeting player** — the card has no "you control" qualifier, so the bypass applies for any player, scoped to the static controller's opponents' creatures.
- `player_ignores_hexproof` now matches only player-scoped grants (`affected = None`) + the per-player transient (Detection Tower), preserving existing behavior. The only battlefield `IgnoreHexproof` emitter is this new code; Detection Tower uses the unchanged transient path.
- `can_target` remains the single hexproof chokepoint; **Shroud is never bypassed** (checked before the bypass).

### Ward suppression — continuous effect, ward-only (CR 702.21a + CR 611.3 + CR 613.11)
- New `SuppressedTriggerEvent::BecomesTargeted` (serde-additive leaf, exhaustive `Display`), reusing the existing `StaticMode::SuppressTriggers`.
- "Ward abilities of those creatures don't trigger" turns the ward triggered ability off at the **inline ward-trigger gate only** — the becomes-target event still occurs (this is a continuous effect, **not** a CR 603.2g prevention), so non-ward becomes-target triggers on the same permanent are unaffected.

### Parser
`parse_ignore_hexproof_static` is nom-combinator throughout, fails closed (defers on partial subject / `TargetFilter::Any` / trailing prose), and is positioned before multi-sentence splitting so the anaphor keeps its antecedent. Subject extraction via `parse_type_phrase` → class-level.

### Tests
- End-to-end parse test (with the `Flash` keyword) asserts **zero Unimplemented** + both statics + filter reuse (replaces a stale deferral test that passed only via an empty-keywords artifact).
- Runtime discriminators in `game::targeting` (a third player can target the static-controller's opponent's hexproof creature; the static-controller's own creature stays protected) and `game::triggers` (ward suppressed for the opponent's creature, still fires for the controller's own) — both with measured revert-probes.

### Verification (rebased onto current `main`)
`cargo fmt --all --check` = 0; `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings` = 0; `cargo test -p engine` = 0; full `cargo nextest` (workspace) = 17,753 passed / 0 failed. Full-corpus `cargo coverage`: supported +1 (Nowhere to Run gap→0), 0 regressions.


matthewevans (changes requested):
Requesting changes for one live backend blocker.

**[HIGH] Object-scoped IgnoreHexproof bypass ignores whether the static is functioning and whether its condition matches.** `target_ignores_hexproof` iterates `battlefield_active_statics` and accepts any matching `affected` filter, while the neighboring player-scoped helper uses `game_functioning_statics` plus `static_condition_matches_context` before granting the same bypass. That means a conditional or currently non-functioning object-scoped static can still make a target legal as though it had no hexproof. Please make `target_ignores_hexproof` use the same functioning/condition gate with a `StaticCheckContext { target_id: Some(target_id), .. }`, and add a regression covering an inactive or condition-false object-scoped static.

I also checked the other high-signal concern about missing ward suppression. Current head does include the runtime hook and regression in `triggers.rs`, so I am not blocking on that.

lgray:

Both addressed in `5aab028dc` (rebased onto current `main`).

### [HIGH] `target_ignores_hexproof` functioning + condition gate — fixed
`target_ignores_hexproof` now mirrors `player_ignores_hexproof` exactly: it iterates `game_functioning_statics` (so a non-functioning / ability-suppressed / phased-out source grants nothing, plus command-zone scope) and gates each static through `static_condition_matches_context` with `StaticCheckContext { target_id: Some(target_id), .. }`.

One precision note from tracing the old path: `battlefield_active_statics` → `active_static_definitions` *already* applied the CR 604.1/613.1 condition gate, but in **source** context (`recipient = None`). So a source-scoped false condition was already denied; the live defect was a **recipient-referencing** condition being evaluated against `recipient = None` — wrongly **denying** a condition that was true for the target — plus the missing functioning/command-zone parity. The `target_id: Some(target_id)` wiring (your suggested fix) is exactly what corrects it.

Regression: `scoped_ignore_hexproof_respects_recipient_condition_multiplayer` (3-player) — a `RecipientMatchesFilter{land}` condition (false for a creature target) **denies** the bypass; `RecipientMatchesFilter{creature}` (true) **grants** it; `assert_ne!` enforces they differ. Revert-probe (measured): restoring the pre-fix `battlefield_active_statics` body fails the condition-TRUE assertion, so the gate is load-bearing.

Gates (worktree, rebased onto post-merge `main`): `cargo fmt --all --check` = 0; `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -D warnings` = 0; `cargo test -p engine` = 0 (14042 passed).

### Re: the ward-suppression "missing from triggers.rs" note
That is a false positive — the ward-suppression runtime + test are in this PR: `fn ward_becomes_target_suppressed` at `triggers.rs:1067` (called at `:1740`) and the test `ward_suppressed_by_becomes_targeted_suppress_static` at `:12804` (`git diff main...HEAD -- crates/engine/src/game/triggers.rs` = +174/-1). No change needed there.

matthewevans (approved):
Approved current head `5aab028dc`. The prior object-scoped hexproof blocker is addressed with `game_functioning_statics` plus `StaticCheckContext { target_id: Some(target_id) }`; the ward-suppression runtime hook and discriminating tests are present; exact-head CI is green.

PR #4563, opened 2026-06-28
fix(server): release state + connections guards before broadcast_play…

### LINKED ISSUE #4548: [Bug]: Full (server-hosted) multiplayer deadlocks on game creation: broadcast_player_slots re-locks the state mutex the caller already holds
__Note:__ the following issue text is entirely LLM-generated (using Claude Opus 4.8). I noticed this project makes heavy use of LLMs, so I figured it would be fine to post as-is in this particular case without rephrasing in my own words (as I usually do whenever I have an LLM draft an issue text). I verified the claims myself:

---

## Build/version

v0.9.0 (`6bd7432`); also present in v0.8.0.

## Game mode

Online multiplayer, server-hosted (a `phase-server` running in its default full mode, without `--lobby-only` / `PHASE_LOBBY_ONLY`).

## What I was doing

Running a self-hosted `phase-server` in full mode (without `--lobby-only` / `PHASE_LOBBY_ONLY`) so two players can play server-hosted multiplayer games, and creating a normal multiplayer game from the client.

## Actual behavior

Creating a server-hosted multiplayer game wedges the server. Concretely:

1. The host who creates the first game receives a join code, but the game never appears in the other player's lobby list.
2. The next attempt to create a server-hosted game (by either player) hangs; the server never responds and no game is created.

Once this happens the server stays wedged until restarted, and restarting only buys one more create before it wedges again.

## Root cause

This is a self-deadlock on the `state` mutex. `SharedState` is a non-reentrant async mutex:

https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L73

The standard multiplayer create handler takes that lock and holds it for the whole block (there is no `drop` before the end):

https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L3734-L3735

Near the end of that block, while the lock is still held, it calls `broadcast_player_slots`:

https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L3854

And `broadcast_player_slots` re-acquires the same `state` lock:

https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L1581-L1600

Because tokio's `Mutex` is not reentrant, the handler blocks on itself there forever. That single fact explains all three symptoms:

- `GameCreated` is sent just before the deadlocking call, so the host still gets a join code.
- The `LobbyGameAdded` fan-out is on the following lines and is never reached, so the game is never broadcast to subscribers:
  https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L3856-L3865
- The wedged task never releases `state`, so every subsequent handler that needs it (including the next `CreateGameWithSettings`) blocks indefinitely.

The same `broadcast_player_slots(state, ...)` call while holding `state` appears in the join paths too, which look affected for the same reason:

https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L4010

https://github.com/phase-rs/phase/blob/6bd7432c348d6bc32e911b4a7701cb0281009eb6/crates/phase-server/src/main.rs#L4123

A fix would be to drop the `state` guard before calling `broadcast_player_slots`, or have the handler pass the already-gathered slot information to a variant that does not re-lock.

## Steps to reproduce

1. Run the official server image in full mode (note: without `PHASE_LOBBY_ONLY`):

   ```sh
   docker run -d --platform linux/amd64 --name phase-server \
     -p 9374:9374 -e PHASE_CORS_ORIGIN='*' \
     ghcr.io/phase-rs/phase-server:v0.9.0
   ```

2. Open two client sessions and point both at the self-hosted server (`ws://<host>:9374/ws`) via the self-hosted server field.
3. In client A, create a server-hosted multiplayer game (not P2P).
4. Observe that A gets a join code but the room never shows up in client B's lobby list.
5. From client B (or A), create another server-hosted game; it never completes.

The deadlock is deterministic: the first server-hosted game created after a server start is enough to wedge it. I also confirmed it by driving two raw WebSocket connections against a clean server (no reverse proxy and no custom client involved), so it is purely server-side.

## Expected behavior

Creating a server-hosted multiplayer game should complete, broadcast `LobbyGameAdded` to lobby subscribers so the room is visible to the other player, and leave the server able to handle further creates and joins.

## Logs

Server logs for the first create in full mode, lightly trimmed; note that it stops after the lobby registration and never logs the fan-out:

```
INFO  phase_server: server mode selected mode=Full
INFO  ws_message: CreateGameWithSettings display_name=Alice public=true has_password=false timer=None deck_size=60 ai_seats=0 has_peer_id=false
INFO  ws_message: server_core::session: game session created game=G44XHQ player_count=2
INFO  ws_message: game created via lobby game=G44XHQ host=Alice players=2
DEBUG ws_message: lobby_broker::lobby: lobby game registered game=G44XHQ host=Alice
   (no further output for this request; the task is now blocked in broadcast_player_slots)
```

A subsequent server-hosted create is received but never produces a session (blocked on `state.lock()`):

```
INFO  ws_message: CreateGameWithSettings display_name=R0 public=true has_password=false timer=None deck_size=60 ai_seats=0 has_peer_id=false
   (no "game session created" ever follows)
```

## Note

Full mode looks like the untrodden path here; the public deployment and all README examples use `--lobby-only` (where games are peer-to-peer and this handler is not exercised), which is likely why this has not surfaced. Lobby-only mode works fine.

dale053:
## Summary

- `broadcast_player_slots` re-acquires both the `state` and `connections` `tokio::Mutex` guards internally, but the `CreateGameWithSettings` full-mode handler held both guards alive across the `.await` — Tokio's `Mutex` is non-reentrant, so the call deadlocked permanently on first use
- Fix: wrap each `MutexGuard` (`mgr`, `conns`, `lob_guard`) in an explicit `{ }` block so it is unconditionally dropped before any `.await` that could re-acquire the same lock
- Two regression tests in `issue_4548_deadlock_tests` verify that `broadcast_player_slots` completes within 2 s when called with no locks held and when the lobby lock is held (the only other guard that spans the call)

## Root cause

`crates/phase-server/src/main.rs`, `CreateGameWithSettings` standard multiplayer path:

```rust
// OLD — mgr and conns declared here, lived until end of else-block
let mut mgr = state.lock().await;       // lock A
// ...
let mut conns = connections.lock().await; // lock B
// ... lots of async work ...
broadcast_player_slots(state, connections, &game_code).await;
//   ↑ calls state.lock().await (lock A) → permanent deadlock
//   ↑ calls connections.lock().await (lock B) → permanent deadlock
```

Tokio's async state machine keeps `MutexGuard`s in the generated `Future` struct until the end of the lexical scope, even after their last syntactic use — NLL does not help here across `.await` points.

## Test plan

- [ ] `cargo test -p phase-server issue_4548` — both regression tests pass
- [ ] `cargo clippy -p phase-server -- -D warnings` — clean
- [ ] Manual: start server in Full mode, create a multiplayer game, verify host receives `GameCreated` + `PlayerSlotsUpdate` and subsequent game creation attempts succeed

## Anchored on

- `crates/phase-server/src/main.rs:3650` — AI game path already correctly wraps `state.lock()` in a `{ }` block and drops it before messaging the client
- `crates/phase-server/src/main.rs:1508` — disconnect handler wraps `state.lock()` in `{ }` to release before the `broadcast_player_slots` call at line 1515

## Gate A

```
exit: 0
```

(No parser files changed; script found no combinator violations.)

Fixes #4548

Model: claude-sonnet-4-6
Thinking: medium
Tier: Standard


matthewevans (changes requested):
Requesting changes for current-head blockers.

**[HIGH] Stray gitlink/submodule is in the PR.** The diff adds `gittensory` as `new file mode 160000` (`a4429ce8…`) alongside the server fix. That is unrelated scope contamination and cannot ship in a backend deadlock PR. Please remove it from the branch entirely.

**[MED] Regression tests do not exercise the deadlocking production path.** The bug is in `CreateGameWithSettings`: it used to hold `state` and `connections` guards across the later `broadcast_player_slots(...).await`. The new tests only call `broadcast_player_slots` directly after creating a game with no `state`/`connections` guard held, and one test intentionally holds only the lobby lock, which `broadcast_player_slots` does not acquire. Those tests pass even if the handler regresses to holding `mgr`/`conns` across the production broadcast. Please add coverage that drives the actual `CreateGameWithSettings` full-mode handler path (or a factored helper that owns the same lock sequence) and would fail/timeout with the old guard lifetime.

**[MED] Avoid cloning the full public-games list for the newly registered game.** At `crates/phase-server/src/main.rs:3844`, the PR calls `lob.public_games()` and scans the cloned list only to find `game_code`. The lobby already exposes `public_game(&game_code)`, which clones just that entry and keeps the lock-scope fix intact.

matthewevans:
Thanks for looking into this. The core diagnosis here is directionally right: the full-mode `CreateGameWithSettings` path was holding async mutex guards across a later `broadcast_player_slots(...).await`, and that broadcast re-acquires the same locks.

That said, this PR is not in a mergeable state and I am going to supersede it with #4570 rather than build on this branch.

Current-head blockers:

- The branch adds an unrelated `gittensory` gitlink/submodule. That is scope contamination and cannot ship in a server deadlock fix.
- The regression tests do not exercise the production path that deadlocked. Calling `broadcast_player_slots` directly with no `state`/`connections` guard held would pass even if `CreateGameWithSettings` regressed back to holding those guards across the broadcast.
- The implementation missed an existing lobby API and cloned/scanned the full public-games list instead of using `public_game(&game_code)`.

For changes touching server, infrastructure, concurrency, persistence, auth, or other architecture-sensitive code, please use Frontier-tier models with an explicit review loop before submitting. Green CI and a plausible diagnosis are not enough for this class of change; the review needs to catch scope contamination, production-path test discrimination, and existing API reuse before the PR is opened.

matthewevans (changes requested):
Current head `ee8e9cd7fbdca1b6969b7d6dbd184aae8880d0f9` is still not mergeable.

The PR still adds the unrelated root-level `gittensory` gitlink/subproject entry. That is scope contamination and cannot ship in a server deadlock fix. The branch also remains superseded by the cleaner follow-up path already discussed, so I am keeping this blocked rather than approving/enqueueing it.

matthewevans (changes requested):
Reviewed current head `b5698bc1d777387d66663c51eded00d7247b3529`.

**[MED] Regression tests still do not exercise the deadlocking production path.** Evidence: `crates/phase-server/src/main.rs:6795` and `crates/phase-server/src/main.rs:6817` call `broadcast_player_slots` directly, while the fixed production behavior is the `CreateGameWithSettings` full-mode path reaching `broadcast_player_slots` at `crates/phase-server/src/main.rs:3897` after releasing the `state` and `connections` guards. Why it matters: these tests would still pass if the handler regressed to holding those guards across the production broadcast, because they never drive the handler or the same lock sequence. Suggested fix: add a handler-level async regression for `CreateGameWithSettings` full multiplayer creation, or factor the lock sequence into a helper that the handler and test both call, so the test times out/fails on the old guard lifetime.

matthewevans (changes requested):
Reviewed current head `5fe0d3bdd4463ff8fdd2262b0e8a1fb62cae4f16`.

The lock-sequence regression is now materially better: `create_and_connect_multiplayer_session` is the same helper the `CreateGameWithSettings` full-mode handler uses for the state/connections phases, and the new async test drives that helper before calling `broadcast_player_slots`, so the earlier production-path coverage concern is addressed.

**[MED]** The PR still clones and scans the full public-games list for a single newly registered game. Evidence: `crates/phase-server/src/main.rs:3962` calls `lob.public_games().into_iter().find(...)`, while `crates/lobby-broker/src/lobby.rs:293` exposes `public_game(&game_code)` for this exact single-game lookup. Why it matters: this was already called out in the first review and by Gemini; it keeps avoidable allocation/list construction on the hot create-game path and bypasses the lobby API that prevents this pattern. Suggested fix: replace the block with `if public { lob.public_game(&game_code) } else { None }` while still taking the snapshot inside the lobby lock.

matthewevans (approved):
Reviewed current head after the branch update (`1489ac35e4de22cd7357278ce80214d5025d9094`). The prior blockers are addressed: the stray gitlink is gone, the regression now drives the production `create_and_connect_multiplayer_session` lock-scoping path, and the lobby snapshot uses the broker `public_game(&game_code)` lookup instead of a manual `public_games()` scan. I also checked the current diff is scoped to `crates/phase-server/src/main.rs`, `git diff --check origin/main...HEAD` passed, and `cargo fmt --all --check` passed locally.

PR #4568, opened 2026-06-28
fix(parser): UNSUPPORTED one-off: Run for Your Life

ntindle:
## Summary

**Root cause:** UNSUPPORTED one-off: Run for Your Life

## Cards corrected
- Run for Your Life

## Fix
Cluster 62 ("Run for Your Life" — Escape on instants/sorceries) was a class bug, not a one-off: the em-dash alternative-cost keyword family in parse_keyword_from_oracle held 10 siblings (cycling, ward, flashback, bestow, buyback, evoke, echo, embalm, eternalize, escalate) but was missing escape. Escape worked on permanents only via a position-sensitive dedicated intercept in oracle.rs that is unreachable for spells (the is_spell catch-all consumes the line first), so every instant/sorcery with Escape fell through to Effect::Unimplemented.

FIX (pure extension, no new types/variants): (1) Registered the escape-em-dash branch in parse_keyword_from_oracle, delegating to the existing parse_escape_keyword, gated by a nom tag("escape\u{2014}") combinator — joining escape to the family it always belonged to. (2) Deleted the now-dead dedicated escape intercept in oracle.rs and trimmed the resulting unused parse_escape_keyword import. (3) Refreshed two stale comments (Priority 8f rationale and the old intercept site). Escape extraction is now card-type-agnostic: both generic is_keyword_cost_line guards (Priority 9 spells, Priority 13 permanents) extract escape uniformly.

TESTS: building-block test parse_keyword_from_oracle_escape_em_dash (oracle_keyword.rs), type-agnostic dispatch test escape_keyword_extracted_on_instants_and_sorceries across Instant/Sorcery/Creature/Enchantment (oracle.rs), and a runtime integration file run_for_your_life_escape.rs (3 tests) built from the REAL parsed Oracle text via from_oracle_text: castable-with-4-cards, not-castable-with-3, and a full end-to-end cast that declares a target, pays the 4-card exile additional cost (WaitingFor::PayCost ExileFromZone, stack variant Escape), resolves, and asserts the targeted creature gained haste + exactly 4 cards moved graveyard to exile + the instant returned to graveyard.

VERIFICATION (cargo direct, worktree not under Tilt): cargo fmt clean; cargo clippy -p engine --all-targets -D warnings exit 0; cargo test -p engine exit 0 (2 new parser unit tests + 3 new integration tests pass; all pre-existing Uro/Kroxa escape tests still green); oracle-gen on all 8 class cards (run for your life, cling to dust, from the catacombs, fruit of tizerus, glimpse of freedom, satyr's cunning, sleep of the dead, sweet oblivion) -> each emits exactly one Escape keyword, 0 Unimplemented; Run for Your Life parses Escape NonMana(Composite[Mana{2}{U}{R}, Exile{count:4, Graveyard}]) plus AddKeyword(Haste) + CantBeBlocked body. Parser diff gate (vs HEAD) passes — zero string-dispatch in added parser lines. check-parser-combinators.sh exits 1 but ONLY from PRE-EXISTING committed code between its baseline ff799f899 and HEAD (lower.split, starts_with("This spell costs"), contains("strive")); my added lines add zero forbidden patterns (verified vs HEAD).

JUDGEMENT CALLS: (a) Used the tag(...).parse(text).is_ok() form for the escape gate instead of the plan's literal if let Ok((_, _)) — identical behavior, matches the existing craft-branch idiom in the same function, avoids a possible clippy redundant-pattern lint. (b) The fluent SpellCast driver does not handle the ExileFromZone PayCost window, so I drove the cast with raw CastSpell + a bounded driver loop (mirroring the Uro and Nourishing Shoal tests) and proved full resolution including the haste grant — a complete card-test runtime proof on the previously-unreachable instant path. (c) Reverted one unintended side-effect: running oracle-gen for verification rewrote crates/engine/data/oracle-subtypes.json (added "Shi-Ar" from current MTGJSON); that file was clean at session start and unrelated to the fix, so I restored it to keep the diff surgical.

DEVIATION: Did NOT run gen-card-data.sh (plan Step 7). card-data.json is a multi-MB downstream generated artifact regenerated by the card-data Tilt resource / release pipeline; oracle-gen already proves the parser output is correct for all 8 cards, and regenerating it would create large shared-file churn the orchestrator must stage. Left to the pipeline.

RISKS: Low. Pure parser change, no runtime/AI/frontend/types touched. Snapshot fixtures contain no escape cards, so no insta snapshots change. The /review-impl loop should mainly re-confirm the permanent escape path (Uro/Kroxa) — covered by passing existing tests, which now flow through the Priority 13 generic guard with byte-identical parse_escape_keyword output.

## Files changed
- crates/engine/src/parser/oracle_keyword.rs
- crates/engine/tests/integration/run_for_your_life_escape.rs
- crates/engine/tests/integration/main.rs

## CR references
- CR 702.138a — Escape: cast from graveyard for an alternative cost (verified grep ^702.138 in docs/MagicCompRules.txt) — new parse_keyword_from_oracle escape branch comment, replacement comment at old intercept site, and all three test sites
- CR 702.10a — Haste is a static ability (verified grep ^702.10a) — integration test haste assertion
- CR 601.2c — player announces target choices before cost payment (verified grep ^601.2c) — integration test target-ordering comment
- CR 608.2n — as the final part of an instant's resolution it is put into its owner's graveyard (verified grep ^608.2n; corrected from an initial mis-citation of 608.2m, which is the leaves-stack-mid-resolution rule) — integration test graveyard assertion

## Verification
- `cargo fmt --all` — passed
- `check-parser-combinators.sh (scoped to upstream/main merge-base f407b89)` — passed
- `cargo clippy -p engine --all-targets -- -D warnings` — passed
- `oracle-gen data --filter 'run for your life'` — passed
Cards confirmed re-parsed correctly: Run for Your Life


matthewevans (approved):
Approved. I reviewed current head `054eeb02d8f3f4d8bd72dc9b53955cc148ccffaa` against the maintainer bar.

What I checked:
- The implementation does not add a Run for Your Life one-off. It registers `escape—...` in the existing `parse_keyword_from_oracle` keyword-cost path, so the generic spell/permanent keyword guards extract Escape uniformly.
- The parser unit coverage verifies the keyword extraction across instant, sorcery, creature, and enchantment surfaces and asserts the line does not leave an `Unimplemented` ability.
- The integration test builds Run for Your Life from real Oracle text, verifies graveyard castability only with four other cards, pays the exile cost through the public action pipeline, and proves the resolved haste grant plus graveyard/exile zone outcomes.
- CI is green and the PR is backend/parser scoped.

This is an appropriately small parser fix with runtime proof for the previously-missed instant Escape path.

PR #4574, opened 2026-06-29
fix(engine): Tolarian Winds draws after discarding hand

### LINKED ISSUE #4572: Tolarian Winds discards hand but does not draw that many cards
## Report

Tolarian Winds discards the caster's hand when cast, but it does not draw that many cards afterward.

## Expected behavior

After discarding all cards in hand, the caster should draw the same number of cards that were discarded.

## Oracle text

Verified from the coverage export used by the bug classifier:

> Discard all the cards in your hand, then draw that many cards.

## Classifier analysis

**Matched clause**: Discard all the cards in your hand, then draw that many cards. (supported: true)

**Reasoning**: The coverage export marks the ability supported and parses it as a Discard effect with count cards in hand (you), followed by a child Draw effect with count event amount. That AST matches the Oracle text, so the reported behavior points to a runtime defect: the discard resolves, but the discarded-card count is not being forwarded/resolved for the following draw effect.

## Notes

- Local client/public/card-data.json is not present in this checkout, so the Oracle text and parse details were verified from /tmp/coverage-data.json, fetched from the preview coverage export.
- Likely area to inspect: event amount propagation from Discard into chained Draw effects.

kiannidev:
## Summary
- Stamp `last_effect_count` from discard events in the generic ability chain body, matching the existing `player_scope` path
- Lets chained `Draw` effects with `EventContextAmount` resolve to the number of cards discarded

Fixes #4572

## Test plan
- [x] Added `discard_hand_then_draw_event_context_amount_without_player_scope` regression test
- [x] `cargo test -p engine discard_hand_then_draw_event_context_amount_without_player_scope`

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -6200,15 +6200,22 @@ fn resolve_chain_body(
     }
 
     // CR 609.3: Extract the numeric result emitted by this parent effect for
-    // `QuantityRef::PreviousEffectAmount` in sub-abilities. The event class is
-    // selected by the parent `Effect` so unrelated numeric side effects from the
-    // same resolution are not mixed together: damage to a battle removes defense
-    // counters and also deals damage, but "damage dealt this way" must read only
-    // `DamageDealt`; Coalition Relic's "counter removed this way" must read only
-    // `CounterRemoved`.
-    if let Some(amount) =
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head. The fix is at the shared effect-chain event accounting seam, exact-head CI is green, Gemini had no substantive findings, and the regression covers the discard-hand then draw-that-many chain that failed before. Approving for merge queue.

PR #4575, opened 2026-06-29
fix(parser): Jocasta graveyard return enters tapped and attacking

### LINKED ISSUE #4566: Jocasta, Automaton Avenger returns without tapped and attacking qualifiers
## Report

Jocasta, Automaton Avenger second triggered ability triggers from the graveyard when you attack with your commander, but Jocasta is not returned to the battlefield tapped and attacking.

## Expected behavior

When you attack with your commander while Jocasta is in your graveyard, the optional trigger should be able to return Jocasta to the battlefield tapped and attacking.

Oracle text:

> Whenever you attack with your commander, if this card is in your graveyard, you may return it to the battlefield tapped and attacking.

## Classifier analysis

**Matched clause**: `Whenever you attack with your commander, if this card is in your graveyard, you may return it to the battlefield tapped and attacking.` (supported: true)

**Reasoning**: The coverage export marks this graveyard trigger supported and parses it as a `YouAttack` trigger with an optional `ChangeZone` child targeting the triggering source and moving it to the battlefield. However, the parsed child details only include `to: battlefield` and `target: triggering source`; they do not encode the Oracle text qualifiers `tapped` or `attacking`. The runtime report matches that AST gap: the trigger fires, but the return effect does not place Jocasta onto the battlefield tapped and attacking.

## Notes

- Reporter used the name "Jocasta, Automation Avenger"; card data canonical name is "Jocasta, Automaton Avenger".
- Likely area to inspect: parsing/synthesis for `return it to the battlefield tapped and attacking` on self-recursive graveyard attack triggers, including `ChangeZone.enter_tapped` and `enters_attacking` propagation.

kiannidev:
## Summary
- Parse "whenever you attack with your commander" as a `YouAttack` trigger gated on `IsCommander`.
- When a trigger intervening-if pins the source in the graveyard, stamp implicit "return it to the battlefield tapped and attacking" effects with `ChangeZone.origin = Graveyard` and bind the target to `SelfRef`.
- Add parser and integration regression tests for Jocasta, Automaton Avenger.

Fixes #4566

## Test plan
- [x] `cargo test -p engine trigger_jocasta_commander_attack_return_tapped_and_attacking`
- [x] `cargo test -p engine --test integration jocasta_returns`


matthewevans (changes requested):
**[MED]** The self-return origin stamping is broader than the implicit self-return shape it is meant to fix. Evidence: `crates/engine/src/parser/oracle_trigger.rs:347` recursively walks every sub/else ability, and `crates/engine/src/parser/oracle_trigger.rs:364` rewrites any `ParentTarget` or `TriggeringSource` `ChangeZone` to `SelfRef` when it goes to battlefield or hand. Why it matters: a graveyard/exile-functioning trigger whose condition says "this card is in your graveyard/exile" can still have a later sub-ability where `ParentTarget` legitimately refers to another object selected or moved by the parent effect; rewriting that nested target to `SelfRef` makes the parser silently return the source instead while the line looks supported. Suggested fix: restrict the stamp to the specific implicit self-return `ChangeZone` shape that has no real parent object to inherit, and add a regression proving a nested `ParentTarget` return remains a parent-target return.

matthewevans (approved):
Prior blocker is addressed. I re-reviewed the current head and the origin stamping is now limited to the top-level execute effect, only covers implicit self-return targets, and the regression test preserves nested ParentTarget returns. CI is green and the backend parser/runtime coverage discriminates the Jocasta path.

PR #4580, opened 2026-06-29
fix(engine): Winter Soldier attack trigger and Hero counter rider (#4560)

### LINKED ISSUE #4560: [Card Bug] Winter Soldier, Reborn Avenger attack trigger does not fire
## Card(s)
Winter Soldier, Reborn Avenger

## Build/version
Unknown

## Game mode
Unknown

## Actual behavior
Reported: Winter Soldier, Reborn Avenger's triggered ability is not triggering.

Coverage classification confirms the card's attack-trigger line is not represented faithfully: deployed coverage parses the full line as a `replacement` / `Moved` node with only a malformed `PutCounter` child, rather than an attack trigger that returns a target creature card from the graveyard and conditionally adds an additional +1/+1 counter if that returned creature is a Hero.

## Expected behavior
When Winter Soldier attacks, its triggered ability should trigger and target a creature card in its controller's graveyard with mana value less than or equal to Winter Soldier's power. On resolution, that card should return to the battlefield. If a Hero enters this way, it should enter with an additional +1/+1 counter on it.

## Classifier analysis

**Matched clause**: `Whenever ~ attacks, return target creature card with mana value less than or equal to ~'s power from your graveyard to the battlefield. If a Hero enters this way, it enters with an additional +1/+1 counter on it.` (supported: true)

**Reasoning**: The parsed AST is unfaithful to the Oracle text. The clause should lower as an attack trigger with a graveyard target and a return-to-battlefield effect, plus a conditional extra +1/+1 counter for Heroes entering this way. Instead, coverage reports category `replacement`, label `Moved`, and a single `PutCounter` child whose counter detail contains most of the reanimation sentence. That explains the reported trigger failure and gives the fix a concrete parser/IR starting point.

## Steps to reproduce
1. Control Winter Soldier, Reborn Avenger.
2. Have a legal target creature card in your graveyard with mana value less than or equal to Winter Soldier's power.
3. Attack with Winter Soldier.
4. Observe that the attack trigger does not appear to trigger / resolve as Oracle text specifies.

## Notes
Local checkout source data did not contain this card in `data/mtgish-cards.json`, but the deployed preview coverage artifact did include it and showed the malformed parse above.

kiannidev:
## Summary
- Fix misclassification of Winter Soldier, Reborn Avenger's oracle text as an enters-with replacement; trigger lines with reflexive "enters this way" counter riders now route through trigger dispatch.
- Parse present-tense "enters this way" zone-change gates and the "it enters with an additional +1/+1 counter" rider on graveyard reanimation.
- Add parser and integration tests covering parse output and in-game Hero reanimation with an extra counter.

Fixes #4560

## Test plan
- [x] `cargo test -p engine --lib winter_soldier_reborn_avenger`
- [x] `cargo test -p engine --test integration issue_4560`


matthewevans (changes requested):
**[MED]** The Hero "enters with" rider is modeled as a later counter effect instead of an entry-time counter. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:884` emits `Effect::PutCounter`, which resolves through the normal counter effect path (`crates/engine/src/game/effects/counters.rs:1021`) after the zone move, while actual "enters with" counters are applied inside the zone-change entry pipeline (`crates/engine/src/game/zone_pipeline.rs:1758`). Why it matters: "it enters with an additional +1/+1 counter" is an entry replacement-style property of the battlefield entry, so entry-time observers/replacements can differ even though the final counter count test passes. Suggested fix: carry the conditional Hero rider into the `ChangeZone`/entry pipeline as conditional `enter_with_counters` (or an equivalent entry-time structure), and add a test with an entry-time observer that distinguishes entering with the counter from receiving it afterward.

matthewevans (changes requested):
**[MED]** The Hero "enters with" rider is still modeled as a post-entry counter effect. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:884` still emits `Effect::PutCounter { target: TargetFilter::ParentTarget }` for "it enters with an additional +1/+1 counter on it`; that runs after the parent `ChangeZone` has already moved the object, instead of carrying the counter into the entry pipeline via `ChangeZone.enter_with_counters`. Why it matters: entry-time observers/replacements can distinguish entering with the counter from receiving the counter afterward, so the final counter-count integration test does not prove the rules behavior. Suggested fix: carry the conditional Hero rider into the `ChangeZone`/entry pipeline as conditional entry counters, and add a regression that distinguishes entry-time counters from a later `PutCounter`.

matthewevans (changes requested):
**[MED]** The regression coverage still does not distinguish entry-time counters from a post-entry counter effect. Evidence: `crates/engine/tests/integration/issue_4560_winter_soldier.rs:183` only asserts the returned Hero's final battlefield zone and final +1/+1 counter count, while the prior failure mode could produce the same final state with a later `PutCounter`; the parse assertion at `crates/engine/tests/integration/issue_4560_winter_soldier.rs:112` proves the AST shape but does not exercise an entry-time observer/replacement through the runtime pipeline. Why it matters: the original bug class was timing-sensitive: "enters with" must be visible during battlefield entry, not merely after resolution. Suggested fix: add a runtime regression with an entry-time observer/replacement/counter-modifier interaction that fails if the Hero counter is applied after the zone move rather than inside the entry pipeline.

kiannidev:
Merged latest `main`, resolved `main.rs` conflict. Hero counter rider uses `conditional_enter_with_counters` on `ChangeZone` (not post-entry `PutCounter`); parse + integration tests updated accordingly. CI re-running — ready for re-review.

matthewevans (changes requested):
**[MED]** Resolution-choice `ChangeZone` paths drop the new conditional entry counters. Evidence: `crates/engine/src/game/effects/change_zone.rs:587` creates `WaitingFor::EffectZoneChoice` without carrying `conditional_enter_with_counters`, and `crates/engine/src/game/engine_resolution_choices.rs:2795` resumes with `enter_with_counters: vec![]` / `conditional_enter_with_counters: vec![]`. Why it matters: any resolution-time "return/put a card ... If a [type] enters this way, it enters with ..." shape can parse into the new field but enter without the rider when it goes through the choice path. Suggested fix: carry the conditional specs or resolved per-object counters through `EffectZoneChoice`, then compute/merge per-object counters before `process_one_zone_move`.

**[MED]** The prior timing-test blocker is still not fixed. Evidence: `crates/engine/tests/integration/issue_4560_winter_soldier.rs:183` only asserts the final battlefield zone, and `crates/engine/tests/integration/issue_4560_winter_soldier.rs:188` only asserts the final counter count. Why it matters: a post-entry `PutCounter` implementation can satisfy the same final-state assertion, so the regression still does not prove the Hero counter is applied as an entry-time counter rather than after the zone move. Suggested fix: add a runtime test with an ETB counter replacement/observer that fails if the Hero counter is applied after the object enters.

matthewevans (changes requested):
**[MED]** The `EffectZoneChoice` conditional-entry-counter path still has no discriminating regression. Evidence: `crates/engine/src/game/engine_resolution_choices.rs:2648` now carries `conditional_enter_with_counters` through `WaitingFor::EffectZoneChoice`, and `crates/engine/src/game/engine_resolution_choices.rs:2806`-`2816` applies it when a selected card is moved, but the new Winter Soldier regression drives the normal targeted `ChangeZone` trigger path (`crates/engine/tests/integration/issue_4560_winter_soldier.rs:249`) and the existing `change_zone.rs` test fixtures still construct `EffectZoneChoice`/`ChangeZone` cases with `conditional_enter_with_counters: vec![]`. Why it matters: the last blocker was specifically that resolution-choice `ChangeZone` paths dropped this rider; without a choice-path fixture, a future regression can erase the newly threaded field while the Winter Soldier tests still pass. Suggested fix: add a focused regression that reaches the `EffectZoneChoice` `SelectCards` branch with a non-empty `conditional_enter_with_counters` rider and verifies the selected matching card enters with the counter while a non-matching card does not.

matthewevans (changes requested):
**[MED] Missing regression for the previously blocked `EffectZoneChoice` conditional-counter path.**

The production path now carries `conditional_enter_with_counters` into `WaitingFor::EffectZoneChoice` (`crates/engine/src/game/effects/change_zone.rs`) and applies it in the `SelectCards` branch (`crates/engine/src/game/engine_resolution_choices.rs`), but the new `EffectZoneChoice` regression still initializes `conditional_enter_with_counters: vec![]`. The Winter Soldier integration tests cover the targeted attack-trigger path and final Hero counter result, but they do not discriminate the specific interactive `EffectZoneChoice` `SelectCards` branch that previously dropped this field.

Please add a focused regression that parks or reaches `WaitingFor::EffectZoneChoice` with a non-empty `conditional_enter_with_counters` rider, selects at least one matching Hero and one non-matching creature/card through `GameAction::SelectCards`, and asserts only the matching selected object enters with the extra counter. That would fail if the field were dropped during the `EffectZoneChoice` round trip.

matthewevans (changes requested):
Holding this head. CI is failing in the PR's own Winter Soldier parser coverage:  panics at  with . Because that failure is directly in the touched behavior, I can't treat it as merge-when-ready noise. Please update the parser/test path so the current Winter Soldier representation is internally consistent and the new entry-time counter modeling remains covered.

matthewevans (changes requested):
Holding this head. CI is failing in the PR Winter Soldier parser coverage: `parser::oracle_trigger::tests::winter_soldier_reborn_avenger_attack_reanimation_trigger` panics at `crates/engine/src/parser/oracle_trigger.rs:14474` with `Hero counter rider`. Because that failure is directly in the touched behavior, I cannot treat it as merge-when-ready noise. Please update the parser/test path so the current Winter Soldier representation is internally consistent and the new entry-time counter modeling remains covered.

matthewevans (changes requested):
**[HIGH]** The `EffectZoneChoice` resume path still drops conditional entry counters in production. Evidence: `crates/engine/src/game/engine_resolution_choices.rs:2797` looks up the resolving ability only by scanning `state.stack`, but `resolve_top` pops the entry before resolving and stores it in `state.resolving_stack_entry` (`crates/engine/src/game/stack.rs:192`, `crates/engine/src/game/stack.rs:460`). The new regression masks this by manually pushing a fake stack entry before `SelectCards` (`crates/engine/src/game/effects/change_zone.rs:4457`), so it exercises a state shape production resolution does not have. Why it matters: when a real resolving `ChangeZone` pauses for `EffectZoneChoice`, `stack_ability` is `None`, so the branch falls back to `enter_with_counters.clone()` and silently ignores `conditional_enter_with_counters`; the Hero selected through this path will not get the entry-time counter. Suggested fix: resolve the ability from the active resolving entry as well as the live stack, and change the regression to model production by using `state.resolving_stack_entry` rather than pushing a fake live stack entry.

matthewevans (changes requested):
**[MED]** Conditional entry counters are lost or copied incorrectly after a selected zone move pauses. Evidence: `crates/engine/src/game/engine_resolution_choices.rs:2812` computes `per_obj_enter_counters` for the current selected card, but the pause stashes only that resolved vector into `PendingChangeZoneIteration.enter_with_counters` (`crates/engine/src/game/engine_resolution_choices.rs:2863`, `crates/engine/src/game/engine_resolution_choices.rs:2895`); the drain path then reconstructs `ChangeZoneIterationCtx` with `conditional_enter_with_counters: vec![]` (`crates/engine/src/game/effects/mod.rs:744`) and reuses the same `enter_with_counters` for every remaining object. Why it matters: if the first selected object pauses for an Aura attachment or replacement/as-enters choice, later selected objects no longer get per-object conditional evaluation, so later Heroes can miss the counter or non-Heroes can inherit the paused object's counter. Suggested fix: carry the conditional counter specs plus resolving ability provenance through `PendingChangeZoneIteration` and recompute `enter_with_counters_for_object` for each remaining object during drain, or store per-object resolved counters for the remaining queue.

matthewevans (changes requested):
**[MED]** Conditional "enters this way" counters are skipped on the single-pick fast paths. Evidence: `crates/engine/src/game/effects/change_zone.rs:460` and `crates/engine/src/game/effects/change_zone.rs:525` both call `execute_zone_move` with only `&effect_enter_with_counters`; unlike the main multi-object loop and the `EffectZoneChoice` path, neither branch evaluates `effect_conditional_enter_with_counters` with `enter_with_counters_for_object`. Why it matters: a `ChangeZone` effect that resolves to one eligible/random object can enter a matching Hero without the entry-time counter even though the multi-object/choice path is now correct. Suggested fix: compute the per-object counter vector in both single-pick branches before moving the object, or route those branches through `process_one_zone_move` with `ChangeZoneIterationCtx`, and add a single-eligible conditional-counter regression.

matthewevans (approved):
Reviewed current head. The Winter Soldier entry-time counter modeling now carries conditional entry-counter specs through the normal, EffectZoneChoice, pause/resume, and single-pick ChangeZone paths with discriminating engine coverage. Merge when ready.

PR #4582, opened 2026-06-29
test(parser): wishboard outside-game fetch regression (#2276)

### LINKED ISSUE #2276: Parser: optional wishboard / outside-the-game fetch dropped (#2231 split D)
Sub-issue of #2231 (`Optional_YouMay` swallowed-clause cluster). One structural pattern per issue.

## Pattern: wishboard / outside-the-game optional fetch

`you may reveal a [card type] card you own from outside the game and put it into your hand` (optionally `… or choose a face-up [type] card you own in exile`)

The optional "from outside the game" fetch (Wish cycle, Learn/wishboard, Karn) is silently dropped while the card stays **supported**.

## Repro

```bash
target/tool/coverage-report data --all --warning-detector Optional_YouMay --warning-limit 500 \
  | jq '.cards[] | select(.warnings[].pattern | test("outside the game"))'
```

## Expected

Represented as an optional fetch from outside the game (CR 731 "outside the game"; CR 408 sideboard in non-tournament play). If the wishboard zone/optional fetch can't be represented, the ability must not stay fully supported.

## Affected cards (10)

Burning Wish · Coax from the Blind Eternities · Cunning Wish · Glittering Wish · Golden Wish · Granted · Karn, the Great Creator · Living Wish · Retriever Phoenix · Vivien, Arkbow Ranger

kiannidev:
## Summary
- Add swallow-check regression tests for Living Wish (creature/land filter), Karn, the Great Creator's -2, and Vivien, Arkbow Ranger's -5 outside-the-game fetches.
- Confirms these lines lower to `SearchOutsideGame` with up-to optional selection and do not emit `Optional_YouMay` swallow warnings.

Fixes #2276

## Test plan
- [x] `cargo test -p engine --lib optional_you_may_accepts_wishboard`
- [x] `cargo test -p engine --lib optional_you_may_accepts_outside_game`


matthewevans (changes requested):
Thanks for adding coverage here. I don't think this is ready to land yet because the new regression test can pass for the wrong reason.

`check_swallowed_clauses` returns immediately when any parsed ability contains `Effect::Unimplemented`, so each `!has_swallowed_detector(..., "Optional_YouMay")` assertion is only meaningful if the fixture first proves the card did not fall back to `Unimplemented`. This matters for these examples because a partial parse can still expose a `SearchOutsideGame` shape while the swallowed-clause detector is globally suppressed by an unsupported sibling/sub-ability.

Please add explicit guards for each parsed fixture, e.g. assert the relevant ability trees do not contain `def_tree_has_unimplemented`, before the negative swallowed-detector assertions. That will make this a real regression test instead of a detector-suppression check.

matthewevans (approved):
Reviewed current head `985543263148bbf7f129405293224ad74e19f3f8`. The prior blocker is addressed: the regression now guards against Unimplemented parses and checks the expected SearchOutsideGame shapes for Living Wish, Karn, and Vivien.

PR #4583, opened 2026-06-29
test(parser): APNAP regression for Protection Racket (#2241)

### LINKED ISSUE #2241: Parser drops APNAP ordering clauses while coverage still marks affected cards supported
## Summary

`./scripts/gen-card-data.sh` reports `1` `Swallow:APNAP` diagnostic. The affected card is otherwise marked supported by coverage.

## Repro

```bash
./scripts/gen-card-data.sh
target/tool/coverage-report data --all --warning-detector APNAP --warning-limit 5
```

Fresh drilldown:

```text
matched_cards: 1
supported_cards: 1
single_gap_cards: 0
example: Protection Racket
```

## Expected

Ordering clauses such as `starting with...` / `in turn order` should be represented in the choice/effect workflow. If unsupported, the affected ability should not remain fully supported.

kiannidev:
## Summary
- Add a swallow-check regression test for Protection Racket's "repeat the following process for each opponent in turn order" upkeep trigger.
- Confirms the parser stamps opponent iteration scope and does not emit `APNAP` swallow warnings.

Fixes #2241

## Test plan
- [x] `cargo test -p engine --lib apnap_accepts_protection_racket`


matthewevans (changes requested):
Thanks for adding the APNAP coverage. I see two issues that need to be fixed before this can land:

1. This PR appears to include the wishboard regression test from #4582 in addition to the Protection Racket test. Please rebase/split so this PR only carries the #2241 APNAP change.

2. The Protection Racket assertion is negative-only. `check_swallowed_clauses` suppresses swallowed-clause diagnostics whenever any parsed ability contains `Effect::Unimplemented`, so `assert!(!has_swallowed_detector(&parsed, "APNAP"))` can pass even if the trigger body is still unsupported. Please also assert the parsed trigger execute tree has no `Unimplemented` and pin the positive APNAP shape, such as the expected opponent `player_scope`, before checking the detector warning is absent.

matthewevans (approved):
Reviewed current head b35f678e. This now addresses the previous APNAP-test blocker: the unrelated wishboard coverage is gone, and the Protection Racket regression asserts no Unimplemented plus the positive opponent player_scope before the negative swallowed-detector check. CI is green.

PR #4584, opened 2026-06-29
test(parser): Duration_NextTurn regression for Amplifire (#2239)

### LINKED ISSUE #2239: Parser drops Duration_NextTurn clauses while coverage still marks affected cards supported
## Summary

`./scripts/gen-card-data.sh` reports `1` `Swallow:Duration_NextTurn` diagnostic. The affected card is otherwise marked supported by coverage.

## Repro

```bash
./scripts/gen-card-data.sh
target/tool/coverage-report data --all --warning-detector Duration_NextTurn --warning-limit 5
```

Fresh drilldown:

```text
matched_cards: 1
supported_cards: 1
single_gap_cards: 0
example: Amplifire
```

## Expected

`until your next turn` / next-turn duration clauses should be represented as typed durations or event-window conditions, or the affected parse should be marked unsupported.

kiannidev:
## Summary
- Add a swallow-check regression test for Amplifire's "until your next turn" base P/T modification on its upkeep trigger.
- Confirms the duration is represented in the parsed AST without `Duration_NextTurn` swallow warnings.

Fixes #2239

## Test plan
- [x] `cargo test -p engine --lib duration_next_turn_accepts_amplifire`


matthewevans (changes requested):
Thanks for adding the Amplifire regression. This test needs one more positive assertion before it is useful.

Right now it only checks that `Duration_NextTurn` was not reported. `check_swallowed_clauses` returns early when any parsed ability contains `Effect::Unimplemented`, so the assertion would also pass if Amplifire stopped parsing and the detector was suppressed. Please assert the trigger/effect tree has no `Unimplemented` and that the parsed P/T layer effect actually carries the expected next-turn duration, then keep the negative swallowed-detector assertion.

matthewevans (changes requested):
**[MED]** Amplifire regression can still pass when parsing falls back to `Unimplemented`. Evidence: `crates/engine/src/parser/swallow_check.rs:3685`. Why it matters: `check_swallowed_clauses` suppresses detectors when any ability tree has `Effect::Unimplemented`, so the negative `Duration_NextTurn` assertion is not meaningful without a prior no-`Unimplemented` guard. Suggested fix: assert the Amplifire trigger/effect tree has no `Unimplemented` before the swallowed-detector assertion.

**[MED]** The duration assertion is not tied to the P/T layer effect. Evidence: `crates/engine/src/parser/swallow_check.rs:3663`. Why it matters: the helper accepts any `UntilNextTurnOf { Controller }` anywhere in the trigger chain, so it does not prove the parsed `SetPowerDynamic`/`SetToughnessDynamic` layer clause carries the expected duration requested by the previous review. Suggested fix: locate the P/T layer ability specifically, assert it contains the expected dynamic base power/toughness modifications, and assert that same ability has `Duration::UntilNextTurnOf { player: PlayerScope::Controller }`.

matthewevans (changes requested):
**[MED]** Amplifire regression can still pass if parsing falls back to `Unimplemented`. Evidence: `crates/engine/src/parser/swallow_check.rs:3691` still asserts only that `Duration_NextTurn` is absent; there is no prior guard that the Amplifire trigger/effect tree contains no `Effect::Unimplemented`. Why it matters: `check_swallowed_clauses` suppresses detectors when any ability tree has `Unimplemented`, so the negative detector assertion can pass vacuously. Suggested fix: assert the trigger/effect tree has no `Unimplemented` before checking the swallowed-detector result.\n\n**[MED]** The duration assertion is still not tied to the P/T layer effect. Evidence: `crates/engine/src/parser/swallow_check.rs:3670` walks the whole execute tree looking for any `Duration::UntilNextTurnOf { player: Controller }`; it does not locate the dynamic base power/toughness layer clause. Why it matters: the previous review asked to prove the parsed P/T layer effect carries the next-turn duration, and this helper would pass if an unrelated sub-ability had that duration while the P/T modification remained unscoped. Suggested fix: locate the specific dynamic base power/toughness modification ability and assert that same ability has `Duration::UntilNextTurnOf { player: PlayerScope::Controller }`.

matthewevans (changes requested):
**[MED]** Amplifire regression can still pass if the P/T clause is preserved as `Unimplemented`. Evidence: `crates/engine/src/parser/swallow_check.rs:3669` only finds any descendant with `Duration::UntilNextTurnOf`, and `crates/engine/src/parser/swallow_check.rs:3691` still only asserts `Duration_NextTurn` is absent; there is no guard that the Amplifire trigger tree has no `Effect::Unimplemented` or that the duration is attached to the parsed base-power/base-toughness layer effect. Why it matters: `check_swallowed_clauses` returns early when any ability tree has `Unimplemented`, so the negative detector assertion can pass while the actual P/T clause remains unsupported. Suggested fix: add a structural guard like `!def_tree_has_unimplemented(execute)` and assert the timed sub-ability is the expected parsed P/T `GenericEffect`/continuous modification, then keep the `Duration_NextTurn` negative assertion.

kiannidev:
Merged `main`, implemented split dynamic base P/T parsing for Amplifire, and strengthened the regression with no-`Unimplemented` + P/T-layer duration guards per review. CI re-running — ready for re-review.

matthewevans (approved):
Reviewed current head. Prior blockers are fixed: the regression now asserts no `Effect::Unimplemented` and ties the duration check to the parsed P/T layer effect path. The parser change is narrow, reuses existing quantity parsing, and current GitHub checks are green.

PR #4585, opened 2026-06-29
test(parser): Condition_AsLongAs regression for equipment (#2234)

### LINKED ISSUE #2234: Parser drops Condition_AsLongAs clauses while coverage still marks affected cards supported
## Summary

`./scripts/gen-card-data.sh` reports `39` `Swallow:Condition_AsLongAs` diagnostics across `39` cards. All `39` are otherwise marked supported by coverage, so `as long as` conditions may be missing from exported AST while coverage stays green.

## Repro

```bash
./scripts/gen-card-data.sh
target/tool/coverage-report data --all --warning-detector Condition_AsLongAs --warning-limit 5
```

Fresh drilldown:

```text
matched_cards: 39
supported_cards: 39
single_gap_cards: 0
examples: Braided Net, Bronze Horse, Camel, Champion's Helm, Cloud, Midgar Mercenary
```

## Expected

`as long as` clauses should be represented as typed static/ability/trigger conditions. If the parser cannot represent the condition, the affected ability should not remain counted as fully supported.

kiannidev:
## Summary
- Add swallow-check regression tests for Bronze Horse and Champion's Helm "as long as" qualifiers folded into attached-subject static grants.
- Confirms these lines do not emit `Condition_AsLongAs` swallow warnings when the qualifier is represented structurally.

Fixes #2234

## Test plan
- [x] `cargo test -p engine --lib condition_as_long_as_accepts_bronze_horse`

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -1912,6 +1912,7 @@ pub(crate) fn lower_effect_chain_ir(ir: &EffectChainIr) -> AbilityDefinition {
     fold_token_it_has_grants_into_token_statics(&mut result);
     nest_whenever_this_turn_token_cleanup_delayed_trigger(&mut result);
     rewire_result_anchored_subchain(&mut result);
+    fold_enters_this_way_counter_rider(&mut result);
     wire_optional_cast_decline_fallback(&mut result);
     retarget_counter_additional_cost_to_target(&mut result);
     // CR 608.2c + CR 608.2b: resolve a chained tap/untap anaphor against a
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
Thanks for adding the condition regression. This has the same issue as the other swallowed-clause regression tests: the assertions only prove the warning is absent, not that the parser represented the condition.

`check_swallowed_clauses` suppresses detectors when any parsed ability/static contains `Effect::Unimplemented`, so `!has_swallowed_detector(..., "Condition_AsLongAs")` can pass if the relevant static failed to parse. Please add positive guards for both fixtures: no `Unimplemented` in the relevant static/effect tree, plus an assertion for the condition/filter shape that should represent the `as long as` clause.

matthewevans (approved):
Prior blocker is addressed on the updated branch. The regression now guards both fixtures with no-`Unimplemented` checks plus positive AST-shape assertions before checking that `Condition_AsLongAs` is not reported, which makes the test discriminate against the swallowed-condition failure mode.

kiannidev:
Rebased onto latest `main`, fixed combinator gate annotation on test assertion. CI re-running — ready for re-review.

matthewevans (changes requested):
**[LOW]** The parser-combinator escape hatch is on the wrong line. Evidence: `crates/engine/src/parser/swallow_check.rs:3227` still trips the parser gate for the added `.contains("as long as")` call on the current head. Why it matters: the behavioral assertions look substantially improved, but this head is not clean/enqueueable while the parser gate flags the new string method. Suggested fix: move `// allow-noncombinator: test assertion on parsed replacement description` onto the same line as the `.contains(...)` call, or avoid the forbidden string method.

matthewevans (approved):
Prior blocker is fixed on the updated head. The allow-noncombinator annotation now covers the actual test-only `contains("as long as")` assertion, `scripts/check-parser-combinators.sh` passes, and the PR diff against current `main` remains limited to the discriminating swallow-check regression.

matthewevans (approved):
Re-approving the maintainer fixup head. The only post-approval change binds the expected phrase in the test assertion so rustfmt no longer moves the parser-gate escape hatch; local `cargo fmt --all --check`, `scripts/check-parser-combinators.sh`, and `git diff --check` passed before push.

PR #4586, opened 2026-06-29
fix(parser): UNSUPPORTED one-off: The Master, Formed Anew

ntindle:
## Summary

**Root cause:** UNSUPPORTED one-off: The Master, Formed Anew

## Cards corrected
- The Master, Formed Anew

## Fix
Test minimal output to isolate schema parsing.

## Files changed

## CR references
- CR 122.1

## Verification
- `cargo fmt --all` — pass
- `./scripts/check-parser-combinators.sh <upstream/main merge-base 4e103aabc>` — pass
- `cargo run --profile tool --features cli --bin oracle-gen -- data --filter "the master, formed anew"` — pass
Cards confirmed re-parsed correctly: The Master, Formed Anew

### DIFF
--- a/crates/engine/src/game/engine_replacement.rs
+++ b/crates/engine/src/game/engine_replacement.rs
@@ -2787,6 +2787,71 @@ mod tests {
         );
     }
 
+    /// CR 400.1 + CR 122.1: The Master, Formed Anew — the copy source is "a
+    /// creature card in exile with a takeover counter on it". `find_copy_targets`
+    /// must scan EXILE (per `InZone { Exile }`) and honor the takeover-counter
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed current head b91eb278. The production change is a narrow generalized parser fix: a counter-presence suffix may trail an explicit zone suffix, so phrases like "creature card in exile with a takeover counter on it" lower to both InZone and Counters predicates. I verified the runtime side uses the encoded zone in find_copy_targets and then applies the normal target filter, so the counter predicate is enforced against objects in exile. CI is green and the added parser/replacement/runtime tests cover the corrected parse shape and target-selection semantics.

PR #4587, opened 2026-06-29
fix(parser): parse Stridehangar appended Thopter token replacement

### LINKED ISSUE #654: Stridehangar Automaton — bonus Thopter replacement effect unimplemented
Reported in Discord: https://discord.com/channels/1485498006781427802/1506321223653658697/1506321223653658697

**Thread:** Stridehangar Automaton
**Cards:** stridehangar automaton
**Parser status:** has_gaps

## Summary
When artifact tokens are created under the controller's control, Stridehangar Automaton should create those tokens *plus* an additional 1/1 colorless Thopter artifact creature token with flying (a replacement effect). Instead, no bonus Thopter is created.

## Oracle text
Thopters you control get +1/+1.
If one or more artifact tokens would be created under your control, those tokens plus an additional 1/1 colorless Thopter artifact creature token with flying are created instead.

## Parser gap
The second line is parsed as `Effect::Unimplemented { name: "replacement_structure", ... }` — the replacement parser matches the line but cannot produce a typed replacement definition.

## Fix shape
Add a replacement definition to `oracle_replacement.rs` for the "tokens plus an additional TOKEN" pattern (a subset-expansion replacement: original tokens are preserved, an extra token is emitted alongside). Cover the class: Stridehangar Automaton, Anointed Procession-adjacent patterns.

---
<sub>report_id: `discord:1506321223653658697:1506321223653658697:0` · discord: `1506321223653658697/1506321223653658697`</sub>

andriypolanski:
## Summary

[Stridehangar Automaton](https://github.com/phase-rs/phase/issues/654) should create artifact tokens **plus** an additional 1/1 colorless Thopter artifact creature token with flying when one or more artifact tokens would be created under your control (CR 614.1a / CR 111.1). The replacement line was classified correctly but failed to parse because `normalize_additional_token_descriptor` injected a spurious leading article before P/T-prefixed specs (`"a 1/1 colorless Thopter …"`), causing `parse_token_description` to return `None` and the line to fall through as `replacement_structure` unimplemented.

## Root cause

For `"those tokens plus an additional 1/1 colorless Thopter artifact creature token with flying are created instead"`, the `"an additional "` strip left `"1/1 colorless Thopter …"`. The normalizer skipped article injection because `parse_count_expr` mis-read the leading `1` in `1/1` as a bare count — but bare `1/1 …` still breaks `parse_token_description` (count prefix consumes `1`, leaving `/1 …`). The fix prepends `"a "` for P/T-led descriptors so the count prefix consumes the article and P/T parsing sees `1/1` intact (same path as `"create a 1/1 colorless Phyrexian Mite …"`).

## Changes

| File | Change |
|------|--------|
| `crates/engine/src/parser/oracle_replacement.rs` | `additional_token_descriptor_needs_leading_article`: prepend `"a "` for subtype-only and P/T-led appended token specs |
| `crates/engine/src/parser/oracle_replacement.rs` (tests) | `parses_stridehangar_additional_thopter_token_replacement` parser regression |
| `crates/engine/tests/integration/issue_654_stridehangar_automaton.rs` | Runtime: Treasure ETB creates Treasure + Thopter; Food ETB does not append Thopter |

## Test plan

- [x] `parses_stridehangar_additional_thopter_token_replacement` — `TokenCoreTypeMatches(Artifact)` gate + `additional_token_spec` with 1/1 flying Thopter
- [x] `parses_peregrin_took_additional_food_token` — unchanged Food-token path still parses
- [x] `stridehangar_adds_thopter_when_artifact_token_would_be_created` — ETB Treasure → 1 Treasure + 1 Thopter on battlefield
- [x] `stridehangar_replacement_does_not_fire_for_non_artifact_tokens` — Squirrel creature token (non-artifact) does not append Thopter
- [ ] Manual: Stridehangar on board, create multiple artifact tokens (e.g. Treasure + Clue) — verify one Thopter per replacement event batch

## Closes

Fixes https://github.com/phase-rs/phase/issues/654

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -6164,13 +6164,12 @@ fn normalize_additional_token_descriptor(descriptor: &str) -> Option<String> {
 
 /// True when the descriptor needs a leading `"a "` before `parse_token_description`.
 /// Subtype-only specs (`"Food token"`) have no count and require an article.
-/// P/T-led specs (`"1/1 colorless Thopter …"`) also require one: without it
-/// `parse_token_count_prefix` mis-reads the leading `1` in `1/1` as a bare
-/// count and leaves `/1 …`, breaking P/T parsing (Stridehangar Automaton).
+/// P/T-led specs (`"1/1 …"`, `"10/10 …"`) also require one: without it
+/// `parse_token_count_prefix` mis-reads the leading digits before `/` as a bare
+/// count and leaves `/toughness …`, breaking P/T parsing.
 fn additional_token_descriptor_needs_leading_article(descriptor: &str) -> bool {
     let trimmed = descriptor.trim_start();
-    let bytes = trimmed.as_bytes();
-    if bytes.len() >= 2 && bytes[0].is_ascii_digit() && bytes[1] == b'/' {
+    if nom_primitives::parse_pt_value(trimmed).is_ok() {
         return true;
     }
     parse_count_expr(trimmed).is_none()
@@ -14004,7 +14003,6 @@ mod tests {
 
     /// CR 614.1a + CR 111.1: Stridehangar Automaton (#654) — artifact-token-gated
     /// "those tokens plus an additional 1/1 colorless Thopter …" replacement.
-    /// The appended spec carries explicit P/T; article injection must not run.
     #[test]
     fn parses_stridehangar_additional_thopter_token_replacement() {
         let text = "If one or more artifact tokens would be created under your control, those tokens plus an additional 1/1 colorless Thopter artifact creature token with flying are created instead.";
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
Thanks for the quick Stridehangar fix. I need one parser-generalization issue fixed before this can merge:

`additional_token_descriptor_needs_leading_article` currently detects P/T-led descriptors with `bytes[1] == b'/'`. That handles `1/1 ...`, but it does not handle the same grammar class with multi-digit power, e.g. `10/10 colorless Eldrazi creature token`. In that case `parse_count_expr("10/10 ...")` still parses `10` as a count, article injection is skipped, and `parse_token_description` receives `/10 ...` as the body.

Please make the P/T-led detection cover the full numeric P/T shape rather than the one-digit case, and add a small parser regression for a multi-digit P/T descriptor so this stays class-level rather than Stridehangar-specific.

matthewevans (approved):
Current head `f93fb284a0713a37534f3c39fb3935d26a7a6232` addresses the prior parser-generalization blocker. The P/T-led descriptor check now reuses the existing `parse_pt_value` building block rather than a one-digit slash check, and the added 10/10 regression plus scenario-runner Stridehangar tests cover both parser shape and runtime replacement behavior.

PR #4590, opened 2026-06-29
feat(engine): Heroic Defiance — gate +3/+3 on most-common-color among permanents

nickmopen:
## Summary
Implements Heroic Defiance: *"Enchanted creature gets +3/+3 unless it shares a color with the most common color among all permanents or a color tied for most common."*

Previously the Aura's `+3/+3` grant parsed but the `unless <condition>` rider was dropped, so the bonus applied unconditionally.

## Changes
- **New `StaticCondition::SharesColorWithMostCommonColorAmongPermanents`** + runtime evaluator (CR 105.2 + CR 611.3a): a source color is "most common" when its board-wide count equals the maximum, counting every color **tied** for the lead. The static gate wraps it in `Not` for the "unless" clause.
- **Parser**: recognize the most-common-color phrase in the shared static-condition parser, and fix the Aura `"+X/+X unless <condition>"` grant to parse from the unless-stripped body and attach the **negated** condition (the grant applies precisely when the condition is false).
- Wired the condition through the layers / coverage / quantity / trigger / effect exhaustive sites.

## Tests
- **Parser regression**: the `+3/+3` carries `Not(SharesColorWithMostCommonColorAmongPermanents)`.
- **Runtime regression**: covers a strict majority (source shares vs. a non-sharing source returns false), a **tie** (every tied color counts), and a colorless source.

## Verification
`cargo fmt`, `cargo clippy -p engine --lib`, and the parser-combinator gate are clean. `oracle_static` (917), `game::conditions` (4), and `game::layers` (189) suites pass. Oracle-gen confirms Heroic Defiance now emits `AddPower 3 / AddToughness 3` gated on `Not(shares-most-common-color)`.

### DIFF
--- a/crates/engine/src/game/conditions.rs
+++ b/crates/engine/src/game/conditions.rs
@@ -177,6 +177,39 @@ pub(crate) fn eval_source_has_dealt_damage(state: &GameState, source_id: ObjectI
     state.objects_that_dealt_damage.contains(&source_id)
 }
 
+/// CR 105.2 + CR 611.3a: True when the source permanent shares a color with the
+/// most common color among all battlefield permanents — counting every color
+/// tied for most common. Backs
+/// `StaticCondition::SharesColorWithMostCommonColorAmongPermanents` (Heroic
+/// Defiance, whose static gate wraps it in `Not` for the "unless" clause).
+pub(crate) fn eval_shares_color_with_most_common_color(
+    state: &GameState,
+    source_id: ObjectId,
+) -> bool {
+    use crate::types::mana::ManaColor;
+    use std::collections::HashMap;
+
+    // CR 105.2a: only the five colors count; the histogram is over every colored
+    // battlefield permanent (the source itself included).
+    let mut counts: HashMap<ManaColor, usize> = HashMap::new();
+    for &id in crate::game::targeting::zone_object_ids(state, Zone::Battlefield).iter() {
+        if let Some(obj) = state.objects.get(&id) {
+            for color in &obj.color {
+                *counts.entry(*color).or_insert(0) += 1;
+            }
+        }
+    }
+    let Some(&max) = counts.values().max() else {
+        return false; // no colored permanent — there is no "most common color"
+    };
+    // A source color is most-common (or tied) iff its board-wide count equals the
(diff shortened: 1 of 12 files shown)


matthewevans (changes requested):
Current head `942acf7df2a802eb437153ecfef7097bdb649614` is not ready to merge.

**[HIGH] Heroic Defiance gates on the Aura's color, not the enchanted creature's color.** The parser emits an unscoped `SharesColorWithMostCommonColorAmongPermanents` condition, and the continuous-effect path evaluates non-recipient conditions before per-recipient filtering. Evidence: `crates/engine/src/game/layers.rs:2677` calls `source_condition_gate_passes(...)`; `crates/engine/src/game/layers.rs:1114` evaluates this condition via `eval_shares_color_with_most_common_color(state, source_id)`; `crates/engine/src/game/conditions.rs:185` defines the checked object as that source id. Heroic Defiance says "Enchanted creature gets +3/+3 unless it shares a color...", so the checked object has to be the enchanted creature/recipient, not the Aura source. Please model this as recipient-scoped condition evaluation and add a production layer test with an Aura attached to a differently colored creature.

**[HIGH] The new attached-subject `unless` split strips `unless` clauses out of granted ability text.** `crates/engine/src/parser/oracle_static/grammar.rs:686` splits any attached/equipped predicate on `" unless "` before quoted/granted ability classification. That breaks cards such as Sunken Field, whose granted activated ability contains ability-level text: "Counter target spell unless its controller pays {1}". The canonical quoted-inner classifier is at `crates/engine/src/parser/oracle_static/keyword_grant.rs:1558`, but this split can move the inner `unless` onto the static grant itself instead. Please only peel trailing static-grant `unless` after proving it is outside quoted/granted ability text, and add discriminating tests for Sunken Field/Coral Net-style granted abilities retaining their inner `unless`.

matthewevans (approved):
Reviewed current head `180cc5039b07b6cd9e8697ac8038affc6828a91d`. The prior recipient-scoping blocker is fixed via retained recipient-context condition evaluation, the quoted inner-unless regression is guarded, and exact-head CI is green.

PR #4592, opened 2026-06-29
fix(parser): UNSUPPORTED one-off: As Foretold

ntindle:
## Summary

**Root cause:** UNSUPPORTED one-off: As Foretold

## Cards corrected
- As Foretold

## Fix
Fixed As Foretold by lowering its pay-zero free-cast line to StaticMode CastFromHandFree OncePerTurn with a dynamic Cmc cap of time counters on the source. Parser-layer only.

## Files changed
- crates/engine/src/parser/oracle_static/restriction.rs
- crates/engine/src/parser/oracle_classifier.rs

## CR references
- CR 109.5
- CR 118.9
- CR 118.9a
- CR 122.1
- CR 202.3
- CR 601.2a
- CR 601.2b

## Verification
- `cargo fmt --all` — clean (no files changed, no commit needed)
- `./scripts/check-parser-combinators.sh <merge-base>` — clean (exit 0)
- `cargo clippy -p engine --all-targets -- -D warnings` — clean (exit 0; all 7 cluster-changed files are in the engine crate and pass)
- `cargo test -p engine` — clean (all tests passed, 0 failed)
- `cargo run --profile tool --features cli --bin oracle-gen -- data --filter "as foretold"` — confirmed: As Foretold parses correctly (Phase/Upkeep PutCounter time trigger + CastFromHandFree OncePerTurn static with Cmc LE CountersOn(time) condition); no Unimplemented/Unknown
Cards confirmed re-parsed correctly: As Foretold


matthewevans (changes requested):
**[MED] As Foretold is lowered onto a hand-only free-cast path even though the permission applies to any spell you cast.** Evidence: `crates/engine/src/parser/oracle_static/restriction.rs:2568` documents that the new `DefaultCastPermission` path only covers hand plus command-zone casts and explicitly excludes spells cast from graveyard/exile via a separate permission, while `crates/engine/src/game/casting.rs:7792` rejects `CastSpellForFree` unless the object is in `Zone::Hand`, and `crates/engine/src/game/casting.rs:10098` only enumerates `player_data.hand` candidates. Why it matters: As Foretold's text has no hand-zone restriction, so this can make the card appear supported while legal casts from exile/graveyard/other permission-created zones cannot use the `{0}` alternative cost. Suggested fix: model this as a general once-per-turn alternative-cost modifier that composes with every valid cast permission, or keep As Foretold coverage unsupported until the non-hand cast paths are represented and tested.

matthewevans (approved):
Reviewed the current head after the prior changes-requested review. This now keeps As Foretold honest as unsupported instead of lowering it onto the hand/default CastFromHandFree path, while preserving the upkeep Phase trigger and suppressing only duplicate swallow noise once Unimplemented already records the unsupported line. The supported CastFromHandFree sibling path remains covered, no hard-stop paths are touched, CR citations were verified locally, and exact-head CI is green.

PR #4596, opened 2026-06-29
fix(parser): Tamiyo flip return-transformed binds to source, not TriggeringSource (#4543)

### LINKED ISSUE #4543: [Card Bug] Tamiyo Flip keeps her in exile
### Card(s)

tamiyo inquisitive student

### Build/version

v.09.0 7c1c1cf

### Game mode

None

### Actual behavior

When tamiyo's effect triggers and resolves, the card instead of coming back on the battlefield as a planeswalker, it stays in exile

### Expected behavior

When tamiyo's ability triggers, and the card goes in exile, it should come back on the battlefield flipped, as a planeswalker with abilities

### Steps to reproduce

_No response_

_No response_

minion1227:
Closes #4543

## Problem

Tamiyo, Inquisitive Student — *"When you draw your third card in a turn,
exile Tamiyo, then return her to the battlefield transformed under her
owner's control."* — left Tamiyo permanently in exile. The exile step
worked; the "return her ... transformed" step moved nothing.

## Root cause

The exile clause correctly targets the source (`~` → `SelfRef`), but the
"return her" sub-clause's bare pronoun bound to `TriggeringSource`. The
trigger subject is the player (*"you draw …"*), so the source has no entry
in the `CardDrawn` event — `TriggeringSource` resolved to nothing and the
return `ChangeZone` had an empty target set.

The SelfRef-anaphor rewrite (CR 608.2c) that handles *"exile ~, then return
it"* was gated on a **typed** trigger subject (it covers Ajani, Nacatl
Pariah's *"Whenever one or more Cats die"* path). Tamiyo's player subject
makes that guard false, so the rewrite was skipped.

## Fix

Extend the rewrite to the transform-flip-from-exile class via a precise
structural predicate: a prior `ChangeZone → Exile` targeting `SelfRef`
followed by a **transformed** return (`ChangeZone → Battlefield` with
`enter_transformed`) binds the anaphor to the source regardless of subject
shape (CR 712.2).

The `enter_transformed` requirement deliberately scopes the change to the
flip class. It leaves untouched:
- **Cecil, Dark Knight** — *"untap ~. … transform it"* keeps `ParentTarget`.
- **Aetherling** — plain *"exile ~. return it"* keeps its existing binding.

Both were caught as regressions by the full suite when an earlier, broader
version of this fix dropped the guard outright; the structural predicate is
the corrected, minimal-blast-radius form.

## Tests

Added `tamiyo_flip_return_transformed_binds_to_self_ref`: asserts the exile
step is `ChangeZone → Exile (SelfRef)` and the return step is
`ChangeZone → Battlefield`, `enter_transformed`, target `SelfRef`.

The identical runtime AST (`exile ~ SelfRef → return SelfRef transformed`)
is already exercised end-to-end by the existing
`ajani_nacatl_pariah_returns_transformed_when_another_cat_dies` integration
test, which drives draw/trigger/resolve and asserts the card returns
transformed as a planeswalker.

## Verification

- `cargo fmt --all` — clean
- `cargo test -p engine --lib` — **14157 passed, 0 failed**
- Rebased onto latest `main`; `behind_by: 0`, auto-mergeable (no conflicts)

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -21482,7 +21482,45 @@ pub(crate) fn parse_effect_chain_ir(
         // *named clause*, not the trigger condition. The SelfRef prior-clause
         // test is structural over the parsed clause and its nested sub-ability
         // chain — no string heuristic.
-        if typed_trigger_subject
+        //
+        // CR 712.2 (issue #4543): the typed-subject branch covers the typed
+        // members of this class (Ajani). The transform-flip-from-exile cards
+        // (Tamiyo, Inquisitive Student: "When you draw your third card in a turn,
+        // exile Tamiyo, then return her to the battlefield transformed") have a
+        // PLAYER subject, so `typed_trigger_subject` is false and the bare "her"
+        // would default to `TriggeringSource` — but the CardDrawn event has no
+        // resolvable source, so the return fizzles and the permanent is stranded
+        // in exile. Detect that specific structural class — a prior "exile ~"
+        // (ChangeZone→Exile, SelfRef) followed by a TRANSFORMED return
+        // (ChangeZone→Battlefield, `enter_transformed`) — and bind to the source
+        // there too. The `enter_transformed` requirement keeps Cecil-style
+        // "untap ~. transform it" (ParentTarget) and plain "exile ~. return it"
+        // (Aetherling) untouched.
+        // Narrowly: only the BROKEN binding (`TriggeringSource`, which the
+        // CardDrawn/player-subject event cannot resolve — Tamiyo). Cards whose
+        // return already binds correctly (`TrackedSet`/`ParentTarget` — the
+        // saga and flip-walker classes like Azusa's Many Journeys, Jace, Vryn's
+        // Prodigy) are left untouched: they are bound by their own (working)
+        // rewrites and must not be re-pointed.
+        let exile_then_return_transformed = matches!(
+            &clause.effect,
+            Effect::ChangeZone {
+                destination: Zone::Battlefield,
+                target: TargetFilter::TriggeringSource,
+                enter_transformed: true,
+                ..
+            }
+        ) && clauses.last().is_some_and(|prev| {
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[MED]** The regression only asserts the lowered parser shape, not the production trigger behavior that failed. Evidence: `crates/engine/src/parser/oracle_trigger.rs:14521` parses the Tamiyo trigger and checks the return sub-ability target is `SelfRef`, but the PR adds no scenario/integration test under `crates/engine/tests` that draws the third card, resolves the trigger, and verifies Tamiyo returns transformed from exile. Why it matters: the bug was observable at resolution time as the permanent being stranded in exile; parser shape alone does not prove the card moves through the real draw-trigger/stack/zone-change pipeline. Suggested fix: add a runtime regression that starts Tamiyo on the battlefield, draws the third card, resolves the trigger, and fails if the return still resolves against `TriggeringSource` or an empty target set.

matthewevans (changes requested):
**[MED]** The current head still does not exercise the production path that failed. Evidence: `crates/engine/src/parser/oracle_trigger.rs:14521` adds a `parse_trigger_line` shape test for the Tamiyo text, and the PR diff still only changes `crates/engine/src/parser/oracle_effect/mod.rs` plus `crates/engine/src/parser/oracle_trigger.rs`; there is no scenario/integration test under `crates/engine/tests` that draws the third card, resolves the trigger, and verifies Tamiyo returns transformed from exile. Why it matters: the reported regression is runtime-visible as the source permanent being stranded in exile, and the maintainer bar requires a bug-fix PR to drive the real trigger/stack/zone-change pipeline with a discriminating test. Suggested fix: add a runtime regression that starts Tamiyo on the battlefield, draws the third card, resolves the trigger, and fails if the return still resolves against `TriggeringSource` or an empty target set.

matthewevans (approved):
Maintainer review: previous runtime-coverage blocker is resolved by the new draw-event/stack/zone-change regression, and the diff is scoped to the Tamiyo #4543 parser binding fix.

PR #4597, opened 2026-06-29
fix(parser): keep another leg-local in singular X or Y disjunctions (closes #4513)

### LINKED ISSUE #4513: Morkrut Necropod: sacrifice another creature or land missing FilterProp::Another
## Card

Morkrut Necropod

## Oracle

Menace

Whenever this creature attacks or blocks, sacrifice another creature or land.

## Problem

Cluster 33 (`FilterProp::Another` dropped): the sacrifice target filter for "another creature or land" omits `FilterProp::Another` on the creature disjunct, so the parser allows sacrificing the attacking/blocking creature itself.

Backlog reference: `docs/parser-misparse-backlog.md` §33.

## Expected

- Trigger mode `AttacksOrBlocks`.
- `Effect::Sacrifice` with `TargetFilter::Or([Typed(Creature + Another), Typed(Land)])`.
- Land disjunct must NOT carry `FilterProp::Another`.

GildardoDev:
## Summary

Morkrut Necropod's "sacrifice another creature or land" (#4513) parsed its sacrifice target with no source-exclusion on either leg, so the source could sacrifice ITSELF. The sacrifice-count grammar consumes "another " as a bare count of 1 and discards the exclusion, so the filter built from the remainder ("creature or land") never carried `FilterProp::Another`.

## Fix

- New typed signal `CountWord { SourceExclusion, Plain }` plus a sibling combinator `parse_count_expr_with_exclusion` in `oracle_util.rs`. The central `parse_count_expr` and its ~50 callers are untouched (no ripple).
- In the sacrifice imperative branch, when the count word was the source-exclusion "another", re-apply `FilterProp::Another` to every leg of an `Or` target. "Another" means "not this source permanent": it is required on any leg the source's type could match and is harmless (vacuous) on legs it cannot (a land is never the source creature). Marking only some legs would leave a self-sacrifice hole for a source that matches a non-first leg — e.g. Mukotai Soulripper, an artifact creature, in "sacrifice another creature or artifact".

## Scope

Single-type "sacrifice another `<type>`" (e.g. Disciple of Bolas) source-exclusion is intentionally left for a follow-up: applying it reduces a single-eligible mandatory sacrifice to one option, which routes it through the engine's auto-sacrifice fast-path, where a follow-on "that creature's power" reference is mis-resolved — a separate, pre-existing engine bug outside this parser change.

## Tests

- End-to-end Morkrut trigger test (`parse_trigger_line` to `Effect::Sacrifice` target): asserts `FilterProp::Another` on both the creature and land legs.
- Production-path integration regression: Morkrut made a creature-Land by Ashaya, Soul of the Wild, then attacks, asserting the source is NOT offered to its own "sacrifice another creature or land" trigger while the other eligible permanent still is. This guards the identity-vs-type case (a type-changing effect making the source match a non-first leg) that a per-leg-type exclusion would miss.

## Verification

`cargo fmt` clean; parser combinator gate exit 0; engine parser and `oracle_effect` suites green; clippy clean. The card-data parse-diff confirms the change is confined to "sacrifice another X or Y" disjunction cards (Morkrut Necropod, Mukotai Soulripper, the "creature or artifact" group, etc.), each now correctly excluding its own source on every leg.

Closes #4513

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -1306,10 +1306,18 @@ pub(super) fn parse_targeted_action_ast(
                 min_count,
             });
         }
-        let (count, after_count) = super::super::oracle_util::parse_count_expr(rest).unwrap_or((
-            crate::types::ability::QuantityExpr::Fixed { value: 1 },
-            rest,
-        ));
+        // Use the exclusion-aware sibling so a source-exclusion "another"
+        // ("sacrifice another creature or land") is reported as a typed
+        // `CountWord::SourceExclusion` instead of being silently consumed as a
+        // bare count of 1. `parse_count_expr` discards the word's identity, so
+        // without this the parsed target never regains `FilterProp::Another`
+        // and the source could sacrifice itself (Morkrut Necropod, #4513).
+        let (count, after_count, count_word) =
+            super::super::oracle_util::parse_count_expr_with_exclusion(rest).unwrap_or((
+                crate::types::ability::QuantityExpr::Fixed { value: 1 },
+                rest,
+                super::super::oracle_util::CountWord::Plain,
+            ));
         let (target_text, _) = super::strip_optional_target_prefix(after_count.trim_start());
         // Strip the "of their choice" / "of your choice" confirmation suffix —
         // CR 701.16b makes player choice the default, so the phrase is a no-op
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
**[MED]** The tests stop at `parse_type_phrase` and do not exercise the production Morkrut trigger/effect path the PR claims to fix. Evidence: `crates/engine/src/parser/oracle_target.rs:7620` and `crates/engine/src/parser/oracle_target.rs:7647` call `parse_type_phrase` directly; no full `parse_oracle_text`/trigger test verifies "Whenever ~ attacks/blocks, sacrifice another creature or land" lowers to a sacrifice target whose `Another` property is present only on the creature leg. Why it matters: #4513 is a card-level parser bug; helper tests can pass while trigger/effect parsing, sacrifice wiring, or generated card-data output still misrepresents the supported card. Suggested fix: add a full parser regression for Morkrut Necropod's attack/block trigger, or equivalent production entry, asserting the sacrifice target shape end-to-end.

matthewevans (changes requested):
**[HIGH]** The new leg-local redistribution can make the source a legal sacrifice under supported type-changing effects. Evidence: `crates/engine/src/parser/oracle_target.rs:2275` removes `FilterProp::Another` from the shared props for bare singular disjunctions, and `crates/engine/src/parser/oracle_target.rs:3316` only re-adds it when the sibling leg's primary type matches the first leg, so `another creature or land` leaves the land leg without source exclusion. At runtime `FilterProp::Another` is an identity check, not a type check (`crates/engine/src/game/filter.rs:3638`), and the repo already supports Ashaya making nontoken creatures into Forest lands while retaining their creature type (`crates/engine/tests/integration/ashaya_nontoken_lands.rs:21` and `:64`). Why it matters: Morkrut under Ashaya (or any similar type-adding effect) will match the land leg and can be offered as a sacrifice to its own "sacrifice another creature or land" trigger, which is exactly the self-sacrifice class this PR is trying to prevent. Suggested fix: keep `Another` on every leg for source identity exclusion, or make the parser carry enough source-type/provenance information to prove a leg can never match the source; add a production-path regression with Morkrut gaining Land (Ashaya-style) and assert the source is not a legal sacrifice choice.

GildardoDev:
Thanks, this is exactly right. Reworked to keep `FilterProp::Another` on every leg of the disjunction, so the source is excluded by identity regardless of which leg its type matches, the Ashaya creature-Land case you describe, and likewise an artifact creature in "sacrifice another creature or artifact". I also dropped the earlier `oracle_target` leg-local redistribution entirely, since it was the source of the hole.

Added the production-path regression you asked for: Morkrut made a creature-Land by Ashaya, asserting it is not offered to its own "sacrifice another creature or land" trigger while the other eligible permanent still is. Re-requesting review.

matthewevans (approved):
Maintainer review: current head fixes the Morkrut/Ashaya source-exclusion bug at the parser/filter seam, includes a discriminating runtime regression, and CI is green.

PR #4598, opened 2026-06-29
fix(parser): delayed "sacrifice it unless you pay" rides the delayed trigger (#4369)

### LINKED ISSUE #4369: Ashling the limitless — You can pay the cost to keep the creature on the battlefield successfully without having the ma…
Reported in Discord: https://discord.com/channels/1485498006781427802/1519524066292142261/1519524066292142261

**Thread:** Ashling the limitless
**Cards:** _none detected_
**Parser status:** no_card

## Summary
You can pay the cost to keep the creature on the battlefield successfully without having the mana to do so.

---
<sub>report_id: `discord:1519524066292142261:1519524066292142261:0` · discord: `1519524066292142261/1519524066292142261`</sub>

minion1227:
Closes #4369

## Problem

Ashling, the Limitless — *"Whenever you sacrifice a nontoken Elemental,
create a token that's a copy of it. ... At the beginning of your next end
step, sacrifice it unless you pay {W}{U}{B}{R}{G}."* — let you keep the
token without the mana: the 5-color cost was demanded when the token was
**created** (on the sacrifice trigger), not at the next end step on the
delayed sacrifice.

## Root cause

The trigger-level `extract_unless_pay_modifier` scanned the whole
multi-sentence effect, found the "unless" in the **delayed** end-step
sentence, and hoisted the cost onto the **parent** trigger — so the
payment (CR 118.12a) surfaced at the wrong time. The cost belongs to the
`CreateDelayedTrigger`'s inner sacrifice (CR 603.7a).

## Fix (parser-only, class-general)

1. `extract_unless_pay_modifier` declines to hoist when the "unless" lives
   in a sentence that **starts with a delayed temporal prefix** (reusing the
   canonical `strip_temporal_prefix` recognizer, widened to `pub(crate)`).
2. `parse_unless_alt_cost` gains *"you pay {explicit mana}"* and *"you pay an
   amount of {E} equal to …"* arms, so the per-clause resolution path can
   extract the delayed sub-clause's cost (it previously handled only
   `SelfManaCost` / life / non-mana shapes).
3. In `parse_effect_chain_ir`'s temporal-prefix branch, the chunk-level
   `unless_pay` is moved onto the delayed **inner** def (and consumed),
   instead of the outer `CreateDelayedTrigger` wrapper.

The runtime delayed-trigger payment path already exists
(`build_resolved_from_def` preserves `unless_pay`; `effects::mod` surfaces it
at resolution).

## Tests

Covers **both** members of the class:
- `ashling_delayed_end_step_unless_pay_rides_delayed_sacrifice` — mana
  `{W}{U}{B}{R}{G}`.
- `satya_delayed_end_step_energy_unless_rides_delayed_sacrifice` — Satya,
  Aetherflux Genius's dynamic energy *"an amount of {E} equal to its mana
  value"*.

Each asserts the parent trigger carries no `unless_pay` and the delayed
inner sacrifice carries the cost.

## Verification

- `cargo fmt --all` — clean
- `cargo test -p engine --lib` — **14158 passed, 0 failed**

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -4238,7 +4238,7 @@ pub(super) fn strip_temporal_suffix(text: &str) -> (&str, Option<DelayedTriggerC
 /// CR 603.7a: Strip temporal prefix indicating a delayed trigger condition.
 /// Symmetric to `strip_temporal_suffix` but handles prefix form:
 /// "At the beginning of the next end step, untap up to two lands."
-pub(super) fn strip_temporal_prefix(text: &str) -> (&str, Option<DelayedTriggerCondition>) {
+pub(crate) fn strip_temporal_prefix(text: &str) -> (&str, Option<DelayedTriggerCondition>) {
     let lower = text.to_lowercase();
     if let Some((condition, rest)) = nom_on_lower(text, &lower, |i| {
         alt((
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -20996,6 +20996,17 @@ pub(crate) fn parse_effect_chain_ir(
             if let Some(sub) = inner_clause.sub_ability {
                 inner_def.sub_ability = Some(sub);
             }
+            // CR 118.12a (issue #4369): a payment-unless on the delayed
+            // instruction itself ("...next end step, sacrifice it unless you pay
+            // {cost}" — Ashling, the Limitless; Satya, Aetherflux Genius) is
+            // extracted at chunk level into `unless_pay`, but it belongs to the
+            // DELAYED sacrifice, not the outer trigger. Move it onto the inner
+            // delayed def and consume it so it is not also applied to the outer
+            // CreateDelayedTrigger wrapper. The trigger-level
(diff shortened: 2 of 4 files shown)


matthewevans (changes requested):
**[MED] Runtime regression is only covered by parser-shape assertions.**

The new tests in `crates/engine/src/parser/oracle_trigger.rs` exercise `parse_trigger_line` and inspect that `CreateDelayedTrigger.unless_pay` is populated, but they do not drive the production game path that regressed: resolving the parent trigger, creating the token, confirming no payment prompt is surfaced immediately, advancing to the next end step, and then verifying the delayed sacrifice prompt plus pay/decline behavior.

That matters because the reported failure is a runtime timing bug, not just an AST placement bug. A parser-only check can pass while the delayed-trigger/stack/payment pipeline still asks at the wrong time or fails to sacrifice/pay correctly. Please add an engine scenario/integration regression for Ashling or Satya that proves the cost is offered only when the delayed trigger resolves and that both the paid and unpaid paths behave correctly.

matthewevans (approved):
Approved after current-head re-review. The prior runtime-test blocker is addressed: the new integration test drives Oracle parsing through sacrifice trigger creation, verifies no early UnlessPayment prompt at token creation, then verifies both decline and pay outcomes when the delayed end-step trigger resolves.

PR #4604, opened 2026-06-29
fix(parser): Phase-triggered token-copier's delayed "sacrifice it" binds the created token, not the source (#4601)

### LINKED ISSUE #4601: [Card Bug] "Mishra, Eminent One" sacrifices itself at the end step instead of the token it created
### Card(s)

Mishra, Eminent One

### Build/version

v0.10.0

### Game mode

None

### Actual behavior

After ending a turn in which Mishra, Eminent One copied an artifact with its ability, it will sacrifice itself and display pop-up asking to put him into graveyard or the command zone. The token also will force sacrifice if it's a legendary artifact that was copied which is incorrect (described in #4444 ). 

### Expected behavior

Mishra, Eminent One should not sacrifice itself because only the token created by it needs to be sacrificed. 

### Steps to reproduce

1. Cast Mishra, Eminent One
2. Cast an artifact (if none exists)
3. Go to combat and copy any valid artifact
4. Proceed to end step

[game-state-turn-9-2026-06-29T13-34-13-995Z.zip](https://github.com/user-attachments/files/29472275/game-state-turn-9-2026-06-29T13-34-13-995Z.zip)

minion1227:
Closes #4601

## Problem

**Mishra, Eminent One** —

> At the beginning of combat on your turn, create a token that's a copy of target noncreature artifact you control, except its name is Mishra's Warform and it's a 4/4 Construct artifact creature in addition to its other types. It gains haste until end of turn. **Sacrifice it at the beginning of the next end step.**

— sacrificed **Mishra itself** at the next end step instead of the token it created, and popped a "graveyard or command zone" prompt for the legendary commander. The temporary copy (the intended victim) was left on the battlefield.

## Root cause

The delayed "Sacrifice it" is a bare-pronoun anaphor whose antecedent is the just-created token. The populate-anaphor repair pass (`resolve_populated_token_anaphors` → `rewrite_populated_anaphor_in_effect`, in `parser/oracle_effect/lower.rs`) already rebinds that pronoun to `TargetFilter::LastCreated` — but only for the `ParentTarget` and `TriggeringSource` defaults, which is what an **ETB-triggered** token-copier produces (Flameshadow Conjuring, Inalla, Archmage Ritualist: "Whenever a creature enters … Exile it at the beginning of the next end step").

Mishra fires off a **Phase trigger** ("At the beginning of combat on your turn, …"), which has *no* triggering object. The imperative parser therefore lowers the bare "it" to `TargetFilter::SelfRef` (the source) — a case the rewrite never touched. So the delayed sacrifice stayed bound to Mishra (`SelfRef`) instead of the token (`LastCreated`).

## Fix (parser-only, class-general)

In the `CreateDelayedTrigger` arm of `rewrite_populated_anaphor_in_effect` — already gated on a preceding token-creating effect — also rebind a `SelfRef` cleanup target to `LastCreated`, via a new `rewrite_delayed_cleanup_self_ref_to_last_created` (the `SelfRef` companion to the existing `rewrite_parent_target_to_last_created`, mirroring the established `rewrite_oneshot_selfref_to_chosen` precedent).

Scope is deliberately limited to the **destructive cleanup** effects that remove/move the temporary token — `Sacrifice` / `Destroy` / `Bounce` / `ChangeZone`. `Pump` / `Attach` / `SetTapState` are intentionally excluded: there a delayed `SelfRef` ("~ gets +1/+1 until end of turn") more plausibly means the source, so it is left untouched.

This covers the Phase-triggered temporary-token class (Mishra, Eminent One and siblings), not just the one card.

CR 603.7c (delayed triggered ability) + CR 608.2c (read the whole text; resolve the anaphor) — both verified against `docs/MagicCompRules.txt`.

The separate legendary-token-sacrifice concern from #4444 is out of scope for this PR.

## Testing

- New parser regression `phase_trigger_token_copier_sacrifice_anaphor_binds_created_token` (modeled on the existing `etb_token_copier_exile_anaphor_binds_created_token`), asserting the delayed sacrifice binds `LastCreated`. Verified it **fails on `main`** (`Some(SelfRef)`) before the change and passes after.
- Full engine suite: **14157 passed / 0 failed**.
- `clippy` clean; `cargo fmt` clean; parser-combinator / engine-authority / test-card-data-load / skill-doc gates all pass.

Diff: 2 files, +78 (fix + regression test).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -2754,7 +2754,13 @@ pub(super) fn rewrite_parent_target_to_last_created(effect: &mut Effect) {
         }
         | Effect::Pump { target, .. }
         | Effect::Attach { target, .. }
-        | Effect::ChangeZone { target, .. } => {
+        | Effect::ChangeZone { target, .. }
+        // CR 603.7c + CR 608.2c (issue #4601 review): a delayed cleanup that
+        // puts the temporary token on top/bottom of a library ("… put it on the
+        // bottom of its owner's library at the beginning of the next end step")
+        // lowers its bare-"it" to `ParentTarget`/`TriggeringSource` just like the
+        // other move/cleanup forms — rebind to the created token.
+        | Effect::PutAtLibraryPosition { target, .. } => {
             // CR 603.7c + CR 608.2c: inside an ETB-triggered token-copier (e.g.
             // Flameshadow Conjuring / Inalla: "create a token that's a copy of
             // that creature. … Exile it at the beginning of the next end step"),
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[MED]** Delayed token cleanup still misses library-position cleanup effects. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:2795-2803` only rewrites delayed-cleanup `SelfRef` for `Sacrifice`, `Destroy`, `Bounce`, and `ChangeZone`; `Effect::PutAtLibraryPosition { target, .. }` is another cleanup/move effect with the same target-shape but is left bound to `SelfRef`. Why it matters: a temporary token cleanup that puts "it" on top/bottom of a library will still act on the source permanent instead of the token created earlier in the chain. Suggested fix: include `PutAtLibraryPosition { target, .. }` in `rewrite_delayed_cleanup_self_ref_to_last_created` and add a parser regression for that cleanup form.

matthewevans (approved):
Reviewed current head. The delayed cleanup rebinding now covers both sacrifice and library-position cleanup effects, with discriminating parser/runtime coverage and green CI.

PR #4607, opened 2026-06-29
fix(engine): BecomesBlocked "that creature" binds the blocked attacker, not the blocker (#4599)

### LINKED ISSUE #4599: She-Hulk, Wallbreaker does not trigger when Heroes become blocked
## Report

She-Hulk, Wallbreaker triggered ability is not triggering when a Hero you control becomes blocked, so it also does not apply the +1/+1 counter.

## Expected behavior

Whenever a Hero you control becomes blocked, the ability should trigger and put a +1/+1 counter on that Hero for each creature blocking it.

## Oracle text

Verified from the refreshed coverage export used by the bug classifier:

> Trample
> Other Heroes you control have trample.
> Whenever a Hero you control becomes blocked, put a +1/+1 counter on that Hero for each creature blocking it.

## Classifier analysis

**Matched clause**: Whenever a Hero you control becomes blocked, put a +1/+1 counter on that Hero for each creature blocking it. (supported: true)

**Reasoning**: The coverage export marks the combat trigger supported and parses it as a BecomesBlocked trigger watching a Hero you control. Its child is a PutCounter effect targeting the parent target, with the counter count represented as the number of creatures blocking the source. That AST matches the Oracle text, so the reported behavior points to a runtime defect in the BecomesBlocked trigger/event path, the parent-target binding, or counter application after the block event.

## Notes

- Local client/public/card-data.json is not present in this checkout, so Oracle text and parse details were verified from /tmp/coverage-data.json after refreshing the preview coverage export.
- Likely area to inspect: combat block event emission and matching for BecomesBlocked triggers, including propagation of the blocked Hero as the parent target and ObjectCount with BlockingSource.

minion1227:
Closes #4599

## Problem

**She-Hulk, Wallbreaker** —

> Trample
> Other Heroes you control have trample.
> **Whenever a Hero you control becomes blocked, put a +1/+1 counter on that Hero for each creature blocking it.**

— never applied the counter. The trigger fired correctly, but "that Hero"
resolved to the wrong object, so the `PutCounter` effect missed She-Hulk
entirely and no counter was placed.

## Root cause

"That Hero" lowers to `TargetFilter::ParentTarget`, whose block-event
resolution goes through `blocked_attacker_from_event` in
`crates/engine/src/game/targeting.rs`.

That helper only handled the **`Blocks`** trigger shape: it filtered the
`BlockersDeclared` assignments down to the pair whose **blocker** is the trigger
source ("Whenever ~ blocks a creature, … *that creature*" → the attacker the
source blocked). For a **`BecomesBlocked`** trigger the source is the
**attacker** (the blocked Hero), or an observer of it — never the blocker — so
the blocker-side filter matched nothing and the function returned `None`.
`ParentTarget` then fell through to the unrelated event arms
(Stationed / Crewed / ZoneChanged) and was never resolved, so the counter had no
subject.

(The trigger itself fires fine — `matching_becomes_blocked_events` handles the
`valid_card` match. The defect was purely the "that Hero" target binding.)

## Fix (additive, both trigger shapes)

When the blocker-side filter is empty, fall back to the attacker carried in the
matcher-narrowed single `(blocker, attacker)` assignment — that attacker is
exactly the blocked creature "that [creature]" refers to. The `Blocks` path is
unchanged (it returns early on a blocker match), so "Whenever ~ blocks … that
creature" keeps binding the attacker it blocked.

This covers the whole `BecomesBlocked` "that [creature]" class (counters, pumps,
etc.), not just She-Hulk.

- `crates/engine/src/game/targeting.rs` — `blocked_attacker_from_event` gains
  the attacker-side fallback.
- CR 509.1 (blocked status) + CR 608.2c (resolving "that"/"it" to the
  established object) — both verified against `docs/MagicCompRules.txt`.

## Verification

- New runtime combat regression
  (`she_hulk_wallbreaker_becomes_blocked_4599`): drives a real
  attack → declare-blocker sequence and asserts a `Plus1Plus1` counter lands on
  the blocked Hero (0 before the fix → 1 after).
- Full engine lib **14159 passed / 0 failed**; 49 block-trigger integration
  tests pass; `clippy` clean.

### DIFF
--- a/crates/engine/src/game/targeting.rs
+++ b/crates/engine/src/game/targeting.rs
@@ -955,9 +955,22 @@ fn blocked_attacker_from_event(
     let crate::types::events::GameEvent::BlockersDeclared { assignments } = event else {
         return None;
     };
-    let mut attackers = assignments
+    // CR 509.1 + CR 608.2c: For a `Blocks` trigger ("Whenever ~ blocks a
+    // creature, … that creature") the source is the BLOCKER, so "that creature"
+    // is the attacker it was assigned to.
+    let mut blocked = assignments
         .iter()
         .filter_map(|(blocker, attacker)| (*blocker == source_id).then_some(*attacker));
+    if let Some(first) = blocked.next() {
+        return blocked.all(|attacker| attacker == first).then_some(first);
+    }
+    // CR 509.1 + CR 608.2c (issue #4599): For a `BecomesBlocked` trigger
+    // ("Whenever a Hero you control becomes blocked, put a +1/+1 counter on that
+    // Hero …" — She-Hulk, Wallbreaker) the source is the ATTACKER (the blocked
+    // creature), or an observer of it, never the blocker — so the blocker-side
+    // filter above is empty. The matcher narrows the event to the single
+    // matched `(blocker, attacker)` pair, so "that [creature]" is that attacker.
+    let mut attackers = assignments.iter().map(|(_, attacker)| *attacker);
     let first = attackers.next()?;
     attackers.all(|attacker| attacker == first).then_some(first)
 }
--- a/crates/engine/tests/she_hulk_wallbreaker_becomes_blocked_4599.rs
+++ b/crates/engine/tests/she_hulk_wallbreaker_becomes_blocked_4599.rs
@@ -1,0 +1,101 @@
+//! Runtime regression for GitHub issue #4599 — She-Hulk, Wallbreaker's
+//! "becomes blocked" trigger.
+//!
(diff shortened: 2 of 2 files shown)


matthewevans (approved):
Reviewed the current head for the #4599 runtime path. The BecomesBlocked matcher narrows BlockersDeclared to a single matched assignment before resolution, ParentTarget now binds the blocked attacker for observer/self triggers without breaking existing Blocks semantics, and the new scenario test exercises the real combat/stack path that would fail on the prior behavior. Exact-head CI is green.

PR #4610, opened 2026-06-29
fix(parser): nested single-quoted token ability keeps its clause (Urza's Saga ch. II) (#4605)

### LINKED ISSUE #4605: Urza's Saga chapter II ability activates but does not create Construct
## Report

Urza's Saga chapter II granted activated ability is selectable, but activating it does not create the Construct token.

## Expected behavior

After chapter II resolves, Urza's Saga should have the activated ability {2}, {T}: create a 0/0 colorless Construct artifact creature token with the artifact-count power/toughness ability. Activating that ability should pay the cost, tap Urza's Saga, and create the Construct token.

## Oracle text

Local client/public/card-data.json is not present in this checkout. Oracle text was verified from the refreshed coverage export and the local generated integration fixture:

> (As this Saga enters and after your draw step, add a lore counter. Sacrifice after III.)
> I — This Saga gains "{T}: Add {C}."
> II — This Saga gains "{2}, {T}: Create a 0/0 colorless Construct artifact creature token with 'This token gets +1/+1 for each artifact you control.'"
> III — Search your library for an artifact card with mana cost {0} or {1}, put it onto the battlefield, then shuffle.

## Classifier analysis

**Matched clause**: II — This Saga gains "{2}, {T}: Create a 0/0 colorless Construct artifact creature token with 'This token gets +1/+1 for each artifact you control.'" (supported: true)

**Reasoning**: The coverage export marks chapter II supported as a CounterAdded chapter trigger that grants an ability to Urza's Saga. The local generated fixture shows the granted activated ability is structurally present with a Composite cost ({2} plus Tap) and a Token effect creating one 0/0 colorless Artifact Creature Construct with dynamic +1/+1 per artifact controlled. Since gameplay can select the granted ability but no Construct is created, the likely defect is in runtime execution of granted activated abilities or Token effect resolution from a granted ability, rather than missing parser support.

## Notes

- Existing closed issue #278 covers chapter III mana-cost filtering and is not a duplicate of this chapter II token-creation report.
- Likely areas to inspect: activation resolution for abilities granted by continuous effects, preservation of the granted ability definition through chapter II, and token creation from granted activated abilities on lands.

minion1227:
Closes #4605

## Problem

**Urza's Saga** chapter II grants `{2}, {T}: Create a 0/0 colorless Construct artifact creature token with 'This token gets +1/+1 for each artifact you control.'`. Activating it created **no token** — the granted ability's effect parsed as `Pump` (artifact-count power/toughness, target Any), i.e. the inner ability lifted out of the single-quoted span, instead of a token-creation effect.

## Root cause

The clause splitter (`split_clause_sequence`) ends a clause on a sentence-period whenever it is not inside a **double**-quoted span. Its comment assumed any open single-quote could only be a possessive/contraction apostrophe — true at the top level (Oracle text uses double quotes for quoted abilities), but **not** when a token-creation clause is itself nested inside a double-quoted granted ability. There the inner token ability genuinely uses **single** quotes (`with 'This token gets +1/+1 … control.'`). The `.` inside that span bisected the clause, dropped the closing quote, and broke token dispatch.

## Fix (class-general — all nested token grants: Sagas, ability-granting cards)

Track whether the open single-quote is a genuine nested ability quote, opened only at an anchored boundary (`with '` / `and '` / `or '` / `, '`) — the same anchors `extract_token_static_abilities` already uses. A sentence-period inside such a span no longer closes the clause. Possessive/contraction apostrophes (`~'s`, `can't`) are unaffected, so the existing `"~'s … turn. Goad it."` (The Master, Mesmerist) sentence split is preserved.

CR 111.1 (a token has the abilities defined in its creation effect), verified against `docs/MagicCompRules.txt`.

## Verification

- New regression `urzas_saga_chapter_two_granted_ability_creates_token` asserts the granted ability's effect is `Effect::Token`, not `Pump` (fails before, passes after).
- Full engine lib **14181 passed / 0 failed**; clippy clean.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -550,7 +550,7 @@ pub(super) fn split_clause_sequence(text: &str) -> Vec<ClauseChunk> {
     let mut chars = text.chars().peekable();
     let mut paren_depth = 0usize;
     let mut in_single_quote = false;
-    // CR 111.1 (issue #4605): a created token has the abilities defined in its
+    // CR 111.3 (issue #4605): a created token has the abilities defined in its
     // creation effect. Tracks whether the currently-open single-quoted span is a
     // genuine NESTED ability quote — the inner ability of a token/permanent
     // created inside a double-quoted granted ability (Urza's Saga:
@@ -650,7 +650,7 @@ pub(super) fn split_clause_sequence(text: &str) -> Vec<ClauseChunk> {
             // split here and reset the phantom single-quote state.
             // `in_double_quote` (a genuine quoted ability) still suppresses the
             // split.
-            // CR 111.1 (issue #4605): inside a genuine nested ability quote
+            // CR 111.3 (issue #4605): inside a genuine nested ability quote
             // (`single_quote_is_ability`) the `.` belongs to the quoted text
             // (`... with 'This token gets +1/+1 for each artifact you control.'`)
             // and must NOT close the clause — otherwise the span is bisected and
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Maintainer review: approved. Backend parser fix is scoped to the clause splitter seam, mirrors the existing token static single-quote anchors, preserves possessive/contraction handling, and has a discriminating saga parser regression that would fail if the sentence split regressed. CI is green on the exact head.

PR #4614, opened 2026-06-29
fix: honor hexproof in per-opponent fanout targeting

mike-theDude:


matthewevans (approved):
Reviewed the updated head after merging current `main`. Two passes found no blocking correctness issue in the hexproof/per-opponent fanout targeting fix. Gemini's comments were performance-only and not merge blockers for this PR.

PR #4621, opened 2026-06-29
fix(parser): "choose a creature you don't control" routes to an opponent target

minion1227:
## Problem

The parser handled the affirmative `choose a creature you control` (→ `TargetOnly { Creature, You }`) but lowered the **negated** form `choose a creature you don't control` to `Effect::Unimplemented`. Roughly **193 cards** carry the "you don't control" choose clause (e.g. *Sadistic Shell Game*, *Ticking Mime Bomb*), so the whole class fell through to the unimplemented fallback.

## Root cause

`is_choose_as_targeting` (`crates/engine/src/parser/oracle_effect/mod.rs`) decides whether a "choose ..." phrase is really a targeting instruction by scanning for affirmative controller suffixes:

```
scan_contains_phrase(x, "you control" / "they control" / "an opponent controls")
```

The string `"you don't control"` does **not** contain the substring `"you control"`, so the check returned `false`, the phrase matched none of the choose handlers, and it fell through to `Unimplemented`.

Notably, `parse_target` **already** resolves `"you don't control"` → `ControllerRef::Opponent` (via `parse_zone_controller` in `oracle_nom/filter.rs`). The targeting layer was correct; only the choose-as-targeting gate failed to recognize the negated controller phrase.

## Fix

Add the negated-controller arms (`"don't control"` / `"doesn't control"`) to **both** controller-reference checks in `is_choose_as_targeting` — the article path and the general fallback — mirroring the existing affirmative arms. The phrase now routes through the normal targeting path and resolves to `TargetOnly { Creature, Opponent }`.

Surgical change: **1 file, +51/−2**.

## Tests

Added regression `choose_creature_you_dont_control_routes_to_opponent_target`, asserting:
- `choose a creature you control` still resolves to `You` (no regression on the affirmative form),
- `choose a creature you don't control` resolves to `Opponent`,
- the third-person `they choose a creature you don't control` also resolves to `Opponent`.

## Verification

- Live parse probe confirmed `Unimplemented` → `TargetOnly { Opponent }`.
- Parser-combinator gate green; full engine lib suite green; clippy clean.

## Scope / coverage notes

This unlocks the controller-reference half of the class. Cards with **co-occurring** unsupported clauses are only **partially** unlocked — e.g. *Sadistic Shell Game* also carries "starting with the next opponent in turn order", which remains a separate gap. No `Closes #` link: this was surfaced via `cargo parser-gaps`, not a filed issue.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -16977,13 +16977,25 @@ fn is_choose_as_targeting(rest: &str) -> bool {
         // Must reference controller to be targeting-like.
         // "they control" covers "target opponent chooses a creature they control"
         // where "they" refers to the targeted player (CR 608.2d).
+        // CR 109.4 + CR 102.2: the NEGATED controller form "a creature you
+        // don't control" is equally a controller-scoped selection — only this
+        // exact phrase is admitted, because `parse_zone_controller`
+        // (`oracle_nom/filter.rs`) maps `"you don't control"` →
+        // `ControllerRef::Opponent` and nothing else. Broader negated forms
+        // ("they don't control", "an opponent doesn't control") are NOT
+        // consumed by `parse_target`, so admitting them here would route the
+        // choice as targeting-like while silently dropping the controller
+        // restriction — they stay out of the gate until the shared controller
+        // parser learns them. Without this arm "choose a creature you don't
+        // control" (Sadistic Shell Game, etc.) falls through to `Unimplemented`.
         // Exclude "from among" patterns (Cataclysm-family multi-category selection)
         // which require engine infrastructure not yet implemented.
         if !scan_contains_phrase(after_article, "from among")
             && (scan_contains_phrase(after_article, "you control")
                 || scan_contains_phrase(after_article, "opponent controls")
                 || scan_contains_phrase(after_article, "an opponent controls")
-                || scan_contains_phrase(after_article, "they control"))
+                || scan_contains_phrase(after_article, "they control")
+                || scan_contains_phrase(after_article, "you don't control"))
         {
             return true;
         }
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The new gate admits negated-controller forms that the target parser does not consume. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:16980` and `crates/engine/src/parser/oracle_effect/mod.rs:16995` admit any `don't control` / `doesn't control`, but `crates/engine/src/parser/oracle_nom/filter.rs:99` only maps `you don't control`, and `crates/engine/src/parser/oracle_effect/imperative.rs:3115` ignores the parse remainder after `parse_target(rest)`. Why it matters: phrases the comment explicitly claims, like `they don't control` / `an opponent doesn't control`, can now be routed as targeting-like and silently lose the controller restriction. Suggested fix: only admit exact negated-controller forms that `parse_target` can consume, or extend the shared controller parser for every admitted form and add tests that prove full consumption.

minion1227:
Thanks — good catch, and it was also the cause of the failing coverage / parser-honesty check. Fixed in `da924b0d0`.

The gate now admits only the exact negated form `parse_zone_controller` consumes — `"you don't control"` — at both sites (the article path and the general fallback). The broad `"don't control"` / `"doesn't control"` scans are gone, so `they don't control` / `an opponent doesn't control` are no longer routed as targeting-like and can't silently lose the controller restriction. The comment was corrected to state this explicitly, and a negative test asserts those unconsumable forms do **not** produce an `Opponent` target:

```rust
for unconsumed in [
    "choose a creature they don't control",
    "choose a creature an opponent doesn't control",
] {
    assert!(!opponent_filter(&parse_effect(unconsumed)), ...);
}
```

`choose_creature_you_dont_control_routes_to_opponent_target` (positive + the new negative cases) passes locally; also rebased onto current `main`.

matthewevans (approved):
Reviewed current head. Prior blocker is fixed: the negated-controller choose gate is now limited to the exact `you don't control` phrase consumed by the shared controller parser, and the regression coverage guards both the accepted form and the broader forms that must not be routed yet.

PR #4623, opened 2026-06-30
fix(parser): support 'any player controls X' static conditions

GildardoDev:
## Anchored on

Knight of Grace (verified Oracle text):

> First strike, hexproof from black
> Knight of Grace gets +1/+0 as long as any player controls a black permanent.

Knight of Malice:

> First strike, hexproof from white
> Knight of Malice gets +1/+0 as long as any player controls a white permanent.

## Problem

"as long as any player controls a [type]" static conditions parsed to `StaticCondition::Unrecognized`, so the conditional P/T buff never applied. Reproduced via `parse_inner_condition`: "you control a black permanent" and "an opponent controls a black permanent" both lower to `IsPresent`, but "any player controls a black permanent" yielded `Unrecognized` (likewise "an artifact", "a Forest", etc.).

## Root cause

`parse_you_control_a` (`crates/engine/src/parser/oracle_nom/condition.rs`) dispatched the control subject over only two arms: `you control ` → `ControllerRef::You` and `an opponent controls ` → `ControllerRef::Opponent`. "any player controls" had no arm and fell through to `Unrecognized`.

## Fix (class-general)

- The verb dispatch now yields `Option<ControllerRef>` and gains an `any player controls ` arm → `None`. Per CLAUDE.md, `Option<ControllerRef>` is the idiomatic way to express "any controller" rather than introducing a sentinel variant.
- For the `None` case, the filter gets battlefield presence with no controller, via a new `inject_battlefield_presence` helper that `inject_controller` now shares (the `InZone { Battlefield }` logic is no longer duplicated). At runtime a Typed filter with `controller: None` applies no controller restriction (`game/filter.rs`: `controller_ref_player` is consulted only for `Some(_)`), so a permanent controlled by ANY player satisfies the gate — exactly "any player controls".
- Covers the whole class: Knight of Grace, Knight of Malice, and any "as long as any player controls X" condition. The `you control` / `an opponent controls` / `you control no` siblings are byte-for-byte unchanged (regression-pinned).

CR 109.4 (only battlefield/stack objects have a controller, so presence is battlefield-scoped) + CR 611.3a (continuous "as long as" re-evaluation).

## Tests

- "any player controls a black permanent" → `IsPresent` with `controller: None` + `InZone { Battlefield }` + black color.
- Required-article guard: bare-plural "any player controls creatures" still rejected (a count, not a presence gate).
- Existing "you control a" / "an opponent controls a red permanent" parses unchanged.

## Verification

`cargo fmt` clean; parser combinator gate exit 0; full engine parser suite green (incl. 960 control-condition tests); `oracle_effect` suite green; clippy clean. The coverage parse-diff confirms "any player controls X" now lowers to `IsPresent` instead of `Unrecognized`.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -2959,24 +2959,34 @@ fn parse_creatures_are_attacking_count_ge(input: &str) -> OracleResult<'_, Stati
 }
 
 /// CR 109.5 (you = the controller of the object the ability is on) +
-/// CR 102.2 / CR 102.3 (an opponent of that player). Parse "you control
-/// a/an/another [type]" AND "an opponent controls a/an/another [type]" →
-/// `IsPresent` whose
-/// filter carries the matched `ControllerRef` (You / Opponent).
+/// CR 102.2 / CR 102.3 (an opponent of that player) + CR 109.4 (any player).
+/// Parse "you control a/an/another [type]", "an opponent controls a/an/another
+/// [type]", and "any player controls a/an/another [type]" → `IsPresent` whose
+/// filter carries the matched `ControllerRef` (You / Opponent) or, for "any
+/// player controls", no controller restriction at all.
 ///
 /// The verb is parameterized over the controller axis: the leading verb phrase
-/// selects `ControllerRef::You` or `ControllerRef::Opponent`, and the SAME
-/// downstream parse (required article, `parse_type_phrase`, full-consume) runs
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head after the merge from main. The change is at the existing `parse_you_control_a` control-condition seam, composes the existing verb/article/type-phrase parser path, keeps battlefield scoping for CR 109.4, and leaves the `you control` / `an opponent controls` siblings pinned. No blocking findings remain.

PR #4626, opened 2026-06-30
fix(engine): Heartwood Storyteller — each of the casting player's opponents may draw (#4361)

### LINKED ISSUE #4361: Heartwood Solo Stories — [[Heartwood Storyteller]] only allows its owner to draw cards, even when its owner casts spell…
Reported in Discord: https://discord.com/channels/1485498006781427802/1519294283952427108/1519294283952427108

**Thread:** Heartwood Solo Stories
**Cards:** heartwood storyteller

## Summary
[[Heartwood Storyteller]] only allows its owner to draw cards, even when its owner casts spells.

### heartwood storyteller
Whenever a player casts a noncreature spell, each of that player's opponents may draw a card.

---
<sub>report_id: `discord:1519294283952427108:1519294283952427108:0` · discord: `1519294283952427108/1519294283952427108`</sub>

minion1227:
## Problem

Heartwood Storyteller — *"Whenever a player casts a noncreature spell, each of that player's opponents may draw a card."* — had two defects:

1. **Wrong recipient.** The draw was routed to the source's controller, so only the Storyteller's controller drew, regardless of who cast the spell. The recipient must be *each opponent of the casting player* (the triggering player), which is **caster-relative**, not controller-relative — they differ whenever an opponent is the caster.
2. **Dropped "may".** The per-recipient optional was lost, so the draw was mandatory.

## Fix

Route the recipient through the standard `player_scope` fan-out with a new, caster-anchored player scope:

- **`PlayerFilter::OpponentOfTriggeringPlayer`** — each opponent of the player identified by `state.current_trigger_event` (the caster), resolved live via `extract_player_from_event`. Two-player `p.id != caster` (CR 102.2; teams CR 102.3 intentionally not modeled, matching the existing `Opponent` convention). Fails closed (no recipient) when no trigger event is in scope.
- **Parser (two nom arms, ASCII + curly apostrophe):**
  - `clause_shell::try_peel_opponent_may_prefix` — `"each of that player's opponents may "` → `optional` + scope (the per-recipient "may").
  - `oracle_effect/lower::strip_each_player_subject` — `"each of that player's opponents "` → scope (mandatory form).
- **Per-recipient "may" (CR 608.2d).** The optional rides `execute.optional`, so the `player_scope` fan-out prompts **each opponent independently** via `WaitingFor::OptionalEffectChoice` — mirroring the Tempting Offer cycle (Tempt with Bunnies). The trigger-level `optional` stays false (the controller does not decide for the opponents).
- Live resolver arms in `matches_player_scope`, `resolve_player_count`, and `speed_effects::players_for_filter` (kept in sync so a count can never desync from the recipient set); rules-justified fail-closed arms elsewhere; explicit arms at every exhaustive `PlayerFilter` site (no wildcard fallbacks).

## Tests

- **Runtime** (`tests/heartwood_storyteller_opponents_draw.rs`): in a 2-player game with the Storyteller under A — A casts a noncreature spell → **B** is prompted and draws (A unchanged); B casts → **A** is prompted and draws. Holding the controller constant at A while the recipient flips with the caster discriminates the caster anchor from the old controller-anchored bug. Plus a decline path (per-recipient "may") and the noncreature gate.
- **Parser shape** + a **building-block unit test** for `matches_player_scope` (event by B → A matches, B doesn't; no event → neither).

CR annotations (102.2 / 603.2 / 608.2d / 109.4) verified against the Comprehensive Rules.

Closes #4361

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -1616,6 +1616,7 @@ fn fmt_player_filter(pf: &PlayerFilter) -> String {
         PlayerFilter::OwnersOfCardsExiledBySource => "owners of cards exiled with source",
         PlayerFilter::TriggeringPlayer => "the triggering player",
         PlayerFilter::OpponentOtherThanTriggering => "each other opponent",
+        PlayerFilter::OpponentOfTriggeringPlayer => "each of that player's opponents",
         PlayerFilter::OpponentOfTriggeringPlayerNotAttacked => {
             "opponents of the attacking player who aren't being attacked"
         }
@@ -6570,6 +6571,7 @@ fn player_filter_feature(scope: &PlayerFilter) -> (&'static str, FeatureSupport)
         PlayerFilter::OwnersOfCardsExiledBySource => ("OwnersOfCardsExiledBySource", Handled),
         PlayerFilter::TriggeringPlayer => ("TriggeringPlayer", Handled),
         PlayerFilter::OpponentOtherThanTriggering => ("OpponentOtherThanTriggering", Handled),
+        PlayerFilter::OpponentOfTriggeringPlayer => ("OpponentOfTriggeringPlayer", Handled),
         // CR 506.2 + CR 508.6: count-only filter resolved by `resolve_player_count`
         // (Suppressor Skyguard's intervening-if). Handled like the other count filters.
         PlayerFilter::OpponentOfTriggeringPlayerNotAttacked => {
--- a/crates/engine/src/game/effects/deal_damage.rs
+++ b/crates/engine/src/game/effects/deal_damage.rs
@@ -1452,6 +1452,16 @@ fn collect_matching_players(
                         });
                         triggering != Some(p.id)
                     }
+                    // CR 102.2 + CR 603.2: Each opponent of the triggering
+                    // (casting) player, resolved live from the trigger event;
+                    // fail closed when no event is in scope. Mirrors the
+                    // recipient predicate in `matches_player_scope` so the
+                    // variant has one consistent meaning across all consumers.
+                    PlayerFilter::OpponentOfTriggeringPlayer => state
+                        .current_trigger_event
+                        .as_ref()
+                        .and_then(|e| crate::game::targeting::extract_player_from_event(e, state))
+                        .is_some_and(|caster| p.id != caster),
                     // CR 608.2c + CR 701.38: Match each player who cast a vote
                     // for the recorded choice index. Mirrors the
                     // `ZoneChangedThisWay` arm — consults the transient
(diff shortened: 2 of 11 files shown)


matthewevans (changes requested):
**[MED]** `OpponentOfTriggeringPlayer` is implemented in some player-scope consumers but hardcoded to no players in the damage fanout paths. Evidence: `crates/engine/src/game/effects/deal_damage.rs:1457` and `crates/engine/src/game/effects/deal_damage.rs:1664` both return `false`, while the same new filter is resolved from `state.current_trigger_event` in the other scope/count consumers. Why it matters: adding a global `PlayerFilter` variant creates a reusable recipient/count scope; leaving damage consumers fail-closed means damage effects using the same phrase will silently hit no players, and the implementation is not sibling-complete. Suggested fix: route these arms through the same `current_trigger_event` / `extract_player_from_event` logic as the other consumers, or centralize the player-filter predicate so the variant has one consistent meaning.

minion1227:
Fixed in `38bc0dba9` — thanks. Both `deal_damage.rs` arms (`damage_targets` for `EffectScope::All` and the `ability`-scoped path) now resolve `OpponentOfTriggeringPlayer` live from the trigger event, identical to `matches_player_scope` / `resolve_player_count` / `players_for_filter`:

```rust
PlayerFilter::OpponentOfTriggeringPlayer => state
    .current_trigger_event
    .as_ref()
    .and_then(|e| crate::game::targeting::extract_player_from_event(e, state))
    .is_some_and(|caster| p.id != caster),
```

So the variant now has one consistent meaning across every consumer — a future "deal damage to each of that player's opponents" card resolves correctly instead of hitting no one. I kept the predicate inlined per-site to match the established pattern of the sibling `OpponentOtherThanTriggering` (also inlined at each of these sites) rather than centralizing only this one variant. `cargo check -p engine` is clean.

matthewevans (changes requested):
**[MED]** `OpponentOfTriggeringPlayer` treats teammates as opponents in team formats. Evidence: `crates/engine/src/game/effects/mod.rs:363`; sibling consumers repeat the same `p.id != caster` predicate in `crates/engine/src/game/quantity.rs:4641`, `crates/engine/src/game/effects/deal_damage.rs:1464`, `crates/engine/src/game/effects/deal_damage.rs:1678`, and `crates/engine/src/game/effects/speed_effects.rs:173`, while the engine already has CR 102.3-aware `players::is_opponent` at `crates/engine/src/game/players.rs:95`. Why it matters: in Two-Headed Giant/team games, Heartwood-style "that player's opponents" incorrectly includes the caster's teammate, and this shared filter bakes that wrong scope into every fanout/count path. Suggested fix: resolve the caster from `current_trigger_event`, then use `crate::game::players::is_opponent(state, caster, p.id)` consistently and add a 2HG regression.

minion1227:
Addressed the review in 51f2f5027.

**[MED] teammates counted as opponents (2HG):** all five `OpponentOfTriggeringPlayer` consumers now route through the existing CR 102.3-aware `players::is_opponent(state, caster, p.id)` instead of the raw `p.id != caster` predicate — `matches_player_scope` (effects/mod.rs), both damage-fanout predicates (deal_damage.rs), the `PlayerCount` consumer (quantity.rs), and the `ChangeSpeed` recipient set (speed_effects.rs). This matches the sibling `OpponentOfTriggeringPlayerNotAttacked` arms. Added a 2HG regression (`opponent_of_triggering_player_excludes_two_headed_giant_teammate`) asserting the caster's teammate is excluded while the opposing team matches.

**[MED] damage fanout hardcoded `false`:** the two damage paths (deal_damage.rs:1460 / :1674) already resolve live from `current_trigger_event` on the current branch tip — they are no longer `false` stubs, and now share the same `is_opponent` predicate so the variant has one consistent meaning across every consumer.

`cargo clippy -p engine --lib` clean; the new 2HG unit test, the existing `#4361` filter unit test, and the Heartwood integration suite all pass locally.

matthewevans (approved):
Re-reviewed the current head after the two stale changes-request rounds. The damage fanout paths now resolve `OpponentOfTriggeringPlayer` from the trigger event instead of dropping the recipients, and all of the new scope consumers now use the existing `players::is_opponent` helper rather than treating teammates as opponents. The included 2HG regression covers that teammate exclusion, and the Heartwood Storyteller scenario exercises the intended per-opponent optional draw behavior.

There is a small follow-up-quality note: the enum doc comment still describes the old raw `p.id != caster` shape, but the runtime implementation and tests are now correct. That is not blocking this fix.

PR #4628, opened 2026-06-30
fix(engine): skip impossible targeted return after sacrificed token ceases to exist

### LINKED ISSUE #4571: Engine crash: object exists
**What happened**
I casted "it that betrays" and then cast "all is dust". All is dust makes the opponent sacrifice their creatures. And it that betrays takes all their sacrifice creatures and puts them on my battlefield. Nasty ik...

**Diagnostic**
```
Build: v0.9.0 (c57cc5e)
Diagnostic tag: getLegalActions-panic
Game id: 1c2ef566-0406-4449-9e70-d6b4727531b7
Game mode: ai
User agent: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Mobile Safari/537.36

Panic:
panicked at crates/engine/src/game/zones.rs:598:45: object exists
```

jonathanchang31:
## Summary
- skip impossible targeted `ChangeZone` moves when the referenced object no longer exists
- add a regression for `It That Betrays` + `All Is Dust` when the sacrificed set includes a token
- close the `object exists` panic path from issue `#4571`

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Tests
- [ ] Refactor

## Behavior Proof
- `process_one_zone_move` now returns `Done` immediately if the snapshotted `ObjectId` is no longer present in `state.objects`
- this preserves the existing `CR 400.7` / `CR 111.7` behavior: once the object is gone, the zone move is impossible and should no-op rather than panic
- the regression constructs the `It That Betrays` reanimation shape with a sacrificed colored token and asserts only the nontoken permanent returns

## Validation
- `cargo fmt --all`
- targeted regression is currently queued behind an unrelated active workspace build in this environment:
  - `cargo test -p engine all_is_dust_with_token_sacrifice_does_not_crash_it_that_betrays_return -- --nocapture`

Closes #4571.


jonathanchang31:
Addressed the redundant `state.objects` lookup review on this PR in commit `1932ad1a4`. `process_one_zone_move` now binds `state.objects.get(&obj_id)` once, uses that bound object for the emblem check, and reuses `obj.zone` for the origin read.

matthewevans (changes requested):
**[HIGH]** The PR's own regression test fails, so the claimed production behavior is not fixed. Evidence: `crates/engine/tests/issue_4571_it_that_betrays_all_is_dust.rs:87` drives the real cast pipeline, then `crates/engine/tests/issue_4571_it_that_betrays_all_is_dust.rs:89` asserts the sacrificed nontoken permanent returned; GitHub Actions shard 1 fails at that assertion on this head. Why it matters: the crash guard may avoid one missing-object panic path, but the It That Betrays + All Is Dust scenario still does not satisfy the rule behavior the PR claims to cover. Suggested fix: fix the trigger/continuation path so the nontoken permanent returns under P0 while the sacrificed token remains gone, then keep this regression test green.

matthewevans (changes requested):
**[HIGH]** The PR's own regression still fails at the exact head. Evidence: `crates/engine/tests/issue_4571_it_that_betrays_all_is_dust.rs:87` drives the cast pipeline and `:89` asserts the nontoken permanent returned; GitHub check runs for `f31b3938` show `Rust tests (shard 1/2)` failing. Why it matters: the missing-object no-op may avoid one panic, but the claimed It That Betrays / All Is Dust behavior is still not fixed. Suggested fix: fix the trigger/continuation path so the nontoken permanent returns under P0 while the sacrificed token remains gone, then keep this regression green.

Reviewed current head `f31b3938060c5c89d841d99be6f7db880d1434e7`.

matthewevans:
@jonathanchang31 — a quick status check on this PR.

Changes were requested on **2026-07-01** and there haven't been any updates to the branch or the discussion since. No pressure, and no judgment about the work — PRs just go stale, and we'd rather nudge than let one sit indefinitely.

**If you're still working on this**, comment or push anything at all and this notice clears automatically. Happy to re-review as soon as there's a new commit, and if the requested changes are unclear or you disagree with them, say so — pushback with reasoning is welcome and often right.

**If nothing changes in the next 7 days**, we'll close the PR to keep the review queue honest. Closing isn't rejection: the branch stays, and you (or anyone) can reopen or resubmit whenever you want to pick it back up.

Thanks for the contribution either way.

PR #4629, opened 2026-06-30
fix(parser): support 'no opponent controls X' static conditions

GildardoDev:
## Anchored on

Kavu Runner (verified Oracle text):

> This creature has haste as long as no opponent controls a white or blue creature.

Skittish Kavu:

> This creature gets +1/+1 as long as no opponent controls a white or blue creature.

## Problem

"as long as no opponent controls a [type]" static conditions parsed the condition to `StaticCondition::Unrecognized`, so the conditional keyword/buff never applied. The modification ("has haste" / "gets +1/+1") parsed fine; only the condition was unrecognized.

## Root cause

The control-condition family dispatcher in `crates/engine/src/parser/oracle_nom/condition.rs` handles "you control a/an X", "an opponent controls X", "any player controls X", "you control no X", and "you don't control a X", but had no arm for the negated-opponent-presence form "no opponent controls a X".

## Fix (class-general)

Added `parse_no_opponent_controls_a` to the control-condition `alt()`: "no opponent controls a/an [type]" → `Not(IsPresent { filter, controller: Opponent })`. It mirrors `parse_you_dont_control_a` with the opponent controller axis, reusing `parse_type_phrase` (which already handles the "white or blue creature" color disjunction), `inject_controller(.., Opponent)`, and the `Not(IsPresent)` wrapper. The required-article guard rejects the bare-plural count form. Covers Kavu Runner, Skittish Kavu, and any "no opponent controls X" condition.

CR 102.2 / CR 102.3 (an opponent of that player) + CR 109.4 (battlefield-scoped controller presence) + CR 611.3a (continuous "as long as" re-evaluation).

## Tests

- "no opponent controls a creature" → `Not(IsPresent)` with `controller: Opponent` + `InZone { Battlefield }` + creature.
- "no opponent controls a white or blue creature" (Kavu Runner / Skittish Kavu) → parses to `Not(...)` instead of the `Unrecognized` fallthrough.
- Required-article guard: bare-plural "no opponent controls creatures" rejected.

## Verification

`cargo fmt` clean; parser combinator gate exit 0; engine parser and `oracle_effect` suites green; clippy clean.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -2728,6 +2728,8 @@ pub(crate) fn parse_control_conditions(input: &str) -> OracleResult<'_, StaticCo
         parse_you_dont_control_a,
         // "you control no [type]" → Not(IsPresent)
         parse_you_control_no,
+        // "no opponent controls a/an [type]" → Not(IsPresent { Opponent })
+        parse_no_opponent_controls_a,
         // CR 702: "a creature you control has <keyword>" — subject-first
         // presence check (Odric, Lunarch Marshal). Grouped into the control
         // family so the parent dispatcher's `alt` arity stays within bounds.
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head. This is a narrow parser-condition fix at the right seam: `no opponent controls ...` now composes through the existing control-condition parser family with regression coverage for the affected static-condition class, and the prior CR citation concern is resolved.

PR #4631, opened 2026-06-30
fix(engine): emit ZoneChanged for Incubate so ETB triggers fire

### LINKED ISSUE #4238: Altar of the Brood bug — Incubating doesn't trigger Altar of the Brood, even though it creates a token (i.
Reported in Discord: https://discord.com/channels/1485498006781427802/1518595329279262842/1518595329279262842

**Thread:** Altar of the Brood bug
**Cards:** altar of the brood

## Summary
Incubating doesn't trigger Altar of the Brood, even though it creates a token (i.

### altar of the brood
Whenever another permanent you control enters, each opponent mills a card.

---
<sub>report_id: `discord:1518595329279262842:1518595329279262842:0` · discord: `1518595329279262842/1518595329279262842`</sub>

claytonlin1110:
## Summary
- Fixes #4238 — Incubate's token creation never emitted `GameEvent::ZoneChanged`, so no "permanent enters the battlefield" trigger (Altar of the Brood, Soul Warden, Panharmonicon, etc.) ever fired for an incubated token.
- `incubate.rs` was creating its token via the low-level `zones::create_object` allocator and skipping the battlefield-entry bookkeeping (`mark_layers_entered`, `record_battlefield_entry`, `record_token_created`, and the `ZoneChanged` event) that every other token-creation path performs.
- This is a known bug class — `conjure.rs` and `gift_delivery.rs` already carry the identical fix for the same root cause; `incubate.rs` was the one production path that still skipped it. Audited the remaining `create_object` call sites (`manifest.rs`, `populate.rs`, `create_emblem.rs`) and confirmed none have the same bug in production code.

## Test plan
- [x] Added `incubate_records_zone_change_for_etb_triggers` — verifies the `ZoneChanged` event and `zone_changes_this_turn` record are emitted (mirrors `conjure.rs`'s existing test).
- [x] Added `incubate_token_fires_another_permanent_enters_trigger` — end-to-end regression test that builds an Altar-of-the-Brood-style permanent via `GameScenario`, resolves Incubate, and asserts the trigger is queued on the stack with the correct `Mill` effect.
- [x] `cargo clippy -p engine --lib --tests -- -D warnings` clean.
- [x] `cargo test -p engine --lib` — 14,247 passed, 0 failed.

### DIFF
--- a/crates/engine/src/game/effects/counters.rs
+++ b/crates/engine/src/game/effects/counters.rs
@@ -391,6 +391,28 @@ fn apply_pending_counter_post_action(
             crate::game::layers::mark_layers_entered(state, object_id);
             crate::game::restrictions::record_battlefield_entry(state, object_id);
             crate::game::restrictions::record_token_created(state, object_id);
+            // CR 603.6a: finalize the deferred ZoneChanged here, once the
+            // token's counters have actually settled, so ETB trigger
+            // observers (Altar of the Brood, Soul Warden, etc.) see the
+            // Incubator's final counter count rather than firing early on a
+            // pre-replacement-choice snapshot (issue #4238).
+            if let Some(zone_change_record) = state.objects.get(&object_id).map(|obj| {
+                obj.snapshot_for_zone_change(
+                    object_id,
+                    None,
+                    crate::types::zones::Zone::Battlefield,
+                )
+            }) {
+                state
+                    .zone_changes_this_turn
+                    .push(zone_change_record.clone());
+                events.push(GameEvent::ZoneChanged {
+                    object_id,
+                    from: None,
+                    to: crate::types::zones::Zone::Battlefield,
+                    record: Box::new(zone_change_record),
+                });
+            }
             true
         }
         PendingCounterPostAction::FinalizeTokenEntry {
--- a/crates/engine/src/game/effects/incubate.rs
+++ b/crates/engine/src/game/effects/incubate.rs
@@ -64,6 +64,36 @@ pub fn resolve(
         obj.reset_for_battlefield_entry(state.turn_number, entry_timestamp);
     }
 
+    // Battlefield entry: incremental re-derive candidate for this Incubator
+    // token (escalates to a full pass if it sources effects, carries
+    // counters, etc.).
+    crate::game::layers::mark_layers_entered(state, obj_id);
+    crate::game::restrictions::record_battlefield_entry(state, obj_id);
+    crate::game::restrictions::record_token_created(state, obj_id);
+
+    // CR 603.6a: The Incubator token enters the battlefield as a zone change
+    // from outside the game (`from: None`) — emit `ZoneChanged` so every ETB
+    // trigger matcher (Altar of the Brood's "another permanent you control
+    // enters", Soul Warden, Panharmonicon, etc.) fires for it through the
+    // same code path used for normal token creation. Without this the
+    // Incubator enters silently and no ETB ability ever triggers (issue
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
Thanks for fixing the missing ETB event path here. I found one ordering issue that needs to be corrected before this can land.

**[MED] Emit the Incubate ETB event only after its entry counters finish.** In `crates/engine/src/game/effects/incubate.rs`, the new battlefield-entry bookkeeping and `ZoneChanged` event happen before `add_counter_with_replacement(...)` applies the Incubate +1/+1 counters. The normal token path applies `enter_with_counters` first, then records battlefield/token-created bookkeeping and emits `ZoneChanged`/`TokenCreated`; if counter replacement pauses, it defers `FinalizeTokenEntry` until after the counter choice resolves. With the current Incubate ordering, the `ZoneChanged` snapshot and `zone_changes_this_turn` record do not include the counters that the token entered with, and a replacement pause can expose the ETB event before the replacement-processed counters are complete. Please move the Incubate entry-event/bookkeeping block after the counter path, and make the pending-counter continuation finalize the same entry event/bookkeeping rather than emitting it early.

Evidence: this PR emits the Incubate `ZoneChanged` before `add_counter_with_replacement`; the existing token implementation applies `spec.enter_with_counters` before entry events and uses `PendingCounterPostAction::FinalizeTokenEntry` for paused counter replacement. Confidence: high.

claytonlin1110:
Fixed in b16eb6ce6 — thanks for catching this.

Moved the bookkeeping (`mark_layers_entered`/`record_battlefield_entry`/`record_token_created`) and `ZoneChanged` event block to *after* the `add_counter_with_replacement(...)` check, mirroring `token.rs`'s canonical `counters → bookkeeping → ZoneChanged` order.

For the paused-replacement case (`add_counter_with_replacement` returns `false`), `incubate.rs` already stashed a `PendingCounterPostAction::InjectPredefinedTokenAbilities` continuation that runs once the counter choice resolves — but its consumer in `counters.rs` (`apply_pending_counter_post_action`) only did the bookkeeping calls and never emitted `ZoneChanged` at all. So the bug class was actually twofold: the immediate path fired the event too early, and the paused path never fired it. Both are fixed now:

- `incubate.rs`: counters apply first; bookkeeping + `ZoneChanged` (with the post-replacement counter count already on the object) only run after.
- `counters.rs`'s `InjectPredefinedTokenAbilities` arm now also emits the deferred `ZoneChanged` once the pending counter additions drain.

Added `incubate_paused_counter_replacement_defers_zone_changed_until_resolved`, which forces the CR 616.1 ordering prompt with Doubling Season + Hardened Scales on a single `AddCounter`, asserts no `ZoneChanged` leaks before the choice resolves, and asserts the eventual snapshot carries the post-replacement counter count. Full suite: `cargo clippy -p engine --lib --tests -- -D warnings` clean, `cargo test -p engine --lib` 14,253 passed.

matthewevans (approved):
Re-reviewed the current head after the Incubate ordering fix and the update-branch merge. The immediate path now applies the Incubate counters through the replacement pipeline before battlefield-entry bookkeeping and `ZoneChanged`; the paused replacement path defers that same entry event until `drain_pending_counter_additions` runs the `InjectPredefinedTokenAbilities` post-action. The new regression covers the important failure mode: no `ZoneChanged` leaks before the CR 616.1 counter-ordering choice resolves, and the eventual snapshot observes the final counter count.

I also verified the update-branch merge locally: the new head contains the reviewed commit plus `origin/main`, and the current diff against `origin/main` is limited to `counters.rs` and `incubate.rs`.

PR #4634, opened 2026-06-30
fix(engine): delayed zone-change effects no-op on objects that ceased to exist (#3883)

### LINKED ISSUE #3883: kari zev, skyship raider — when she attacks, she creates ragavan.
Reported in Discord: https://discord.com/channels/1485498006781427802/1517625483506749591/1517625483506749591

**Thread:** kari zev, skyship raider
**Cards:** the end
**Parser status:** has_gaps

## Summary
when she attacks, she creates ragavan.

### the end

---
<sub>report_id: `discord:1517625483506749591:1517625483506749591:0` · discord: `1517625483506749591/1517625483506749591`</sub>

claytonlin1110:
## Summary
- Fixes #3883: resolving Kari Zev, Skyship Raider's delayed "exile that token at end of combat" trigger crashed the engine if the created Ragavan token was blocked and killed before the trigger fired.
- Root cause: the delayed trigger snapshots the token's `ObjectId` at creation. Once the token dies, CR 704.5d removes it from `state.objects` entirely (tokens that leave the battlefield cease to exist). The zone-move pipeline then hit a bare `.expect("object exists")` panic in `move_to_zone` ([`zones.rs:598`](crates/engine/src/game/zones.rs#L598)).
- `process_one_zone_move` in `crates/engine/src/game/effects/change_zone.rs` now treats a target object that's no longer in `state.objects` as a silent no-op (CR 400.7: a zone change can't happen to an object that's left the game). This covers any delayed or queued zone-change effect with a stale snapshot target, not just this card.

## Test plan
- [x] Added `kari_zev_ragavan_dying_in_combat_does_not_crash_delayed_exile` in `crates/engine/tests/integration/issue_3883_kari_zev_ragavan.rs`, which declares Kari Zev as an attacker, lets the attack trigger create Ragavan, blocks and kills Ragavan in combat, then drives the delayed exile trigger to resolution.
- [x] Verified the new test reproduces the exact reported panic (`zones.rs:598:45 "object exists"`) when the fix is reverted.
- [x] `cargo test -p engine --test integration issue_3883` passes with the fix in place (both the pre-existing token-creation test and the new crash regression test).
- [x] `cargo fmt --all`

### DIFF
--- a/crates/engine/src/game/effects/change_zone.rs
+++ b/crates/engine/src/game/effects/change_zone.rs
@@ -778,6 +778,19 @@ pub(crate) fn process_one_zone_move(
     obj_id: ObjectId,
     events: &mut Vec<GameEvent>,
 ) -> ZoneMoveResult {
+    // CR 400.7 + CR 111.7: An object that has already left the game (a token
+    // that ceased to exist via CR 704.5d, or any other object removed from
+    // `state.objects` since this effect snapshotted it as a target) cannot
+    // undergo a further zone change — skip silently rather than falling
+    // through to `move_to_zone`, which assumes the object is live and panics
+    // otherwise. A delayed trigger that snapshots `TargetFilter::LastCreated`
+    // (e.g. Kari Zev, Skyship Raider's "exile that token at end of combat")
+    // is the common path here: the token dies in combat before the delayed
+    // trigger fires.
+    if !state.objects.contains_key(&obj_id) {
+        return ZoneMoveResult::Done;
+    }
+
     // CR 114.5: Emblems cannot be moved between zones.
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Reviewed the current head after update-branch. The fix is at the right seam: `process_one_zone_move` is the shared ChangeZone iteration gate before `move_to_zone`, so it is the appropriate place to turn an already-ceased object into a no-op rather than letting a delayed zone-change effect reach `move_to_zone`'s live-object assumptions.

The regression drives the real Kari Zev delayed-trigger path: Ragavan is created, blocked, killed, removed from `state.objects`, and then the delayed end-of-combat exile resolves without panicking. Gemini's lookup-style note is non-blocking for this small fix.

PR #4636, opened 2026-06-30
fix(engine): Three Blind Mice Chapter II/III never occur (#2004)

### LINKED ISSUE #2004: Three Blind Mice - saga chapters II and III don't occur — Three Blind Mice Steps to Reproduce: Cast Three Blind Mice.
Reported in Discord: https://discord.com/channels/1485498006781427802/1511077817960628298/1511077817960628298

**Thread:** Three Blind Mice - saga chapters II and III don't occur
**Cards:** three blind mice
**Parser status:** has_gaps

## Summary
Three Blind Mice
Steps to Reproduce:
Cast Three Blind Mice.

### three blind mice
(As this Saga enters and after your draw step, add a lore counter. Sacrifice after IV.)
I — Create a 1/1 white Mouse creature token.
II, III — Create a token that's a copy of target token you control.
IV — Creatures you control get +1/+1 and gain vigilance until end of turn.

---
<sub>report_id: `discord:1511077817960628298:1511077817960628298:0` · discord: `1511077817960628298/1511077817960628298`</sub>

claytonlin1110:
## Summary

Fixes #2004 — Three Blind Mice's Chapter II/III ("Create a token that's a copy of target token you control") never occurred.

- **Root cause**: "target token you control" parses to `TargetFilter::Typed { type_filters: [], controller: Some(You), properties: [Token] }` — `type_filters` is correctly empty because a token can be any permanent type. But the legal-target enumerator (`find_legal_targets_with_context` in `targeting.rs`) treated *any* `Typed` filter with empty `type_filters` as the "target opponent"-shaped player-only filter, ignoring `properties` entirely. So the only "legal target" the ability ever saw was the controller player, not the tokens on the battlefield — `CopyTokenOf` then found no object to copy and silently no-opped.
- **Fix**: require `properties` to also be empty before collapsing a `Typed` filter to player-only target enumeration (mirroring the existing, correct guard in `TargetFilter::chosen_player_index`). This is a class-level fix — any "target token [you control]"-shaped filter, not just this card, now falls through to real object enumeration.
- Saga chapter parsing itself (`oracle_saga.rs`) and the `CounterAdded` threshold-crossing trigger matcher were independently verified correct via existing tests and were not the source of the bug.

## Test plan

- [x] Added `crates/engine/tests/three_blind_mice.rs` — drives a real two-turn game (CR 714.3b turn-based lore-counter addition) and asserts Chapter II's `CopyTokenOf` trigger reaches the stack with the correct object target and resolves into a second Mouse token.
- [x] `cargo test -p engine --lib` — 14,262 passed, 0 failed
- [x] `cargo test -p engine --tests` — full integration suite green, no regressions
- [x] `cargo fmt --all`
- [x] `cargo clippy -p engine --lib --tests -- -D warnings` — clean

### DIFF
--- a/crates/engine/src/game/targeting.rs
+++ b/crates/engine/src/game/targeting.rs
@@ -163,13 +163,19 @@ fn find_legal_targets_with_context(
         return targets;
     }
 
-    // Typed filter with no type_filters targets players, not permanents.
-    // e.g. "target opponent" → Typed { type_filters: [], controller: Opponent }
-    // The "any other target" shape (handled above as `is_any_other_target`) is
-    // the sole exception: it adds players above and falls through to the object
+    // Typed filter with no type_filters AND no properties targets players, not
+    // permanents. e.g. "target opponent" → Typed { type_filters: [], controller:
+    // Opponent }. A non-empty `properties` list (e.g. `FilterProp::Token` for
+    // "target token you control") describes an object characteristic that has
+    // no meaning for a player, so it must fall through to the object
+    // enumeration below instead of collapsing to players-only here (issue #2004
+    // — "target token you control" was wrongly resolving to the controller
+    // player instead of enumerating tokens). The "any other target" shape
+    // (handled above as `is_any_other_target`) is the sole property-bearing
+    // exception: it adds players above and falls through to the object
     // enumeration below instead of collapsing to players-only here.
     if let TargetFilter::Typed(ref tf) = filter {
-        if tf.type_filters.is_empty() && !is_any_other_target {
+        if tf.type_filters.is_empty() && tf.properties.is_empty() && !is_any_other_target {
             let controller = &tf.controller;
             for player in &state.players {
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
Thanks for the focused fix. The targeting change itself matches the bug: a typed filter with object properties like `Token` should not be treated as the player-only empty-typed shape used by `target opponent`.

One required fix before this can land: the new regression test cites `CR 714.3b` for the turn-based action that adds lore counters during the controller's precombat main phase, but the current Comprehensive Rules text has that action at `CR 714.3c`. `CR 714.3b` is the Read Ahead replacement-effect rule, so the annotation is misleading in both places it appears in the test comments. Please update those comments to `CR 714.3c`.

matthewevans (approved):
Approved after re-review: the CR 714.3c citation is corrected, the targeting fix is at the legal-target enumeration seam, and the Three Blind Mice regression drives the runtime Saga/stack path that would fail if the properties guard were reverted.

PR #4637, opened 2026-06-30
fix(engine): Dread Return into empty library breaks Thassa's Oracle win condition

### LINKED ISSUE #1365: Dread return used on thassa's oracle when no cards in lib — So i used Hermit druid to put my deck in the grave, then us…
Reported in Discord: https://discord.com/channels/1485498006781427802/1509563314605264916/1509563314605264916

**Thread:** Dread return used on thassa's oracle when no cards in lib
**Cards:** dread return, hermit druid, thassa's oracle
**Parser status:** has_gaps

## Summary
So i used Hermit druid to put my deck in the grave, then used dread return on Thassa's Oracle to get a win, it resolved and thassa's looked to be put

### dread return
Return target creature card from your graveyard to the battlefield.
Flashback—Sacrifice three creatures. (You may cast this card from your graveyard for its flashback cost. Then exile it.)
### hermit druid
{G}, {T}: Reveal cards from the top of your library until you reveal a basic land card. Put that card into your hand and all other cards revealed this way into your graveyard.
### thassa's oracle
When this creature enters, look at the top X cards of your library, where X is your devotion to blue. Put up to one of them on top of your library and the rest on the bottom of your library in a random order. If X is greater than or equal to the number of cards in your library, you win the game. (Each {U} in the mana costs of permanents you control counts toward your devotion to blue.)

---
<sub>report_id: `discord:1509563314605264916:1509563314605264916:0` · discord: `1509563314605264916/1509563314605264916`</sub>

jsdevninja:
## Summary
- Fixes #1365 — Thassa's Oracle reanimated with an empty library (e.g. Hermit Druid mills the deck, then Dread Return brings it back) silently failed its win condition instead of winning the game.
- Root cause: the ETB chain is `Dig` (look at top X) → `PutAtLibraryPosition` (put up to one on top, rest on bottom) → `WinTheGame` (devotion >= library size). With an empty library, `Dig` correctly looks at zero cards, but the chained `PutAtLibraryPosition` still ran with `target: ParentTarget` and no selected cards — the engine's generic "no parent target found" self-fallback (`targeting.rs`) resolved that to Thassa's Oracle itself, moving it from the battlefield onto its own library. That corrupted both its devotion contribution and the library count *before* `WinTheGame`'s condition was evaluated, turning a true `1 >= 0` win into a false `0 >= 1` non-win.
- Fix: a transient `last_dig_found_nothing` flag is set whenever a `Dig` (or its `PriorLook` variant) genuinely looks at zero cards, and cleared at every fresh chain resolution. A chained `ParentTarget` consumer now resolves to no target instead of self when this flag is set — a class-level fix for any "look at X, put some/the rest somewhere" card shape, not a Thassa's-Oracle special case. Every other self-fallback use (e.g. Avenging Angel's LTB self-return) is unaffected.

## Test plan
- [x] New runtime regression test (`thassas_oracle_dread_return_empty_library_still_wins`) drives the real parser output through the full ability chain with an empty library; fails pre-fix (Oracle ends up in the library, no win) and passes post-fix.
- [x] All existing `dig.rs` tests (24), `put_on_top.rs` tests (13, including the Avenging Angel self-fallback cases), and existing Thassa's Oracle tests (3) pass unchanged.
- [x] Full engine suite: 14,263 passed, 0 failed.
- [x] `cargo fmt --all` — no changes needed.
- [x] `cargo clippy -p engine --lib --tests -- -D warnings` — clean.

### DIFF
--- a/crates/engine/src/game/effects/additional_phase.rs
+++ b/crates/engine/src/game/effects/additional_phase.rs
@@ -196,6 +196,7 @@ mod tests {
             sub_link: crate::types::ability::SubAbilityLink::ContinuationStep,
             modal: None,
             mode_abilities: vec![],
+            dig_found_nothing_for_parent_target: false,
         }
     }
 
--- a/crates/engine/src/game/effects/dig.rs
+++ b/crates/engine/src/game/effects/dig.rs
@@ -76,6 +76,12 @@ pub fn resolve(
 
     let library_owner = super::resolve_player_for_context_ref(state, ability, library_owner_filter);
 
+    // CR 401.5 + CR 608.2c: This Dig's own outcome — not a stale value from an
+    // earlier link in the same chain — decides whether a chained `ParentTarget`
+    // consumer may self-fallback. Reset here; the two "found nothing" returns
+    // below (and in `resolve_from_prior_look`) set it back to `true`.
+    state.last_dig_found_nothing = false;
+
     // CR 701.20e + CR 608.2c: PriorLook means the card set was already populated
     // by a preceding look-only Dig (e.g. Birthing Ritual: sacrifice sits between
     // the look step and the choice step). Read from private_look_ids so that
@@ -105,6 +111,15 @@ pub fn resolve(
     // CR 401.5: If a library has fewer cards than required, use as many as available.
     let count = dig_num.min(player.library.len());
     if count == 0 {
+        // CR 608.2c: Nothing was looked at — a chained `ParentTarget` consumer
+        // ("put up to one of them on top … the rest on the bottom") has no
+        // cards to act on and must not fall back to acting on this ability's
+        // own source (issue #1365).
+        state.last_dig_found_nothing = true;
+        events.push(GameEvent::EffectResolved {
+            kind: EffectKind::from(&ability.effect),
+            source_id: ability.source_id,
+        });
         return Ok(());
     }
 
@@ -250,6 +265,10 @@ fn resolve_from_prior_look(
 ) -> Result<(), EffectError> {
     let cards = state.private_look_ids.clone();
     if cards.is_empty() {
+        // CR 608.2c: mirrors the empty-library branch in `resolve` (issue
+        // #1365) — no cards were looked at, so a chained `ParentTarget`
+        // consumer must not self-fallback.
+        state.last_dig_found_nothing = true;
         events.push(GameEvent::EffectResolved {
             kind: EffectKind::Dig,
             source_id: ability.source_id,
@@ -519,6 +538,17 @@ mod tests {
         let result = resolve(&mut state, &ability, &mut events);
         assert!(result.is_ok());
         assert!(matches!(state.waiting_for, WaitingFor::Priority { .. }));
+        assert!(
+            state.last_dig_found_nothing,
+            "an empty-library Dig must flag that it found nothing, so a chained \
+             ParentTarget consumer does not self-fallback (issue #1365)"
+        );
+        assert!(
+            events
+                .iter()
+                .any(|e| matches!(e, GameEvent::EffectResolved { .. })),
+            "an empty-library Dig must still emit EffectResolved"
+        );
     }
 
     #[test]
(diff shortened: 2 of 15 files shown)


matthewevans (changes requested):
**[MED]** The empty-Dig outcome is modeled as a raw global boolean and read by generic `ParentTarget` resolution. Evidence: `crates/engine/src/types/game_state.rs:6873` adds `last_dig_found_nothing: bool`, and `crates/engine/src/game/targeting.rs:583` suppresses every unresolved `TargetFilter::ParentTarget` while that flag is set. Why it matters: this loses the distinction between “no Dig has run”, “Dig looked at zero cards”, and “Dig looked at N cards”, and it makes the shared targeting chokepoint depend on one effect family’s transient outcome instead of keeping the no-op at the `PutAtLibraryPosition`/Dig-tail seam that actually needs it. Suggested fix: replace the boolean with typed transient outcome data, e.g. `last_dig_count: Option<usize>` or an equivalent small enum, and consume the `Some(0)` case only in the `PutAtLibraryPosition { target: ParentTarget, .. }` Dig-tail path before falling through to generic `resolved_targets`.

matthewevans (changes requested):
**[MED]** The stale-flag guard still suppresses the unrelated LTB self-return before clearing the flag. Evidence: `crates/engine/src/game/effects/put_on_top.rs:54` computes `dig_found_nothing_for_parent_target` from `state.last_dig_found_nothing`, then `put_on_top.rs:60` clears it, and `put_on_top.rs:69` still uses the already-captured true value to skip `resolved_targets`. The new regression at `put_on_top.rs:710` sets the flag to true and then expects the Avenging Angel-class `ParentTarget` self-return to happen, but this control flow makes `effective_targets` empty and falls into the “requires a target” path instead. Why it matters: the PR still fails its own stale-flag scenario, so a previous empty Dig can break an unrelated `PutAtLibraryPosition { ParentTarget }` resolution. Suggested fix: make the Dig-tail no-op distinguish a real preceding empty Dig for this specific continuation from an unrelated stale flag, preferably with typed transient outcome data instead of a raw `bool`, and ensure this test fails before the fix and passes after.

jsdevninja:
Addressed the review finding: the empty-Dig signal was a raw global bool read directly by the shared resolved_targets chokepoint, which could leak forward and suppress an unrelated PutAtLibraryPosition { ParentTarget } self-return later in the same resolution.

Fixed by replacing it with a typed, per-ability signal: ResolvedAbility::dig_found_nothing_for_parent_target, stamped only by effects::apply_parent_chain_context at the exact parent→child hand-off immediately following an empty-library Dig, which also consumes (clears) the transient global flag at that single hand-off. put_on_top.rs now checks only this per-ability field — targeting.rs (the shared chokepoint) is untouched. Added a regression test (stale_dig_found_nothing_does_not_suppress_unrelated_ltb_self_return) that sets the global flag directly and asserts an unrelated Avenging Angel-class LTB self-return still succeeds — it fails against the prior approach and passes now.

Full engine suite (14,266 tests) + integration tests + cargo clippy -p engine --tests -- -D warnings all green.

matthewevans (changes requested):
**[LOW]** New inline imports were added inside the regression test. Evidence: `crates/engine/src/game/effects/dig.rs` adds `use crate::parser::oracle_effect::parse_effect_chain;`, `use crate::types::ability::AbilityKind;`, and `use crate::types::mana::{ManaCost, ManaCostShard};` inside `thassas_oracle_dread_return_empty_library_still_wins`. The repository instructions require all imports at file scope, and we have been applying that rule consistently on these sweeps. Suggested fix: move those imports into the file's module-level import block (or use already-imported symbols if present). The substantive empty-Dig handoff fix looks like it addresses my earlier blocker; this is the remaining cleanup before I can approve/update/enqueue.

jsdevninja:
Fixed the import-scope nit: moved the regression test's imports to module scope. `parse_effect_chain` and `AbilityKind` were already imported at module level (the inline copies were redundant duplicates), and `ManaCost`/`ManaCostShard` are now imported there too.

Full lib suite (14,266 tests) + `cargo clippy -p engine --lib -- -D warnings` both clean.

matthewevans (approved):
Reviewed current head `39f372346f04beb99833173206f88e3f5484696e` after the import cleanup. The earlier blockers are addressed: the empty-Dig outcome is no longer read from generic targeting, the signal is stamped onto the immediate child at the Dig handoff and consumed there, the stale-flag regression covers unrelated ParentTarget self-return, and the Thassa's Oracle empty-library runtime test drives the parsed effect chain through resolution.

Gemini's older raw-bool concern is stale against this head for the important seam: the shared targeting chokepoint no longer reads the transient flag, and the remaining relay is scoped to the immediate handoff plus the PutAtLibraryPosition Dig-tail consumer.

PR #4638, opened 2026-06-30
feat(engine): support Greymond, Avacyn's Stalwart (choose-two-keywords grant and conditional Human anthem)

GildardoDev:
## Anchored on

Greymond, Avacyn's Stalwart ({2}{W}{W}, Creature — Human Soldier, 3/4), verified Oracle text:

> As Greymond, Avacyn's Stalwart enters, choose two abilities from among first strike, vigilance, and lifelink.
> Humans you control have each of the chosen abilities.
> As long as you control four or more Humans, Humans you control get +2/+2.

## Problem

Greymond was fully unsupported — its parse yielded multiple `Effect::Unimplemented` / `StaticCondition::Unrecognized` (no as-enters keyword choice, no plural "the chosen abilities" grant).

## Root cause

Two gaps in otherwise-shipped infrastructure:
- The as-enters keyword choice (`ChoiceType::Keyword`) was single-pick only — no way to "choose two abilities from among …".
- The chosen-keyword static grant (`ContinuousModification::AddChosenKeyword`) and its accessor read only the FIRST chosen keyword (documented limitation), so "have each of the chosen abilities" could never grant both.

(Line 3, the conditional tribal anthem, already parses via the existing control-count condition + typed-anthem path — confirmed by test, no code needed.)

## Fix (class-general — "choose N keywords", not Greymond-only)

- **Parameterize, don't proliferate:** added a single `count: usize` field to the existing `ChoiceType::Keyword` (default 1, serde-back-compatible) rather than a new `TwoKeywords` variant. The chosen pair persists as TWO ordinary `ChosenAttribute::Keyword` entries. **Zero new enum variants.** (Ran through the `add-engine-variant` gate: leaf parameterization on the count axis, within CR 608.2d.)
- New `GameObject::chosen_keywords()` accessor reads ALL chosen keywords; the existing `AddChosenKeyword` apply path (battlefield + off-zone) now grants every one, source-scoped via `effect.source_id` (two Greymonds with different pairs don't leak).
- Parser: a new `try_parse_keyword_choice_from_among` branch handles "choose N abilities/keywords from among <Oxford-and list>" (reusing the existing `FromAmong` splitter; trims the sentence-final period); the L2 anaphor `alt()` is extended with "each of the chosen abilities" → the existing static `AddChosenKeyword`.
- Lockstep: one constructor in `crates/mtgish-import` updated to `count: 1` (mandatory cross-crate compile update for the shared-enum shape change; not feature work).

CR 608.2d (making a choice), 613.1f (Layer 6 ability grant), 614.12c (persisted as-enters choice), 702.15b (lifelink), 611.3a / 400.7 (continuous re-eval / zone-change attribute clearing) — all grep-verified against `docs/MagicCompRules.txt`.

## Tests

- Parser: L1 exact text → `ChoiceType::Keyword { [FirstStrike,Vigilance,Lifelink], count: 2 }` + persist; negative siblings (bare "first strike, vigilance, or lifelink" stays `count: 1`; "left or right" stays `Labeled`); L2 → a `StaticDefinition` (not a trigger) with a Human-subtype you-control filter + `AddChosenKeyword`; L3 → both the count condition and the +2/+2 subject carry `Subtype("Human")` with comparator `>= 4`; whole-card parse → zero `Unimplemented`/`Unrecognized`.
- Runtime (GameScenario): cast + answer persists two chosen keywords; each Human you control gains BOTH (and only the) chosen abilities, non-Humans none; **two Greymonds with different pairs** → a Human gets the source-scoped union; **3 Humans → no +2/+2, 4 → +2/+2, back to 3 → off** (live); off-zone plural read; serde back-compat round-trip; AI candidate generation yields exactly the 3 pairs.

## Verification

`cargo fmt` clean; full-workspace `cargo check` clean (incl. mtgish-import); parser combinator gate exit 0; engine clippy `-D warnings` clean; engine parser (7014) / oracle_effect (2269) / layers (400) suites green; all new parser + 6 runtime fixtures pass.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -1801,7 +1801,7 @@ fn fmt_choice_type(ct: &ChoiceType) -> String {
         ChoiceType::Word => "word",
         ChoiceType::Artist => "artist",
         // CR 608.2d: "choose an ability" — Urborg / Walking Sponge prompt.
-        ChoiceType::Keyword { options } => {
+        ChoiceType::Keyword { options, .. } => {
             return format!(
                 "ability from: {}",
                 options
--- a/crates/engine/src/game/effects/choose.rs
+++ b/crates/engine/src/game/effects/choose.rs
@@ -198,9 +198,52 @@ pub(crate) fn bind_named_choice(
     source_id: Option<ObjectId>,
 ) {
     if let Some(obj_id) = source_id {
-        if let Some(attr) = ChosenAttribute::from_choice(choice_type.clone(), choice) {
+        // CR 608.2d: A multi-keyword choice (`ChoiceType::Keyword { count > 1 }`,
+        // e.g. Greymond's "choose two abilities from among ...") arrives as one
+        // comma-joined answer ("First Strike, Vigilance"). Split it on ',' and
+        // trim each token (tolerating "A, B" / "A,B" / "A ,B" whitespace
+        // variants) and persist one `ChosenAttribute::Keyword` per token so each
+        // chosen ability is independently readable by the `AddChosenKeyword`
+        // plural grant. The single-keyword path (count == 1, and every other
+        // choice type) produces a single attribute, byte-identical to before.
+        let attrs: Vec<ChosenAttribute> = match choice_type {
+            ChoiceType::Keyword { options, count } if *count > 1 => choice
+                .split(',')
+                .filter_map(|token| {
+                    ChosenAttribute::from_choice(
+                        ChoiceType::Keyword {
+                            options: options.clone(),
+                            count: 1,
+                        },
+                        token.trim(),
+                    )
+                })
+                .collect(),
+            _ => ChosenAttribute::from_choice(choice_type.clone(), choice)
+                .into_iter()
+                .collect(),
+        };
+        if !attrs.is_empty() {
             if let Some(obj) = state.objects.get_mut(&obj_id) {
-                obj.chosen_attributes.push(attr);
+                // CR 608.2d: A keyword choice represents the CURRENT answer set,
+                // not an accumulation. A source that makes a fresh keyword choice
+                // each time its effect resolves (Angelic Skirmisher — "At the
+                // beginning of combat on your turn, choose first strike, vigilance,
+                // or lifelink. Creatures you control gain that ability until end of
+                // turn") must REPLACE its prior keyword answer, otherwise the
+                // `AddChosenKeyword` plural read would grant every historical
+                // choice. Clear only `ChosenAttribute::Keyword` (Greymond's single
+                // as-enters bind clears nothing, so its behavior is unchanged);
+                // every other chosen-attribute kind (Color, Subtype, CardName,
+                // Label, …) is untouched so RemoveChosenKeyword/Urborg and the
+                // anchor-word/Morophon cards keep accumulating per their own rules.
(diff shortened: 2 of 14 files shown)


matthewevans (changes requested):
**[HIGH]** `AddChosenKeyword` now grants every historical keyword choice stored on the source, but `bind_named_choice` only appends keyword attributes and never replaces the previous choice. Evidence: `crates/engine/src/game/effects/choose.rs:208`-`228` builds one or more `ChosenAttribute::Keyword` values and pushes them onto `obj.chosen_attributes` without clearing the prior keyword choice, while `crates/engine/src/game/layers.rs:4037`-`4052` changes `AddChosenKeyword` to read `src.chosen_keywords()` and grant the full list. That is correct for one Greymond answer, but it regresses existing repeated single-keyword choice cards such as Angelic Skirmisher: choose first strike one combat, choose lifelink on a later combat, and the source now grants both because both historical `ChosenAttribute::Keyword` entries remain. The new runtime helper masks the problem by explicitly clearing all keyword attributes before seeding (`crates/engine/tests/greymond_avacyns_stalwart.rs:58`-`64`), but the real `ChooseOption` path does not do that. Please make the stored keyword choice represent the current answer set for the relevant choice instance, or otherwise scope plural Greymond reads so old single-choice answers cannot leak into later grants, and add a regression that drives two successive keyword choices through the real handler and verifies only the current choice(s) are granted.

GildardoDev:
Good catch, thanks. Reworked so a keyword choice represents the current answer set rather than accumulating: when binding a `ChoiceType::Keyword`, `bind_named_choice` now retain-filters out the source's prior `ChosenAttribute::Keyword` entries before pushing the new answer (scoped to `Keyword` only, so Subtype/Color/Name/RemoveChosenKeyword/Urborg are unaffected, and it stays source-scoped per object). Greymond's single count=2 bind clears nothing and is unchanged; repeated single-keyword choosers like Angelic Skirmisher now replace their prior choice instead of granting every historical one.

Also took gemini's split robustness fix (`split(',')` + `trim()` per token).

Added a regression (`repeated_keyword_choice_replaces_prior_answer`) that drives two successive keyword choices through the real `ChooseOption` / `bind_named_choice` path and asserts only the current keyword is granted; it fails without the clear. Re-requesting review.

matthewevans (approved):
Reviewed the updated head after merging main.

Evidence checked:
- `ChoiceType::Keyword` is parameterized with `count`, with legacy JSON defaulting to `count: 1` and `count: 1` serialization remaining byte-stable.
- The prior blocker is addressed at the binding seam: `bind_named_choice` clears prior `ChosenAttribute::Keyword` entries before storing the current keyword answer set.
- `AddChosenKeyword` reads the source-scoped plural chosen keyword set in both layer and off-zone paths.
- The regression drives the real `ChooseOption` / `bind_named_choice` path twice and asserts the second single-keyword choice replaces the first.
- Greymond has runtime coverage for cast-time count-2 choice persistence, plural keyword grant, source scoping, live Human-count anthem, serde, and AI pair candidates.
- Local post-update gates passed: `cargo fmt --all --check`, `scripts/check-parser-combinators.sh`, `git diff --check`; CR citations resolve in local `docs/MagicCompRules.txt`.

CI is pending on the updated branch; merge-when-ready can wait for required checks.

PR #4640, opened 2026-06-30
fix(parser): support "more cards in hand than each opponent" condition

GildardoDev:
## Anchored on

Okina Nightwatch and Secretkeeper, both carrying the static condition "As long as you have more cards in hand than each opponent, ..." (Okina gets +3/+3; Secretkeeper gets +2/+2 and has flying). The same phrasing appears on 7 cards in total.

## Problem

The condition "you have more cards in hand than each opponent" parsed to `StaticCondition::Unrecognized`, so the conditional buff on both cards never applied.

## Root cause

`parse_hand_size_predicate` (the shared hand-size comparison combinator) had a "more cards in hand than you" arm but none for "than each opponent", and no other path in `parse_inner_condition` recognized it.

## Fix

Added a single arm to `parse_hand_size_predicate` for "more cards in hand than each opponent" that yields `QuantityComparison { HandSize(controller) GT HandSize(Opponent { aggregate: Max }) }`. "Each opponent" is universal (more than all of them), so the bar is the maximum opponent hand, which is the mirror of the existing "more life than an opponent" arm using Min for its existential "an". It reuses the existing `QuantityComparison`, `QuantityRef::HandSize`, `PlayerScope::Opponent`, and `AggregateFunction` types with no new variants.

## Tests

Parser unit test: "as long as you have more cards in hand than each opponent" parses to the expected `QuantityComparison` (HandSize controller GT HandSize Opponent{Max}). Whole-card parses of Okina Nightwatch and Secretkeeper were confirmed to yield zero `Unrecognized` / `Unimplemented`.

## Verification

`cargo fmt` clean; parser combinator gate exit 0; engine clippy `-D warnings` clean; the new parser test and the existing hand-comparison suite pass.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -1931,6 +1931,32 @@ fn parse_hand_size_predicate(rest: &str, player: PlayerScope) -> Option<(&str, S
         ));
     }
 
+    // CR 402: "more cards in hand than each opponent" → HandSize(player) GT
+    // HandSize(Opponent{Max}). "each opponent" is universal (more than ALL of
+    // them), so the predicate is "your hand > the maximum opponent hand" — the
+    // Max aggregate, mirroring how "more life than an opponent" uses Min for its
+    // existential "an". Okina Nightwatch, Secretkeeper.
+    if let Ok((rest, _)) =
+        tag::<_, _, OracleError<'_>>("more cards in hand than each opponent").parse(rest)
+    {
+        return Some((
+            rest,
+            StaticCondition::QuantityComparison {
+                lhs: QuantityExpr::Ref {
+                    qty: QuantityRef::HandSize { player },
+                },
+                comparator: Comparator::GT,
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed head `e75540ed50a08b34028a7b64226c4ab723bf2a71`.

This is a narrow backend parser fix in the right place: `parse_hand_size_predicate` already owns hand-size predicate suffixes, and the new branch keeps the existing nom-combinator style instead of adding a separate parser path. Modeling `each opponent` as `HandSize(Controller) > HandSize(Opponent { Max })` matches the universal comparison: the controller must exceed the largest opposing hand. The added parser test covers the full `parse_condition("as long as ...")` entry point and would fail if this phrase regressed to `Unrecognized` again.

Local review gates passed: `cargo fmt --all --check`, `scripts/check-parser-combinators.sh`, and `git diff --check origin/main...HEAD`. CI is still allowed to finish under merge-when-ready.

Issue #4544, opened 2026-06-28
[Card Bug] Abhorrent Oculus manifested cards


Ryu3394:
### Card(s)

Abhorrent Oculus

### Build/version

v.09.0 7c1c1cf

### Game mode

None

### Actual behavior

When Abhorrent oculus ability triggers, if you choose a non-permanent card, it disappears: the card will not be put on the battlefield or in any other space. 
If you manifest a permanent card  you cannot look at it, and if it is a creature, you can no longer flip it in any way.

### Expected behavior

When Abhorrent oculus ability triggers, 1 manifested creature should come on battlefiled face down, even if you chose a non-permanent card.
You should be able to look at you face down cards and flip them if they are creatures

mike-theDude:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `manifest dread` (supported: true)

**Reasoning**: Abhorrent Oculus is parsed with ManifestDread on each opponent upkeep. The reported behavior is runtime support for face-down manifested cards: nonpermanents should still become face-down 2/2 creatures, owners should be able to inspect them, and legal creature cards should be turnable face up.

---
_Posted by bug-coverage-classifier skill._

matthewevans:
A new Discord report identifies a related visibility failure: when an opponent resolves Abhorrent Oculus’s manifest-dread trigger, the reporting player can see the two cards from which that opponent is choosing. Linked to the new report.

matthewevans:
Verified against current `origin/main`: merged PR #4641 implements the reported nonpermanent Manifest Dread path and includes `issue_4544_abhorrent_oculus_manifest_nonpermanent.rs`. The current face-down visibility filter grants the controller the CR 708.5 look permission, and the manifest turn-face-up action remains available for creature cards. Closing as already shipped.

PR #4641, opened 2026-06-30
fix(engine): manifest dread drops a non-permanent card instead of manifesting it face down

luciferlive112116:
## Summary

Fixes manifest (manifest dread / morph-class face-down entries) dropping a chosen **non-permanent** (instant/sorcery) card: instead of manifesting it face down as a 2/2, the card was left in its origin zone. Addresses the engine half of #4544 (Abhorrent Oculus manifesting non-permanent cards "disappear"). A manifested card is a colorless 2/2 creature regardless of its hidden card types (CR 708.2a), so an instant/sorcery must manifest like any other card.

### Root cause

`zone_pipeline::deliver_replaced_zone_change` commits a battlefield entry by calling `zones::move_to_zone` (zone_pipeline.rs), and only applies the face-down profile (`apply_face_down_entry_profile`) **afterward**. But `move_to_zone` contains the CR 304.4 / CR 400.4a guard that rejects an instant/sorcery entering the battlefield unless `obj.face_down` is set (zones.rs). At the time of the `move_to_zone` call the object is not yet flagged face down, so the guard bounces a non-permanent manifest back to its origin zone.

- A **land** manifests fine (it is a permanent type, so the guard doesn't apply).
- A **default/typeless** scenario card manifests fine (not an instant/sorcery).
- Only **instant/sorcery** cards hit the guard — matching the report.

### Fix

In `deliver_replaced_zone_change`, flag the object face down just before the battlefield move dispatch when the delivered event carries a `face_down_profile`, so the legality guard sees the incoming face-down entry. The full face-down profile is still applied right after the move. A Library/Hand → Battlefield manifest never hits the `face_down`-clearing reset branches (which key on `from` == Exile/Battlefield/Stack), so the flag survives until the profile is applied.

This mirrors the already-tested invariants in `zones.rs` (`face_down_instant_can_enter_battlefield`, `face_down_spell_stays_face_down_when_resolving_to_battlefield`).

> This is a minimal, narrowly-scoped ordering fix (flag `face_down` before the entry legality guard runs) plus a regression test, with no AST changes and no new effect/parser logic. Happy to route it through `/engine-implementer` if you'd prefer the full pipeline for an engine-behavior change.

## CR references

- CR 708.2a — a face-down permanent is a 2/2 creature with no name/types/abilities
- CR 708.3 — turned face down before it enters the battlefield
- CR 304.4 / CR 400.4a — instants/sorceries can't be put onto the battlefield (the guard being bypassed for face-down entries)

## Verification

- Added regression test `crates/engine/tests/integration/issue_4544_abhorrent_oculus_manifest_nonpermanent.rs` (registered in `tests/integration/main.rs`): Abhorrent Oculus manifests a chosen instant face down as a 2/2 on the battlefield and graves the other looked-at card. It fails on `main` (the card stays in the library) and passes with this fix.
- The fix is backed by the existing `zones.rs` invariant tests `face_down_instant_can_enter_battlefield` and `face_down_spell_stays_face_down_when_resolving_to_battlefield`, which already prove that an object flagged face down before `move_to_zone` enters the battlefield legally and stays face down.
- Suggested local check: `cargo test -p engine --lib --test integration -- manifest face_down` (covers `issue_3245`, `issue_3311`, and the new `issue_4544` regression test). Please confirm via CI as part of review.


matthewevans (changes requested):
**[HIGH]** The pre-move face-down mutation is not rolled back when the battlefield entry is rejected. Evidence: `crates/engine/src/game/zone_pipeline.rs:1641` sets `obj.face_down = true` before calling `move_to_zone`, but `crates/engine/src/game/zones.rs:552` returns early when `is_blocked_from_entering_battlefield` matches a `CantEnterBattlefieldFrom` static such as Grafdigger's Cage (`crates/engine/src/game/zones.rs:1189`). Why it matters: a manifest/morph-style move that is blocked from entering the battlefield leaves the card in its original zone with `face_down` incorrectly set, corrupting hidden/card state even though the zone change never happened. Suggested fix: avoid committing the temporary guard flag unless the move succeeds, or snapshot and restore the prior `face_down` value when the object remains in its original zone; add a regression where a face-down library/graveyard-to-battlefield entry is blocked and the card stays unchanged.

matthewevans (changes requested):
**[HIGH]** The rejected-entry path still strands the object face down in its original zone. Evidence: `crates/engine/src/game/zone_pipeline.rs:1652` mutates `obj.face_down = true` before `zones::move_to_zone`, but `move_to_zone` can return without moving at `crates/engine/src/game/zones.rs:556` when `is_blocked_from_entering_battlefield` matches a `CantEnterBattlefieldFrom` static such as Grafdigger's Cage. The new `apply_face_down_entry_profile` assertion only runs after a successful move, so it does not repair this early-return path. Why it matters: a manifest/morph-style move blocked from entering the battlefield corrupts the card left in Library/Graveyard/etc. by marking it face down even though no zone change occurred. Suggested fix: snapshot and restore the previous `face_down` value when the object remains in its original zone after `move_to_zone`, or avoid persistent mutation for the guard preflight; add a regression with a `CantEnterBattlefieldFrom` static blocking a face-down library/graveyard-to-battlefield entry and asserting the card remains unchanged.

luciferlive112116:
Thanks for the review — addressed the HIGH rejected-entry issue.

The pre-move face-down flag is no longer committed unconditionally:

1. **Snapshot + rollback.** `deliver_replaced_zone_change` now snapshots the prior `face_down` before the preflight, computes `entered_battlefield` (whether the object actually ended up on the battlefield after `move_to_zone`), and restores the prior value when the entry was rejected. So a `CantEnterBattlefieldFrom` static such as Grafdigger's Cage (CR 614.1d) no longer strands the card face down in its origin zone.
2. **Profile gated on actual entry.** `apply_face_down_entry_profile` is now gated on `entered_battlefield` rather than the requested destination (`to == Battlefield`), so a rejected entry never morphs the hidden card in place — it is left fully unchanged.
3. **Regression added.** `blocked_battlefield_entry_does_not_strand_card_face_down`: a face-down library→battlefield manifest blocked by Grafdigger's Cage stays in the library with `face_down == false` and no profile applied (`back_face` is `None`).

The successful path is unchanged: on a real entry the flag is re-asserted by `apply_face_down_entry_profile`, so the rollback is inert there.

matthewevans (changes requested):
**[LOW]** New inline imports remain inside the added regression test. Evidence: `crates/engine/src/game/zone_pipeline.rs:3015` imports `FilterProp`, `StaticDefinition`, `TargetFilter`, `TypeFilter`, and `TypedFilter` inside `blocked_battlefield_entry_does_not_strand_card_face_down`, and `zone_pipeline.rs:3018` imports `StaticMode` inside the same function. Why it matters: the repo instructions require imports at file/module scope, and we have been applying this consistently on current PRs before approval. Suggested fix: move these imports to the surrounding test module's import block and leave the test body import-free.

The rejected-entry rollback shape now addresses my earlier blocker; this import cleanup is the remaining issue before this can be approved/updated/enqueued.

matthewevans (changes requested):
**[LOW]** New inline imports remain inside the added regression test. Evidence: `crates/engine/src/game/zone_pipeline.rs:3019`. Why it matters: repo instructions explicitly require imports at file/module scope. Suggested fix: move `FilterProp`, `StaticDefinition`, `TargetFilter`, `TypeFilter`, `TypedFilter`, and `StaticMode` into the surrounding test module import block.

Reviewed current head `edb76b251833731e77667e602658aa46d135e8c8`.

luciferlive112116:
Addressed the remaining **LOW** (inline test imports).

- `FilterProp`, `StaticDefinition`, `TargetFilter`, `TypeFilter`, `TypedFilter`, and `StaticMode` are hoisted from `blocked_battlefield_entry_does_not_strand_card_face_down` into the surrounding `face_down_exile_entry_tests` module import block, leaving the test body import-free (`d141c9af1`).
- Also wrapped an over-long `obj.toughness` 2/2 assertion that tripped `cargo fmt --all -- --check` (`eca9b05fb`).

Local verification: `cargo check -p engine --lib --tests` clean, `cargo fmt --all -- --check` clean. Happy to merge latest `main` in if you'd prefer it un-behind before enqueue — otherwise ready for re-review.

luciferlive112116:
Merged latest `main` in (`a14859404`) so it is no longer behind. Clean merge (no conflicts); net PR diff vs `main` is unchanged (+364 / -1, the manifest fix only). Verified locally against current `main`: `cargo fmt --all -- --check`, `cargo check -p engine --lib --tests`, and `cargo clippy -p engine --lib --tests -- -D warnings` all clean. Ready for re-review / enqueue.

matthewevans (approved):
Reviewed current head. The face-down entry path now preflights non-permanent battlefield entry correctly, rolls back if entry is rejected, and reasserts the final face-down profile after exile exit cleanup. The tests cover the exile-origin non-permanent manifest case and the blocked-entry rollback case.

PR #4525, opened 2026-06-28
feat(engine): Alchemy perpetual P/T modifier (perpetually gets +N/+M)

philluiz2323:
Fixes #4524.

## Summary

Adds the Alchemy **perpetual P/T modifier** form — "[object] perpetually gets +N/+M" — as PerpetualModification::ModifyPowerToughness, parser arm, and resolver delta on persistent ase_power/ase_toughness with layer flush.

Cards: **Heir to Dragonfire** (~ perpetually gets +3/+3), **Tiana, Angelic Mechanic** (that Vehicle perpetually gets +1/+0).

## Implementation method (required)

- [ ] Produced via the `/engine-implementer` pipeline (plan → review-plan → implement → review-impl → commit)
- [x] **Not** `/engine-implementer` — explain why below

> Focused parser + resolver increment on the existing `ApplyPerpetual` / `perpetual_mods` infrastructure from #4104; too small for the full pipeline.

## CR references

None (digital-only Alchemy mechanic, consistent with `Intensify` / `Conjure`).

## Verification

- Parser unit tests: self-subject `~ perpetually gets +3/+3` and anaphoric `that vehicle perpetually gets +1/+0` → `ModifyPowerToughness`.
- Resolver regression: +3/+3 on 1/1 base updates `base_*` and live P/T after `flush_layers`.
- CI covers `cargo fmt`, clippy, parser gate, and engine tests.

PR #4642, opened 2026-06-30
feat(engine): Alchemy perpetual self-cost modification (perpetually gains "this spell costs {N} less/more to cast")

GildardoDev:
## Anchored on

The Alchemy (digital-only) "perpetual self-cost modification" class: an effect that permanently grants a card a change to its own cast cost, written as `perpetually gains "This spell costs {N} less to cast"` (~17 cards) and `perpetually gains "This spell costs {N} more to cast"` (~6 cards). ~23 cards today, plus every future card with the same grant template.

## Problem

`perpetually gains "This spell costs {N} less/more to cast"` had no engine support. The existing `PerpetualModification` enum covered base P/T, +N/+M, keyword grants, and type-change (philluiz's #4525/#4529/#4616), but no perpetual cost modifier, so these cards' grants did nothing.

## Root cause

Two layers were missing: the perpetual-modification type had no cost variant, and the perpetual-grant parser had no arm for the quoted self-cost static.

## Fix (class-general, parameterized)

- **One new leaf variant** `PerpetualModification::ModifyCost { mode: CostModifyMode, amount: ManaCost }` — direction (`Reduce`/`Raise`) and amount are parameters, not separate variants (passed the add-engine-variant gate; the axis lies within CR 601.2f). Zero other new types.
- **Parser:** a new `try_parse_perpetual_modify_cost` arm in the perpetual-grant cluster, parsing the quoted `"This spell costs {N} less/more to cast"` body with nom combinators (`delimited` + `parse_mana_cost` + `value(Reduce/Raise)`), reusing a factored-out `parse_perpetual_self_subject` helper shared with the keyword-grant arm. Honest-red `None` on any unconsumed tail.
- **Realization (reuse, not reinvent):** `apply_perpetual_modification` injects a synthetic self-spell `StaticMode::ModifyCost` static (`affected = SelfRef`, `active_zones = self_spell_cost_mod_active_zones()`) into **both** the live `static_definitions` and the persistent `base_static_definitions` — mirroring how the `GrantKeywords` arm writes both `keywords` and `base_keywords`. The live copy makes the discount visible to a from-hand cast immediately; the base copy survives the battlefield layer reset (`static_definitions = base.clone()`). `apply_perpetual_modification` runs once per `ApplyPerpetual` resolution (single caller), so there is no double-injection; distinct grants intentionally stack.
- **Consumption: no cost-calc changes.** The existing `collect_self_spell_cost_modifiers` consumes the injected static unchanged, including the raise-before-reduce ordering and the `{0}` floor.

CR 601.2f (cost determination, increase/reduction, `{0}` floor); "perpetually" is a digital-only Alchemy keyword action with no CR entry (annotated as such, matching the existing perpetual code).

## Tests

- **Runtime (real cast pipeline, `GameRunner::cast(...).resolve()`):** reduce lowers self-cast cost; raise increases it; `{0}` floor; the modifier **persists across a hand→library→hand zone change**; two grants **stack and floor** correctly; and a from-hand cast funded with **only the reduced `{2}` pool** resolves (sorcery → graveyard, CR 608.2n) — proving the reduction applies on the real payment path, not just to the computed cost.
- **Parser (the class):** reduce/raise/anaphoric forms map to `ModifyCost`; negative honest-red (unconsumed tail → `Unimplemented`); `perpetually gains flying` still routes to `GrantKeywords` (no regression).
- **serde:** additive `#[serde(tag="kind")]` round-trip.

## Verification

`cargo fmt` clean; `cargo check -p engine` clean; engine clippy `-D warnings` clean; parser combinator gate exit 0; the 6 new runtime + 3 parser tests pass; broad perpetual (25) and casting-cost (944) suites stay green. Engine-only change, no cross-crate lockstep.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -21517,6 +21517,423 @@ mod tests {
         }
     }
 
+    /// Apply an Alchemy "perpetually gains \"This spell costs {N} less/more to
+    /// cast\"" grant to `spell_id` through the REAL resolution path
+    /// (`parse_effect_chain` → `build_resolved_from_def` → `resolve_ability_chain`),
+    /// exactly as the `ApplyPerpetual` effect runs in production. Returns nothing;
+    /// the synthetic self-spell `ModifyCost` static is injected onto the object.
+    fn grant_perpetual_cost_mod(state: &mut GameState, spell_id: ObjectId, oracle: &str) {
+        use crate::game::ability_utils::build_resolved_from_def;
+        use crate::game::effects::resolve_ability_chain;
+        use crate::types::ability::AbilityKind;
+        let def = parse_effect_chain(oracle, AbilityKind::Activated);
+        let ability = build_resolved_from_def(&def, spell_id, PlayerId(0));
+        let mut events = Vec::new();
+        resolve_ability_chain(state, &ability, &mut events, 0).unwrap();
+    }
+
+    /// Build a generic {N}-cost sorcery in hand for the perpetual-cost tests.
+    fn create_generic_sorcery_in_hand(
+        state: &mut GameState,
+        card_id: u64,
+        name: &str,
+        generic: u32,
+    ) -> ObjectId {
+        let obj_id = create_object(
+            state,
+            CardId(card_id),
+            PlayerId(0),
+            name.to_string(),
+            Zone::Hand,
+        );
+        let obj = state.objects.get_mut(&obj_id).unwrap();
+        obj.card_types.core_types.push(CoreType::Sorcery);
+        obj.mana_cost = ManaCost::Cost {
+            shards: vec![],
+            generic,
+        };
+        obj_id
+    }
+
+    fn self_cost_generic(state: &GameState, spell_id: ObjectId) -> u32 {
+        let mut mana_cost = state.objects.get(&spell_id).unwrap().mana_cost.clone();
+        apply_self_spell_cost_modifiers(state, PlayerId(0), spell_id, &mut mana_cost);
+        match mana_cost {
+            ManaCost::Cost { generic, .. } => generic,
+            other => panic!("expected ManaCost::Cost, got {other:?}"),
+        }
+    }
+
+    /// Reduce: a {4}-cost card granted `Reduce {2}` casts for {2}; without the
+    /// grant the same card costs {4} (revert).
+    #[test]
+    fn perpetual_cost_reduce_lowers_self_cast_cost() {
+        let mut state = setup_game_at_main_phase();
+        let spell = create_generic_sorcery_in_hand(&mut state, 4_001, "Reduce Test", 4);
+
+        // Revert baseline: no grant → full {4}.
+        assert_eq!(
+            self_cost_generic(&state, spell),
+            4,
+            "ungranted cost is {{4}}"
+        );
+
+        grant_perpetual_cost_mod(
+            &mut state,
+            spell,
+            "~ perpetually gains \"This spell costs {2} less to cast\".",
+        );
+        // CR 601.2f: {4} − {2} = {2}.
+        assert_eq!(
+            self_cost_generic(&state, spell),
+            2,
+            "Reduce {{2}} lowers {{4}} to {{2}}"
+        );
+    }
+
+    /// Raise: a {4}-cost card granted `Raise {2}` casts for {6}.
+    #[test]
+    fn perpetual_cost_raise_increases_self_cast_cost() {
+        let mut state = setup_game_at_main_phase();
+        let spell = create_generic_sorcery_in_hand(&mut state, 4_002, "Raise Test", 4);
+
+        grant_perpetual_cost_mod(
+            &mut state,
+            spell,
+            "~ perpetually gains \"This spell costs {2} more to cast\".",
+        );
+        // CR 601.2f: {4} + {2} = {6}.
+        assert_eq!(
+            self_cost_generic(&state, spell),
+            6,
+            "Raise {{2}} raises {{4}} to {{6}}"
+        );
+    }
+
+    /// Floor: a {1}-cost card granted `Reduce {3}` floors at {0} (not negative).
+    #[test]
+    fn perpetual_cost_reduce_floors_at_zero() {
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
**[HIGH]** The parser does not accept the real quoted-cost Oracle text. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:6795` consumes only `" to cast"`, and the surrounding `delimited(..., char('"'))` at `crates/engine/src/parser/oracle_effect/mod.rs:6798` requires the closing quote immediately afterward. The new tests use synthetic text with the period outside the quote, e.g. `crates/engine/src/game/casting.rs:21584`, but real cards in this class use punctuation inside the granted quoted ability, such as `perpetually gain "This spell costs {1} less to cast."` and comma-separated forms like `"This spell costs {1} less to cast,"`. Why it matters: the PR's claimed ~23-card class still falls through instead of producing `PerpetualModification::ModifyCost`, so the runtime implementation is unreachable for the real card text. Suggested fix: parse the granted quoted ability with terminal punctuation inside the quote (`to cast` followed by optional `.`/`,` before the closing quote) and add parser/runtime fixtures using exact Oracle strings from representative less, more, and comma-separated grant forms.

**[MED]** The subject parser only covers self and `that ...` forms, not several real producers in the claimed class. Evidence: `parse_perpetual_self_subject` only accepts a leading `that ...` branch or the fixed self subjects listed at `crates/engine/src/parser/oracle_effect/mod.rs:6725`; it does not accept `it perpetually gains ...`, `cards in your hand ... perpetually gain ...`, or `each <type> card ... perpetually gains ...`, all of which appear in the current Alchemy cost-grant class described by the PR. Why it matters: even after the quote punctuation is fixed, broad-card and pronoun forms such as `It perpetually gains "This spell costs ..."` remain unsupported while the PR advertises the whole class. Suggested fix: either route those subjects through existing target/selection machinery into `ApplyPerpetual` or narrow the claim/tests to the exact supported subject forms and leave unsupported class members honest-red.

GildardoDev:
Thanks, both fair catches.

**[HIGH] punctuation inside the quote** — the quoted-cost body parser now accepts a terminal `.`/`,` inside the quote (`opt(alt((tag("."), tag(","))))` before the closing `"`), factored into one shared `parse_quoted_self_spell_cost_body` helper. My earlier tests put the period outside the quote, which masked this; the new tests use the exact Oracle strings.

**[MED] subjects** — added the mass `[type] cards in your hand perpetually gain "..."` form, using the same latched `ApplyPerpetual` + `PerpetualModification::ModifyCost` mechanism rather than the continuous-static path. A new arm in `perpetual_target_object_ids` enumerates a non-battlefield-zone target filter (`extract_in_zone` + `zone_object_ids` + `matches_target_filter`, before the source fallback so a no-match scan never buffs the source), and the per-card `SelfRef` cost static is applied to each matching hand card. This covers Blooming Cactusfolk, Charged Conjuration, Fearsome Whelp, Fountainport Charmer, plus the `that card` anaphor (Paths of Tuinvale, Talion's Throneguard). The type axis is parsed with strict combinators (no `take_until`).

**Deliberately honest-red (Unimplemented, not silently mis-parsed):**
- `It perpetually gains "..."` and `Choose a card in your hand. It ...` — these need a card-in-hand-pick latch the engine doesn't have yet (would be a separate interactive-effect feature).
- `a random ... card in ... hand perpetually gains "..."` — no random-single-pick primitive; the existing shipped arm is unchanged (I did not turn "a random card" into "all matching").
- Absorb Energy's `... that share a card type with that spell ...` rider, `{X}` bodies, and trailing riders (`... and draws a card`) — the mass arm rejects any unconsumed tail, so these fall through cleanly to Unimplemented.

Runtime cast-pipeline tests added: a matching hand card's cost actually drops while non-matching and opponent-hand cards don't. Re-requesting review.

matthewevans (changes requested):
Thanks for addressing the earlier quoted-punctuation issue. I found one remaining blocking correctness issue in the new mass-hand arm.

`that player's hand` is currently parsed as `controller: None` on the hand-card filter. That is not a scoped reference to the target/triggering player; in the existing filter machinery `None` means any controller. The new resolver then sees `InZone { Hand }`, enumerates `zone_object_ids(state, Zone::Hand)`, and that function returns cards from every player's hand. The result is that a clause like `Nonland cards in that player's hand perpetually gain "This spell costs {1} less to cast."` can apply the cost modifier to both players' matching hand cards instead of only that player's matching cards.

Please keep this honest-red until the player referent is modeled, or bind it through an existing scoped player axis such as `ControllerRef::TargetPlayer` / trigger player context as appropriate, and add a resolver-level regression with matching cards in both players' hands proving only the referenced player's cards are modified. The parser test currently asserts `controller: None`, which locks in the wrong behavior.

GildardoDev:
Good catch, you're right that `controller: None` means any controller, so the hand-zone resolver would have enumerated every player's hand and modified both players' matching cards.

Fixed by making `that player's hand` honest-red: the mass arm now only parses `... in your hand ...` (`controller = ControllerRef::You`, the spell's controller). `that player's hand` no longer parses and falls through to `Unimplemented`, until there's a scoped-player axis to bind the referent. All four covered cards (Blooming Cactusfolk, Charged Conjuration, Fountainport Charmer, Fearsome Whelp) use "in your hand", so coverage is unchanged.

The parser test that asserted `controller: None` now asserts `that player's hand` is honest-red (not `ApplyPerpetual`). And the existing resolver-level test already proves the `your hand` scoping with cards in both players' hands - a matching card in the opponent's hand under the `You` filter is left unmodified while the controller's matching card is reduced. Re-requesting review.

matthewevans (approved):
Reviewed the current head. The scoped-hand blocker is fixed by leaving `that player's hand` unsupported instead of mapping it to an unscoped hand filter, and the new resolver tests cover matching cards, nonmatching cards, and opponent hand cards. The self-cost modifier is represented as a `StaticMode::ModifyCost` on `SelfRef` and consumed by the existing cast-cost path, with real cast-pipeline coverage for the resulting cost change.

PR #4645, opened 2026-06-30
test(parser): direct-call regression for Optional_YouMay Prototype reminder exemption

### LINKED ISSUE #4644: Add direct-call regression test for detect_optional_you_may Prototype reminder-text exemption
## Summary

`crates/engine/src/parser/swallow_check.rs` guards the `Optional_YouMay`
swallowed-clause detector almost entirely through full-pipeline `parse_named(...)`
fixtures. The companion `Modal_DynamicMaxDropped` detector, by contrast, also has
**direct-call** unit tests (`modal_dynamic_max_dropped_fires_on_modal_without_dynamic_cap`,
`modal_dynamic_max_dropped_silent_*`) that exercise the detector function in
isolation with crafted inputs.

The `detect_optional_you_may` exemption for **Prototype keyword reminder text**
(`"(You may cast this spell with different mana cost, color, and size. It keeps its
abilities and types.)"`, CR 702.160a) is not pinned by any direct-call test. That
exemption is currently only covered indirectly. A parser change that happens to keep
the existing full-pipeline fixtures green could silently weaken or invert the
reminder-text gate without any test failing.

## Expected

A direct-call unit test (mirroring the `detect_modal_dynamic_max_dropped_*` pattern)
that asserts:

- a bare `"you may <verb>"` optional effect with no AST representation **fires**
  `Optional_YouMay` (CR 117.3a), and
- the Prototype keyword reminder text stays **silent** (CR 702.160a),

so the reminder-text exemption is regression-locked at the detector boundary.

luciferlive112116:
## Summary

Adds a direct-call regression test for `detect_optional_you_may` in

The new test pins the detector contract at its boundary:

- A bare `"you may <verb>"` optional effect with **no** AST representation must
  fire `Optional_YouMay` (CR 117.3a).
- The **Prototype** keyword reminder text
  (`"(You may cast this spell with different mana cost, color, and size...)"`,
  CR 702.160a) must stay **silent** — it is keyword reminder text, not a
  resolution-time optional effect.

Previously this exemption was only covered indirectly through full-pipeline
`parse_named(...)` fixtures. Pinning it directly means a future parser change that
keeps those fixtures green can no longer silently weaken or invert the
reminder-text gate.

Closes #4644.

## Test plan

- [ ] `optional_you_may_fires_on_bare_verb_and_exempts_prototype_reminder` passes
      (added in `swallow_check.rs` tests).
- [ ] No behavior change — test-only addition; reuses the established

### DIFF
--- a/crates/engine/src/parser/swallow_check.rs
+++ b/crates/engine/src/parser/swallow_check.rs
@@ -3611,6 +3611,64 @@ mod tests {
         }
     }
 
+    /// CR 117.3a + CR 702.160a: direct-call contract for `detect_optional_you_may`.
+    /// A bare "you may <verb>" optional effect with no AST representation must fire
+    /// `Optional_YouMay`, while the Prototype keyword reminder text
+    /// ("You may cast this spell with different mana cost...") is keyword reminder
+    /// text — not a resolution-time optional effect — and must stay silent. This
+    /// pins the reminder-text exemption directly: reverting that gate re-fires the
+    /// detector on every Prototype card, and dropping the final `diagnostics.push`
+    /// drops the fire case.
+    #[test]
+    fn optional_you_may_fires_on_bare_verb_and_exempts_prototype_reminder() {
+        let parsed = crate::parser::oracle::ParsedAbilities {
+            abilities: Vec::new(),
+            triggers: Vec::new(),
+            statics: Vec::new(),
+            replacements: Vec::new(),
+            extracted_keywords: Vec::new(),
+            modal: None,
+            additional_cost: None,
+            casting_restrictions: Vec::new(),
+            casting_options: Vec::new(),
+            solve_condition: None,
+            strive_cost: None,
+            parse_warnings: Vec::new(),
+        };
+
+        // Fire: an unrepresented "you may draw a card" optional effect.
+        let mut fire = Vec::new();
+        super::detect_optional_you_may(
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The added CR evidence cites the wrong rule for the fire case. Evidence: `crates/engine/src/parser/swallow_check.rs:3614` says bare `you may <verb>` is covered by `CR 117.3a`, but `docs/MagicCompRules.txt` has `117.3a` as priority timing, not optional effects or choices. Why it matters: swallow-check tests are parser/rules guardrails, and incorrect CR annotations create false verification confidence. Suggested fix: replace the citation with a verified rule that actually covers optional choices/effects, or remove the CR citation from this direct detector contract.

**[LOW]** As a standalone PR, this private-helper direct-call test is too narrow for the contribution bar. Evidence: `crates/engine/src/parser/swallow_check.rs:3623` constructs an empty `ParsedAbilities` and calls `detect_optional_you_may` directly instead of exercising `check_swallowed_clauses` or a full parse path; the existing file already has production-path Optional_YouMay fixtures. Why it matters: this adds review/maintenance surface without proving a user-facing parser regression. Suggested fix: fold it into a consolidated swallow-check test PR with correct CR evidence, or replace it with a production-path regression that would fail for a real Oracle text/card class.

matthewevans:

PR #4647, opened 2026-06-30
test(parser): direct-call regression for Duration_UntilEndOfTurn serialized-AST fallback

### LINKED ISSUE #4646: Add direct-call regression test for detect_duration_until_eot serialized-AST marker fallback
## Summary

`detect_duration_until_eot` in `crates/engine/src/parser/swallow_check.rs` has two
distinct code paths:

1. the structured `def_tree_has_duration` walk over parsed abilities, and
2. a serialized-AST marker fallback (`json_has_any` for `"duration":"UntilEndOfTurn"`
   / `"UntilEndOfCombat"`) that exists specifically because the structured walk
   **cannot descend into `Effect::Token`-granted abilities** (CR 611.2a).

Path (2) is currently only covered indirectly via the full-pipeline
`duration_until_eot_accepts_token_granted_trigger` fixture. There is no direct-call
unit test (unlike the `detect_modal_dynamic_max_dropped_*` detector, which has them)
that locks the fallback contract in isolation. A refactor of the serialized-AST
marker strings or the `json_has_any` gate could regress without any existing test
failing, as long as the token-fixture's parse output happens to stay stable.

## Expected

A direct-call unit test asserting:

- an `"until end of turn"` phrase with **no** structured duration and **no** AST
  marker **fires** `Duration_UntilEndOfTurn`, and
- the same phrase with a serialized `"duration":"UntilEndOfTurn"` marker stays
  **silent** (the token-granted-trigger fallback),

pinning the fallback at the detector boundary.

luciferlive112116:
## Summary

Adds a direct-call regression test for `detect_duration_until_eot` in

`detect_duration_until_eot` falls back to a serialized-AST marker check
(`json_has_any` for `"duration":"UntilEndOfTurn"`) precisely because the structured
`def_tree_has_duration` walk cannot descend into `Effect::Token`-granted abilities
(CR 611.2a). The new test pins both halves of that contract at the detector
boundary:

- an `"until end of turn"` phrase with no structured duration and no AST marker
  **fires** `Duration_UntilEndOfTurn`, and
- the same phrase with a serialized `"duration":"UntilEndOfTurn"` marker stays
  **silent**.

Previously the fallback was only exercised indirectly through the
`duration_until_eot_accepts_token_granted_trigger` full-pipeline fixture.

Closes #4646.

## Test plan

- [ ] `duration_until_eot_fires_without_marker_and_exempts_serialized_duration`
      passes (added in `swallow_check.rs` tests).

### DIFF
--- a/crates/engine/src/parser/swallow_check.rs
+++ b/crates/engine/src/parser/swallow_check.rs
@@ -4178,6 +4178,65 @@ mod tests {
         assert!(!has_swallowed_detector(&parsed, "Duration_UntilEndOfTurn"));
     }
 
+    /// CR 611.2a: direct-call contract for `detect_duration_until_eot`. The
+    /// structured `def_tree_has_duration` walk cannot descend into
+    /// `Effect::Token`-granted abilities, so the detector falls back to a
+    /// serialized-AST marker check. This pins that fallback at the detector
+    /// boundary: an "until end of turn" phrase with no structured duration AND no
+    /// AST marker fires, while the same phrase with a `"duration":"UntilEndOfTurn"`
+    /// marker in the serialized AST stays silent.
+    #[test]
+    fn duration_until_eot_fires_without_marker_and_exempts_serialized_duration() {
+        let parsed = crate::parser::oracle::ParsedAbilities {
+            abilities: Vec::new(),
+            triggers: Vec::new(),
+            statics: Vec::new(),
+            replacements: Vec::new(),
+            extracted_keywords: Vec::new(),
+            modal: None,
+            additional_cost: None,
+            casting_restrictions: Vec::new(),
+            casting_options: Vec::new(),
+            solve_condition: None,
+            strive_cost: None,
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[LOW]** As a standalone PR, this private-helper direct-call test is too narrow for the contribution bar. Evidence: `crates/engine/src/parser/swallow_check.rs:4189` constructs an empty `ParsedAbilities` and calls `detect_duration_until_eot` directly with handcrafted JSON instead of exercising `check_swallowed_clauses` or a real parse fixture; the existing `duration_until_eot_accepts_token_granted_trigger` fixture already covers the production fallback. Why it matters: this creates a new PR/issue/review cycle for a small detector implementation detail without adding meaningful production-path regression coverage. Suggested fix: fold this into a consolidated swallow-check test PR, or make it a production-path regression that demonstrates a real Oracle class would fail without the fallback.

matthewevans:

Issue #2240, opened 2026-06-04
Parser drops ActivateLimit clauses while coverage still marks affected cards supported


mike-theDude:
## Summary

`./scripts/gen-card-data.sh` reports `1` `Swallow:ActivateLimit` diagnostic. The affected card is otherwise marked supported by coverage.

Related umbrella/context issue: #2229 (`Condition_If` swallowed clauses).

## Repro

```bash
./scripts/gen-card-data.sh
target/tool/coverage-report data --all --warning-detector ActivateLimit --warning-limit 5
```

Fresh drilldown:

```text
matched_cards: 1
supported_cards: 1
single_gap_cards: 0
example: Luxurious Locomotive
```

## Expected

Activation-use limits should be represented as typed activation limits/constraints. If the parser cannot represent them, the affected ability should not remain fully supported.

PR #4649, opened 2026-06-30
test(parser): direct-call regression pinning the ActivateLimit marker set

### LINKED ISSUE #4648: Add direct-call regression test pinning the detect_activate_limit marker set
## Summary

`detect_activate_limit` in `crates/engine/src/parser/swallow_check.rs` recognizes a
set of usage-limit marker phrases (CR 605 / CR 602.5e):

- `"activate this ability only once each"`
- `"activate this ability only twice each"`
- `"activate this ability no more than"`
- `"activate only once each turn"`
- `"activate only twice each turn"`

The ActivateLimit regressions section (#2240) currently has a single full-pipeline
fixture (`activate_limit_accepts_crew_once_per_turn_cadence`). There is no
direct-call unit test (unlike `detect_modal_dynamic_max_dropped_*`) that locks the
marker set itself. Trimming or mistyping one of these marker strings could regress
detection silently, since the one full-pipeline fixture only exercises the
`"activate only once each turn"` form.

## Expected

A direct-call unit test asserting that:

- each accepted usage-limit marker form **fires** `ActivateLimit` when no AST limit
  represents it, and
- text carrying no usage-limit phrase (e.g. a plain mana ability) stays **silent**,

pinning the marker set at the detector boundary.

luciferlive112116:
## Summary

Adds a direct-call regression test for `detect_activate_limit` in

The new test pins the usage-limit marker set at the detector boundary (CR 605,
CR 602.5e):

- each accepted marker form (`"activate this ability only once each ..."`,
  `"activate this ability no more than ..."`, `"activate only twice each turn"`)
  **fires** `ActivateLimit` when no AST limit represents it, and
- a plain mana ability with no usage-limit phrase stays **silent**.

Previously the ActivateLimit section (#2240) only had one full-pipeline fixture
(`activate_limit_accepts_crew_once_per_turn_cadence`), which exercises a single
marker form. Trimming or mistyping one of the other marker strings could regress
silently.

Closes #4648.

## Test plan

- [ ] `activate_limit_fires_per_marker_and_silent_without_marker` passes (added in
      `swallow_check.rs` tests).

### DIFF
--- a/crates/engine/src/parser/swallow_check.rs
+++ b/crates/engine/src/parser/swallow_check.rs
@@ -5186,6 +5186,53 @@ this spell's mana cost.\nAttacking creatures get -3/-0 until end of turn.",
         assert!(!has_swallowed_detector(&parsed, "ActivateLimit"));
     }
 
+    /// CR 605 + CR 602.5e: direct-call contract for `detect_activate_limit`. A
+    /// usage-limit phrase with no AST limit representation must fire `ActivateLimit`
+    /// for each accepted marker form, while text carrying no usage-limit phrase
+    /// stays silent. This pins the marker set at the detector boundary, companion
+    /// to the full-pipeline cadence fixture above.
+    #[test]
+    fn activate_limit_fires_per_marker_and_silent_without_marker() {
+        let parsed = crate::parser::oracle::ParsedAbilities {
+            abilities: Vec::new(),
+            triggers: Vec::new(),
+            statics: Vec::new(),
+            replacements: Vec::new(),
+            extracted_keywords: Vec::new(),
+            modal: None,
+            additional_cost: None,
+            casting_restrictions: Vec::new(),
+            casting_options: Vec::new(),
+            solve_condition: None,
+            strive_cost: None,
+            parse_warnings: Vec::new(),
+        };
+
+        for marker in [
+            "activate this ability only once each turn.",
+            "activate this ability no more than twice each turn.",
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The added CR evidence does not support the usage-limit contract being tested. Evidence: `crates/engine/src/parser/swallow_check.rs:5189` cites `CR 605 + CR 602.5e`; `CR 605` is the mana-ability section, and `CR 602.5e` only covers “Activate only as an instant,” not once/twice/no-more-than usage limits. Why it matters: these comments are meant to be rules guardrails, but the cited rules do not describe the behavior under test. Suggested fix: replace the citations with verified rules that actually cover activation restrictions/limits, or remove the CR citation from this detector-boundary test.

**[LOW]** As a standalone PR, this private-helper direct-call test is too narrow for the contribution bar. Evidence: `crates/engine/src/parser/swallow_check.rs:5195` constructs an empty `ParsedAbilities` and calls `detect_activate_limit` directly; the current file already has a full-pipeline cadence fixture for `ActivateLimit`. Why it matters: marker-set unit coverage can be useful, but this standalone PR/issue adds process and maintenance overhead disproportionate to the value. Suggested fix: fold it into a consolidated swallow-check test PR with correct CR evidence, or add production-path fixtures for real usage-limit Oracle forms that are not already covered.

matthewevans:

PR #4651, opened 2026-06-30
test(parser): direct-call regression pinning the ActivateOnlyDuring marker set

### LINKED ISSUE #4650: Add direct-call regression test pinning the detect_activate_only_during marker set
## Summary

`detect_activate_only_during` in `crates/engine/src/parser/swallow_check.rs`
flags restricted-timing activation clauses (CR 605.1c) when neither
`"activate only during"` nor `"activate this ability only during"` is represented
as an activation constraint on a parsed ability.

Unlike the companion `Modal_DynamicMaxDropped` detector, this detector has no
direct-call unit test locking its marker set in isolation. A change to the marker
strings or the `any_ability_has_constraint` exemption could regress detection
silently.

## Expected

A direct-call unit test asserting that:

- each accepted timing-restriction marker form **fires** `ActivateOnlyDuring` when
  no constraint represents it, and
- text carrying no timing-restriction phrase stays **silent**,

pinning the marker set at the detector boundary.

luciferlive112116:
## Summary

Adds a direct-call regression test for `detect_activate_only_during` in

The new test pins the restricted-timing marker set at the detector boundary
(CR 605.1c):

- each accepted marker form (`"activate only during ..."`,
  `"activate this ability only during ..."`) **fires** `ActivateOnlyDuring` when no
  activation constraint represents it, and
- a plain mana ability with no timing-restriction phrase stays **silent**.

Closes #4650.

## Test plan

- [ ] `activate_only_during_fires_per_marker_and_silent_without_marker` passes
      (added in `swallow_check.rs` tests).

### DIFF
--- a/crates/engine/src/parser/swallow_check.rs
+++ b/crates/engine/src/parser/swallow_check.rs
@@ -5186,6 +5186,54 @@ this spell's mana cost.\nAttacking creatures get -3/-0 until end of turn.",
         assert!(!has_swallowed_detector(&parsed, "ActivateLimit"));
     }
 
+    /// CR 605.1c: direct-call contract for `detect_activate_only_during`. A
+    /// restricted-timing phrase ("Activate only during ...") with no activation
+    /// constraint represented on any parsed ability must fire `ActivateOnlyDuring`
+    /// for each accepted marker form, while text with no such phrase stays silent.
+    /// Pins the marker set at the detector boundary.
+    #[test]
+    fn activate_only_during_fires_per_marker_and_silent_without_marker() {
+        let parsed = crate::parser::oracle::ParsedAbilities {
+            abilities: Vec::new(),
+            triggers: Vec::new(),
+            statics: Vec::new(),
+            replacements: Vec::new(),
+            extracted_keywords: Vec::new(),
+            modal: None,
+            additional_cost: None,
+            casting_restrictions: Vec::new(),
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The added CR evidence cites the wrong rule for restricted activation timing. Evidence: `crates/engine/src/parser/swallow_check.rs:5189` says `CR 605.1c` covers “Activate only during ...”; in `docs/MagicCompRules.txt`, `605.1c` is not present, and the `605.1` section is about mana abilities, not timing restrictions. Why it matters: parser rules tests need verified CR evidence, and this annotation points future maintainers at the wrong rule family. Suggested fix: replace it with a verified rule that actually describes activation timing restrictions, or remove the CR citation from this direct detector contract.

**[LOW]** As a standalone PR, this private-helper direct-call test is too narrow for the contribution bar. Evidence: `crates/engine/src/parser/swallow_check.rs:5195` constructs an empty `ParsedAbilities` and calls `detect_activate_only_during` directly, rather than exercising the parser/swallow-check production path on representative Oracle text. Why it matters: a standalone new-contributor PR should add production-relevant regression value, not only pin a private marker scan. Suggested fix: fold it into a consolidated swallow-check test PR with correct CR evidence, or replace it with production-path fixtures for real timing-restriction Oracle forms.

matthewevans:

PR #4653, opened 2026-06-30
test(parser): direct-call regression pinning the Replacement_Instead contract

### LINKED ISSUE #4652: Add direct-call regression test pinning the detect_replacement_instead contract
## Summary

`detect_replacement_instead` in `crates/engine/src/parser/swallow_check.rs` flags a
swallowed CR 614 "instead" replacement when Oracle text contains `" instead"` but no
`ReplacementDefinition` (and none of the structural exemptions) captures it. It
carries several exemption gates — `parsed_has_conditional_modal_max`,
`any_ability_has_target_replacement`, `any_ability_has_exile_parent_rider`,
`any_ability_has_instead_condition`, and `any_text_field_contains(parsed, "instead")`.

These exemptions are well covered by full-pipeline fixtures
(`replacement_instead_accepts_*`), but there is no direct-call unit test (unlike
`detect_modal_dynamic_max_dropped_*`) that pins the base contract: marker present +
no representation + no exemption ⇒ fire; no marker ⇒ silent.

## Expected

A direct-call unit test asserting:

- an unrepresented `"... instead"` replacement clause that escapes every exemption
  **fires** `Replacement_Instead`, and
- text with no `" instead"` marker stays **silent**,

pinning the base contract at the detector boundary.

luciferlive112116:
## Summary

Adds a direct-call regression test for `detect_replacement_instead` in

The new test pins the base contract at the detector boundary (CR 614):

- an unrepresented `"... instead"` replacement clause that escapes every structural
  exemption (target-replacement rider, exile-parent rider, instead-condition
  sub_ability, "instead" text field) **fires** `Replacement_Instead`, and
- text with no `" instead"` marker stays **silent**.

The exemption paths are already covered by full-pipeline `replacement_instead_accepts_*`
fixtures; this adds the missing isolated base-contract coverage.

Closes #4652.

## Test plan

- [ ] `replacement_instead_fires_on_unrepresented_clause_and_silent_without_marker`
      passes (added in `swallow_check.rs` tests).

### DIFF
--- a/crates/engine/src/parser/swallow_check.rs
+++ b/crates/engine/src/parser/swallow_check.rs
@@ -4020,6 +4020,62 @@ mod tests {
         assert!(!has_swallowed_detector(&parsed, "Replacement_Instead"));
     }
 
+    /// CR 614: direct-call contract for `detect_replacement_instead`. An
+    /// " instead" replacement clause that produces no `ReplacementDefinition` and
+    /// is not captured by any of the structural exemptions (target-replacement
+    /// rider, exile-parent rider, instead-condition sub_ability, or an "instead"
+    /// text field) must fire `Replacement_Instead`; text with no " instead" marker
+    /// stays silent. Pins the marker + final-push contract at the detector
+    /// boundary, companion to the full-pipeline exemption fixtures above.
+    #[test]
+    fn replacement_instead_fires_on_unrepresented_clause_and_silent_without_marker() {
+        let parsed = crate::parser::oracle::ParsedAbilities {
+            abilities: Vec::new(),
+            triggers: Vec::new(),
+            statics: Vec::new(),
+            replacements: Vec::new(),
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[LOW]** As a standalone PR, this private-helper direct-call test is too narrow for the contribution bar. Evidence: `crates/engine/src/parser/swallow_check.rs:4031` constructs an empty `ParsedAbilities` and calls `detect_replacement_instead` directly; the same section already has production-path `replacement_instead_accepts_*` fixtures for the structural exemptions. Why it matters: this adds a PR and issue for a private detector base case, but does not prove a real parser/card regression beyond the implementation detail. Suggested fix: fold it into a consolidated swallow-check test PR, or add a production-path fixture where real Oracle text should intentionally emit `Replacement_Instead` when unrepresented.

matthewevans:

PR #4643, opened 2026-06-30
fix(parser): controller-scoped "you lost life this turn" intervening-if condition

minion1227:
## Problem

The gained/lost-life intervening-if condition matrix was missing exactly one subject: the controller-scoped **"you lost [N or more] life this turn"**. Opponent (`parse_opponent_lost_life_this_turn`) and any-player (`parse_player_lost_life_this_turn`) "lost life" parsed, and the controller *gained* sibling parsed too — but `"you lost ..."` silently dropped its condition (left `None`).

As a result, phase-trigger intervening-ifs such as **The Book of Vile Darkness** lost **both** the gate and their effect:

> At the beginning of your end step, if you lost 2 or more life this turn, create a 2/2 black Zombie creature token.

## Fix

Add `parse_you_lost_life_this_turn` as the exact mirror of `parse_you_gained_life_this_turn`, wired into `parse_life_history_condition`. It resolves via the existing `QuantityRef::LifeLostThisTurn { player: Controller }` — **no new variants**. Handles both `"you lost N or more life this turn"` (≥ N) and the bare `"you lost life this turn"` (≥ 1) forms, plus the `"you've lost ..."` contraction.

CR 119.3 (life-total changes) + CR 603.4 (intervening-if).

## Tests

Two inline parser regressions in `oracle_trigger.rs`:
- `trigger_if_you_lost_2_or_more_life` — The Book of Vile Darkness; asserts condition `LifeLostThisTurn{Controller} ≥ 2` **and** `execute.is_some()` (the dropped effect).
- `trigger_if_you_lost_life_this_turn_no_minimum` — bare form → `≥ 1`.

Both pass; engine compiles clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4654, opened 2026-06-30
fix(parser): controller-scoped "you lost life this turn" intervening-if condition

minion1227:
The gained/lost x you/opponent/player condition matrix was missing exactly
one subject: the controller-scoped "you lost [N or more] life this turn".
Opponent and any-player "lost life" parsed, and controller "gained life"
parsed, but "you lost ..." silently dropped its condition (left None), so
phase-trigger intervening-ifs such as The Book of Vile Darkness ("if you lost
2 or more life this turn, create a 2/2 black Zombie") lost both the gate and
their effect.

Add parse_you_lost_life_this_turn as the mirror of
parse_you_gained_life_this_turn, wired into parse_life_history_condition.
Resolves via the existing QuantityRef::LifeLostThisTurn { Controller } — no
new variants.

CR 119.3 + CR 603.4.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -4309,6 +4309,10 @@ fn parse_life_history_condition(input: &str) -> OracleResult<'_, StaticCondition
         parse_player_lost_life_this_turn,
         // "you gained life this turn" / "you gained N or more life this turn"
         parse_you_gained_life_this_turn,
+        // "you lost life this turn" / "you lost N or more life this turn" — the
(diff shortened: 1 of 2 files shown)


matthewevans:
Closing this as a duplicate/no-op PR: the same branch/head was already merged through #4643 as squash commit `32577d50f` on `main`, and `git diff origin/main HEAD` for this PR is empty.

I am not using this duplicate as a second merge vehicle. If there is a follow-up parser normalization issue to handle, it should come as a fresh PR against current `main` with only that delta.

Issue #4385, opened 2026-06-26
Risen reef — Comes back to hand after it's cast ad infinitum.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1519783596926308562/1519783596926308562

**Thread:** Risen reef
**Cards:** _none detected_
**Parser status:** no_card
**Extraction confidence:** 0.60

## Summary
Comes back to hand after it's cast ad infinitum.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1519783596926308562:1519783596926308562:0` · discord: `1519783596926308562/1519783596926308562`</sub>

mike-theDude:
Card is Risen Reef: "Whenever this creature or another Elemental you control enters, look at the top card of your library. If it's a land card, you may put it onto the battlefield tapped. If you don't put the card onto the battlefield, put it into your hand." The trigger is fully parsed as a Dig(1) → conditional-battlefield-else-hand chain with no `Unimplemented`/`Unknown` entries, and nothing in the Oracle text or parsed effect chain describes Risen Reef itself returning to hand — only the revealed library card does. The report text ("comes back to hand after it's cast ad infinitum") is too vague to tell whether the reporter means the dug card repeatedly hitting hand (expected behavior whenever multiple Elementals enter), an actual self-bounce of Risen Reef, or an unrelated interaction with other cards in play; no concrete game state, decklist, or turn sequence was provided. Given the ambiguity, this needs a reproducible game state (what else was on the battlefield/stack, exact sequence of triggers) before a root cause in the zone-change/trigger-loop machinery can be identified.

*Status: needs-repro — classifier: cannot-determine*

PR #4655, opened 2026-06-30
test(engine): regression coverage for Risen Reef's land-or-hand ETB cycle (#4385)

claytonlin1110:
## Summary
- Investigates Discord report #4385: "Risen Reef ... Comes back to hand after it's cast ad infinitum."
- The suspected mechanism — the `Dig -> ChangeZone(Battlefield) -> ChangeZone(Hand)` chain's `ParentTarget` resolving to the trigger source (Risen Reef itself) instead of the dug card — does **not** reproduce against current `main`. `apply_parent_chain_context` already injects the revealed card id from `state.last_revealed_ids` into the sub-ability's targets before its condition or resolution runs, falling back to self-target only when nothing was revealed (an empty-library edge case, separately hardened in #4637).
- Adds permanent end-to-end regression tests (via `GameScenario`/`GameRunner`, real cast/trigger/combat pipeline) locking in correct behavior for the "look at the top card, if it's a land you may put it onto the battlefield [tapped], otherwise put it into your hand" Oracle-text cycle:
  - Risen Reef: cast-from-hand self ETB, sibling-Elemental ETB, accept branch (land enters tapped), decline branch (card to hand) — never bounces Risen Reef itself either way.
  - Rulik Mons, Warren Chief: same skeleton but attack-triggered with a token-creation rider instead of a hand fallback — confirms the fix generalizes across trigger mode and rider effect type.
  - Fecund Greenshell: structural self/other condition phrasing variant.
- CR citations (608.2c, 701.20e, 614.1c) verified against `docs/MagicCompRules.txt` before being added.

No production code changes — investigation found the engine already correct for this card family; this PR is test coverage only, per the repo's "test the building block" convention.

## Test plan
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy -p engine --lib --tests -- -D warnings` clean
- [x] `cargo test -p engine --lib risen_reef` — 8/8 new tests pass
- [x] Full `cargo test -p engine --lib` run clean (14k+ tests, 0 failures) prior to rebase onto latest `main`

### DIFF
--- a/crates/engine/src/game/mod.rs
+++ b/crates/engine/src/game/mod.rs
@@ -88,6 +88,9 @@ pub mod merge;
 #[cfg(test)]
 #[path = "omnath_tests.rs"]
 mod omnath_tests;
+#[cfg(test)]
+#[path = "risen_reef_tests.rs"]
+mod risen_reef_tests;
 // Tests for `merge` live in a sibling file (declared here, not in `merge.rs`,
 // so `merge.rs` stays implementation-only).
 pub mod archenemy;
--- a/crates/engine/src/game/risen_reef_tests.rs
+++ b/crates/engine/src/game/risen_reef_tests.rs
@@ -1,0 +1,344 @@
+//! Regression tests for issue #4385 (Discord: "Risen Reef ... Comes back to
+//! hand after it's cast ad infinitum").
+//!
+//! The reported hypothesis was that `TargetFilter::ParentTarget` in the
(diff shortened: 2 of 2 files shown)


matthewevans:
Closing this rather than iterating: this is a standalone test-only PR, and the PR body says the suspected failure does not reproduce on current main. We are not accepting standalone test-only coverage PRs from the sweep; if this area needs work, it should come with a substantive engine/parser fix or a maintainer-requested consolidated regression suite.

PR #4660, opened 2026-06-30
fix(data): auto-unlock released sets and infer MSH legalities (#4365)

### LINKED ISSUE #4365: Marvel Super Heroes Cards Showing Up Banned — All cards from the new Marvel Super Heroes set (which are legal following…
Reported in Discord: https://discord.com/channels/1485498006781427802/1519416385590132966/1519416385590132966

**Thread:** Marvel Super Heroes Cards Showing Up Banned
**Cards:** _none detected_

## Summary
All cards from the new Marvel Super Heroes set (which are legal following the recent pre-release) are showing up banned.

---
<sub>report_id: `discord:1519416385590132966:1519416385590132966:0` · discord: `1519416385590132966/1519416385590132966`</sub>

claytonlin1110:
## Summary

- Auto-ungate sets listed in `GATED_SETS` once their MTGJSON `releaseDate` has passed, so Marvel Super Heroes (MSH/MSC/TMSH) cards are no longer marked Banned in every format if the env var is not manually cleared after release.
- Synthesize format legalities during card-data generation when AtomicCards has empty legality data for newly released sets (common during the release window).
- Pin the release-gate evaluation date via `GATED_SETS_AS_OF` in `gen-card-data.sh` for deterministic CI/local generation.

Fixes #4365.

## Test plan

- [x] `cargo test -p engine --lib set_catalog set_gating legality_inference`
- [x] `cargo test -p engine --test issue_4365_msh_legality`
- [ ] Regenerate card-data with `./scripts/gen-card-data.sh` and confirm MSH cards show Standard-legal (not banned) in deck builder
- [ ] Verify deploy card-data job picks up the change on merge


matthewevans (changes requested):
Thanks for taking a swing at this. I think there is a useful fix in here, but I don't think we should merge the legality inference layer in its current shape.

The release-gating part looks valuable: using MTGJSON `SetList.releaseDate` to stop treating a released set as gated is the right layer for this problem, and keeping unknown/missing set metadata gated is appropriately conservative.

The legality inference part is where I think this PR gets too broad. If the issue is that MSH was still using stale/pre-release MTGJSON data, the fix should be to refresh authoritative card data once the set is released and make the release gate respect MTGJSON release dates. Once MTGJSON is refreshed, the legalities should come from the source data, not from Phase guessing based on set type.

Right now this adds a local, global legality inference system: when MTGJSON legalities are empty, Phase infers format legality from set type and exports that into card data. Those exported legalities are then consumed by search and deck validation, so a false `Legal` value becomes user-visible behavior. That is a much larger blast radius than the release-gating bug requires.

The current templates also look unsafe. For example, commander products are inferred legal in Brawl/Standard Brawl, and masters-style products are inferred legal in formats like Modern, Pioneer, Brawl, and Standard Brawl. Those are not format guarantees we should invent locally. If upstream legalities are stale, we should refresh or surface that staleness; we should not silently fabricate legality data.

I would support narrowing this PR to:

1. auto-unlock configured gated sets once `SetList.releaseDate` has passed;
2. keep missing/unknown set metadata conservative;
3. add evidence/tests around the MSH release-gating path;
4. avoid adding broad legality inference unless there is a separate design showing that upstream legalities remain missing after refresh and exactly which formats can be inferred safely.

So my request is: please remove or substantially narrow the legality inference system from this PR. The release-date gate cleanup is useful, but the current inference layer is an under-designed authority for format legality, and I don't think that belongs in this bug fix.

matthewevans (approved):
Reviewed the revised release-gating implementation at b0ccd10d. The broad legality inference layer from the earlier review has been removed; the remaining change is scoped to resolving configured gated sets through MTGJSON SetList.releaseDate, conservatively keeping unknown/missing metadata gated, and preserving MTGJSON as the source of real legalities. I also checked the duplicate/security gates and local merge-tree; the PR is suitable for merge-when-ready.

PR #4664, opened 2026-06-30
feat(engine): support "whenever you activate a loyalty ability of [planeswalker]" triggers

GildardoDev:
## Anchored on

The "Whenever you activate a loyalty ability of [planeswalker]" triggered-ability class, which parsed to `TriggerDefinition { mode: Unknown(...) }` (unclassified). The 4-card class:
- Chandra's Regulator, Keral Keep Disciples — "a Chandra planeswalker" (subtype-scoped)
- Elspeth's Talent, Rowan's Talent — "enchanted planeswalker" (Aura-host-scoped)

## Problem

These triggers had no engine classification, so none of the four cards fired.

## Root cause

There was no way to recognize "a player activated a *loyalty* ability." `GameEvent::AbilityActivated` is emitted for every non-mana activation and carries no loyalty discriminator, and there was no `TriggerMode` for it.

## Fix (discriminator field, not a new event)

- **`ActivatedAbilityKind { Normal, Loyalty }`** added as a `#[serde(default)]` field on the existing `GameEvent::AbilityActivated` (CR 602.2 + 606.2). The kind is computed (via `is_loyalty_ability_cost`) at every `AbilityActivated` emission site — including the **targeted** loyalty-finalize path (`casting_targets.rs`, where most loyalty abilities resolve), so the common case actually fires. This reuses the existing event rather than a parallel `LoyaltyAbilityActivated` event, so `copy_spell` and the anaphor resolvers (which already match `AbilityActivated`) keep working unchanged — "copy that ability" still copies.
- **`TriggerMode::LoyaltyAbilityActivated`** + `match_loyalty_ability_activated`: fires when `kind == Loyalty`, the activator controls the trigger source (CR 602.2a), and the source planeswalker matches `valid_card`. The two filter shapes reuse existing `TargetFilter` slots: `Typed(Planeswalker + Subtype)` for "a Chandra planeswalker", `AttachedTo` for "enchanted planeswalker" (the Bracers precedent). The planeswalker subtype is parsed via the existing `parse_subtype`, so the class covers any planeswalker subtype (Chandra is data, not code).
- Parser arm in `try_parse_ability_activation_trigger` (nom combinators only).

No new `GameEvent` variant; the new `TriggerMode` is wired through the matcher dispatch + registry, both trigger-index maps, the analysis axis, and serde round-trip. CR 205.3j (planeswalker subtypes), 602.2a, 603.2, 606.1/.2/.3/.5.

## Tests

- **Targeted loyalty fires (the key case):** a targeted loyalty ability of a Chandra planeswalker fires Keral Keep Disciples' trigger — revert-verified (hardcoding `kind: Normal` at the targeted path makes it fail).
- **"copy that ability" copies at runtime** (Rowan's Talent / Chandra's Regulator) — proving the discriminator field preserves the copy path.
- Non-Chandra planeswalker → no fire; non-loyalty activated ability (`kind: Normal`) → no fire; enchanted-host fires only for the host.
- All 4 cards parse to `LoyaltyAbilityActivated` with the correct `valid_card` and a non-Unimplemented effect.
- serde round-trip of the new mode; back-compat default (legacy `AbilityActivated` without `kind` → `Normal`).

## Verification

`cargo fmt`/`check -p engine`/clippy `-D warnings`/parser combinator gate all clean; `cargo check -p mtgish-import` clean (no cross-crate ripple); ninjutsu / copy_spell / loyalty / broad trigger suites stay green; the 4 new runtime tests + the matcher/parser/serde unit tests pass.

Model: claude-opus-4-8
Tier: Frontier


matthewevans (changes requested):
**[LOW] The loyalty-activation matcher cites the wrong CR authority for “you.”** Evidence: `crates/engine/src/game/trigger_matchers.rs:4094`/`:4115` cite `CR 602.2a` for “you = controller of the trigger source”, but `docs/MagicCompRules.txt` defines “you” in `CR 109.5`; `CR 602.2a` is about announcing an activated ability and the ability object’s controller. Why it matters: CR annotations are treated as verified rules evidence, and this points future maintainers at the wrong rule for the trigger-source controller check. Suggested fix: cite `CR 109.5` for “you,” keeping `CR 602.2a` only if the annotation is specifically about the activated ability’s controller provenance.

**[LOW] The “enchanted planeswalker” parser tests cite an unrelated Aura-entry rule.** Evidence: `crates/engine/src/parser/oracle_trigger.rs:4933`, `:27720`, and `:27738` cite `CR 303.4f`, but `docs/MagicCompRules.txt` has the relevant attachment-reference rules at `CR 303.4b` (“enchanted”) and `CR 303.4m` (“enchanted [object or player]”). Why it matters: the parser behavior is correct, but the annotation sends maintainers to the wrong rule section. Suggested fix: update these comments to cite `CR 303.4b`/`CR 303.4m` as appropriate.

GildardoDev:
Thanks, both fixed.

- **CR 109.5 for "you":** updated the matcher's annotations (`trigger_matchers.rs`) to cite CR 109.5 ("you"/"your" refer to the object's controller) for the trigger-source controller check, dropping the CR 602.2a reference there.
- **CR 303.4m for "enchanted planeswalker":** updated the parser arm + its tests (`oracle_trigger.rs`) to cite CR 303.4m (an ability referring to the "enchanted [object]"), replacing the incorrect CR 303.4f (Aura entry).

Also took gemini's adjacent suggestions: same CR 109.5 correction, and simplified the kind derivation at the targeted-loyalty emission site (`is_some_and` instead of the nested `map_or`). No behavior change; the runtime tests are unaffected. Re-requesting review.

matthewevans (changes requested):
**[MED] Targeted `[−X]` loyalty abilities still emit `Normal` after target selection.** Evidence: `[−X]` loyalty abilities are modeled through the X-cost path (`crates/engine/src/parser/oracle.rs:4388`), and that path clears `pending_act.activation_cost` before target selection at `crates/engine/src/game/casting_costs.rs:2620`; target finalization then decides `ActivatedAbilityKind::Loyalty` by checking `pending.activation_cost` at `crates/engine/src/game/casting_targets.rs:453`. Why it matters: targeted `[-X]` loyalty abilities lose their loyalty kind during target selection, so “whenever you activate a loyalty ability” triggers miss this subclass. Suggested fix: preserve the activation kind on `PendingCast`, or classify targeted finalization from `activated_ability_kind(state, pending.object_id, ability_index)` instead of the cleared cost.

**[LOW] One new CR citation still uses the wrong Aura rule for “enchanted planeswalker.”** Evidence: `crates/engine/src/game/trigger_matchers.rs:5857` cites `CR 303.4f`, but `docs/MagicCompRules.txt` shows `CR 303.4f` is an Aura entering-battlefield choice rule; `CR 303.4b`/`CR 303.4m` cover “enchanted” attachment references. Why it matters: the parser comments were fixed, but this matcher-test annotation still points maintainers at the wrong rule. Suggested fix: change this comment to `CR 303.4b + CR 303.4m`.

GildardoDev:
both fixed.

**[MED] targeted `[−X]` loyalty.** You're right — the X-cost path clears `pending.activation_cost` before the targeted finalize, so deriving the kind from it mis-classified `[−X]` abilities as `Normal`. The targeted finalize (`casting_targets.rs`) now classifies via `activated_ability_kind(state, pending.object_id, ability_index)` — the source ability's *printed* cost, which is stable. `is_loyalty_ability_cost` recognizes both `AbilityCost::Loyalty` and the `[−X]` `RemoveCounter { X loyalty counters }` form, so `[−X]` loyalty abilities now classify as `Loyalty`. This mirrors the non-targeted path in `planeswalker.rs`. Added a test asserting a `[−X]` loyalty ability's printed cost classifies as `Loyalty` (and a non-loyalty sibling as `Normal`).

**[LOW] CR citation.** The matcher-test annotation now cites `CR 303.4b` + `CR 303.4m` (the "enchanted" attachment-reference rules), matching the parser-side fix.

Re-requesting review.

matthewevans (approved):
Re-reviewed current head after branch update (`7fb1a925`). The prior blockers are addressed.

Evidence checked:
- targeted activation finalization now classifies `AbilityActivated.kind` from the source ability's printed cost via `planeswalker::activated_ability_kind`, so targeted `[-X]` loyalty abilities do not become `Normal` after target selection.
- the non-targeted activation paths use the same classifier, keeping the event model consistent.
- the enchanted-planeswalker CR citation was corrected to the Aura host rules, and parser/runtime tests cover Chandra subtype matching, non-loyalty negatives, and enchanted-host positive/negative cases through the scenario runner.

This is a feature-surface engine addition and is ready for merge-when-ready.

PR #4671, opened 2026-06-30
feat(engine): die-roll result-filter triggers ("whenever you roll a [N]")

GildardoDev:
## Anchored on

The "Whenever you roll a [specific result]" die-roll triggered-ability class, which parsed to `TriggerMode::Unknown(...)`. Real forms: exact ("roll a 1"), disjunction ("roll a 1 or 2"), and threshold ("roll a 3 or higher" / "or more"). Cards: Complaints Clerk, Dee Kay (Finder of the Lost), Night Shift of the Living Dead, Monoxa (Midway Manager), Mr. House (President and CEO), Atomwheel Acrobats.

## Problem

The die-roll trigger event (`GameEvent::DieRolled`), mode (`RolledDie`/`RolledDieOnce`), and matcher (`match_rolled_die`) are all shipped, but the matcher only filtered by die **sides**, never by the rolled **result** — so result-specific triggers were unclassified and never fired.

## Fix (mirrors the shipped `coin_flip_result` pattern)

- **`DieResultFilter { Exact(Vec<u8>), AtLeast(u8) }`** — `Exact` covers a single face ("roll a 1") and disjunctions ("roll a 1 or 2"); `AtLeast` covers the GE threshold ("roll a N or higher"). A set is needed for the disjunction, so a single `(Comparator, u8)` can't express it; this is the same shape as the existing `CoinFlipResult` filter.
- **`die_result: Option<DieResultFilter>`** field on `TriggerDefinition` (`#[serde(default, skip_serializing_if = ...)]`, like `coin_flip_result`).
- **`match_rolled_die`** now reads `DieRolled.result` and applies the filter (`Exact` → result ∈ set; `AtLeast` → result >= n). Per CR 706.7, a non-numeric (planar) roll (`result: None`) never satisfies a numeric filter, while a `None` filter (e.g. existing "roll a d20") is unaffected.
- **Parser:** `parse_die_roll_result` (nom only) parses "a N" / "a N or M" / "a N or higher"/"or more", threaded through the die-roll trigger parser; reuses `RolledDieOnce` (no new mode).

No new event, no new mode. CR 706.2 (die-roll result) and CR 706.7 (planar-die exclusion).

## Scope (honest)

This classifies and fires the die-result **trigger** for the class. The trigger now fires on exactly the right rolled result. Several of the named cards have *additional* effect clauses that are pre-existing `Unimplemented`/partial gaps unrelated to this change — e.g. Monoxa's "if the roll was N or higher" escalations, Mr. House's "6+ instead" + its activated ability, Dee Kay's roll-2 "and you gain 1 life", Night Shift's roll-modifier line. Those are separate, known gaps; this PR does not claim to complete those cards' effects, only to classify the trigger correctly.

## Tests

Parser: each form → `RolledDieOnce` + correct `die_result` (`Exact([1])`, `Exact([1,2])`, `AtLeast(3)`, `AtLeast(4)`); "roll a die" / "one or more dice" still parse with `die_result: None` (no regression). Matcher (the discriminating matrix): `Exact([1])` fires on result 1 not 2; `AtLeast(3)` fires on 3/6 not 2; `Exact([1,2])` fires on 1 and 2 not 3; numeric filter does not fire on a `None` (planar) result. Runtime: a "roll a 1" trigger fires only on a rolled 1.

## Verification

`cargo fmt`/`check -p engine`/clippy `-D warnings`/parser combinator gate all clean; `cargo check -p mtgish-import` clean (no cross-crate ripple); the new parser/matcher/runtime tests pass; the existing `rolled_die` + `coin_flip` suites stay green.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -2,8 +2,9 @@ use std::collections::HashMap;
 use std::sync::LazyLock;
 
 use crate::types::ability::{
-    AbilityTag, CoinFlipResult, ControllerRef, DamageKindFilter, DestinationConstraint, EffectKind,
-    OriginConstraint, TargetFilter, TargetRef, TriggerDefinition, TypedFilter,
+    AbilityTag, CoinFlipResult, ControllerRef, DamageKindFilter, DestinationConstraint,
+    DieResultFilter, EffectKind, OriginConstraint, TargetFilter, TargetRef, TriggerDefinition,
+    TypedFilter,
 };
 use crate::types::events::{GameEvent, PlayerActionKind};
 use crate::types::game_state::GameState;
@@ -3639,11 +3640,27 @@ pub(super) fn match_rolled_die(
     state: &GameState,
 ) -> bool {
     if let GameEvent::DieRolled {
-        player_id, sides, ..
+        player_id,
+        sides,
+        result,
     } = event
     {
         if trigger.die_sides.is_some_and(|required| required != *sides) {
             return false;
+        }
+        // CR 706.2: result-face filter. CR 706.7: a planar (non-numeric) roll has
+        // result == None and never satisfies a numeric filter; a None filter is unaffected.
+        if let Some(filter) = &trigger.die_result {
+            let Some(rolled) = *result else {
+                return false;
+            };
+            let ok = match filter {
+                DieResultFilter::Exact(faces) => faces.contains(&rolled),
+                DieResultFilter::AtLeast(min) => rolled >= *min,
+            };
+            if !ok {
+                return false;
+            }
         }
         valid_player_matches(trigger, state, *player_id, source_id)
     } else {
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Reviewed current head `978c46008f60828fe011255c4bd80eff9bb4d318`.

The result filter is at the right seam: `TriggerDefinition` carries the typed die-result predicate, the die-roll trigger parser composes the existing actor/object axes and distinguishes `a d20`/`a die` before result faces, and `match_rolled_die` filters numeric `DieRolled` events while leaving planar/non-numeric rolls out of numeric result matches. The production trigger test drives `process_triggers` and would fail if the matcher ignored the result face.

I considered Gemini's `BTreeSet` suggestion; for this small parsed face list, `Vec<u8>` is adequate and keeps the parser/order representation simple without introducing extra machinery. CI is still pending, but that is not a review blocker for merge-queue auto-merge.

PR #4675, opened 2026-06-30
Add Aven Mindcensor

bruno-branco-brito:
## Summary
Adds engine support for **Aven Mindcensor** by closing its one coverage gap — the static *"If an opponent would search a library, that player searches the top four cards of that library instead."* (Flash and Flying already work.)

Implements the **class** of top-N library-search restrictions, not just this card, via a new data-carrying `StaticMode::RestrictLibrarySearchToTop { who: ProhibitionScope, count: u32 }` (parameterized by searcher scope and count — not hardcoded to opponents/4). The line is routed to **Priority-7 static dispatch** (it leads with "…would search…", which the Priority-8 `"would "` replacement classifier would otherwise swallow — and there is no search *replacement* event), parsed with nom combinators (mirroring The Valeyard's "If an opponent would …" routing), and enforced **inline in `search_library::resolve()`** by capping the **library** candidate set to the top `count` cards for searchers matched by `who` — mirroring the existing `CantSearchLibrary` seam (`is_search_muzzled`).

Rules-correct per CR 701.23f: the restriction keys on **who searches** (an opponent of the source's controller), applies to **any** library that player searches (their own, the controller's, or a third player's), search-triggers and per-turn tracking still fire, and the whole library still shuffles. Multi-zone tutors cap only the library portion; a library with fewer than `count` cards is searched in full.

## Files changed
- `crates/engine/src/types/statics.rs`
- `crates/engine/src/parser/oracle_static/restriction.rs`
- `crates/engine/src/parser/oracle_classifier.rs`
- `crates/engine/src/game/effects/search_library.rs`
- `crates/engine/src/game/coverage.rs`

## CR references
- `CR 701.23f` — replacing "search a library" with "search a portion": other instructions still apply, search-triggers still fire (the rule's example names Aven Mindcensor)
- `CR 701.23` — Search
- `CR 614.1a` — "instead" → replacement effect
- `CR 609.3` — "do as much as possible" (multi-zone tutor caps only the library portion)

## Track
Non-developer

## LLM
Model: claude-opus-4-8[1m]
Thinking: medium

## Verification
Local verification skipped — see CI status checks.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

## Tier
Frontier


matthewevans (approved):
Reviewed current head `c673bd66d8f125525e0a4ea9090eff6a4e218779`. The Aven Mindcensor implementation is not test-only: it adds a parameterized searcher-scoped top-N static, parses the Aven class with a count parameter, enforces it in the central SearchLibrary resolver before SearchChoice construction, and includes runtime coverage for opponent search, controller self-search, searcher-vs-owner divergence, multi-zone searches, search tracking, and multiple sources.

PR #4676, opened 2026-06-30
Fix bullet-line modal triggers binding "that player" to the wrong player (Grenzo)

### LINKED ISSUE #2346: Grenzo, Havoc Raiser: trigger targets the player who dealt damage instead of the player who received it
## Reported behavior

When Grenzo, Havoc Raiser's trigger fires after combat damage, both modes (Goad and Exile top card) affect the player who dealt the damage (Grenzo's controller) rather than the player who received the combat damage.

## Oracle text (verified from card-data.json)

```
Whenever a creature you control deals combat damage to a player, choose one —
• Goad target creature that player controls.
• Exile the top card of that player's library. Until end of turn, you may cast that card and you may spend mana as though it were mana of any color to cast that spell.
```

## Expected vs actual

- **Expected**: 'That player' refers to the player who received combat damage (the opponent). Goad targets a creature the damaged player controls; ExileTop exiles from the damaged player's library.
- **Actual**: The Goad mode targets a creature the controller (Grenzo's owner) controls (`controller: You` in parsed data). ExileTop targets `ParentTarget` which resolves to the controller's library.

## Notes

Confirmed in card-data.json: `mode_abilities[0].effect.target.controller == "You"` (should be opponent/damaged player). The trigger mode is `DamageDone` and 'that player' should be bound to the event's damaged player, not the source's controller.

## Source

https://discord.com/channels/1485498006781427802/1512209004124700754/1512209004124700754

---
<sub>report_id: `discord:1512209004124700754:1512209004124700754:0` · via bug-triage</sub>

GildardoDev:
**Anchored on:** #2346

**Problem:** Grenzo, Havoc Raiser reads "Whenever a creature you control deals combat damage to a player, choose one — • Goad target creature that player controls. • Exile the top card of that player's library. Until end of turn, you may cast that card...". Both modes goaded and exiled from the wrong player: "that player" resolved to Grenzo's own controller instead of the player who was dealt combat damage.

**Root cause:** Bullet-line modal triggers route through the `TriggeredModal` block path, which never threaded the trigger condition's relative-player scope into the mode bodies. With no scope, "that player controls" / "that player's library" fell back to `ControllerRef::You` (oracle_target.rs). The scope-correct machinery already existed and was wired into the inline "; or" modal path via `relative_player_scope_for_condition`, but the bullet-line path bypassed it. Fixing the parser then exposed a second defect: the triggered-modal mode-choice handler computed the chosen mode's legal targets after `current_trigger_event` had already been cleared, so `TriggeringPlayer` resolved to nothing and the ability died with "No legal targets available".

**Fix:**
- Parser (oracle_modal.rs, `TriggeredModal` arm): derive the scope from the single-authority `relative_player_scope_for_condition` and thread it through a scope-aware modal builder, matching the inline path. The existing `modal_subject` (`TriggeringSource` for "that creature") threading is preserved alongside it, not replaced.
- Runtime (engine_modes.rs, `handle_triggered_mode_choice`): re-establish the trigger event context around legal-target computation and restore it on every return path, so `TriggeringPlayer` resolves to the damaged player.

Both changes route through shared authorities, so any bullet-line modal combat-damage / attack / target-player trigger using "that player" anaphora is fixed, not just Grenzo.

**Tests:**
- Parser (oracle_trigger.rs, `damage_done_bullet_modal_uses_triggering_player_for_that_player`): parses Grenzo's bullet-line text; asserts the Goad mode's `controller` and the exile mode's `player` are `TriggeringPlayer`, not `You`. Reverting the scope line reproduces `Some(You)`.
- Runtime (grenzo_havoc_raiser_modal_2346.rs, GameScenario): P0's attacker deals combat damage to P1; the Goad mode's only legal targets are P1's creatures and never P0's. Reverting the parser fix offers P0's creatures; reverting only the runtime fix reproduces "No legal targets available".

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, and ./scripts/check-parser-combinators.sh all clean. Regression green: modal 145, triggered_modal 8, damage_done 13, broad trigger 2008, plus the new runtime test. The inline "; or" modal path is unchanged. No new engine variant.

Closes #2346.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/game/engine_modes.rs
+++ b/crates/engine/src/game/engine_modes.rs
@@ -436,8 +436,26 @@ fn handle_triggered_mode_choice(
         .pending_trigger
         .take()
         .ok_or_else(|| EngineError::InvalidAction("No pending trigger".to_string()))?;
+    // CR 603.2 + CR 109.4: Re-establish the trigger event context for
+    // the duration of mode-target computation. The modal was paused for mode
+    // choice (`trigger_dispatch`) AFTER restoring the context to its pre-dispatch
+    // value, so `state.current_trigger_event` is now unset. A chosen mode body
+    // whose target filter references the triggering event — e.g. Grenzo, Havoc
+    // Raiser's "Goad target creature that player controls" (`ControllerRef::
+    // TriggeringPlayer`) — must resolve "that player" to the damaged player while
+    // its legal targets are computed here, exactly as the dispatch-time
+    // `filter_modes_by_target_legality` did. Without this, the Goad slot finds no
+    // legal target and `build_target_slots_labelled` errors ("No legal targets
+    // available"). Restored on every return path below.
+    let trigger_event_batch = state.pending_trigger_event_batch.clone();
+    let mode_context_snapshot = triggers::push_trigger_event_context(
+        state,
+        trigger.trigger_event.as_ref(),
+        &trigger_event_batch,
+        trigger.subject_match_count,
+    );
     // CR 700.2 / CR 700.2b: slots + per-mode labels built together (Finding 4).
-    let (target_slots, mode_labels) = build_target_slots_labelled(
+    let (target_slots, mode_labels) = match build_target_slots_labelled(
         state,
         &mode_abilities,
         &indices,
@@ -447,7 +465,13 @@ fn handle_triggered_mode_choice(
         &resolved.context,
         // CR 107.1b: Triggered abilities don't use a chosen X here.
         None,
-    )?;
+    ) {
+        Ok(pair) => pair,
+        Err(err) => {
+            triggers::restore_trigger_event_context(state, mode_context_snapshot);
+            return Err(err);
+        }
+    };
     let target_constraints = target_constraints_from_modal(&modal);
 
     trigger.ability = resolved;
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**[LOW]** The current head repeatedly cites the wrong CR rule for the "that player"/damaged-player binding. Evidence: `docs/MagicCompRules.txt:2618` shows `CR 603.7c` is about delayed triggered abilities continuing to affect a particular object if it changes characteristics; it is not the rule establishing the damaged player as `TriggeringPlayer` for Grenzo's combat-damage trigger. This appears in the new comments/tests around `engine_modes.rs`, `oracle_modal.rs`, `oracle_trigger.rs`, and `gr​enzo_havoc_raiser_modal_2346.rs`. The implementation direction looks reasonable because it reuses the existing `relative_player_scope_for_condition`/`ParseContext.relative_player_scope` path, but the repo requires verified CR annotations, so please replace the incorrect `CR 603.7c` citations with verified rule references or remove that specific citation where no exact CR rule applies.

Also note that GitHub currently reports this PR as `DIRTY`/conflicting against `main`; I did not approve or enqueue it from the sweep.

matthewevans (changes requested):
**[LOW]** This is still blocked at current head `4daabedf`: the incorrect `CR 603.7c` citations remain. Evidence: `docs/MagicCompRules.txt:2618` says `CR 603.7c` is about delayed triggered abilities continuing to affect a particular object after characteristic changes; it is not the damaged-player / `TriggeringPlayer` binding rule. Current-head examples include `crates/engine/src/game/engine_modes.rs`, `crates/engine/src/parser/oracle_modal.rs`, the new `grenzo_bullet_modal_tests` block in `crates/engine/src/parser/oracle_trigger.rs`, and `crates/engine/tests/grenzo_havoc_raiser_modal_2346.rs`. The implementation still looks directionally reasonable because it reuses `relative_player_scope_for_condition` / `ParseContext.relative_player_scope`, but the CR annotations need to be corrected or narrowed to verified rules before this can be approved.

GildardoDev:
Good catch, thanks. Replaced the incorrect CR 603.7c citations in the new comments and tests: kept the verified co-citations (CR 109.4 control, CR 115.1 targets, CR 506.2 defending player, CR 603.2 trigger conditions) and dropped 603.7c where no exact CR governs the "that player" anaphora, since that binding is a parser back-reference rather than a single rule. The rebase onto main also cleared the earlier conflict, so it should no longer report DIRTY. Re-requesting review.

matthewevans (approved):
Reviewed current head `605d7a337e124e0e77bfda374087493bc9abc720`. The prior CR annotation blocker is addressed on the new code/tests, the implementation reuses the existing `relative_player_scope_for_condition` / `ParseContext.relative_player_scope` path rather than adding a Grenzo-specific special case, and the parser plus runtime regressions cover the bullet-line modal form. GitHub reports the PR clean/green at this head.

Issue #4223, opened 2026-06-23
Blocks persist through zone changes — I attacked with Strangleroot Geist.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1519065815540830339/1519065815540830339

**Thread:** Blocks persist through zone changes
**Cards:** strangleroot geist, voice of victory, yawgmoth, thran physician
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
I attacked with Strangleroot Geist.

### strangleroot geist
Haste
Undying (When this creature dies, if it had no +1/+1 counters on it, return it to the battlefield under its owner's control with a +1/+1 counter on it.)
### voice of victory
Mobilize 2 (Whenever this creature attacks, create two tapped and attacking 1/1 red Warrior creature tokens. Sacrifice them at the beginning of the next end step.)
Your opponents can't cast spells during your turn.
### yawgmoth, thran physician
Protection from Humans
Pay 1 life, Sacrifice another creature: Put a -1/-1 counter on up to one target creature and draw a card.
{B}{B}, Discard a card: Proliferate. (Choose any number of permanents and/or players, then give each another counter of each kind already there.)

---
<sub>report_id: `discord:1519065815540830339:1519065815540830339:0` · discord: `1519065815540830339/1519065815540830339`</sub>

chimstead:
Full text of issue: "I attacked with Strangleroot Geist. My opponent blocked with Voice of Victory. I activated Yawgmoth, Thran Physician to sacrifice Strangleroot Geist. It returned to play with a +1/+1 counter because of undying. However, it returned tapped, attacking, and blocked by voice of victory. It should instead have returned untapped as a creature not in combat."

matthewevans:
Reporter follow-up: the original summary is too vague. Full repro detail from Discord: Strangleroot Geist attacked, was blocked by Voice of Victory, then was sacrificed to Yawgmoth, Thran Physician. It returned with undying but remained tapped, attacking, and blocked; expected it to return as a creature not in combat.

mike-theDude:
[...]

PR #4679, opened 2026-06-30
Fix combat roles persisting through battlefield exits (CR 506.4)

andriypolanski:
## Summary

- Clear live combat membership when a permanent leaves the battlefield during `apply_zone_exit_cleanup` (CR 506.4).
- Combat role is still snapshotted into the zone-change record before cleanup (CR 603.10a), so look-back triggers (`"if it was blocking"`, `"attacking creatures die"`, etc.) are unaffected.
- Fixes stale block/attack UI lines when the same `ObjectId` re-enters the battlefield (Undying, blink, exile-and-return) while combat is still in progress.

## Problem

Reported via Discord: attacking with Strangleroot Geist showed block assignments persisting after zone changes.

The engine already called `capture_combat_status` at zone exit but never removed the object from `combat.attackers`, `combat.blocker_assignments`, or `combat.blocker_to_attacker`. Because phase.rs reuses storage identity (`ObjectId`) across zone changes, a blocker that died and returned via Undying still appeared in `blocker_to_attacker`, and the client continued rendering block lines.

## Solution

Wire the existing `remove_object_from_combat` building block into the battlefield-exit path in `apply_zone_exit_cleanup`, after combat status has been captured in `move_to_zone` / `move_to_library_at_index`.

## Files changed

- `crates/engine/src/game/zones.rs`
  - Call `remove_object_from_combat` on battlefield exit.
  - Add regression test `issue_4223_combat_role_cleared_on_battlefield_exit`.

## Test plan

- [x] `cargo test -p engine --lib zones::tests::issue_4223_combat_role_cleared_on_battlefield_exit`
- [x] `cargo test -p engine --lib remove_from_combat::`
- [x] `cargo test -p engine --lib filter::tests::zone_change_record_combat_properties_match_snapshot`
- [ ] Manual: attack with a creature, block it, kill the blocker in combat, resolve Undying — returned creature must not show block lines
- [ ] Manual: attacker dies and returns mid-combat — must not remain in the attackers list

## Closes

- https://github.com/phase-rs/phase/issues/4223

### DIFF
--- a/crates/engine/src/game/effects/remove_from_combat.rs
+++ b/crates/engine/src/game/effects/remove_from_combat.rs
@@ -54,12 +54,20 @@ pub fn remove_object_from_combat(state: &mut GameState, oid: crate::types::ident
         let attackers_before = combat.attackers.len();
         combat.attackers.retain(|a| a.object_id != oid);
         attacker_removed = combat.attackers.len() != attackers_before;
-        // Remove as blocker from all attacker assignments
+        // Drop attacker-keyed forward assignments (oid was blocking nobody as a key,
+        // but was an attacker with blockers assigned to it).
+        combat.blocker_assignments.remove(&oid);
+        // Remove as blocker from all remaining attacker assignments
         for blockers in combat.blocker_assignments.values_mut() {
             blockers.retain(|b| *b != oid);
         }
-        // Remove reverse blocker lookup
+        // Remove reverse lookup when oid was a blocker
         combat.blocker_to_attacker.remove(&oid);
+        // Prune oid from every blocker's attacker list (oid was an attacker)
+        combat.blocker_to_attacker.retain(|_, attackers| {
+            attackers.retain(|id| *id != oid);
+            !attackers.is_empty()
+        });
         // Remove any pending damage assignments for this object
         combat.damage_assignments.remove(&oid);
     }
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[HIGH]** Departing attackers still leave stale block-assignment edges behind. Evidence: `crates/engine/src/game/effects/remove_from_combat.rs:50` removes the object from `combat.attackers`, removes the object only when it appears as a blocker in `blocker_assignments` values, and removes only `blocker_to_attacker[oid]`; it never removes `blocker_assignments[oid]` or prunes `oid` from other `blocker_to_attacker` values. This PR calls that helper from `apply_zone_exit_cleanup`, and the new attacker half of the regression test only checks that `combat.attackers` no longer contains the attacker (`crates/engine/src/game/zones.rs:1584`), so the stale attacker-keyed assignment map and blocker reverse lookup remain undetected. Why it matters: CR 506.4 requires the permanent to be removed from combat, not just from the attacker list; leaving the old blocker relations means a same-ObjectId return can still inherit stale combat relation data through `blocker_assignments` / `blocker_to_attacker`. Suggested fix: make `remove_object_from_combat` remove the departing object as an attacker key and prune it from every reverse attacker list (dropping empty entries), then extend the zone-exit regression to assert both maps are clear for attacker departure as well as blocker departure.

matthewevans (approved):
Maintainer review: passes the authorized review bar at head `c27d9f1b26b6811f2070d4064ddf596849532674`.

PR #4682, opened 2026-07-01
fix(parser): set-base-P/T via the "change ~'s base power [and toughness] to <value>" verb form

minion1227:
The parser recognized the intransitive "<subject>'s base power [and toughness] become[s] <value>" copula (Moon Girl, Pupu UFO, Sita Varma) but not the transitive "change <subject>'s base power [and toughness] to <value>" verb frame, so the whole "change … to" class fell to `Effect::Unimplemented`.

Generalize `try_parse_subject_base_pt_set_clause_ast` to accept both surface verbs: strip the optional "change"/"you may change" prefix and parameterize the copula (`parse_base_pt_copula`) so the possessive subject grammar and the layer-7b `SetPower(Dynamic)`/`SetToughness(Dynamic)` emission (CR 208.1 + CR 613.4b) are shared verbatim. The "change … to" value side is bare (no "equal to" lead-in): fixed "N/M", a single-axis quantity, or a paired "<X>'s power and toughness" / "the power and toughness of <X>" referent that `parse_pt_pair_referent` splits into two same-object single-axis quantities — each routed back through the existing `parse_base_pt_axis_quantity`, so every recognized referent scope (target, event-context, aggregate) composes with **no duplicated grammar and no new engine variant**.

**Unlocks:** Riptide Mangler (base power ← target creature's power), Arni Brokenbrow (base power ← 1 plus the greatest power among other creatures you control), and the dual-axis pair Shape Stealer / Eldrazi Mimic (base P/T ← that creature's P/T).

**Deferred to a follow-up** (documented in code): the inverted-genitive "change" form (Brine Hag, Exuberant Wolfbear — the copula " to " collides with a " to " inside the subject) and toughness-only axes (Sentinel, Wall of Tombstones).

**Tests:** 5 new building-block tests (single-axis target referent, "you may change" + offset-aggregate, dual-axis paired referent, fixed value, and a guard that the verb frame requires a base-P/T subject). `cargo fmt` + `clippy` + the parser combinator gate are clean; 2290 `oracle_effect` + 112 `subject`-module tests green (become-form regression intact).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/subject.rs
+++ b/crates/engine/src/parser/oracle_effect/subject.rs
@@ -616,6 +616,91 @@ fn parse_base_pt_set_value(remainder: &str) -> Option<(BasePtSetValue, &str)> {
     Some((BasePtSetValue::Dynamic(expr), ""))
 }
 
+/// CR 208.1 + CR 613.4b: Parse the copula that separates a base-P/T subject from
+/// its value. The intransitive "become[s] " form and the transitive
+/// "change … to " form (Riptide Mangler, Shape Stealer, Halfdane) share one
+/// downstream value/emission path; `is_change` selects the token so the two
+/// surface verbs are a single parameterized copula rather than duplicated arms.
+fn parse_base_pt_copula(input: &str, is_change: bool) -> OracleResult<'_, ()> {
+    if is_change {
+        value((), tag(" to ")).parse(input)
+    } else {
+        value((), (tag(" become"), opt(tag("s")), tag(" "))).parse(input)
+    }
+}
+
+/// CR 208.1 + CR 613.4b: value side of the transitive "change <subject>'s base
+/// power [and toughness] to <value>" frame. Unlike the "become[s] equal to"
+/// copula, the "change … to" frame introduces the value with a bare " to ", so
+/// the value is a fixed "N/M" (Brine Hag), a paired "<X>'s power and toughness"
+/// referent (Shape Stealer, Halfdane), or a bare single-axis quantity (Riptide
+/// Mangler). Each form routes to the exact same building block the copula form
+/// uses — no value grammar is duplicated.
+fn parse_change_base_pt_value(remainder: &str) -> Option<(BasePtSetValue, &str)> {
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed the updated head after the base-branch merge. This stays within the existing set-base-P/T parser path, parameterizes the change/to verb frame over the existing base P/T value emission, and covers fixed, single-axis dynamic, paired-referent dynamic, and guarded non-base-P/T change cases. No blocking findings from code review; CI remains the merge gate.

matthewevans (changes requested):
Blocking this after the refreshed-head CI result: the parser combinator gate flags the new `lower.strip_suffix(marker)` dispatch in `crates/engine/src/parser/oracle_effect/subject.rs`. This is parser grammar matching, so it needs to be expressed with nom combinators, or explicitly justified with the documented `allow-noncombinator` escape hatch only if it is genuinely structural rather than dispatch.\n\nThe implementation direction still looks right, but this needs to satisfy the parser mandate before it can merge.

minion1227:
Thanks — addressed. The `lower.strip_suffix(marker)` dispatch in `parse_pt_pair_referent` is now expressed with nom combinators: the possessor is captured via `take_until(marker)` + `tag(marker)` (prefix-oriented, mirroring the possessive-subject grammar in the same function), requiring the remainder to be empty so only a genuine suffix matches. The `format!` reconstruction that follows only builds the two single-axis referent strings fed back into the existing `parse_base_pt_axis_quantity` — it is sub-parser input, not dispatch.

Also rebased onto latest `main` (was BEHIND). Local `scripts/check-parser-combinators.sh` now passes (exit 0), `cargo fmt`/`clippy` clean, and the base-P/T tests (43, incl. the 5 change-frame tests) are green.

matthewevans (approved):
Reviewed the current head after the parser-combinator follow-up. The previous blocker is addressed: the paired-referent suffix is now parsed through nom (`take_until` + `tag`) and still routes through the existing base-P/T quantity builders. No blocking findings from this narrow current-head review.

PR #4683, opened 2026-07-01
Fix taps-for-mana delayed triggers parsing to an unmatchable Unknown mode (High Tide)

### LINKED ISSUE #4673: [Card Bug] High Tide does not add additional blue mana from Islands
## Card(s)

High Tide

## Build/version

Unknown

## Game mode

Unknown

## Actual behavior

After casting High Tide, Islands tapped for mana still produce only one blue mana.

## Expected behavior

After High Tide resolves, until end of turn, whenever any player taps an Island for mana, that player should add an additional `{U}`. An Island that normally produces `{U}` should effectively produce two blue mana for that tap.

## Steps to reproduce

1. Control an untapped Island.
2. Cast and resolve High Tide.
3. Tap the Island for mana before end of turn.
4. Observe the mana pool.

Actual: the mana pool increases by one blue mana.

Expected: the mana pool increases by two blue mana total: the Island's normal `{U}` plus High Tide's additional `{U}`.

## Oracle text verified from coverage data

> Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}.

## Classifier analysis

**Matched clause**: `Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}.` (supported: true)

**Reasoning**: Live coverage marks High Tide as supported and parses the clause as `CreateDelayedTrigger` with `when = whenever event this turn` and `duration = until end of turn`. The AST claims to represent the delayed trigger, but runtime mana production still only applies the Island's base mana ability. This is therefore an actionable supported-aspect defect, likely in the runtime path for delayed triggers that watch mana taps / additional mana production.

GildardoDev:
**Anchored on:** #4673

**Problem:** High Tide's "Until end of turn, whenever a player taps an Island for mana, that player adds an additional {U}" registered its delayed trigger with the inner mode `TriggerMode::Unknown` instead of `TapsForMana`. An Unknown-mode trigger has no matcher, so it never fires and the additional {U} is silently dropped. The same defect bound "that player" to the caster instead of the tapping player. The bug is class-wide: Bubbling Muck (Swamp/{B}) and the bare-instant form reproduce identically.

**Root cause:** The delayed-trigger split (`try_parse_whenever_this_turn`) strips the leading "whenever " keyword before dispatching the condition, but the taps-for-mana recognizer `parse_taps_for_mana_actor_line` mandatorily required that keyword (`preceded(alt((tag("whenever "), tag("when "))), ...)`), so on the keyword-stripped text it failed and fell to the `Unknown` fallback. The same keyword-mandatory recognizer gated `condition_matches_taps_for_mana_event`, so `relative_player_scope_for_condition` returned no scope. A second, downstream seam then surfaced: `parse_subject_application` had no arm for a resolved `TriggeringPlayer`/`DefendingPlayer` scope and defaulted "that player" to `ParentTargetController`.

**Fix:**
- `parser/oracle_trigger.rs` — made the leading keyword optional (`preceded(opt(alt(...)), ...)`) in `parse_taps_for_mana_actor_line`, so the keyword-stripped delayed-trigger condition is recognized as `TapsForMana` and its "that player" scope resolves to `TriggeringPlayer`. Strictly more permissive; keyworded callers (printed-permanent triggers) still match.
- `parser/oracle_effect/subject.rs` — added an arm mapping a resolved `TriggeringPlayer`/`DefendingPlayer` relative-player scope to the context filter, so "that player" honors the triggering player instead of defaulting to the caster. The arm only fires when the subject is absent (the delayed-trigger split path), so regular triggers are unaffected.

Both changes route through the existing `relative_player_scope_for_condition` single authority. No new engine enum variants. Class-general across permanent type and mana color.

**Tests:**
- Parser `high_tide_delayed_trigger_taps_for_mana_mode_and_filter`: real Oracle text lowers to `CreateDelayedTrigger{ WheneverEvent{ TapsForMana, valid_card Island } }` with the additional-{U} rider and `UntilEndOfTurn`; asserts the mode is not `Unknown`. Reverting the `opt()` reproduces `Unknown`.
- Parser `high_tide_delayed_trigger_that_player_binds_triggering_player`: the mana recipient binds to `TriggeringPlayer`. Reverting the subject.rs arm reproduces `ParentTargetController`.
- Parser `bubbling_muck_delayed_trigger_taps_for_mana_class_general`: Swamp/{B} sorcery sibling, same assertions.
- Runtime `high_tide_runtime_bonus_mana_routes_to_triggering_player_and_expires_at_eot` (2-player GameScenario): after High Tide resolves, P0 tapping an Island gains {U}{U} while P1 is unaffected; P1 tapping their Island routes the bonus to P1 (triggering player, not caster); after end-of-turn cleanup the trigger is purged and a later tap yields only the base {U}.

The Sea Devils III's "adds an additional mana of any type that land produced" rider is a separate, pre-existing unimplemented gap and is intentionally not addressed here.

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, ./scripts/check-parser-combinators.sh, and ./scripts/check-skill-doc.sh all clean. Regression green: taps_for_mana 14, delayed_trigger 58, broad trigger 2014, the printed-permanent Mana Flare taps-for-mana test, and the aura taps-for-mana class (Wild Growth/Utopia Sprawl/Fertile Ground) all pass; full engine lib suite 14377 passed / 0 failed.

Closes #4673.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/parser/oracle_effect/subject.rs
+++ b/crates/engine/src/parser/oracle_effect/subject.rs
@@ -1929,6 +1929,20 @@ pub(super) fn parse_subject_application(
                 Some(ControllerRef::ParentTargetController)
             ) {
                 TargetFilter::ParentTargetController
+            } else if matches!(
+                ctx.relative_player_scope,
+                Some(ControllerRef::TriggeringPlayer | ControllerRef::DefendingPlayer)
+            ) {
+                // CR 608.2c + CR 106.12a: An explicit triggering/defending player
+                // scope established by `relative_player_scope_for_condition`
+                // (e.g. the instant/sorcery taps-for-mana delayed trigger split:
+                // "whenever a player taps <type> for mana, that player adds …" —
+                // High Tide, Bubbling Muck) makes "that player" the triggering
+                // player, NOT the parent target's controller. Without this arm the
+                // scope is resolved but silently discarded, defaulting to
+                // `ParentTargetController` below. `ctx_filter` is the matching
+                // event-context ref for the parsed subject phrase.
+                ctx_filter
             } else if ctx.subject.is_some() {
                 ctx_filter
             } else {
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -7667,11 +7667,21 @@ fn split_taps_for_mana_for_clause(text: &str) -> Option<(String, Option<Vec<Mana
 /// CR 603.2 + CR 605.1a: Shared nom dispatch for "Whenever [an opponent / a player]
 /// taps … for mana" trigger conditions. Used by trigger-line dispatch and by
 /// `condition_matches_taps_for_mana_event` for `"that player"` scope binding.
+///
+/// The leading "whenever "/"when " keyword is OPTIONAL. Printed-permanent
+/// dispatch (`try_parse_player_trigger`) passes the full keyworded condition,
+/// but the instant/sorcery delayed-trigger split path
+/// (`try_parse_whenever_this_turn` in `oracle_effect`) strips the "whenever "
+/// keyword before handing the condition to `parse_trigger_condition`, so the
+/// keyword is already gone by the time this recognizer runs on High Tide /
+/// Bubbling Muck. `opt()` accepts both forms; because both callers require the
+/// recognizer to consume the ENTIRE candidate line (empty remainder), the
+/// looser match cannot false-positive on mid-sentence text.
 fn parse_taps_for_mana_actor_line(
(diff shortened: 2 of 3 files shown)


matthewevans (approved):
Reviewed the updated head after the base-branch merge. The change keeps taps-for-mana delayed triggers in the existing trigger parser/matcher path, binds "that player" to TriggeringPlayer for this event context, and includes parser plus runtime coverage for High Tide/Bubbling Muck behavior. No blocking findings from code review; CI remains the merge gate.

PR #4686, opened 2026-07-01
fix(parser): cast prohibition "during that player's next turn" sets a next-turn expiry

minion1227:
The clause "Each opponent can't cast instant or sorcery spells **during that player's next turn**" (Sphinx's Decree, Azor, the Lawbringer) parsed the prohibition itself (`ProhibitActivity{CastSpells}`) but **dropped** the trailing "during that player's next turn" duration into an `Unimplemented` sub_ability, leaving the restriction to expire at the **current** turn's cleanup — so opponents were wrongly free to cast instants/sorceries on their next turn (wrong game result).

Mirror the attack-prohibition branch (`try_parse_that_player_cant_attack_prohibition`, which already recognizes both "during their next turn" and "during that player's next turn"): after the spell filter, consume the trailing next-turn clause and set `duration = UntilEndOfNextTurnOf { player: Controller }`.

**No new variant, no runtime change** — `add_restriction::fill_runtime_fields` already lowers that duration to `RestrictionExpiry::UntilEndOfNextTurnOf`, and for the `OpponentsOfSourceController` scope it anchors on the controller, whose next turn follows each opponent's single intervening next turn in a round-robin (CR 514.2 + CR 500.7 + CR 109.5). The `EndOfTurn` expiry set at parse time is the placeholder `fill_runtime_fields` overrides from `ability.duration`, exactly as the attack branch does.

**Tests:** 3 new — the "that player's next turn" form (asserts the captured `UntilEndOfNextTurnOf` duration and that no `Unimplemented` sub_ability is produced), the "during their next turn" determiner variant, and a "this turn" regression guard that must keep `EndOfTurn`. `cargo fmt` + `clippy` + the parser combinator gate are clean; 2301 `oracle_effect` + 7 `add_restriction` tests green.

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -523,12 +523,21 @@ fn is_blocked_by_cant_cast_spells(
         let GameRestriction::ProhibitActivity {
             source,
             affected_players,
+            expiry,
             activity: ProhibitedActivity::CastSpells { spell_filter },
-            ..
         } = restriction
         else {
             return false;
         };
+        // CR 514.2 + CR 500.7: a still-pre-armed `UntilEndOfNextTurnOf` cast ban
+        // ("Each opponent can't cast … during that player's next turn" — Sphinx's
+        // Decree / Azor, fanned out per opponent) is not yet in force; it takes
+        // effect only once the restricted player's untap step converts it to
+        // `EndOfTurn` (turns.rs). Mirror the activate-abilities gate so the ban
+        // does not leak onto the creating turn.
+        if matches!(expiry, RestrictionExpiry::UntilEndOfNextTurnOf { .. }) {
+            return false;
+        }
         let source_controller = state
             .objects
             .get(source)
--- a/crates/engine/src/game/casting_tests.rs
+++ b/crates/engine/src/game/casting_tests.rs
@@ -37675,3 +37675,79 @@ mod plot_from_library {
         );
     }
 }
+
+/// CR 500.7 + CR 514.2 + CR 109.5: Sphinx's Decree / Azor — "Each opponent can't
+/// cast … during that player's next turn." The ban must be dormant on the
+/// creating turn (opponents may still respond that turn) and take force on each
+/// opponent's OWN next turn. Resolving the restriction fans it out per opponent;
+/// the pre-armed marker only bites once that opponent's untap step arms it.
+#[test]
+fn cant_cast_next_turn_arms_on_each_opponents_own_turn() {
+    use crate::types::ability::{
+        Duration, Effect, GameRestriction, PlayerScope, ProhibitedActivity, ResolvedAbility,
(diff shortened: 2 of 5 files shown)


matthewevans (changes requested):
**[HIGH]** The next-turn cast ban is anchored on the source controller, so it is dormant during the affected opponents' next turns. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:2761`-`:2768` stamps `Duration::UntilEndOfNextTurnOf { player: PlayerScope::Controller }` for "Each opponent can't cast ... during that player's next turn"; `crates/engine/src/game/effects/add_restriction.rs:105`-`:133` only re-anchors that duration to a restricted player when `affected_players` has already been lowered to `SpecificPlayer`, but this parser path leaves it as `OpponentsOfSourceController`; and `crates/engine/src/game/casting.rs:583` ignores `RestrictionExpiry::UntilEndOfNextTurnOf` while it is still pre-armed. Why it matters: Sphinx's Decree/Azor will not actually stop opponents from casting on their next turns. Suggested fix: lower this shape to per-affected-player/specific-player restrictions or another representation that arms on each opponent's own next turn, and add a runtime test that resolves the card, advances to an opponent's next turn, and proves the prohibited spell is rejected.

minion1227:
Good catch — fixed, and you were right that the controller-anchored marker never took force. Two engine changes make it rules-correct (verified with a runtime test):

1. **Per-opponent fan-out** (`add_restriction::resolve`): an `OpponentsOfSourceController` + `UntilEndOfNextTurnOf` prohibition is now split into one `SpecificPlayer(opponent)` restriction per opponent, so `fill_runtime_fields` anchors each on that opponent's **own** next turn (armed at their untap step). Every other shape — including Kang's `AllPlayers` self-anchored power-up ban — passes through unchanged.

2. **Pre-armed cast gate** (`is_blocked_by_cant_cast_spells`): mirrors the activate-abilities gate you pointed at, skipping a still-pre-armed `UntilEndOfNextTurnOf` so the ban stays dormant on the creating turn (opponents may still respond that turn) and bites only once the opponent's untap step arms it.

New runtime test `cant_cast_next_turn_arms_on_each_opponents_own_turn`: resolves the restriction, asserts the per-opponent fan-out, advances to the opponent's next turn (runs their untap step), and proves the prohibited spell is rejected there while remaining castable on the creating turn. Also rebased onto latest `main`.

`add_restriction` (7) + `cant_cast` (43) + `game::casting` (697) tests green; fmt + clippy + parser combinator gate clean.

matthewevans (approved):
Reviewed current head 579c797b31ada65aa4eca4fde8c3fe391ea13033. The follow-up fixes the prior next-turn anchoring blocker: parser preserves the duration, restrictions fan out per affected opponent, and the runtime test exercises the dormant-to-armed transition through execute_untap.

PR #4688, opened 2026-07-01
feat(engine): S01 standard-set completion — 11 cards / 7 mechanic groups

lgray:

Completes the S01 standard-set card cluster — 11 cards across 7 mechanic groups, each landed through an adversarial review + verification gate.

## Groups
- **Reflexive object-property conditions** (card-type / exact-power / compound or-if) + the current-phase condition — Dose of Dawnglow.
- **Excess-damage "this way" condition** (Torch the Witness, Orbital Plunge) + convergence of the `DamageDealtThisTurn` excess flag onto a typed `DamageChannel`.
- **Full Bore** — target-scoped "cast for its warp cost" condition (`CastVariantPaid { subject: ObjectScope::Target }`, CR 608.2c / CR 702.185a).
- **ConditionInstead not-swap trailing tail** (CR 608.2c) — Throw from the Saddle, Take the Fall, That's Rough Buddy, Evil's Thrall.
- **Kylox's Voltstrider** (MKM) — collect-evidence activated ability with the cost charged through the single-authority interactive-cost resolver (fixes a CR 701.59 free-animate), plus the cast-from-exile library-bottom rider.
- **Quistis Trepe + Tinybones the Pickpocket** — cast-from-graveyard with any-type mana (CR 609.4b), multiplayer-correct target (`Owned { TriggeringPlayer }`).
- **Summoner's Grimoire** (MKM) — conditional enters-tapped-and-attacking on a granted put-from-hand, a zero-new-variant leaf parameterization of the ChangeZone enters-modifier cluster (`enters_modified_if`, CR 614.12).

## Verification (rebased on `upstream/main` @ `2222dbeed`)
- `cargo clippy --workspace --all-targets --features engine/proptest -D warnings`: **0 warnings**
- `cargo test -p engine --features proptest`: **16531 passed, 0 failed** (15 ignored). All 7 load-bearing S01 tests pass.
- **card-data coverage-regression** (built HEAD and a `2222dbeed` baseline against the *same* MTGJSON + *identical* committed `known-tokens.toml`, so the diff is purely parser-source):
  - **REGRESSED (engine) = 0** — fatal CI gate passes.
  - ORACLE CHANGED = 0 (clean parser-only diff).
  - **GAINED = 19** — 10 named S01 cards flipped `false → true` plus 9 class-level bonus unlocks; Tinybones + Evil's Thrall were already supported and were improved without a coverage flip.
  - 1 non-fatal coverage-honesty flip (Masterful Ninja) — HEAD honestly flags a nonsensical clause the baseline silently swallowed (swallowed-clause diagnostic 969 → 947).


matthewevans (changes requested):
**[HIGH]** Excess-damage fight conditions read the wrong fighter. Evidence: `crates/engine/src/game/effects/mod.rs:5037` selects the first object target of the fight when computing `PreviousEffectAmount` excess, but the parser accepts text like "the creature the opponent controls is dealt excess damage this way" where the recipient is the second fight target. Why it matters: dual-target fight cards such as The Last Agni Kai can add mana based on excess damage to the controller's creature instead of the opponent's creature. Suggested fix: carry the parsed damage-recipient subject into the lowered condition, or make the fight excess resolver select the actual recipient named by the condition, with a dual-target runtime test.

**[HIGH]** Full-cost "you may cast ... from graveyard" effects lower to a delayed permission instead of a during-resolution cast. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:16884` turns Quistis/Tinybones-style text into `Effect::CastFromZone`, and `crates/engine/src/game/effects/cast_from_zone.rs:297`-`:308` only casts during resolution for free-cast/graveyard-free-cast paths; otherwise it grants lingering permissions. Why it matters: these triggers are one-shot casts during resolution, but the PR changes timing and can allow/deny spells based on a later priority window instead. Suggested fix: implement a paid during-resolution cast path that uses the real mana cost plus `mana_spend_permission`, or keep this text unsupported until that path exists.

**[MED]** Graveyard replacement riders can leak from an unconsumed cast permission. Evidence: `crates/engine/src/game/casting_costs.rs:5832`-`:5841` scans any `ExileWithAltCost { graveyard_replacement: Some(..) }` on the object instead of the selected permission that authorized the current cast, while the adjacent enters-with-counter path uses selected-permission authority at `:5851`-`:5856`. Why it matters: a stale or foreign permission can redirect the resolving spell to exile/library/hand even when the current cast did not use that rider. Suggested fix: add a selected-permission helper analogous to `selected_exile_alt_cost_permission_enters_with_counter` and use it when installing `apply_spell_graveyard_replacement_rider`.

**[MED]** Legacy serialized excess-damage quantities lose their excess-only gate. Evidence: `crates/engine/src/types/ability.rs:4649` replaces the former `excess_only` bool with a defaulted `channel: DamageChannel`, but there is no alias/compat deserializer mapping old `excess_only: true` payloads to `DamageChannel::Excess`. Why it matters: old card data/saves/scenarios reload as total-damage checks and can fire on any matching damage. Suggested fix: add serde compatibility for the old bool field and a legacy JSON test for `excess_only: true`.

**[MED]** Legacy serialized `ExileWithAltCost` permissions lose their exile-on-resolution rider. Evidence: `crates/engine/src/types/ability.rs:2118` replaces `exile_instead_of_graveyard_on_resolve` with defaulted `graveyard_replacement`, but there is no alias/compat mapping from the old bool to `Some(SpellStackToGraveyardReplacement::Exile)`. Why it matters: in-flight saved/reconnected permissions reload without the required graveyard replacement and resolve to the graveyard. Suggested fix: add serde compatibility for the old bool field and cover it with a legacy `CastingPermission` round-trip test.

matthewevans (changes requested):
Thanks for the quick follow-up. I re-reviewed the current head (`e21efa8890aa0c36ed57d586d84adf0b69c889c8`) and the main blockers from the previous review are still present.

**[HIGH]** The fight excess-damage condition still reads the wrong fighter. `previous_effect_excess_amount_from_events` still selects the first object target on the Fight ability (`crates/engine/src/game/effects/mod.rs:5037-5040`) and then sums excess damage dealt to that object (`:5041-5056`). For templates like The Last Agni Kai, the first target is the source creature, but the condition is about excess damage dealt to the opponent's creature, so the condition can pass/fail on the wrong damage event.

**[HIGH]** Full-cost resolution-time graveyard casts still become lingering permissions. The parser now forwards `ManaSpendPermission::AnyTypeOrColor`, but `handle_cast_from_zone` only performs an immediate graveyard cast when `without_paying_mana_cost` is true (`crates/engine/src/game/effects/cast_from_zone.rs:288-297`). Quistis/Tinybones-style text says to cast the targeted graveyard card during resolution while paying its normal cost, so the non-free path still falls through to `grant_lingering_permissions` at `:308` instead of creating the required during-resolution paid-cast choice.

**[MED]** The graveyard replacement rider is still read from any permission on the object, not the permission consumed by this cast. `finalize_cast_with_phyrexian_choices` still scans `obj.casting_permissions.iter().find_map(...)` for the first `ExileWithAltCost { graveyard_replacement: Some(...) }` (`crates/engine/src/game/casting_costs.rs:5832-5840`). That can apply another grant's replacement rider to this cast; it needs the same selected-permission authority used by `selected_exile_alt_cost_permission_enters_with_counter`.

**[MED]** The serialized-state migration gaps are still present. `DamageDealtThisTurn` now defaults `channel` to total (`crates/engine/src/types/ability.rs:4644-4650`), and `ExileWithAltCost` now defaults `graveyard_replacement` to `None` (`:2118-2119`), but there is no compatibility path for legacy `excess_only: true` or `exile_instead_of_graveyard_on_resolve: true`. Existing saved/generated JSON using those fields will deserialize as total damage / no replacement rider.

matthewevans (approved):
Maintainer review: passes the authorized review bar at head `10be31afa74eb17a1345adbf420c348ec7c676db`.

PR #4692, opened 2026-07-01
fix(parser): set-base-P/T toughness-only axis ("change ~'s base toughness to <value>")

minion1227:
The base-P/T set-clause parser handled the power-only and both-axes forms ("~'s base power [and toughness] …") but not the **toughness-only** axis, so "change ~'s base toughness to <value>" fell to `Effect::Unimplemented`.

Two symmetric adjustments in `try_parse_subject_base_pt_set_clause_ast`:
- `parse_base_pt_axes` now accepts a leading "base toughness" (`set_power:false, set_toughness:true`), mirroring the existing power-only axis. Tried first; it cannot shadow the both-axes form, which always opens with "base power".
- the possessive subject capture anchors on `"'s base "` instead of `"'s base power"` so the toughness-only characteristic word reaches `parse_base_pt_axes` for classification.

The per-axis emission (`SetToughnessDynamic` only, base power untouched) and every value form (fixed / single-axis dynamic / paired referent) are reused verbatim — **no new variant**. Unlocks **Wall of Tombstones** ("change this creature's base toughness to 1 plus the number of creature cards in your graveyard") and completes the axis symmetry for the class.

Sentinel additionally needs the "blocking or blocked by ~" combat-relation target filter in its value referent — a separate gap, left for its own change.

**Tests:** toughness-only change-form test (asserts `SetToughnessDynamic` and that base power is not touched). `cargo fmt` + `clippy` + the parser combinator gate are clean; 2307 `oracle_effect` + 118 `subject`-module tests green (power/both-axes regression intact).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/subject.rs
+++ b/crates/engine/src/parser/oracle_effect/subject.rs
@@ -560,6 +560,19 @@ struct BasePtSetAxes {
 /// "base " on the second noun and the optional "and toughness" conjunct are each
 /// a single combinator.
 fn parse_base_pt_axes(input: &str) -> OracleResult<'_, BasePtSetAxes> {
+    // CR 208.1: "base toughness" alone (toughness-only, symmetric with the
+    // power-only axis) — Sentinel / Wall of Tombstones set base toughness without
+    // touching base power. Tried first; it cannot shadow the both-axes form,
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Maintainer review: passes the authorized review bar at head `7d8f10fc760363c86b9535d480f22caa559ac129`.

PR #4698, opened 2026-07-01
Fix Twilight Prophet caster gaining 0 life from the revealed card's mana value

### LINKED ISSUE #1375: Twilight Prophet — Not triggering at my upkeep, I do have ascend
Reported in Discord: https://discord.com/channels/1485498006781427802/1509672840633389226/1509672840633389226

**Thread:** Twilight Prophet
**Cards:** _none detected_
**Parser status:** no_card

## Summary
Not triggering at my upkeep, I do have ascend

---
<sub>report_id: `discord:1509672840633389226:1509672840633389226:0` · discord: `1509672840633389226/1509672840633389226`</sub>

GildardoDev:
**Anchored on:** #1375

**Problem:** Twilight Prophet's Ascend upkeep — "reveal the top card of your library and put it into your hand. Each opponent loses X life and you gain X life, where X is that card's mana value" — gives the caster the wrong life total: **you gain 0 instead of the revealed card's mana value.** (The "each opponent loses X" half happens to work, but only by a target-slot coincidence; the "you gain X" half is observably broken.)

Note on the report: #1375 was filed as "not triggering at my upkeep, I do have ascend." The trigger firing itself is actually fine on current `main` — Ascend grants the city's blessing at ten or more permanents, the "if you have the city's blessing" intervening-if evaluates, and the ability resolves and reveals the card. The observable defect the player hit is this drain: the reveal happens but the caster's life doesn't change (gain 0), which reads as "nothing happened at upkeep." This PR fixes that root cause.

**Root cause:** The "where X is that card's mana value" binding fell through to `parse_cda_quantity`, which hard-maps "that card's mana value" to `QuantityRef::ObjectManaValue { scope: ObjectScope::Target }`. The upkeep trigger has no object target — the revealed card is an anaphoric reference, not a target (CR 115.10a). The chained `LoseLife` coincidentally finds the revealed card in a propagated target slot, but the deeper chained `GainLife` has an empty target slot, so `ObjectScope::Target` reads nothing there and X resolves to 0 for the caster.

**Fix:** A nom guard in `parse_where_x_quantity_expression` routes only the literal "that card's mana value" (and its "converted mana cost" synonym) through `parse_event_context_quantity`, so the anaphoric referent binds `ObjectScope::Demonstrative` — resolved against the revealed card via the LKI-snapshotted `effect_context_object` (CR 608.2c, CR 202.3). "that spell's mana value" (EventSource) and "that creature's/permanent's mana value" (Target, which is correct for targeted cards like Feeding Grounds) are deliberately excluded, and power/toughness is untouched. Class-general across the "that card's mana value" cards; the ambiguous demonstrative-sense creature references (e.g. Aether Mutation) are a separate antecedent-disambiguation problem and are intentionally not addressed here.

**Tests:**
- Runtime (live-parsed, so the parser guard is load-bearing): P0 with the city's blessing reveals a mana-value-5 card at upkeep; asserts the caster **gains 5** and each opponent **loses 5**. Reverting the guard makes the caster gain 0 → the gain assertion fails.
- Runtime no-regression: a live-parsed "put X counters on target creature, where X is that creature's mana value" card still resolves X against its target.
- Parser: "that card's mana value" → Demonstrative (revert → Target); "that spell's mana value" → EventSource preserved; "that creature's mana value" → Target preserved; full-card Twilight Prophet parse asserts both drains bind Demonstrative.

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, ./scripts/check-parser-combinators.sh, and ./scripts/check-skill-doc.sh all clean. Both guard-toggle revert-fails proven. Regression green: quantity 675, mana_value 193, object_mana 10, plus the Dark Confidant and Yuriko integration tests. No new engine variant; zero net test-fixture change.

Closes #1375.

Model: claude-opus-4-8
Tier: Frontier


matthewevans (changes requested):
**[MED]** The normalized guard returns the untrimmed expression. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:6935` trims trailing periods into `expression`, but `crates/engine/src/parser/oracle_effect/lower.rs:7043` still calls `parse_event_context_quantity(where_x_expression)`. Why it matters: if the caller passes `that card's mana value.`, the new guard can fire and then return `None`, so the bug survives for punctuation-bearing inputs despite the normalization. Suggested fix: pass the already-trimmed `expression` to `parse_event_context_quantity` and cover the trailing-period case.

Reviewed current head `9d7392ed909ef2e577b628f3cac1e8663f6e8fcc`.

matthewevans (approved):
Reviewed current head c78e409b3ecc777419a25c70cc7da7c382082336. The prior blocker is fixed: the where-X event-context quantity path now passes the trimmed expression into parsing, and coverage includes both Twilight Prophet and Living Armor regressions.

Issue #4388, opened 2026-06-26
Itlimoc, Cradle of the Sun and Gaea's Cradle. — Having trouble activating these during opponents turns.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1519811300996284416/1519811300996284416

**Thread:** Itlimoc, Cradle of the Sun and Gaea's Cradle.
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Having trouble activating these during opponents turns.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1519811300996284416:1519811300996284416:0` · discord: `1519811300996284416/1519811300996284416`</sub>

mike-theDude:
Pipeline findings:

- Card data/parser side looks good: both Gaea's Cradle and Itlimoc, Cradle of the Sun are fully parsed as mana abilities.
- Added an engine regression in `crates/engine/tests/integration/issue_3647_gaeas_cradle.rs` covering opponent-turn priority. It verifies that:
  - Gaea's Cradle is exposed through `legal_actions_full` as the single-option `TapLandForMana` shortcut.
  - Itlimoc is exposed through `legal_actions_full` as the multi-option `ActivateAbility` action for the variable green ability.
  - Both actions successfully resolve while player 0 has priority during player 1's turn, producing the expected green mana.
- Verified locally with:
  - `cargo fmt --all`
  - `cargo fmt --all -- --check`
  - `cargo test -p engine --test integration issue_3647_gaeas_cradle -- --nocapture`

Conclusion: the engine/backend action surface already allows these activations when the player actually receives priority. If the player-visible problem is that the window disappears before they can click the land, the remaining suspect is current client auto-pass policy for quiet mana-only priority windows. Those windows are intentionally skipped unless full control or an appropriate phase stop is active; changing that globally would make all harmless mana sources halt opponent-turn priority.

matthewevans:
[...]

Issue #544, opened 2026-05-20
Krark-Clan Ironworks — The game auto-passes in this board state when it shouldn't, because I can activate Krark-Clan Ir…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1506424625502879795/1506424625502879795

**Thread:** Krark-Clan Ironworks
**Cards:** krark-clan ironworks
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
The game auto-passes in this board state when it shouldn't, because I can activate Krark-Clan Ironworks for mana

### krark-clan ironworks
Sacrifice an artifact: Add {C}{C}.

---
<sub>report_id: `discord:1506424625502879795:1506424625502879795:0` · discord: `1506424625502879795/1506424625502879795`</sub>

PR #4702, opened 2026-07-01
fix(engine): hold priority for mana on opponent's turn (CR 117.1d, #4388)

andriypolanski:
## Summary

- **Root cause:** `auto_pass_recommended` treated standalone mana abilities (Gaea's Cradle, Itlimoc, basic lands) as non-meaningful priority actions. On an opponent's turn, the client auto-passed through priority windows before the player could tap those lands, even though the engine already exposed the actions in `legal_actions_by_object`.
- **Fix:** When the priority player is not the active player and `activatable_object_mana_actions` is non-empty, do not recommend auto-pass (CR 117.1d).
- **Scope:** Engine-only; no parser or frontend changes. Sacrifice-for-mana (KCI, #544) behavior unchanged. Auto-pass on your own turn with only mana available is unchanged.

## Changed files

| File | Change |
|------|--------|
| `crates/engine/src/ai_support/mod.rs` | Hold priority when grouped mana is available on opponent's turn |
| `crates/engine/tests/integration/issue_4388_mana_on_opponents_turn.rs` | Integration regression (legal actions + auto-pass + activation) |

## Test plan

- [x] `cargo test -p engine --test integration issue_4388`
- [x] `cargo test -p engine auto_pass_holds_priority_for_grouped`
- [x] `cargo test -p engine auto_pass_does_not_skip_non_mana` (existing behavior preserved)
- [ ] Manual: on opponent's turn with priority, tap Gaea's Cradle / Itlimoc without Full Control enabled
- [ ] Manual: on your own main phase with only mana available, auto-pass still advances as before

## Rules

- CR 117.1d — priority player may activate their mana abilities during another player's turn
- CR 605.3a — mana abilities do not use the stack

### DIFF
--- a/crates/engine/src/ai_support/mod.rs
+++ b/crates/engine/src/ai_support/mod.rs
@@ -789,15 +789,6 @@ pub fn auto_pass_recommended(state: &GameState, actions: &[GameAction]) -> bool 
         _ => return false,
     };
 
-    // CR 117.1d (issue #4388): On an opponent's turn the priority player may
-    // activate their own mana abilities (Gaea's Cradle, Itlimoc, basic lands).
-    // Those actions live only in `legal_actions_by_object`, not the flat list
-    // consumed here — without this guard auto-pass fires through the window
-    // before the frontend can offer a tap.
-    if state.active_player != player && !activatable_object_mana_actions(state).is_empty() {
-        return false;
-    }
-
     if auto_passes_initial_priority_by_default(state) {
         return true;
     }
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[HIGH]** Opponent-turn ordinary mana now disables frontend auto-pass whenever any mana source is available. Evidence: `crates/engine/src/ai_support/mod.rs:797` and `client/src/game/controllers/gameLoopController.ts:110`. Why it matters: the default game loop will stop on every opponent priority window for any untapped land or mana ability even when the only available action is standalone mana, regressing normal play and performance. Suggested fix: keep grouped mana actions exposed, but only block auto-pass for meaningful mana activations, such as the existing sacrifice-for-mana path, or when the user explicitly sets full control/phase stops.

Reviewed current head `68272d5250b3285674a3968944677cc53d6f2716`.

matthewevans:
Thanks for the contribution. This PR is currently test-only against current main; it does not include a production behavior change. We're closing standalone test-only PRs so the contribution queue stays focused on fixes that change behavior. If you want to continue on #4388, please reopen with the behavioral fix and focused regression coverage together.

PR #4704, opened 2026-07-01
fix(parser): lower "if you put a [type] onto the battlefield this way" rider (Spelunking)

dean-alley:
## Summary

Fixes **Spelunking** (LCI, Modern 4-of). Its ETB rider — *"If you put a Cave onto the battlefield this way, you gain 4 life."* — was swallowed (`Condition_If`), which dropped the `GainLife` gate and marked the card `supported: false`. The rest of the card already worked on HEAD (the `Lands you control enter untapped` replacement half; the `draw a card, then you may put a land card…` stem).

**Oracle text:** `When this enchantment enters, draw a card, then you may put a land card from your hand onto the battlefield. If you put a Cave onto the battlefield this way, you gain 4 life.\nLands you control enter untapped.`

## Root cause

The active-voice put-onto-battlefield gate combinator `parse_you_put_onto_battlefield_this_way_clause` already existed, but was only wired into `strip_if_you_do_conditional`'s `when ` branch — never the leading `if ` path that Spelunking's rider takes. So `you put a Cave onto the battlefield this way` reached only the passive-voice recognizer (`a Cave is/was put …`) and failed to lower.

## Fix

Compose the existing active-voice combinator into `parse_outcome_this_way_condition` — the shared recognizer reached by both the leading-`if` (`strip_leading_general_conditional` → `try_nom_condition_as_ability_condition`) and trailing-suffix paths — alongside the passive-voice `parse_zone_changed_this_way_clause` it already delegated to. Both grammatical voices now lower to the identical `AbilityCondition::ZoneChangedThisWay`.

Parser-only, additive (a fallback `alt` branch), no change to the sacrificed/exiled/discarded or passive-voice paths.

## Build for the class

This unlocks the whole **"if you put a [type] onto the battlefield this way, [bonus]"** fetch/ramp payoff rider class — the verb arm is the reusable unit and the type token (`Cave`) is the only card-specific part.

## CR

- **CR 608.2c** — later text on a card may reference an earlier instruction in the same effect ("this way").
- **CR 400.7j** — if an effect causes an object to move to a public zone, other parts of that effect can find that object (how the rider finds the Cave the preceding put-land instruction moved to the battlefield).

## Tests

- `leading_you_put_a_cave_onto_battlefield_spelunking` — leading-`if` active-voice gate → `ZoneChangedThisWay { Cave }`.
- `outcome_you_put_a_creature_onto_battlefield_active_voice` — suffix-form active-voice gate stays in lockstep.
- `spelunking_etb_put_cave_gains_life_conditionally` — end-to-end `parse_oracle_text` of the full card: `GainLife` sub-ability carries `ZoneChangedThisWay { Cave }`, no `Unimplemented` node, replacement half still parses.

`cargo fmt --all` clean; `cargo clippy -p engine --lib` clean; the full `parser::oracle_effect::conditions` unit suite (95 tests) passes.


matthewevans (changes requested):
**[LOW]** Function-local imports violate the repo's no-inline-import rule. Evidence: `crates/engine/src/parser/oracle_effect/conditions.rs:5036`. Why it matters: the repository explicitly requires all imports at file scope, and this new parser path adds a local `use` inside rule-touching code. Suggested fix: move `parse_you_put_onto_battlefield_this_way_clause` and `parse_zone_changed_this_way_clause` imports to the module import section or call them through the module path.

Reviewed current head `a440c57f36d693967a4e0cb08aaeaf54f21891c0`.

matthewevans:
Holding current head for parser evidence before re-review/approval. This PR changes parser code, but the PR packet still has no current coverage parse-diff sticky to check the card-level blast radius against the Spelunking scope.

matthewevans:

matthewevans:
Rechecked current head a80ac17 after CI went green. The previous changes-requested blocker is still present, just at the rebased location: `crates/engine/src/parser/oracle_effect/conditions.rs` still has a function-local `use crate::parser::oracle_nom::condition as nom_cond;` inside the parser branch. The repo rule requires imports at file scope, so I am leaving the requested change active.

matthewevans (changes requested):
Thanks for the update. One remaining repo-policy issue blocks this for me:

**[LOW]** Function-local imports still violate the repo's no-inline-import rule. Evidence: `crates/engine/src/parser/oracle_effect/conditions.rs:5935` and `crates/engine/src/parser/oracle_tests.rs:48`. Why it matters: the repository requires all imports at file scope, and this PR adds new local `use` statements in parser/test code. Suggested fix: move the imported items to the module import section or call them through their module paths.

matthewevans (approved):
Maintainer review: current head resolves the prior inline-import blocker, keeps the active/passive `this way` parser at the existing condition seam, has current no-change parse-diff evidence, and CI is green. Approving for merge queue.

PR #4713, opened 2026-07-01
fix(parser): UNSUPPORTED full-set one-off: Mondassian Colony Ship

ntindle:
## Summary

**Root cause:** UNSUPPORTED full-set one-off: Mondassian Colony Ship

## Cards corrected
- Mondassian Colony Ship

## Fix
Fixed cluster 91 — Mondassian Colony Ship (WHO). The card's ChaosEnsues→TurnFaceDown ability was already fully implemented on current upstream/main; the spec snapshot was stale. The real defect was a SILENT MISPARSE on the Attacks trigger: "it gets +1/+1 for each other creature its controller controls that shares a creature type with it" parsed to a flat Pump{Fixed(1)/Fixed(1)} and swallowed the entire for-each multiplier (SwallowedClause/DynamicQty warning). Two surgical fixes, no new engine enum variants (add-engine-variant gate N/A — all primitives already exist):

FIX 1 (parser, oracle_target.rs::parse_controller_suffix): added a nom alt() arm recognizing the anaphoric controller suffix "its controller controls" / "their controller controls". Binds to ControllerRef::TriggeringPlayer in a trigger-subject context and ControllerRef::ParentTargetController otherwise — a verbatim mirror of resolve_pronoun_target/resolve_it_pronoun's ctx.subject discriminator, so "its controller" resolves to the same anaphor as the sibling "shares … with it" clause. After the fix the trigger parses to Pump{power/toughness = Quantity(ObjectCount{Typed{Creature, controller: TriggeringPlayer, [Another, SharesQuality{CreatureType, reference: TriggeringSource}]}})}, verified via oracle-gen (parse_warnings: null) and in the full 35k-card export (no Unimplemented, no Unknown, both triggers present).

FIX 2 (runtime, game/effects/pump.rs — CLASS-LEVEL latent bug also affecting Shared Animosity): the eager Pump/PumpAll resolver resolved the dynamic P/T count with recipient: None, so FilterProp::Another excluded the ability SOURCE (the Plane in the command zone) instead of the creature being pumped — overcounting by one (attacker wrongly counted). Parameterized pt_modifications with an Option<ObjectId> recipient and added pt_uses_recipient (gates on the existing quantity_expr_uses_recipient); when the amount references the pumped object the count is now resolved per-recipient via resolve_quantity_with_targets_and_recipient, otherwise the single shared resolution is reused. Applied identically to resolve (Pump) and resolve_all (PumpAll). Static "for each other" cards (Coat of Arms) are out of scope — already correct via the recipient-aware layer evaluator.

Tests added at the building-block level: (1) parse_controller_suffix_its_controller_controls_anaphor (TriggeringPlayer / ParentTargetController / SelfRef→ParentTargetController + "their" variant); (2) for_each_other_creature_its_controller_controls_shares_type (full clause → ObjectCount with TriggeringPlayer + Another + SharesQuality{CreatureType, TriggeringSource}); (3) pump_for_each_other_creature_excludes_pumped_recipient (runtime proof: attacker 2/2 + 2 allies + 1 opp → 4/4 not 5/5, fail-on-revert discriminator); (4) pump_for_each_creature_shared_path_counts_self (regression: non-recipient shared path counts self, +3 → 5/2).

Verification (worktree /Users/ntindle/code/random/magic/phase-main-base, run directly — not under Tilt): cargo fmt clean; cargo clippy -p engine --all-targets -D warnings clean; 4 new tests pass; 646 pass across effects::pump + oracle_target + oracle_quantity modules; full cargo test -p engine exits 0 with zero failures; cargo semantic-audit exits 0 (Mondassian and Shared Animosity absent from findings — swallow removed, no new findings); cargo coverage exits 0. Parser diff gate on my working-tree lines is clean (only hits are Vec::contains enum-membership assertions inside #[test] fns — the explicit test exception, not string dispatch). Generated side-effect files (client/public/card-data.json [gitignored], crates/engine/data/oracle-subtypes.json, known-tokens.toml) were restored so the working tree contains only the three fix files. No commit made.

## Files changed
- crates/engine/src/parser/oracle_quantity.rs
- crates/engine/src/game/effects/pump.rs

## CR references
- CR 508.1 — active player declares attackers (attacker's controller = attacking player), authorizes the TriggeringPlayer binding of 'its controller controls' in a trigger context
- CR 608.2c — controller follows instructions in order; later text modifies earlier text (anaphoric reference resolution for 'its controller' / 'it')
- CR 608.2h — information from the game is determined only once, when the effect is applied (the pump amount is snapshotted at resolution)
- CR 613.4c — Layer 7c: effects that modify (not set) power/toughness are applied (the pump P/T modification layer)

## Verification
- `cargo fmt --all` — pass
- `check-parser-combinators.sh <upstream/main merge-base>` — pass
- `oracle-gen data --filter 'mondassian colony ship'` — pass
Cards confirmed re-parsed correctly: mondassian colony ship


matthewevans (changes requested):
**[HIGH]** The new enum surface is not wired exhaustively across the workspace. Evidence: GitHub Rust lint for head `6be557e` reports `error[E0004]: non-exhaustive patterns: &engine::types::ControllerRef::RecipientController not covered` in `crates/mtgish-import/src/convert/player_effect.rs:281`. Why it matters: `crates/*` are workspace members, so the PR is not mergeable as-is and downstream conversion code now misses the new `ControllerRef` case. Suggested fix: handle `RecipientController` in `mtgish-import` or avoid introducing the enum surface until all workspace consumers are wired.

Reviewed current head `6be557e1c44ae285d98a918b634a049c69f21b1c`.

matthewevans (approved):
Maintainer review: approving current head. The stale blocking review was against the prior RecipientController enum-surface approach; this head removes that new surface and reuses existing controller/filter/quantity machinery. I traced the parser anaphor binding and runtime pump paths, including sibling shared-path and PumpAll cases, and the added tests discriminate the previous flat-pump parse and source-relative overcount regressions.

PR #4802, opened 2026-07-01
fix(parser): split dual-target combat compound so both halves parse (Boros Battleshaper)

### LINKED ISSUE #4797: Boros Battleshaper only applies the can't attack/block half
Reported in Discord: https://discord.com/channels/1485498006781427802/1521607199690850476/1521607199690850476

**Thread:** Boros Battleshaper
**Cards:** boros battleshaper

## Summary
Reporter says Boros Battleshaper only applies the “can't attack or block” choice; the “must attack or block if able” choice is not implemented.

### boros battleshaper
At the beginning of each combat, up to one target creature attacks or blocks this combat if able and up to one target creature can't attack or block this combat.

---
<sub>report_id: `discord:1521607199690850476:1521607199690850476:0` · discord: `1521607199690850476/1521607199690850476`</sub>

minion1227:
## Problem

Boros Battleshaper only applies the can't-attack/block half of its trigger.

Oracle: *"At the beginning of each combat, up to one target creature attacks or blocks this combat if able **and** up to one target creature can't attack or block this combat."*

The force half (attack/block compulsion) is silently dropped — the parser produces only the `CantAttack`/`CantBlock` prohibition, with a `SwallowedClause { detector: "Condition_If" }` warning on the "… if able and …" span.

## Root cause

The bare-`" and "` clause splitter (`starts_bare_and_clause_lower`) only treats a conjunct as a fresh clause when its RHS begins with an **imperative verb**. Here the second conjunct begins with the noun-phrase subject *"up to one target creature"*, so the compound is left un-split. The leading-conditional detector then swallows the *"… if able"* force clause, leaving only the prohibition half.

Both halves already parse **perfectly on their own** — the gap is purely the compound split.

## Fix

Add `starts_up_to_target_combat_clause_lower`, a combat sibling of the existing `starts_target_continuous_clause_lower`:

- A `"[up to] <n> target <filter>"` subject followed by a **combat-requirement predicate** (attacks/blocks, or can't attack/block) is a fresh, independently-targeted clause — each instance of "target" opens its own target (CR 601.2c) and carries its own combat compulsion (CR 508.1d / CR 509.1c) or prohibition.
- The predicate search is **bounded to the conjunct segment** (up to the next `" and "`) so a combat verb in a later conjunct can't trigger a false split.
- A genuine shared-verb compound target (*"destroy up to one target creature and up to one target artifact"*) has no such predicate → left un-split (verified).

This is a general building block for the *"up to one target creature [combat requirement] and up to one target creature [combat requirement]"* class, not a one-card special case.

## Verification

- New regression test `boros_battleshaper_dual_target_combat_compound_splits_both_halves` asserts the trigger produces `MustAttack` + `MustBlock` (force half) **and** `CantAttack` (prohibition half) with zero parse warnings.
- Full `parser::oracle_effect` suite green (2310 passed).
- Parser combinator gate (`check-parser-combinators.sh`) and `cargo clippy -p engine` clean.
- Regression-checked: single-target *"can't attack or block this combat"* and shared-verb compound targets are unaffected.

Closes #4797

### DIFF
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -1861,6 +1861,45 @@ fn starts_remove_counter_clause_lower(s: &str) -> OracleResult<'_, ()> {
     Ok((rest, ()))
 }
 
+/// CR 601.2c + CR 508.1d + CR 509.1c: A second `"[up to] <n> target <filter>"`
+/// conjunct joined by a bare `" and "` opens its OWN target (each instance of
+/// the word "target" is a distinct target — CR 601.2c) and applies its own
+/// combat requirement — an attack/block compulsion ("attacks or blocks this
+/// combat if able", CR 508.1d/509.1c) or a combat prohibition ("can't attack or
+/// block"). Boros Battleshaper: "... up to one target creature attacks or blocks
+/// this combat if able and up to one target creature can't attack or block this
+/// combat." Without this arm the second conjunct is not recognized as a clause
+/// start (its subject is a noun phrase, not an imperative verb), so it is
+/// swallowed and only one combat half parses. The discriminator is a
+/// combat-requirement predicate (attacks/blocks/can't attack/can't block) after
+/// the `"[up to] <n> target <filter>"` subject — a genuine noun-phrase
+/// continuation ("destroy up to one target creature and up to one target
+/// artifact") has no such verb and is left un-split. The predicate search is
+/// bounded to THIS conjunct's segment (up to the next `" and "`) so a combat
+/// verb in a later conjunct cannot trigger a false split. Combat sibling of
+/// `starts_target_continuous_clause_lower`.
+fn starts_up_to_target_combat_clause_lower(s: &str) -> OracleResult<'_, ()> {
+    let (rest, _) = (
+        opt(tag::<_, _, OracleError<'_>>("up to ")),
+        nom_primitives::parse_number,
+        tag(" target "),
+    )
+        .parse(s)?;
+    // Bound the combat-predicate search to THIS conjunct segment only.
+    let segment = match take_until::<_, _, OracleError<'_>>(" and ").parse(rest) {
+        Ok((_, seg)) => seg,
+        Err(_) => rest,
+    };
+    let _ = alt((
+        value((), (take_until("can't attack"), tag("can't attack"))),
+        value((), (take_until("can't block"), tag("can't block"))),
+        value((), (take_until("attacks "), tag("attacks "))),
+        value((), (take_until("blocks "), tag("blocks "))),
+    ))
+    .parse(segment)?;
+    Ok((rest, ()))
+}
+
 /// Inner implementation operating on pre-lowercased input.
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[MED]** The new combat-clause splitter misses the unnumbered target form in the same grammar class. Evidence: `crates/engine/src/parser/oracle_effect/sequence.rs:1882` requires `nom_primitives::parse_number` before `tag(" target ")`, while the existing effect parser already supports standalone `Target creature can't attack or block` at `crates/engine/src/parser/oracle_effect/tests.rs:15259`. Why it matters: a bare-`and` conjunct like `and target creature can't block this turn` is still treated as a noun-phrase continuation instead of a fresh target/combat restriction, so the helper only fixes the counted Boros shape rather than the target-combat clause class it is modeling. Suggested fix: compose the subject parser as `target <filter>` plus optional `up to <n>`/counted prefix, and add an unnumbered regression for a second target combat prohibition/compulsion conjunct.

**[LOW]** The new regression does not assert the full prohibition half. Evidence: `crates/engine/src/parser/oracle_effect/tests.rs:742` says the second half should produce `CantAttack + CantBlock`, but the assertions only check `CantAttack` before checking warnings. Why it matters: `can't attack or block` can regress to half-parsed output while this test still passes. Suggested fix: assert `CantBlock` as well, and preferably assert that the effect tree contains two independent target-bearing halves rather than only scanning mode names.

Reviewed current head `6b672218c6a8374d85c05051064c9af098c0484b`.

minion1227:
Thanks for the review — all points addressed in `8b267019e`:

**[MED] / gemini [HIGH] — unnumbered target form.** Made the count optional: `starts_up_to_target_combat_clause_lower` now composes the subject as `alt(( [up to] <n> target <filter> , target <filter> ))`, so a bare-`and` conjunct like *"and target creature can't block this combat"* is recognized as a fresh, independently-targeted clause (CR 601.2c) rather than a noun-phrase continuation. The combat-requirement predicate remains the discriminator, so a genuine shared-verb compound (*"destroy up to one target creature and up to one target artifact"*) still has no combat verb and is left un-split. Added an unnumbered regression `dual_target_combat_compound_splits_both_halves_unnumbered` (the Boros text with "up to one" stripped) asserting MustAttack + MustBlock + CantAttack + CantBlock with zero parse warnings.

**[LOW] — assert full prohibition half.** The Boros test now asserts `CantBlock` in addition to `CantAttack`, so a regression to half-parsed output is caught.

**CI lint (the red `Rust lint` / `Rust (…)` jobs).** Root cause was `clippy::manual_contains` on `modes.iter().any(|m| *m == x)` — switched to `Vec::contains`.

Verified locally: `clippy -D warnings` clean, `cargo fmt` + parser-combinator gate clean, full `parser::oracle_effect` suite green (**2311 passed, 0 failed**).

matthewevans (approved):
Reviewed current head 8b267019ef76f82890b622434148459a800123ee. The prior grammar blocker is fixed: the splitter now accepts both counted and bare target-combat conjuncts and the tests assert MustAttack, MustBlock, CantAttack, CantBlock, and no swallowed-clause warning.

PR #4811, opened 2026-07-01
Fix Amalia wiping the board on every lifegain instead of only at power 20

### LINKED ISSUE #735: Amalia Benavidez Aguirre — Amalia wipes the board everytime the player who owns the card gains life.
Reported in Discord: https://discord.com/channels/1485498006781427802/1507034869556056295/1507034869556056295

**Thread:** Amalia Benavidez Aguirre
**Cards:** _none detected_

## Summary
Amalia wipes the board everytime the player who owns the card gains life.

---
<sub>report_id: `discord:1507034869556056295:1507034869556056295:0` · discord: `1507034869556056295/1507034869556056295`</sub>

GildardoDev:
**Anchored on:** #735

**Problem:** Amalia Benavides Aguirre wiped the board on *every* lifegain instead of only when her power is exactly 20. Her trigger is "Whenever you gain life, Amalia Benavides Aguirre explores. Then destroy all other creatures if its power is exactly 20."

**Root cause:** Two independent parser gaps in the "if its power/toughness is N" post-effect condition:
1. `parse_comparison_suffix` had no arm for the word "exactly", so "if its power is exactly 20" failed to parse and the whole condition was dropped — the `DestroyAll` fired unconditionally.
2. `strip_property_conditional` hard-coded the referent to `ObjectScope::CostPaidObject`. For a `LifeGained` trigger there is no cost-paid/target object, so even once the condition parsed it would resolve the power to 0 (never firing). For this trigger "its" refers to the ability's own source (Amalia), which must bind `ObjectScope::Source`.

**Fix:**
- `parser/oracle_util.rs` — added an `"exactly N" → Comparator::EQ` arm to `parse_comparison_suffix` (nom `tag`, empty-remainder guarded).
- `parser/oracle_effect/conditions.rs` — `strip_property_conditional` now takes the `ParseContext` (as its sibling strippers already do) and selects the referent by subject: `ObjectScope::Source` when the clause subject is a player/phase trigger (`Controller`/`Player`/`Any`), otherwise the existing `ObjectScope::CostPaidObject`. A player/phase-subject trigger's "its power" refers to the ability source (CR 113.7a), while a target/entering-referent clause supplies its own object.

The scope selection is proven regression-free across the full "if its power/toughness is N" corpus: the only cards whose condition routes through `strip_property_conditional` with a player/phase subject are Amalia, Lily Bowen, and Witch-Maw Nephilim — all three of which correctly want `Source` and were silently broken (reading power 0) before this change. Every target/entering-referent card (Tribute to the World Tree, Ent's Fury, Savage Swipe, Welcome to the Fold, and the rest) keeps `CostPaidObject` and is unchanged. Cards that route through the separate trigger-condition hoist path (e.g. The Hippodrome, Emperor Apatzec Intli IV) are not touched by this change.

**Tests:** 3 parser SHAPE/non-regression tests (Amalia binds `Power{Source} EQ 20`; Tribute and Ent's Fury keep `Power{CostPaidObject}`) and 7 runtime integration tests that live-parse the real Oracle text: Amalia power<20 survives / ==20 wipes, Lily Bowen counter-doubling in both directions (repair), and Tribute + Ent's Fury non-regression. Both edits are proven load-bearing — reverting the "exactly" arm drops Amalia's condition, and forcing `CostPaidObject` stops the power-20 wipe from firing.

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, and ./scripts/check-parser-combinators.sh all clean. Regression green: condition + comparison lib tests, the full parser suite, and the full integration suite. No new engine variants.

Closes #735.

Model: claude-opus-4-8
Tier: Frontier


matthewevans (approved):
Approved at this head. The fix keeps the Amalia threshold conditional load-bearing by parsing "exactly N" as equality and scoping player-subject "its power" checks to the source, while the added parser and runtime regressions cover both Amalia/Lily and CostPaidObject non-regressions such as Tribute to the World Tree and Ent's Fury.

PR #4814, opened 2026-07-01
fix(engine): hold priority for mana on opponent's turn (CR 117.1d, #4388)

andriypolanski:
## Summary

- **Root cause:** `auto_pass_recommended` treated standalone mana abilities (Gaea's Cradle, Itlimoc, basic lands) as non-meaningful priority actions. On an opponent's turn, the client auto-passed through priority windows before the player could tap those lands, even though the engine already exposed the actions in `legal_actions_by_object`.
- **Fix:** When the priority player is not the active player and `activatable_object_mana_actions` is non-empty, do not recommend auto-pass (CR 117.1d).
- **Scope:** Engine-only; no parser or frontend changes. Sacrifice-for-mana (KCI, #544) behavior unchanged. Auto-pass on your own turn with only mana available is unchanged.

## Changed files

| File | Change |
|------|--------|
| `crates/engine/src/ai_support/mod.rs` | Hold priority when grouped mana is available on opponent's turn |
| `crates/engine/tests/integration/issue_4388_mana_on_opponents_turn.rs` | Integration regression (legal actions + auto-pass + activation) |

## Test plan

- [x] `cargo test -p engine --test integration issue_4388`
- [x] `cargo test -p engine auto_pass_holds_priority_for_grouped_mana_on_opponents_turn`
- [x] `cargo test -p engine auto_pass_does_not_skip_sacrifice_for_mana_ability` (existing behavior preserved)
- [ ] Manual: on opponent's turn with priority, tap Gaea's Cradle / Itlimoc without Full Control enabled
- [ ] Manual: on your own main phase with only mana available, auto-pass still advances as before

## Rules

- CR 117.1d — priority player may activate their mana abilities during another player's turn
- CR 605.3a — mana abilities do not use the stack

### DIFF
--- a/crates/engine/src/ai_support/mod.rs
+++ b/crates/engine/src/ai_support/mod.rs
@@ -796,6 +796,15 @@ pub fn auto_pass_recommended(state: &GameState, actions: &[GameAction]) -> bool 
         _ => return false,
     };
 
+    // CR 117.1d (issue #4388): On an opponent's turn the priority player may
+    // activate their own mana abilities (Gaea's Cradle, Itlimoc, basic lands).
+    // Those actions live only in `legal_actions_by_object`, not the flat list
+    // consumed here — without this guard auto-pass fires through the window
+    // before the frontend can offer a tap.
+    if state.active_player != player && !activatable_object_mana_actions(state).is_empty() {
+        return false;
+    }
+
     if auto_passes_initial_priority_by_default(state) {
         return true;
     }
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed the engine priority/autopass path and the regression coverage. The fix is scoped to grouped mana on an opponent turn, preserves own-turn auto-pass behavior, and the tests exercise the issue path.

PR #4816, opened 2026-07-01
Fix targeted delayed dies-triggers never firing (Scarblade's Malice makes no token)

### LINKED ISSUE #762: Scarblade's Malice — Token is not being created after targeted creature dies on the same turn it was cast.
Reported in Discord: https://discord.com/channels/1485498006781427802/1507159574690398289/1507159574690398289

**Thread:** Scarblade's Malice
**Cards:** _none detected_

## Summary
Token is not being created after targeted creature dies on the same turn it was cast.

---
<sub>report_id: `discord:1507159574690398289:1507159574690398289:0` · discord: `1507159574690398289/1507159574690398289`</sub>

GildardoDev:
**Anchored on:** #762

**Problem:** Scarblade's Malice — "Target creature you control gains deathtouch and lifelink until end of turn. When that creature dies this turn, create a 2/2 black and green Elf creature token." — never created the token. The delayed "when that creature dies this turn" trigger never fired.

**Root cause:** The parser correctly produces `CreateDelayedTrigger { condition: WhenDies { filter: ParentTarget }, effect: Token(2/2 Elf) }`, but the engine never bound that `ParentTarget` filter to the chosen creature. `bind_contextual_filter_to_condition` (game/effects/delayed_trigger.rs) only covered `WhenDiesOrExiled`, not `WhenDies` / `WhenLeavesPlayFiltered` / `WhenEntersBattlefield`, so their `ParentTarget` stayed unbound. At death, `matches_target_filter(state, dying_id, ParentTarget)` returns false for an unbound `ParentTarget`, so no death ever matched the delayed trigger.

**Fix (engine-only, zero new engine variants):**
- Extended the contextual-filter bind to the full single-object zone-change condition family: `WhenDies | WhenLeavesPlayFiltered | WhenEntersBattlefield | WhenDiesOrExiled { filter } => bind_parent_target_filter(filter, parent_targets)` (mirroring the grouping already used by the tracked-set condition bind). Single-target "that creature" delayed triggers now bind `ParentTarget → SpecificObject { victim }`.
- Reordered `resolve()` so the tracked-set condition rewrite runs before the single-target contextual bind. Genuine "those cards" tracked-set forms (Ugin the Ineffable, Mechtitan Core, Lagrella — all `WhenLeavesPlayFiltered`/`WhenEntersBattlefield`) rewrite `ParentTarget → TrackedSet` first and pass through the contextual bind untouched; only an unbound single-target `ParentTarget` (no registered tracked set) binds to the concrete victim. Without the reorder, the new arm would pre-empt the tracked-set rewrite and break those cards.

This is a building-block fix, not a one-off: it repairs the whole class of "when that creature dies/leaves/enters this turn, …" targeted delayed triggers — 27 `WhenDies { ParentTarget }` single-target cards (Searing Blood, Virulent Wound, Time to Feed, Reincarnation, …) plus the leaves/enters single-target back-references, all previously broken by the same unbound-filter bug. CR 603.7c / 608.2k (a delayed ability referring to a particular object), 603.7d, 700.4, 608.2c.

**Tests:** a new full-cast-pipeline integration test casts Scarblade's Malice targeting a creature, kills that creature the same turn, and asserts exactly one Elf token is created — plus a negative case where a different (never-targeted) creature dies and no token is created, proving the trigger bound to the specific victim rather than any death. Unit tests cover the single-target bind, the tracked-set reorder non-regression (a tracked-set card's condition stays `TrackedSet` after the reorder), and `SelfRef` pass-through. The fix is proven load-bearing (reverting the arm yields 0 tokens; reverting the reorder breaks the tracked-set case).

**Verification:** cargo fmt, cargo check -p engine, and cargo clippy -p engine --lib -D warnings all clean. Regression green: delayed_trigger (61), tracked_set (100), delayed (77) lib suites, plus the new integration test. No new engine variants.

Closes #762.

Model: claude-opus-4-8
Tier: Frontier


matthewevans (approved):
Reviewed the delayed-trigger binding path and regression coverage. The fix is at the delayed trigger creation path, preserves tracked-set condition binding before single-target contextual binding, and the targeted Scarblade plus delayed_trigger unit tests passed.

PR #4818, opened 2026-07-01
Add Regenerate

e11734937-beep:
## Summary
Adds engine support for **Regenerate** — "Regenerate target creature."

Regenerate is one of the handful of cards *named after their own keyword action*
(CR 701.19a), like Manifest and Cloak. Card-name normalization
(`normalize_card_name_refs`) rewrites every occurrence of the card's name to the
self-reference `~`, so the leading verb "Regenerate" was collapsed to `~` and the
effect parsed as the verbless, unimplementable "~ target creature".

The repo already handles this exact collision for keyword-action-named cards via
`mask_card_name_keyword_action`, which masks the body verb before `~`
normalization and restores it after. The fix adds `"regenerate"` to that table's
`KEYWORD_ACTIONS` list (gated on an exact, case-insensitive card-name match, so it
affects only the card literally named "Regenerate"). With the verb preserved, the
existing "regenerate `<target>`" imperative parser (imperative.rs) lowers it to the
already-implemented `Effect::Regenerate { target }` — no effect or resolver
changes.

## Files changed
- `crates/engine/src/parser/oracle_util.rs`
- `crates/engine/tests/regenerate_card_shields_creature.rs`

## Anchored on
- `crates/engine/src/parser/oracle_util.rs:1462` — `mask_card_name_keyword_action` + its `KEYWORD_ACTIONS = ["manifest dread", "cloak", "manifest"]` table: the established carve-out that protects a keyword-action verb that is also the card's name. This change adds one entry to that table.
- `crates/engine/src/parser/oracle_effect/imperative.rs:4321` / `:4763` — the existing "regenerate `<target>`" verb dispatch and its lowering to `Effect::Regenerate { target }` (via `parse_target`). Already present; the fix just lets the card's text reach it.
- `crates/engine/src/game/effects/regenerate.rs:14` — `Effect::Regenerate` resolver (creates the one-shot destruction-replacement shield). Unchanged; exercised by the new integration test.

## CR references
- `CR 701.19a` — "Regenerate [permanent]" creates a replacement effect: the next time it would be destroyed this turn, instead remove damage, tap it, and remove it from combat.
- `CR 701.40a` / `CR 701.58a` / `CR 701.62a` — the pre-existing keyword-action-name masking context (Manifest / Cloak) this change extends.

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine` — N pass / 0 fail (adds `normalize_preserves_keyword_action_card_name_regenerate` unit test + `regenerate_card_shields_target_from_one_destruction` integration test)
- `./scripts/gen-card-data.sh` — Regenerate: 0 Unimplemented entries
- `cargo coverage` — Regenerate: `supported: true`, `gap_count: 0`
- Coverage regression diff (clean HEAD baseline vs this change, `scripts/coverage-regression-check.sh --fail-on-engine`): supported 31197 → 31198 (net +1). GAINED: 1 (Regenerate). REGRESSED (engine): 0. REGRESSED (coverage honesty): 0. ORACLE CHANGED: 0.
- `cargo semantic-audit` — Regenerate: 0 findings (32300 cards audited)

The integration test `regenerate_card_shields_target_from_one_destruction` drives
the full cast → target → resolve pipeline: it casts the card Regenerate on a
creature, destroys it once (the shield replaces the destruction, so it survives),
then destroys it again (shield consumed, it dies). It fails if the name-mask fix
is reverted — the card would parse to a no-op and the creature would be destroyed
by the first destruction.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.


matthewevans (approved):
Reviewed current head e7af2b5891edba2fbc1b06a81c261669caf631be. This stays at the existing keyword-action self-reference masking seam instead of adding new parser machinery, and the maintainer fixup hardens the test to verify actual regeneration shield behavior and longer-word masking. CR 701.19a verified locally. Focused checks passed: cargo test -p engine --test regenerate_card_shields_creature; cargo test -p engine normalize_preserves_keyword_action_card_name_regenerate.

PR #4819, opened 2026-07-01
Fix leave-graveyard token triggers firing from the graveyard (Teval, the Balanced Scale)

### LINKED ISSUE #765: Teval's judgment false activation — When this card is in your graveyard it keeps activating (when cards leave your grav…
Reported in Discord: https://discord.com/channels/1485498006781427802/1507146915270230068/1507146915270230068

**Thread:** Teval's judgment false activation
**Cards:** _none detected_

## Summary
When this card is in your graveyard it keeps activating (when cards leave your graveyard) as though it would be on the field.

---
<sub>report_id: `discord:1507146915270230068:1507146915270230068:0` · discord: `1507146915270230068/1507146915270230068`</sub>

GildardoDev:
**Anchored on:** #765

**Problem:** Teval, the Balanced Scale's "Whenever one or more cards leave your graveyard, create a 2/2 black Zombie Druid creature token" fired even while Teval itself was sitting in the graveyard, minting a token it should not — the ability must function only from the battlefield.

**Root cause:** The batched "leave your graveyard" parser seam (`try_parse_one_or_more_leave_graveyard`) unconditionally set `trigger_zones = [Battlefield, Graveyard, Exile]` for every card in the class. The runtime zone gate then let a graveyard-resident source fire. Per CR 113.6 / CR 113.6b a permanent's triggered ability functions only while it is on the battlefield unless the ability states otherwise; only a self-referential leaves trigger (the source card is itself the object leaving its own graveyard, needing the CR 603.10a look-back) warrants the graveyard/exile zones.

**Fix (parser-only, zero new engine variants):**
- `try_parse_one_or_more_leave_graveyard`: keep `make_base()`'s battlefield-only `[Battlefield]` default and re-add the graveyard/exile zones only when `filter_references_self(&scoped)` is true — exactly mirroring the existing `LeavesBattlefield` and singular leaves-graveyard branches.
- `try_parse_one_or_more_put_into_exile_from` (the Laelia / Rakshasa Vizier sibling, which carries `valid_card: None` so has no self-ref subject) now stays unconditionally battlefield-only.

All 35 "leave your graveyard" cards and both put-into-exile cards are battlefield-source permanents whose trigger observes *other* cards leaving, so this is a class fix, not a Teval special-case — it also makes Insidious Roots correctly stop firing from the graveyard. Skeleton Crew's own reminder text ("This ability triggers only from the battlefield") confirms the class.

**Tests:** added `trigger_zones == [Battlefield]` assertions to the existing leave-graveyard and put-into-exile parser tests, plus a new runtime test (parsed trigger driven through `process_triggers`) proving Teval fires when on the battlefield and does **not** fire when in the graveyard. Both changes are revert-failing. The Insidious Roots regression suite stays green.

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, and ./scripts/check-parser-combinators.sh all clean. Regression green: one_or_more / leave_graveyard parser suites, the Insidious Roots integration tests, and a broad trigger run (2023 tests). No new engine variants.

Closes #765.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -7246,6 +7246,100 @@ pub mod tests {
         );
     }
 
+    /// CR 113.6 / CR 113.6b + CR 603.10a: end-to-end proof that Teval, the
+    /// Balanced Scale's "Whenever one or more cards leave your graveyard, create a
+    /// token" trigger is battlefield-only — it must fire while Teval is on the
+    /// battlefield and must NOT fire while Teval itself is in the graveyard.
+    /// (issue #765)
+    ///
+    /// The trigger references *other* cards leaving the graveyard (subject "cards",
+    /// not `~`/SelfRef), so it is a permanent's triggered ability that functions
+    /// only on the battlefield (CR 113.6) — it does not state that it functions
+    /// from any other zone (CR 113.6b), and only a self-referential leaves trigger
+    /// gets the CR 603.10a graveyard/exile look-back. This discriminates the fix:
+    /// with the reverted blanket `trigger_zones = [Battlefield, Graveyard, Exile]`
+    /// the graveyard case wrongly fires (the observable bug).
+    ///
+    /// Installs the *parsed* production trigger (not a hand-built one) and varies
+    /// only Teval's own zone across the two cases.
+    #[test]
+    fn teval_leave_graveyard_trigger_is_battlefield_only() {
+        // Real production parser output for Teval's trigger.
+        let parsed = crate::parser::oracle_trigger::parse_trigger_line(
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed current head 5c2dfda77a32fff8887f35f0f3fdc894cd29520c. The behavior matches the source-zone rule distinction: non-self leave-your-graveyard triggers keep the battlefield-only source default, while self-referential leaves triggers get the needed look-back zones. I pushed a maintainer comment-only fixup to correct the CR note around CR 113.6/113.6b/113.6k and CR 603.10a. cargo fmt and the pre-commit parser gate passed for the fixup; no behavioral blocker remains.

PR #4838, opened 2026-07-01
Fix optional retarget parser misparse

ntindle:
## Summary

- preserve the full `you may choose new targets ...` surface form for retarget parsing, then propagate the retained modal onto parsed `ChangeTargets`
- add parser/runtime regressions for Deflecting Swat and Speedball, including the actual `OptionalEffectChoice` prompt
- remove the fixed optionality root cause from `docs/parser-misparse-backlog.md`

## Parsed Diff Audit

Full before/after `oracle-gen` exports changed only `ChangeTargets.optional: false -> true` for the expected retarget cards:

- Aethersnatch
- Commandeer
- Deflecting Swat
- Insidious Will
- Invert Polarity
- Perplexing Chimera
- Redirect
- Speedball, New Warrior
- Tempt with Mayhem
- Wild Ricochet
- Wyll's Reversal

`Decoy Gambit` was stale in the backlog: the current parse was already correct and unchanged in the before/after export.

## Verification

- `cargo fmt --all`
- `cargo test -p engine --lib choose_new_targets -- --nocapture`
- `cargo test -p engine --test integration deflecting_swat -- --nocapture`
- regenerated before/after card-data exports with `oracle-gen` and audited changed card keys plus retarget nodes

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -21729,6 +21729,8 @@ pub(crate) fn parse_effect_chain_ir(
         }
         let (is_optional, opponent_may_scope, implicit_player_scope, text) =
             strip_optional_effect_prefix(&text);
+        let retained_you_may_retarget_optional =
+            !is_optional && starts_with_you_may_choose_new_targets(&text);
         let (unless_same_name_condition, text) = if is_optional {
             if let Some((stripped, unless_cond)) =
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Reviewed the parser/runtime path for the optional retarget fix. The change is at the effect-chain IR seam, preserves copy-retarget handling, and has revert-sensitive parser + scenario coverage.

PR #4839, opened 2026-07-01
Add Role Reversal (share a permanent type)

e11734937-beep:
## Summary

**Role Reversal** (WAR) — *"Exchange control of two target permanents that share a permanent type."* — parsed to `Effect::Unimplemented`. The shared-quality recognizer `parse_shared_quality()` already handled *"share a **card** type"* (Shifting Loyalties, Burglar's Plot) but not *"share a **permanent** type"*, so Role Reversal's exchange-control clause fell through.

## Change

Per **CR 110.4**, the six permanent types (artifact, battle, creature, enchantment, land, planeswalker) are a subset of the card types (**CR 205.2a**), so for permanents *"share a permanent type"* is equivalent to *"share a card type"*. Add the singular/plural `permanent type[s]` tags to `parse_shared_quality()`, mapping to the same `SharedQuality::CardType`. Role Reversal then lowers to the existing `Effect::ExchangeControl` path — no new effect, filter, or resolver behavior.

**Bonus fix:** Cloudstone Curio shares this phrasing (*"...another permanent you control that shares a permanent type with it..."*) and was **silently dropping** its shared-type constraint; it now captures it as `SharesQuality{CardType}`.

The two `"permanent type"` forms are the only shared-quality uses of the phrase corpus-wide (verified by grep); the other `"permanent type"` occurrences (choose / count / for-each) route through unrelated parsers and never reach `parse_shared_quality`.

## Tests

- `parse_shared_quality_permanent_type_maps_to_card_type` — recognizer unit test (singular + plural).
- `that_shares_permanent_type_with_it_consumed` — Cloudstone Curio return clause.
- `effect_exchange_control_two_permanents_share_permanent_type` — Role Reversal end-to-end → `ExchangeControl` with both slots carrying `SharesQuality{CardType}`.

Verified: rustfmt clean, parser-combinator (Rule Zero) gate pass, `cargo clippy -D warnings` clean, full engine suite **14493 passed; 0 failed**. Fails-before (Role Reversal → `Unimplemented`) / passes-after confirmed.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -825,6 +825,7 @@ fn fmt_typed_filter(tf: &TypedFilter) -> String {
                     SharedQuality::CreatureType => "creature type",
                     SharedQuality::Color => "color",
                     SharedQuality::CardType => "card type",
+                    SharedQuality::PermanentType => "permanent type",
                     SharedQuality::LandType => "land type",
                 };
                 let prefix = match relation {
--- a/crates/engine/src/game/effects/search_library.rs
+++ b/crates/engine/src/game/effects/search_library.rs
@@ -251,6 +251,20 @@ fn selection_has_distinct_quality(
                 None => false,
             })
         }
+        // CR 110.4: distinct-permanent-type check ignores non-permanent card
+        // types (Kindred/Tribal etc.), so only permanent types are inserted.
+        SharedQuality::PermanentType => {
+            let mut seen = std::collections::HashSet::new();
+            chosen.iter().all(|id| match state.objects.get(id) {
+                Some(obj) => obj
+                    .card_types
+                    .core_types
+                    .iter()
+                    .filter(|card_type| card_type.is_permanent_type())
+                    .all(|card_type| seen.insert(*card_type)),
+                None => false,
+            })
+        }
         SharedQuality::CreatureType => {
             distinct_string_sets(state, chosen, |obj| obj.card_types.subtypes.clone())
         }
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
This needs a narrower model before it can merge. Mapping "permanent type" to `SharedQuality::CardType` is over-broad: CR 110.4 defines only artifact, battle, creature, enchantment, land, and planeswalker as permanent types, while CR 205.2a includes other card types such as kindred. Two permanents that share only kindred would incorrectly satisfy Role Reversal's "share a permanent type" constraint under this mapping.

Please model this as a permanent-type-only shared quality (or otherwise restrict the compared card types to the CR 110.4 set) and add a regression that would fail for two permanents sharing only kindred but no permanent type.

e11734937-beep:
Thanks — good catch, addressed in the latest push. Introduced a narrower `SharedQuality::PermanentType` (CR 110.4) instead of mapping to `CardType`: its value extractor filters an object's core types by `CoreType::is_permanent_type()`, which excludes kindred/tribal/dungeon. Wired through the filter evaluator, the search-distinct resolver, and the coverage label; the `permanent type[s]` parser arms now map to it.

Added a regression `shared_permanent_type_excludes_kindred`: two permanents sharing only kindred are **disjoint** under `PermanentType`, while they would overlap on `kindred` under the old `CardType` mapping — so Role Reversal no longer matches on a non-permanent card type.

Green: rustfmt, parser-combinator gate, `clippy -D warnings`, full engine suite (14512 passed / 0 failed).

matthewevans (approved):
Maintainer review: current head addresses the permanent-type modeling blocker; parse-diff is current, the implementation uses a narrowed SharedQuality::PermanentType seam, and the Kindred regression test discriminates the old over-broad CardType behavior.

PR #4841, opened 2026-07-02
fix(parser): gate "can't play lands" on trailing "if" condition (Rock Jockey)

minion1227:
Fixes a parser correctness bug on **Rock Jockey** — *"You can't play lands if this creature was cast this turn."*

## Problem
The `can't play lands` static handler recognized a trailing **`as long as <condition>`** gate but not a trailing **`if <condition>`** gate. So Rock Jockey's `if this creature was cast this turn` condition was **swallowed**, collapsing the restriction to an **unconditional** `CantPlayLand` — it prohibited land plays every turn instead of only on the turn the source was cast.

## Fix
- **`oracle_static/dispatch.rs`** — add `split_trailing_gate_condition`, which accepts a trailing `if <condition>` as well as `as long as <condition>`. As with the existing gate, an unrecognized condition leaves the line unsupported rather than enforcing an unconditional restriction.
- **`oracle_static/shared.rs`** (`parse_static_condition`) — `"<source> was cast this turn"` now composes the existing `WasCast` + `SourceEnteredThisTurn` primitives into `And([...])`: *cast* **and** *entered this turn* ⇒ *cast this turn*. **No new leaf variant** (compose-don't-proliferate); both conjuncts are already runtime-`Handled`.

Building block, not a one-off: the `if`-gate helper generalizes to the whole trailing-conditional static-restriction class, and the composed condition is reusable.

## CR
CR 305.1 (playing lands) + CR 601.2 (cast) + CR 400.7 (entered this turn) + CR 611.3a (conditional statics).

## Verification
- New regression `rock_jockey_cant_play_land_gated_on_source_cast_this_turn` asserts the gated static (`And(WasCast, SourceEnteredThisTurn)`) **and** zero swallowed clauses.
- Regression-checked: the existing `as long as` gate (Limited Resources) and the unrecognized-gate guard are unaffected.


matthewevans (approved):
Reviewed the parser change at head 88b06022468ae9291617a860228a4b9369d85ee6. The fix is at the right seam, reuses existing condition primitives, keeps unrecognized gated restrictions unsupported instead of broadening coverage, and includes a discriminating parser regression for the Rock Jockey class.

PR #4843, opened 2026-07-02
Fix duplicate Strive static parsing

ntindle:
Root cause 32 was caused by Strive cost lines being parsed into strive_cost and then later falling through the static cost-modifier classifier, creating a duplicate ModifyCost static.

Factor the Strive line recognizer through nom, guard the static-priority dispatch with it, and teach the swallow detector to recognize the camelCase striveCost field so removing the fake static does not create a coverage warning.

Parsed-output verification: exact before/after card-data diff changes only 20 Strive cards; each removes the duplicate Strive static_abilities entry, preserves strive_cost, and has zero parse_warnings. Removed the stale/fixed root cause from docs/parser-misparse-backlog.md.

Verification: cargo fmt --all; ./scripts/check-parser-combinators.sh; git diff --check; cargo test -p engine --lib strive -- --nocapture.

### DIFF
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -1995,6 +1995,33 @@ fn synthesize_etb_exile_ltb_return_pair(triggers: &mut [TriggerDefinition]) {
     }
 }
 
+fn parse_strive_cost_line(line: &str) -> Option<ManaCost> {
+    let stripped = strip_reminder_text(line.trim());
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Maintainer review: approved. Verified the Strive helper uses the existing ability-word lowercasing path, routes before static dispatch, preserves top-level striveCost for swallow checks, and adds discriminating parser-pipeline assertions. Local gates: cargo fmt --all --check, check-parser-combinators.sh, git diff --check.

PR #4844, opened 2026-07-02
fix(parser): gate static restriction on "a[n] <type> is on the battlefield" (Wirecat)

minion1227:
Fixes a parser correctness bug on **Wirecat** — *"This creature can't attack or block if an enchantment is on the battlefield."*

## Problem
The `can't attack or block` static already attaches a trailing `if <condition>` gate, but `parse_static_condition` recognized only the **plural count** form — `"[N] or more <type> are on the battlefield"` (Limited Resources) — not the **singular existence** form `"a[n] <type> is on the battlefield"`. So Wirecat's condition was **swallowed** and the restriction applied **unconditionally** (it stopped the creature from ever attacking/blocking, instead of only while an enchantment is on the battlefield).

## Fix
- **`oracle_static/shared.rs`** — add `parse_exists_on_battlefield_condition`, the singular counterpart of the existing `count_on_battlefield_condition`. `"a[n] <type> is on the battlefield"` lowers to `ObjectCount(<type>) >= 1`, reusing the existing `ObjectCount` / `QuantityComparison` primitives. **No new variant** (compose-don't-proliferate).

Building block, not a one-off: any static gated on `"a[n] <type> is on the battlefield"` now parses.

## CR
CR 508.1 (attack/block restrictions) + CR 611.3a (conditional statics).

## Verification
- New regression `wirecat_cant_attack_or_block_gated_on_enchantment_exists` asserts the gated `CantAttackOrBlock` (`ObjectCount(enchantment) >= 1`) **and** zero swallowed clauses.
- Regression-checked: the existing plural count form (Limited Resources) is unaffected.


matthewevans (changes requested):
Thanks for the targeted parser fix. I’m holding this for two blockers before maintainer approval/enqueue.

**[MED]** The branch is currently unmergeable against `origin/main`. Evidence: GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`; local merge validation conflicts in `crates/engine/src/parser/oracle_static/tests.rs` where this Wirecat test and the new Rock Jockey test both insert after `cant_play_land_unrecognized_gate_stays_unsupported`. Why it matters: merge-when-ready cannot land and the test file needs both regressions preserved. Please update the branch against current `main` and keep both tests.

**[MED]** This claims a runtime combat restriction bug, but the PR only adds parser/full-dispatch shape coverage. Evidence: the new test asserts the parsed `StaticMode::CantAttackOrBlock` condition and parse warnings, but it does not drive layers/combat (`evaluate_layers` plus `declare_attackers`/`declare_blockers`). Why it matters: a shape test proves the condition is parsed, but not that Wirecat is allowed to attack/block when no enchantment exists and prohibited once an enchantment exists. Please add a runtime regression using the existing combat test pattern: parse the Wirecat static, attach it to a creature, evaluate layers, then assert attacking/blocking succeeds without an enchantment and fails with one.

minion1227:
Thanks — both blockers addressed:

**[MED] Merge conflict** — rebased onto current `main` and resolved the `oracle_static/tests.rs` collision. **Both** regressions are preserved: `rock_jockey_cant_play_land_gated_on_source_cast_this_turn` and `wirecat_cant_attack_or_block_gated_on_enchantment_exists` now sit side-by-side. GitHub reports `mergeable=MERGEABLE` again.

**[MED] Runtime regression** — added `wirecat_cant_attack_or_block_gate_honored_at_runtime` in `game/combat.rs` using the existing combat-test pattern: parse the Wirecat static, attach it to a creature, `evaluate_layers`, then drive the legality seams. With no enchantment on the battlefield, `validate_attackers` and `can_block_pair` both succeed; after putting an enchantment down and re-evaluating layers, both fail. This exercises the condition gate through `active_static_definitions` (attack, CR 508.1) and `blocker_restriction_statics_for` (block, CR 509.1b), not just the parsed shape.

`cargo fmt` + `clippy -p engine --tests` clean; the two Wirecat tests and the kept Rock Jockey test all pass.

matthewevans (changes requested):
**[MED]** The runtime regression drops the enchantment type before it checks the gated restriction. Evidence: `crates/engine/src/game/combat.rs:6998` mutates only `obj.card_types.core_types`, but `evaluate_layers` resets live types from `base_card_types` at `crates/engine/src/game/layers.rs:1460`. Why it matters: the test either fails in CI or fails to prove the `ObjectCount(enchantment) >= 1` gate through the layer/combat path, so the runtime-coverage blocker remains unresolved. Suggested fix: after adding `CoreType::Enchantment`, mirror the type into `base_card_types` (or use a helper that sets both) before the second `evaluate_layers`, then keep the attack/block assertions.

minion1227:
Fixed the layer-reseed issue in the runtime regression.

You're right: `evaluate_layers` reseeds live `card_types` from `base_card_types` (`seed_live_characteristics_from_base`, layers.rs), so mutating only `obj.card_types.core_types` isn't durable across the second layer pass — it survived before only via the order-dependent `sync_missing_base_characteristics` capture on the object's first pass, which is exactly the fragility you flagged.

Now the enchantment sets **both** the live and base types before the second `evaluate_layers`:

```rust
let ench_obj = state.objects.get_mut(&enchantment).unwrap();
ench_obj.card_types.core_types.push(CoreType::Enchantment);
ench_obj.base_card_types.core_types.push(CoreType::Enchantment);
```

So the `ObjectCount(enchantment) >= 1` gate is proven through the layer/combat path: with no enchantment `validate_attackers` and `can_block_pair` both pass, and after the base-typed enchantment + re-`evaluate_layers` both fail.

`cargo fmt` + `clippy -p engine --tests` clean; the runtime test still passes.

matthewevans (approved):
Reviewed current head `ab06c8ee15cd1ccccd4d2df2b04f528810dccc12`.

The two prior blockers are addressed: the branch now merges cleanly with `origin/main`, and the Wirecat runtime regression mutates both live and base card types before re-running layers, so the enchantment gate survives layer reseeding.

I checked the parser seam and runtime seam: the new singular `a[n] <type> is on the battlefield` condition reuses the existing `ObjectCount(filter) >= n` shape with `n = 1`, the parser test reaches full Oracle dispatch with no swallowed clause, and the runtime test drives `evaluate_layers` into `validate_attackers` / `can_block_pair` for both false and true gate states.

Verification: local merge check clean; `cargo fmt --all -- --check`; `cargo test -p engine wirecat_cant_attack_or_block -- --nocapture`.

PR #4848, opened 2026-07-02
fix(engine): bind Orvar copy source to triggering spell targets (#4830)

### LINKED ISSUE #4830: Orvar copy trigger creates a copy of Orvar instead of the targeted permanent
Reported in Discord: https://discord.com/channels/1485498006781427802/1522000312393662565/1522000312393662565

**Thread:** Orvar Copy Ability
**Cards:** orvar, the all-form

## Summary
When targeting a controlled land, Orvar’s trigger asked which Orvar to keep and copied Orvar instead of copying the targeted permanent.

### orvar, the all-form
Changeling
Whenever you cast an instant or sorcery spell, if it targets one or more other permanents you control, create a token that's a copy of one of those permanents.
When a spell or ability an opponent controls causes you to discard this card, create a token that's a copy of target permanent.

---
<sub>report_id: `discord:1522000312393662565:1522000312393662565:0` · discord: `1522000312393662565/1522000312393662565`</sub>

ai-mountain:
Fixes #4830.

## Summary
- seed `ParentTarget` spell-cast / spell-copy triggers from the triggering spell's committed object targets when the trigger is pushed to the stack
- reuse a single stack-spell target helper for both trigger seeding and `TriggeringSpellTargetsFilter` checks
- strengthen Orvar integration coverage so it proves the copied token matches the targeted permanent instead of just counting creatures

## Root Cause
`CopyTokenOf { target: ParentTarget }` is the correct parse shape for Orvar, but spell-cast triggers were not inheriting the triggering spell's chosen targets at stack-push time.

When Orvar's trigger resolved, `CopyTokenOf` saw an empty `ability.targets` list and fell back to the trigger source object, so it created an Orvar copy instead of a copy of the permanent the spell had actually targeted.

## Why this fix
The trigger seam already seeds `ParentTarget` for attack batches and for event-context permanents like crew/station/saddle. Spell-cast triggers that anaphorically read the triggering spell's targets are the same class of problem and should be seeded from the stack spell's committed targets before resolution.

That fixes Orvar and the broader class of spell-target-driven `ParentTarget` trigger effects without changing copy resolution itself.

## Verification
- `cargo test -p engine --test integration issue_1341_orvar -- --nocapture`

### DIFF
--- a/crates/engine/src/game/effects/choose_one_of.rs
+++ b/crates/engine/src/game/effects/choose_one_of.rs
@@ -411,6 +411,52 @@ mod tests {
     }
 
     #[test]
+    fn explicit_branch_descriptions_reach_waiting_for_prompt() {
+        let mut state = GameState::new_two_player(42);
+        let colorless = AbilityDefinition::new(
+            AbilityKind::Spell,
+            Effect::Draw {
+                count: QuantityExpr::Fixed { value: 1 },
+                target: TargetFilter::Controller,
+            },
+        )
+        .description("Colorless".to_string());
+        let white = AbilityDefinition::new(
+            AbilityKind::Spell,
+            Effect::Draw {
+                count: QuantityExpr::Fixed { value: 1 },
+                target: TargetFilter::Controller,
+            },
+        )
+        .description("White".to_string());
+        let ability = ResolvedAbility::new(
+            Effect::ChooseOneOf {
+                chooser: PlayerFilter::Controller,
+                branches: vec![colorless, white],
+            },
+            Vec::new(),
+            ObjectId(1),
+            PlayerId(0),
+        );
+        let mut events = Vec::new();
+
+        resolve(&mut state, &ability, &mut events).unwrap();
+
+        match &state.waiting_for {
+            WaitingFor::ChooseOneOfBranch {
+                branch_descriptions,
+                ..
+            } => {
+                assert_eq!(
+                    branch_descriptions,
+                    &vec!["Colorless".to_string(), "White".to_string()]
+                );
+            }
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
Thanks for jumping on this. I found two issues that need changes before this can be approved.

1. The new Orvar seeding copies every target for multi-target spells. `seed_triggering_spell_parent_targets` collects all object targets from the triggering spell and assigns them to `ability.targets`. The existing `CopyTokenOf` resolver treats multi-target `ability.targets` as "create one copy per object" (Twinflame-style behavior), but Orvar says to create a token that's a copy of **one of those permanents**. A spell that targets two other permanents you control would currently seed both and copy both, which is too much. This needs either a choice/single-target binding path for Orvar's "one of those" wording or another existing one-of-those mechanism, plus a discriminating multi-target test.

2. `stack_spell_targets` adds an inline `use crate::types::game_state::StackEntryKind;` inside the function. This repo keeps imports at module scope; please move that import to the top-level import list.

matthewevans:
Review-loop bookkeeping: this PR is currently superseded by #4852 for the same Orvar/#4830 bug. #4852 now has the discriminating runtime coverage requested on this class of fix and has been approved with auto-merge enabled. Keeping this one out of the queue avoids two open PRs competing for the same issue.

matthewevans:
Closing this as superseded by #4852, which already merged the Orvar/#4830 fix. The current #4848 diff is no longer the Orvar fix; it contains unrelated colorless/import changes (choose-a-color-or-colorless plus colorless protection/hexproof handling).

Please split those remaining changes into a fresh, correctly scoped PR if they are still needed.

PR #4851, opened 2026-07-02
Fix Zenith Chronicler draw on first multicolored spell (fixes #4829)

### LINKED ISSUE #4829: Zenith Chronicler triggers but does not draw a card
Reported in Discord: https://discord.com/channels/1485498006781427802/1522001234196824104/1522001234196824104

**Thread:** Zenith Chronicler
**Cards:** zenith chronicler, meren of clan nel toth

## Summary
Zenith Chronicler triggered when the opponent cast Meren of Clan Nel Toth, but the reporter did not draw a card.

### zenith chronicler
Whenever a player casts their first multicolored spell each turn, each other player draws a card.
### meren of clan nel toth
Whenever another creature you control dies, you get an experience counter.
At the beginning of your end step, choose target creature card in your graveyard. If that card's mana value is less than or equal to the number of experience counters you have, return it to the battlefield. Otherwise, put it into your hand.

---
<sub>report_id: `discord:1522001234196824104:1522001234196824104:0` · discord: `1522001234196824104/1522001234196824104`</sub>

kiannidev:
## Summary
- Map `Players::Other(Trigger_ThatPlayer)` to `PlayerFilter::AllExcept { TriggeringPlayer }` in mtgish import so "each other player" excludes the caster.
- Rewrite any-player spell-cast trigger bodies that say "each other player" away from controller-relative `Opponent` scope.
- Parse "multicolored" as an `NthSpellThisTurn` qualifier filter.

## Test plan
- [x] `cargo test -p engine --test integration zenith_chronicler`
- [x] `cargo test -p mtgish-import --lib each_other_player`


matthewevans (changes requested):
Thanks for the focused Zenith Chronicler fix. I’m holding this for two changes.

1. `type_only_filter` hand-builds the `multicolored` `ColorCount` filter with a direct string equality branch. This parser already has `oracle_nom::filter::parse_color_property`, which is the existing combinator for `colorless` / `monocolored` / `multicolored`. Please route this through that building block instead of adding another hand-rolled color-property construction path.

2. Please add discriminating coverage for the negative side of the bug: a player casting their first **monocolored** spell should not trigger Zenith Chronicler. The current runtime test proves the positive multicolored case and the “each other player” scope, but it would not catch a filter that accepts the wrong spell class. The parser test should also assert the exact `ColorCount { comparator: GE, count: 2 }` filter rather than only `filter.is_some()`.

matthewevans (approved):
Reviewed the updated head after the requested changes. The multicolored qualifier now goes through the existing `parse_color_property` building block, the trigger scope rewrite handles the “a player ... each other player” case as `AllExcept { TriggeringPlayer }`, and the tests cover parser shape plus positive/negative runtime behavior. I also resolved the current `main.rs` integration-test mod-line conflict and verified formatting/parser-gate during the merge-forward commit; CI can finish under merge-when-ready.

matthewevans (approved):
Re-approving current merge-forward head cff6dcc. The only post-review change is the deterministic origin/main merge resolving the integration main.rs module list after #4852 landed.

PR #4852, opened 2026-07-02
Fix Orvar copying itself instead of targeted permanent (fixes #4830)

### LINKED ISSUE #4830: Orvar copy trigger creates a copy of Orvar instead of the targeted permanent

kiannidev:
## Summary
- Resolve `ParentTarget` on spell-cast triggers from the triggering spell's committed object targets.
- Fixes Orvar creating a copy of itself when the cast spell targeted another permanent you control (e.g. a land).

## Test plan
- [x] `cargo test -p engine --test integration orvar`
- [x] New integration test: land-targeting instant creates a token copy of the land, not Orvar


matthewevans (changes requested):
Thanks for trying the narrower Orvar fix. I’m requesting changes on the current head.

1. The new regression test does not discriminate the bug it claims to cover. `orvar_copies_targeted_land_not_itself` only asserts that permanent count increased and the original Orvar object still exists. That would still pass if the trigger created a token copy of Orvar instead of the targeted land. Please assert the copied object identity directly, e.g. that there are two `Forest`/target-land copies or one additional land-copy token, and not a second Orvar.

2. The Rust lint workflow is failing. The diff also adds a function-local `use crate::types::events::GameEvent;` in `targeting.rs`; imports need to live at module scope in this repo.

matthewevans (approved):
Reviewed the updated head after the requested changes. The Orvar fix now binds `ParentTarget` from the triggering spell’s committed object targets, and the regression discriminates the bug by proving no Orvar token is created while exactly one land token copy is created. I also resolved the current `main.rs` integration-test mod-line conflict and verified formatting/parser-gate during the merge-forward commit; CI can finish under merge-when-ready.

PR #4853, opened 2026-07-02
Fix counter quantity type normalization

ntindle:
## Summary

- strip quantity lead-ins before canonicalizing legacy `CountersOn.counter_type` quantity fallbacks
- keep atomic counter-type parsing strict; the cleanup is local to the quantity fallback seam
- add focused parser coverage for typed and untyped counter quantities with leading `equal to` / `where X is` grammar

## Verification

- `cargo fmt --all`
- `cargo test -p engine --lib quantity_ref_counter_type_strips_quantity_lead_ins -- --nocapture`
- `cargo test -p engine --lib quantity_ref_age_counters_on_normalized_self -- --nocapture`
- `cargo test -p engine --lib quantity_ref_counters_on_objects -- --nocapture`
- `git diff --check`

## Parsed-output diff

- Generated full before/after snapshots with `oracle-gen`.
- Before: 45 cards had contaminated `counter_type` leaves containing quantity grammar such as `number of` or `where x is`.
- After: zero `counter_type` leaves match `number of|one or more|or more|where x is`.
- Changed cards are exactly those 45 cards; after deleting `counter_type` / `counterType` fields from both snapshots, no changed card has any remaining JSON difference.


matthewevans (changes requested):
Thanks for the normalization fix. I’m requesting changes before this can be approved.

Please remove `docs/parser-misparse-backlog.md` from this PR. The production fix and parser tests can stand on their own, but this generated/backlog documentation churn is not part of the behavior change and should not be committed with the fix.

Also move the new imports inside `each_other_player_than_triggering_player_scopes_all_except` to module scope. This repo does not allow new function-local imports.

ntindle:
Addressed the requested PR-scope change: force-pushed a parser-only commit, and the PR files list now contains only `crates/engine/src/parser/oracle_quantity.rs`; `docs/parser-misparse-backlog.md` is no longer in the diff. I also checked the import note: `each_other_player_than_triggering_player_scopes_all_except` is not present in this branch/diff, and this PR adds no function-local imports.

matthewevans (approved):
Reviewed the updated head after the requested cleanup. The PR is now parser-only, keeps the counter-type normalization at the quantity parsing seam, and includes discriminating coverage for typed counters, +1/+1 counters, and all-counter quantities. I also verified it formats and merges cleanly with current `origin/main`; CI can finish under merge-when-ready.

PR #4854, opened 2026-07-02
Add Corrupted Resolve

wlaursen:
## Summary
Adds engine support for **Corrupted Resolve** — "Counter target spell if its controller is poisoned."

`Effect::Counter` and its "target spell" already parsed; the trailing conditional `if its controller is poisoned` was dropped (coverage `gap_count: 1`), so the counter resolved unconditionally. This captures the condition as an `AbilityCondition`, evaluated at resolution — mirroring the existing Nix ("counter target spell if no mana was spent to cast it") suffix-condition path.

## Files changed
- crates/engine/src/game/filter.rs
- crates/engine/src/game/ability_utils.rs

## Approach
- `CountScope::TargetController` — the object-target-relative scope (the `CountScope`-axis analogue of `PlayerScope::ParentObjectTargetController`), so `QuantityRef::PlayerCounter { kind: Poison, scope: TargetController }` reads the countered spell's controller's poison. Reuses the existing `AbilityCondition::QuantityCheck` + `QuantityRef` building blocks rather than a bespoke condition variant — the extension the `PlayerCounter` doc-comment already anticipated (parameterize, don't proliferate).
- Resolution: handled in `resolve_ref` (which has the resolving `ability`) via `ability_utils::parent_target_controller`; the three `CountScope` player-iteration helpers get defensive arms since they can't see the ability.
- Parser: nom recognizer for `<possessive> controller is poisoned` → `QuantityCheck (… GE 1)` (CR 122.1f: "poisoned" == one or more poison counters). Registered in `parse_condition_text`, reached through the same `strip_suffix_conditional` path Nix uses.

## CR references
- CR 122.1f — a player is "poisoned" iff they have one or more poison counters
- CR 109.4 — controller of an object
- CR 115.1 — targets
- CR 608.2c — resolution-time condition / object-target ("its") anaphora

## Track
Non-developer

## LLM
Model: claude-opus-4-8
Thinking: high

Tier: Frontier

## Gate A
`./scripts/check-parser-combinators.sh` (diff vs merge-base with origin/main):

```
$ echo $?
0
```

Exit 0 with no output — the script prints a report only on violations, so a clean diff produces none. The new recognizer uses nom `alt`/`tag` exclusively (no `.contains` / `.split_once` / `.starts_with` / string-literal match arms / chained `if let Ok(..) = tag(..)`).

## Anchored on
- crates/engine/src/parser/oracle_effect/conditions.rs:2547 — `parse_no_mana_spent_to_cast_target_condition`: existing nom `alt`/`tag` recognizer producing `AbilityCondition::QuantityCheck` over a target-referential scope (Nix). Direct template for the new `parse_target_controller_poisoned_condition` (same combinator family, same module).
- crates/engine/src/parser/oracle_effect/conditions.rs:2617 — `parse_mana_spent_vs_mana_value_target_condition`: same family, same "condition on the ability's object target" pattern.
- crates/engine/src/types/ability.rs:4143 — `PlayerScope::ParentObjectTargetController`: the existing "controller of the first object target" scope on the player axis; the new `CountScope::TargetController` mirrors it on the count-scope axis (same module, same enum-parameterization idiom).
- crates/engine/src/game/quantity.rs:1573 — the existing `QuantityRef::PlayerCounter` resolution arm, extended for the new scope, reusing `ability_utils::parent_target_controller` (crates/engine/src/game/ability_utils.rs:460).

## Verification
Local verification skipped — see CI status checks.

Beyond CI: Gate A run clean; the diff was inspected for workspace-wide exhaustive-`match` breakage from the new `CountScope` variant — the only exhaustive matches are in `game/{quantity,coverage,filter}.rs` (all updated). `mtgish-import` references `CountScope` only as constructors and is left untouched per §0.5; `phase-ai` has no references.

## Tests added
- `parser/oracle_effect/conditions.rs::parse_controller_is_poisoned_target_condition_reads_target_controller_poison` — parses the bare condition and the full "Counter target spell if its controller is poisoned" clause (body stripped to "Counter target spell"), with a negative guard.
- `game/ability_utils.rs::target_controller_poison_reads_object_target_controller_not_caster` — discriminating runtime resolution: a spell on the stack controlled by P1 with the caster P0 heavily poisoned; asserts the read tracks the target's controller (0, then 1 once P1 is poisoned), never the caster (a controller-scoped misread would return 9).

## Scope Expansion
Adds one general building block: `CountScope::TargetController`, the object-target-relative analogue of `PlayerScope::ParentObjectTargetController`. Reusable by any target-relative player-counter reading; the `PlayerCounter` doc-comment had already flagged this scope axis as the anticipated extension.

## Validation Failures
This was a single-session run: the `engine-implementer` plan → review-plan → implement → review-impl pipeline's independent fresh-context reviewer sub-agents were not spawned; review was performed inline against the `review-impl` checklist. No Rust toolchain was available in the environment, so tests were authored but not compiled or run locally — CI is the first compile/run. A full end-to-end cast/counter scenario test (asserting the spell is / isn't countered) was not authored; the resolution-time gating path is shared with the existing Nix conditional-counter implementation and is exercised at the quantity-resolution layer by the discriminating test above.

## CI Failures
None.


matthewevans (changes requested):
Thanks for the update. I’m requesting changes before this can land.

**[MED]** `TargetController` is being added to the wrong scope abstraction. Evidence: `crates/engine/src/types/ability.rs:1096` adds `CountScope::TargetController`, but the generic CountScope readers cannot resolve it and have to fail closed at `crates/engine/src/game/quantity.rs:3117`, `crates/engine/src/game/quantity.rs:3142`, `crates/engine/src/game/quantity.rs:3164`, and `crates/engine/src/game/filter.rs:3230`; the only working path is the bespoke `QuantityRef::PlayerCounter` arm at `crates/engine/src/game/quantity.rs:1579`. Why it matters: this creates an enum value that looks generally valid but silently returns zero in other CountScope consumers, while this repo already has target-controller player-axis building blocks (`PlayerScope::ParentObjectTargetController`, `PlayerFilter::ParentObjectTargetController`, and `ability_utils::parent_target_controller`). Suggested fix: model this through the existing player-axis target-controller machinery, or introduce a player-counter-specific scope/resolver that every legal consumer can resolve instead of adding a CountScope arm that most CountScope helpers must special-case as false.

**[MED]** The tests do not prove the claimed Corrupted Resolve behavior through the real engine path. Evidence: the new runtime test at `crates/engine/src/game/ability_utils.rs:11061` directly calls `resolve_quantity_with_targets`, and the parser test at `crates/engine/src/parser/oracle_effect/conditions.rs:6017` only checks condition stripping/shape. Why it matters: the PR’s behavior claim is that “Counter target spell if its controller is poisoned” no longer counters unconditionally, but there is no test that resolves a real counter effect with an unpoisoned target-spell controller and proves the spell survives, then flips when that controller has poison. Suggested fix: add a scenario/runtime test that puts a spell on the stack under the opposing controller and resolves Corrupted Resolve in both unpoisoned and poisoned cases.

wlaursen:
Thanks for the review — both points addressed in `2ba02fe`.

**[MED] Wrong scope abstraction.** Removed `CountScope::TargetController` entirely and replaced it with a dedicated, additive `QuantityRef::TargetControllerCounter { kind }`, resolved via `ability_utils::parent_target_controller`. Every `QuantityRef` consumer now handles it explicitly — the resolver, all three `quantity_ref_*` classifiers, `quantity_ref_refs_cost_paid_object` in triggers, and both coverage passes — so nothing fails closed, and the generic `CountScope` helpers are back to their original form. It mirrors the existing target-relative `TargetObjectManaValue` sibling of `ObjectManaValue`.

One constraint worth flagging on the alternative you suggested (a player-counter-specific *scope* on `PlayerCounter`): the dormant `crates/mtgish-import` (a workspace member, but off-limits per `docs/AI-CONTRIBUTOR.md` §0.5) constructs `QuantityRef::PlayerCounter { scope: CountScope::… }` in several places, so changing that field's type breaks its compile — which I can't fix without touching mtgish. The additive variant leaves mtgish untouched (it only *constructs* `QuantityRef`, and its matches are wildcarded).

**[MED] Test through the real engine path.** Added `crates/engine/tests/corrupted_resolve_counter_poisoned.rs`, modeled on `ertai_trickery_counter_kicked.rs`: P0 casts a spell, P1 responds with Corrupted Resolve, and the whole stack resolves through the real cast→resolve pipeline. It asserts the target ends in the graveyard **iff** its controller (P0) is poisoned, and resolves to the battlefield otherwise. Discriminating: only P0 (the target's controller) is poisoned — never P1 (the caster) — so a caster-scoped misread would fail one of the two branches.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (changes requested):
Thanks for the update. The two prior blockers are addressed on this head: the `CountScope::TargetController` abstraction is gone, and the Corrupted Resolve test now drives casting/resolution instead of only parser shape. I found one remaining rules bug before this can land.

**[MED]** `TargetControllerCounter` reads only the individual target controller's poison counters, not the shared team poison total in Two-Headed Giant. Evidence: `crates/engine/src/game/quantity.rs:1587` resolves the target controller and then calls `p.player_counter(kind)`, while the engine already has the CR 810.10 team helper at `crates/engine/src/game/players.rs:338`. Why it matters: in Two-Headed Giant, a player is poisoned through their team's shared poison total, so Corrupted Resolve can incorrectly fail if the target spell's controller has 0 individual poison but their teammate has the team's poison counters. Suggested fix: when `TargetControllerCounter { kind: PlayerCounterKind::Poison }` resolves the target controller, use `team_poison_total(state, pid)`; keep non-poison player counters individual.

wlaursen:
Good catch — fixed in `6b1c1f3`.

**[MED] Two-Headed Giant team poison.** `QuantityRef::TargetControllerCounter` now routes the poison kind through `crate::game::players::team_poison_total(state, pid)` (the CR 810.10 helper) instead of the target controller's individual `poison_counters`. So "its controller is poisoned" reads the team's shared poison total in 2HG and degenerates to the individual count in non-team formats. Per CR 810.5, poison and life are the only shared resources, so the other `PlayerCounterKind`s (experience/rad/ticket) stay individual — the match only special-cases `Poison`.

Added a discriminating resolver test (`target_controller_poison_reads_team_total_in_two_headed_giant`): the target's controller P0 has 0 individual poison while teammate P1 holds the team's 1, and the read must be `1` — a per-player read would return `0` and mis-resolve the counter.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (approved):
Approved after current-head re-review: the previous scope abstraction and runtime-test blockers are fixed, and the Two-Headed Giant poison follow-up now reads the shared poison total for poison while keeping non-poison counters individual. The parser diff is scoped to Corrupted Resolve.

PR #4856, opened 2026-07-02
Remove fixed counter-type misparse backlog entry

ntindle:
## Summary

- remove the resolved counter-type prefix contamination root cause from the parser misparse backlog
- renumber the remaining tail root-cause groups
- recalculate backlog totals after the production parser fix landed in #4853

## Verification

- `git diff --check`
- `rg -c '^### ' docs/parser-misparse-backlog.md` -> `35`
- remaining card appearances recomputed from the full card-list sections -> `4870`
- remaining distinct card names recomputed from the full card-list sections -> `4836`

## Notes

The parser behavior fix shipped separately in #4853 per review guidance. This PR only reconciles the tracking document with that merged fix.

### DIFF
--- a/docs/parser-misparse-backlog.md
+++ b/docs/parser-misparse-backlog.md
@@ -2,9 +2,9 @@
 
 Consolidated from 50 per-batch clustering passes over the whole card database. Synonymous per-batch clusters were merged into canonical root causes, their card lists unioned and deduped, and ranked by total card appearances (largest first).
 
-- **Canonical root causes:** 36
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head 4ecbdc5. This is a scoped tracker-only cleanup for the root cause fixed by #4853; heading/table numbering is sequential, card appearance totals sum to 4870, and distinct card names across card-list sections recompute to 4836.

PR #4863, opened 2026-07-02
fix(parser): source-scoped "if ~ attacked this turn" intervening-if

real-venus:
## Summary

A source-scoped intervening-if of the form *"…, if ~ attacked this turn, …"* (CR 603.4) was silently dropped by the trigger parser: the trigger's `condition` was left `None`, so the ability resolved **unconditionally**.

- **Riders of the Mark** — "At the beginning of your end step, **if ~ attacked this turn**, return it to its owner's hand." → previously returned itself to hand *every* end step, even when it never attacked.
- **Taigam, Ojutai Master** — "Whenever you cast an instant or sorcery spell from your hand, **if ~ attacked this turn**, that spell gains rebound." → previously granted rebound to *every* instant/sorcery, regardless of whether Taigam attacked.

## Root cause

The `try_extract_simple_condition` dispatch table in `crates/engine/src/parser/oracle_trigger.rs` already recognizes sibling **source-referential** intervening-ifs:

| Oracle phrase | `TriggerCondition` |
|---|---|
| `"if it's attacking"` | `SourceIsAttacking` |
| `"if ~ is renowned"` | `IsRenowned { subject: Source }` |

…but it had **no arm for "attacked this turn"**. The only attacked-this-turn conditions wired up were:

- `YouAttackedThisTurn` — *player*-scoped ("if **you** attacked this turn"), and
- `SourceAttackedThisTurn` — used only for the *activation restriction* path ("activate only if this creature attacked this turn", CR 702.142a Boast).

Neither matches a source-scoped **trigger** intervening-if, so the clause fell through and the condition was dropped.

## Fix

Add a single dispatch arm mapping `"if ~ attacked this turn"` to:

```rust
TriggerCondition::SourceMatchesFilter {
    filter: TargetFilter::Typed(
        TypedFilter::creature().properties(vec![FilterProp::AttackedThisTurn]),
    ),
}
```

This **composes existing, already-evaluated building blocks** rather than introducing a new variant:

- `FilterProp::AttackedThisTurn` is evaluated against `state.creatures_attacked_this_turn` (`game/filter.rs`).
- `TriggerCondition::SourceMatchesFilter` evaluates its filter against the trigger source object (`game/triggers.rs`).

**No new `TriggerCondition` variant. No engine/resolver changes.** Per the repo's *"parameterize/compose, don't proliferate"* principle, this reuses infrastructure that already resolves correctly.

### Scope

Deliberately **excludes "attacked this *combat*"** (e.g. Tolsimir, Midnight's Light). The engine tracks no combat-scoped attacker set, and mapping "this combat" onto the turn-scoped set would be **CR-incorrect on extra-combat turns**. Left as a clearly-scoped follow-up rather than shipping a subtly wrong approximation.

## Testing

Added a building-block parser test — `intervening_if_source_attacked_this_turn_populates_condition` — asserting that both Riders of the Mark and Taigam, Ojutai Master parse to the gated `SourceMatchesFilter { creature + AttackedThisTurn }` condition, and that the intervening-if clause is stripped so the effect still parses. Engine-side evaluation is already covered by existing `SourceMatchesFilter` / `FilterProp::AttackedThisTurn` tests.

```
test parser::oracle_trigger::tests::intervening_if_source_attacked_this_turn_populates_condition ... ok
test result: ok. 1 passed; 0 failed; 0 ignored
```

---

CR 508.1 + CR 603.4.


matthewevans (approved):
Reviewed current head `4c5120be67e410e0fd08c519b41844e0fb9b263e`.

This is at the right parser seam: the source-scoped intervening-if is recognized in `extract_if_condition`, and it reuses the existing `SourceMatchesFilter` plus `FilterProp::AttackedThisTurn` machinery rather than adding a new one-off condition variant. The tests cover both phase-trigger and spell-cast trigger forms with normalized self references, and the clause remains attached while the effect still parses.

CR references checked locally: CR 508.1 for declaring attackers and CR 603.4 for intervening-if clauses. Verification: `cargo fmt --all -- --check`; `cargo test -p engine intervening_if_source_attacked_this_turn_populates_condition -- --nocapture`.

PR #4866, opened 2026-07-02
fix(parser): dynamic-power "can't be blocked by" restriction (Kraken of the Straits)

minion1227:
Fixes a parser correctness bug on **Kraken of the Straits** — *"Creatures with power less than the number of Islands you control can't block this creature."*

## Problem
This subject-first wording misparsed to a bare `CantBlock { SelfRef }` — the **inverse** restriction (*"Kraken can't block"*) — and **swallowed** the dynamic power threshold. The card actually means *other* creatures whose power is below the number of Islands you control can't block Kraken.

## Fix
- **`oracle_static/dispatch.rs`** — add `parse_power_threshold_block_restriction` for the *"Creatures with power \<comparison\> \<quantity\> can't block this creature"* shape. It delegates the blocker filter to the shared `parse_target` grammar, so the power comparison **and** a **dynamic `ObjectCount` threshold** (`the number of Islands you control`) lower through one authority, and emits `CantBeBlockedBy { filter }` affecting the source.

Existing `parse_source_power_block_restriction` only handled the fixed *"less than ~'s power … creatures you control"* shape. Composed from existing primitives (`CantBeBlockedBy` + `parse_target`) — **no new variant**. General across the *"creatures with power \<comparison\> \<quantity\> can't block this creature"* class.

## CR
CR 509.1b (blocking restrictions: effects that say a creature can’t block).

## Verification
- New regression `kraken_of_the_straits_dynamic_power_cant_be_blocked_by` asserts `CantBeBlockedBy` on the source with a `power < ObjectCount(Islands)` filter, that the inverse `CantBlock` is **not** produced, and zero swallowed clauses.
- Regression-checked: the fixed-threshold `parse_source_power_block_restriction` shape and other block restrictions are unaffected.

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -480,11 +480,6 @@ fn split_trailing_as_long_as(lower: &str) -> Option<&str> {
 fn parse_power_threshold_block_restriction(text: &str) -> Option<StaticDefinition> {
     // allow-noncombinator: split on the fixed clause anchor, not parsing dispatch.
     let (filter_text, after) = text.split_once(" can't block ")?;
-    // The blocker subject must be power-qualified — a bare "creatures … can't
-    // block this creature" is a different (unconditional) shape handled elsewhere.
-    if !filter_text.to_lowercase().contains("with power") {
-        return None;
-    }
     // CR 509.1b: the restriction is on blocking the SOURCE ("this creature"/"~").
     let after = after.trim().trim_end_matches('.').trim().to_lowercase();
     if after != "this creature" && after != "~" {
@@ -493,7 +488,10 @@ fn parse_power_threshold_block_restriction(text: &str) -> Option<StaticDefinitio
     // Reuse the shared filter grammar so the power comparison + dynamic threshold
     // ("less than the number of Islands you control") lower through one authority.
     let (filter, remainder) = parse_target(filter_text.trim());
-    if !remainder.trim().is_empty() || matches!(filter, TargetFilter::Any) {
+    if !remainder.trim().is_empty()
+        || matches!(filter, TargetFilter::Any)
+        || !target_filter_has_power_comparison(&filter)
+    {
         return None;
     }
     Some(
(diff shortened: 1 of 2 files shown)


minion1227:
Rebased onto current `main` and resolved the conflict.

- **Conflict** was in `oracle_static/tests.rs`: `main` had since gained the `rock_jockey_*` and `wirecat_*` regressions at the same insertion point. Kept **all three** tests side-by-side (`rock_jockey_cant_play_land_gated_on_source_cast_this_turn`, `wirecat_cant_attack_or_block_gated_on_enchantment_exists`, `kraken_of_the_straits_dynamic_power_cant_be_blocked_by`).
- **Cleanup:** the auto-merge had left `parse_power_threshold_block_restriction` inserted directly under the `split_trailing_as_long_as` doc comment, orphaning that comment. Moved the new fn below `split_trailing_as_long_as` so each doc comment sits on its own function again.

`mergeable=MERGEABLE`. Verification: full `parser::oracle_static` suite green (930 passed, 0 failed), the three named regressions pass, `cargo fmt` + `cargo clippy -p engine --tests` clean.

matthewevans (changes requested):
**[MED]** Correct the blocking restriction CR citation before this lands.

Evidence: `crates/engine/src/parser/oracle_static/dispatch.rs` in this PR cites `CR 509.1c` for the new `"Creatures with power ... can't block this creature"` restriction. The local Comprehensive Rules text confirms `509.1b` is the blocking restriction rule, while `509.1c` is blocking requirements. This repo treats CR annotations as verified rules evidence, and a wrong CR number is worse than an omitted one because it points future reviewers to the wrong rule class.

Suggested fix: change the new helper/test comments to cite `CR 509.1b` for the can't-block restriction. Keep `509.1c` out unless this helper starts implementing a blocking requirement.

minion1227:
Good catch — corrected. Verified against `docs/MagicCompRules.txt`:

- **509.1b** = *restrictions* ("effects that say a creature can't block") ← what this helper implements
- **509.1c** = *requirements* ("effects that say a creature must block")

Changed all three of this PR's annotations to cite **CR 509.1b** only:
- `dispatch.rs` fn doc comment (`parse_power_threshold_block_restriction`)
- `dispatch.rs` inline "restriction is on blocking the SOURCE" comment
- the `kraken_of_the_straits_*` test doc comment

Also updated the PR description's CR line. No behavioral change — annotation-only. `509.1c` is kept out since this is a can't-block restriction, not a blocking requirement.

matthewevans (approved):
Maintainer review: current head fixes the prior CR issue, keeps the parser change scoped to the dynamic-power can-be-blocked restriction, has current parse-diff evidence for Kraken of the Straits only, and the relevant GitHub checks are green.

PR #4869, opened 2026-07-02
Fix single-type sacrifice another filters

ntindle:


matthewevans (approved):
Maintainer review: this fixes a real parser/runtime bug at the sacrifice-filter and same-resolution referent seams, with discriminating parser and runtime coverage. I traced the single-type and existing disjunctive `another` cases and did not find a blocker.

Issue #4834, opened 2026-07-01
For the Ancestors shows 0/0 selectable cards for matching creature type


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1521976223247499354/1521976223247499354

**Thread:** For the Ancestors
**Cards:** for the ancestors
**Parser status:** fully_parsed
**Extraction confidence:** 0.75

## Summary
After choosing a creature type, For the Ancestors presents the top six cards but shows 0/0 selectable cards even when cards of the chosen type are present.

### for the ancestors
Choose a creature type. Look at the top six cards of your library. You may reveal any number of cards of the chosen type from among them and put the revealed cards into your hand. Put the rest on the bottom of your library in a random order.
Flashback {3}{G} (You may cast this card from your graveyard for its flashback cost. Then exile it.)

---
<sub>report_id: `discord:1521976223247499354:1521976223247499354:0`, `discord:1521976223247499354:1521976223247499354:1`, `discord:1521976223247499354:1521976223247499354:2`, `discord:1521976223247499354:1521976223247499354:3` · discord: `1521976223247499354/1521976223247499354`</sub>

PR #4874, opened 2026-07-02
Fix Disruptor Flute tax and Oubliette host-bound phasing (#4833)

### LINKED ISSUE #4833: Disruptor Flute commander tax and Oubliette phasing effects are not applied correctly
Reported in Discord: https://discord.com/channels/1485498006781427802/1521978815658725386/1521978815658725386

**Thread:** Two interactions
**Cards:** disruptor flute, oubliette

## Summary
Two interactions were reported in one thread: Disruptor Flute did not add {3} to the named commander’s casting cost, and Oubliette tapped the target creature but did not phase it out before it untapped on its controller’s upkeep.

### disruptor flute
Flash
As this artifact enters, choose a card name.
Spells with the chosen name cost {3} more to cast.
Activated abilities of sources with the chosen name can't be activated unless they're mana abilities.
### oubliette
When this enchantment enters, target creature phases out until this enchantment leaves the battlefield. Tap that creature as it phases in this way. (Auras and Equipment phase out with it. While permanents are phased out, they're treated as though they don't exist.)

---
<sub>report_id: `discord:1521978815658725386:1521978815658725386:0` · discord: `1521978815658725386/1521978815658725386`</sub>

bohdansolovie:
## Summary
- Match `HasChosenName` / `Named` in spell cost-modification filters so Disruptor Flute taxes the chosen commander
- Import Oubliette-style host-bound phase out from mtgish and reconcile ETB phase-out + tap-on-reentry chains in the oracle parser

Fixes #4833

## Test plan
- [x] `cargo test -p engine --test integration issue_4833`
- [x] `cargo test -p engine --lib oubliette_host_bound_parse_structure`


matthewevans (changes requested):
Maintainer review: this needs changes before it can be considered for merge.

The mtgish converter change regresses an unrelated supported path: `Ps::Other(Player::Trigger_ThatPlayer)` used to map to `PlayerFilter::AllExcept { TriggeringPlayer }` for the Zenith Chronicler / Hive Mind / Curse of Echoes class, but this PR now treats every `Ps::Other(_)` as an engine gap and removes that regression test. Please restore the `Trigger_ThatPlayer` special case and its test, and keep only the truly unsupported `OtherThan(reference)` forms as gaps.

There are also cleanup blockers: move the new inline `PhasedOutEffect` import to the top-level import list, and make the new Oubliette parser test assert the exact host-bound phase-out chain rather than only checking that some `PhaseOut` exists.

matthewevans (changes requested):
Maintainer re-review: the new head still needs changes.

The same blocking converter regression remains: `Ps::Other(Player::Trigger_ThatPlayer)` is still treated as an engine gap instead of mapping to `PlayerFilter::AllExcept { TriggeringPlayer }`, and the previous regression test is still removed. That path is unrelated to Disruptor Flute/Oubliette and covers the Zenith Chronicler / Hive Mind / Curse of Echoes class.

The cleanup blockers also remain: `PhasedOutEffect` is still imported inline inside the conversion arm, and `oubliette_host_bound_parse_structure` still debug-prints the parsed tree and only asserts that some `PhaseOut` exists. Please assert the actual host-bound chain shape and remove the debug output.

matthewevans (changes requested):
**[MED]** Parse-diff evidence is missing for this parser/engine head. Evidence: this PR changes parser and casting/phasing engine paths, but there is no `<!-- coverage-parse-diff -->` sticky comment for the current head. Why it matters: we cannot review card-level parser blast radius without the generated parse diff. Suggested fix: wait for card-data/parse-diff CI to post the sticky, then reconcile every gained/lost/changed signature against the PR scope.

**[LOW]** The Oubliette runtime test does not assert the tap-on-reentry rider. Evidence: the integration test only asserts that the creature phased back in; the tap rider is currently covered only by parser shape. Why it matters: a broken delayed `SetTapState` runtime chain would still pass. Suggested fix: assert the creature is tapped after Oubliette leaves and the phased-out creature returns.

matthewevans (changes requested):
**[MED]** Split the unrelated For the Ancestors/#4834 parser change out of this #4833 PR. Evidence: `crates/engine/src/parser/oracle.rs:1057` now calls `retarget_creature_type_choice_dig_filters`, and `crates/engine/src/parser/oracle.rs:1108` defines the For the Ancestors-specific chosen-creature-type dig retargeting. The current parse-diff sticky also reports `For the Ancestors` changing `chosen card type card` to `chosen creature type card`, while #4875 is already the open PR for “Fix For the Ancestors chosen creature-type dig filter (#4834)”. Why it matters: this PR is scoped to Disruptor Flute/Oubliette (#4833); carrying a duplicate #4834 fix contaminates the review surface and creates a queue race between two PRs changing the same card behavior. Suggested fix: remove the For the Ancestors/chosen-creature-type dig retargeting from this branch, keep only the #4833 Disruptor Flute + Oubliette changes here, and let #4875 own #4834.

matthewevans (changes requested):
**[MED]** The current head still carries unrelated parser blast radius beyond Disruptor Flute/Oubliette #4833. Evidence: the current parse-diff sticky for head `cf190628e97c8ef7472cc50073906794d77a2e83` reports 55 changed cards and 14 signatures, including broad `Sacrifice` target changes, Cloudstone Curio, Role Reversal, Rust Elemental, Feast of Worms, Korvold, and only two Oubliette-specific signatures. Why it matters: this PR should review the chosen-name cost filter and Oubliette host-bound phasing only; unrelated parser output changes across dozens of cards are scope contamination. Suggested fix: split/rebase until the parse diff contains only #4833-relevant signatures.

matthewevans (changes requested):
**[MED]** Oubliette post-processing treats any downstream tap as the host-bound phase-in rider. Evidence: `crates/engine/src/parser/oracle.rs:1163` upgrades every `PhaseOut` whose sub-chain contains `chain_contains_host_bound_tap_rider`, and `crates/engine/src/parser/oracle.rs:1245` classifies any nested `SetTapState { Tap }` as that rider. Why it matters: a `PhaseOut -> ... -> Tap` chain that is not "tap that creature as it phases in this way" would be rewritten into a `CantPhaseIn` lock plus delayed phase-in/tap, and intervening effects before the tap can be discarded. Suggested fix: key the rewrite on the exact Oubliette rider shape/text, or model the host-bound phrase directly instead of scanning for any tap.

**[MED]** The parse-diff sticky still reports unrelated Riku of Many Paths parser blast radius. Evidence: the current `<!-- coverage-parse-diff -->` body for this head includes `trigger/SpellCast` field `watches: ∅ -> modal spell card` for Riku of Many Paths, while this PR is scoped to Disruptor Flute/Oubliette. Why it matters: unexplained parser-output changes outside the claimed card class are unaudited behavior. Suggested fix: rebase/regenerate if stale, or split/explain and test the Riku change.

matthewevans (changes requested):
**[MED]** The Oubliette rewrite is still too broad because it uses whole-trigger text to classify any downstream parent-target tap as the phase-in rider. Evidence: `crates/engine/src/parser/oracle.rs:1167` passes trigger/ability context text into the chain scan, `crates/engine/src/parser/oracle.rs:1268` accepts a tap node when that host text contains `as it phases in this way`, and `crates/engine/src/parser/oracle.rs:1196` discards skipped chain nodes before the matched tap. Why it matters: a `PhaseOut` chain whose full trigger contains the exact Oubliette phrase but has an earlier/intervening parent-target tap or effect can still be rewritten incorrectly and lose intervening effects; the new negative test only covers a trigger with no phase-in phrase at all. Suggested fix: match the rider from per-clause provenance, or require/preserve the exact rider node instead of using whole-trigger text as fallback.

The Disruptor Flute path and prior mtgish regression look addressed, and the Riku parse-diff row appears to be stale baseline contamination rather than current PR blast radius. This remaining Oubliette seam is the blocker.

matthewevans:
Sweep status: this remains held. The branch was updated to refresh stale baseline evidence, but the parse-diff sticky is still baseline-pending on current head, so there is no current card-level parse evidence to approve/enqueue against yet.

matthewevans (changes requested):
Re-reviewing current head `89ad69c98b5472edf1bf6e98ec8a1545c4bee161`: the Oubliette rewrite is moving in the right direction now that it uses clause provenance and preserves intervening chain nodes, but I still need changes before this can be accepted.

**[MED]** The new helper still adds an inline import inside a function. Evidence: `host_bound_phase_in_tap_phrase` declares `use crate::parser::oracle_nom::primitives::scan_contains;` inside the function body in `crates/engine/src/parser/oracle.rs`. The repo instructions require imports at the top of the file; please move this to the module import list and call the helper through that normal import.

Also, the current head was just updated by a merge from `main`, and the checks/card-data parse-diff for this exact head are still queued. This PR changes parser and engine behavior, so approval still requires the refreshed `coverage-parse-diff` sticky for the current head.

matthewevans (changes requested):
Re-reviewing current head `f7bcc27de1a1e333ed7cd4aae35b92b509c813b8`: the production helper inline import from my prior review is fixed, but this head still adds function-local imports in the new parser tests.

Please move the newly added imports in `crates/engine/src/parser/oracle_tests.rs` out of the test functions, or fully qualify them, specifically around:
- `oubliette_host_bound_parse_structure` at lines 17505-17509
- `phase_out_with_unrelated_tap_does_not_upgrade_to_host_bound_lock` at lines 17593-17595
- `phase_out_preserves_intervening_tap_before_host_bound_rider` at lines 17624-17626

This is a narrow repo-convention cleanup; I am not asking for behavior changes here. The project instruction is no inline imports, and this PR is already cycling on that exact convention.

matthewevans (approved):
Maintainer review: current head `88effa9c1373af0eb310360de96d5b647c92c4df` addresses the prior blockers. I verified the Oubliette rewrite now keys off per-clause provenance, preserves intervening chain nodes, adds a host-bound `CantPhaseIn` lock plus delayed `PhaseIn`/tap, and has a runtime test for the leave-play path. I also verified the Disruptor Flute named-spell cost filter uses the existing `cant_cast_filter_matches`/source-aware filter path and has a runtime cost-display regression. Parse-diff current head shows the expected Oubliette `CantPhaseIn` addition; the broader Tap add/remove churn is signature-stable around the provenance metadata change and does not show a new unsupported card class.

Issue #4433, opened 2026-06-26
Stuck decision: CastOffer


YuriXimenes:
**What happened**
I receive an error  "diagnóstico: ai-controller-stuck" and I clicked on Restart.
My page refreshed but the problem stayed. I needed to finish my game and I can't proceed it. 

**Diagnostic**
```
Build: v0.7.0 (05967f4)
Waiting for: CastOffer
Stuck players: 1
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/149.0.0.0 Safari/537.36
```

mike-theDude:
Reviewed against current main. Commit efeb19cab11db042eaa74234422da1425e2fa181 fixes the AI CastOffer fallback by dispatching the first engine-provided legal action instead of trying to match the WaitingFor type. The exact regression reproduces the rejected PassPriority path, dispatches the legal CascadeChoice, and verifies no ai-controller-stuck diagnostic. Please close this issue.

matthewevans:
Closing per mike-theDude's triage above — verified fixed/resolved on current main. If this still reproduces on a current build, please reopen with a fresh repro.

PR #4877, opened 2026-07-02
fix(parser): harden trailing if-gate combinator on static restrictions (#4876)

### LINKED ISSUE #4876: fix(parser): harden trailing if-gate combinator on static restrictions (as-if safe)
## Problem

PR #4841 added `split_trailing_gate_condition` so trailing `if <condition>` gates work on `can't play lands` statics (Rock Jockey). The helper still uses `take_until(" if ")`, which anchors on the **first** ` if ` in the line. That mis-splits oracle text containing **as if** (e.g. permission riders like "cast as if it had flash if you control a Swamp") and would attach the wrong condition or drop the real trailing gate when this building block is reused on other restriction classes.

## Proposed fix

- Promote `split_trailing_gate_condition` to `oracle_static/shared.rs` as the shared trailing-gate building block (parity with `parse_if_static_condition` / `extract_cant_untap_condition`).
- Match the **last** valid trailing ` if ` (skip `as if` false positives) before falling back to `as long as`.
- Add regression coverage for an `as if` + trailing `if` line on `can't play lands`, and extend the full Rock Jockey oracle dispatch test to assert the gated `CantPlayLand` static on card-name-normalized text.

## CR

CR 611.3a (conditional statics) + CR 305.1 (land plays).

## Class

Parser building-block hardening in the same family as Rock Jockey / Wirecat trailing-gate fixes.

claytonlin1110:
## Summary

Re-scoped per maintainer feedback: trailing `if` / `as long as` gate peel hardening plus **concrete corpus static fixes** with honest fail-closed semantics.

### Infrastructure
- Promote `split_trailing_gate_condition` to `oracle_static/shared.rs` with nom word-boundary scan
- `as if` false-positive guard with UTF-8 char-boundary safety (em-dash panic fix)
- `parse_if_static_condition` routes through last-valid peel with original-case preservation

### Corpus unlocks
| Card / pattern | Fix | Behavior |
|----------------|-----|----------|
| **Shauku-style** | `there's another <type> on the battlefield` → `ObjectCount(type) >= 2` | Gated `CantAttack` gains `condition` |
| **Dual-gated attack/block** | Split `can't attack if <A> and can't block if <B>` before `can't block` dispatch | Emits `CantAttack` + `CantBlock` **only when both gates parse** |
| **The Fallen Apart** | Leg/arm gates not yet modeled | **Fail closed** — no unconditional combat statics (matches `can't play lands` unrecognized-gate policy) |

Rock Jockey / Wirecat / Limited Resources were already fixed on main (#4841 / #4844 / #4433).

Fixes #4876

## Test plan

- [x] `cargo check -p engine`, `cargo fmt --all`
- [x] Rock Jockey, `as if` fixture, em-dash peel, Shauku another-creature gate
- [x] Dual-gate happy path (both gates parse) + Fallen Apart fail-closed regression
- [ ] CI parse-diff / coverage

### DIFF
--- a/crates/engine/src/parser/oracle_nom/primitives.rs
+++ b/crates/engine/src/parser/oracle_nom/primitives.rs
@@ -869,6 +869,86 @@ where
     None
 }
 
+/// Like [`scan_at_word_boundaries`] but returns the **last** successful match,
+/// together with the byte offset where that match began.
+pub fn scan_last_at_word_boundaries_with_offset<'a, O, F>(
+    text: &'a str,
+    mut combinator: F,
+) -> Option<(usize, O, &'a str)>
+where
+    F: FnMut(&'a str) -> nom::IResult<&'a str, O, OracleError<'a>>,
+{
+    let mut remaining = text;
+    let mut offset = 0;
+    let mut last = None;
+    while !remaining.is_empty() {
+        if let Ok((rest, val)) = combinator(remaining) {
+            last = Some((offset, val, rest));
+        }
+        if let Some(rel) = remaining.find(' ') {
+            offset += rel + 1;
+            remaining = remaining[rel + 1..].trim_start();
+        } else {
+            break;
+        }
+    }
+    last
+}
+
+/// Like [`scan_last_at_word_boundaries_with_offset`] but only retains matches
+/// whose offset passes `accept_offset`.
+pub fn scan_last_valid_at_word_boundaries_with_offset<'a, O, F, P>(
+    text: &'a str,
+    mut combinator: F,
+    mut accept_offset: P,
+) -> Option<(usize, O, &'a str)>
+where
+    F: FnMut(&'a str) -> nom::IResult<&'a str, O, OracleError<'a>>,
+    P: FnMut(usize) -> bool,
+{
+    let mut remaining = text;
+    let mut offset = 0;
+    let mut last = None;
+    while !remaining.is_empty() {
+        if let Ok((rest, val)) = combinator(remaining) {
+            if accept_offset(offset) {
+                last = Some((offset, val, rest));
+            }
+        }
+        if let Some(rel) = remaining.find(' ') {
+            offset += rel + 1;
+            remaining = remaining[rel + 1..].trim_start();
+        } else {
+            break;
+        }
+    }
+    last
+}
+
+/// Like [`scan_at_word_boundaries`] but returns the **last** successful match.
+///
+/// Use for terminal riders ("... if <condition>", "... as long as <condition>")
+/// where an earlier `as if` phrase must not steal the gate.
+pub fn scan_last_at_word_boundaries<'a, O, F>(
+    text: &'a str,
+    mut combinator: F,
+) -> Option<(O, &'a str)>
+where
+    F: FnMut(&'a str) -> nom::IResult<&'a str, O, OracleError<'a>>,
+{
+    let mut remaining = text;
+    let mut last = None;
+    while !remaining.is_empty() {
+        if let Ok((rest, val)) = combinator(remaining) {
+            last = Some((val, rest));
+        }
+        remaining = remaining
+            .find(' ')
+            .map_or("", |i| remaining[i + 1..].trim_start());
+    }
+    last
+}
+
 /// Check whether `phrase` appears at any word boundary in `text`.
 ///
 /// More precise than `str::contains()` — matches complete phrases at word
@@ -1185,6 +1265,44 @@ mod tests {
         let result = scan_preceded("the creature enters", |i| {
             tag::<_, _, OracleError<'_>>("dies").parse(i)
         });
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn test_scan_last_at_word_boundaries_picks_terminal_if_gate() {
+        let (_, tail) = scan_last_at_word_boundaries(
+            "you may cast as if they had flash if you control a zombie",
+            |i| tag::<_, _, OracleError<'_>>("if ").parse(i),
+        )
+        .expect("terminal if gate");
+        assert_eq!(tail, "you control a zombie");
+    }
+
+    #[test]
+    fn test_scan_last_at_word_boundaries_with_offset_picks_terminal_if_gate() {
+        let (_, _, tail) = scan_last_at_word_boundaries_with_offset(
+            "you can't play lands as if there were no rule if ten or more lands are on the battlefield",
+            |i| tag::<_, _, OracleError<'_>>("if ").parse(i),
+        )
+        .expect("terminal if gate");
+        assert_eq!(tail, "ten or more lands are on the battlefield");
+    }
+
+    #[test]
+    fn test_scan_last_at_word_boundaries_with_offset_skips_as_if() {
+        let result = scan_last_valid_at_word_boundaries_with_offset(
+            "you can't play lands as if there were no rule",
+            |i| tag::<_, _, OracleError<'_>>("if ").parse(i),
+            |if_offset| {
+                if_offset < 3
+                    || tag::<_, _, OracleError<'_>>("as ")
+                        .parse(
+                            &"you can't play lands as if there were no rule"
+                                [if_offset - 3..if_offset],
+                        )
+                        .is_err()
+            },
+        );
         assert!(result.is_none());
     }
 
--- a/crates/engine/src/parser/oracle_nom/primitives.rs
+++ b/crates/engine/src/parser/oracle_nom/primitives.rs
@@ -1294,12 +1294,12 @@ mod tests {
             "you can't play lands as if there were no rule",
             |i| tag::<_, _, OracleError<'_>>("if ").parse(i),
             |if_offset| {
-                if_offset < 3
+                let Some(start) = if_offset.checked_sub(3) else {
+                    return true;
+                };
+                !"you can't play lands as if there were no rule".is_char_boundary(start)
                     || tag::<_, _, OracleError<'_>>("as ")
-                        .parse(
-                            &"you can't play lands as if there were no rule"
-                                [if_offset - 3..if_offset],
-                        )
+                        .parse(&"you can't play lands as if there were no rule"[start..if_offset])
                         .is_err()
             },
         );
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -451,21 +451,6 @@ pub(crate) fn parse_damage_not_removed_during_cleanup(
     )
 }
 
-/// Split a trailing " as long as <condition>" rider off a static line, returning
-/// the condition text when present (combinator form, no string-method dispatch).
-fn split_trailing_as_long_as(lower: &str) -> Option<&str> {
-    opt(preceded(
-        (
-            take_until::<_, _, OracleError<'_>>(" as long as "),
-            tag(" as long as "),
-        ),
-        rest,
-    ))
-    .parse(lower)
-    .ok()
-    .and_then(|(_, condition)| condition)
-}
-
 /// CR 509.1b: "Creatures with power <comparison> <quantity> can't
 /// block this creature." — a can't-be-blocked-by restriction whose blocker
 /// filter gates on a power threshold that may be DYNAMIC (Kraken of the Straits:
@@ -524,26 +509,6 @@ fn filter_prop_has_power_comparison(prop: &FilterProp) -> bool {
         FilterProp::AnyOf { props } => props.iter().any(filter_prop_has_power_comparison),
         _ => false,
     }
-}
-
-/// CR 611.3a: A static restriction may carry a trailing gate introduced by
-/// either `" as long as <condition>"` (continuous) or `" if <condition>"` (state
-/// gate) — e.g. Rock Jockey: "You can't play lands if this creature was cast
-/// this turn." Returns the condition text for `parse_static_condition`. The
-/// `as long as` form is tried first so a card carrying both keywords anchors on
-/// the continuous form; a bare `if` gate is the fallback. As with
-/// `split_trailing_as_long_as`, an unrecognized condition downstream leaves the
-/// line unsupported rather than enforcing the restriction unconditionally.
-fn split_trailing_gate_condition(lower: &str) -> Option<&str> {
-    split_trailing_as_long_as(lower).or_else(|| {
-        opt(preceded(
-            (take_until::<_, _, OracleError<'_>>(" if "), tag(" if ")),
-            rest,
-        ))
-        .parse(lower)
-        .ok()
-        .and_then(|(_, condition)| condition)
-    })
 }
 
 pub(crate) fn parse_static_line_inner(
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -1948,14 +1948,21 @@ fn split_trailing_if_condition_tp<'a>(tp: &'a TextPair<'a>) -> Option<&'a str> {
     Some(tp.original.get(start..)?.trim_start())
 }
 
+/// CR 508.1c + CR 509.1b: Split the gated combat tail `<A> and can't block if <B>`
+/// after the leading `"can't attack if "` marker has been consumed.
+fn parse_dual_gated_combat_condition_tails(input: &str) -> OracleResult<'_, (&str, &str)> {
+    let (input, attack_cond) = take_until(" and can't block if ").parse(input)?;
+    let (input, _) = tag(" and can't block if ").parse(input)?;
+    let (input, block_cond) = terminated(rest, opt(tag("."))).parse(input)?;
+    Ok((input, (attack_cond, block_cond)))
+}
+
 /// CR 508.1c + CR 509.1b: Split a compound "~ can't attack if <A> and can't block
 /// if <B>" static into two gated restrictions (The Fallen Apart).
 fn parse_dual_gated_cant_attack_block(input: &str) -> OracleResult<'_, (&str, &str, &str)> {
     let (input, subject) = take_until("can't attack if ").parse(input)?;
     let (input, _) = tag("can't attack if ").parse(input)?;
-    let (input, attack_cond) = take_until(" and can't block if ").parse(input)?;
-    let (input, _) = tag(" and can't block if ").parse(input)?;
-    let (input, block_cond) = terminated(rest, opt(tag("."))).parse(input)?;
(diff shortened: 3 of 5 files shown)


matthewevans (changes requested):
**[MED]** The new trailing-gate parser bypasses the mandated nom combinator path. Evidence: `crates/engine/src/parser/oracle_static/shared.rs` uses `rfind` plus byte slicing for grammar boundary detection. Why it matters: parser dispatch needs to stay in the composable nom path so this shared restriction parser does not grow ad hoc string parsing. Suggested fix: express the last-valid `if` / `as long as` peel as a nom/scan combinator and return the condition slice through one shared path.

**[LOW]** The combat if-gate helper cites the blocker requirement rule for restriction behavior. Evidence: the modified helper/comment still points at `CR 509.1c` for block restrictions; `CR 509.1b` is the blocker restriction rule, while `CR 509.1c` is requirements. Suggested fix: cite `CR 509.1b` for block restrictions and `CR 508.1c` for attack restrictions.

Also: the parse-diff sticky (`<!-- coverage-parse-diff -->`) is not present yet for this parser PR, so the card-level blast radius is still missing review evidence.

matthewevans (changes requested):
**[MED]** The current parse-diff does not match this trailing-if hardening PR’s scope. Evidence: the `<!-- coverage-parse-diff -->` sticky for head `4316161b8044bfb809b64cbba2105319288b3750` reports 54 changed cards and 12 signatures, including broad `Sacrifice` target changes plus Cloudstone Curio and Role Reversal, none of which are explained by a shared trailing `if` gate scanner for static restrictions. Why it matters: even if the prior nom-combinator blocker is fixed, we cannot approve a parser PR whose generated card-level blast radius is unrelated to the stated Rock Jockey/Wirecat-style trailing condition class. Suggested fix: rebase/split the branch so the parse diff is limited to the trailing-if condition behavior this PR intends to change.

matthewevans (changes requested):
**[HIGH]** The current parse-diff still does not match this trailing-if hardening PR’s scope. Evidence: the `<!-- coverage-parse-diff -->` sticky for the current sweep still reports 54 cards / 12 signatures, including broad Sacrifice target changes and unrelated cards such as Cloudstone Curio and Role Reversal, while this PR’s stated scope is the shared trailing-if gate parser. Why it matters: approving this would accept an unexplained card-level parser blast radius unrelated to the intended restriction-gate change. Suggested fix: regenerate the current-head parse-diff and either explain a tightly scoped result or rebase/split until the diff matches the trailing-if change.

claytonlin1110:
## Parse-diff scope fix

The 54-card / 12-signature sticky (Sacrifice targets, Cloudstone Curio, Role Reversal) was **baseline contamination from a stale branch**, not parser blast radius from the trailing-if gate work.

**Evidence:** that sticky used baseline `main \`dfbd2e3009f9\`` while the PR head at the time was missing intervening main parser commits (`56391ceb9` Sacrifice another-filters, `c4b8367e3` Role Reversal, etc.). The diff was effectively *main − stale PR*, not *this PR’s engine delta*.

**Fix:** rebased `clayton/trailing-if-gate-hardening` onto current `upstream/main` (`572d533e8`). Branch is now exactly 3 commits / 5 parser files vs main; merge-base equals main tip.

Fresh CI on head `447370433` should produce a tightly scoped parse-diff once the baseline for the merge parent publishes (prior run showed *Baseline pending* for `19c5df049`, which is expected until R2 catches up).

matthewevans (changes requested):
**[MED]** This now looks like speculative parser hardening rather than a PR that clears the maintainer value bar. Evidence: the current `<!-- coverage-parse-diff -->` sticky for head `4aee701c921c6583683281a46f2c357f7fa74a2e` reports “No card-parse changes detected,” while the linked issue was opened by the PR author and describes a hypothetical future reuse case rather than a current corpus card misparse. Why it matters: this PR adds new shared parser scanning surface and tests but does not demonstrate a real card/class unlock or current production parser bug, so the blast radius/value tradeoff is not strong enough to merge as-is. Suggested fix: either tie the change to a real current card/parser failure with a scoped parse-diff or close/split this down to whatever existing-card regression actually needs the shared helper.

claytonlin1110:
Reopened — PR description updated to address the parse-diff / value-bar review directly.

**Headline:** this is intentionally a **zero card-parse-change** consolidation PR, not a corpus unlock. #4841 already covers Rock Jockey; the `as if` + terminal `if` regression is a class fixture for the shared peel.

Rebased onto latest `main` (`978cde04c`); branch head is now `e3db861d6`. CI should refresh the parse-diff sticky.

matthewevans (changes requested):
**[MED]** Thanks for clarifying the intended scope, but this still does not clear the maintainer value bar for this sweep. Evidence: the PR now explicitly says it is “not claiming a card unlock,” and the current parse-diff evidence remains “No card-parse changes detected,” while the linked issue is author-created and describes a hypothetical future reuse/regression class rather than a current card/parser failure. Why it matters: this adds shared parser surface and synthetic fixtures without fixing a live misparse or unlocking a real card class, which is the wrong tradeoff for an elevated-scrutiny contributor PR. Suggested fix: close this consolidation PR or re-scope it around a concrete current corpus failure with a current parse-diff proving the behavior change.

claytonlin1110:
Re-scoped and rebased onto latest main (51f17859). Added corpus-facing fixes: dual-gated attack/block split (The Fallen Apart) and \	here's another <type> on the battlefield\ → ObjectCount >= 2 (Shauku-style gates). Peel hardening + as-if/em-dash guards retained. Ready for CI parse-diff.

matthewevans (changes requested):
**[HIGH] The dual-gated attack/block split drops unmodeled `if` gates and emits unconditional restrictions.**

Evidence: `try_parse_dual_gated_cant_attack_and_cant_block` creates `CantAttack` / `CantBlock` definitions first, then only attaches a condition when `parse_static_condition(...)` succeeds; if the condition is unrecognized, the function still returns the static definitions with `condition = None` (`crates/engine/src/parser/oracle_static/shared.rs:1965` and `:1971`). The new The Fallen Apart test encodes that unsafe behavior by asserting both conditions are `None` for `"it has no legs"` / `"it has no arms"` while still requiring the statics to be emitted (`crates/engine/src/parser/oracle_static/tests.rs:493`).

Why it matters: "can't attack if <A> and can't block if <B>" must not become unconditional `can't attack` / `can't block` when `<A>` or `<B>` is not modeled. That makes the card parse as supported while enforcing stricter combat restrictions than Oracle text.

Suggested fix: fail closed for this compound split unless both gates parse to conditions, or preserve unrecognized gate conditions in the AST so coverage stays honest and runtime does not enforce them unconditionally.

claytonlin1110:
Addressed HIGH review: dual-gated attack/block split now requires both \if\ gates to parse (same fail-closed policy as unrecognized \can't play lands\ gates). Fallen Apart stays unsupported until leg/arm conditions are modeled; added happy-path test with both gates recognized (Shauku-style + Wirecat-style conditions).

matthewevans (changes requested):
**[MED]** The current parse-diff still contains an unexplained change outside this PR's static-restriction scope. Evidence: the current `<!-- coverage-parse-diff -->` sticky for head `60dd1ddee70f421f405638bd9fdd83ed947b58dd` reports Multani's Presence changing from the raw `Whenever a spell you've cast is countered` trigger signature to typed `Countered (active in=battlefield, watches=you)`, while this PR's changed files are limited to `oracle_nom/primitives.rs`, `oracle_static/*`, and parser tests. Why it matters: a trailing-if/static-restriction hardening PR should not silently alter unrelated trigger coverage; approving it would accept a card-level parser blast radius that is not explained by the diff. Suggested fix: either explain and test why this trigger signature change is actually caused by the static-gate work, or rebase/regenerate/split until the parse-diff is limited to the static restriction class this PR is meant to cover.

matthewevans:
Additional blocker from follow-up review: the new dual-gated `can't attack if <A> and can't block if <B>` splitter appears to preempt attached-subject parsing and hardcodes `SelfRef`. The splitter runs before the attached-subject owner path in `oracle_static/shared.rs`, then emits both restrictions with `affected(TargetFilter::SelfRef)`. That means a valid attached-subject form such as `Enchanted creature can't attack if <A> and can't block if <B>` would apply the restrictions to the Aura/source instead of the enchanted creature. Please either scope this splitter to self-reference subjects only, or parse/thread the subject filter through it before emitting the two restrictions.

claytonlin1110:
Addressed MED parse-diff review: Multani's Presence trigger signature change was **not** caused by this PR's static-gate work — the branch was stale behind main and was effectively reverting #4903 (\ix(parser): trigger on your own countered spell (Multani's Presence)\). Rebased onto current \upstream/main\ (\e9cb2001\); diff is now limited to the 5 oracle_nom/oracle_static files. CI parse-diff should no longer show unrelated trigger coverage.

matthewevans (changes requested):
**[MED]** The dual-gated attack/block splitter still preempts attached-subject restrictions and hardcodes the affected object to `SelfRef`. Evidence: `parse_static_line_multi_inner` runs `try_parse_dual_gated_cant_attack_and_cant_block` before the normal attached-subject/static dispatch, and `parse_dual_gated_cant_attack_block` uses `take_until("can't attack if ")`, so it will match text such as `Enchanted creature can't attack if <A> and can't block if <B>`. The emitted definitions then use `.affected(TargetFilter::SelfRef)` for both restrictions. Why it matters: for Aura/equipment-style statics, `SelfRef` is the enchantment/equipment source, not the enchanted/equipped creature that the restriction text names. Suggested fix: either parse and preserve the leading subject in this splitter, or only let this special splitter handle explicit source/self-subject forms and leave attached-subject forms to the established subject parser.

claytonlin1110:
Addressed MED attached-subject review: dual-gated attack/block split now runs only when the leading subject is a self-ref (~ / this creature / …) and \ttached_subject_filter\ is absent. Enchanted/equipped forms defer to the established subject parser instead of emitting \SelfRef\ combat statics.

matthewevans (changes requested):
**[MED]** The "there's another <type> on the battlefield" condition drops the source-relative `Another` building block. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:2328` parses the clause, but `crates/engine/src/parser/oracle_static/shared.rs:2343` counts the plain type filter and compares it to `>= 2` instead of representing "another" as source-relative exclusion. Why it matters: `ObjectCount(creature) >= 2` only approximates the Oracle text and loses the engine's existing `FilterProp::Another` semantics. Suggested fix: parse the counted filter with `FilterProp::Another` and compare against `>= 1`, or leave the condition unsupported until the source/recipient binding is represented correctly.

**[MED]** The parse-diff sticky still includes unrelated Riku of Many Paths parser blast radius. Evidence: the current `<!-- coverage-parse-diff -->` body for this head reports `trigger/SpellCast` field `watches: ∅ -> modal spell card` for Riku of Many Paths, while this PR is scoped to trailing-if/static-restriction parsing. Why it matters: the review cannot treat unrelated card-output changes as safe without explanation and tests. Suggested fix: rebase/regenerate if stale, or split/explain and test the Riku change.

claytonlin1110:
Addressed MED reviews: (1) \	here's another <type> on the battlefield\ now uses \FilterProp::Another\ + \ObjectCount >= 1\ instead of plain type count >= 2. (2) Rebased onto current \upstream/main\ (\dad23ee98\) — Riku modal-spell parse-diff was stale-branch noise from missing #4907, not this PR.

matthewevans:
Holding approval on current head `612af4a39887790131e1daabfe1720ca827fcd3a`: the prior `Another` / attached-subject static-condition findings appear addressed in this pass, but the current parse-diff sticky is still not clean enough for approval.

The sticky reports an unrelated Riku of Many Paths `trigger/SpellCast` `watches: ∅ -> modal spell card` change, while this branch is scoped to static parser/static tests. Please regenerate/rebase the parse-diff evidence so it only shows the intended Shauku / The Fallen Apart / static-structure changes, or explain why the Riku trigger signature belongs in this PR.

matthewevans (commented):
Holding approval on current head `612af4a39887790131e1daabfe1720ca827fcd3a`: the prior `Another` / attached-subject static-condition findings appear addressed in this pass, but the current parse-diff sticky is still not clean enough for approval.

The sticky reports an unrelated Riku of Many Paths `trigger/SpellCast` `watches: ∅ -> modal spell card` change, while this branch is scoped to static parser/static tests. Please regenerate/rebase the parse-diff evidence so it only shows the intended Shauku / The Fallen Apart / static-structure changes, or explain why the Riku trigger signature belongs in this PR.

matthewevans (changes requested):
One current-head blocker remains.

**[LOW]** The new `another` gate has an incorrect CR citation. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:2346` cites `CR 613.7` for “another <type> excludes the ability source,” but the local Comprehensive Rules text has `613.7` as timestamp ordering for continuous effects. That citation is unrelated to this parser behavior.

Please remove the CR citation or replace it with a verified relevant one. The prior `Another` / attached-subject behavior and parse-diff scope otherwise look addressed on this pass.

matthewevans (changes requested):
**[MED]** The dual-gated attached-subject splitter still preempts attached grant+restriction lines and drops the leading grant. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:1945`, `crates/engine/src/parser/oracle_static/shared.rs:1966`, and `crates/engine/src/parser/oracle_static/shared.rs:1996`. Why it matters: a valid class like `Enchanted creature gets +2/+2 and can't attack if <A> and can't block if <B>` can be accepted by this early splitter, but it returns only `CantAttack`/`CantBlock`, so the pump or keyword grant never reaches the established splitters. Suggested fix: require the captured subject to be exactly the attached/self subject before splitting, or thread the leading predicate through the existing grant splitter and append the two gated restrictions.

matthewevans (changes requested):
**[LOW]** The grant-plus-dual-restriction regression test over-constrains the future-correct shape. Evidence: `crates/engine/src/parser/oracle_static/tests.rs:626`. Why it matters: a correct parser that preserves the grant and also emits both gated combat restrictions would have two combat statics and fail this test, so the test locks in partial support instead of only guarding against the prior `only two combat statics, grant dropped` regression. Suggested fix: assert the bad shape directly, for example `!(defs.len() == 2 && defs.iter().all(|d| matches!(d.mode, StaticMode::CantAttack | StaticMode::CantBlock)))`, while keeping the continuous-grant assertion.

matthewevans (changes requested):
**[MED]** Grant-plus dual-gated attached-subject lines still drop both combat restrictions. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:1995`, `crates/engine/src/parser/oracle_static/evasion.rs:961`, `crates/engine/src/parser/oracle_static/grammar.rs:804`, and `crates/engine/src/parser/oracle_static/tests.rs:621`. Why it matters: `Enchanted creature gets +2/+2 and can't attack if <A> and can't block if <B>` now preserves the pump, but the gated `CantAttack`/`CantBlock` clauses are still not emitted, so the enchanted creature can attack/block when Oracle says it cannot. Suggested fix: thread the leading attached-subject grant through the dual-gate splitter, or extend the attached-subject grant splitter to append both gated combat statics, and require those statics in the regression when both gates parse.

matthewevans:
Holding approval on current head `e5302cf7144003330b7c8ec37414be21a8eaa9ef`: the prior grant-plus-dual-restriction code blocker appears addressed in this patch. The regression now requires the pump grant plus both gated `CantAttack` / `CantBlock` statics, and the implementation appends those gated combat statics from the grant's affected target.

I am not approving/enabling merge-when-ready yet because this is parser-surface and the required parse-diff sticky is still the older evidence from baseline `main dad23ee985ed` rather than current-head evidence for `e5302cf...`; the branch also still reports behind. Please refresh the branch/CI so the parse-diff sticky reflects the current head.

matthewevans (changes requested):
**[HIGH]** Current head cannot compile, so card-data cannot produce current parse-diff evidence. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:2486` and `crates/engine/src/parser/oracle_static/shared.rs:2492` define private helpers that the current CI reports as unused under `-D warnings`. Why it matters: the branch update cannot get us a reviewable current-head parse-diff while `oracle-gen`/Rust checks fail before coverage generation. Suggested fix: remove the duplicate unused helpers, or route the production parser through one helper instead of leaving it dead.

**[MED]** The grant-plus-dual-gate splitter still drops the intended restrictions. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:2043` consumes `" and can't attack if "`, then `crates/engine/src/parser/oracle_static/shared.rs:2053` passes the remaining tail into `parse_dual_gated_cant_attack_block`, whose first parse step still expects a `"can't attack if "` marker at `crates/engine/src/parser/oracle_static/shared.rs:1953`. Why it matters: after the compile blocker is fixed, the new regression at `crates/engine/src/parser/oracle_static/tests.rs:751` should still fail because the dual gated combat statics are never emitted. Suggested fix: parse the attack/block condition tails directly after the scan, or reconstruct the expected prefix before calling the existing helper.

matthewevans:
Updated the branch to current `main` to refresh the expired parse-diff baseline. New head is `98a011cf911c40f325061273683008f36ed76159`.

Holding review until CI posts a current `<!-- coverage-parse-diff -->` sticky for this new head; the previous sticky was `Baseline pending`, so the card-level parser evidence needed for approval is still unavailable right now.

matthewevans:
Holding review on current head `daee732869994d23d5100ed8c7fac4c21a16daac`: the latest `<!-- coverage-parse-diff -->` sticky was updated before this head was pushed, and card-data is running again. Because this PR changes parser source, I need the current-head parse-diff before I can re-review or approve.

matthewevans:
Holding review on current head `5f1273d396c3109d7a2f014fc52c07f4be1c5545`: the current parser head was pushed after the latest coverage-parse-diff sticky, and Card data is still running for this head. Because this PR changes parser source, so I need current-head parse-diff evidence before re-review or approval.

I’m not approving or enqueueing from this pass. Once the current-head evidence lands, this should resurface for review.

matthewevans (approved):
Current head `5f1273d396c3109d7a2f014fc52c07f4be1c5545` is approve-ready.

The prior grant/static-condition blockers are addressed with the fail-closed dual-gate handling, the parser changes stay on the existing nom/static restriction seam, and the regression tests cover the relevant trailing-if combinations. The current parse-diff is fresh and scoped to The Fallen Apart only.

PR #4879, opened 2026-07-02
fix(parser): source-scoped "if ~ attacked or blocked this turn" intervening-if

real-venus:
## Summary

The intervening-if *"…, if ~ attacked or blocked this turn, …"* (CR 603.4) was silently dropped on triggered abilities — the trigger's `condition` was left `None`, so the ability resolved **unconditionally**. **Inferno Hellion** shuffled itself into its owner's library at *every* end step, even on a turn it neither attacked nor blocked.

## Root cause

The `try_extract_simple_condition` dispatch table in `oracle_trigger.rs` handles the attacked-only sibling (`"if ~ attacked this turn"`, added in #4863) but had **no arm for the attacked-**or**-blocked form**.

## Fix

Add one dispatch arm mapping `"if ~ attacked or blocked this turn"` to:

```rust
TriggerCondition::SourceMatchesFilter {
    filter: TargetFilter::Typed(
        TypedFilter::creature().properties(vec![FilterProp::AttackedOrBlockedThisTurn]),
    ),
}
```

Reuses the existing, already-evaluated `FilterProp::AttackedOrBlockedThisTurn` (checked against `state.creatures_attacked_this_turn` / `creatures_blocked_this_turn` in `game/filter.rs`). **No new `TriggerCondition` variant, no engine/resolver changes** — the direct sibling of #4863.

**Scope:** turn-scoped (`"this turn"`) only. The `"this combat"` form (Clockwork cycle, Kjeldoran Home Guard) needs combat-scoped tracking the engine doesn't keep; mapping it to the turn set would be CR-incorrect on multi-combat turns, so it's intentionally excluded.

## Testing

`intervening_if_source_attacked_or_blocked_this_turn_populates_condition` asserts Inferno Hellion gets the gated condition and the effect still parses. Passes:

```
test parser::oracle_trigger::tests::intervening_if_source_attacked_or_blocked_this_turn_populates_condition ... ok
```

CR 508.1 + CR 509.1 + CR 603.4.

### DIFF
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -3883,6 +3883,25 @@ fn extract_if_condition(text: &str) -> (String, Option<TriggerCondition>) {
                     ),
                 },
             ),
+            // CR 508.1 + CR 509.1 + CR 603.4: source-scoped "if ~ attacked or
+            // blocked this turn" — the sibling of the attacked-only arm above,
+            // gating on the source creature having attacked OR blocked this turn
+            // (Inferno Hellion). Reuses the existing, already-evaluated
+            // `FilterProp::AttackedOrBlockedThisTurn` (checked against
+            // `state.creatures_attacked_this_turn` / `creatures_blocked_this_turn`)
+            // via `SourceMatchesFilter`, so no new `TriggerCondition` variant is
+            // needed. The turn-scoped "this turn" form only; the "this combat"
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Reviewed current head `b55759c9735cbdae0c456a48d92846c911b48838`. This clears the parser bar: the change is scoped to the missing source-scoped `attacked or blocked this turn` intervening-if condition, CI is green, and the parse-diff sticky shows the intended Inferno Hellion condition change only.

PR #4881, opened 2026-07-02
Reclassify stale unless-payment backlog entry

ntindle:
## Summary
- replace the stale root-33 unless-payment bucket with the still-real static pay/action-to-ignore-effect gap
- remove the five listed cards whose current parser output already models the alleged gap or is no longer an unless-payment card
- keep Lost in Thought tracked under the reclassified root cause

## Current parser evidence
- Lava Blister has unless_pay with ParentTargetController payer and deal-6-damage effect cost
- Lethargy Trap has an alternative {U} casting option gated by three or more attacking creatures, not an unless-payment clause
- Lim-Dul's Hex has trigger unless_pay with TriggeringPlayer payer and {B} or {3} cost
- Read the Runes has the repeated discard branch with unless_pay sacrifice-permanent cost
- Trapped in the Tower has the creature-without-flying enchant filter and static lock effects, and is not an unless-payment card
- Lost in Thought still lacks the ignore-this-effect-until-EOT exception and remains tracked

## Verification
- targeted parser output generated with oracle-gen for all six original cards
- python backlog count check: roots=33 appearances=4853 distinct=4819
- git diff --check HEAD~1..HEAD

### DIFF
--- a/docs/parser-misparse-backlog.md
+++ b/docs/parser-misparse-backlog.md
@@ -3,8 +3,8 @@
 Consolidated from 50 per-batch clustering passes over the whole card database. Synonymous per-batch clusters were merged into canonical root causes, their card lists unioned and deduped, and ranked by total card appearances (largest first).
 
 - **Canonical root causes:** 33
-- **Distinct cards implicated:** 4824
-- **Total card appearances across root causes:** 4858 (a card may appear under more than one root cause when it exhibits multiple distinct misparses)
+- **Distinct cards implicated:** 4819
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head `ee8df04a07a1c4d3fc4627890ecbb1e6186339fd`. This is a docs-only backlog correction: it removes stale unless-payment entries that the PR body accounts for and narrows root cause 33 to the remaining static pay/action-to-ignore-effect gap. No engine behavior changes are included.

PR #4885, opened 2026-07-02
fix(parser): parse parenthetical gate on cant-cast prohibitions (Alhammarret)

### LINKED ISSUE #4883: fix(parser): parse parenthetical gate on cant-cast prohibitions (Alhammarret)
## Problem

Casting prohibitions like Alhammarret, High Arbiter:

> Your opponents can't cast spells with the chosen name **(as long as this creature is on the battlefield).**

`parse_cant_cast_type_spells` in `restriction.rs` currently **strips** trailing parenthetical riders via `rfind(" (")` instead of parsing them as `StaticCondition` gates. The prohibition is therefore enforced **unconditionally** — opponents cannot cast spells with the chosen name even after Alhammarret leaves the battlefield.

This is the same class of bug fixed for Rock Jockey in #4841 (trailing `if` / `as long as` gate swallowed → unconditional restriction), but on the **parenthetical** gate form used by cant-cast statics.

## Proposed fix

- Add a composable `split_parenthetical_gate_condition` helper (sibling of `split_trailing_gate_condition`) that accepts `(as long as <condition>)` and `(if <condition>)`.
- In `parse_cant_cast_type_spells`, peel the parenthetical gate **before** pattern dispatch; attach via `parse_static_condition`. If a parenthetical gate is present but its condition is unrecognized, leave the line unsupported (`None`) rather than enforcing unconditionally — parity with Limited Resources / Rock Jockey guards.
- Regression test on Alhammarret full Oracle text asserting `CantBeCast` + `SourceInZone { Battlefield }` gate.

## CR

CR 101.2 (who can cast) + CR 611.3a (conditional statics) + CR 113.6b (source zone).

## Related

- #4841 — trailing `if` gate on `CantPlayLand` (Rock Jockey)

claytonlin1110:
Fixes #4883

## Summary

- **`static_helpers.rs`** — add `split_parenthetical_gate_condition` and `attach_parsed_static_gate`, composable siblings of Rock Jockey's `split_trailing_gate_condition` (#4841). Accept trailing `(as long as <condition>)` and `(if <condition>)` parenthetical gates.
- **`restriction.rs`** — `parse_cant_cast_type_spells` now peels parenthetical gates before pattern dispatch and attaches them via `parse_static_condition`. Unrecognized gates leave the line unsupported (`None`) instead of enforcing an unconditional prohibition.
- **`tests.rs`** — Alhammarret regression asserting `CantBeCast` + `SourceInZone { Battlefield }`, plus `if`-gate parity, unrecognized-gate guard, and full-card dispatch swallow check.

## Problem

Alhammarret, High Arbiter: *"Your opponents can't cast spells with the chosen name (as long as this creature is on the battlefield)."*

The parenthetical gate was **stripped** by `rfind(" (")`, collapsing the restriction to an **unconditional** `CantBeCast` — opponents could never cast spells with the chosen name even after Alhammarret left the battlefield.

Same bug class as #4841 (Rock Jockey trailing `if` gate swallowed).

## CR

CR 101.2 (who can cast) + CR 611.3a (conditional statics) + CR 113.6b (source zone).

## Test plan

- [x] `cargo fmt --all`
- [ ] `cargo test -p engine cant_cast_spells_with_chosen_name_parenthetical alhammarret_cant_cast cant_cast_spells_unrecognized_parenthetical`
- [ ] `cargo clippy -p engine -- -D warnings`

### DIFF
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -789,14 +789,30 @@ fn try_parse_graveyard_keyword_static_with_continuation(line: &str) -> Option<St
 /// and then delegating to `parse_static_line_multi` so compound forms
 /// (e.g., cross-mode conjunctions) emit all their constituent statics
 /// rather than silently dropping the extras.
-fn parse_static_line_with_graveyard_keyword_continuation(line: &str) -> Vec<StaticDefinition> {
-    if let Some(def) = try_parse_graveyard_keyword_static_with_continuation(line) {
-        return vec![def];
+///
+/// When `raw_line_for_cant_cast_gates` is set (oracle dispatch only), cant-cast
+/// gate parentheticals stripped by `strip_reminder_text` are re-applied without
+/// passing raw reminder parentheticals through the general static parser.
+fn parse_static_line_with_graveyard_keyword_continuation(
+    line: &str,
+    raw_line_for_cant_cast_gates: Option<&str>,
+    card_name_for_cant_cast_gates: Option<&str>,
+) -> Vec<StaticDefinition> {
+    let mut defs = if let Some(def) = try_parse_graveyard_keyword_static_with_continuation(line) {
+        vec![def]
+    } else if let Some(def) = try_parse_graveyard_keyword_grant_static(line) {
+        vec![def]
+    } else {
+        parse_static_line_multi(line)
+    };
+    if let (Some(raw), Some(card_name)) =
+        (raw_line_for_cant_cast_gates, card_name_for_cant_cast_gates)
+    {
+        defs = crate::parser::oracle_static::apply_raw_parenthetical_cant_cast_gate(
+            defs, raw, card_name,
+        );
     }
-    if let Some(def) = try_parse_graveyard_keyword_grant_static(line) {
-        return vec![def];
-    }
-    parse_static_line_multi(line)
+    defs
 }
 
 /// CR 614.6 + CR 701.26b: A single `<subject> can't <P1> and can't <P2>`
@@ -823,8 +839,8 @@ fn parse_static_replacement_compound(
     let left = format!("{subject} can't {p1}");
     let right = format!("{subject} can't {p2}");
 
-    let left_statics = parse_static_line_with_graveyard_keyword_continuation(&left);
-    let right_statics = parse_static_line_with_graveyard_keyword_continuation(&right);
+    let left_statics = parse_static_line_with_graveyard_keyword_continuation(&left, None, None);
+    let right_statics = parse_static_line_with_graveyard_keyword_continuation(&right, None, None);
     let left_repl = parse_replacement_line(&left, card_name);
     let right_repl = parse_replacement_line(&right, card_name);
 
@@ -1209,6 +1225,8 @@ where
             .statics
             .extend(parse_static_line_with_graveyard_keyword_continuation(
                 modeled_sentence,
+                None,
+                None,
             ));
         result.abilities.push(make_unimplemented(unmodeled_tail));
         return true;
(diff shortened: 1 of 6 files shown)


matthewevans (changes requested):
**[HIGH]** Current-head parser evidence is incomplete because the parser gate failed and no parse-diff sticky exists. Evidence: `gh pr checks 4885` reports `Rust lint (fmt, clippy, parser gate)` failed on run `28573982199`, and the inspect packet has `parse_diff.present=false` for head `b9b7bf95f62a100daba1011fabe3f135fc19a820`. Why it matters: this PR changes parser source, so the current card-level parse-diff is required evidence before approval. Suggested fix: fix the Rust/parser gate failure and let CI publish the current-head coverage parse-diff.

**[MED]** The new parenthetical parser helper bypasses the mandated combinator path. Evidence: `crates/engine/src/parser/oracle_static/static_helpers.rs:1245` implements `split_parenthetical_gate_condition` with `trim_end_matches`, `rsplit_once`, and `strip_suffix` for grammar boundary parsing. Why it matters: new parser code is expected to compose nom axes so parenthetical gates stay reusable and precise, instead of adding another ad hoc string parser. Suggested fix: express the parenthetical gate as a small nom combinator over body + parenthetical introducer + condition, reusing `parse_static_condition` for the condition payload.

**[MED]** Non-gate parentheticals now become parse-failing residue for cant-cast lines. Evidence: `crates/engine/src/parser/oracle_static/restriction.rs:977` only strips the parenthetical when `split_parenthetical_gate_condition` recognizes `as long as` or `if`; otherwise it falls back to the full `after_cant_cast` text with the parenthetical still attached. Gemini raised the same consumer issue on this head. Why it matters: the previous parser stripped trailing parentheticals before matching the prohibition body; this change can regress benign reminder/non-gate parentheticals into unsupported lines while trying to fix Alhammarret. Suggested fix: split the body from a trailing parenthetical independently, then attach a condition only when the parenthetical is a recognized gate; benign/unrecognized non-gate parentheticals should not poison the body parse unless they are rules-bearing gates that must remain unsupported.

matthewevans (changes requested):
**[HIGH]** Current-head parse-diff shows broad unrelated parser regression outside the Alhammarret gate fix. Evidence: the `<!-- coverage-parse-diff -->` sticky now reports the intended Alhammarret `CantBeCast(opponents)` condition change, but also removes existing `Continuous` graveyard/static keyword grants for Varolz, The Cave of Skulls, Underworld Breach, Stormcage Containment Facility, The Master of Keys, Wire Surgeons, Jurassic Park, and Fumiko while adding generic `ability/static_structure` for several of those cards. Why it matters: preserving raw parentheticals for cant-cast gates must not make unrelated static abilities fall out of their established parsers or coverage becomes dishonest across multiple card classes. Suggested fix: narrow the raw-parenthetical path to the cant-cast parser seam, or strip/route benign parentheticals so the existing static keyword-grant parsers keep their prior output; repush with a parse-diff scoped to Alhammarret and intentional sibling changes only.

matthewevans (changes requested):
**[HIGH]** Current-head parse-diff still shows broad unrelated parser changes outside the Alhammarret parenthetical gate fix. Evidence: the updated `<!-- coverage-parse-diff -->` sticky for head `8b9c21de9aec5b8323dd6d443400faa3f280096d` reports 13 cards / 26 signatures, including unrelated Old Fogey keyword/trigger/replacement changes, Harbor Serpent condition changes, multiple protection-grant changes, Zask/Yawgmoth's Agenda graveyard-cast-affects changes, and an Explosive Revelation PutAtLibraryPosition removal. Why it matters: this PR is not safe to merge as a narrow cant-cast gate fix while it changes/regresses parser output for unrelated card classes. Suggested fix: rebase/update against current main, reduce the diff to the Alhammarret-class gate behavior, and wait for a parse-diff sticky whose card changes match the claimed scope.

matthewevans (approved):
Approved current head. I re-checked the current parse-diff sticky against `b7a57bcc7937bba5392daae9075b42b0d711bea7`: it is scoped to Alhammarret's `CantBeCast(opponents)` condition gaining `source is in battlefield`, with CI green and the prior broad parser-regression blocker gone. The parenthetical split is now a nom-based parser helper, benign/unrecognized parentheticals are covered, and the full-card dispatch test would fail if the gate were stripped into an unconditional prohibition.

PR #4888, opened 2026-07-02
Support "Skip the [step] step of that turn" (Savor the Moment)

e11734937-beep:
## Summary

Two extra-turn cards fell through to `Effect::Unimplemented` on their **second** sentence:

- **Savor the Moment** (SHM) — *"Take an extra turn after this one. Skip the untap step of that turn."*
- **Time Bends to My Will** (scheme) — *"When you set this scheme in motion, take an extra turn after this one. Skip the untap step of that turn."*

The extra-turn clause already parsed; only *"Skip the untap step of that turn."* was unimplemented. This unlocks both cards fully.

## Change

Add a `"skip the <step> of that turn"` arm to `try_parse_skip_next_step` (`crates/engine/src/parser/oracle_effect/mod.rs`), the current-turn-anaphor sibling of the existing `"skip your next <step> step"` arm. It reuses the shared `parse_skip_step_name` vocabulary and lowers to the existing `Effect::SkipNextStep { target: Controller, step, count: 1, scope: NextOccurrence }` — no new effect, scope, or resolver.

Per CR 614.10a, a "next"/first-unskipped-occurrence skip is exactly right here: "that turn" is the extra turn just created, which is the controller's next turn, so its untap step is the next occurrence of that step for the controller. The arm is whole-clause anchored (the remainder after "of that turn" must be empty modulo a trailing period), so compound clauses fall through to their own recognizers. Corpus grep confirms *"skip the untap step of that turn"* is the only occurrence of this shape.

## Tests

- `skip_the_untap_step_of_that_turn_parses_as_step_skip` — the isolated clause lowers to `SkipNextStep`.
- `savor_the_moment_full_card_has_no_unimplemented` — full-card reach guard: the extra-turn effect parses, the skip chains, and there is zero `Unimplemented`.

Fails-before (clause → `Unimplemented`) / passes-after confirmed. Green: rustfmt, parser-combinator (Rule Zero) gate, `clippy -D warnings`, full engine suite (14518 passed / 0 failed).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -7736,6 +7736,33 @@ fn try_parse_skip_next_step(tp: TextPair, ctx: &ParseContext) -> Option<ParsedEf
         }
     }
 
+    // CR 614.10 + CR 614.10a + CR 502.3: "Skip the [step] step of that turn" —
+    // the current-turn-anaphor sibling of the "skip your next [step] step" form
+    // above. Printed on extra-turn spells that take an extra turn, then skip a
+    // step of it (Savor the Moment, Time Bends to My Will: "Take an extra turn
+    // after this one. Skip the untap step of that turn."). "That turn" is the
+    // extra turn just created, which is the controller's next turn, so its
+    // untap step is exactly the next occurrence of that step for the controller
+    // (CR 614.10a: a skip waits for the first occurrence that isn't skipped).
+    // The subject is elided to the resolving controller. Lowers to the same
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved. Evidence checked at current head: the parse-diff is scoped to Savor the Moment / Time Bends to My Will replacing the unimplemented skip clause with existing SkipNextStep, the change reuses the existing step-skip parser/resolver surface instead of adding machinery, and CI is green. Tests include both an isolated clause case and full-card no-Unimplemented reach guard. For future first-time AI-assisted PRs, please keep using docs/AI-CONTRIBUTOR.md as the contribution checklist.

PR #4890, opened 2026-07-02
Absorb reveal-until rest pile after damage

ntindle:
## Summary
- Absorb reveal-until rest-pile clauses that appear after intervening damage into the original RevealUntil effect instead of emitting a standalone bottom-of-library effect.
- Add regression coverage using the Explosive Revelation shape.
- Remove the fixed/reclassified root-32 entries from docs/parser-misparse-backlog.md.

## Verification
- cargo fmt --all
- ./scripts/check-parser-combinators.sh
- cargo test -p engine --lib reveal_until_rest_pile_after_intervening_damage_is_absorbed -- --nocapture
- cargo clippy -p engine --lib -- -D warnings
- Full parser export before/after comparison changed exactly one card key: explosive revelation. The JSON diff only removes the stray PutAtLibraryPosition sibling under the damage effect; RevealUntil keeps its hand destination and library rest destination.
- Ship worktree checks: git diff --check HEAD~1..HEAD, cargo fmt --all --check

### DIFF
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -5483,6 +5483,18 @@ pub(super) fn parse_followup_continuation_ast(
                 reorder_all: false,
             })
         }
+        // CR 701.20a + CR 608.2c: A reveal-until rest-pile clause may be
+        // separated from the RevealUntil by a transparent intervening effect
+        // ("~ deals damage equal to that card's mana value. Put that card into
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Approved. Evidence checked at current head: the parse-diff is scoped to Explosive Revelation and only removes the stray PutAtLibraryPosition sibling; the implementation absorbs the rest-pile clause into the existing RevealUntil continuation path without adding new parser surface. CI is green and the regression covers the intended shape.

PR #4892, opened 2026-07-02
Fix Hama Pashar dungeon room trigger doubling (fixes #4358)

### LINKED ISSUE #4358: Hamar Pasher, Ruin still seeking — [[Hamar Pasher, Ruin Seeker]] doesnt create another dungeon room trigger event.
Reported in Discord: https://discord.com/channels/1485498006781427802/1519256951832182836/1519256951832182836

**Thread:** Hamar Pasher, Ruin still seeking
**Cards:** _none detected_

## Summary
[[Hamar Pasher, Ruin Seeker]] doesnt create another dungeon room trigger event.

---
<sub>report_id: `discord:1519256951832182836:1519256951832182836:0` · discord: `1519256951832182836/1519256951832182836`</sub>

kiannidev:
## Summary
- Parse "Room abilities of dungeons you own trigger an additional time" as `DoubleTriggers` with a new `RoomEntered` cause.
- Apply trigger doubling when dispatching synthetic dungeon room triggers.

## Test plan
- [x] `cargo test -p engine --lib hama_pashar_room`
- [x] `cargo test -p engine --test integration issue_4358`
- [x] `cargo fmt --all -- --check`

Fixes #4358


matthewevans (dismissed):
**[HIGH]** Hama Pashar currently doubles the wrong player’s dungeon room triggers. Evidence: `crates/engine/src/game/triggers.rs:4758` adds `TriggerCause::RoomEntered` as `matches!(event, Some(GameEvent::RoomEntered { .. }))`, while `apply_trigger_doubling` only checks that the pending trigger controller matches the doubler controller and does not compare the `RoomEntered.player_id` / dungeon owner to the doubler controller. Why it matters: Hama Pashar says “Room abilities of dungeons you own,” so a room trigger controlled by the Hama controller but caused by another player’s dungeon event would be doubled incorrectly, and the current regression only covers the controller’s own dungeon. Suggested fix: make the `RoomEntered` cause check bind to the `player_id`/owner in the event, and add a negative test where the Hama controller does not own the dungeon whose room was entered.

matthewevans (changes requested):
**[HIGH]** This needs one more pass before merge: the Hama Pashar parser branch is new Oracle parser dispatch written as `scan_contains` phrase matching in `crates/engine/src/parser/oracle_static/dispatch.rs`, which violates the repo parser bar for new parser work. The existing nearby trigger-doubling block also uses scan helpers, but new coverage should not extend that style; decompose the phrase with nom combinators / existing trigger-doubling parsing structure instead. Also address the current unresolved CR-annotation comments in `triggers.rs` for the synthetic trigger-doubling path and the `RoomEntered` cause arm. Once those are fixed, the production trace for ownership looks sound: `queue_room_trigger` constructs the pending room trigger with `controller == RoomEntered.player_id`, so the dismissed owner-mismatch blocker is not an active concern.

matthewevans (changes requested):
**[LOW]** This adds an npm lockfile to a pnpm-managed client without package metadata changes. Evidence: `client/package-lock.json:1` is newly added, while `client/package.json:8` continues to use pnpm scripts and no package-manager metadata changed. Why it matters: this creates a second dependency authority that can drift from `pnpm-lock.yaml` and confuse local or automation installs. Suggested fix: remove `client/package-lock.json` and keep the existing pnpm lockfile as the single client lock authority.

matthewevans (approved):
Reviewed current head `03f7c0a7fc07a637455ee54a5abcb28ac8743246`. The prior package-lock blocker is gone, the Hama Pashar static is now parsed through composed nom tags rather than `scan_contains` dispatch, and the runtime test covers the dungeon room trigger doubling path. The parse-diff is scoped to the Room-trigger class. I do not see a remaining blocker.

PR #4895, opened 2026-07-02
Parse attached restriction lists with activation locks

ntindle:
Summary:
- Parse attached-subject restriction lists before the companion activated-ability lock.
- Emit the missing restriction statics for Bound in Gold style Aura text and Avacyn's Collar style Equipment text.
- Remove fixed or stale cards from root cause 31 in docs/parser-misparse-backlog.md.

Intentional parser output changes:
- Bound in Gold: adds CantBlock and CantCrew alongside the existing CantAttack and CantBeActivated(ManaAbilities).
- Avacyn's Collar, the Symbol of Her Church: adds CantBlock and CantTransform alongside the existing CantAttack and CantBeActivated(None).

Verification:
- cargo fmt --all
- cargo fmt --all --check
- ./scripts/check-parser-combinators.sh
- git diff --check
- RUSTC_WRAPPER= cargo test -p engine --lib cant_be_activated_compound_aura -- --nocapture
- RUSTC_WRAPPER= CARGO_TARGET_DIR=/home/ntindle/code/phase/target cargo test -p engine --lib cant_be_activated_compound -- --nocapture
- RUSTC_WRAPPER= CARGO_TARGET_DIR=/home/ntindle/code/phase/target cargo test -p engine --lib static_enchanted_creature_cant_attack -- --nocapture
- RUSTC_WRAPPER= CARGO_TARGET_DIR=/home/ntindle/code/phase/target cargo clippy -p engine --lib -- -D warnings
- focused root31 export diff: only Bound in Gold changed among the seven triaged root-31 cards
- coverage sticky diff: Bound in Gold and Avacyn's Collar changes are intentional same-root-cause fixes

Notes:
- Source checkout cargo commands used RUSTC_WRAPPER= because Cargo plus the default sccache wrapper fails in this sandbox with Operation not permitted.
- Ship-worktree checks passed: git diff --check HEAD~1..HEAD and cargo fmt --all --check.

### DIFF
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -682,6 +682,67 @@ pub(crate) fn attached_subject_filter<'a>(tp: &TextPair<'a>) -> Option<(TargetFi
     None
 }
 
+/// CR 602.5: Parses the activation-prohibition tail of compound static text.
+fn parse_activation_compound_tail(input: &str) -> OracleResult<'_, ()> {
+    value(
+        (),
+        (
+            tag(", and "),
+            opt(alt((tag("its "), tag("their ")))),
+            alt((
+                tag("activated abilities can't be activated"),
+                tag("activated abilities can\u{2019}t be activated"),
+            )),
+            opt((tag(" unless they're "), tag("mana abilities"))),
+            opt(tag(".")),
+        ),
+    )
+    .parse(input)
+}
+
+fn rule_static_predicate_to_activation_compound_mode(
+    predicate: RuleStaticPredicate,
+) -> Option<StaticMode> {
+    match predicate {
+        RuleStaticPredicate::CantAttack => Some(StaticMode::CantAttack),
+        RuleStaticPredicate::CantBlock => Some(StaticMode::CantBlock),
+        RuleStaticPredicate::CantAttackOrBlock => Some(StaticMode::CantAttackOrBlock),
+        RuleStaticPredicate::CantCrew => Some(StaticMode::CantCrew),
+        RuleStaticPredicate::CantUntap
+        | RuleStaticPredicate::CantBeActivated
+        | RuleStaticPredicate::CantBeSacrificed
+        | RuleStaticPredicate::MustAttack
+        | RuleStaticPredicate::MustBlock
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[MED]** The new compound parser cannot enter for the `CantCrew`-only sibling it appears to support. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:1158` only dispatches this compound activation branch when the line contains `can't attack` or `can't block`, but `parse_activation_compound_restriction_modes` maps `RuleStaticPredicate::CantCrew` at `crates/engine/src/parser/oracle_static/shared.rs:702`; a line like `Enchanted permanent can't crew Vehicles, and its activated abilities can't be activated...` would skip this branch entirely. Why it matters: the new helper is presented as a restriction-list parser, but one of its explicit modes is unreachable unless an attack/block phrase is also present. Suggested fix: broaden the entry predicate to the same combat restriction parser used inside the branch, and add a focused test for the `can't crew Vehicles, and its activated abilities...` form.

**[MED]** Required parse-diff evidence is still absent for this parser-source PR. Evidence: the inspect packet for current head `d4cad78b4a508c176616f56284b53102b22fd984` has `parse_diff.state = absent`, and the card-data/coverage CI job is still queued. Why it matters: parser PR approval requires the sticky card-level diff so we can verify Bound in Gold is the only intended parser output change and no unrelated cards moved. Suggested fix: let CI produce the `<!-- coverage-parse-diff -->` sticky and ensure it is scoped to the claimed card/class before requesting re-review.

matthewevans (commented):
I reviewed the current `98a656e748ad7f11f1f18fd61c927c5e66ad52bc` head. The earlier crew-only entry blocker appears addressed: the new predicate path now uses `attached_activation_compound_modes`, and there is focused coverage for `Enchanted permanent can’t crew Vehicles, and its activated abilities...`.

I am not approving yet because this is parser output work and the required current-head parse-diff sticky is still absent while card-data/coverage CI is pending. I will re-check once that evidence lands for this head.

matthewevans (approved):
Reviewed current head with parse-diff evidence; the prior crew-only blocker is covered and the parser movement is scoped to Bound in Gold / Avacyn's Collar restriction-list outputs.

PR #4897, opened 2026-07-02
fix(parser): Tinfoil Helm — copula "has +N/+M" pump + bare-list protection expansion

real-venus:
## Summary

**Tinfoil Helm** — "Equipped creature has +2/+2 and has protection from Aliens, Birds, Eldrazi, Lizards, Mutants, Robots, Yetis, and hybrid mana." — parsed to a single static carrying only `Protection(CardType("aliens"))`: the **+2/+2 pump was dropped** and the **protection list truncated** to its first quality. Two parser gaps, both fixed for the whole class:

### 1. Copula "has +N/+M" pump (`parse_fixed_pt_in_text`)
The fixed-P/T scanner only recognized the verb "gets"/"get", so the copula "has +2/+2" produced no modification. Broadened the verb `alt` to `gets/get/has/have` (keeping the **sign-required** `parse_pt_modifier`), with a **counter-suffix negative-lookahead** so counter-placement lines are never misread as pumps — e.g. **Melira, Sylvok Outcast** ("…can't have -1/-1 counters put on them") must not become a −1/−1 anthem. *(CR 613.4c / CR 122.1a)*

### 2. Bare comma-list protection (`expand_protection_parts`)
Only "from"-prefixed continuation members were expanded, so "protection from A, B, C, and D" dropped every member after the first. Widened the fast-path guard to route a bare protection list to the slow path, and added a three-way continuation branch: a bare **quality** (color, creature subtype, "hybrid mana") continues the active protection/hexproof prefix, while a genuine trailing **keyword** (vigilance, first strike) resets the list. A recognized creature **subtype takes precedence** over a keyword-name collision (reuses `oracle_util::is_subtype_word`). *(CR 702.16a / 702.16g / 702.16i)*

## Notes
- **No new enum variant** — runtime already matches subtype protection via `ProtectionTarget::CardType` fall-through in `game/keywords.rs`.
- "protection from hybrid mana" parses to an inert quality — an **honest deferred runtime gap**, not a false coverage claim.
- Nom combinators only; `mtgish/` untouched.

## Testing
Building-block + full-pipeline tests: Tinfoil → `AddPower{2}`+`AddToughness{2}` + 8 `Protection` entries; **Melira → no pump** (revert-failing regression guard); copula unit tests reject "has protection…"/"+N/+N counter"; bare-list expansion for colors/subtypes; trailing-keyword reset. **1165 parser tests pass, 0 fail** (`oracle_keyword` + `oracle_static` + `oracle_util`).

### DIFF
--- a/crates/engine/src/parser/oracle_keyword.rs
+++ b/crates/engine/src/parser/oracle_keyword.rs
@@ -35,6 +35,25 @@ use crate::types::zones::Zone;
 /// Plate) or "each color with the most votes" (Council Guardian) carry
 /// additional qualifiers and pass through unchanged for a future dynamic
 /// handler.
+///
+/// CR 702.16i: also expands a bare comma-list continuation — after a
+/// "protection from A" prefix, subsequent list members that are neither
+/// "from …"-prefixed continuations nor genuine trailing keywords (e.g. Tinfoil
+/// Helm's "protection from aliens, birds, …, and hybrid mana", or a color list
+/// "protection from white, blue, red") each become their own protection entry.
+/// A genuine trailing keyword (Akroma's "vigilance") resets the prefix and
+/// passes through unexpanded.
+fn protection_prefix(i: &str) -> nom::IResult<&str, &'static str, OracleError<'_>> {
+    // (prefix_with_space, emit_prefix_no_space) — strip the prefix+space,
+    // emit the prefix without space. Shared by the prefix branch and the
+    // fast-path guard so both recognize the same set.
+    alt((
+        value("protection from", tag("protection from ")),
+        value("hexproof from", tag("hexproof from ")),
+    ))
+    .parse(i)
+}
+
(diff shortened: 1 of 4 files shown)


matthewevans (commented):
I reviewed the current parser diff at `f51d92a5228a910773b91793b99d1d67ddc275b0`. I am not approving yet because this is parser coverage work and the required parse-diff sticky is not present on the PR yet, with CI still pending.

From the code diff, the approach is in the right seam: it extends the existing keyword-list expansion and fixed-P/T scanner rather than adding card-specific dispatch, and the tests include the Tinfoil Helm full static parse plus counter/keyword negative cases. I will re-check once parse-diff evidence lands for this head.

matthewevans (approved):
Reviewed current head with green CI and parse-diff evidence; the parser changes are scoped to the copula P/T grant and bare protection-list expansion classes.

PR #4880, opened 2026-07-02
fix(parser): attach board-state count gate to self-spell cost reduction (Hour of Revelation)

minion1227:
Fixes a parser correctness bug on **Hour of Revelation** — *"This spell costs {3} less to cast if there are ten or more nonland permanents on the battlefield."*

## Problem
The trailing `if [board-state]` gate was **swallowed**, so the cost reduction applied **unconditionally** (Hour of Revelation would always cost {3} less, regardless of the board). The self-spell cost-mod condition extractor tried `parse_cost_modifier_condition` and `parse_inner_condition`, but neither recognized the **existential** phrasing `"there are [N] or more [type] on the battlefield"` (the existing count helper anchors on `"[N] or more [type] are on the battlefield"`).

## Fix (both edits in test-free files)
- **`oracle_static/shared.rs`** — add `parse_there_are_count_on_battlefield_condition`, the existential counterpart of `count_on_battlefield_condition`, with the same `ObjectCount(type) >= N` shape. Wired into `parse_static_condition` alongside the count/existence helpers.
- **`oracle_static/static_helpers.rs`** — fall back to `parse_static_condition` in the self-spell cost-mod trailing-condition extraction, so board-state gates attach.

Composed from existing `ObjectCount` / `QuantityComparison` primitives — **no new variant**. General across the *"costs {N} less if there are [N] or more [type] on the battlefield"* class.

## CR
CR 601.2f (cost modification) + CR 611.3a (conditional statics).

## Verification
- Full `parser::oracle_static` suite green — **930 passed, 0 failed**.
- New regression `hour_of_revelation_cost_reduction_gated_on_battlefield_count` asserts the `ModifyCost` static gains an `ObjectCount(nonland permanents) >= 10` gate with zero swallowed clauses.
- `cargo fmt` + `cargo clippy -p engine -- -D warnings` clean.
- Regression-checked: the existing `"[N] or more [type] are on the battlefield"` count form and other cost-mod conditions are unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4898, opened 2026-07-02
fix(parser): opponent-scoped "control N or more" cost-reduction gate (Lashwhip Predator)

minion1227:
Fixes a parser correctness bug on **Lashwhip Predator** — *"This spell costs {2} less to cast if your opponents control three or more creatures."*

## Problem
The trailing `if [board-state]` gate was **swallowed**, so the cost reduction applied **unconditionally**. `parse_static_condition` handled the *"you control N or more X"* controller scope but not *"your opponents control N or more X"*.

## Fix
- **`oracle_static/shared.rs`** — add `parse_opponents_control_count_condition` → `ObjectCount(type controlled by an opponent) >= N`, injecting `ControllerRef::Opponent` onto the counted filter (mirroring the existing `you control` count path). Wired into `parse_static_condition`.

Composed from existing `ObjectCount` / `QuantityComparison` primitives — **no new variant**. General across the *"costs {N} less if your opponents control [N] or more [type]"* class, and reusable by any static gated on an opponent-scoped count.

## CR
CR 601.2f (cost modification) + CR 611.3a (conditional statics).

## Verification
- Full `parser::oracle_static` suite green — **933 passed, 0 failed**.
- New regression `lashwhip_predator_cost_reduction_gated_on_opponents_creature_count` asserts the `ModifyCost` static gains an opponent-scoped `ObjectCount(creatures) >= 3` gate with zero swallowed clauses.
- Follow-up to #4880 (self-spell cost-mod board-state gates); the `you control` count path is unaffected.

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -2915,13 +2915,32 @@ fn controlled_battlefield_subtype_filter(subtype: String) -> TargetFilter {
     )
 }
 
-/// Canonical combinator: "you control N or more [type]" → QuantityComparison.
+/// Parse the leading controller-scope phrase of a "control N or more" count
+/// condition, returning the `ControllerRef` the count is scoped to:
+/// "you control " → `You`, "your opponents control " → `Opponent`.
+///
+/// CR 109.3: object control is a per-player property; this is the single axis
+/// that distinguishes the self-scoped ("you control three or more creatures")
+/// and opponent-scoped ("your opponents control three or more creatures",
+/// Lashwhip Predator) forms of the same `ObjectCount >= N` threshold.
+fn parse_control_scope_prefix(input: &str) -> OracleResult<'_, ControllerRef> {
+    alt((
+        value(ControllerRef::You, tag("you control ")),
+        value(ControllerRef::Opponent, tag("your opponents control ")),
+    ))
+    .parse(input)
+}
+
+/// Canonical combinator: "you control / your opponents control N or more [type]"
+/// → QuantityComparison.
 ///
 /// Single authority for this pattern — called from `oracle_static.rs` and
-/// `oracle_trigger.rs` to avoid three-way duplication.
+/// `oracle_trigger.rs` to avoid three-way duplication. The controller scope is
+/// parameterized on the `you control` / `your opponents control` axis
+/// (`parse_control_scope_prefix`) so both forms share one parse path.
 /// Returns the remainder after the type phrase (may be non-empty for trailing text).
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[HIGH]** This adds the new condition at the static-only fallback seam instead of the shared condition parser. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:1459` already delegates static conditions to `nom_condition::parse_inner_condition` as the single game-state condition path, but the new `parse_opponents_control_count_condition` is added later in the static parser at `crates/engine/src/parser/oracle_static/shared.rs:1559`/`:2062`; the shared `oracle_nom::condition` parser already owns the adjacent `you control N or more` and `an opponent controls N or more` count grammars at `crates/engine/src/parser/oracle_nom/condition.rs:2742` and `:6372`. Why it matters: the same “your opponents control N or more …” condition will still fail anywhere that reaches `parse_inner_condition` directly, so this fixes Lashwhip through one static path while proliferating a duplicate condition grammar outside the shared authority. Suggested fix: extend the shared condition parser/controller axis to accept `your opponents control ` as `ControllerRef::Opponent`, let `parse_static_condition` pick it up through the existing delegate, and add a `parse_inner_condition` building-block test plus the Lashwhip static fixture.

Also, the required parser parse-diff sticky is currently absent on this PR, so it would remain unapprovable until that evidence is present even after the seam is fixed.

minion1227:
Thanks — addressed the [HIGH] finding by moving the opponent-scope count grammar into the shared `oracle_nom::condition` authority instead of the static-only seam.

- Parameterized `parse_control_count_ge` on the controller-scope axis via a new `parse_control_scope_prefix` combinator (`you control ` → `You`, `your opponents control ` → `Opponent`), then `inject_controller(filter, ctrl)`. `parse_static_condition` picks it up through its existing delegate to `parse_inner_condition`, so triggers/intervening-ifs that reach the shared condition parser directly now handle the opponent form too (it also makes `unless your opponents control N or more …` resolve correctly via the `Not`-wrapping consumer in `evasion.rs`).
- Deleted the duplicate `parse_opponents_control_count_condition` / `opponents_control_count_condition` from `oracle_static/shared.rs`.
- Added a `parse_inner_condition` building-block test (`parse_inner_condition_control_count_ge_controller_scope_axis`, both controller scopes) and kept the Lashwhip static fixture.

Full `oracle_static` (933) + `oracle_nom::condition` (357) suites green, clippy clean. CR 109.3 + CR 603.4 + CR 611.3a.

matthewevans (approved):
Maintainer review: the previous wrong-seam finding is resolved by parameterizing the shared control-count condition parser; parse-diff is limited to Lashwhip Predator's ReduceCost gate. Approving for merge-when-ready.

PR #4899, opened 2026-07-02
Fix Ixhel corrupted poison-scoped exile trigger (fixes #3882)

### LINKED ISSUE #3882: ixhel, scion of atraxa — it's not triggering the "corrupted" ability keyword.
Reported in Discord: https://discord.com/channels/1485498006781427802/1517601521909633064/1517601521909633064

**Thread:** ixhel, scion of atraxa
**Cards:** _none detected_

## Summary
it's not triggering the "corrupted" ability keyword.

---
<sub>report_id: `discord:1517601521909633064:1517601521909633064:0` · discord: `1517601521909633064/1517601521909633064`</sub>

kiannidev:
## Summary
- Strip "who has/have N or more poison counters" relative clauses into `PlayerFilter::PlayerAttribute` when parsing each-player subjects
- Accept singular "who has" in the shared player-counter attribute grammar used by quantity refs
- Add parser and integration regression tests for Ixhel, Scion of Atraxa's corrupted end-step exile

Fixes #3882

## Test plan
- [x] `cargo test -p engine --lib strip_each_player_subject_poison_counter_clause`
- [x] `cargo test -p engine --lib ixhel_corrupted_end_step_trigger_parses_poison_scoped_exile`
- [x] `cargo test -p engine --test integration issue_3882`
- [x] `cargo clippy -p engine --tests -- -D warnings`


matthewevans (changes requested):
**[HIGH]** The new subject-path parser duplicates only the counter subset of an existing player-attribute grammar. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:4206` adds a fresh `strip_player_attribute_clause` that manually parses only `who has/have N or more <kind> counters`, while the existing quantity authority at `crates/engine/src/parser/oracle_quantity.rs:1372` already builds `PlayerFilter::PlayerAttribute` from the same population/attribute shape and covers counter, hand-size, cards-drawn, and battlefield-entry predicates; the adjacent controls subject path follows the intended pattern by delegating to a shared core (`parse_controls_permanent_object`) at `crates/engine/src/parser/oracle_effect/lower.rs:4187`. Why it matters: this fixes Ixhel by proliferating a second, narrower grammar, so future `each opponent with N cards in hand` / `who drew N cards` / `who had N permanents enter` subject forms will keep failing or require another duplicate parser. Suggested fix: factor/export a shared player-attribute predicate core (population + attr clause, or attr-after-population) and have both the quantity path and `strip_each_player_subject` consume it, then add subject-path class tests for at least one non-counter attribute sibling.

Also, the required parser parse-diff sticky is absent on the current head, so the PR would remain unapprovable until that evidence appears even after the parser seam is fixed.

matthewevans (changes requested):
**[MED]** The new subject-path player attribute parser exposes only the counter subset of the existing attribute grammar. Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:4217` delegates only to `parse_player_counter_attribute_suffix`, while `crates/engine/src/parser/oracle_quantity.rs:1374` already treats counters, hand size, cards drawn, and battlefield-entry predicates as one `PlayerAttribute` class. Why it matters: Ixhel is fixed, but sibling subject forms with the same semantics still need another parser change instead of reusing the established building block. Suggested fix: factor/export a shared attribute-suffix parser covering the full existing `PlayerAttribute` set and have both quantity parsing and `strip_each_player_subject` consume it, with at least one non-counter subject-path test.

matthewevans:
Holding approval after update-branch to current head `58c932b5eda1ee8995c2cda25565b0d1d3bd6043`: the code review pass on the pre-update head looked clean and the branch update is a merge from `main`, but the required parse-diff sticky is still the older pre-update evidence (`baseline: main 699ddac2de81`, posted before this head). Because this is parser-surface, I need a current-head parse-diff before approving, labeling, or enabling merge-when-ready.

matthewevans (approved):
Maintainer review: the current head addresses my prior blocker. The subject-path parser now delegates to the shared `parse_player_attribute_attr_clause` grammar rather than adding a poison-counter-only parser, so this is parameterized across the existing player-attribute class. I also verified the runtime path: `PlayerFilter::PlayerAttribute` reads the scalar from each candidate player, and the new integration tests cover both the corrupted and below-threshold Ixhel cases through trigger resolution.

Parse-diff evidence is present and matches the expected Ixhel-only change; the latest head commit is comment-only, so it does not invalidate the behavioral parse evidence.

PR #4900, opened 2026-07-02
Add Coastal Wizard

e11734937-beep:
## Summary
Adds engine support for **Coastal Wizard** — "{T}: Return this creature and another target creature to their owners' hands." — and, from the same fix, **Lady Sun**.

The compound bounce splits on " and " into "Return ~" (bounce self) plus the
verbless second conjunct "another target creature".
`try_split_targeted_compound`'s verb-carry-forward — which re-parses a verbless
trailing conjunct by prepending the primary effect's verb — only fired when that
conjunct began with `"target "`. "another target creature" begins with
`"another "`, so the carry-forward was skipped and the second creature was
orphaned (`Effect::Unimplemented`); only the source returned to hand.

The fix extends that one guard to also accept a `"another target "` head (CR
608.2c). "another target creature" is already a first-class `FilterProp::Another`
target used by ~249 supported cards, so the re-parse of "Return another target
creature" lowers to a real bounce with no other changes. The fix is generic — it
covers any `<verb> X and another target Y` compound, not just these two cards.

## Files changed
- `crates/engine/src/parser/oracle_effect/mod.rs` — carry-forward guard accepts `"another target "`
- `crates/engine/tests/coastal_wizard_bounce_self_and_another.rs` — integration test

## Anchored on
- `crates/engine/src/parser/oracle_effect/mod.rs` — the existing `"target "` verb-carry-forward in `try_split_targeted_compound` ("exile target creature and target artifact"). This change adds `"another target "` to the same guard.
- `crates/engine/src/parser/oracle_target.rs` (~line 510 / 1703) — `parse_target` already recognizes "another target creature" → `FilterProp::Another`; ~249 supported cards use it, so the carried-forward "Return another target creature" parses without new target work.

## CR references
- `CR 608.2c` — resolving each part of a compound instruction; the trailing verbless conjunct inherits the primary effect's verb.
- `CR 115.1` — "another target creature" is a target constrained to a creature other than the source (`FilterProp::Another`).

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine` — N pass / 0 fail (adds `coastal_wizard_returns_self_and_another_target_to_hand`)
- `./scripts/gen-card-data.sh` — Coastal Wizard / Lady Sun: 0 Unimplemented entries
- `cargo coverage` — both: `supported: true`, `gap_count: 0`
- Coverage regression diff (clean HEAD baseline vs this change, `scripts/coverage-regression-check.sh --fail-on-engine`): supported 31197 → 31199 (net +2). GAINED: 2 (Coastal Wizard, Lady Sun). REGRESSED (engine): 0. REGRESSED (coverage honesty): 0. ORACLE CHANGED: 0. (The ~249 existing "another target creature" cards are unchanged.)
- `cargo semantic-audit` — Coastal Wizard & Lady Sun: 0 findings (32301 cards audited)

The integration test drives the real activate → target → resolve pipeline:
activate the `{T}` ability, target a second creature, and assert BOTH the source
and the targeted creature return to hand. It flips if the fix is reverted — the
second bounce conjunct would be dropped and the targeted creature would remain on
the battlefield.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.


matthewevans:
Reviewed the current head. I do not see a code-level blocker in the narrow verb-carry-forward change, but this touches parser output and the required parse-diff sticky is absent for this head. Please wait for the parse-diff evidence to publish before this can be approved or queued.

matthewevans (commented):
Reviewed the current head. I do not see a code-level blocker in the narrow verb-carry-forward change, but this touches parser output and the required parse-diff sticky is absent for this head. Holding approval/queue until the parse-diff evidence publishes.

matthewevans (approved):
Maintainer review: narrow parser carry-forward fix; parse-diff is limited to Coastal Wizard/Lady Sun and the runtime test exercises the activate/resolve path. Approving for merge-when-ready.

PR #4902, opened 2026-07-02
Support "loses A or B" keyword removal (Urborg)

e11734937-beep:
## Summary

**Urborg** — `{T}: Add {B}.` / `{T}: Target creature loses first strike or swampwalk until end of turn.` — had its second activated ability fall through to `Effect::Unimplemented`. The keyword-loss recognizer already handled the `" and "` conjunction (via the shared `split_keyword_list`), but deliberately did not split on `" or "`, so `"first strike or swampwalk"` stayed a single token and `map_keyword` failed.

## Change

In the keyword-loss path of `keyword_grant.rs`, split each keyword-loss segment on `" or "` (scoped to the *loss* path only), emitting one `ContinuousModification::RemoveKeyword` per keyword — the identical multi-removal shape the `" and "` path already produces. Uses nom combinators (`separated_list1` over the loss segment); no new effect or enum.

**Why "or" removes both (not a choice):** the card has no "of your choice"/"choose" language. Per CR 613, removing an ability an object doesn't have is a no-op, so "loses first strike or swampwalk" removes whichever the creature has — modeled as two `RemoveKeyword`s, which is correct in every case and matches the established `" and "` behavior. The split is confined to the loss path, preserving the grant-path distinction (a "has A or B" grant would be a choice, not both).

## Tests

- `loses_keyword_or_keyword_splits_into_two_remove_keywords` — building block: `"loses first strike or swampwalk"` → `[RemoveKeyword(FirstStrike), RemoveKeyword(Landwalk("Swamp"))]`, and equal to the `" and "` form.
- `urborg_loses_first_strike_or_swampwalk_parses_end_to_end` — full card: 2 abilities, the 2nd a `GenericEffect { duration: UntilEndOfTurn, target: Typed(Creature) }` with both removals, zero `Unimplemented`.

Fails-before (clause → `Unimplemented`) / passes-after confirmed. Green: rustfmt, parser-combinator (Rule Zero) gate, `clippy -D warnings`, full engine suite (14522 passed / 0 failed).


matthewevans (changes requested):
The Urborg part is heading in the right direction, but this head changes another card outside the claimed scope and needs to be tightened before it can merge.

**[HIGH]** The new `" or "` splitter leaks into `can't have ... or ...` riders and duplicates Arcane Lighthouse's shroud removal. Evidence: the current parse-diff sticky reports Arcane Lighthouse changed from `remove Hexproof, remove Shroud` to `remove Hexproof, remove Shroud, remove Shroud`; the new loop applies `split_lose_or_segments` to every `part` produced by `split_keyword_list` in `crates/engine/src/parser/oracle_static/keyword_grant.rs`, so a tail like `can't have hexproof or shroud` can be split and the trailing `shroud` maps as another `RemoveKeyword`. Why it matters: this is unintended parser blast radius on an existing supported card and the tests only cover the bare Urborg phrase, not the sibling loss-plus-cant-have form already modeled in `append_cant_have_keyword_denials`. Suggested fix: constrain the `or` splitting to actual keyword-loss tokens before any `can't have`/`can't have or gain` rider, and add a regression test that Arcane Lighthouse-style text still emits exactly one Hexproof removal and one Shroud removal.

matthewevans (changes requested):
**[MED]** The new `or` splitter regresses an existing keyword-loss card by duplicating one removal. Evidence: `crates/engine/src/parser/oracle_static/keyword_grant.rs:1433` plus the current `coverage-parse-diff` sticky for this head, which shows Arcane Lighthouse changing from `remove Hexproof, remove Shroud` to `remove Hexproof, remove Shroud, remove Shroud`. Why it matters: this PR is intended to split Urborg's old-frame `or` loss clause, but the implementation also changes an unrelated supported card's parse signature, so coverage is no longer honest for the existing class. Suggested fix: adjust the loss-list splitting so each keyword is emitted exactly once, and add a regression assertion for Arcane Lighthouse / `opponents' creatures lose hexproof and shroud until end of turn` alongside the Urborg case.

matthewevans:
Additional blocker from follow-up review: this should not parse Urborg as removing both listed abilities. The engine already has the `ChoiceType::Keyword { count: 1 }` / `ContinuousModification::RemoveChosenKeyword` path for this exact Urborg/Walking Sponge class (`crates/engine/src/types/ability.rs` documents Urborg there), and the reference import represents Urborg as `ChooseACheckableAbility` followed by losing `TheChosenAbility`. The current branch instead emits two `RemoveKeyword`s, so a creature with both first strike and swampwalk would lose both rather than the chosen one. The fix should route this form through the existing chosen-keyword authority, not a broad inclusive `or` splitter.

e11734937-beep:
Thanks — fixed the blast radius. The `" or "` split now only applies when *every* segment is itself a keyword (`map_keyword(...).is_some()`), so a `can't have hexproof or shroud` rider is left intact and no longer duplicates a removal. Added the requested regression `arcane_lighthouse_lose_and_cant_have_or_rider_no_duplicate_removal` (fails-before: 3 removals; passes-after: 2). Verified with the corpus parse-diff: across all 35,397 faces exactly one card changed — Arcane Lighthouse 3 → 2 removals — with 0 other regressions (coverage-regression-check: 0 regressed, +1 Urborg). Full suite 14554 green, clippy clean.

matthewevans (changes requested):
**[HIGH]** Urborg still bypasses the existing chosen-keyword authority and removes both listed abilities. Evidence: `crates/engine/src/parser/oracle_static/keyword_grant.rs:1450` splits `"first strike or swampwalk"` into two unconditional `RemoveKeyword` modifications, while `crates/engine/src/types/ability.rs:350` documents the existing `ChoiceType::Keyword { count: 1 }` / `RemoveChosenKeyword` path for the Urborg/Walking Sponge class and `crates/engine/src/game/layers.rs:8578` verifies that only the chosen landwalk ability is removed. Why it matters: a target with both abilities should lose the chosen one, not every listed option. Suggested fix: parse this form through a persisted keyword choice plus `ContinuousModification::RemoveChosenKeyword`, and update the current tests that assert two unconditional removals.

matthewevans (changes requested):
**[HIGH]** Urborg still removes both listed abilities instead of choosing one. Evidence: `crates/engine/src/parser/oracle_static/keyword_grant.rs:1450` pushes every `or` segment as `RemoveKeyword`, and `crates/engine/src/parser/oracle_static/tests.rs:22255` / `crates/engine/src/parser/oracle_static/tests.rs:22319` assert both FirstStrike and Swampwalk removals; the parse-diff sticky still reports Urborg added as `remove FirstStrike, remove Landwalk`. Why it matters: a target with both abilities should lose only the chosen keyword, and the repo already has the `ChoiceType::Keyword { count: 1 }` plus `ContinuousModification::RemoveChosenKeyword` authority documented for this class. Suggested fix: parse this form through a persisted keyword choice and downstream `RemoveChosenKeyword`, then update the tests and parse-diff to prove Urborg chooses one keyword.

e11734937-beep:
Thanks for the detailed pointer — you're right that "loses first strike or swampwalk" should be a keyword *choice* (lose one), not both, via a persisted `ChoiceType::Keyword { count: 1 }` → `RemoveChosenKeyword`. That's a larger choice-flow change than the parser split this PR took, so I'll close this rather than layer more onto the branch, and revisit it as a proper choice-model PR. Appreciate the review!

e11734937-beep:
Reworked per your guidance — `loses A or B` now routes through a persisted keyword choice instead of unconditional removals: the head is `Effect::Choose { ChoiceType::Keyword { count: 1 }, persist }` and the sub-ability applies `ContinuousModification::RemoveChosenKeyword`, so a creature with both abilities loses only the chosen one (CR 702.14). Reverted the earlier remove-both split in keyword_grant.rs (now byte-identical to main), so the Arcane Lighthouse `can't have … or …` rider is intact (regression kept).

One note: the existing `RemoveChosenKeyword` runtime (layers.rs:4333) + `ChoiceType::Keyword` were in place, but no parser actually emitted them — I wrote the parser wiring modeled on `try_parse_become_choice` (subject.rs), splitting the keyword list via the shared `try_parse_keyword_choice`. Runtime coverage: the existing `test_remove_chosen_keyword_strips_only_chosen_landwalk_variant` (+ siblings) pass. Full suite 14623 green, clippy clean; corpus parse-diff shows Urborg is the only real change (the other diffs reproduce in a HEAD-vs-HEAD self-diff = export nondeterminism).

PR #4903, opened 2026-07-02
Trigger on your own countered spell (Multani's Presence)

e11734937-beep:
## Summary

**Multani's Presence** — *"Whenever a spell you've cast is countered, draw a card."* — parsed its trigger to `Unknown("Whenever a spell you've cast is countered")` and silently dropped it. The engine already models this via `TriggerMode::Countered`, but only the *active* side ("a spell or ability you control counters a spell") had a recognizer.

## Change

Add `parse_own_spell_countered_line` to `try_parse_player_trigger` (`crates/engine/src/parser/oracle_trigger.rs`) — the *passive* dual of the existing countering-side arm. It recognizes `"whenever|when a spell you've cast|you control is countered"` (whole-clause `all_consuming`, nom combinators only) and emits `TriggerMode::Countered` with `valid_card = Typed(controller = You)`.

At runtime, `match_countered` (`game/trigger_matchers.rs`) evaluates `valid_card` against the `SpellCountered` event's `object_id` (the countered spell), so the `You` controller filter restricts the trigger to *your own* countered spell — the passive dual of the active arm, which gates the countering source via `valid_source`. The `Draw` child was already supported.

**Semantics:** a spell's controller is its caster (CR 108.4), so "you've cast" ≡ "you control" for a spell; both route to `ControllerRef::You`. The trigger fires only for your own countered spell, never an opponent's.

CR annotations grep-verified: CR 701.6a (countering), CR 603.2 (trigger conditions), CR 108.4 (a spell's controller).

## Tests

- `trigger_a_spell_youve_cast_is_countered` / `trigger_a_spell_you_control_is_countered` — both forms parse to `Countered` with `valid_card = Controller(You)`, `valid_source = None`.
- `countered_trigger_valid_card_gates_own_spell` — **runtime**: your countered spell fires the trigger; an opponent's countered spell does NOT.

Fails-before (trigger → `Unknown`) / passes-after confirmed. Green: rustfmt, parser-combinator (Rule Zero) gate, `clippy -D warnings`, full engine suite (14546 passed / 0 failed), `cargo coverage` clean. The new arm is a whole-clause `all_consuming` match on a clause that previously parsed to `Unknown`, so no other card's parse is affected.

### DIFF
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -4678,6 +4678,73 @@ mod tests {
     }
 
     #[test]
+    fn countered_trigger_valid_card_gates_own_spell() {
+        // CR 701.6a + CR 108.4: Multani's Presence -- "Whenever a spell you've
+        // cast is countered". The trigger gates the COUNTERED spell via
+        // `valid_card = Controller(You)`, so it fires only when the countered
+        // spell's controller matches the trigger source's controller. Prove
+        // the chosen `You` filter is honored: an own countered spell fires,
+        // an opponent's countered spell does not.
+        let mut state = setup();
+        let source = create_object(
+            &mut state,
+            CardId(1),
+            PlayerId(0),
+            "Multani's Presence".to_string(),
+            Zone::Battlefield,
+        );
+        // A spell you control (owner/controller P0 == source controller).
+        let own_spell = create_object(
+            &mut state,
+            CardId(2),
+            PlayerId(0),
+            "Your Countered Spell".to_string(),
+            Zone::Stack,
+        );
+        // A spell an opponent controls (controller P1 != source controller).
+        let opponent_spell = create_object(
+            &mut state,
+            CardId(3),
+            PlayerId(1),
+            "Opponent's Countered Spell".to_string(),
+            Zone::Stack,
+        );
+        let countering_source = create_object(
+            &mut state,
+            CardId(4),
+            PlayerId(1),
+            "Some Counterspell".to_string(),
+            Zone::Battlefield,
+        );
+        let mut trigger = make_trigger(TriggerMode::Countered);
+        trigger.valid_card = Some(TargetFilter::Typed(
+            TypedFilter::default().controller(ControllerRef::You),
+        ));
+
+        // Your spell is countered -> trigger fires (regardless of who countered it).
+        let own_event = GameEvent::SpellCountered {
+            object_id: own_spell,
+            countered_by: countering_source,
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Approved current head `58bd95b2abad18558b08505b586a8d921a688336`.

Evidence checked: parse-diff is scoped to Multani's Presence (one card / two signatures), the parser-combinator gate passes locally, CI is green, and the implementation is at the existing `TriggerMode::Countered` seam. Hand trace: the parser maps "spell you've cast / spell controlled by you is countered" to `valid_card = controller(You)`, and `match_countered` applies `valid_card` to the countered stack object while preserving the separate `valid_source` check for the countering spell/ability. The added matcher tests cover own-spell vs opponent-spell behavior and the parser test would fail if the trigger remained unimplemented. CR references used here were verified against the local rules text.

e11734937-beep:
Hi @matthewevans — quick labeling question. The card implementations I've been landing (Regenerate, Role Reversal, Coastal Wizard, this countered-spell trigger, etc.) are coming through as `bug`. Since these add new card support rather than fix regressions, would `feature` — or `quality` for the larger ones — be the more appropriate label, or is `bug` intentional for coverage-completion work? Happy to follow whatever convention you prefer; just want to tag them right going forward. Thanks for the fast merges!

matthewevans:
> Hi @matthewevans — quick labeling question. The card implementations I've been landing (Regenerate, Role Reversal, Coastal Wizard, this countered-spell trigger, etc.) are coming through as `bug`. Since these add new card support rather than fix regressions, would `feature` — or `quality` for the larger ones — be the more appropriate label, or is `bug` intentional for coverage-completion work? Happy to follow whatever convention you prefer; just want to tag them right going forward. Thanks for the fast merges!

Heya! If your PRs are being misclassified please let me know. In the case of adding new _card_ support (wiring it up) that would be `enhancement` and is actually worth more than the `feature` label.

`quality` label is at my discretion and is for PRs that are submitted that effectively check all the boxes. Right architectural seam, elegant code, discriminating tests, little to no review churn. :)

Hope this helps!

e11734937-beep:
That's super helpful, thank you! Yes — these are all new-card wiring (Regenerate, Role Reversal, Coastal Wizard, the skip-step form, this countered-spell trigger, Valakut Fireboar), so `enhancement` fits. Would you be open to re-tagging the already-merged ones from `bug` → `enhancement` when you get a moment? And I'll aim for the `quality` bar (right seam, discriminating tests, minimal churn) on the ones that warrant it. Really appreciate the guidance!

PR #4907, opened 2026-07-02
Fix Riku, Of Many Paths triggering on every spell instead of only modal spells

### LINKED ISSUE #750: Riku, Of Many Paths triggering on warp spells — Riku, of Many Paths is specifically ruled to only trigger on instants/s…
Reported in Discord: https://discord.com/channels/1485498006781427802/1507069598036394014/1507069598036394014

**Thread:** Riku, Of Many Paths triggering on warp spells
**Cards:** _none detected_
**Parser status:** has_gaps

## Summary
Riku, of Many Paths is specifically ruled to only trigger on instants/sorcs that say something to the effects of "choose x".

---
<sub>report_id: `discord:1507069598036394014:1507069598036394014:0` · discord: `1507069598036394014/1507069598036394014`</sub>

GildardoDev:
**Anchored on:** #750

**Problem:** Riku, Of Many Paths — "Whenever you cast a modal spell, choose up to X…" — triggered on *every* spell its controller cast, not just modal spells.

**Root cause:** The cast-trigger parser could not express "modal spell" as a filter. In the "you cast a …" branch, the "modal" qualifier ran through `parse_type_phrase`, which doesn't recognize "modal" as a type, producing an empty filter; `has_meaningful_type_constraint()` was then false, so `valid_card` was left `None` and the qualifier was silently dropped. With no spell filter, the `SpellCast` trigger matched every cast.

**Fix:** Add `FilterProp::Modal` (CR 700.2) — an arity-0 intrinsic spell predicate alongside `HasSingleTarget`/`Foretold`. The parser now peels an optional leading "modal " qualifier (nom `opt(tag)`) before the type-phrase parse and attaches `FilterProp::Modal` to the filter's properties, so `valid_card` is assigned. At runtime it evaluates against the stack object's static printed modality (`obj.modal.is_some()`), which is present from object creation. The predicate is wired at every exhaustive `FilterProp` match site: the live evaluation, the two population classifiers and coverage render, conservative `false` at the two record-snapshot matchers (which carry no modal field), and the safe (apply-constant) group of the AI memoization classifier — with no new wildcard added. It is class-general (Riku plus Your Wish Is My Command, and any future "modal spell" reference), not a Riku special-case.

**Tests:** Three parser tests (Riku's trigger `valid_card` gains `FilterProp::Modal`; "modal instant spell" keeps the type *and* Modal; a plain "cast a spell" trigger gets no Modal) and two full-cast-pipeline runtime tests — Riku fires when a modal spell (Abrade) is cast and does **not** fire on a plain non-modal spell (Shock). Reverting the parser attach restores `valid_card == None` and makes Shock wrongly fire, so both directions are load-bearing.

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, and ./scripts/check-parser-combinators.sh all clean. Regression green: filter, spell_cast, filterprop, trigger, and coverage suites. The new variant is additive on the serialized surface (`{"type":"Modal"}`), so existing card data loads unchanged.

Closes #750.

Model: claude-opus-4-8
Tier: Frontier

### DIFF
--- a/crates/engine/src/ai_support/filter.rs
+++ b/crates/engine/src/ai_support/filter.rs
@@ -612,6 +612,9 @@ fn filterprop_reads_only_candidate_fp(p: &FilterProp) -> bool {
         | FilterProp::HasManaAbility
         | FilterProp::HasNoAbilities
         | FilterProp::Named { .. }
+        // CR 700.2: modality reads `obj.modal`, an apply()-constant printed-card
+        // field on the candidate — safe to memoize.
+        | FilterProp::Modal
         | FilterProp::IsCommander => true,
 
         // SAFE only when the embedded `QuantityExpr` is memo-safe.
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -873,6 +873,7 @@ fn fmt_typed_filter(tf: &TypedFilter) -> String {
                 ));
             }
             FilterProp::HasSingleTarget => parts.push("single target".into()),
+            FilterProp::Modal => parts.push("modal spell".into()),
             FilterProp::FaceDown => parts.push("face-down".into()),
             FilterProp::TargetsOnly { filter } => {
                 parts.push(format!("targets only {}", fmt_target(filter)));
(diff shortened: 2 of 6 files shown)


matthewevans (approved):
Maintainer sign-off: scoped engine/parser fix for modal spell cast triggers. Evidence checked: current parse-diff only changes Riku's SpellCast watch filter; implementation uses a typed FilterProp::Modal at the filter seam; parser and cast-pipeline tests discriminate the over-trigger regression; CI/card-data are green.

PR #4908, opened 2026-07-02
fix(parser+engine): Jabari's Influence — scope target to "attacked you this turn" + chain the dropped -1/-0 counter

real-venus:
## Summary

**Jabari's Influence** — "Gain control of target nonartifact, nonblack creature that attacked you this turn and put a -1/-0 counter on it." — parsed to a bare `Effect::GainControl`: it **dropped both** the "attacked you this turn" target restriction **and** the "-1/-0 counter" conjunct.

## Root cause (one coupled seam)

The target parser had **no arm for the defender-scoped "attacked you this turn"** relative clause, so `parse_target` stopped at "that" and left `"that attacked you this turn and put a -1/-0 counter on it"` as the remainder. `try_split_targeted_compound` keys its `" and "` boundary off that remainder — so the split point was wrong and only `GainControl` survived.

## Fix

- **Parameterize `FilterProp::AttackedThisTurn` → `AttackedThisTurn { defender: Option<ControllerRef> }`**, mirroring the existing `FilterProp::Attacking { defender }` sibling: `None` = board-wide (unchanged), `Some(You)` = "attacked you this turn". Evaluation reuses the existing per-defender ledger `creature_attacked_defenders_this_turn` via `attacking_defender_matches` — **no new engine state, no CR-incorrect board-wide over-approximation**.
- A new **permissive** target arm in `parse_that_clause_suffix` consumes "attacked you this turn" and deliberately (no terminator guard) leaves the trailing `" and put a -1/-0 counter on it"` intact — so the existing compound splitter **auto-chains `PutCounter(-1/-0, ParentTarget)`**, fixing the counter drop for free.

## Notes

- **serde** is internally-tagged, so `defender: None` serializes byte-identically to the old unit variant → **legacy card-data unaffected**.
- `mtgish-import` carries the same mechanical `{ defender: None }`/`{ .. }` compile-fixes the `Attacking` parameterization already required (workspace clippy builds it).
- Opponent scope is eval/render-supported but parser-deferred (no card needs it yet) — honest gap, not silent.

## Testing

Discriminating tests, both defects **revert-failing**: full-card `parse_effect_chain` (GainControl target carries `AttackedThisTurn{Some(You)}` + chained `PutCounter{PowerToughness{-1,0}, ParentTarget}`); a **3-player runtime** test (a creature that attacked *another* player is NOT a legal target); plus board-wide/negation regression + serde-shape tests. All affected engine suites green.

CR 508.1a + CR 508.1b + CR 508.6 + CR 122.1.


matthewevans (changes requested):
**[MED]** The fix still needs a production-path regression test for the combined Jabari's Influence behavior. Evidence: the added coverage proves the parser shape (`GainControl` gains `AttackedThisTurn { defender: Some(You) }` and a `PutCounter` sub-ability) and directly tests the filter matcher, but it does not cast/resolve the parsed spell through the real action pipeline. Why it matters: this PR claims to fix both target legality and the chained `-1/-0` counter; parser-shape tests would still pass if the target prompt, parent-target binding, or resolution chain regressed after parsing. Suggested fix: add a `GameScenario`/cast-resolution test for Jabari's Influence in a multiplayer or defender-distinguishing setup: one creature attacked the caster, another attacked only a different player, only the former is legal/selected, control changes, and the `-1/-0` counter lands on that same gained-control creature.

I checked Gemini's CR annotation note against the current head; the relevant zone-change branch already has a `CR 400.7` comment immediately above `FilterProp::AttackedThisTurn { .. }`, so I am not treating that as a remaining blocker.

matthewevans (changes requested):
**[MED]** Defender-scoped attack history is not threaded through the mtgish converter. Evidence: `crates/mtgish-import/src/schema/types.rs:5174` already models `Permanents::AttackedPlayerThisTurn(Box<Player>)`, `data/mtgish-cards.json:14206` uses that exact source shape for Jabari's Influence with `_Player: "You"`, but `crates/mtgish-import/src/convert/filter.rs:506` still falls through to `EnginePrerequisiteMissing` instead of producing `FilterProp::AttackedThisTurn { defender: Some(ControllerRef::You) }`. Why it matters: this PR adds the right engine/parser building block but leaves the parallel card import path unable to represent the same card/class. Suggested fix: add a converter arm for `Permanents::AttackedPlayerThisTurn(player)` at least for the `You` axis, mirroring the existing player-scope conversion patterns, and keep unsupported axes strict-failing rather than erasing the defender.

matthewevans:
Holding approval on current head `716805192ee51718fc8fe5c3859a273df1348adb`: the prior Jabari-specific blocker appears addressed in this pass, including the mtgish converter threading and runtime coverage, but the current parse-diff evidence is still not acceptable for approval.

The sticky still reports unrelated Riku `SpellCast.watches` and Valakut `SwitchPT` changes, while the local diff for this repush is limited to the Jabari/attacked-you path. Parser/engine PRs need current card-level parse-diff evidence with explained blast radius before merge-when-ready. Please update/rebase and rerun coverage parse-diff until the sticky reflects only intended Jabari-scope changes, or explicitly justify each unrelated signature.

matthewevans (commented):
Holding approval on current head `716805192ee51718fc8fe5c3859a273df1348adb`: the prior Jabari-specific blocker appears addressed in this pass, including the mtgish converter threading and runtime coverage, but the current parse-diff evidence is still not acceptable for approval.

The sticky still reports unrelated Riku `SpellCast.watches` and Valakut `SwitchPT` changes, while the local diff for this repush is limited to the Jabari/attacked-you path. Parser/engine PRs need current card-level parse-diff evidence with explained blast radius before merge-when-ready. Please update/rebase and rerun coverage parse-diff until the sticky reflects only intended Jabari-scope changes, or explicitly justify each unrelated signature.

matthewevans (approved):
Approved after current-head review. The defender-scoped `AttackedThisTurn { defender: Option<ControllerRef> }` parameterization is scoped to the attack-history axis, mtgish now maps Jabari’s `AttackedPlayerThisTurn(You)` to the defender-scoped filter, and runtime coverage exercises cast, target legality, control change, and the chained -1/-0 counter. The non-Jabari parse-diff entries are stale baseline contamination rather than current PR blast radius.

real-venus:
@matthewevans 

**CI blocked by a Rust-test-shard timeout, not a test failure — requesting an admin-merge or a shard-timeout bump.**

This PR is approved, conflict-free with current `main`, and green on every check except the `Rust ... coverage-gate` aggregator — which fails only because both `Rust tests` shards hit the **15-min `timeout-minutes` cap** (`ci.yml:92`):

- Shard logs show `Finished test profile in ~6m` → `Starting ~9,200 tests` → `Error: The operation was canceled`. **No test fails** (nextest reports failures by name); it's purely the 15-min wall.
- Compile alone eats ~6 min, leaving <9 min for the suite. It won't self-heal on re-run: `rust-cache` only saves on a *successful* job, so a timeout means the cache never warms for this branch → every run is a cold ~6-min compile → times out again.

This looks like a repo-wide capacity issue that grew past the budget around v0.13.0, which this branch inherits by being up to date with `main`.

Could you either **admin-merge** (the code's approved and no test actually fails), or **bump `timeout-minutes`** / **split the shards 2→4**? 
Happy to do whatever's cleanest on my end. Thanks!

PR #4909, opened 2026-07-02
fix(parser): parse Eidolon's distributive dual-subject dynamic pump

### LINKED ISSUE #4777: Eidolon of Countless Battles as a creature — When cast as a creature, [[Eidolon of Countless Battles]] immediately get…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521826810084462632/1521826810084462632

**Thread:** Eidolon of Countless Battles as a creature
**Cards:** eidolon of countless battles

## Summary
When cast as a creature, [[Eidolon of Countless Battles]] immediately get moved to the graveyard as it has 0 toughness, while I'm pretty sure that the

### eidolon of countless battles
Bestow {2}{W}{W} (If you cast this card for its bestow cost, it's an Aura spell with enchant creature. It becomes a creature again if it's not attached.)
This creature and enchanted creature each get +1/+1 for each creature you control and +1/+1 for each Aura you control.

---
<sub>report_id: `discord:1521826810084462632:1521826810084462632:0` · discord: `1521826810084462632/1521826810084462632`</sub>

nickmopen:
## Problem (#4777)

Eidolon of Countless Battles' static —

> This creature and enchanted creature each get +1/+1 for each creature you control and +1/+1 for each Aura you control.

— was mis-parsed as a **fixed `+1/+1` `Pump`** ability. Two independent gaps combined:

1. The distributive **"each"** after the multi-subject list (`this creature and enchanted creature each get …`) broke `contextual_continuous_subject_filter`, so the whole static failed and fell back to the imperative pump path.
2. Even with the subject fixed, the **repeated** `+N/+M for each X and +N/+M for each Y` pump collapsed to a fixed value — `parse_continuous_gets_has` splits on the first `for each` only, and the second term was silently dropped (the `DynamicQty` swallow the parser flagged).

## Fix (both in the clean `oracle_static/anthem.rs`)

- Strip the distributive `" each"` marker in `contextual_continuous_subject_filter` so the compound subject unions into `Or { SelfRef, <enchanted host> }`. No single subject legitimately ends in `" each"`, so it only fires on the distributive form.
- Add `parse_repeated_for_each_pt_modifications`, which splits on the `" and "` that introduces another `+n/+m for each` pump term (leaving a single term's embedded `for each A and B` count-list to the existing single-term path) and accumulates each term's dynamic modifications via the existing `parse_dynamic_for_each_pt_modifications`. Returns `None` unless ≥2 whole pump terms parse, so every non-repeated case is untouched.

## Result

Eidolon now parses to one continuous static on `{source, enchanted host}` with dynamic `+1/+1` per creature you control **and** `+1/+1` per Aura you control. The spurious `Pump` ability and the `DynamicQty` swallow warning are both gone.

## Testing

- New regression test `static_self_and_enchanted_each_repeated_dynamic_pump`.
- Full engine lib suite green: `cargo test -p engine --lib` → **14,551 passed, 0 failed**.
- `cargo fmt -p engine --check` clean.

Closes #4777

### DIFF
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -580,6 +580,16 @@ pub(crate) fn contextual_continuous_subject_filter(
     subject_original: &str,
     condition: Option<&StaticCondition>,
 ) -> Option<TargetFilter> {
+    // CR 613.1: a distributive "each" trailing a multi-subject list ("this
+    // creature and enchanted creature each get +1/+1 …", Eidolon of Countless
+    // Battles) shares one predicate across every named subject. Strip the marker
+    // from both views so the compound subject parses; the `Or` union below
+    // already applies the predicate to each branch. Stripping each view with the
+    // same combinator keeps them aligned without slicing one by the other's byte
+    // length (safe when case-folding changes width).
+    let subject_lower = strip_trailing_distributive_each(subject_lower);
+    let subject_original = strip_trailing_distributive_each(subject_original);
+
     if subject_lower == "that creature" {
         return condition
             .and_then(exactly_one_creature_you_control_filter)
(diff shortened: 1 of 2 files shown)


nickmopen:
Thanks for the review — addressed all three points and rebased onto latest `main`:

- **[HIGH] `anthem.rs` subject slice** — removed the `subject_original[..subject_lower.len()]` slice entirely. The distributive-`each` strip now runs the same combinator (`strip_trailing_distributive_each`, a `take_until(" each")`/`tag`) over each view independently, so the two stay aligned without ever slicing one by the other's byte length (safe when case-folding changes width).
- **[MEDIUM] R1 / string-method dispatch** — converted all parsing dispatch to nom combinators: `segment_starts_pump_term` is now `preceded(opt(alt((tag("gets "), tag("get ")))), preceded(parse_pt_modifier, tag(" for each ")))`, the `" and "` split is a `take_until`/`tag` scan (`split_on_and`), and the verb check is `segment_has_gets_verb`. `scripts/check-parser-combinators.sh` passes clean.

Local verification: `cargo fmt --all --check`, `cargo clippy -p engine --all-targets -- -D warnings`, `cargo test -p engine --lib` (14,551 passed / 0 failed), and the parser-combinator gate all green.

matthewevans (approved):
Reviewed current head `cfe9311b94a125cd24865a9c4a2d52889bc668bc`. The parser evidence is now current and scoped: the coverage parse-diff is limited to Eidolon of Countless Battles and Nighthowler, matching the distributive dual-subject dynamic-pump change, and CI is green. I do not see a remaining blocker.

PR #4910, opened 2026-07-02
feat(pr-review): contributor intelligence + transient-failure hardening for pr_review.py

matthewevans:
Rework scripts/pr_review.py around the JSONL event log as sole canonical
store and add an advisory contributor block to every packet and recommend
output: standing (private-overrides.json contributor_standing or derived
trusted), scrutiny (normal|elevated|maintainer_attention) with reasons,
windowed signal recurrence (60d, PR-attributed signals only), and
first-contribution detection. Only a standing override of 'skip' changes
advisory_action, ordered after hard_stop so safety wins.

record validates optional per-review 'signals' against
QUALITY_SIGNAL_WEIGHTS (single authority); all login grouping is
case-folded via fold_login across analytics, packets, and compact.
parse_diff sticky-comment classification requires the github-actions
author and rides on packets for engine-surface review evidence.

run_json now retries transient gh/GraphQL failures (3 attempts, backoff)
and surfaces gh's captured stderr on final failure — the expensive scan
query intermittently times out server-side and a single blip was killing
whole review sweeps.

Skills updated to consume the contributor block (pr-review-loop sweep
protocol, pr-contribution-handler standing gauge, review-impl parse-diff
evidence gate). 36/36 tests.

### DIFF
--- a/.claude/skills/pr-contribution-handler/SKILL.md
+++ b/.claude/skills/pr-contribution-handler/SKILL.md
@@ -18,6 +18,7 @@ This skill lands contributor work, but **only after it meets the maintainer's ba
 2. **CI/Tilt green is necessary, NOT sufficient.** Green CI proves it compiles and existing tests pass. It does **not** prove correctness, no-regression, or performance. Never present "CI green" as evidence a PR is ready. For every PR you must additionally:
    - **Trace the changed logic by hand**, end to end, for the target case AND 2–3 sibling cases in the class AND the obvious edge cases (multiplayer, zero/empty, interaction with existing effects). Confirm it actually produces the rules-correct result — not merely that it "conforms to CLAUDE.md."
    - **Verify the tests DISCRIMINATE.** For every assertion, ask: *would this fail if the fix were reverted?* An assertion that passes both before and after the fix is coverage theater. A bug-fix PR must have at least one runtime test that drives the engine through the real pipeline (`apply()` / scenario runner) and would fail without the change. Name any behavior with no discriminating test as a gap, and add the missing test before enqueue.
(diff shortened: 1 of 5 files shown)


matthewevans:
Re: gemini-code-assist review — dispositions:

**Accepted (landing in an immediate follow-up PR — this PR is in the merge queue and a push would drop it):**
- `event.get("quality", {})` null-guards (×2, `build_analytics_model` + `command_compact`): fixed as `(event.get("quality") or {})`, matching the existing `contributor_login_for_event` pattern; regression test added for explicit `"quality": null`.
- `result.get("data", {})` GraphQL null-guards (×3, `gh_pr_refresh_chunk` / `fetch_open_prs` / `gh_pr_view`): fixed with `or {}` chains. Largely shielded today because `gh` exits non-zero on GraphQL errors and `run_json` raises, but it's an external boundary and the guard is free.
- Shared `LOCK_SH` in `all_events`: fixed. It completes the locking protocol with `append_event`'s `LOCK_EX` so a reader can't observe a partially flushed line.

**Declined:**
- Optional `fcntl` import + `if fcntl:` guards: making the lock optional silently drops the multi-agent append-atomicity guarantee on the canonical event log — corrupting `review-events.jsonl` quietly on an unlocked platform is worse than failing loudly with `ImportError`. "Portable" in the docstring means portable across repos/hosts (state dir outside the repo), and review hosts are POSIX. If Windows support is ever needed, the right change is a real lock (`msvcrt.locking` or a lockfile), not no lock.

matthewevans:
Correction to the disposition note above: the accepted fixes were pushed to this PR directly (4eca0cee22) rather than a follow-up — includes the null-guards, the `LOCK_SH` read lock, a regression test for explicit `"quality": null`, and a no-behavior-change structure pass (constants consolidated, `ReviewContext` extraction, section map). The two `fcntl`-optional suggestions remain declined per the rationale above. 37/37 tests.

PR #4914, opened 2026-07-02
Parse the it-entered-this-turn static condition (Crew Captain, Drownyard Behemoth, ...)

GildardoDev:
**Problem:** Four cards with an "as long as it entered this turn" static gate — Crew Captain ("has indestructible as long as it entered this turn"), Drownyard Behemoth, Thrasta, Tempest's Roar, and Zurgo and Ojutai (all "has hexproof as long as it entered this turn") — dropped that condition to `StaticCondition::Unrecognized`, so the keyword grant never turned on.

**Root cause:** The entered-this-turn arm in `parse_source_state_conditions` (parser/oracle_nom/condition.rs) hardcoded the `~` subject (`tag("~ entered this turn")`) and never normalized the bound pronoun "it", so `it entered this turn` matched no arm and fell through to `Unrecognized`.

**Fix:** Extracted a class-general `parse_source_entered_this_turn` that composes the existing `parse_self_source_subject` helper — which already accepts `it`/`~`/`this creature`/the card name — with the entered predicate, emitting the **existing** `StaticCondition::SourceEnteredThisTurn`. This mirrors the sibling `parse_source_hasnt_dealt_damage` pattern, is strictly more general than the old `~`-only arm (so canonical `~ entered ...` templating still parses), and requires zero engine change and no new variants. `SourceEnteredThisTurn` is already engine-evaluated and bridged to `TriggerCondition::SourceEnteredThisTurn`, so all four cards now correctly gain their keyword only on the turn they entered.

**Tests:** parser unit tests asserting `it entered this turn` (and the long "entered the battlefield this turn" form) resolve to `SourceEnteredThisTurn`, plus a full-card static-line test for Crew Captain's indestructible and the shared hexproof shape — both revert-failing (reverting the arm restores `Unrecognized`). The existing `~`/battlefield/"this &lt;type&gt;" forms stay green.

**Verification:** cargo fmt, cargo check -p engine, cargo clippy -p engine --lib -D warnings, and ./scripts/check-parser-combinators.sh all clean. Regression green: entered_this_turn (15), source_state (3), the full condition suite (359) and static suite (896). No new engine variants.

Model: claude-opus-4-8
Tier: Frontier


matthewevans (changes requested):
**[HIGH]** Bare `it entered this turn` is accepted in the context-free condition parser, bypassing the existing self-vs-attached pronoun guard. Evidence: `crates/engine/src/parser/oracle_nom/condition.rs:1727` accepts every `parse_self_source_subject` form as `SourceEnteredThisTurn`, while `crates/engine/src/parser/oracle_static/shared.rs:1534` calls that parser before the static-specific fallback, and `crates/engine/src/parser/oracle_static/anthem.rs:764` intentionally rewrites `it` to source only when `affected == SelfRef`. Why it matters: attached-subject statics like `Enchanted creature ... as long as it entered this turn` would now bind `it` to the Aura/Equipment source instead of the enchanted/equipped creature, turning an honest gap into wrong coverage. Suggested fix: keep bare `it entered...` out of the context-free grammar and rewrite it to source only in the existing `TargetFilter::SelfRef` static-condition path, with a negative attached-subject regression test.

GildardoDev:
Good catch, you're right that binding bare "it" context-free would mis-cover attached-subject auras. I pulled the bare-"it" handling out of the context-free grammar and instead rewrite it to source only on the existing TargetFilter::SelfRef path, so an Aura's "as long as it entered this turn" stays an honest gap. Added a negative attached-subject regression test asserting that case stays Unrecognized.

matthewevans (approved):
Approved after current-head review. The previous bare-pronoun blocker is fixed: context-free condition parsing no longer accepts bare `it entered this turn`, the `it entered...` rewrite is limited to the SelfRef static path, and attached-subject Aura/Equipment cases remain an honest `Unrecognized` gap. Parse-diff is scoped to intended source-entered-this-turn static conditions.

PR #4906, opened 2026-07-02
Add Valakut Fireboar: switch its power and toughness (pronoun form)

e11734937-beep:
## Summary
Adds engine support for **Valakut Fireboar** (ZEN) — "Whenever this creature attacks, switch its power and toughness until end of turn." — by teaching the effect parser the pronoun surface form **"switch its power and toughness"**.

`Effect::SwitchPT` and its layer-7d runtime already exist (Invert, Inversion Behemoth are the supported siblings). The parser recognized two surface forms — prepositional "switch the power and toughness of `<target>`" and possessive "switch `<target>`'s power and toughness" — but not the bare pronoun "switch **its** power and toughness". `parse_target` does not treat a bare "its" as a target, so both branches missed it and the trigger effect fell to `Effect::Unimplemented`; the P/T swap never fired.

The fix adds a third branch that matches "switch its power and toughness" and resolves the "its" anaphor through the shared `resolve_it_pronoun(ctx)` helper — the same anaphor resolution already used by the sacrifice/counter clauses. For Valakut Fireboar the trigger subject is this creature, so "its" resolves to `SelfRef` (the source switches its own P/T). The change is generic: any effect written as "switch its power and toughness" now lowers to `Effect::SwitchPT`.

## Files changed
- `crates/engine/src/parser/oracle_effect/imperative.rs` — pronoun branch in the SwitchPT parser
- `crates/engine/src/parser/oracle_effect/tests.rs` — parser unit test `effect_switch_pt_pronoun_its`
- `crates/engine/tests/valakut_fireboar_switch_pt_on_attack.rs` — integration test

## Anchored on
- `crates/engine/src/parser/oracle_effect/imperative.rs` (~4516) — the existing prepositional and possessive SwitchPT branches; this adds the pronoun sibling immediately before the possessive one.
- `crates/engine/src/parser/oracle_effect/mod.rs:172` — `resolve_it_pronoun` already maps "it/its" to `SelfRef` or `TriggeringSource` from `ctx.subject`; reused verbatim.
- `crates/engine/src/game/effects/switch_pt.rs` — `Effect::SwitchPT` runtime (layer-7d `SwitchPowerToughness`); unchanged. Invert and Inversion Behemoth exercise this handler today.

## CR references
- `CR 613.4d` — switching a creature's power and toughness is a single continuous effect applied in layer 7d.
- `CR 608.2c` — the triggered ability's instruction resolves as written; "its" refers to the object established by the trigger (here, the attacking creature).

## Track
Developer

[...]

PR #4812, opened 2026-07-01
Add Identity Crisis

e11734937-beep:
## Summary
Adds engine support for **Identity Crisis** — "Exile all cards from target player's hand and graveyard."

The card is a mass exile whose origin is a *union* of the targeted player's hand
and graveyard (CR 400.7). The parser's generic mass-exile path infers a **single**
origin zone via `infer_origin_zone`, which captured only the first zone ("hand")
and left the trailing "and graveyard" to be mis-split by `try_split_targeted_compound`
into an orphaned, unsupported "graveyard" child clause — so graveyard cards were
never exiled.

The fix mirrors the repo's existing multi-zone-origin recognizer for name-hate
exile (`try_parse_multi_zone_same_name_exile` → `MultiZoneSameNameExile`, used by
Deadly Cover-Up / Lost Legacy / Eradicate): a dedicated combinator recognizes the
`cards from <possessive> <zone> and <zone>` grammar and encodes the zone union on
the target filter via `FilterProp::InAnyZone { zones }`, with `origin: None` so the
already-existing `ChangeZoneAll` resolver reads the multi-zone origin from the
filter. No game-layer/resolver changes — the runtime path is the one Lost Legacy
already exercises. The recognizer requires ≥2 zones, so single-origin exiles
(e.g. Haunting Echoes, "exile all cards from target player's graveyard") keep
taking the unchanged generic path.

## Files changed
- `crates/engine/src/parser/oracle_effect/imperative.rs`
- `crates/engine/src/parser/oracle_effect/mod.rs`
- `crates/engine/src/parser/oracle_effect/tests.rs`
- `crates/engine/tests/identity_crisis_multi_zone_exile.rs`

## Anchored on
- `crates/engine/src/parser/oracle_effect/imperative.rs:2171` — `try_parse_multi_zone_same_name_exile`: the existing dedicated multi-zone-origin exile recognizer (Deadly Cover-Up / Lost Legacy). My `try_parse_multi_zone_player_exile` follows the same `alt((tag(...)))` owner-axis + zone-list structure.
- `crates/engine/src/parser/oracle_effect/imperative.rs:2742` — `MultiZoneSameNameExile` lowering: constructs `Effect::ChangeZoneAll { origin: None, target: Typed(... InAnyZone { zones } ...), destination: Exile }`. My lowering builds the identical shape (minus the `SameNameAsParentTarget` name constraint).
[...]

PR #4917, opened 2026-07-02
Recognize "there are no [type] on the battlefield" condition (Sarcomancy, Call to the Grave)

e11734937-beep:
## Summary

Triggers/conditions of the form **"there are no [type] on the battlefield"** were dropped — the intervening-`if` parsed to `condition: null`, so cards like **Sarcomancy** ("At the beginning of your upkeep, if there are no Zombies on the battlefield, ~ deals 1 damage to you"), **Drop of Honey**, **Porphyry Nodes**, and **Task Mage Assembly** ("when there are no creatures on the battlefield, sacrifice ~") lost their gating condition.

The engine already lowered the subject-first form **"no [type] are on the battlefield"** (Call to the Grave) to `ObjectCount(<filter>) == 0`; only the existential *there*-form was missing.

## Change

Extend `parse_no_on_battlefield` (`crates/engine/src/parser/oracle_nom/condition.rs`) to recognize both surface forms with one shared output:

- `"there are no <type> on the battlefield"` (existential there-form) — **new**
- `"no <type> are on the battlefield"` (subject-first) — existing

Both lower to `StaticCondition::QuantityComparison { ObjectCount(<filter>) == 0 }`. A prefix `alt` selects the surface form and its trailing anchor; `<type>` is parsed via the shared `parse_type_phrase`, so a subtype ("Zombies"), a card type ("lands"), or a token phrase ("Reflection tokens") scopes the emptiness check exactly. No controller restriction — "no creatures" counts *any* player's matching permanents (CR 603.4 intervening-`if`; CR 110.1).

Nom combinators only (`alt`/`value`/`tag`/`take_until`); no new condition type.

## Tests

Added recognizer tests covering: the there-form ("there are no creatures / Zombies on the battlefield"), the subject-first form, card-type ("no lands"), the battlefield anchor requirement, and a guard that "there are no" is not confused with a count condition. All assert the `ObjectCount == 0` lowering with the correct filter and no controller restriction.

Fails-before (there-form → dropped condition) / passes-after confirmed. Green: rustfmt, parser-combinator (Rule Zero) gate, `clippy -D warnings`, full engine suite, and `cargo coverage` (net gain, 0 regressions).


matthewevans (approved):
Maintainer review: parser condition fix is scoped and idiomatic. It extends the existing battlefield-absence condition parser rather than adding a sibling shape, keeps the output as the existing ObjectCount == 0 predicate, and the parse-diff is limited to the expected Sarcomancy / Task Mage Assembly condition gains.

e11734937-beep:
Hi @matthewevans — quick labeling note on this one: it landed tagged only `bug`, but it adds a new parser capability (recognizing "there are no [type] on the battlefield" as a condition — Sarcomancy, Call to the Grave), which seems to be the same class as the card/condition adds you tagged `enhancement`/`quality` (e.g. #4812, #4900, #4906). Would `enhancement` be the right fit here? Totally your call — thanks for the guidance on labeling earlier.

e11734937-beep:
Hello @matthewevans,

Thanks for merging these, and for the careful reviews. A calibration question on the `quality` label so I can aim future PRs at it: the triage applied it to some of these minimal-churn parser fixes (#4812, #4903) but not to others of similar scope — e.g. #4804 (+65/−4, definite-article reveal), #4839 (+82/−0, shared-quality), and this one (#4917, a focused condition recognizer). Per the label's description ("high-quality minimal to no-churn PRs") they read similar to me.

Is there a distinction I'm missing, or just triage variance? Happy to match whatever bar you'd prefer. Thanks!

PR #4918, opened 2026-07-02
fix(parser): recognize filtered attacker presence conditions (Nemesis Trap, Slingbow Trap)

### LINKED ISSUE #4916: fix(parser): recognize filtered attacker presence conditions (Nemesis Trap, Slingbow Trap)
## Problem

`parse_creature_attacking_you` (`crates/engine/src/parser/oracle_nom/condition.rs`) only recognizes the bare literal `"a creature is attacking you"`. Any qualifier in front of "creature" — a color, a subtype, or a "with `<keyword>`" clause — falls through unrecognized, so the whole condition is dropped and the clause it gates becomes `Unimplemented`:

- **Nemesis Trap** — "If a white creature is attacking, you may pay {B}{B} rather than pay this spell's mana cost."
- **Slingbow Trap** — "If a black creature with flying is attacking, you may pay {G} rather than pay this spell's mana cost."
- **Charging Hooligan** — "...If a Rat is attacking, this creature gains trample until end of turn."

The sibling combinator `parse_creatures_are_attacking_count_ge` only covers the *numeric* "N or more creatures are attacking" form (Lethargy Trap), and it's unfiltered too — neither existing arm covers a *filtered, singular* existential attacker check.

## Proposed fix

Generalize `parse_creature_attacking_you` into a `parse_type_phrase`-driven combinator — the same shape as the already-generalized `parse_creature_has_keyword` ("a `<filter>` creature has `<keyword>`") a few functions below it — that accepts an optional type/color/keyword qualifier ahead of "is attacking[ you]" and folds it into the `IsPresent` filter via the existing `add_filter_property` helper, instead of hardcoding `TypedFilter::creature()`.

Because `parse_inner_condition` is the single shared condition authority for the engine, fixing it there should reach every caller that gates on it through the existing bridges: static "as long as" conditions, casting-restriction / alternative-cost leading-`if` gates (`oracle_casting.rs`), and trigger intervening-if — without touching any of those call sites.

## CR

CR 508.1 (declaring attackers) + CR 105.1 (color) + CR 702.9 (flying) + CR 118.9 (alternative costs).

## Class

Filtered attacker-presence conditions: any card gating a static, cost, or trigger on "a `<qualified>` creature is attacking[ you]" rather than the bare/unfiltered form. Confirmed against the local MTGJSON snapshot: at least the Zendikar Trap cycle (Nemesis Trap, Slingbow Trap) plus Charging Hooligan.

jsdevninja:
## Summary

- Closes #4916.
- Generalizes `parse_creature_attacking_you` (`crates/engine/src/parser/oracle_nom/condition.rs`) into `parse_filtered_creature_is_attacking`, which delegates the attacker qualifier to `parse_type_phrase` (mirroring the already-generalized `parse_creature_has_keyword`) instead of hardcoding `TypedFilter::creature()`.
- Unlocks "a `<filter>` creature is attacking[ you]" for any type/color/keyword qualifier — previously only the bare literal "a creature is attacking you" parsed.
- Because `parse_inner_condition` is the single shared condition authority, this reaches every caller that gates on it (static "as long as" conditions, casting-restriction / alternative-cost leading-`if` gates, trigger intervening-if) without touching any of those call sites.
- Real cards unlocked: **Nemesis Trap** ("If a white creature is attacking, you may pay {B}{B} rather than pay this spell's mana cost."), **Slingbow Trap** ("If a black creature with flying is attacking, you may pay {G} rather than pay this spell's mana cost."), and the intervening-if half of **Charging Hooligan**.
- Existing bare-form behavior is preserved: Confront the Assault / Swat Away / Heroic Return ("a creature is attacking you") and Fated Clash / Qasali Ambusher ("a creature is attacking") still resolve to the same `IsPresent` shape as before.

CR annotations grep-verified against `docs/MagicCompRules.txt`: CR 508.1 (declaring attackers), CR 105.1 (color), CR 702.9 (flying), CR 118.9 (alternative costs).

## Tests

- `oracle_nom/condition.rs`: `test_filtered_creature_is_attacking_color` (Nemesis Trap's bare condition), `test_filtered_creature_is_attacking_color_and_keyword` (Slingbow Trap's bare condition) — both assert the `IsPresent` filter carries the color/keyword `FilterProp` plus `Attacking { defender: None }`.
- `oracle_casting.rs`: `alt_cost_leading_if_filtered_attacking_creature_color_binds` and `alt_cost_leading_if_filtered_attacking_creature_color_and_keyword_binds` — full leading-`if` alternative-cost lines for Nemesis Trap and Slingbow Trap, asserting the `SpellCastingOption` now carries a `QuantityComparison(ObjectCount(filter) >= 1)` condition instead of being dropped.
- Existing `test_a_creature_is_attacking_you` (bare form) is unchanged and should still pass, confirming no regression on the previously-supported cards.

### DIFF
--- a/crates/engine/src/parser/oracle_casting.rs
+++ b/crates/engine/src/parser/oracle_casting.rs
@@ -742,6 +742,7 @@ mod tests {
         ControllerRef, FilterProp, ParsedCondition, PlayerFilter, QuantityExpr, QuantityRef,
         TargetFilter, TypeFilter,
     };
+    use crate::types::keywords::Keyword;
     use crate::types::mana::{ManaColor, ManaCost};
     use crate::types::zones::Zone;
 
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[MED]** Attached-subject attacker exclusions are incomplete. Evidence: `crates/engine/src/parser/oracle_nom/condition.rs:2992` only rejects `equipped creature` / `enchanted creature`, while `parse_attached_condition_subject_core` also owns `enchanted permanent`, `enchanted artifact`, and `enchanted land` at `crates/engine/src/parser/oracle_nom/condition.rs:797`. Why it matters: those broader self-referential Aura-host phrases can still fall through `parse_type_phrase` and become board-wide `IsPresent(... + Attacking)` checks instead of staying on the attached-subject seam. Suggested fix: reject all attached-subject combat prefixes handled by `parse_attached_subject_combat_state`, and add negative tests for the non-creature enchanted subjects.

matthewevans (approved):
Reviewed current head: the attached-subject follow-up is fixed at the shared condition seam, tests cover the filtered attacker cases and the non-creature attached-subject exclusions, and current-head parser/card-data checks passed.

PR #4924, opened 2026-07-02
Add end-to-end regression tests for copy-retarget across intervening clauses

GildardoDev:
## Summary
The implementation this PR originally proposed has since landed on main via #5010 (binding "choose new targets for the copy/copies" across intervening clauses, plus multi-copy patching). Rather than duplicate that work, this PR is now rebased onto main and reduced to the end-to-end regression coverage for the copy-retarget building block, complementing #5010's in-module unit tests by driving the full parser through `parse_oracle_text`.

## Tests (crates/engine/tests/copy_retarget_past_rider.rs)
- Narset's Reversal — a Bounce rider ("... then return it to its owner's hand") between the copy and the retarget clause; the copy still gains `MayChooseNewTargets` and the rider is preserved.
- Spinerock's Tyrant — a static-grant rider ("those spells gain wither") between the copy and the retarget clause.
- Increasing Vengeance — the flashback "copy that spell twice instead" nests a second `CopySpell`; both the base copy and the instead copy gain `MayChooseNewTargets`.
- Twincast — no-rider control (copy immediately followed by the retarget clause) stays correct.

All four pass against current main (#5010's implementation). CR 707.10c.

## Gate A
```
(exit 0 — no violations)
```

Model: claude-opus-4-8
Tier: Frontier


matthewevans:
### Parse changes introduced by this PR · 3 card(s), 1 signature(s)  (baseline: main `7ff76715a824`)

#### 3 card(s) · ability/orphaned_copy_retarget · removed: `orphaned_copy_retarget`
Examples: Increasing Vengeance, Narset's Reversal, Spinerock Tyrant

matthewevans (changes requested):
**[MED]** Conditional “copy twice instead” branches keep `KeepOriginalTargets`. Evidence: `crates/engine/src/parser/oracle_effect/sequence.rs:2829` returns after patching the root `CopySpell`, so it never descends into a `CopySpell` under `sub_ability`; `crates/engine/src/parser/oracle_effect/lower.rs:1193`-`1208` stores generic instead overrides as a root effect plus conditional sub-ability, and `crates/engine/tests/integration/throw_instead_tail_class.rs:464` shows the parse-diff affected `Increasing Vengeance` shape. Why it matters: when Increasing Vengeance is cast from a graveyard, the two replacement copies should also allow new targets, but this PR only patches the base copy path. Suggested fix: make the copy-retarget patch recurse through all reachable copy branches, including conditional instead `sub_ability`/`else_ability`, and add an Increasing Vengeance parser assertion covering both branches.

GildardoDev:
Fixed both. `set_copy_retarget_in_ability` now patches every reachable copy: it descends the `sub_ability` chain and the `else_ability` branch and accumulates instead of returning after the first match, and the apply-site backward scan no longer breaks after the first copy-bearing def (non-copy riders still return false and are skipped). So Increasing Vengeance's flashback "copy that spell twice instead" copy — nested in the conditional `sub_ability` — now gains `MayChooseNewTargets` alongside the base copy, and any card that makes multiple copies retargets all of them.

Added `increasing_vengeance_retargets_both_base_and_instead_copies` asserting both `CopySpell` branches carry `MayChooseNewTargets` (fails before the fix, when only the base copy did). Rebased on latest main; full engine suite + clippy green.

matthewevans:
Re-reviewed current head `f01724de`. The code-side blocker from my prior review appears addressed: the retarget patch now walks every reachable copy branch (`effect`, `sub_ability`, and `else_ability`), and the Increasing Vengeance regression asserts both the base copy and the conditional "copy twice instead" copy carry `MayChooseNewTargets`.

Holding approval for required parser evidence only: the parse-diff sticky is still updated at `2026-07-02T17:48:31Z`, but the current head commit is from `2026-07-03T08:09:56Z`. Because this is parser-surface work, I need a parse-diff for the current head before clearing the stale changes-requested state or enqueueing.

matthewevans:
Reviewed current head again. I am keeping this blocked because the same parser seam has since landed on `main` with more precise plural/singular copy-retarget handling, and this branch now conflicts with current `main`. Please do not rebase this implementation as-is; if there is unique test value here, it should be ported onto the mainline implementation instead.

GildardoDev:
Heads up on the scope change: the parser fix this PR proposed landed independently via #5010, so I've rebased onto main and pared this down to just the end-to-end integration regression tests (they drive parse_oracle_text for Narset's Reversal, Spinerock's Tyrant, Increasing Vengeance's two copies, and the Twincast no-rider control), which main doesn't currently have at the integration level. All four pass against #5010's implementation. The earlier changes-requested on the implementation is moot now that the code is gone.

matthewevans (approved):
Maintainer approval: current head is test-only, preserves all four parser-path regressions, and is registered under the integration test harness. The stale implementation review is superseded by the current head; local fmt/parser gate and Tilt clippy, engine tests, and card-data resources are green.

PR #4925, opened 2026-07-02
fix(parser): gate keyword statics on graveyard-card presence, distribute "the same is true for" (Cairn Wanderer)

### LINKED ISSUE #4774: Cairn is cracked — [[Cairn Wanderer]] gets all the keywords of his static ability even though theres no creatures with…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521722966495465482/1521722966495465482

**Thread:** Cairn is cracked
**Cards:** cairn wanderer

## Summary
[[Cairn Wanderer]] gets all the keywords of his static ability even though theres no creatures with those keywords in any graveyard.

### cairn wanderer
Changeling (This card is every creature type.)
As long as a creature card with flying is in a graveyard, this creature has flying. The same is true for fear, first strike, double strike, deathtouch, haste, landwalk, lifelink, protection, reach, trample, shroud, and vigilance.

---
<sub>report_id: `discord:1521722966495465482:1521722966495465482:0` · discord: `1521722966495465482/1521722966495465482`</sub>

dhgoal:
## Problem

Cairn Wanderer gains **every** listed keyword unconditionally, even with no matching creature card in any graveyard. Two compounding parser defects on the graveyard-keyword static:

1. **Gate-swallow.** "As long as a creature card with `<kw>` is in a graveyard, this creature has `<kw>`" had no combinator for the `<card> is in a graveyard` gate, so it fell to `StaticCondition::Unrecognized`, which evaluates as **always true** at runtime — the same gate-swallow class fixed for trailing "if" gates in #4841. The same bug made **Tarmogoyf** grant trample with no land card in any graveyard.
2. **Distributive-tail mis-tokenization.** The "The same is true for `<keyword list>`" continuation was not distributed per keyword.

## Fix (parser-only)

- **`parse_card_in_graveyard`** (`oracle_nom/condition.rs`): nom combinator, sibling of `parse_creature_has_keyword`, parsing "[a/an] `<type-phrase>` [with `<keyword>`] is in a graveyard" → `IsPresent(Typed + InZone{Graveyard} [+ WithKeyword])`. Covers the whole class — Tarmogoyf's bare card-type gate and Cairn's card-type + keyword gate.
- **`push_graveyard_keyword_same_is_true_tail`** (`oracle.rs`): distributes the "same is true for" list, cloning the gated static per keyword and swapping **both** the granted keyword and the gate's `WithKeyword`, so each keyword is granted only while a card with **that** keyword is in a graveyard (CR 611.3a per-keyword conditional continuous static). Plain-`StaticDefinition` analogue of the trigger path's `attach_same_is_true_keywords`, reusing `try_parse_same_is_true_continuation` and `rewrite_condition_keyword`.

- **Graveyard layer invalidation** (`game/zones.rs` + `game/layers.rs`): the zone-change seam only re-armed layers for battlefield/hand moves, so a graveyard membership change never re-evaluated a graveyard-gated static in production. Added a graveyard entry/exit invalidation gated on `any_active_static_reads_zone_membership(Graveyard)` (scans the static-effect-source index — O(generators)). The detector covers both `IsPresent(InZone{Graveyard})` presence gates **and** `QuantityComparison` count gates via an exhaustive, wildcard-free `QuantityExpr`/`QuantityRef` walk (`GraveyardSize`, `ZoneCardCount{Graveyard}`, graveyard-scoped `ObjectCount`/`Aggregate`, `DistinctCardTypes` from graveyard). Mill/deaths/discard stay cheap when no such static is live; layers re-evaluate only when a graveyard move can flip a live gate.

The one runtime addition is that graveyard invalidation; `IsPresent` + `FilterProp::InZone`/`WithKeyword` already evaluate the gate itself, and a card's printed keywords are visible in the graveyard.

## Anchored on
- `crates/engine/src/parser/oracle_nom/condition.rs:3116` — `parse_creature_has_keyword`: the sibling `OracleResult<StaticCondition>` combinator that runs `parse_type_phrase` then `add_filter_property(FilterProp::WithKeyword)` into an `IsPresent` gate. `parse_card_in_graveyard` mirrors it exactly, substituting the `is in a graveyard` predicate and `FilterProp::InZone{Graveyard}`, and is registered in the same condition `alt()` dispatcher (`parse_remaining_state_presence_conditions`), just as `parse_creature_has_keyword` is registered in `parse_control_conditions` (condition.rs:2779).
- `crates/engine/src/parser/oracle_effect/mod.rs:17379` — `attach_same_is_true_keywords`: the "same is true for" distributor that clones a template per keyword and swaps both the `AddKeyword` modification and the gate keyword via `rewrite_condition_keyword`. `push_graveyard_keyword_same_is_true_tail` is the plain-`StaticDefinition` analogue of this exact pattern and reuses the same `rewrite_condition_keyword` / `try_parse_same_is_true_continuation` building blocks.

## Gate A
```
(no output; exit 0 — no combinator-purity violations)
```

## Tests

- `condition_card_with_keyword_in_graveyard_types_to_is_present` — the new gate types to `IsPresent(Creature + WithKeyword{Flying} + InZone{Graveyard})`.
- `static_has_keyword_as_long_as` — **tightened** from `Unrecognized` to the typed gate (Tarmogoyf no longer grants trample unconditionally).
- `cairn_wanderer_distributes_graveyard_keyword_gate_per_keyword` — full-dispatch: **11** gated statics (10 continuation + the modeled flying), each granting exactly the keyword it gates on (flying-gates-on-flying, trample-gates-on-trample); asserts **no** static grants `Keyword::Unknown` and that the unqualified `protection`/`landwalk` clauses surface as an `Effect::Unimplemented` residual.
- `cairn_graveyard_keyword_static_grants_flying_only_while_flyer_in_graveyard` (**runtime**, `game/layers.rs`) — drives `evaluate_layers` + `has_keyword`: empty graveyard → no Flying; a flying creature card in a graveyard → Cairn gains Flying; remove it → grant falls away. Fails on revert of the gate.
- `graveyard_arrival_reevaluates_graveyard_gated_static_via_zone_move` (**production path**, `game/zones.rs`) — mills a flying creature card Library→Graveyard via `move_to_zone` **with no manual `mark_full`**, then asserts layers are dirty and `evaluate_layers` grants Cairn Flying. Fails on revert of the zone-seam invalidation.
- `graveyard_arrival_does_not_dirty_layers_without_graveyard_gated_static` — proves the invalidation is scoped: no graveyard-gated static → milling a card does **not** dirty layers (no per-graveyard-move perf regression).
- `graveyard_count_gated_static_reevaluates_via_zone_move` (**production path**) — a `QuantityComparison(GraveyardSize >= 1)`-gated static grants its keyword after a card is milled into the graveyard via `move_to_zone`; fails on revert of the count-gate detector branch.

See **## Verification** below.

## Notes

Unqualified `protection` / `landwalk` in the list resolve to `Keyword::Unknown` via the shared `Keyword::from_str` (no quality/subtype to type). Rather than clone them into inert `AddKeyword(Unknown)` statics — which `game/coverage.rs` would read as Continuous-mode *supported*, falsely marking the card covered — they are emitted as an explicit `Effect::Unimplemented` residual (via the existing `make_unimplemented` helper, the same mechanism the chosen/every-type `same is true for` path uses), so they stay a loud unsupported gap. Typing generic `protection`/`landwalk` is a separate keyword-vocabulary change, left as a follow-up.

## Verification

- `cargo fmt --all` — clean
- `cargo clippy -p engine` — clean (`-D warnings`). Full-workspace `cargo clippy-strict` is not runnable in my sandbox (a non-engine binary target needs OpenSSL/`pkg-config`, unavailable, no sudo); CI **Rust lint (fmt, clippy, parser gate)** runs it green.
- `./scripts/check-parser-combinators.sh` — clean (exit 0, no violations) — see **## Gate A**
- `cargo test -p engine --lib -- parser:: game::layers:: game::zones::` — **7435 passed / 0 failed** (includes the Cairn runtime + graveyard presence/count invalidation production-path tests). The full sharded `cargo test -p engine` runs green in CI (**Rust tests 1/2 + 2/2**).
- Card-data / coverage / semantic-audit — the local pipeline needs an MTGJSON `AtomicCards.json` download unavailable in my sandbox; CI **Card data (generate, validate, coverage)** runs `gen-card-data` + `coverage` + the parser-honesty regression check green on this head. Honesty note: unqualified `protection` / `landwalk` are emitted as an explicit `Effect::Unimplemented` residual (see **## Notes**) so `coverage` marks them unsupported rather than the card reading as fully covered; the 11 parameter-typed keywords are gated and supported.

## CR

CR 611.3a (conditional continuous static, re-evaluated each moment) + CR 702 (keyword abilities) + CR 404 (graveyard zone). Verified against `docs/MagicCompRules.txt`.

Closes #4774

Tier: Frontier
Model: claude-opus-4-8
Thinking: high

### DIFF
--- a/crates/engine/src/game/layers.rs
+++ b/crates/engine/src/game/layers.rs
@@ -25,7 +25,7 @@ use crate::types::ability::{
     AbilityCost, AbilityDefinition, AbilityKind, ActivationRestriction, BasicLandType,
     CastingPermission, ChosenSubtypeKind, CommanderOwnership, ContinuousModification,
     CopiableValues, Duration, Effect, FilterProp, ManaContribution, ManaProduction, PlayerScope,
-    QuantityExpr, StaticCondition, StaticDefinition, TargetFilter, TypedFilter,
+    QuantityExpr, QuantityRef, StaticCondition, StaticDefinition, TargetFilter, TypedFilter,
 };
 use crate::types::attribution::EffectRef;
 use crate::types::card_type::{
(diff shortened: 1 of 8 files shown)


matthewevans:
### Parse changes introduced by this PR · 1 card(s), 13 signature(s)  (baseline: main `6d6783e8df5e`)

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Deathtouch in graveyard creature is present, mods=grant Deathtouch)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with DoubleStrike in graveyard creature is present, mods=grant DoubleStrike)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Fear in graveyard creature is present, mods=grant Fear)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with FirstStrike in graveyard creature is present, mods=grant FirstStrike)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Flying in graveyard creature is present, mods=grant Flying)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Haste in graveyard creature is present, mods=grant Haste)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Lifelink in graveyard creature is present, mods=grant Lifelink)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Reach in graveyard creature is present, mods=grant Reach)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Shroud in graveyard creature is present, mods=grant Shroud)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Trample in graveyard creature is present, mods=grant Trample)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · added: `Continuous (affects=self, conditional=with Vigilance in graveyard creature is present, mods=grant Vigilance)`
Examples: Cairn Wanderer

#### 1 card(s) · static/Continuous · removed: `Continuous (affects=self, conditional=unrecognized, mods=grant Flying, grant FirstStrike, grant DoubleStrike, grant Deathtouch, grant Haste, grant Lifelink, gr…`
Examples: Cairn Wanderer

#### 1 card(s) · ability/unknown · added: `unknown`
Examples: Cairn Wanderer

dhgoal:
Thanks — this parse-diff is the intended transformation for #4774. The single `conditional=unrecognized` grant (which granted every keyword regardless of graveyard contents — the reported bug) is replaced by one `Continuous` static **per keyword**, each gated on a creature card *with that keyword* being in a graveyard (CR 611.3a).

On the two `grant Unknown` rows: those are the bare `protection` and `landwalk` entries in Cairn's list. The shared `Keyword::from_str` only types `protection`/`landwalk` when a parameter is present (`Protection(target)` / `Landwalk(subtype)`); parameterless generic forms have no quality/subtype to bind, so they resolve to `Keyword::Unknown` and stay inert. I left them as loud gap markers rather than silently dropping them (coverage-honesty) or inventing a bogus target/subtype. The other 11 parameter-free keywords gate correctly (spot-checked in `cairn_wanderer_distributes_graveyard_keyword_gate_per_keyword`: flying-gates-on-flying, trample-gates-on-trample). Happy to fold proper generic-protection/landwalk typing into a follow-up if you'd prefer.

matthewevans (changes requested):
Thanks for the follow-up. I agree the main parse-diff direction is the intended one for #4774, but I found one coverage-honesty blocker before this can land.

**[MED]** Generic `landwalk` / `protection` continuation clauses are parsed into inert supported statics. Evidence: `crates/engine/src/parser/oracle.rs:1285` clones every continuation keyword into `ContinuousModification::AddKeyword`, while `crates/engine/src/game/coverage.rs:3893` only checks static-mode support and does not mark `AddKeyword(Keyword::Unknown(_))` unsupported. Why it matters: the parse-diff's bare `landwalk` and `protection` rows are not semantically modeled, but they are no longer represented as an explicit unsupported residual, so Cairn Wanderer can become coverage-supported while those two keyword clauses do nothing useful. Suggested fix: keep unqualified `landwalk` / `protection` as an explicit unsupported residual for this parser path, or add typed support for generic landwalk/protection before cloning those continuations into statics.

For first-time AI-assisted contributions, please make sure future updates keep the `docs/AI-CONTRIBUTOR.md` evidence gates explicit: production entry, current-head parse diff, and tests that fail on revert.

dhgoal:
Thanks for the quick and thorough look — much appreciated.

**Current-head parser evidence.** The branch had drifted behind `main`, so I've rebased onto current `main` and force-pushed; head is now `d42dcdd5` (`behind_by 0` at push). That re-triggers CI, so the `<!-- coverage-parse-diff -->` sticky will refresh for this exact head once the run's `workflow_run` completes. The blast radius it should show (scoped to Cairn Wanderer / Tarmogoyf as claimed):
- **removed** the single `Continuous (conditional=unrecognized, mods=grant Flying, FirstStrike, …)` — the unconditional over-grant that was the #4774 bug (and the same `Unrecognized`→always-true swallow that made Tarmogoyf grant trample with an empty graveyard);
- **added** one `Continuous` static **per keyword**, each gated on `IsPresent(creature card WITH that keyword in a graveyard)` (CR 611.3a). Two rows show `grant Unknown` — the bare `protection` / `landwalk` entries, which have no quality/subtype to bind so they resolve to `Keyword::Unknown` via the shared `from_str`; I left them as honest coverage gaps rather than silently dropping them or inventing a target/subtype.

**AI-CONTRIBUTOR gates.** Reviewed `docs/AI-CONTRIBUTOR.md` and added the missing `## Verification` section to the PR body (Gate A / `## Anchored on` / `Tier: Frontier` / `Model:` were already there). Two items I couldn't run in my sandbox, so I've cited the authoritative CI jobs instead of fabricating local output: full-workspace `clippy-strict` needs OpenSSL/`pkg-config` (no sudo) → CI **Rust lint** is the green run; the card-data/coverage/semantic-audit pipeline needs an MTGJSON download → CI **Card data (generate, validate, coverage)** (incl. the parser-honesty regression check) is the green run. Locally: `cargo fmt` clean, `cargo clippy -p engine` clean, combinator gate clean, `cargo test -p engine --lib -- parser::` 7186 pass / 0 fail.

Happy to adjust anything once the refreshed sticky lands — and glad to split proper generic-`protection`/`landwalk` typing into a follow-up if you'd prefer those handled rather than left as gaps.

dhgoal:
You're right — thanks, that's a real coverage-honesty hole and I've fixed it. Pushed `443701ff` (rebased onto current `main`).

**Fix.** `push_graveyard_keyword_same_is_true_tail` now partitions the continuation: keywords that resolve to a real `Keyword` are cloned into their per-keyword gated statics as before, but any that resolve to `Keyword::Unknown` (unqualified `protection` / `landwalk` — no quality/subtype to type) are **no longer cloned into inert `AddKeyword(Unknown)` statics**. They're emitted as an explicit `Effect::Unimplemented` residual via the existing `make_unimplemented` helper — the same residual mechanism the chosen-/every-type `push_same_is_true_static_tail` path already uses. So `game/coverage.rs` now sees a genuine unsupported item for Cairn instead of a Continuous-mode static that reads as supported; the card can't go coverage-supported while those two clauses do nothing.

**Evidence gates.**
- **Tests that fail on revert.** Added a discriminating *runtime* test — `cairn_graveyard_keyword_static_grants_flying_only_while_flyer_in_graveyard` (`game/layers.rs`) — that drives the real layer pipeline (`evaluate_layers` + `has_keyword`), not just parsed AST: empty graveyard → Cairn has **no** Flying; a flying creature *card* in a graveyard → Cairn **gains** Flying; remove it → the grant falls away. This fails on revert of the gate (the pre-fix `Unrecognized`→always-true condition would keep Flying with an empty graveyard). The parse test additionally asserts **no** static grants `Keyword::Unknown` and that the unqualified clauses surface as the Unimplemented residual naming `landwalk` / `protection`.
- **Current-head parse diff.** Rebased to current `main`, so the sticky refreshes for this head — it should now show **+11 gated grants (10 continuation + the modeled flying) / −1** unconditional `Unrecognized` grant, plus the landwalk/protection residual, Cairn Wanderer only.
- Local: `cargo test -p engine --lib -- parser:: game::layers::` 7389 pass / 0 fail; `clippy -p engine` clean; wasm-dev check clean. (Full `clippy-strict` / card-data pipeline still cited via the CI **Rust lint** and **Card data** jobs per the sandbox limits noted in the PR body.)

I opted for the "explicit unsupported residual" path over adding typed generic `landwalk`/`protection` — the latter is a broader keyword-vocabulary change I'd rather not fold into this fix. Happy to do that as a follow-up if you'd like those two actually modeled.

matthewevans (changes requested):
Thanks for the update. I still need one production-path fix before this can land.

The new graveyard-gated statics depend on `InZone(Graveyard)`, but zone movement currently only marks layers dirty for battlefield/hand changes. The added runtime coverage manually calls `mark_full()` before `evaluate_layers()`, so it does not catch the production path where a card moves into or out of a graveyard and Cairn/Tarmogoyf-style conditions should re-evaluate.

Please move the invalidation to the zone-change seam, or the equivalent layer invalidation seam, for graveyard membership changes that can affect active static conditions, and add a production-path test that uses the normal zone move path without manually forcing layers dirty.

dhgoal:
Good catch — you're right that the manual `mark_full()` in the layers test hid the real gap. Fixed at the zone-change seam. Pushed `95f3550b` (rebased onto current `main`).

**Invalidation.** `zones::move_to_zone` only re-armed layers for `to == Battlefield || to == Hand || from == Hand`, so a graveyard membership change never re-evaluated a graveyard-gated static. Added, right beside that mark:

```rust
if (to == Zone::Graveyard || from == Zone::Graveyard)
    && crate::game::layers::any_active_static_reads_zone_membership(state, Zone::Graveyard)
{
    crate::game::layers::mark_layers_full(state);
}
```

**Scoped, not blanket.** `any_active_static_reads_zone_membership` walks the static-effect-source index (`for_each_static_effect_source`, O(generators)) and returns true only when a live `Continuous` static has an enabling condition whose filter tests `FilterProp::InZone { Graveyard }` (`IsPresent`, incl. `Not`/`And`/`Or` compounds). So routine graveyard churn — deaths, mill, discard — stays cheap when no such static exists; the full re-eval only fires when it can actually matter. I used the full-invalidation path rather than extending the battlefield-entry incremental narrowing (`entered_object_perturbs_static_condition`), which is entry-into-battlefield scoped and, per its own comment, assumes `IsPresent` is battlefield population — a graveyard-scoped `IsPresent` is exactly the case it doesn't model, and a full pass evaluates the `InZone { Graveyard }` filter correctly.

**Production-path test.** `graveyard_arrival_reevaluates_graveyard_gated_static_via_zone_move` (`game/zones.rs`) mills a flying creature card **Library → Graveyard through `move_to_zone`, with no manual `mark_full`**, then asserts layers are dirty and — after a normal `evaluate_layers` — Cairn has Flying. Library→Graveyard deliberately bypasses the pre-existing hand/battlefield mark, so it exercises the new seam specifically and fails on revert of it. Paired with `graveyard_arrival_does_not_dirty_layers_without_graveyard_gated_static`, which proves the gate stays clean (no dirty) when no graveyard-membership static is active.

Verification: `cargo test -p engine --lib -- game::zones:: game::layers:: parser::` 7433 pass / 0 fail; `clippy -p engine` + wasm-dev check clean; rebased current `main`.

matthewevans (changes requested):
Thanks for the follow-up. The Cairn/Tarmogoyf `IsPresent(... InZone(Graveyard))` invalidation path is now in the right production seam, but the detector is still too narrow for the supported static-condition class.

**[MED]** Graveyard-count static conditions still do not dirty layers. Evidence: `crates/engine/src/game/zones.rs:725` only marks layers for graveyard moves when `any_active_static_reads_zone_membership(..., Zone::Graveyard)` returns true, but `crates/engine/src/game/layers.rs:1959` only detects `StaticCondition::IsPresent` filters with `FilterProp::InZone`; supported graveyard-count gates parse through `QuantityComparison` instead, e.g. `crates/engine/src/parser/oracle_nom/condition.rs:9747` (`GraveyardSize`) and `crates/engine/src/parser/oracle_static/tests.rs:5131` (“fewer than eight cards in your graveyard”). Why it matters: a graveyard move can cross one of those thresholds without dirtying layers, leaving continuous effects like type removal, keyword grants, or P/T changes stale until some unrelated layer invalidation happens. Suggested fix: extend the zone-read detector through `QuantityExpr` / `QuantityRef` for graveyard-count reads (`GraveyardSize`, `ZoneCardCount` targeting `Zone::Graveyard`, and nested aggregates as applicable) and add a `move_to_zone` regression that fails on revert.

dhgoal:
Right again — the detector only saw `IsPresent`/`InZone` and missed the count path. Extended it through `QuantityComparison` → `QuantityExpr`/`QuantityRef`. Pushed `c48422a9` (rebased onto current `main`).

**Detector.** `static_condition_reads_zone_membership` now also matches `QuantityComparison { lhs, rhs, .. }` and checks both operands via a new `quantity_expr_reads_zone`, which recurses the expression tree (mirroring `quantity::quantity_expr_uses_object_count` — `Sum`/`Max`/`Offset`/`ClampMin`/`Difference`/etc.) down to `quantity_ref_reads_zone`. The leaf classifier flags a zone read for:
- `GraveyardSize` (→ `Zone::Graveyard`);
- `ZoneCardCount { zone }` / `TargetZoneCardCount { zone }` (via a `ZoneRef`→`Zone` map), plus a graveyard-scoped `ZoneCardCount.filter`;
- `ObjectCount` / `ObjectCountDistinct` / `ObjectCountBySharedQuality` / `Aggregate` whose `TargetFilter` carries `FilterProp::InZone { Graveyard }`;
- `DistinctCardTypes` sourced from `CardTypeSetSource::Zone { Graveyard }` (Tarmogoyf's card-types-in-graveyards read).

It's **exhaustive and wildcard-free** (every other `QuantityRef` variant is enumerated as `false`), mirroring `quantity_ref_uses_object_count`, so a future zone-reading quantity is forced through this classification rather than silently under-escalating.

**Production-path test.** `graveyard_count_gated_static_reevaluates_via_zone_move` (`game/zones.rs`): a static gated on `QuantityComparison(GraveyardSize >= 1)` grants Trample only after a card is milled Library→Graveyard through `move_to_zone` (no manual `mark_full`) and layers re-evaluate. It fails on revert of the `QuantityComparison`/`QuantityRef` branch.

Verification: `cargo test -p engine --lib -- game::zones:: game::layers:: parser::` 7435 pass / 0 fail; `clippy -p engine` clean (`-D warnings`); wasm-dev clean; rebased current `main`.

matthewevans:
Reviewed current head `c48422a9` after the latest follow-up. The previous graveyard-count blocker looks addressed: `static_condition_reads_zone_membership` now follows `QuantityComparison` through `QuantityExpr`/`QuantityRef`, and `graveyard_count_gated_static_reevaluates_via_zone_move` exercises the actual `move_to_zone` path without a manual layer mark.

I'm not approving/enqueueing this from the sweep yet because the parse-diff sticky is still from `2026-07-02T22:28:28Z`, before this latest push, and the current-head Card data job that refreshes that evidence is still pending. Once that current-head sticky lands, this should resurface for final handler routing if the diff remains scoped.

matthewevans:
Holding review on current head `9cca04c9e09e9b92fcbbc1ccac06052e1ab3c0e7`: the branch was just updated/merged with `main`, so the existing `<!-- coverage-parse-diff -->` sticky predates the current head and card-data is running again. I need a current-head parse-diff before I can re-review this parser PR.

matthewevans:
@dhgoal looking into the parse-diff issue. Nothing you need to do on your end. Thanks for the contribution :)

matthewevans (approved):
Current-head review is clean. The prior coverage-honesty and graveyard layer invalidation blockers are addressed: unsupported generic continuations remain explicit, graveyard static-condition reads invalidate at the zone-change seam, count-based graveyard gates are covered, and the current parse-diff is scoped to Cairn Wanderer.

PR #4926, opened 2026-07-02
fix(parser): gate "can't attack" on "if there's another creature on the battlefield" (Shauku, Endbringer)

minion1227:
## Problem

Shauku, Endbringer reads "Shauku can't attack **if there's another creature on the battlefield**." The trailing `if` gate was swallowed, so the `CantAttack` static parsed with `condition: null` and applied **unconditionally** — Shauku could never attack, even when it is the only creature on the battlefield.

## Fix

Add `parse_there_is_exists_on_battlefield_condition` in the test-free `oracle_static/shared.rs`: the singular existential ("there's a[n]/another \<type\> on the battlefield") counterpart of the existing "there are N or more \<type\> on the battlefield" count form, mapping to `ObjectCount(\<type\>) >= 1`. Registered in `parse_static_condition` alongside the sibling existence/count helpers.

The `another ` article is preserved into the type phrase so `parse_type_phrase` attaches the source-exclusion `Another` prop (**CR 611.3a**): the source's own presence must not satisfy its own gate, otherwise the restriction would never lift.

No new enum variant — reuses the merged existence-condition family (Wirecat #4844, Hour of Revelation #4880).

### Before / after

```
before: CantAttack{SelfRef}  condition: null           (1 swallowed clause)
after:  CantAttack{SelfRef}  condition: ObjectCount(Creature, Another) >= 1
```

## Tests

- New `shauku_cant_attack_gated_on_another_creature_exists` regression: asserts the `CantAttack` static carries the `ObjectCount(Creature) >= 1` gate with the `Another` source-exclusion prop, and that full-card dispatch produces the gated static with no swallowed clause.
- 941 `oracle_static` tests pass; `cargo fmt` + `clippy -D warnings` clean.


matthewevans (approved):
Maintainer review: this is scoped and idiomatic. The current head extends the existing static-condition parser with a composed nom arm for the singular `there is/there's ... on the battlefield` existential, reuses the existing `ObjectCount(filter) >= 1` condition shape, preserves `another` as `FilterProp::Another`, and the parse-diff is limited to the expected Shauku / Glacial Crasher CantAttack conditions. Approving for merge queue.

PR #4927, opened 2026-07-02
fix(parser): gate combat restrictions on trailing "as long as" conditions (Seer of the Bright Side)

minion1227:
## Problem

The can't-attack / can't-attack-or-block / can't-block static arms peel a trailing `unless [cost]` or `if [board-state]` gate, but **silently drop a trailing `as long as [board-state]` gate** — enforcing the restriction **unconditionally**.

Seer of the Bright Side (`This creature can't attack or block **as long as it has a stun counter on it**.`) could therefore never attack or block, regardless of whether it actually has a stun counter. Ethrimik, Imagined Fiend (`**As long as you control another creature**, Ethrimik can't attack or block.`) and every other `can't attack/block as long as X` card were likewise mis-gated.

## Fix

`as long as` and `if` both express a continuous game-state gate on a static ability (**CR 611.3a**). Added `parse_as_long_as_static_condition` (a direct mirror of `parse_if_static_condition`) in the test-free `oracle_static/shared.rs`, and wired it into the restriction arms in `oracle_static/dispatch.rs` between the `unless` and `if` peels — matching `split_trailing_gate_condition`'s existing `as long as`-before-`if` precedence.

The condition bodies already parse via `parse_static_condition` → the shared `parse_inner_condition` authority (counters, control-count, existence, …); **only the keyword peel was missing**, so this closes the whole class without touching the condition grammar and without duplicating any shared parser.

### Before / after (Seer of the Bright Side)

```
before: CantAttackOrBlock{SelfRef}  condition: null   (1 swallowed clause → applies always)
after:  CantAttackOrBlock{SelfRef}  condition: HasCounters{ OfType(stun), min: 1 }
```

## Tests

- New `cant_attack_or_block_gated_on_trailing_as_long_as` regression: Seer's `as long as it has a stun counter` rider must gate the restriction (`HasCounters`), with no swallowed clause; plus `if` and `can't block as long as` cases to lock in that the other keyword peels still attach their conditions.
- 941 `oracle_static` tests pass; `cargo fmt` + `clippy -D warnings` clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -2042,10 +2042,13 @@ pub(crate) fn parse_static_line_inner(
         let mut def = StaticDefinition::new(StaticMode::CantBlock)
             .affected(TargetFilter::SelfRef)
             .description(text.to_string());
-        // CR 509.1c: a trailing "unless [cost]" or "if [board-state]" clause
-        // scopes the restriction; attach whichever is present.
-        if let Some(condition) =
-            parse_unless_static_condition(&tp).or_else(|| parse_if_static_condition(&tp))
+        // CR 509.1c + CR 611.3a: a trailing "unless [cost]", "as long as
+        // [board-state]", or "if [board-state]" clause scopes the restriction;
+        // attach whichever is present. "as long as" is tried before "if" to match
+        // `split_trailing_gate_condition`'s precedence.
+        if let Some(condition) = parse_unless_static_condition(&tp)
+            .or_else(|| parse_as_long_as_static_condition(&tp))
+            .or_else(|| parse_if_static_condition(&tp))
         {
             def.condition = Some(condition);
         }
@@ -2086,10 +2089,14 @@ pub(crate) fn parse_static_line_inner(
         let mut def = StaticDefinition::new(mode)
             .affected(TargetFilter::SelfRef)
             .description(text.to_string());
-        // CR 508.1: a trailing "unless [cost]" or "if [board-state]" clause
-        // scopes the restriction; attach whichever is present.
-        if let Some(condition) =
-            parse_unless_static_condition(&tp).or_else(|| parse_if_static_condition(&tp))
+        // CR 508.1 + CR 611.3a: a trailing "unless [cost]", "as long as
+        // [board-state]", or "if [board-state]" clause scopes the restriction;
+        // attach whichever is present. "as long as" is tried before "if" to match
+        // `split_trailing_gate_condition`'s precedence (Seer of the Bright Side:
+        // "... can't attack or block as long as it has a stun counter on it.").
+        if let Some(condition) = parse_unless_static_condition(&tp)
+            .or_else(|| parse_as_long_as_static_condition(&tp))
+            .or_else(|| parse_if_static_condition(&tp))
         {
             def.condition = Some(condition);
         }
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[MED]** The new `as long as` peel also matches `for as long as` duration text. Evidence: `parse_as_long_as_static_condition` splits on raw `" as long as "` (`crates/engine/src/parser/oracle_static/shared.rs:1911-1913`), and the generic `can't attack` arm applies it to every self-restriction (`crates/engine/src/parser/oracle_static/dispatch.rs:2097-2099`). The parse-diff sticky confirms the leak: Promise of Loyalty changed from no `static/CantAttack.conditional` to `has 1+ counters`, but that card's text is `for as long as it has a vow counter`, which is duration/provenance text rather than a trailing static gate. Why it matters: `for as long as` belongs in the effect duration pipeline (`Duration::ForAsLongAs`) or should remain unsupported; accepting it here is wrong-seam parser support. Please reject `for as long as` in this helper, add a Promise of Loyalty regression/parse-diff guard, and only handle that class through the duration/effect seam if intended.

**[LOW]** The block-restriction CR citation points at the requirements subrule instead of the restrictions subrule. Evidence: the `can't block` arm cites `CR 509.1c` (`crates/engine/src/parser/oracle_static/dispatch.rs:2045`), but `CR 509.1b` is the restriction rule for effects that say a creature can't block; `509.1c` is requirements. Please update the block-restriction citation to `CR 509.1b` where the code is about `can't block` restrictions.

minion1227:
Thanks — both addressed in 58371eb63.

**[MED] `for as long as` leak.** `parse_as_long_as_static_condition` now rejects the `for as long as` form before splitting on the bare `as long as`, so duration/provenance text stays with the `Duration::ForAsLongAs` / effect seam and is no longer mis-attached as a static gate. Promise of Loyalty ("… can't attack you or planeswalkers you control **for as long as** it has a vow counter on it") no longer gains a bogus static `HasCounters` gate. Added a building-block regression guard (`as_long_as_peel_rejects_for_as_long_as_duration`) using the real Promise of Loyalty text: it asserts the `for as long as` form → `None` and the bare Seer `as long as` form → `Some` (no over-rejection).

**[LOW] CR citation.** Corrected the `can't block` arm to **CR 509.1b** (block *restrictions*; 509.1c is block *requirements*), verified against `docs/MagicCompRules.txt`.

Parser-combinator gate green, 951 `oracle_static` tests pass, clippy clean. The CI parse-diff should now show Promise of Loyalty reverting to no `static/CantAttack.conditional`.

matthewevans:
Current-head code review looks clean: the `for as long as` leak is addressed by rejecting that duration form before the bare `as long as` split, and the `can't block` citation now points to CR 509.1b.

Holding final approval until the current-head parse-diff sticky refreshes. The existing `<!-- coverage-parse-diff -->` comment is still the older pre-fix body and still lists Promise of Loyalty contamination, so I need the refreshed card-level diff for head `58371eb63bb82cd7b72c3d43e7e7c51768c51beb` before approving/enqueueing.

minion1227:
Heads up on the head change: I rebased onto current main (`62a319618`) — new head `d8dcaba8d`, **code byte-identical** to `58371eb63` (empty diff on all three files). The prior parse-diff baseline (`0ad57af`) was orphaned and never published a coverage snapshot, so the sticky was stuck on "baseline pending"; rebasing onto a commit with a published snapshot lets it re-baseline.

The **refreshed sticky is now populated** (baseline main `039e5a25b396`) and confirms the fix: **Promise of Loyalty is gone** from the diff. The only remaining changes are the intended combat-restriction gates —

- Seer of the Bright Side → `CantAttackOrBlock.conditional`: `has 1+ counters`
- Ethrimik / Imagined Fiend → `CantAttackOrBlock.conditional`: `another creature is present`

No `for as long as` / duration-text contamination. Ready for re-review/enqueue when you get a chance.

matthewevans (approved):
Current-head review is clean. The previous `for as long as` overmatch is no longer present in the refreshed parse diff, the change is scoped to the combat-restriction gate seam, and the tests cover the bare `as long as` gate plus the duration-seam rejection.

PR #4929, opened 2026-07-02
fix(parser): source-scoped "if ~ has counters on it" intervening-if

real-venus:
## Summary

A source-scoped intervening-if of the form *"…, if ~ has counters on it, …"* (CR 603.4) was silently dropped on triggered abilities — the trigger's `condition` was left `None`, so the ability fired/prompted **unconditionally**:

- **The Ozolith** — "At the beginning of combat on your turn, **if ~ has counters on it**, you may move all counters…" → previously offered the counter-move *every* combat even while holding zero counters.
- **Denry Klin, Editor in Chief** — "Whenever a nontoken creature you control enters, **if ~ has counters on it**, proliferate." → previously proliferated on every creature ETB regardless.

## Root cause

The `try_extract_simple_condition` dispatch table in `oracle_trigger.rs` handles sibling source-scoped intervening-ifs (`"if it's attacking"` → `SourceIsAttacking`, `"if ~ attacked this turn"`, `"if ~ is renowned"`) but had **no arm for "has counters on it"**.

## Fix

Add one dispatch arm mapping `"if ~ has counters on it"` to the existing, already-evaluated:
```rust
TriggerCondition::HasCounters { counters: CounterMatch::Any, minimum: 1, maximum: None }
```
(evaluated in `game/triggers.rs`). **No new `TriggerCondition` variant, no engine/resolver change.** Distinct from the past-tense event-subject *"if it had counters on it"* (`HadCounters`), which is handled separately.

## Testing

`intervening_if_source_has_counters_on_it_populates_condition` asserts both The Ozolith and Denry Klin parse to the gated `HasCounters` condition (and the effect still parses). Passes.

CR 603.4 + CR 122.


matthewevans:
### Parse changes introduced by this PR · 2 card(s), 2 signature(s)  (baseline: main `86e5db5ca7c0`)

#### 1 card(s) · trigger/ChangesZone · field `condition`: `∅` → `has 1+ counters`
Examples: Denry Klin, Editor in Chief

#### 1 card(s) · trigger/Phase · field `condition`: `∅` → `has 1+ counters`
Examples: The Ozolith

matthewevans:
Sweep status: this remains held. Local code review did not find a blocker, but the parse-diff sticky is still baseline-pending on current head, so I am waiting for current card-level parse evidence before any approval/enqueue decision.

matthewevans (approved):
Reviewed current head. The fix now delegates the source-scoped counter gate to the existing counter-condition combinator, preserves the trigger-side HasCounters bridge/evaluator, covers The Ozolith and Denry Klin through parse_trigger_line including the short-name normalization path, and the current parse-diff is scoped to the two intended trigger conditions.

PR #4931, opened 2026-07-02
Recover the count and filter when a verb cost continues past an and-clause

### LINKED ISSUE #2343: Mechtitan Core: token missing Haste (and 'all colors'), exile cost only removes self not four others
## Reported behaviors

Two related bugs on Mechtitan Core:

1. **Token missing Haste**: The Mechtitan token created by the activated ability does not have haste, despite the oracle text specifying it. card-data.json confirms the token keywords list is `[Flying, Vigilance, Trample, Lifelink]` — Haste is absent. The token is also not marked as all colors.
2. **Exile cost only removes self**: When activating Mechtitan Core, only the Core itself is exiled. The four other artifact creatures/Vehicles that must also be exiled as part of the activation cost are not removed. The parsed cost structure shows only `Exile(count:1, filter:SelfRef)` + `Exile(count:1, filter:Any)` rather than `Exile(count:4, filter:artifact-creatures-or-vehicles)`.
3. **Return trigger missing**: The 'when that token leaves the battlefield, return all cards exiled...' trigger is not present in card-data.

## Oracle text (verified from card-data.json)

```
{5}, Exile this Vehicle and four other artifact creatures and/or Vehicles you control: Create Mechtitan, a legendary 10/10 Construct artifact creature token with flying, vigilance, trample, lifelink, and haste that's all colors. When that token leaves the battlefield, return all cards exiled with this Vehicle except this card to the battlefield tapped under their owners' control.
Crew 2
```

## Expected vs actual

- **Expected**: Activating for {5} + exiling self + four other artifact creatures/vehicles creates a 10/10 Flying Vigilance Trample Lifelink **Haste** **all-colors** token, with a return trigger on that token.
- **Actual**: Only self is required in exile cost (plus 1 generic other); token has no haste; return trigger missing.

## Source

https://discord.com/channels/1485498006781427802/1512229915037143142/1512229915037143142
https://discord.com/channels/1485498006781427802/1512230367610929323/1512230367610929323

---
<sub>report_id: `discord:1512229915037143142:1512229915037143142:0` · via bug-triage</sub>

GildardoDev:
Closes #2343

## Anchored on
- crates/engine/src/parser/oracle_cost.rs:166 — `is_from_among_remove_counter_cost` reconstructs a cost fragment and re-checks it through `parse_single_cost`; same building-block reuse this change applies to the verb-prefixed continuation.
- crates/engine/src/parser/oracle_cost.rs:139 — `fixup_from_among_remove_counter_parts` re-joins split cost parts and re-parses them via `parse_single_cost`, the existing precedent for repairing a cost that the top-level splitter fragmented.

## Problem
Mechtitan Core's activation cost is `{5}, Exile this Vehicle and four other artifact creatures and/or Vehicles you control`. Activating it exiled the Core plus a single arbitrary permanent instead of the Core plus **four** artifact creatures/Vehicles you control. (The issue also reported the assembled token missing Haste / "all colors" and the exiled cards not returning — both already fixed on main; this was the remaining sub-bug, so it closes the issue.)

## Root cause
`fixup_bare_noun_continuations` in `oracle_cost.rs` rehydrates a bare noun-phrase continuation that follows a verb-cost (`Exile A and <continuation>`). It re-parsed the continuation with `parse_target(&format!("target {stripped}"))` and a hardcoded `count: 1`. `parse_target` does not consume a leading count word or the "other" property, so `four other artifact creatures and/or Vehicles you control` collapsed to `count: 1` with an empty `Typed` filter — the count and the whole disjunctive filter were dropped.

## Fix
Route the continuation through the verb-prefixed cost parser instead of the degenerate target parse: reconstruct `"{verb} {continuation}"` from the tracked `PrecedingVerb` and call `parse_single_cost`, which already extracts the count, the "other"/`Another` property, and the full (possibly disjunctive) filter for the standalone form. A `match (verb, rehydrated)` only replaces the cost when the reparse yields the same verb's cost variant; otherwise it leaves the honest `Unimplemented` rather than fabricate a wrong cost. This is a class fix for `<Verb> <self/A> and <N> other <type> you control` (exile / sacrifice / tap), not a Mechtitan special case — it also repairs Sword of the Ages (`Sacrifice this artifact and any number of creatures you control`), which previously dropped its continuation to `Unimplemented`.

CR 601.2b / CR 601.2f: additional costs (exiling / sacrificing / tapping permanents) are part of the total cost; the full count and filter must be recovered so the cost is paid in full (CR 118.3).

## Tests
Three building-block regression tests in `parser::oracle_cost::tests` (assert on generic phrasings, not card-name replay):
- `cost_exile_self_and_count_other_you_control_recovers_count_and_filter` — the Mechtitan cost parses to `[Mana{5}, Exile{1, SelfRef}, Exile{4, Or[Typed{Artifact,Creature, You, Another}, Typed{Subtype(Vehicle), You, Another}]}]`; asserts `count == 4` and both disjunction legs' concrete types + `You` + `Another`.
- `cost_sacrifice_and_count_other_continuation_recovers_count` — the sacrifice-verb generalization recovers `count == 2` with the `other … you control` filter.
- `cost_sacrifice_article_continuations_stay_count_one` — regression guard: `A, B, and C` article continuations still each parse as independent `count: 1` sacrifices.

Both count-recovering assertions fail on the pre-fix code (which produced `count: 1` / empty filter).

## Verification
- `cargo test -p engine` — 14581 lib tests + all integration suites pass, 0 failures.
- `cargo test -p engine --lib parser::oracle_cost::tests` — 98 pass.
- `cargo clippy -p engine --all-targets` — 0 warnings.
- `cargo fmt --all` — clean.

## Gate A
```
(exit 0 — no violations)
```

Model: claude-opus-4-8
Tier: Frontier
Thinking: High

### DIFF
--- a/crates/engine/src/parser/oracle_cost.rs
+++ b/crates/engine/src/parser/oracle_cost.rs
@@ -196,7 +196,56 @@ fn fixup_bare_noun_continuations(costs: &mut [AbilityCost]) {
             }
             AbilityCost::TapCreatures { .. } => last_verb = Some(PrecedingVerb::TapCreatures),
             AbilityCost::Unimplemented { description } if last_verb.is_some() => {
+                if description.trim().is_empty() {
+                    continue;
+                }
+                let verb = last_verb.unwrap();
                 let lower = description.to_lowercase();
+                // CR 601.2b/f + #2343 (Mechtitan Core): a continuation that names an
+                // explicit count of two or more objects ("four other artifact
+                // creatures and/or Vehicles you control") must recover that true
+                // count and the full (possibly disjunctive) filter — the historical
+                // `count: 1` + `parse_target` path dropped both. Scope the recovery
+                // to explicit counts >= 2 so single-object continuations keep their
+                // previous parse unchanged (this fix moves no parser surface outside
+                // the explicit-multi-count class). `parse_type_phrase` (the exile
+                // arm's own consumption-aware primitive) must consume the whole
+                // object phrase into a concrete filter, so an unsupported rider
+                // stays an honest `Unimplemented` rather than a false-green cost.
+                if let Some((count, rest)) = parse_number(&lower).filter(|(n, _)| *n >= 2) {
+                    let filter_text = strip_count_article_prefix(rest.trim())
+                        .trim_end_matches('.')
+                        .trim();
+                    let (filter, remainder) = parse_type_phrase(filter_text);
+                    if remainder.trim().is_empty() && !matches!(filter, TargetFilter::Any) {
+                        costs[i] = match verb {
+                            PrecedingVerb::Sacrifice => {
+                                AbilityCost::Sacrifice(SacrificeCost::count(filter, count))
+                            }
+                            PrecedingVerb::Exile { zone } => AbilityCost::Exile {
+                                count,
+                                zone: extract_filter_zone(&filter).or(zone),
+                                filter: Some(filter),
+                            },
+                            PrecedingVerb::TapCreatures => AbilityCost::TapCreatures {
+                                requirement: TapCreaturesRequirement::count(count),
+                                filter,
+                            },
+                        };
+                    }
+                    // An explicit-count continuation is terminal: only the
+                    // full-consumption/non-`Any` branch above may rehydrate it. If
+                    // that did not fire (an unmodeled rider left a remainder, or an
+                    // `Any` filter), leave the continuation as an honest
+                    // `Unimplemented` — never fall through to the article/count-1
+                    // fallback below, which would emit a broad `count: 1` cost that
+                    // both drops the unmodeled rider and loses the real count.
+                    continue;
+                }
+                // Baseline single-object rehydration (unchanged pre-existing
+                // behavior): a bare "<article> <type>" continuation of the verb
+                // ("Sacrifice a green creature, a white creature, and a blue
+                // creature"). Left exactly as before so this fix does not move it.
                 let stripped = strip_article(description, &lower);
                 if stripped.is_empty() {
                     continue;
@@ -205,24 +254,20 @@ fn fixup_bare_noun_continuations(costs: &mut [AbilityCost]) {
                 if matches!(filter, TargetFilter::Any) {
                     continue;
                 }
-                match last_verb.unwrap() {
+                costs[i] = match verb {
                     PrecedingVerb::Sacrifice => {
-                        costs[i] = AbilityCost::Sacrifice(SacrificeCost::count(filter, 1));
+                        AbilityCost::Sacrifice(SacrificeCost::count(filter, 1))
                     }
-                    PrecedingVerb::Exile { zone } => {
-                        costs[i] = AbilityCost::Exile {
-                            count: 1,
-                            zone,
-                            filter: Some(filter),
-                        };
-                    }
-                    PrecedingVerb::TapCreatures => {
-                        costs[i] = AbilityCost::TapCreatures {
-                            requirement: TapCreaturesRequirement::count(1),
-                            filter,
-                        };
-                    }
-                }
+                    PrecedingVerb::Exile { zone } => AbilityCost::Exile {
+                        count: 1,
+                        zone,
+                        filter: Some(filter),
+                    },
+                    PrecedingVerb::TapCreatures => AbilityCost::TapCreatures {
+                        requirement: TapCreaturesRequirement::count(1),
+                        filter,
+                    },
+                };
             }
             _ => {
                 last_verb = None;
@@ -1670,8 +1715,8 @@ fn parse_mana_cost_nom(
 mod tests {
     use super::*;
     use crate::types::ability::{
-        ControllerRef, DiscardSelfScope, ObjectScope, QuantityExpr, SharedQuality, TypeFilter,
-        TypedFilter,
+        ControllerRef, DiscardSelfScope, FilterProp, ObjectScope, QuantityExpr,
+        SacrificeRequirement, SharedQuality, TypeFilter, TypedFilter,
     };
     use crate::types::counter::CounterMatch;
     use crate::types::mana::{ManaCost, ManaCostShard};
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The bare-continuation fixup can turn unsupported sacrifice fragments into supported sacrifice costs with an over-broad filter. Evidence: `fixup_bare_noun_continuations` rehydrates an `Unimplemented` continuation by calling `parse_single_cost(&format!("sacrifice {description}"))` and then accepts any `AbilityCost::Sacrifice` result (`crates/engine/src/parser/oracle_cost.rs:216-219`). But the sacrifice parser builds its filter through `parse_target(&format!("target {}", filter_text))` and ignores the returned remainder (`crates/engine/src/parser/oracle_cost.rs:548`), so unsupported leftover text can still be accepted as if the full phrase parsed cleanly. This is the same false-green class the comment says it is avoiding: a continuation that should remain `Unimplemented` can become a broad supported cost. Please make the rehydrated sacrifice path validate full consumption / non-`Any` target shape, or tighten the sacrifice parser so unsupported continuations stay unimplemented, and add a regression that asserts the exact parsed cost rather than only the intended happy path.

There is also no current `<!-- coverage-parse-diff -->` evidence on this parser-surface PR. The loop requires that card-level diff before acceptance, because this change can alter parser support beyond Mechtitan Core.

GildardoDev:
Fixed. Added a `bare_continuation_fully_parses` gate that runs before rehydration: it strips the leading count / "any number of" / article, then requires `parse_type_phrase` (the same consumption-aware primitive the exile arm uses) to fully consume the phrase into a concrete, non-`Any` filter. If any leftover text remains, the continuation is left as an honest `Unimplemented` rather than rehydrated. This closes the sacrifice/tap arms' dropped-`parse_target`-remainder path you flagged — an unsupported rider can no longer be silently dropped into a false-green supported cost.

Added a negative regression that asserts the exact parsed cost: `Sacrifice a creature and the biggest creature an opponent controls` keeps the unsupported continuation as `Unimplemented` (and does not fabricate a second sacrifice). The positive tests already assert the exact recovered count + filter. Rebased on latest main; the force-push re-triggers the `coverage-parse-diff` job for the card-level diff.

matthewevans:

matthewevans (changes requested):
**[MED]** The current parser gate failure is a real blocker, not a transient CI issue. Evidence: the `Rust lint (fmt, clippy, parser gate)` job fails in `check-parser-combinators.sh` on added lines in `crates/engine/src/parser/oracle_cost.rs`, specifically `.strip_prefix("any number of ")` and the verbatim `description == "the biggest creature an opponent controls"` check. This parser surface needs to use the project’s nom-combinator patterns rather than new string-dispatch escape hatches.

**[MED]** The fresh parse-diff is still too broad for this fix. Evidence: the current `<!-- coverage-parse-diff -->` sticky reports seven changed cards and added cost signatures such as bare `Artificers`, `Hand of Vecna`, `Reveal the creature type you chose`, `and/or Monks you control`, and `pay its mana cost`. Those look like split-fragment spillover, not just recovery of the intended verb-cost continuation shape. Either those deltas need to be made semantically correct with targeted evidence/tests, or the continuation gate needs to stay narrower so unrelated unsupported fragments remain honest.

GildardoDev:
Both addressed.

**Parser gate:** replaced the `.strip_prefix("any number of ")` with a nom `tag`, and dropped the verbatim `description == "..."` check in the test for a structural `matches!(.., Unimplemented { .. })` assertion. `check-parser-combinators.sh` now passes.

**Parse-diff breadth:** narrowed the rehydration gate to require a determiner-led object phrase — an article, a count word, `another`/`other`, or `any number of`. The split-fragment spillover you cited (bare `Artificers`, `and/or Monks you control`, `Hand of Vecna`, `Reveal the creature type you chose`, `pay its mana cost`) has no leading determiner, so it now stays honest `Unimplemented` instead of being rehydrated — only the intended `<verb> <self/A> and <N> [other] <type> [you control]` continuation is recovered. Added `cost_split_list_fragment_without_determiner_stays_unimplemented` locking that in. The force-push re-triggers the coverage-parse-diff for the narrowed card set.

matthewevans (changes requested):
Requesting changes on current head `6faea539`.

The parser gate is green now, but the current `<!-- coverage-parse-diff -->` sticky updated at `2026-07-03T09:14:17Z` is still broad: 8 cards / 9 signatures. It still includes the unrelated split fragments called out in the previous review (`Artificers`, `Hand of Vecna`, `Reveal the creature type you chose`, `and/or Monks you control`, `pay its mana cost`) and now also includes Greater Gargadon (`creature`, `or land`).

Even if some of those are now being preserved as honest `Unimplemented` fragments rather than rehydrated into supported costs, this PR is still changing parser surface outside the claimed continuation class. For this kind of parser change, the acceptance bar is a current parse-diff that is intentionally scoped, or explicit per-card justification and regressions for every widened/redified card class.

Please either narrow the fix so the current parse-diff only moves the intended determiner-led verb-continuation cases, or make the broader “previously dropped fragment should become unsupported” redification explicit with per-card tests and scope notes for the affected cards.

GildardoDev:
Narrowed the scope so the parse-diff moves only the intended class.

The recovery now fires **only for an explicit count of two or more** (`parse_number >= 2`) — the actual #2343 shape (`four other artifact creatures and/or Vehicles you control`), recovered strictly via `parse_type_phrase` (full-consumption, non-`Any`). Every single-object / fragment continuation now runs the **exact pre-existing baseline** (`parse_target` + `count: 1`) unchanged, so the unrelated cards you cited — `Artificers`, Greater Gargadon (`creature`/`or land`), `Hand of Vecna`, `and/or Monks you control`, `pay its mana cost` — parse identically to main and do not move. I dropped the determiner-gate helpers and the redification they caused.

Net: the fix touches parser surface only for the `<verb> <self/A> and <N>=2 [other] <type> [you control]` continuation class. Added `cost_single_object_continuation_keeps_count_one_baseline` asserting a single-object continuation keeps its baseline count-1 parse. The force-push re-triggers the coverage-parse-diff so you can confirm the narrowed set. Gate/clippy/fmt green, full engine suite 14722/0.

matthewevans:
Holding current head `d2e241a83a696e201f53dad43133463219af0ccc` until the parse-diff evidence refreshes.

The latest `coverage-parse-diff` sticky I can see is still from `2026-07-03T09:14:17Z`, before the latest narrowing push. This PR changes parser behavior in `oracle_cost.rs`, so I need the current-head card-level diff before I can verify that the previous scope-contamination finding is actually resolved.

matthewevans (changes requested):
**[MED]** Explicit-count continuations can still false-green through the legacy count-1 fallback when the new full-consumption path fails.

Evidence: the new explicit-count branch starts at `crates/engine/src/parser/oracle_cost.rs:215` and correctly requires `parse_type_phrase` to consume the full continuation before producing a count-preserving cost. But when that branch sees `count >= 2` and `parse_type_phrase` leaves a remainder, control falls through to the legacy fallback at `oracle_cost.rs:242-246`, which strips only articles, calls `parse_target("target ...")`, ignores the returned remainder, and then emits a count-1 cost as long as the filter is not `Any`.

That undermines the safety property the comment describes: an unsupported explicit-count continuation with an unmodeled rider can still become a supported broad count-1 cost instead of staying `Unimplemented`. The fix should make explicit-count continuations terminal: once `parse_number >= 2` is detected, only replace the cost if the full-consumption/non-`Any` filter branch succeeds; otherwise leave the continuation unsupported. A regression test should cover an explicit-count continuation with unsupported trailing text and assert it remains `Unimplemented` rather than rehydrating as count 1.

GildardoDev:
Good catch — fixed. The explicit-count branch is now **terminal**: once `parse_number >= 2` is detected, the continuation is only replaced when the full-consumption / non-`Any` `parse_type_phrase` branch succeeds; otherwise it `continue`s and stays an honest `Unimplemented` rather than falling through to the article/count-1 fallback. So an explicit-count continuation with an unmodeled rider can no longer become a broad `count: 1` cost that drops both the rider and the real count.

Added `cost_explicit_count_continuation_with_unmodeled_rider_stays_unimplemented`: `Sacrifice a creature and two artifacts that were dealt damage this turn` — the `that were dealt damage this turn` tail is unmodeled, so `parse_type_phrase` leaves a remainder and the continuation stays `Unimplemented` (before this fix it rehydrated as a count-1 sacrifice). Gate/clippy/fmt green, full engine suite 14740/0.

matthewevans:
Holding current head `da4fe3ef6f491d33b9f5168fd3303f053fe0eee8` for fresh current-head parse-diff evidence. The latest push appears to address the explicit-count fallback blocker by making the explicit-count branch terminal, but all current-head checks and the card-data/coverage diff are still queued/running.

Because this PR has previously had scope-contamination in the parser diff, I need the refreshed `coverage-parse-diff` sticky for this exact head before approval.

matthewevans (approved):
Maintainer review: current head is clean for this parser-cost fix. I checked the diff at `fixup_bare_noun_continuations`, the prior blocker is addressed by making explicit-count continuations terminal unless `parse_type_phrase` fully consumes the filter, and the current parse-diff reports no card-parse changes. The change stays at the existing cost-continuation seam and reuses the existing parser/building-block types.

PR #4933, opened 2026-07-02
fix(parser): harden trailing if-gate splitter against "as if" riders

### LINKED ISSUE #4876: fix(parser): harden trailing if-gate combinator on static restrictions (as-if safe)

nickmopen:
Closes #4876.

## Problem

`split_trailing_gate_condition` (added in #4841 for Rock Jockey's `can't play lands if …` gate) used `take_until(" if ")`, which anchors on the **first** ` if ` in the line. On oracle text containing **as if** — e.g. a permission rider like "cast as if it had flash if you control a Swamp" — it mis-splits inside `as if`, attaching the wrong condition or dropping the real trailing gate when the building block is reused on other restriction classes.

## Fix

- **Promoted** `split_trailing_gate_condition` and its `split_trailing_as_long_as` building block from `oracle_static/dispatch.rs` to `oracle_static/shared.rs` (parity with `parse_if_static_condition` / `extract_cant_untap_condition`), so the shared trailing-gate splitter lives with the other condition building blocks.
- The bare-`if` fallback now **walks each ` if ` boundary** (combinator-based `take_until`/`tag` scan) and keeps the **last** one whose preceding word is not `"as"` — skipping `as if` riders. `" as long as "` is still tried first; an unrecognized condition still leaves the line unsupported rather than enforcing unconditionally.

Verified behavior:

| Input (lowercased) | Gate condition |
|---|---|
| `… can't play lands if this creature was cast this turn` | `this creature was cast this turn` |
| `… can't play lands as if it had flash if you control a swamp` | `you control a swamp` |
| `… can't play lands as long as ten or more lands are on the battlefield` | `ten or more lands are on the battlefield` |
| `… can't play lands` | `None` |
| `cast this as if it had flash` (no real gate) | `None` |

## Tests

- New `cant_play_lands_gate_skips_as_if_rider` — an `as if` + trailing `if` line on `can't play lands` must bind the trailing gate.
- Extended the Rock Jockey dispatch test with **card-name-normalized** text ("Rock Jockey was cast this turn").

`cargo test -p engine --lib` → **14,572 passed / 0 failed**; `cargo fmt --all --check`, `cargo clippy -p engine --all-targets -- -D warnings`, and `scripts/check-parser-combinators.sh` all clean.

CR 611.3a (conditional statics) + CR 305.1 (land plays).

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -451,21 +451,6 @@ pub(crate) fn parse_damage_not_removed_during_cleanup(
     )
 }
 
-/// Split a trailing " as long as <condition>" rider off a static line, returning
-/// the condition text when present (combinator form, no string-method dispatch).
-fn split_trailing_as_long_as(lower: &str) -> Option<&str> {
-    opt(preceded(
-        (
-            take_until::<_, _, OracleError<'_>>(" as long as "),
-            tag(" as long as "),
-        ),
-        rest,
-    ))
-    .parse(lower)
-    .ok()
-    .and_then(|(_, condition)| condition)
-}
-
 /// CR 509.1b: "Creatures with power <comparison> <quantity> can't
 /// block this creature." — a can't-be-blocked-by restriction whose blocker
 /// filter gates on a power threshold that may be DYNAMIC (Kraken of the Straits:
@@ -524,26 +509,6 @@ fn filter_prop_has_power_comparison(prop: &FilterProp) -> bool {
         FilterProp::AnyOf { props } => props.iter().any(filter_prop_has_power_comparison),
         _ => false,
     }
-}
-
-/// CR 611.3a: A static restriction may carry a trailing gate introduced by
-/// either `" as long as <condition>"` (continuous) or `" if <condition>"` (state
-/// gate) — e.g. Rock Jockey: "You can't play lands if this creature was cast
-/// this turn." Returns the condition text for `parse_static_condition`. The
-/// `as long as` form is tried first so a card carrying both keywords anchors on
-/// the continuous form; a bare `if` gate is the fallback. As with
-/// `split_trailing_as_long_as`, an unrecognized condition downstream leaves the
-/// line unsupported rather than enforcing the restriction unconditionally.
-fn split_trailing_gate_condition(lower: &str) -> Option<&str> {
-    split_trailing_as_long_as(lower).or_else(|| {
-        opt(preceded(
-            (take_until::<_, _, OracleError<'_>>(" if "), tag(" if ")),
-            rest,
-        ))
-        .parse(lower)
-        .ok()
-        .and_then(|(_, condition)| condition)
-    })
 }
 
 pub(crate) fn parse_static_line_inner(
@@ -2834,7 +2799,7 @@ pub(crate) fn parse_static_line_inner(
         // gates the restriction. If the rider is present but its condition is
         // NOT recognized, leave the whole line unsupported (return None) rather
         // than marking it a CantPlayLand enforced unconditionally.
-        return match split_trailing_gate_condition(tp.lower) {
+        return match super::shared::split_trailing_gate_condition(tp.lower) {
             Some(condition_text) => Some(def.condition(parse_static_condition(condition_text)?)),
             None => Some(def),
         };
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -181,6 +181,64 @@ pub(crate) fn nom_tag_tp<'a>(tp: &TextPair<'a>, prefix: &str) -> Option<TextPair
         })
 }
 
+/// CR 611.3a: Split a trailing `" as long as <condition>"` rider off a static
+/// line, returning the condition text when present (combinator form, no
+/// string-method dispatch). Shared trailing-gate building block alongside
+/// [`split_trailing_gate_condition`].
+pub(crate) fn split_trailing_as_long_as(lower: &str) -> Option<&str> {
+    opt(preceded(
+        (
+            take_until::<_, _, OracleError<'_>>(" as long as "),
+            tag(" as long as "),
+        ),
+        rest,
+    ))
+    .parse(lower)
+    .ok()
+    .and_then(|(_, condition)| condition)
+}
+
+/// CR 611.3a: A static restriction may carry a trailing gate introduced by
+/// either `" as long as <condition>"` (continuous) or `" if <condition>"` (state
+/// gate) — e.g. Rock Jockey: "You can't play lands if this creature was cast
+/// this turn." Returns the condition text for `parse_static_condition`.
+///
+/// The `as long as` form is tried first so a card carrying both keywords anchors
+/// on the continuous form. For the bare `if` fallback the condition is taken
+/// after the LAST trailing `" if "` that is not the tail of an `"as if"`
+/// permission rider — e.g. "… cast as if it had flash if you control a Swamp"
+/// gates on "you control a Swamp", not "it had flash …". Anchoring on the first
+/// `" if "` (as the original inline helper did) would otherwise attach the wrong
+/// condition or drop the real trailing gate.
+///
+/// As with [`split_trailing_as_long_as`], an unrecognized condition downstream
+/// leaves the line unsupported rather than enforcing the restriction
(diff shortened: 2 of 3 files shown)


matthewevans (changes requested):
**[MED]** The new `as if` regression test drives the production `CantPlayLand` parser with a rules-bearing middle rider that the parser does not model or reject. Evidence: the test parses `"You can't play lands as if it had flash if this creature was cast this turn."` through `parse_static_line_multi` (`crates/engine/src/parser/oracle_static/tests.rs:345-348`), while the production arm accepts any line containing `can't play lands` (`crates/engine/src/parser/oracle_static/dispatch.rs:2789-2795`) and only attaches the trailing gate afterward. That means the test succeeds while silently dropping `as if it had flash`. Why it matters: this makes the parser look like it supports a rules-bearing rider it does not actually represent, which is coverage-dishonest support. Please either test `split_trailing_gate_condition` directly as a splitter unit, or use a production static whose middle `as if` rider is fully modeled; production parsing should not accept and erase an unmodeled rider.

This parser-surface PR also still needs the current `<!-- coverage-parse-diff -->` sticky before acceptance, so the card-level blast radius can be checked against the claimed scope.

nickmopen:
Good catch — fixed. Replaced the production-parser test with a **direct splitter-unit test** (`split_trailing_gate_condition_skips_as_if_rider`) that calls `split_trailing_gate_condition` on the building block itself, so nothing drives the `can't play lands` production arm with an `as if it had flash` rider it doesn't model. The unit test asserts the splitter's full contract:

- `"…can't play lands as if it had flash if you control a swamp"` → `Some("you control a swamp")` (skips `as if`, takes the last ` if `)
- plain trailing `if` → unchanged
- `as long as` still precedes a later bare `if`
- an `as if`-only line (no real gate) → `None`
- no gate → `None`

The Rock Jockey dispatch test (with card-name-normalized text) still exercises the real production path on a fully-modeled line.

Also rebased onto current `main` so the `coverage-parse-diff` sticky recomputes against the current base. `cargo test -p engine --lib`, `fmt --check`, `clippy -D warnings`, and `check-parser-combinators.sh` all clean.

matthewevans:
Thanks for the update. The previous false-green concern is addressed on this head: the `as if` case now tests `split_trailing_gate_condition` directly, which is the right building block for this fix.

Holding this for current-head parse-diff evidence before approval. This PR changes parser source (`crates/engine/src/parser/oracle_static/shared.rs` and `dispatch.rs`), but I do not see a `<!-- coverage-parse-diff -->` sticky for head `fdaf7f6d3663a126a11390b1ef37c41810cfa07f`. Once the parse-diff sticky is present for this head, I can verify the card-level blast radius and finish the review.

nickmopen:
### Parse changes introduced by this PR · head `fdaf7f6d3`

✓ **No card-parse changes detected.**

```json
{ "oracle_changed": 1, "added_cards": [], "removed_cards": [], "clusters": [] }
```

Surfacing this manually because the auto-post `workflow_run` sticky job didn't fire for this fork head — the `Card data (generate, validate, coverage)` check *did* run and uploaded the `parse-diff` artifact, so this is the CI-generated result for `fdaf7f6d3`, not a hand-computed one:

- Run: https://github.com/phase-rs/phase/actions/runs/28619287180
- `parse-diff` artifact: https://github.com/phase-rs/phase/actions/runs/28619287180/artifacts/8050895401

Zero card blast radius is expected here: the change is a behavior-preserving move of `split_trailing_gate_condition`/`split_trailing_as_long_as` into `shared.rs` plus defensive `as if` hardening on the bare-`if` fallback. No card in the pool has a `can't play lands … as if … if …` line, and the existing `can't play lands` cards (Rock Jockey single trailing `if`, Limited Resources `as long as`) resolve to the same condition under both the old first-`if` and new last-`if` logic — so no card's parse output changes.

matthewevans:
Thanks for the follow-up. I’m keeping this on hold for the current head because the parse-diff block here was supplied by the PR author, and the review loop only treats trusted workflow- or maintainer-generated parse-diff evidence as authoritative for parser-source PRs.

The no-change result may be correct, but I need trusted evidence for this head before I can clear the prior review. This is especially important for parser splitter changes where the failure mode is silent parse-scope drift.

matthewevans:

matthewevans (changes requested):
Thanks, the parser shape looks otherwise clean. One remaining repo-policy issue blocks this for me:

**[LOW]** New test uses an inline import despite the repo-wide no-inline-import rule. Evidence: `crates/engine/src/parser/oracle_static/tests.rs:346` introduces `use super::shared::split_trailing_gate_condition;` inside the test body. Why it matters: project instructions require imports at file scope for readability/static analysis, and this is directly avoidable. Suggested fix: move the import to the test module imports or call `super::shared::split_trailing_gate_condition(...)` fully qualified.

matthewevans:
Thanks for this, and apologies for the slow turnaround — the work here is correct, but it has been overtaken by another PR.

`fix(parser): harden trailing if-gate combinator on static restrictions (#4876)` (#4877) merged on 2026-07-03 as `c4541bd3`, closing the same issue (#4876) this PR targets. `origin/main` already carries both halves of your change:

- `split_trailing_as_long_as` and `split_trailing_gate_condition` have been moved from `dispatch.rs` into `oracle_static/shared.rs` — your exact refactor.
- The `as if` rider fix is implemented there via the shared building block `nom_primitives::scan_last_valid_at_word_boundaries_with_offset(...)` guarded by `is_as_if_gate_marker(input, if_offset)`, plus a body-preserving sibling `split_trailing_gate_condition_with_body`.

Your implementation is functionally correct — I hand-traced it and it anchors on the last valid ` if `, skips the `as if` rider, and returns `None` when there is no trailing gate. The difference is that main reaches the same behavior through the shared nom scanner rather than a bespoke `while` loop with `before.rsplit(' ').next() != Some("as")`, which is the direction CLAUDE.md's combinator mandate points.

Because main already defines both functions in `shared.rs`, this branch no longer merges cleanly: the conflict resolves into duplicate definitions of `split_trailing_gate_condition` and `split_trailing_as_long_as` that will not compile. Reconciling it would mean deleting your copies and choosing between two competing implementations of the same fix — a design decision rather than a conflict resolution, so I have not pushed anything to your branch.

Recommending this be closed as superseded by #4877. Nothing was wrong with the analysis or the test (`split_trailing_gate_condition_skips_as_if_rider` genuinely discriminates against a first-` if ` anchor) — it just landed twice. If you want to pick up follow-on work in this area, `split_trailing_gate_condition_with_body` was extended again in #5005 and that seam is still active.

matthewevans:
Thank you for this — and I want to be specific about why I'm closing it, because the work itself is sound.

Your change hardens the trailing `if`-gate splitter so an `as if` permission rider (e.g. "cast as if it had flash **if you control a Swamp**") doesn't get mistaken for the gate. That exact hardening already landed on `main` via #4877, which fixed the same issue (#4876). Main's `oracle_static/shared.rs` now carries `split_trailing_gate_condition`, `split_trailing_as_long_as`, and an `is_as_if_gate_marker` guard — the same three pieces you built here, with the `as if` exclusion done at the same boundary. #4877's version additionally carries a UTF-8 char-boundary guard for the em-dash case. That's why this PR now shows as conflicting: main already rewrote these functions.

So this is superseded rather than wrong — your diagnosis of the `as if` false-positive was correct and matches the merged fix. Closing against #4877. Please keep sending parser work; this was the right instinct.

Issue #4886, opened 2026-07-02
[Card Bug] Stuck at choosing Jinnie's replacement effect


Molamola8964:
### Card(s)

Jinnie Fay, Jetmir's Second

### Build/version

v0.12.0

### Game mode

AI

### Actual behavior

Choosing whether to create a dog or cat token would return to asking me to decide Jinnie's replacement effect, causing a loop.

### Expected behavior

After I choose the desired token, the prompt should disappear and the game should make the token.

### Steps to reproduce

Any time Jinnie's replacement effect comes up. No matter what tokens were being made. I have tried several match with AI but I was not able no create a single cat or dog token.

### Logs/screenshots/game-state

https://github.com/user-attachments/assets/2d12fa00-cfee-44e5-b09b-0206693d806e

mike-theDude:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `If you would create one or more tokens, you may instead create that many 2/2 green Cat... or 3/1 green Dog...` (supported: true)

**Reasoning**: Jinnie Fay replacement is parsed as a supported token replacement choice. Returning to the same decision after choosing Cat/Dog indicates the replacement continuation is not consuming/advancing the pending replacement effect, causing a softlock loop.

---
_Posted by bug-coverage-classifier skill._

mike-theDude:
Reviewed against current main. Commit 11eb20f500b3e7f6caacbdef0cd9c6198b7e6d09 consumes the Jinnie Fay token replacement after the player chooses Cat or Dog instead of presenting the same decision again. The exact regression accepts the replacement, chooses Dog, verifies one Dog and no original Treasure, and confirms the prompt does not recur. Please close this issue.

matthewevans:
Closing per mike-theDude's triage above — verified fixed/resolved on current main. If this still reproduces on a current build, please reopen with a fresh repro.

matthewevans:
Reopened from a new Discord report: Jinnie Fay is offered for an opponent’s token creation and then produces no token regardless of the choice.

rykerwilliams:
Appears **already resolved** — flagging because this is labelled `priority:p0-softlock` and still open, which makes it look like an outstanding softlock.

Fixed on `upstream/main` by commit `11eb20f5` (PR #4938), whose body carries two commits naming this issue:

- `fix(engine): stop Jinnie Fay token replacements from looping (#4886)`
[...]

PR #4938, opened 2026-07-02
fix(engine): stop Jinnie Fay replacement from re-prompting (#4886)

ai-mountain:
## Summary
- treat Jinnie Fay-class `CreateToken -> ChooseOneOf(Token, Token)` replacements as full substitutions so the original token event does not also resolve
- carry the originating replacement `applied` set into the chosen substitute token event so the same replacement cannot immediately re-prompt on its own branch
- add a runtime regression plus supporting token/parser checks around the replacement-choice path

## Files changed
- `crates/engine/src/game/effects/choose_one_of.rs` — stop `ChooseOneOf` completion from clearing the inherited token-choice applied seed
- `crates/engine/src/game/effects/mod.rs` — clear the applied seed only at true full-drain (no pending continuation, repeat iteration, or repeat-until), after all of those have drained
- `crates/engine/src/game/effects/token.rs` — read the inherited applied seed on every token proposal emitted by the replacement continuation
- `crates/engine/src/game/engine_replacement.rs` — replacement-continuation plumbing for the token-choice substitution path
- `crates/engine/src/game/replacement.rs` — seed the applied set on the originating token-choice continuation (Jinnie Fay-class)
- `crates/engine/src/types/game_state.rs` — `post_replacement_token_choice_applied` state field + ownership/lifetime documentation

## Track
Developer

## LLM
Model: GLM-5.2
Thinking: high

## Verification
- `cargo test -p engine create_token_choice_replacement_does_not_reprompt_or_create_original_tokens -- --nocapture`
- `cargo test -p engine create_token_replacement_accepted_applies_full_spec -- --nocapture`
- `cargo test -p engine replacement_instead_accepts_untap_and_token_choice_replacements -- --nocapture`
- `cargo clippy --all-targets -- -D warnings`


matthewevans (changes requested):
Thanks. I found one blocker in the current approach.

The inherited replacement-applied set is only seeded as transient game state and then consumed by the next token proposal. That handles the direct Jinnie Fay branch in the test, but it does not cover the full replacement-continuation class: nested `ChooseOneOf` branches are not detected by `ability_tree_creates_tokens`, and a branch that emits a second substitute token would lose the inherited `applied` set after the first token event. That lets the same replacement become eligible again inside its own substitute effect, which is exactly the loop this is trying to prevent.

Please bind the inherited applied set to the replacement continuation / choice context and apply it to every token proposal emitted by that continuation, including nested `ChooseOneOf` branches, then add coverage for a nested or multi-token branch so this cannot regress.

matthewevans (changes requested):
**[HIGH]** The inherited replacement-applied set can still be dropped across an intervening nested replacement choice inside the paused branch chain. Evidence: branch token proposals read `state.post_replacement_token_choice_applied` at `crates/engine/src/game/effects/token.rs:450`, but `continue_replacement_impl` overwrites that field to `None` for any nested replacement whose own post-effect is not another token-choice shape at `crates/engine/src/game/replacement.rs:6750`. `ReplacementChoice` is one of the resolution-pausing states (`crates/engine/src/game/effects/mod.rs:1812`), and sub-abilities are stashed across such pauses (`crates/engine/src/game/effects/mod.rs:7101`), so a selected Jinnie branch that creates a token, hits another optional replacement choice, then resumes a later token sub-ability loses the original replacement id and can re-prompt the same Jinnie replacement. Why it matters: this is the same replacement-continuation loop class as the earlier blocker, just through an intervening `ReplacementChoice` instead of only nested `ChooseOneOfBranch`. Suggested fix: carry the inherited applied seed on the replacement continuation / pending resolution context and clear it only when the originating token-choice continuation fully drains, then add a regression with a branch token that pauses on another replacement choice before a later token sub-ability.

matthewevans (changes requested):
**[HIGH]** The inherited applied seed can still be cleared by an intervening nested choice before the later token sub-ability drains. Evidence: the seed is global on `GameState` at `crates/engine/src/types/game_state.rs:5967`, and the originating replacement seeds it at `crates/engine/src/game/replacement.rs:6760` when a token replacement parks a token-choice `ChooseOneOf`. But `ChooseOneOf` completion clears that global seed whenever `waiting_for` is back to priority at `crates/engine/src/game/effects/choose_one_of.rs:161`; then the `ChooseBranch` action handler drains any stashed continuation only afterward at `crates/engine/src/game/engine_resolution_choices.rs:2388`. A branch shaped like `ChooseOneOf(non-token choice) -> sub_ability Token` is production-reachable because `resolve_ability_chain` stashes the sub-ability while the nested choice is waiting (`crates/engine/src/game/effects/mod.rs:7101`). When the nested choice resolves, line 161 clears the Jinnie seed before line 2388 drains the token continuation, so the later token proposal loses the already-applied replacement id and can re-prompt the same replacement. Suggested fix: carry the seed on the originating pending choice/continuation frame, or otherwise clear only after that frame and its stashed continuation have drained; add a regression where a Jinnie-class branch pauses on a non-token `ChooseOneOf` before a later token sub-ability.

matthewevans (changes requested):
**[MED]** The token-choice applied seed still clears before a paused `repeat this process` continuation can re-enter and emit later token proposals. Evidence: `crates/engine/src/game/effects/mod.rs:673` gates the full-drain cleanup only on `pending_continuation` and `pending_repeat_iteration`, then clears `post_replacement_token_choice_applied` at `crates/engine/src/game/effects/mod.rs:685` before draining `pending_repeat_until` at `crates/engine/src/game/effects/mod.rs:686`; that drain can immediately re-run the stashed ability through `resolve_ability_chain` at `crates/engine/src/game/effects/mod.rs:721` and `crates/engine/src/game/effects/mod.rs:744`. Why it matters: a Jinnie-class token-choice branch that pauses inside a repeat-until process can lose the inherited replacement id before a later repeated token proposal, reopening the same self-replacement loop class. Suggested fix: treat `pending_repeat_until` as part of the originating continuation frame and clear the seed only after that repeat continuation has fully drained or stopped; add a regression with a token-choice branch that pauses inside a repeat-until loop before a later token proposal.

matthewevans (changes requested):
**[MED]** This PR still carries unrelated parser-backlog doc changes. Evidence: `docs/parser-misparse-backlog.md` now changes the global totals at lines 5-7 and root-cause 31 at line 5221, but #4938 is a Jinnie Fay replacement-loop engine fix; those backlog edits remove the Rush of Dread / The Goose Mother entries that were handled separately in the parser backlog work. Why it matters: it contaminates the engine fix with unrelated ledger churn, which makes the PR harder to review and risks duplicate/conflicting doc accounting. Suggested fix: drop `docs/parser-misparse-backlog.md` from this PR and keep this diff scoped to the replacement-loop files.

I did check the current repeat-until seed-clearing change as well. The prior `pending_repeat_until` concern appears addressed by draining repeat-until before clearing `post_replacement_token_choice_applied` in `crates/engine/src/game/effects/mod.rs` lines 677-697, but the PR still has a queue-safety proof gap because the AI contributor template proof sections are missing. Please keep the behavioral proof on the PR current after removing the unrelated doc edit.

RenzoMXD:
@matthewevans — thanks for confirming the repeat-until fix.

On the `docs/parser-misparse-backlog.md` finding: I checked this against the actual merge-base rather than a tip-vs-tip comparison, and this PR's diff did not touch that file. `gh pr view --json files` (and a `git diff` against both this fork's `main` and `upstream/main` at the real merge-base, `ad98b541e`) showed only the 6 engine files below — no backlog doc.

What happened: this branch's merge-base (`ad98b541e`, "Remove resolved root31 backlog entries") already reflected 31 root causes. `upstream/main` had since moved 14 commits further and independently trimmed the backlog to 30 (removing the Rush of Dread / Goose Mother entries) in a later, unrelated commit. Comparing this branch's tip against current `main`'s tip — rather than against the shared merge-base — showed that later main-side reduction as if this PR were reverting it, when it wasn't.

I've merged current `upstream/main` into the branch (now at `7d998ac3c`) to remove any ambiguity — `docs/parser-misparse-backlog.md` is now byte-identical to `main` and won't appear in any diff regardless of comparison method. No conflicts.

Files touched by this PR (post merge-base):
- `crates/engine/src/game/effects/choose_one_of.rs`
- `crates/engine/src/game/effects/mod.rs`
- `crates/engine/src/game/effects/token.rs`
- `crates/engine/src/game/engine_replacement.rs`
- `crates/engine/src/game/replacement.rs`
- `crates/engine/src/types/game_state.rs`

Ready for re-review on the current head.

matthewevans (changes requested):
Thanks for the correction on the parser-backlog doc finding. I rechecked the merge-base diff for this PR and agree that `docs/parser-misparse-backlog.md` is not in `origin/main...HEAD`; that prior objection should not block this PR.

I found two remaining lifecycle blockers in the current replacement-frame approach:

**[MED]** Token-choice seeding misses a `ChooseOneOf` whose token is in the choice's own tail. Evidence: `crates/engine/src/game/replacement.rs:300` only classifies `ChooseOneOf` by scanning its branches, while `crates/engine/src/game/replacement.rs:6840` only seeds through that predicate; `crates/engine/src/game/effects/mod.rs:7208` can stash a paused effect's `sub_ability`, and `crates/engine/src/game/engine_resolution_choices.rs:2550` drains that tail after `ChooseBranch`. Why it matters: a `CreateToken -> ChooseOneOf(non-token branches).sub_ability(Token)` replacement will not seed the applied set, so the tail token can re-prompt/reapply the originating token replacement. Suggested fix: classify the whole root `AbilityDefinition` for token creation, not only the branches, or carry the applied set on the owned continuation frame itself; add a test for this tail-token shape.

**[MED]** The new global seed is not cleared when replacement/choice teardown abandons the normal drain path. Evidence: `crates/engine/src/types/game_state.rs:6032` adds `post_replacement_token_choice_applied`, normal clearing only happens in `crates/engine/src/game/effects/mod.rs:692`, and the elimination teardown clears adjacent replacement continuation state but not this field at `crates/engine/src/game/elimination.rs:487`. Why it matters: if a live token-choice replacement frame is abandoned before normal full-drain, later token proposals inherit a stale applied set via `crates/engine/src/game/effects/token.rs:450`, suppressing replacements that should apply. Suggested fix: clear the seed anywhere the owning replacement/choice continuation is discarded, or move the seed into the owned pending continuation frame so teardown cannot leave a global latch behind.

matthewevans:
Re-reviewed the latest head. The two most recent requested-change items look addressed in the code: the token-choice classifier now walks the full ability tree including tail `sub_ability`/`else_ability`, and the post-replacement abandonment cleanup now clears the token-choice applied seed through a centralized helper.

I am holding approval until the current CI evidence is complete for this head. At review time the current card-data/parse-diff artifact was not available yet and Rust/card-data checks were still pending.

matthewevans:
I updated the branch to refresh the stale merge-base evidence: the previous parse-diff was comparing against an old base and included unrelated parser/card-data changes despite this PR's current diff being engine-only.

The substantive head reviewed clean for the latest lifecycle issues, but I am holding approval on the new merge head until the refreshed CI and parse-diff/card-data evidence complete.

matthewevans:
Saw the new merge-only head `55c2696903e933a365e25526a48350999fbd3fde`. The prior substantive review remains clean, and the refreshed parse-diff still reports no card-parse changes, but all CI jobs restarted for this head, so I am holding approval until those checks complete on the current commit.

matthewevans (approved):
Maintainer sign-off after re-review: current head is merge-only from the previously reviewed clean implementation, parse-diff reports no card-parse changes, and the current required checks are green.

PR #4939, opened 2026-07-02
Parse exile target of "exile N cards from your graveyard" for N>6 (Ultimecia, Emeritus of Ideation)

e11734937-beep:
## Summary

**Ultimecia, Time Sorceress** and **Emeritus of Ideation** both read *"…exile eight cards from your graveyard. If you do, …"*, but the exile recipient was silently dropped to `TargetFilter::Any` (a `target-fallback`), so the "eight cards from your graveyard" zone filter was lost.

Cause: `parse_target`'s `QUANTIFIED_PREFIXES` list (which strips a leading spelled-out/numeric count so the underlying noun phrase parses) only covers bare numbers up to `"six "`. `"five cards from your graveyard"` (Kroxa, Aegis Sculptor, Young Necromancer) already works — this is a pure parser gap; runtime is unchanged.

## Change

Rather than extend the bare-number prefix list (a **trap** — a bare `"seven "` prefix would strip the leading word of a *number-named card* used as a copy source, mis-parsing **Princess Snowfall**'s "copy of **Seven Dwarves**"), add an **anchored** recognizer ahead of `QUANTIFIED_PREFIXES` (`crates/engine/src/parser/oracle_target.rs`): a `parse_number` immediately followed by the generic `"card"/"cards"` noun. The count discard is thus scoped to genuine "N cards" quantities, and number-named cards stay whole.

After the fix, the exile target parses to `Typed(Card, controller = You, [InZone Graveyard])` — byte-identical to the supported "five cards from your graveyard" sibling.

## Tests

- `quantified_count_before_card_noun_parses_as_zone_filter` — seven..twelve strip to the graveyard filter; includes a **Seven Dwarves collision guard** (number-named card must stay `Any`/unconsumed).
- `ultimecia_exile_eight_from_graveyard_recipient_not_dropped` — real card; asserts the exile recipient is the graveyard filter, not a dropped `Any`.

Fails-before / passes-after confirmed. Green: rustfmt, parser-combinator (Rule Zero) gate, `clippy -D warnings`, full engine suite (14580 passed / 0 failed). Corpus (built against a fresh origin/main baseline): **net +2** (Ultimecia, Time Sorceress; Emeritus of Ideation gained), **0 regressions**, `target-fallback` 53→51, Princess Snowfall confirmed still an honest fallback (not mis-parsed).

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -479,6 +479,29 @@ pub fn parse_target_with_syntax<'a>(
         return parse_target_with_syntax(original_rest, ctx);
     }
 
+    // CR 107.1 + CR 400.7: A fixed spelled-out count before the generic
+    // "card"/"cards" noun (Ultimecia, Time Sorceress and Emeritus of Ideation:
+    // "exile eight cards from your graveyard"). The `QUANTIFIED_PREFIXES` list
+    // below only enumerates bare counts through "six", and it is deliberately NOT
+    // extended for higher counts: a bare "seven "/"ten " prefix would also strip
+    // the leading word of a number-named card used as a copy/target source ("a
+    // copy of Seven Dwarves" -> the Dwarf subtype, "Ten Wizards Mountain"), a
+    // silent mis-parse. Anchoring the strip to the generic card noun keeps the
+    // count discard scoped to genuine "N cards" quantities: the count is dropped
+    // here (callers that need it read it separately) and the recursion parses the
+    // "card[s] from <zone>" remainder as the underlying zone filter, matching the
+    // already-supported five-card siblings (Kroxa and Kunoros, Young Necromancer).
+    if let Ok((rest, _)) = (
+        nom_primitives::parse_number,
+        space1,
+        peek(alt((tag::<_, _, OracleError<'_>>("cards"), tag("card")))),
+    )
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
Thanks. This needs to stay unsupported until the count is preserved through resolution.

The new target parser recognizes `eight cards from your graveyard`, but it discards the `eight` and returns only the graveyard-card filter. The resulting `ChangeZone` still has no count/cardinality metadata, so resolution falls back to selecting a single eligible card. That makes Ultimecia/Emeritus parse green while still implementing the cost incorrectly.

Please thread the parsed cardinality at the exile imperative seam into the resolution-time selection/payment behavior, or keep coverage red with an explicit unsupported marker until the engine can require and move exactly eight cards. Also use the existing bounded `card`/`cards` helper for the anchor so this does not match a longer word by prefix.

e11734937-beep:
Withdrawing — you're right that this parses green while implementing the cost incorrectly (the count is dropped, so resolution moves a single card instead of eight). Correctly supporting it needs the parsed cardinality threaded through the exile imperative into resolution-time selection/payment, which is an engine change beyond this parser-only PR. Better to keep the card honestly unsupported than ship a parse-green/run-wrong fix. Will revisit if/when the count is preserved end-to-end.

PR #4940, opened 2026-07-02
fix(engine): scope Iona's chosen-color cast prohibition to the chosen color

### LINKED ISSUE #4937: [Card Bug] Iona, Shield of Emeria locks opponents out of every color, not just the chosen one
### Card(s)

Iona, Shield of Emeria (and any card with "can't cast spells of the chosen color").

### Build/version

`main` (reproduced against current `main`).

### Game mode

AI / Solo (rules-engine behavior, transport-independent).

### Actual behavior

Iona's static reads “Your opponents can't cast spells of **the chosen color**.” After a color is chosen, opponents are prevented from casting **spells of every color** (and colorless spells) — a total cast-lock. Choosing blue also stops opponents from casting their red, white, black, green, and colorless spells.

### Expected behavior

Only spells whose colors include the **chosen** color should be prohibited (CR 105.2 / CR 105.4). If blue is chosen, opponents can't cast blue spells but can freely cast spells of any other color, exactly like the physical card.

### Repro

1. Put Iona, Shield of Emeria onto the battlefield under player A and choose a color (e.g. blue).
2. As opponent B, try to cast a spell that is **not** the chosen color (e.g. a red spell).
3. Observe: casting is disallowed, even though the spell is not the chosen color.

### Root cause (for triage)

In `crates/engine/src/parser/oracle_static/restriction.rs`, the `"spells of the chosen color"` branch of `parse_cant_cast_type_spells` builds a `StaticMode::CantBeCast` with **no `affected` filter** — unlike its sibling branches `"spells of the chosen type"` (attaches `FilterProp::IsChosenCardType`) and `"spells with the chosen name"` (attaches `TargetFilter::HasChosenName`). An unfiltered `CantBeCast` prohibits every spell.

Additionally, even if the parser attached `FilterProp::IsChosenColor`, the runtime resolver `cant_cast_filter_matches` in `crates/engine/src/game/casting.rs` has a dedicated arm for `IsChosenCardType` (it needs the source permanent's chosen attribute) but **none** for `IsChosenColor` — so the delegated `spell_record_matches_filter` path returns `false` for it. A complete fix needs both the parser filter and a matching runtime arm.

The existing parser test `cant_cast_spells_of_chosen_color` asserts only `def.mode` and never inspects `def.affected`, which is why the dropped filter went unnoticed.

ultrahighsuper:
## Summary

**Iona, Shield of Emeria** — "Your opponents can't cast spells of **the chosen color**" — parsed to an *unfiltered* `CantBeCast`, locking opponents out of spells of **every** color (and colorless) instead of only the chosen one. This scopes the prohibition to the chosen color, matching the physical card. Fixes #4937.

The bug had two halves, both fixed here (mirroring the existing `"chosen type"` / `"chosen name"` sibling branches):

- **Parser** (`parser/oracle_static/restriction.rs`): the `"spells of the chosen color"` branch of `parse_cant_cast_type_spells` built `CantBeCast` with no `affected` filter. It now attaches `FilterProp::IsChosenColor`, exactly like the adjacent `"chosen type"` branch attaches `IsChosenCardType`.
- **Runtime** (`game/casting.rs`): `cant_cast_filter_matches` had a dedicated arm for `IsChosenCardType` (it needs the source permanent's chosen attribute) but none for `IsChosenColor`, so the delegated `spell_record_matches_filter` path returned `false` for it — i.e. a parser-only fix would make Iona prohibit *nothing*. Added the matching arm, resolving the spell's colors (CR 105.2) against the source's chosen color (CR 105.4).

> Small, surgical fix that slots into an existing `alt`-style branch cluster (it mirrors the two sibling branches immediately above it and the sibling runtime arm) rather than introducing a new pattern or variant. A final `/review-impl` pass was run and its notes addressed; the passive-voice "Spells of the chosen color can't be cast" form is intentionally left out of scope (no printed card uses it, so leaving it unimplemented stays coverage-honest).

## CR references

- `CR 101.2` — a "can't" effect takes precedence (the prohibition).
- `CR 105.2` — an object's color(s) come from its mana symbols (matching the spell).
- `CR 105.4` — choosing a color (Iona's chosen color).

## Verification

Toolchain used directly (Tilt not running in this environment):

- `cargo fmt --all` — clean.
- New end-to-end regression test `crates/engine/tests/issue_4937_iona_chosen_color.rs` drives production `can_cast_object_now`: with Iona (chosen color = blue) on the battlefield, an opponent **can't** cast a blue spell but **can** still cast a red spell. Passes with the fix; **fails without it** (revert-probe — reverting either half makes the off-color spell wrongly prohibited).
- Strengthened the existing `cant_cast_spells_of_chosen_color` parser unit test to assert the `IsChosenColor` filter is attached (previously it checked only `def.mode`, which is why the dropped filter went unnoticed).

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -14922,7 +14922,7 @@ fn is_blocked_by_cant_be_cast(
 
         // CR 604.1: Check spell filter if present.
         if let Some(ref affected) = def.affected {
-            if !cant_cast_filter_matches(state, spell_obj, affected, bf_obj) {
+            if !cant_cast_filter_matches(state, spell_obj, affected, bf_obj, caster) {
                 continue;
             }
         }
@@ -14946,18 +14946,31 @@ fn is_blocked_by_cant_be_cast(
 
 /// CR 101.2: Check if a spell matches a CantBeCast affected filter.
 /// Handles type filters, mana value comparisons, chosen name, and chosen card type.
-/// Source-dependent filters (HasChosenName, IsChosenCardType) are resolved here
-/// because they need the source permanent's chosen attributes.
+/// Evaluate a `CantBeCast` affected filter against a spell being cast, with the
+/// prohibiting permanent as the filter source.
+///
+/// Only `HasChosenName` needs a dedicated arm — the shared spell-filter matcher
+/// has no top-level chosen-name variant. Every other filter, including the
+/// chosen-attribute *properties* (`IsChosenColor` per CR 105.2/105.4,
+/// `IsChosenCardType` per CR 205), is evaluated as a normal typed-filter
+/// conjunction through the source-aware `spell_object_matches_filter_from_state`
+/// path. That path resolves each chosen property against the source permanent's
+/// chosen attributes from context, so a prohibition can combine a chosen
+/// attribute with any card-type, controller, or zone axis without a bespoke
+/// per-property matcher here.
 fn cant_cast_filter_matches(
     state: &GameState,
     spell_obj: &super::game_object::GameObject,
     filter: &TargetFilter,
     source_obj: &super::game_object::GameObject,
+    caster: PlayerId,
 ) -> bool {
-    use crate::types::ability::{ChosenAttribute, FilterProp};
+    use crate::types::ability::ChosenAttribute;
 
     match filter {
-        // CR 201.2: "spells with the chosen name" — match spell name against source's chosen name.
+        // CR 201.2: "spells with the chosen name" — the shared spell-filter path
+        // has no top-level chosen-name variant, so match the spell name against
+        // the source's chosen name here.
         TargetFilter::HasChosenName => {
             let chosen_name = source_obj.chosen_attributes.iter().find_map(|a| match a {
                 ChosenAttribute::CardName(n) => Some(n.as_str()),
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
Thanks for the clean repro. The parser shape and the Iona runtime test are pointed in the right direction, but this needs one seam fix before it can land.

**[MED]** `IsChosenColor` is handled as a whole-filter shortcut in `cant_cast_filter_matches`, bypassing the existing typed-filter matcher and ignoring any other axes on the filter. Evidence: `crates/engine/src/game/casting.rs:14993` returns after checking only `spell_obj.color.contains(&chosen_color)`, while `crates/engine/src/game/filter.rs:2561` already evaluates `TypedFilter` as `type_filters && properties`, and `crates/engine/src/game/filter.rs:2688` already has the source-context implementation for `FilterProp::IsChosenColor`. Why it matters: this fixes Iona's current one-property filter, but it proliferates a second chosen-color matcher and would over-block any future/source-produced `CantBeCast` filter that combines chosen color with a card type, controller, zone, or other property. Suggested fix: route `CantBeCast` affected filters through the existing source-aware spell filter path (`spell_object_matches_filter_from_state` or equivalent source-context property evaluation) so `IsChosenColor` is just one property in the normal typed-filter conjunction.

ultrahighsuper:
Addressed — thanks, that's the right call. Pushed `4651b4fa0`.

`cant_cast_filter_matches` no longer has a bespoke chosen-color matcher. `CantBeCast` affected filters now route through the shared source-aware `spell_object_matches_filter_from_state`, so `IsChosenColor` is evaluated as one property in the normal `type_filters && properties` conjunction (its source-context implementation at `filter.rs:2688` reads the source permanent's chosen color). A prohibition that combines chosen color with a card type, controller, or zone now conjuncts correctly instead of short-circuiting on color alone.

While there, I folded the former `IsChosenCardType` arm into the same path too (its property matcher at `filter.rs:2701` is a superset — it also resolves labeled card-type choices via `chosen_card_type()`), leaving only `HasChosenName` as a dedicated arm since the shared path has no top-level chosen-name variant. Net effect is a smaller `cant_cast_filter_matches` with no per-chosen-property matchers.

Verification: the Iona end-to-end test still passes through the new path; `cant_cast` (47), `game::filter` (117), and chosen-* (262) unit tests are green; fmt + `clippy -p engine --lib` clean.

matthewevans (approved):
Reviewed current head: the chosen-color prohibition now uses the shared source-aware spell-filter path, the parser diff is scoped to Iona, and the runtime regression covers chosen-color vs off-color castability.

PR #4973, opened 2026-07-02
feat(parser): modal "As ~ enters, it becomes your choice of" as-enters replacement

real-venus:
## Summary

Implements the **CR 208.2b** modal as-enters replacement — "As ~ enters, it becomes your choice of [mode A], [mode B], (or) [mode C]" — where the controller chooses among several P/T + type + keyword profiles as the creature enters. Unlocks a class of previously-unsupported cards:

- **Primal Plasma** — 3/3, or 2/2 flying, or 1/6 defender
- **Primal Clay** — additive ("in addition to its other types"): keeps Artifact, gains Wall/Defender/etc.
- **Corrupted Shapeshifter** — 3/3 flying, 2/5 vigilance, or 0/12 defender (Devoid parses independently)
- **Aquamorph Entity** — 5/1 or 1/5 (enters branch; see scope note)

## Approach — compose, don't add a new path

There is **no new interactive path**. The choice presentation, persistence, deferred-entry timing (CR 614.12a), AI enumeration, and frontend modal all already exist (the anchor-word as-enters cards exercise them). This PR adds **one parser recognizer** that emits those existing primitives:

- A classifier predicate `is_as_enters_becomes_choice_pattern` routes the "your choice of" form (the existing `is_as_enters_choose_pattern` only matches the verb "choose ").
- The recognizer emits a `Moved` replacement whose execute is `Effect::Choose { ChoiceType::Labeled, persist: true }`, plus one `StaticCondition::ChosenLabelIs`-gated continuous `StaticDefinition` per mode. `synthesize_mode_label` is the single authority feeding both the options and the gates; duplicate labels abort emission (honest gap).
- Each profile is decomposed by the shared `parse_animation_spec`; Primal Clay's additive marker is handled via `locate_in_addition_other_types_marker` (CR 205.1b — retains the printed Artifact type).

**Zero new engine enum variants, zero AI/frontend/serialization surface.**

## Scope note

Aquamorph Entity's "**or is turned face up**" branch is scoped out as an honest `Effect::unimplemented("modal-enters-face-up", …)` gap (CR 614.1e) — `TurnFaceUp` + interactive choice is a separate, unproven path. The enters branch is fully covered.

## Rules

CR 208.2b (governing) + CR 614.1c + CR 614.12a + CR 205.1b. The chosen profile is modeled as a `Duration::Permanent` Layer-7b continuous effect gated on the chosen label, not as a copiable-value modification (CR 707.2) — a pre-existing limitation shared with `parse_as_enters_becomes`, annotated honestly.

## Testing

8 runtime tests drive the real `ChooseOption` → label-persist → `evaluate_layers` path across all 4 cards, including a **discriminating** gated-`SetPower` proof (fixture creature has printed 0/0, so a passing 2/2 can only come from the chosen mode's gated static) and mode-exclusivity assertions. Plus parser-shape and negative-sibling tests (single fixed-mode "in addition", Typed non-self, Mercurial Transformation all route correctly). `cargo test -p engine --lib`: 14,684 passed, 0 regressions.


matthewevans (changes requested):
Thanks for the detailed implementation. I think this is close, but I don't think we can land it until the class-level behavior is covered more directly.

**[MED] The runtime tests bypass the replacement path this PR adds.** `place_and_choose` puts the object directly onto the battlefield with `add_creature_from_oracle` and then manually assigns `runner.state_mut().waiting_for = WaitingFor::NamedChoice` before calling `ChooseOption` (`crates/engine/tests/modal_enters_becomes_choice.rs:59`, `:64`). That proves the label persistence and gated static can work after a hand-built choice, but it does not prove the new `ReplacementDefinition::new(ReplacementEvent::Moved)` actually fires during a real entry, filters on `SelfRef`/`Zone::Battlefield`, pauses at the CR 614.12a deferred-entry point, and then resumes with the label already persisted (`crates/engine/src/parser/oracle_replacement.rs:2008`). This would still pass if the actual enters replacement were never wired into zone movement. Please add at least one production-path regression where the card starts off the battlefield, enters through the replacement-aware movement/cast path, the engine produces the named choice itself, and the selected label then drives the layer result.

**[MED] The synthesized mode key drops characteristics that the mode actually applies.** The comment says the label identity is P/T plus additional characteristics, and the mode modifications include colors, supertypes, card types, and subtypes via `animation_modifications` (`crates/engine/src/parser/oracle_effect/animation.rs:214`, `:222`, `:232`). But `synthesize_mode_label` only includes P/T and keywords (`crates/engine/src/parser/oracle_replacement.rs:1836`). Because duplicate labels abort the whole lowering (`:1998`), two legal modes with the same P/T/keyword text but different type/subtype/color characteristics would collide and be treated as an honest gap even though the parser has enough structured data to distinguish and apply them. Please make the mode key include every characteristic axis that participates in the generated `mode_mods` (or otherwise preserve a stable per-profile identity), and add a collision test where same P/T modes differ by type/subtype/color.

matthewevans:
Re-reviewed the latest head (`e4e6e345`). The two blockers from my prior review look addressed on the backend side:

- the replacement path is now exercised through the real cast/entry/`ChooseOption` flow rather than by manually constructing `waiting_for`, and
- the synthesized labels now include the extra characteristic axes needed to avoid collisions between structurally different modes.

I'm holding approval for one process/evidence reason: the parse-diff comment is still from `2026-07-02T22:39:57Z`, while this head was updated afterward. Because this PR changes parser/card coverage, I need a fresh parse-diff for the current head before approving/enqueueing.

matthewevans (commented):
Current-head hold on `e4e6e345`: I’m not approving/enqueueing yet because the required parser evidence is not usable for this head/base.

The sticky parse-diff body is still comparing against `main` baseline `c2c0162b1427` and includes unrelated parser deltas outside this PR’s modal-as-enters scope, including `unless_payment`, counter/removal changes, trap alternative-cost changes, and Cairn Wanderer-style graveyard keyword statics. The local diff for this PR is the modal-as-enters implementation, so I need a regenerated/current parse-diff whose card-level changes are scoped to that class before clearing the stale changes-requested review.

No approval/enqueue from the sweep until that evidence is clean.

matthewevans:
Holding review on current head `5e12c471`: the required parser evidence is still not scoped enough to clear the stale changes-requested state.

The sticky parse-diff now reflects this head, but it still reports broad unrelated parser deltas outside the modal as-enters replacement class, including `unless_payment`, generic counter/removal changes, trap alternative-cost changes, conditional attack/block changes, and Cairn Wanderer-style graveyard keyword statics. The local PR diff is the modal as-enters implementation (`oracle_replacement`, classifier dispatch, named-choice layer invalidation, and modal-enters tests), so I still need a regenerated/current parse-diff whose card-level changes are attributable to that modal-as-enters class before approving or enqueueing.

matthewevans (approved):
Maintainer review: current head resolves the prior blockers. The parse-diff sticky is current for this head and scoped to the modal as-enters replacement class, the mode labels now include the characteristic axes they gate, and the added production-path tests drive real cast/entry choice behavior rather than a hand-built waiting state.

PR #4977, opened 2026-07-02
fix(parser): gate self keyword-grants on trailing "unless <source state>" (Tadeas, Juniper Ascendant)

minion1227:
## Problem

A self-referential keyword grant gated by a trailing `unless it's <attacking|blocking|tapped|…>` rider parsed to a `static_structure` gap and silently lost its gate:

- Tadeas, Juniper Ascendant: `~ has hexproof **unless it's attacking**.` → no static (dropped)

The self keyword-grant dispatch had `if` and `as long as` arms but **no `unless` arm**, and `parse_continuous_gets_has` split ` as long as ` (resolving the self-pronoun `it's attacking` → `SourceIsAttacking` via `rewrite_self_pronoun_subject`) but had **no ` unless ` split**. So `unless <source state>` reached `parse_continuous_gets_has` with the rider intact and nothing handled it → the whole grant failed.

Confirmed the whole class was affected, not just Tadeas: `~ has hexproof unless it's tapped`, `~ has flying unless it's blocking`, etc. all failed identically, while `unless you control X` (board condition) and `as long as it's attacking` already worked.

## Fix

Add an ` unless ` split to `parse_continuous_gets_has` (test-free `oracle_static/anthem.rs`) mirroring the existing ` as long as ` block (**CR 611.3a**): recurse on the body, parse the condition — applying the SelfRef-only self-pronoun rewrite so `it's attacking` binds to the source — and wrap it in `Not` (an `unless X` grant applies precisely when X is **false**).

A balanced-double-quote guard keeps a granted ability's own inner `unless` (`gains "counter target spell unless its controller pays {1}"`) with its quoted text.

### Before / after (Tadeas)

```
before: (no static — static_structure gap, hexproof gate dropped)
after:  Continuous / AddKeyword(Hexproof)  condition: Not(SourceIsAttacking)
```

## Scope / safety

- Covers the whole self keyword/pump-grant `unless <source state>` class (attacking / blocking / blocked / tapped / etc.).
- **Unchanged:** board-condition `unless` (`Not(IsPresent)`), `as long as`, plain grants, attached-subject (`enchanted creature … unless it's attacking` correctly keeps its host binding as an honest gap rather than a wrong self-rewrite), and quoted granted-ability inner `unless`.

## Tests

- New `self_keyword_grant_unless_source_combat_state_gate` regression: asserts the Continuous `AddKeyword(Hexproof)` gate is `Not(SourceIsAttacking)` with no `static_structure` residue.
- 949 `oracle_static` tests pass; `cargo fmt` + `clippy -D warnings` clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -799,6 +799,47 @@ pub(crate) fn parse_continuous_gets_has(
         }
     }
 
+    // CR 611.3a: Split a trailing " unless [condition]" gate, mirroring the
+    // " as long as " form above. An "unless <cond>" rider grants the modification
+    // precisely when <cond> is FALSE, so the parsed condition is wrapped in `Not`
+    // (Tadeas, Juniper Ascendant: "has hexproof unless it's attacking" → AddKeyword
+    // gated on Not(SourceIsAttacking)). Only peel when the split sits OUTSIDE a
+    // quoted granted ability — a granted ability's own inner "unless" (e.g. "gains
+    // 'counter target spell unless its controller pays {1}'") must stay with the
+    // quoted text; balanced double quotes in the body signal the split is outside
+    // any "...". As with the " as long as " form, the self-pronoun condition
+    // subject ("it's attacking"/"it's tapped") is resolved to the source only for
+    // SelfRef grants — an attached-subject "it" keeps its enchanted/equipped
+    // binding and stays an honest gap.
+    if let Some((before_cond, after_cond)) = tp
+        .split_around(" unless ")
+        .filter(|(body, _)| body.original.chars().filter(|&c| c == '"').count() % 2 == 0)
+    {
+        let continuous_text = before_cond.original;
+        let condition_text = after_cond.original.trim().trim_end_matches('.');
+        if let Some(mut def) =
+            parse_continuous_gets_has(continuous_text, affected.clone(), description)
+        {
+            let typed = if matches!(affected, TargetFilter::SelfRef) {
+                parse_static_condition(&rewrite_self_pronoun_subject(condition_text))
+            } else {
+                parse_static_condition(condition_text)
+            };
+            let condition = match typed {
+                Some(inner) => StaticCondition::Not {
+                    condition: Box::new(inner),
+                },
+                None => StaticCondition::Not {
+                    condition: Box::new(StaticCondition::Unrecognized {
+                        text: format!("unless {condition_text}"),
+                    }),
+                },
+            };
+            def.condition = Some(condition);
+            return Some(def);
+        }
+    }
+
     // CR 613.4c: Handle repeated dynamic pump terms — "gets +N/+M for each X and
     // +P/+Q for each Y" (Eidolon of Countless Battles) — where each term scales
     // by its own count. Try this before the single-"for each" path so the second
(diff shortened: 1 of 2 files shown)


matthewevans:
Sweep status: source review is clean so far on current head. The change is at the existing `parse_continuous_gets_has` gate seam, the new test covers the self keyword-grant `unless it's attacking` path through `parse_oracle_text`, and current card-data/lint completed successfully.

I am holding approval/enqueue because this is a parser PR and the required `coverage-parse-diff` sticky comment has not posted yet for this head. I’ll re-check once the card-level parse-diff evidence is visible.

matthewevans (changes requested):
Thanks for the tight patch. I think this still needs one more pass before we can land it.

The parse-diff now shows Tromokratis moving from `static_structure` to `Continuous(... conditional=not (unrecognized), mods=grant Hexproof)`. That is a real partial parse, not just a harmless representation change: the new `unless` splitter feeds `"it's attacking or blocking"` through `rewrite_self_pronoun_subject`, but that helper only rewrites exact single-state tails like `attacking`, `blocking`, and `blocked`. It does not rewrite the compound `attacking or blocking`, so `parse_static_condition` never sees the canonical `~ is attacking or blocking` form that the condition parser already supports.

Please extend the existing self-pronoun normalization for this same combat-state axis rather than adding a new variant or special case, and add a regression test for `~ has hexproof unless it's attacking or blocking.` / Tromokratis that expects `Not(Or([SourceIsAttacking, SourceIsBlocking]))` with no `Unrecognized` condition. Once that parse-diff no longer contains the Tromokratis unrecognized condition, this should be straightforward to re-review.

minion1227:
Thanks for the catch on Tromokratis — you're right that it was a real partial parse, not just a representation change.

Fixed in `58ccaf14c` by extending the existing self-pronoun normalization on the same combat-state axis: added `"attacking or blocking"` to the exact-match set in `rewrite_self_pronoun_subject`, so `it's attacking or blocking` now normalizes to the canonical `~ is attacking or blocking` that the condition parser already lowers to `Or([SourceIsAttacking, SourceIsBlocking])`. No new variant or special case.

`~ has hexproof unless it's attacking or blocking.` now parses to `Not(Or([SourceIsAttacking, SourceIsBlocking]))` with no `Unrecognized` residue — the Tromokratis entry should be gone from the parse-diff.

Extended the existing regression test (`self_keyword_grant_unless_source_combat_state_gate`) with the Tromokratis compound case: asserts the `Not(Or([SourceIsAttacking, SourceIsBlocking]))` gate and that no `Unrecognized` condition remains. 950 oracle_static tests pass; fmt + clippy clean.

minion1227:
Thanks @matthewevans — addressed in `58ccaf14c`.

Extended the existing self-pronoun normalization on the same combat-state axis: `"attacking or blocking"` is now in the exact-match set of `rewrite_self_pronoun_subject` (no new variant / special case), so the compound tail reaches the canonical `~ is attacking or blocking` form that the condition parser already lowers to `Or([SourceIsAttacking, SourceIsBlocking])`. Tromokratis now parses to `Not(Or([SourceIsAttacking, SourceIsBlocking]))` with no `Unrecognized` residue.

Extended the regression test (`self_keyword_grant_unless_source_combat_state_gate`) with the Tromokratis compound case, asserting the `Or` gate and no `Unrecognized` in the serialized parse. The parse-diff no longer contains the Tromokratis unrecognized condition. Re-requesting review.

matthewevans (approved):
Reviewed current head 86b99a717a74c882bee52dbd4e6adfa1890b59fc. The previous Tromokratis parse-diff blocker is fixed: the current sticky shows Tadeas/Juniper and Tromokratis moving from static_structure to typed Continuous grants, with Tromokratis now gated by not (source is attacking or source is blocking). The implementation stays at the existing parse_continuous_gets_has/static-condition seam, extends the existing self-pronoun normalization axis rather than adding a new variant, and the tests assert both the Tadeas single-state and Tromokratis compound-state paths through parse_oracle_text. Approving for merge queue.

PR #4981, opened 2026-07-03
feat(parser): activated-ability cost mods with mana-ability exemption (Suppression Field, Zirda)

minion1227:
## Problem

Two activated-ability cost-modifier shapes fell through to a `static_structure` gap because the existing handlers only covered the scoped "Activated abilities **of** \<subject\>" form and had no way to express the "unless they're mana abilities" exemption:

- **Suppression Field**: `Activated abilities cost {2} more to activate unless they're mana abilities.` (global)
- **Zirda, the Dawnwaker**: `Abilities you activate that aren't mana abilities cost {2} less to activate.` (activator-scoped)

## Fix

The runtime already supported the `keyword == "activated"` blanket reducer **and** both directions (Reduce/Raise) — only the exemption and these two unfiltered subjects were missing.

- **types** (`statics.rs`): add `exemption: ActivationExemption` to `StaticMode::ReduceAbilityCost`, reusing the same **CR 605.1a** enum that `StaticMode::CantBeActivated` already carries (no bool, no new variant). `#[serde(default)]` keeps already-serialized card-data reading as the un-exempt form.
- **parser** (`oracle_static/dispatch.rs`): a new handler for the two subjects that carry no `of \<subject\>` filter — the **global** form (`affected = None` → all sources, all players) and the **activator** form (`Abilities you activate` → sources you control). Directional (`less`/`more`); the mana-ability exemption is parsed from the prefix `that aren't mana abilities` or the suffix `unless they're mana abilities`.
- **runtime** (`game/casting.rs`): `apply_static_activated_ability_cost_reduction` skips the adjustment for a mana ability when `exemption == ManaAbilities` (**CR 605.1a**), classified via `mana_abilities::is_mana_ability` before the mutable cost borrow.

### Parse result

```
Suppression Field → ReduceAbilityCost{ Raise, "activated", 2, exemption: ManaAbilities }  affected: None
Zirda             → ReduceAbilityCost{ Reduce, "activated", 2, exemption: ManaAbilities } affected: creatures/cards you control
```

## Scope / safety

The scoped `Activated abilities of \<subject\>` form, keyword-tagged reducers (ninjutsu, exhaust, power-up…), and all existing `ReduceAbilityCost` behavior are **unchanged** — `exemption` defaults to `None`.

## Tests

- 3 parser regressions: global + exemption, `you activate` + exemption, and the no-exemption variants (bare global + scoped-form regression).
- A **runtime** test proving the exemption actually works: with no mana available, a discounted non-mana ability (`{2}`→`{0}`) is activatable while an exempt **mana** ability (still `{2}`) is not.
- 14,683 engine tests pass; `cargo fmt` + `clippy -D warnings` clean.

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -14460,7 +14460,7 @@ fn apply_cost_reduction(
     // (active_keyword == None) — so skipping the whole function is equivalent to
     // skipping just the "activated" arm, and clearer.
     if !is_plot_special_action(ability_def) {
-        apply_static_activated_ability_cost_reduction(state, ability_def, source_id);
+        apply_static_activated_ability_cost_reduction(state, ability_def, player, source_id);
     }
 
     // CR 116.2k + CR 702.170: Plot is taken as a special action via a synthesized
@@ -14480,6 +14480,7 @@ fn apply_cost_reduction(
 fn apply_static_activated_ability_cost_reduction(
     state: &GameState,
     ability_def: &mut AbilityDefinition,
+    player: PlayerId,
     source_id: ObjectId,
 ) {
     // CR 601.2f: A `ReduceAbilityCost` static keyed on a keyword (e.g. "power-up")
@@ -14507,6 +14508,7 @@ fn apply_static_activated_ability_cost_reduction(
             minimum_mana,
             dynamic_count,
             exemption,
+            activator,
         } = &def.mode
         else {
             continue;
@@ -14518,6 +14520,25 @@ fn apply_static_activated_ability_cost_reduction(
         // adjustment (Suppression Field's tax, Zirda's discount).
         if *exemption == ActivationExemption::ManaAbilities && ability_is_mana {
             continue;
+        }
+        // CR 602.2: an activator-scoped static ("abilities you activate" — Zirda,
+        // the Dawnwaker; Fluctuator) keys off WHO is activating the ability,
+        // evaluated relative to the static's controller — NOT who controls the
+        // ability's source. Reuse the activator-permission predicate with the
+        // static's controller as the reference point so "you" resolves to the
+        // static controller. An ability on a permanent this player doesn't control
+        // (activatable via `activator_filter`) is still discounted when they
+        // activate it, and an ability on a permanent they DO control but activated
+        // by someone else is not. `None` leaves the source/global scope untouched.
+        if let Some(activator) = activator {
+            if !player_may_begin_activating(
+                state,
+                player,
+                static_source.controller,
+                Some(activator),
+            ) {
+                continue;
+            }
         }
         if def.affected.as_ref().is_some_and(|filter| {
             !super::filter::matches_target_filter(
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -14489,6 +14489,11 @@ fn apply_static_activated_ability_cost_reduction(
     let active_keyword = ability_def
         .ability_tag
         .map(crate::types::ability::AbilityTag::keyword_str);
+    // CR 605.1a: Classify the activating ability BEFORE the mutable cost borrow so
+    // an `ActivationExemption::ManaAbilities` static ("unless they're mana
+    // abilities" / "that aren't mana abilities" — Suppression Field, Zirda) can
+    // skip a mana ability's cost.
+    let ability_is_mana = super::mana_abilities::is_mana_ability(ability_def);
 
     let Some(cost) = ability_def.cost.as_mut() else {
         return;
@@ -14501,11 +14506,17 @@ fn apply_static_activated_ability_cost_reduction(
             amount,
             minimum_mana,
             dynamic_count,
+            exemption,
         } = &def.mode
         else {
             continue;
         };
         if (keyword != "activated" && Some(keyword.as_str()) != active_keyword) || *amount == 0 {
+            continue;
+        }
+        // CR 605.1a: a mana ability bypasses a "unless they're mana abilities"
+        // adjustment (Suppression Field's tax, Zirda's discount).
+        if *exemption == ActivationExemption::ManaAbilities && ability_is_mana {
             continue;
         }
         if def.affected.as_ref().is_some_and(|filter| {
--- a/crates/engine/src/game/casting_tests.rs
+++ b/crates/engine/src/game/casting_tests.rs
@@ -8437,6 +8437,7 @@ fn activated_ability_cost_reduction_applies_to_matching_permanent_type() {
                 amount: 1,
                 minimum_mana: None,
                 dynamic_count: None,
+                exemption: crate::types::statics::ActivationExemption::None,
             })
             .affected(TargetFilter::Typed(TypedFilter {
                 type_filters: vec![TypeFilter::Subtype("Food".to_string())],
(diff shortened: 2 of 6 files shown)


matthewevans:
### Parse changes introduced by this PR · 15 card(s), 9 signature(s)  (baseline: main `9ddc6846c615`)

#### 4 card(s) · static/ReduceAbilityCost(-equip,1) · field `affects`: `you control card` → `∅`
Examples: Arms Scavenger, Bladehold War-Whip, Bureau Headmaster (+1 more)

#### 3 card(s) · static/ReduceAbilityCost(-equip,2) · field `affects`: `you control card` → `∅`
Examples: Cloud, Planet's Champion, Fighter Class, Helitrooper

#### 2 card(s) · static/ReduceAbilityCost(-equip,3) · field `affects`: `you control card` → `∅`
Examples: Fervent Champion, Strong Back

#### 2 card(s) · static/ReduceAbilityCost(-ninjutsu,1) · field `affects`: `you control card` → `∅`
Examples: A-Silver-Fur Master, Silver-Fur Master

#### 2 card(s) · ability/static_structure · removed: `static_structure`
Examples: Suppression Field, Zirda, the Dawnwaker

#### 1 card(s) · static/ReduceAbilityCost(+activated,2) · added: `ReduceAbilityCost(+activated,2)`
Examples: Suppression Field

#### 1 card(s) · static/ReduceAbilityCost(-activated,2,1) · added: `ReduceAbilityCost(-activated,2,1)`
Examples: Zirda, the Dawnwaker

#### 1 card(s) · static/ReduceAbilityCost(-boast,1) · field `affects`: `you control card` → `∅`
Examples: Dragonkin Berserker

#### 1 card(s) · static/ReduceAbilityCost(-cycling,2) · field `affects`: `you control card` → `∅`
Examples: Fluctuator

minion1227:
Gentle nudge — the coverage evidence you were holding for is now posted. The `coverage-parse-diff` sticky is populated and shows exactly the intended surface:

- **Suppression Field, Zirda** — `static_structure` gap removed
- Suppression Field → `ReduceAbilityCost(+activated, 2)` (Raise, mana-ability exemption)
- Zirda → `ReduceAbilityCost(-activated, 2, min 1)` scoped to `you control` (Reduce, mana-ability exemption; the `min 1` is Zirda's own "can't reduce … to less than one mana" clause)

No other cards affected. All checks are green (including Card data / coverage-gate) and merge state is CLEAN. Whenever you have a moment for the semantic diff, this should be ready for re-review / enqueue. Thanks!

matthewevans (changes requested):
Thanks for the follow-up. I think the mana-ability exemption itself is on the right axis, but the Zirda branch is still modeling the wrong scope for `Abilities you activate`.

`dispatch.rs` parses that form into `affected = TypedFilter::card().controller(ControllerRef::You)`, and `apply_static_activated_ability_cost_reduction` only evaluates that source filter. That makes the effect source-controller scoped, not activator scoped. The engine already has a separate activator axis (`player_may_begin_activating` consults `AbilityDefinition::activator_filter`), so this can produce the wrong result for abilities that another player may activate: Zirda should key off who is activating the ability, while the current implementation keys off who controls the ability source.

Please keep the mana-ability exemption parameterized, but separate the activator scope from the affected source filter. In practice this likely means passing the activating `player` through the activated-ability reducer and gating the Zirda-style form on the static source controller/activator, while keeping source filters for actual “of <subject>” text. Please add a discriminating runtime test using an ability with `activator_filter` set so the static only discounts abilities activated by the Zirda controller, not merely abilities on sources that player controls.

minion1227:
Thanks @matthewevans — addressed in `75bc23fe4`. You're right that the Zirda branch was modeling the wrong scope.

**Separated the activator scope from the source filter.** "Abilities you activate" is activator-scoped (CR 602.2 — keyed on *who activates*, relative to the static's controller), not source-controller-scoped. Concretely:

- Added `activator: Option<PlayerFilter>` to `StaticMode::ReduceAbilityCost` (`#[serde(default, skip_serializing_if)]`).
- Threaded the activating `player` through `apply_static_activated_ability_cost_reduction` and gate it via `player_may_begin_activating(state, player, static_source.controller, ..)` — so `Some(PlayerFilter::Controller)` resolves "you" to the static's controller. `None` leaves the source/global scope untouched.
- Parser: both "abilities you activate" forms — blanket Zirda **and** the pre-existing keyword-scoped Fluctuator/Ninjutsu/Equip (same latent mis-scope, fixed as a class) — now emit `activator = Some(Controller)` and **no** `affected` filter. The source-scoped "abilities **of** \<subject\>" forms and the global Suppression Field form are unchanged (source filter / `None` respectively).

**Discriminating runtime test** (`activated_ability_cost_reduction_you_activate_keys_off_activator_not_source_controller`) drives `apply_cost_reduction` in **both** directions with an ability whose `activator_filter = All`:
- opponent activating an ability on the static-controller's **own** permanent → **not** discounted;
- static-controller activating an ability on the **opponent's** permanent → **is** discounted.

A source-controller-scoped implementation (the reverted `affected(controller(You))`) fails each direction oppositely, so both assertions pin the fix. Also flipped the parser test that had asserted the old `affected` filter.

Note the parse-diff will show Zirda's signature drop the `(affects=you control card)` annotation — that source filter *was* the bug; the activator scope now rides on the static (serialized `activator: Controller`) and is proven by the runtime test above.

14690 engine tests pass, clippy + fmt clean.

minion1227:
@matthewevans this is ready for re-review whenever you have a moment. 🙏

The activator-scope fix is in (`75bc23fe4`): "abilities you activate" now keys off the activator relative to the static's controller (CR 602.2), not a source-controller filter — with the discriminating both-directions runtime test you asked for. All checks are green on the current head (10 passed / 1 skipped, 0 pending) and merge state is `MERGEABLE`. Happy to rebase onto current main if you'd prefer it caught up before enqueue.

matthewevans (changes requested):
Thanks for the activator-scope repush. The new runtime test around activator vs. source controller is the right seam, and the `activator: Option<PlayerFilter>` axis is the parameterized shape we want rather than another cost-modifier sibling.

I still can't clear this head because the required parse-diff evidence is not bounded to the PR's claimed surface. The current sticky was updated for this head (`75bc23fe4`, updated 2026-07-03T13:57:59Z) and reports 117 cards / 90 signatures, including Tap add/remove churn, MustBeBlockedByAll moving ability to static, unless_payment changes, modal enters-face-up, Oubliette, Iona, Primal Clay, and other signatures that are not explained by Suppression Field/Zirda or the `ReduceAbilityCost` activator/exemption axis.

The PR body also says existing scoped behavior is unchanged, but the parse-diff shows several `{equip,ninjutsu,boast,cycling} abilities you activate` signatures changing `affects: you control card -> ∅`. That may be the correct activator-scope correction, but it needs to be explicitly owned and bounded in the PR evidence.

Please either regenerate/fix the parse-diff if it is stale/noisy, split the PR so the parse-diff is limited to the intended cost-modifier changes, or explicitly justify and test the additional parser behavior changes here. Until the card-level parser evidence matches the claimed scope, this fails the scope-contamination/evidence gate even though the activator-axis implementation itself is materially better.

minion1227:
Good catch on the evidence gate — the contamination was **branch staleness**, now fixed. The branch was based on a main tip that was 21 commits behind, so the parse-diff (my head vs current-main baseline) surfaced every parser change that had merged in between (Tap churn, MustBeBlockedByAll, unless_payment, modal enters-face-up, Oubliette, Iona, Primal Clay, …) as if it were mine.

**Rebased onto current main (`437aab9bd`)** — clean, zero conflicts (force-pushed `5ba370f4`). The branch is now a linear 2-commit stack on the current tip, and the net diff is bounded to exactly the 6 files this PR owns (statics/casting/dispatch + tests). The refreshed parse-diff should now report only the intended surface.

**Explicitly owning the intended parser surface** (the only card-level changes that should remain):
- **Suppression Field** → `ReduceAbilityCost(+activated, 2)` (global, mana-ability exemption)
- **Zirda, the Dawnwaker** → `ReduceAbilityCost(-activated, 2, min 1)`, now `activator: Controller` + no `affected`
- **`{equip, ninjutsu, boast, cycling} abilities you activate`** cards → `affects: you control card → ∅`. This is the **deliberate activator-scope correction**, not incidental churn: those forms are "abilities **you** activate" (activator-scoped per CR 602.2), and the old `affected = card().controller(You)` source filter was the same mis-scope you flagged on Zirda. They now ride the `activator: Some(Controller)` axis. Covered by the flipped parser tests and the both-directions discriminating runtime test.

If the refreshed sticky still shows anything outside that set, point me at it and I'll chase it down. Re-verified after rebase: `cargo check` + `clippy` clean; full suite was green pre-rebase and CI is re-running the complete gate on the new head now.

matthewevans (approved):
Reviewed current head. This is the right parameterization: activated-ability cost modification now carries mana-ability exemption and activator scope instead of proliferating source-filter special cases. The parse-diff sticky is current and the broader affected-card set is the expected class-level correction.

PR #4984, opened 2026-07-03
fix(parser): resolve "an opponent's graveyard" target to a graveyard card

ultrahighsuper:
## Summary

`parse_target("an opponent's graveyard")` returned a bare Opponent-**player** filter and left `"'s graveyard"` as an unconsumed remainder, instead of a card-in-graveyard filter. The un-bounded `tag("an opponent")` arm of the opponent-player references matched the `"an opponent"` prefix of `"an opponent's graveyard"` *before* the more-specific graveyard-phrase branch could run — leaving that branch (which already lists `"an opponent's graveyard"`) as dead code.

This reorders the graveyard-phrase branch ahead of the opponent-player references, so the possessive form resolves to `Owned{Opponent} + InZone{Graveyard}` — matching the already-correct **no-"an"** sibling (`"opponent's graveyard"`), which was never shadowed and has a passing test. The two possessive forms now agree.

Why reorder rather than word-bound the `"an opponent"` tag: the graveyard-phrase branch is the specific downstream handler for this possessive form, and reordering restores its evidently-intended precedence with zero blast radius. Bare opponent-player references (`"an opponent chosen at random"`, `"each opponent"`, `"opponents"`) contain no `"graveyard"` token, so they still fall through unchanged — pinned by a new guard test.

> Minimal, surgical reordering of two adjacent branches within an existing `parse_target` dispatch cluster (no new pattern, variant, or matcher). A final `/review-impl` pass was run; the broader `"an opponent's <noun>"` shadowing class (the un-bounded tag as root cause) is intentionally left out of scope because graveyard is the only possessive form with a dedicated downstream handler — changing routing for the others has no correct destination and would risk regressions.

## CR references

- `CR 108.3` — owner of a card (the `Owned{Opponent}` property: a card in an opponent's graveyard is owned by that opponent).
- `CR 404.1` — a player's graveyard (the `InZone{Graveyard}` property).

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib parse_target_` — 41 passed, incl. two new tests:
  - `parse_target_an_opponents_graveyard_is_graveyard_filter` — the fix (revert-probe: **fails under the old ordering**, confirming nothing earlier catches the bare phrase and the assertion is non-vacuous).
  - `parse_target_bare_an_opponent_still_player` — guard proving the bare opponent-player reference is unaffected by the reorder.

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -1102,6 +1102,31 @@ pub fn parse_target_with_syntax<'a>(
         );
     }
 
+    // CR 108.3 + CR 404.1: an opponent's graveyard as a target resolves to a
+    // card in that graveyard. This more-specific possessive phrase MUST be tried
+    // before the bare opponent-player references below: the un-bounded
+    // `tag("an opponent")` arm would otherwise match the "an opponent" prefix of
+    // "an opponent's graveyard" and return a bare Opponent-player filter, leaving
+    // "'s graveyard" as an unconsumed remainder. The no-"an" sibling
+    // ("opponent's graveyard") is unaffected either way (no opponent-player tag
+    // matches "opponent's"), so both possessive forms now agree.
+    for phrase in ["opponent's graveyard", "an opponent's graveyard"] {
+        if let Ok((rest, _)) = tag::<_, _, OracleError<'_>>(phrase).parse(lower.as_str()) {
+            return (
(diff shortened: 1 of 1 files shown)


matthewevans:

matthewevans (approved):
Current-head review is clean. This is a narrow parser ordering fix for the bare opponent-graveyard target phrase, with regression coverage and no card-parse blast radius.

PR #4986, opened 2026-07-03
fix(parser): require a word boundary after English number words

ultrahighsuper:
## Summary

`parse_english_number` matched multi-character cardinals with bare `tag()`s and no trailing word-boundary check, so a number word was matched *inside* a longer word:

| input | before | after |
|-------|--------|-------|
| `sixth` | `Ok(("th", 6))` | `Err` |
| `tenth` | `Ok(("th", 10))` | `Err` |
| `tenfold` | `Ok(("fold", 10))` | `Err` |
| `nineteenth` | `Ok(("th", 19))` | `Err` |

The `oracle_util::parse_number` wrapper only applies its boundary guard to matches of ≤2 characters (i.e. `"a"`/`"an"`), so every real number word (≥3 chars) slipped through unguarded.

This adds the same word-boundary guard `parse_article_number` already applies to `"a"`/`"an"`: reject the match when the next character is ASCII-alphanumeric (the token continues, so the cardinal was a spurious substring). Distinct number words that merely share a prefix — `"sixteen"`, `"sixty"` — are matched by their own longer/earlier `alt` arms and are unaffected.

> Small, self-contained parser primitive fix: a word-boundary guard on one `nom` combinator, mirroring the existing `parse_article_number` guard one function above it. No new pattern, variant, or game logic — it's lexical parsing plumbing, so no CR annotation applies. A final `/review-impl` pass was run.

## CR references

None — this is lexical number-word tokenization, not a game rule.

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib parser::` — **7203 passed, 0 failed** (no regression across the whole parser suite).
- New test `test_parse_english_number_requires_word_boundary` is a revert-probe: it **fails without the guard** (`"sixth"` parses as 6) and passes with it, while genuine cardinals with a trailing boundary (`"six cards"`, `"nineteen creatures"`, `"three, then draw"`) and prefix-sharing words (`"sixteen"`, `"sixty"`) still parse correctly.

### DIFF
--- a/crates/engine/src/parser/oracle_nom/primitives.rs
+++ b/crates/engine/src/parser/oracle_nom/primitives.rs
@@ -72,7 +72,7 @@ fn parse_digit_number(input: &str) -> OracleResult<'_, u32> {
 fn parse_english_number(input: &str) -> OracleResult<'_, u32> {
     // Longest-match-first ordering within shared prefixes (e.g. "fourteen" before "four").
     // Split into multiple alt groups to stay within nom's 21-element tuple limit.
-    alt((
+    let (rest, matched) = alt((
         value(100u32, tag("one hundred")),
         parse_hyphenated_english_number,
         parse_english_tens,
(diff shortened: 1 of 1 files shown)


matthewevans:

ultrahighsuper:
Thanks — brought the branch current with `main` (head `700f1fe4f`) so the Card data / coverage job regenerates the parse-diff sticky against this head. This change only adds a word-boundary guard to `parse_english_number` (rejecting a cardinal matched inside a longer word, e.g. `sixth`/`tenfold`), so the parse-diff is a no-change result; the full parser unit suite (`cargo test -p engine --lib parser::`, 7203 tests) is green with the guard, and the new `test_parse_english_number_requires_word_boundary` is a revert-probe (fails without the guard).

matthewevans:
Holding review on current head `700f1fe4fbb2e5c380c30ce5e08d61583fc43069`: Card data / coverage completed for this head, but I still do not see a current `<!-- coverage-parse-diff -->` sticky.

For parser primitive changes, I need the parse-diff evidence before approving, including either scoped real changes or an explicit no-change result tied to this head.

matthewevans (approved):
Reviewed current head `8c40b26237`: right seam in the shared English-number parser primitive, no card-parse changes in the current parse-diff sticky, and the new boundary tests discriminate the reverted behavior (`sixth`/`tenfold` etc. no longer parse as cardinal numbers). Approving for merge queue; pending CI can settle there.

PR #4990, opened 2026-07-03
feat(engine): S07 condition-if standard cluster — 28 cards + token-chain and as-long-as class fixes

lgray:

## What

Completes the **S07 condition-if standard cluster: 28 cards**, all `supported:true gap_count:0`, via 14 commits of building-block work (no card-specific hacks). Highlights by mechanism family:

- **"this way" tracked-set Condition_If** (batch 1, 9 cards) · **cost-object / additional-cost / gift** (batch 2, 5) · **target/identity** (batch 3, 5)
- **Celestial Reunion** — behold-chosen-type additional cost + conditional search destination
- **Target-model infra** — `ParentTargetSlot` indexed resolution (`resolve_parent_slot_from_root`) + delayed/filter sites
- **Malamet Battle Glyph** — entered-this-turn condition detector + a reciprocal-fight class fix (un-broke the previously-uncastable "those creatures fight each other" class: Longstalk Brawl, Duel for Dominance)
- **A Killer Among Us** — secret creature-type choice subsystem + target-hoist gate
- **Trance Kuja / The Rollercrusher Ride** — damage-doubler Condition_If false-positive fixes
- **Cloud, Midgar Mercenary** — DoubleTriggers self+equipment gating (CR 603.2d)
- **Sonic Shrieker / Slumbering Trudge / Avatar Aang** (new `QuantityRef::BendTypesThisTurn`)

## Class fixes beyond the 28

- `fix(parser)`: **"create A, a B, and a C token" 3+-item chains silently dropped the middle token** — N-way conjunctive split with peek-guarded separators; fixes 9 more printed cards (Bestial Menace, Fae Offering, Triplicate Titan, Trostani's Summoner, …) and corrects The Companion of the Wilds' can't-block misattribution.
- Cloud's dispatch "as long as" re-attach rescued 4 sibling statics for free: Ethrimik, Glimpse the Cosmos, Pact Weapon, The Bird Champion.
- Malamet's detector flipped Samut, Vizier of Naktamun for free.
- `fix(engine)`: delayed-trigger `ParentTargetSlot` now snapshots the **full root chain** (multi-clause delayed triggers; Flickerwisp-class single-clause cards unaffected by construction) + context-ref `UnattachAll` surfaced in `effect_parent_ref_slots`.

## Verification

- Every card: **measured** per-card coverage diff (REGRESSED(engine)=0 at each commit; GAINED = exactly the intended cards).
- Discriminating tests with executed revert-fail evidence at each increment; full `cast().resolve()` runtime tests for cast-path work (not condition-only probes).
- All CR annotations grep-verified against the Comprehensive Rules; parser work nom-combinator-first (parser gate clean vs merge base).
- Rebased onto current `main`; full CI-equivalent (fmt, workspace clippy `-D warnings` with proptest, engine test suite) green locally.

https://claude.ai/code/session_01CwQE5oyMqZ9T4BPMsih3Kj


matthewevans (commented):
Picked up current head `7ce29301` in the sweep. This touches broad engine/parser surface, so I’m holding review until the `coverage-parse-diff` sticky is present for this exact head.

That card-level parse diff is required evidence for parser/engine PRs before approval or enqueue. No approval/enqueue from the sweep yet.

matthewevans (changes requested):
**[MED]** Canceling after a behold type choice leaves a stale creature type on the spell object. Evidence: `crates/engine/src/game/casting_costs.rs:1945`, `crates/engine/src/game/casting_costs.rs:4255`, and `crates/engine/src/game/casting.rs:14203`. Why it matters: the choice is written onto the card object, later cast attempts skip `CostTypeChoice` once any creature-type attribute exists, and `handle_cancel_cast` does not rewind that temporary choice, so a canceled cast can force the next cast attempt to reuse the old type. Suggested fix: keep the pre-cost choice in `PendingCast`/cost state, or explicitly remove the temporary `CreatureType` chosen attribute when the cast is canceled.

**[MED]** `CostTypeChoice.options` leaks hidden hand-derived feasibility. Evidence: `crates/engine/src/game/casting_costs.rs:4276`, `crates/engine/src/game/filter.rs:2225`, and `crates/engine/src/game/casting_costs.rs:449`. Why it matters: the public waiting state serializes creature-type options computed from hand cards, so opponents can infer private hand type/count information before only the selected behold cards are revealed. Suggested fix: keep serialized options public/generic and validate feasibility on submit, or add per-viewer redaction for `CostTypeChoice.options`.

**[MED]** `manabrew-compat` cannot prompt for the new `CostTypeChoice` waiting state. Evidence: `crates/engine/src/game/casting_costs.rs:4276` emits `WaitingFor::CostTypeChoice`, but `crates/manabrew-compat/src/lib.rs:730` only maps `NamedChoice` to `chooseType`, and the match falls through to `UnsupportedPrompt` at `crates/manabrew-compat/src/lib.rs:804`. Why it matters: Manabrew agents will fail to cast cards that enter this new behold type-choice step. Suggested fix: add a `WaitingFor::CostTypeChoice` arm mirroring `NamedChoice`, and include it in `source_card_id` using `pending_cast.object_id`.

**[LOW]** The stack UI drops the pending-cast indicator during `CostTypeChoice`. Evidence: `client/src/adapter/types.ts:1227` defines `CostTypeChoice` with `pending_cast`, but `getPendingCastObjectId` only reads adjacent pending-cast states at `client/src/components/stack/StackDisplay.tsx:35`; that value drives `isPending` at `client/src/components/stack/StackDisplay.tsx:330` and the “Casting” badge at `client/src/components/stack/StackEntry.tsx:221`. Why it matters: during this new casting-cost prompt, the frontend no longer marks the in-progress spell on the stack, unlike adjacent cost-choice states. Suggested fix: add `case "CostTypeChoice": return waitingFor.data.pending_cast.object_id;` to `getPendingCastObjectId`.

lgray:

Thanks for the thorough review — all four findings are addressed in `d8dbffbd8` (branch rebased onto current main `bb0339203`; force-push). For the record, the `CostTypeChoice`/behold subsystem was introduced by the Celestial Reunion commit rather than A Killer Among Us, but the analysis applies verbatim.

**1. Cancel leaves stale creature type** — `handle_cancel_cast` now clears the temporary `ChosenAttribute::CreatureType` from the spell object, so backing out fully rewinds the choice (CR 601.2 + CR 733.1 — "the entire action is reversed"). The `already_chosen` guard therefore re-prompts on the next cast. `cancel_after_type_choice_rewinds_and_recast_reprompts` asserts both that the attribute is gone after cancel and that the re-cast re-presents `CostTypeChoice`. (We verified the clear can't collide with persisted post-resolution choices: the only production cast-time writer of `CreatureType` on a cancelable spell object is the behold pre-cost path; `bind_named_choice` writes onto battlefield permanents at resolution.)

**2. Hidden-info leak** — `filter_state_for_viewer` now empties `CostTypeChoice.options` for any viewer who can't see the caster's private zones (the list is derived from beholdable hand cards), mirroring the existing `PayCost` hand-redaction (CR 400.2). The acting player keeps the full list; there's no residual length leak (the Oracle-public restricted-options axis is separate from the feasibility set). Covered by `cost_type_choice_options_redacted_for_opponent_viewer`, which asserts both caster-full and opponent-empty.

**3. manabrew-compat gap** — Added a `CostTypeChoice` arm mirroring `NamedChoice`: `chooseType` with `validTypes`, `source_card_id` from `pending_cast.object_id`, plus a `waiting_for_type` label — no more `UnsupportedPrompt` fall-through. manabrew builds from `filter_state_for_viewer(raw_state, viewer)`, so it inherits the same caster-only visibility. Covered by `cost_type_choice_prompt_maps_to_choose_type_with_source`.

**4. Frontend badge drop** — `getPendingCastObjectId` now includes `CostTypeChoice`, so the stack keeps its "Casting" badge during the prompt.

On the sweep-review hold: the coverage-parse-diff sticky regenerates for this exact head on this push; the parser changes it lists are the intended target-model/behold work from the tranche. Local measurement (main `bb0339203` → head): **REGRESSED(engine)=0, GAINED=38** — no card lost engine support.

One additional item surfaced during our internal review of the fix, out of scope here: a pre-existing latent bug where `already_chosen` reuses a stale creature type when a behold-pre-cost spell is re-cast from graveyard/exile (`chosen_attributes` isn't cleared on zone exit for non-permanents; CR 400.7 makes the re-cast a new object that should re-prompt). Independent of this fix — this fix shrinks the affected surface — and we'll file it separately.

matthewevans (approved):
Current-head re-review of `d8dbffbd8c`: the prior requested-change items are addressed, including cancel rewind for behold type choice, viewer redaction, manabrew prompt/source wiring, and stack badge display. Parse-diff evidence is broad but explainable for the claimed class-level parser/engine work; final fix commit does not alter parser scope. Approving; CI/merge queue can gate readiness.

lgray:

Note on the red **Decision-cost perf gate** on `ec0b9ea4`: this is the card-data-driven trajectory shift the gate itself flags in its log (`card-data hash changed (543c313f…→e8baaebf…) — likely a card-data-driven trajectory shift, not a cost-per-node regression`). The tranche's 38 newly-supported cards change the card pool, which reroutes the three AI mirror scenarios; the 6 over-threshold counters are node-count scans in subsystems (crew eligibility, legend rule, SBA snapshots, layers, state clone) that this PR's engine diff doesn't touch, and the branch carries no change to `perf-baseline.json`.

Per the repo convention (baseline refreshes ship with the paired-seed report attached), we'll submit the baseline refresh as an immediate follow-up PR computed against post-merge main rather than adding a commit to this queued PR — so this branch stays exactly as you reviewed it. Merge past the perf gate at your discretion, or if you'd rather the refresh land first, say the word and we'll reorder.

PR #4992, opened 2026-07-03
Configure multiplayer default server for self-hosted builds

davidomil:
## Summary

Makes the multiplayer lobby server default configurable at build time so self-hosted client bundles can default to their own broker while still keeping the official lobby available in the server picker.

the pipeline):

> Frontend-only client configuration and persistence migration change; this does not touch engine/parser logic or `crates/engine/` rules behavior.

> [!NOTE]

## CR references

None.

## Verification

- `git diff --check origin/main...HEAD` — passed
- `NODE_OPTIONS=--no-experimental-webstorage npx --yes pnpm@10.23.0 exec vitest run src/services/__tests__/serverDetection.test.ts src/stores/__tests__/multiplayerStore.test.ts` — passed, 32 tests
- `npx --yes pnpm@10.23.0 type-check` — passed
- `npx --yes pnpm@10.23.0 exec eslint . --quiet` — passed
- `DEFAULT_MULTIPLAYER_SERVER_URL=wss://selfhost.example/ws node --input-type=module ...` Vite config-load assertion — passed; `__DEFAULT_MULTIPLAYER_SERVER_URL__` resolved to the supplied self-host URL

### DIFF
--- a/client/src/components/lobby/ServerPicker.tsx
+++ b/client/src/components/lobby/ServerPicker.tsx
@@ -124,7 +124,7 @@ export function ServerPicker({ onClose, onApply }: ServerPickerProps) {
                       className="h-3.5 w-auto rounded-[2px] shadow-sm ring-1 ring-black/20"
                     />
                   )}
-                  {preset.label}
+                  {t(preset.labelKey)}
                 </span>
                 <span className="min-w-0 truncate pl-2 font-mono text-[10px] text-slate-500">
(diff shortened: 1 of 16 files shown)


matthewevans (approved):
Maintainer sign-off: reviewed scoped frontend configuration change, added i18n cleanup and stronger default-server coverage, targeted verification passed locally.

PR #4993, opened 2026-07-03
fix(engine): clamp negative discard counts to zero (CR 107.1b)

ultrahighsuper:
## Summary

The discard resolver cast the resolved count `as u32` with no lower bound (`discard.rs`, the `Effect::Discard` arm). A subtractive count — "discard cards equal to [A] minus [B]" (`Sum[A, Multiply{-1, B}]`) — resolves negative when `B > A`, and `-2 as u32` wraps to `4_294_967_294`. In the **"up to"** path that wrapped count is passed **uncapped** into `WaitingFor::DiscardChoice`, so the player is offered a discard of ~4 billion and may legally dump their **entire hand**, when CR 107.1b requires discarding **zero**.

Every sibling resolver already clamps this exact hazard — `draw.rs` (whose comment names the Mr. Foxglove subtractive shape), `sacrifice.rs`, `adapt.rs`, `token.rs`, etc. `discard.rs` was the lone unclamped one. The fix is the same one-liner:

```rust
resolve_quantity_with_targets(state, inner, ability).max(0) as u32,
```

> One-line resolver fix that restores an invariant every sibling resolver already holds (`.max(0)` before the `as u32` cast), plus a regression test. No new pattern, variant, or AST. A final `/review-impl` self-check was run.

## CR references

- `CR 107.1b` — "If a calculation that would determine the result of an effect yields a negative number, zero is used instead."

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib game::effects::discard::tests` — 29 passed.
- New `discard_negative_count_clamps_to_zero` resolves `Effect::Discard { count: UpTo(HandSize{You} − HandSize{Opponent}) }` with the opponent holding more cards, and asserts the presented `DiscardChoice` count is `0`. Revert-probe: **without** the `.max(0)` the presented count is `u32::MAX - 1` (`4294967294`) and the test fails — proving the wrap is real and reachable, and the assertion is non-vacuous (both fixed and buggy paths produce a `DiscardChoice`, differing only in the count).

### DIFF
--- a/crates/engine/src/game/effects/discard.rs
+++ b/crates/engine/src/game/effects/discard.rs
@@ -141,8 +141,13 @@ pub fn resolve(
         } => {
             let (inner, up_to) = count.peel_up_to();
             (
-                // CR 107.1b: Use ability context so X resolves against the caster's chosen value.
-                resolve_quantity_with_targets(state, inner, ability) as u32,
+                // CR 107.1b: a calculation that would yield a negative number
+                // uses zero instead. Clamp before the `as u32` cast — a
+                // subtractive count ("discard cards equal to A minus B" when
+                // B > A) would otherwise wrap to ~4 billion and let the player
+                // discard their whole hand. Ability context also resolves X
+                // against the caster's chosen value. Mirrors `draw.rs`.
+                resolve_quantity_with_targets(state, inner, ability).max(0) as u32,
                 up_to,
                 unless_filter.clone(),
                 target.clone(),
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Maintainer review: localized engine fix at the discard-count seam. The CR 107.1b clamp happens before the signed-to-unsigned cast, mirrors the existing draw-count handling, and the regression test would fail as a huge discard count if the clamp were reverted.

PR #4994, opened 2026-07-03
fix(parser): bind anaphoric "equal to the difference" counter count to the trigger's power comparison (Drizzt Do'Urden)

### LINKED ISSUE #4967: Drizzt Do'Urden — doesn't get counters when larger creatures die
Reported in Discord: https://discord.com/channels/1485498006781427802/1522109777709105234/1522109777709105234

**Thread:** Drizzt Do'Urden
**Cards:** drizzt do'urden
**Parser status:** no_card (heuristic missed the card name)

## Summary
Drizzt is not getting +1/+1 counters when creatures with power greater than Drizzt's power die.

### drizzt do'urden
Double strike
When Drizzt enters, create Guenhwyvar, a legendary 4/1 green Cat creature token with trample.
Whenever a creature dies, if it had power greater than Drizzt's power, put a number of +1/+1 counters on Drizzt equal to the difference.

---
<sub>report_id: `discord:1522109777709105234:1522109777709105234:0` · discord: `1522109777709105234/1522109777709105234`</sub>

dhgoal:
## Problem

Drizzt Do'Urden's dies trigger — "Whenever a creature dies, if it had power greater than Drizzt's power, put a number of +1/+1 counters on Drizzt **equal to the difference**" — was a no-op: the intervening-if parsed to nothing and the bare anaphoric "the difference" count fell to `Effect::Unimplemented`. This was a **deliberately-documented gap** (`oracle_effect/counter.rs` returns `None` for the bare "the difference" rather than emit a wrong 0-count placeholder that would swallow the trailing clause).

## Fix (parser-only)

- **`parse_event_object_pt_vs_source_comparison`** (`oracle_nom/condition.rs`): a nom combinator parsing "it had `<stat>` {greater/less than\[ or equal to\]} ~'s `<stat>`" — the *triggering* object's last-known stat (`ObjectScope::EventSource`, CR 603.10a graveyard look-back) vs the ability source's stat — into a `StaticCondition::QuantityComparison`, bridged onto the trigger's `condition` by the existing `static_condition_to_trigger_condition`.
- **`parse_counter_equal_to`** (`counter.rs`): recognizes the bare "the difference" (via `all_consuming(tag("the difference"))`) and emits a deferred count placeholder, so the effect parses to a well-formed `PutCounter` (counter type + placement target intact) instead of `Unimplemented`.
- **`lower_trigger_ir`** (`oracle_trigger.rs`): binds that placeholder to `QuantityExpr::Difference{ comparison.lhs, comparison.rhs }` through the **existing draw/lose difference-binding block** (`Effect::count_expr_mut`, effect-agnostic). A placeholder that no comparison could bind is downgraded to an explicit `Unimplemented` residual — **coverage-honesty parity with the draw/lose siblings**, never a silently-zero `PutCounter` that would read as supported.

No runtime changes — the runtime already resolves `Difference{Power{EventSource}, Power{Source}}` from the dying creature's last-known power (`quantity.rs` LKI; LKI is snapshotted on the battlefield→graveyard zone change). Built for the **class** "if it had `<stat>` greater than ~'s `<stat>` … equal to the difference" — no name special-case, no new enum variants.

## Anchored on
- `crates/engine/src/parser/oracle_trigger.rs:1276` — the existing `difference_count` / `is_difference_draw` / `is_difference_lose` block that binds "equal to the difference" for the Draw and LoseLife classes (Kozilek, Damia, …). This fix extends that same block, effect-agnostically via `Effect::count_expr_mut()`, to the PutCounter class, and gives the counter path the same unbound→`Unimplemented` loud-gap behavior.
- `crates/engine/src/parser/oracle_nom/condition.rs:1785` — `parse_source_power_toughness_condition` (source stat vs a fixed threshold): the new `parse_event_object_pt_vs_source_comparison` (`:1830`) is its two-object sibling (EventSource stat vs Source stat), placed ahead of the source-state group so "it had `<stat>`" isn't mistaken for the source-subject form.

## Gate A
```
$ ./scripts/check-parser-combinators.sh   (verified via git diff vs origin/main)
CLEAN — no forbidden string-method dispatch in the added parser lines
```

## Verification
- `cargo fmt --all` — clean
- `cargo clippy -p engine --tests -- -D warnings` — clean (exit 0). Full-workspace `clippy-strict` needs OpenSSL/`pkg-config` (unavailable in my sandbox); CI **Rust lint** is the green run.
- Combinator gate — verified by `git diff origin/main -- 'crates/engine/src/parser/**' | grep '^+' | grep -E '<forbidden methods>'` → no matches.
- `cargo test -p engine --lib -- parser:: game::triggers:: game::quantity:: game::effects::counters::` — **7746 passed / 0 failed**. Full sharded `cargo test -p engine` runs green in CI.
- wasm-dev check (`cargo check -p engine-wasm --target wasm32-unknown-unknown` with `-D warnings`) — clean.
- Card-data / coverage / semantic-audit run green in CI (**Card data** job); the local pipeline needs an MTGJSON download unavailable in my sandbox.

## Tests (drive the real pipeline; fail on revert)
- `drizzt_dies_trigger_puts_difference_counters_gated_by_power_comparison` (`parser/oracle_tests.rs`) — full `parse_oracle_text`: the dies trigger's effect is `PutCounter` with a `Difference{ Power{EventSource}, Power{Source} }` count (NOT `Unimplemented`), and the trigger carries the strict-GT power gate.
- `drizzt_difference_counters_from_dying_creature_lki_power` (`game/effects/counters.rs`) — **end-to-end runtime** off the real parsed trigger: a 5-power creature dies → Drizzt (power 2) gains 3 (=5−2) counters; equal (2) and smaller (1) → 0 (strict-GT gate).
- `difference_counter_anaphor_without_comparison_stays_unimplemented` — coverage-honesty: an unbindable "the difference" (no comparison) surfaces as a loud `Unimplemented`, not a silent-0 `PutCounter`.

## CR
CR 122.1 (counters) + CR 208.1 (P/T) + CR 603.4 (intervening-if) + CR 603.10a (last-known-information look-back for a creature that has left the battlefield). Verified against `docs/MagicCompRules.txt`.

Closes #4967

Tier: Frontier
Model: claude-opus-4-8
Thinking: high

### DIFF
--- a/crates/engine/src/game/effects/counters.rs
+++ b/crates/engine/src/game/effects/counters.rs
@@ -3117,6 +3117,102 @@ mod tests {
         );
     }
 
+    /// CR 122.1 + CR 603.4 + CR 603.10a: Drizzt Do'Urden — "Whenever a creature
+    /// dies, if it had power greater than Drizzt's power, put a number of +1/+1
+    /// counters on Drizzt equal to the difference." End-to-end through the real
+    /// parser: a larger-power creature dying gates the trigger on and puts
+    /// `dyingPower - drizztPower` counters (read from LKI, CR 603.10a); an
+    /// equal/smaller creature fails the gate and adds none. Fails on revert
+    /// (parser leaves the effect Unimplemented / drops the gate → 0 counters).
+    #[test]
+    fn drizzt_difference_counters_from_dying_creature_lki_power() {
+        use crate::game::stack::resolve_top;
+        use crate::game::triggers::process_triggers;
+        use crate::types::triggers::TriggerMode;
+
+        // Parse Drizzt's dies trigger from Oracle text (real pipeline).
+        let parsed = crate::parser::parse_oracle_text(
+            "Double strike\n\
+             Whenever a creature dies, if it had power greater than Drizzt's power, \
+             put a number of +1/+1 counters on Drizzt equal to the difference.",
+            "Drizzt Do'Urden",
+            &[],
+            &["Creature".to_string()],
+            &["Elf".to_string(), "Ranger".to_string()],
+        );
+        let dies_trigger = parsed
+            .triggers
+            .iter()
+            .find(|t| {
+                matches!(t.mode, TriggerMode::ChangesZone)
+                    && t.execute
+                        .as_ref()
+                        .is_some_and(|e| matches!(&*e.effect, Effect::PutCounter { .. }))
+            })
+            .unwrap_or_else(|| panic!("Drizzt dies PutCounter trigger not parsed: {parsed:#?}"))
+            .clone();
+
+        // Run the dies scenario with a creature of the given power; return the
+        // number of +1/+1 counters Drizzt ends up with.
+        let run = |dying_power: i32| -> u32 {
+            let mut state = GameState::new_two_player(42);
+
+            let drizzt_id = create_object(
+                &mut state,
+                CardId(1),
+                PlayerId(0),
+                "Drizzt Do'Urden".to_string(),
+                Zone::Battlefield,
+            );
+            {
+                let d = state.objects.get_mut(&drizzt_id).unwrap();
+                d.power = Some(2);
+                d.toughness = Some(3);
+                d.card_types.core_types.push(CoreType::Creature);
+                d.trigger_definitions.push(dies_trigger.clone());
+            }
+
+            let dying_id = create_object(
+                &mut state,
+                CardId(2),
+                PlayerId(1),
+                "Hill Giant".to_string(),
+                Zone::Battlefield,
+            );
+            {
+                let g = state.objects.get_mut(&dying_id).unwrap();
+                g.power = Some(dying_power);
+                g.toughness = Some(3);
+                g.card_types.core_types.push(CoreType::Creature);
+            }
+
+            let mut events = Vec::new();
+            crate::game::zones::move_to_zone(&mut state, dying_id, Zone::Graveyard, &mut events);
+            process_triggers(&mut state, &events);
+            while !state.stack.is_empty() {
+                let mut resolve_events = Vec::new();
+                resolve_top(&mut state, &mut resolve_events);
+            }
+
+            state.objects[&drizzt_id]
+                .counters
+                .get(&CounterType::Plus1Plus1)
+                .copied()
+                .unwrap_or(0)
+        };
+
+        // Larger power (5) than Drizzt (2): gate passes, +1/+1 counters = 5 - 2 = 3.
+        assert_eq!(
+            run(5),
+            3,
+            "5-power creature dying should give Drizzt 3 (=5-2) +1/+1 counters"
+        );
+        // Equal power (2): strict GT gate fails, no counters.
+        assert_eq!(run(2), 0, "equal-power creature must not add counters");
+        // Smaller power (1): gate fails, no counters.
+        assert_eq!(run(1), 0, "smaller-power creature must not add counters");
+    }
+
     /// Regression test: MoveCounters must use LKI when the source has changed zones.
     /// Simulates Essence Channeler's "When this creature dies, put its counters on
     /// target creature you control" — the source is in the graveyard with no counters,
--- a/crates/engine/src/parser/oracle_effect/counter.rs
+++ b/crates/engine/src/parser/oracle_effect/counter.rs
@@ -2,7 +2,7 @@ use crate::parser::oracle_nom::error::{OracleError, OracleResult};
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take_till, take_until};
 use nom::character::complete::space1;
-use nom::combinator::{eof, opt, peek, value};
+use nom::combinator::{all_consuming, eof, opt, peek, value};
 use nom::sequence::terminated;
 use nom::Parser;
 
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
**[MED]** The current parse-diff shows scope/coverage contamination outside Drizzt. Evidence: the current `<!-- coverage-parse-diff -->` sticky reports Conformer Shuriken changed from `ability/put` unsupported coverage to a supported-looking `PutCounter (conditional=target has Unknown (instead), counter=Variable { name: "difference" } P1P1, target=self)`, while the PR title/body scope is Drizzt’s dies trigger. Why it matters: an unbound `difference` placeholder under an unsupported conditional is exactly the false-green coverage shape this PR is trying to avoid; either Conformer’s attack trigger must be bound correctly to its target/equipped-creature comparison, or it must remain an honest unsupported residual. Suggested fix: extend the binding/honesty guard to the Conformer-style conditional branch or keep that card’s parser output unsupported, then regenerate parse-diff so only intentional card-level deltas remain.

**[MED]** The unresolved Gemini whitespace finding is valid. Evidence: `crates/engine/src/parser/oracle_effect/counter.rs:520` only calls `trim_end()` before matching `all_consuming(tag("the difference"))`, so an `after_equal` phrase with leading whitespace fails the anaphor parser even though it is the same semantic phrase. Why it matters: this is a fragile parser edge in the new grammar leaf. Suggested fix: parse the trimmed phrase, and compute the returned remainder from the consumed original slice so clause-boundary behavior remains correct.

**[LOW]** The unbound-placeholder fallback is still hardcoded to a put-counter description. Evidence: `crates/engine/src/parser/oracle_trigger.rs:1348` replaces any unbound count placeholder found through `count_expr_mut()` with `Effect::unimplemented("put", "counters equal to the difference")`. Why it matters: the helper is intentionally effect-agnostic, so the fallback should not misdescribe future effect shapes that carry the same placeholder. Suggested fix: match the current effect variant when building the unimplemented name/description, or restrict this fallback to counter effects only.

dhgoal:
Thanks — all three addressed at `00dd1195`. The root cause of the Conformer leak was that `Effect::count_expr_mut()` only reaches the **top-level** effect, so a placeholder nested one level down escaped both the bind and the guard.

**[MED] Conformer Shuriken false-green — fixed by recursing the effect tree.** Probing the real Conformer-shaped trigger confirmed the placeholder nests: `execute.effect` is the `SetTapState` (the tap), and the `PutCounter { count: Variable{"difference"}, target: SelfRef }` lives in `execute.sub_ability.effect`, gated by a **clause-level** `condition`, *not* a hoisted `def.condition` (which is `None`). Replaced both top-level-only `count_expr_mut()` blocks (the inline bind + the coordinator downgrade) with a single recursive resolver `resolve_difference_anaphor_in_ability` that walks `sub_ability`, `else_ability`, and the one heterogeneous `Box<Effect>` wrapper (`CreateDrawReplacement`). Per placeholder: **bind** to the real `Difference` when a hoisted `QuantityComparison` exists (Drizzt, top-level), or **downgrade to `Effect::Unimplemented`** when there's nothing to bind (Conformer, nested). So Conformer is back to its pre-PR `ability/put` **unsupported** state — no surviving `PutCounter{Variable}` anywhere in its tree.

New test `nested_conditional_difference_anaphor_downgrades_to_unimplemented` parses the Conformer-shaped attacks trigger and asserts (a) a **recursive** walk finds no surviving `PutCounter`/`PutCounterAll` with a `Variable{"difference"}` count anywhere (effect/sub_ability/else_ability), and (b) `has_unimplemented(execute)` — fails on revert of the recursion.

**[MED] Leading-whitespace edge — fixed.** `parse_counter_equal_to` now captures `take_till`'s remainder and matches the anaphor on the **fully-trimmed** phrase (`phrase_raw.trim()`, both ends) so `"equal to  the difference"` still binds, returning the `take_till` remainder to preserve any trailing `", …"` clause. The `parse_cda_quantity` branch keeps its prior `trim_end()` behavior unchanged.

**[LOW] Honest downgrade name — fixed.** Both bind and downgrade are now restricted to `Effect::PutCounter`/`PutCounterAll` (the only variants that ever carry this placeholder), so the `Effect::unimplemented("put", …)` name/description no longer depends on `count_expr_mut` being effect-agnostic.

**Verification:** `cargo fmt` clean; combinator gate (`git diff origin/main -- 'crates/engine/src/parser/**'`) prints nothing; `cargo clippy -p engine --tests -- -D warnings` clean; targeted suites `cargo test -p engine --lib -- parser:: game::triggers:: game::quantity:: game::effects::counters::` → **7747 passed, 0 failed**. The three Drizzt tests (parse-gate, runtime LKI counters, unbound-stays-unimplemented) stay green. Rebased onto latest `main`; the draw/lose difference mechanism is untouched.

matthewevans:
Acknowledged current head `00dd1195`. The follow-up targets the previous findings, but I’m holding re-review until current-head evidence is available: the `coverage-parse-diff` sticky still predates this commit, and card-data / Rust test jobs are still pending on the current CI run. I’ll re-review once the current parse-diff and jobs finish.

matthewevans (approved):
Re-reviewed current head `00dd1195` with the fresh parse-diff. The prior blocker is resolved: Drizzt's intervening-if power comparison now binds the `difference` counter quantity at the trigger seam, and the Conformer Shuriken spillover is kept as unsupported instead of a false-green `PutCounter` with an unbound variable. The runtime test drives the real trigger pipeline/LKI behavior, and the parser tests cover both unbound and nested placeholders.

dhgoal:
Parse-diff regenerated on `00dd1195` (Card data job green). Confirming the Conformer outcome and flagging one honest residual for your call.

**Conformer false-green is gone.** Conformer Shuriken is now `ability/put` (unsupported) on both sides — no `ability/PutCounter{Variable{"difference"}}` anywhere in its tree. Probed baseline vs branch: the nested effect is `Unimplemented { name: "put", … }` in both. So the card is an honest unsupported residual, exactly as requested.

**Residual (your call):** the regenerated diff still lists Conformer as a *net-zero* `ability/put` add+remove of the **same displayed signature**. The only difference is the `Unimplemented` **description** string:

- baseline: `"put a number of +1/+1 counters on ~ equal to the difference"` (the raw clause)
- this branch: `"counters equal to the difference"` (my downgrade's fragment)

Same variant, same structure, same `conditional=target has Unknown (instead)`, same unsupported status — it's a description refinement, not a coverage or false-green change. The reason it isn't byte-identical: the downgrade runs in `lower_trigger_ir` on the already-lowered `AbilityDefinition`, which doesn't carry the per-clause `source_text` (that lives on the effect-chain clause IR). Matching baseline's description exactly would mean either threading per-clause provenance onto the effect or reconstructing the raw clause via string manipulation — the latter is the kind of thing the parser architecture forbids.

Two ways forward, your preference:
- **(A)** accept the honest unsupported residual as-is (Conformer stays unsupported; only the diagnostic string differs), or
- **(B)** I thread the clause `source_text` through so the downgrade emits baseline's exact `Unimplemented` description, dropping Conformer out of the parse-diff entirely (only Drizzt's intended deltas remain).

Happy to do (B) if you'd rather see a Drizzt-only diff — just didn't want to add provenance plumbing unprompted. Everything else from the round-2 review (whitespace, effect-restricted fallback, recursive guard) is in `00dd1195`; all required checks green.

PR #4995, opened 2026-07-03
fix(engine): clamp negative mill and exile-top counts to zero (CR 107.1b)

ultrahighsuper:
## Summary

The **mill** (`mill.rs`) and **exile-top** (`exile_top.rs`) resolvers cast their resolved count `as usize` with no lower bound. Mill shares the Draw/Mill/Discard dynamic-count parser (`oracle_effect/imperative.rs`: *"cross-verb dynamic-count idioms shared by Draw/Mill/Discard"*), so a subtractive count — "mill cards equal to A minus B" when `B > A` — resolves negative. `-2 as usize` wraps to a huge value, and the downstream library-size `min` (CR 701.17b, "a player can't mill/exile more than their library") then mills / exiles the target's **entire library** instead of nothing.

Both get the same one-liner every sibling count-resolver already applies (`draw.rs`, `discard.rs`, `sacrifice.rs`):

```rust
resolve_quantity_with_targets(state, count, ability).max(0) as usize,
```

Mill and exile-top were the two remaining unclamped count-casts among the effect resolvers; the counter/life resolvers already clamp (`.max(0)`) or use negatives as intentional "remove all" sentinels, so they are untouched.

> Two one-line resolver fixes restoring an invariant every sibling resolver already holds (`.max(0)` before the `as usize` cast), each with a resolver-level regression test. No new pattern, variant, or AST. A `/review-impl` self-check was run.

## CR references

- `CR 107.1b` — "If a calculation that would determine the result of an effect yields a negative number, zero is used instead."
- `CR 701.17b` — a player can't mill/exile more cards than are in their library (the downstream `min` that turns the wrap into a whole-library drain).

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib effects::mill::tests` (12) and `effects::exile_top::tests` (12) — all pass.
- Added `mill_negative_count_clamps_to_zero` and `exile_top_negative_count_clamps_to_zero`. Each resolves the effect with `count = HandSize{You} − HandSize{Opponent}` (opponent holds more) and asserts the target's library is untouched (0 milled/exiled). **Revert-probe:** without the `.max(0)`, each test fails because the target's entire library is drained — proving the wrap is real, reachable, and the assertions are non-vacuous.

### DIFF
--- a/crates/engine/src/game/effects/exile_top.rs
+++ b/crates/engine/src/game/effects/exile_top.rs
@@ -18,7 +18,11 @@ pub fn resolve(
         } => (
             // Use resolve_quantity_with_targets so that TargetZoneCardCount (and
             // DivideRounded wrapping it) can resolve against the targeted player.
-            resolve_quantity_with_targets(state, count, ability) as usize,
+            // CR 107.1b: clamp a negative result to zero before the `as usize`
+            // cast — a subtractive count would otherwise wrap huge, and the
+            // downstream library-size `min` would exile the entire library
+            // instead of nothing. Mirrors the guard in `draw.rs` / `discard.rs`.
+            resolve_quantity_with_targets(state, count, ability).max(0) as usize,
             player.clone(),
             *face_down,
         ),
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[MED]** This leaves the same CR 107.1b negative-count wrap in a sibling resolver while claiming the family invariant is restored. Evidence: the PR body says `discard.rs` already applies the same `.max(0)` guard, but current `crates/engine/src/game/effects/discard.rs` still resolves `Effect::Discard` with `resolve_quantity_with_targets(state, inner, ability) as u32`. That has the same failure mode as the fixed sites: a negative dynamic count wraps to a huge unsigned value, then the hand-size `min` path can force the whole hand to be discarded instead of zero. Please include the discard resolver and a matching revert-probe test, or narrow the PR claim and explain why discard is semantically exempt from CR 107.1b here.

ultrahighsuper:
Good catch — this was a branch-staleness artifact, now fixed. Pushed `e9d429a63`.

The discard clamp landed separately in #4993 (merged into `main` at 08:21). This branch was cut *before* that merge, so its base still had the unclamped `discard.rs` you (and Gemini) saw — hence the comment referencing a guard that wasn't yet on the base. I've rebased onto current `main`, so `discard.rs` now carries `.max(0)` in context and the `"Mirrors ... draw.rs / discard.rs"` comments are accurate.

So the CR 107.1b family invariant is complete across the count resolvers: `draw.rs` + `discard.rs` (#4993) + `sacrifice.rs` already clamp; this PR adds the two remaining unclamped sites, **mill** and **exile-top**. I kept it scoped to those two rather than re-touching `discard.rs` since #4993 already owns that fix. The counter/life resolvers were left untouched — they already clamp or use negatives as intentional "remove all" sentinels.

Verification after the rebase: `cargo test -p engine --lib negative_count_clamps_to_zero` passes all three (mill, exile-top, and discard's own test from the base); each new test is a revert-probe (without `.max(0)` the whole library is drained). fmt + `clippy -p engine --lib` clean.

matthewevans (approved):
Current head re-reviewed. The previous sibling-clamp blocker is resolved by the branch now being based on main with the discard clamp present; mill/exile-top use the same CR 107.1b guard, and the added negative-count tests are revert-probes for the actual wrap failure.

PR #5001, opened 2026-07-03
fix(engine): route chosen-player mana to the named recipient (CR 106.4)

ultrahighsuper:
## Summary

"Choose a player. That player adds one mana of any color they choose" (**Spectral Searchlight**, **Stadium Vendors**, **Mad Science Fair Project** mode 2) deposited the mana into the **controller's** pool instead of the chosen player's.

The main resolve path (`mana.rs`, `resolve`) already derives the recipient from `Effect::Mana.target`, resolving player context-refs (`ScopedPlayer`, `SourceChosenPlayer`, …) via `resolve_player_for_context_ref`. But the color-prompt **completion** path, `handle_choose_mana_effect` — which is *always* taken when the production offers more than one color (`AnyOneColor`/`AnyCombination` with the full WUBRG set), since `mana_choice_prompt` inspects only `produced` — destructured `Effect::Mana { …, target: _ }` and hardcoded `recipient = ability.controller`, silently dropping the named recipient.

The fix derives the recipient in the completion path exactly as `resolve` does. Priority still returns to the controller (unchanged — a player receiving mana does not thereby gain priority; only the mana is redirected).

These effects are **not** mana abilities (`is_mana_ability` returns false when `target` is set), so they always route through the stack → `resolve` → prompt → `handle_choose_mana_effect`, meaning the correct-recipient logic in `resolve` was never reached for this whole card class.

> A few-line fix that makes the prompt-completion path derive the mana recipient the same way the non-prompt `resolve` path already does (no new pattern/variant/AST), plus a revert-probe test. A `/review-impl` self-check was run; priority behavior is deliberately left unchanged to keep the change scoped to the recipient bug.

## CR references

- `CR 106.4` — "When an effect instructs a player to add mana, that mana goes into *that* player's mana pool."

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib effects::mana::tests` — 35 passed.
- New `choose_mana_effect_routes_to_named_recipient_not_controller`: P0 controls the effect and chooses opponent P1 as recipient (`scoped_player`), P1 picks Blue; asserts the Blue mana lands in **P1's** pool and **P0's** pool stays empty. Revert-probe: with the recipient hardcoded to the controller the test fails (P1 receives 0), proving the class-level bug and a non-vacuous assertion.

### DIFF
--- a/crates/engine/src/game/effects/mana.rs
+++ b/crates/engine/src/game/effects/mana.rs
@@ -4,15 +4,36 @@ use crate::game::{mana_payment, mana_sources};
 use crate::types::ability::ManaContribution;
 use crate::types::ability::{
     ChoiceValue, Effect, EffectError, EffectKind, LinkedExileScope, ManaProduction,
-    ManaSpendRestriction, ObjectScope, ResolvedAbility, TargetRef,
+    ManaSpendRestriction, ObjectScope, ResolvedAbility, TargetFilter, TargetRef,
 };
 use crate::types::events::GameEvent;
 use crate::types::game_state::{
     GameState, ManaChoice, ManaChoiceContext, ManaChoicePrompt, MayTriggerOrigin, WaitingFor,
 };
 use crate::types::mana::{ManaColor, ManaRestriction, ManaType};
+use crate::types::player::PlayerId;
 
-/// Mana effect: adds mana to the controller's mana pool (CR 106.4).
+/// CR 106.4: The player who receives (and, when a color is chosen, chooses) the
+/// mana produced by an `Effect::Mana`. A subject-led / chosen-player clause
+/// ("Choose a player. That player adds one mana of any color they choose" —
+/// Spectral Searchlight, Stadium Vendors) carries a player context-ref in
+/// `recipient_filter`; a non-player-ref filter or `None` leaves the controller
+/// as the recipient. Shared by the immediate `resolve` path, the color-choice
+/// prompt, and the prompt-completion path so all three agree on the recipient.
+fn mana_effect_recipient(
+    state: &GameState,
+    ability: &ResolvedAbility,
+    recipient_filter: &Option<TargetFilter>,
+) -> PlayerId {
+    match recipient_filter {
+        Some(filter) if filter.is_context_ref() => {
+            super::resolve_player_for_context_ref(state, ability, filter)
+        }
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The named recipient still does not make the mana-color choice on the production path.

Evidence: `resolve()` still publishes the color-choice prompt as `WaitingFor::ChooseManaColor { player: ability.controller, ... }` at `crates/engine/src/game/effects/mana.rs:50-54`. The new recipient logic only runs later in `handle_choose_mana_effect()`, and the added test calls that helper directly rather than starting from `resolve()` / the actual `WaitingFor` prompt path.

Why it matters: for Spectral Searchlight / Stadium Vendors-style text, the PR can now deposit mana into the chosen player's pool after completion, but the controller is still the player asked to choose the color. That is still wrong for "that player adds one mana of any color they choose".

Suggested fix: resolve the player context-ref recipient before publishing `WaitingFor::ChooseManaColor` for resolving mana effects, set the prompt `player` to that recipient, and add a production-path test that starts at `resolve()` and asserts both the prompted player and final mana recipient.

ultrahighsuper:
Good catch — the prompt player was the deeper half. Fixed in `61d30123d`.

Both the production `resolve()` path (which publishes `WaitingFor::ChooseManaColor`) and the completion path `handle_choose_mana_effect` now derive the recipient from `Effect::Mana.target` via a shared `mana_effect_recipient` helper (resolving player context-refs like `ScopedPlayer`). So for "that player adds one mana of any color they choose", the chosen player is now both **prompted to pick the color** and **receives the mana**; priority still returns to the controller.

The test now starts at `resolve()`, asserts the `ChooseManaColor` prompt is directed to the chosen player, then runs the completion and asserts the mana lands in that player's pool (and not the controller's). Revert-probe: neutering the recipient derivation makes the prompt go to the controller and the test fails on the prompt assertion. `effects::mana::tests` (35) + clippy green.

matthewevans (approved):
Approved after re-reviewing the updated head. The previous blocker is addressed: the production `resolve()` path now directs `WaitingFor::ChooseManaColor` to the resolved mana recipient, and `handle_choose_mana_effect()` deposits to the same recipient. The regression test drives the real prompt path and would fail if either the prompt routing or final mana deposit reverted.

PR #5002, opened 2026-07-03
fix(parser): variable-first "X plus/minus N" count offset — restore the dropped bonus (Crater's Claws, Flame Discharge)

e11734937-beep:
## Summary

Count expressions of the form **"X plus N" / "X minus N"** (variable-first) parsed the amount as a bare `X`, **silently dropping the `+N`/`−N`**. So cards like **Crater's Claws** ("X plus 2 damage"), **Flame Discharge** / **Light Up the Night** ("X plus 1 damage"), **Meteor Shower**, and **Hellfire** ("X plus 3 damage") resolved as *X* damage instead of *X+N* — a correctness bug, even though the cards were already classified "supported" via their primary clause.

The engine already had the integer-first dual, **"N plus/minus X"** (Slumbering Trudge's "three minus X", "two plus X"), which lowers to `QuantityExpr::Offset`. Only the variable-first ordering was missing.

## Change

Add the variable-first arm to `parse_count_expr` (`crates/engine/src/parser/oracle_util.rs`), mirroring the existing integer-first offset arm with the roles swapped: after the bare `X` ref, a `plus `/`minus ` connective followed by a **literal integer** (via `parse_number`) yields `Offset { inner: X, offset: +/-N }` (sign folded directly into the offset, no `Multiply` wrapper). Maps to the existing `QuantityExpr::Offset` — no new effect/IR; the resolver already clamps a negative result to zero (CR 107.1b).

**No over-reach:** the arm fires only when a literal integer follows the connective. A dynamic operand (*"X plus the number of creatures you control"*) makes `parse_number` fail, so it falls through to the bare `X` ref with the connective left on the remainder (regression-tested).

## Tests

- `parse_count_expr_x_plus_two` / `parse_count_expr_x_minus_one` — recognizer → `Offset{X, +2}` / `Offset{X, -1}`.
- `parse_count_expr_x_plus_dynamic_stays_bare_x` — dynamic operand is NOT swallowed into an offset.
- `effect_deal_x_plus_two_damage_uses_offset_amount` — end-to-end: "deals X plus 2 damage to any target" → `DealDamage { amount: Offset{X,+2} }` (previously a bare `X` that dropped the +2).

Fails-before (amount was bare `X` / `Unimplemented`) / passes-after confirmed. Green: rustfmt, parser-combinator (Rule Zero) gate, `clippy -D warnings`, full engine suite (14733 passed / 0 failed). `cargo coverage` byte-identical before/after (this is a correctness fix to already-"supported" cards, not a coverage gain — no regression, and the shared `parse_count_expr` changed no other card's count parse: the arm only fires on a literal-integer offset).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/tests.rs
+++ b/crates/engine/src/parser/oracle_effect/tests.rs
@@ -3007,6 +3007,33 @@ fn effect_lightning_bolt() {
 }
 
 #[test]
+fn effect_deal_x_plus_two_damage_uses_offset_amount() {
+    // Flame Discharge / Light Up the Night: "deals X plus N damage" composes the
+    // spell's announced X with a fixed bonus. The DealDamage amount must be
+    // Offset { X, offset: N } via parse_count_expr, not a bare X with a dropped
+    // "plus N" tail (the pre-fix behavior that left the effect unsupported).
+    let e = parse_effect("~ deals X plus 2 damage to any target");
+    match e {
+        Effect::DealDamage { amount, target, .. } => {
+            assert_eq!(target, TargetFilter::Any);
+            match amount {
+                QuantityExpr::Offset { inner, offset } => {
+                    assert_eq!(offset, 2);
+                    assert!(matches!(
+                        *inner,
+                        QuantityExpr::Ref {
+                            qty: QuantityRef::Variable { .. }
+                        }
+                    ));
+                }
+                other => panic!("expected Offset {{X, +2}} amount, got {other:?}"),
+            }
+        }
+        other => panic!("expected DealDamage, got {other:?}"),
+    }
+}
+
+#[test]
 fn effect_damage_to_each_opponent_uses_player_scope() {
     let e = parse_effect("~ deals 1 damage to each opponent");
     assert!(matches!(
--- a/crates/engine/src/parser/oracle_util.rs
+++ b/crates/engine/src/parser/oracle_util.rs
@@ -509,13 +509,68 @@ pub fn parse_count_expr(text: &str) -> Option<(QuantityExpr, &str)> {
                     return Some((expr, ""));
                 }
             }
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
**[MED]** The "literal-only" `X plus/minus N` branch still accepts a second dynamic `X` and turns it into zero.

Evidence: the new branch calls `parse_number(after_op)` at `crates/engine/src/parser/oracle_util.rs:532`, but `parse_number` explicitly maps `X` to `0` at `oracle_util.rs:371-376`. The new no-overreach test covers `"X plus the number ..."`, but not `"X plus X"` or `"X minus X"`.

Why it matters: `X plus X` / `X minus X` would now be consumed as supported `X + 0` / `X - 0` instead of preserving or rejecting the dynamic operand, contradicting the branch's stated literal-only scope and risking another false-green quantity parse.

Suggested fix: use a literal-number parser that excludes `X` for this variable-first offset branch, then add regression tests proving `X plus X` and `X minus X` remain unconsumed/unsupported rather than becoming an offset by zero.

e11734937-beep:
Good catch, fixed. The variable-first offset arm now guards the operand with a nom negative-lookahead — `not(terminated(tag("x"), alt((eof, space1))))` via the nom_on_lower bridge — so it forms an `Offset` only when the operand is a genuine literal, not the standalone `X` (which `parse_number` maps to 0). `X plus X` / `X minus X` now fall through to the bare `X` ref with the connective left on the remainder, exactly as `X plus the number of …` already did. Added the two regressions `parse_count_expr_x_plus_x_not_offset` / `parse_count_expr_x_minus_x_not_offset` (fail-before: they parsed to `Offset{X, 0}`; pass-after: bare `X`). Full suite 14737 green, clippy + parser-combinator gate clean.

matthewevans (approved):
Maintainer review: latest head fixes the prior X-plus-X false-green issue; parser surface and discriminating tests look good. Queuing when checks are ready.

PR #5003, opened 2026-07-03
feat(combat): directional attack restriction — "attack only the nearest opponent in the chosen direction" (CR 508.1c)

real-venus:
## Summary

Implements the **CR 508.1c** global directional attack restriction — "Each player may attack only the nearest opponent in the [last] chosen direction and planeswalkers controlled by that opponent" — unlocking three previously-unsupported cards:

- **Pramikon, Sky Rampart** — `As Pramikon enters, choose left or right.`
- **Mystic Barrier** — `When this enchantment enters and at the beginning of your upkeep, choose left or right.`
- **Teyo, Geometric Tactician** — `[−2]: Choose left or right. Until your next turn, …` (duration-bound)

## Approach — compose existing primitives, no new subsystem

The restriction and its linked "choose left or right" are built from existing building blocks; there is **no new AI, frontend, or serialization surface** (the choice UI, layer grant, and combat-declaration machinery are all reused).

- **`StaticMode::AttackOnlyNeighbor`** (nullary, peer of `MustAttackPlayer`). The direction is **not** a parse-time field — it's read live from the source's chosen direction — so all present and future printings share one variant and one combat gate.
- **`ChosenAttribute::Direction(SeatDirection)`** (peer of `Label`), with a typed `SeatDirection::from_choice_label` helper and a `chosen_direction()` accessor. `bind_named_choice` hijacks the **exact two-option `{Left, Right}`** `Labeled` choice into a typed `Direction` (CR 607.2d linked choice), replacing the prior direction on re-choose ("last chosen"). A `Labeled` choice that merely *includes* "Left" among other options stays a `Label` — so nothing else regresses.
- **Combat gate** in `declare_attackers_with_bands`: for every attacker, the direction is read from each active `AttackOnlyNeighbor` source (CR 109.5), the nearest opponent is resolved **per attacker** via `players::neighbor` (skips eliminated players, wraps, 2-player collapses to the sole opponent), and an attack is rejected unless it targets that neighbor or a planeswalker they control. Existence-gated (a no-op when no such static is on the battlefield); multiple sources **intersect** (each is an independent CR 508.1c check).
- **Teyo's `[−2]`** parses to a duration-bound `GrantStaticAbility` (CR 611.2a "until your next turn"); the granted static lands on Teyo and reads Teyo's chosen direction, and expires at the controller's next turn. Teyo's ETB Wall token and `[+1]` are out of scope (parse independently).

Coverage is via registry membership (`handle_rule_mod` marker); enforcement is **solely** the combat gate (no double enforcement).

## Rules

CR 508.1c (governing) + CR 109.5 (source-controller "you") + CR 102.1 / 103.1 (seating / neighbor order) + CR 607.2d (linked choose ↔ chosen) + CR 611.2a / 611.2c (Teyo duration) + CR 400.7 (chosen attribute cleared on zone change). All verified against the Comprehensive Rules.

## Testing

**10 runtime combat tests** through the real `declare_attackers` path:
- non-neighbor attack rejected / neighbor + their planeswalker allowed,
- global scope (every player restricted, not just the source's controller),
- 2-player collapse (any attack legal),
- no-direction-yet is inert,
- "last chosen" re-choice flips legality,
- two-source intersection,
- **Teyo's grant gates attacks and expires at the controller's next turn** (installed via the real transient-grant → layer-materialization → `bind_named_choice` → prune path, revert-failing).

Plus the `{Left, Right}` hijack hostile fixtures (a 3-option `{Left, Center, Right}` choice stays a `Label`; re-choose replaces the prior direction) and static/effect parser tests. Full `cargo test -p engine` green; clippy clean.

### DIFF
--- a/crates/engine/src/game/combat.rs
+++ b/crates/engine/src/game/combat.rs
@@ -30,24 +30,35 @@ use crate::types::zones::Zone;
 /// rejects on `def.mode != mode` first), so `flag && check_static_ability(..)`
 /// is byte-identical to the original call while skipping the redundant scan.
 ///
-/// Representation: five named compile-time presence flags, NOT a no-bool-flags
-/// anti-pattern — these are five independent existence facts, not one
+/// Representation: named compile-time presence flags, NOT a no-bool-flags
+/// anti-pattern — these are independent existence facts, not one
 /// choice-encoding bool. An `EnumSet` is rejected because `enumset` is not a
 /// workspace dependency and `StaticMode` is not fieldless (it carries data
 /// variants such as `MaxUntapPerType { filter, max }` and `Other(String)`).
-/// Five named flags read clearer than a runtime set lookup.
+/// Named flags read clearer than a runtime set lookup.
 struct CombatStaticGates {
     has_cant_attack: bool,
     has_cant_attack_or_block: bool,
     has_must_attack: bool,
     has_goad: bool,
     has_can_attack_with_defender: bool,
+    /// CR 508.1c: any functioning `StaticMode::AttackOnlyNeighbor` present.
+    has_attack_only_neighbor: bool,
 }
 
 impl CombatStaticGates {
-    /// One `game_functioning_statics` sweep computing all five presence flags.
+    /// One `game_functioning_statics` sweep computing the presence flags.
     /// Does NOT increment the static-full-scan perf counter: this is the single
     /// intended hoisted sweep, not a per-element legality scan.
+    ///
+    /// `has_attack_only_neighbor` is deliberately computed from the SAME
+    /// collection the CR 508.1c enforcement loop iterates
+    /// (`battlefield_active_statics`, battlefield-only), not the broader
+    /// `game_functioning_statics` (battlefield + command zone) used for the
+    /// other flags. This keeps the gate flag and the enforcement in lockstep:
+    /// the flag is set iff a battlefield source will actually be enforced. A
+    /// command-zone grant of `AttackOnlyNeighbor` is out of scope and would
+    /// otherwise set the flag but never be enforced (silent under-enforcement).
     fn compute(state: &GameState) -> Self {
         let mut gates = CombatStaticGates {
             has_cant_attack: false,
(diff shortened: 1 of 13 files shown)


matthewevans (changes requested):
**[MED]** Directional attack restriction resolves the next seat, not the nearest opponent. Evidence: `crates/engine/src/game/combat.rs:2510` uses `players::neighbor(state, attacker_controller, dir)`, but `players::neighbor` is documented/implemented as next-or-previous living player only (`crates/engine/src/game/players.rs:70-80`), while opponent identity is team-based (`crates/engine/src/game/topology.rs:76-78`). Why it matters: in Two-Headed Giant / team formats, an adjacent teammate becomes the required target, so the actual nearest opponent in that direction is rejected and the restriction can make legal attacks impossible. Suggested fix: add/use a direction walker that skips non-opponents with `players::is_opponent(state, attacker_controller, candidate)`, and add a shared-team-turn regression where the adjacent seat is a teammate and the next seat is the legal opponent.

real-venus:
@matthewevans 

Thanks — addressed on the current head (07f4878e7), implemented exactly as you suggested.

- Added `players::nearest_opponent(state, controller, direction)`: it walks the seat ring in the chosen direction and **skips living teammates** via `players::is_opponent` (CR 102.2), returning the first actual opponent (`None` only if the caster is the sole survivor). In a free-for-all it's identical to `neighbor`; in team formats it steps past an adjacent teammate.
- The CR 508.1c gate now calls `nearest_opponent` instead of `players::neighbor` — the `combat.rs` line you cited no longer uses `neighbor`.
- Added the shared-team-turn regression you asked for: a **Two-Headed Giant** scenario where the attacker's adjacent seat is a teammate and the next seat is the legal opponent — attacking that nearest opponent (past the teammate) is now legal, which was impossible before. Plus `nearest_opponent` unit tests covering free-for-all equivalence, the 2HG teammate-skip, and the sole-survivor case.

Full `cargo test -p engine` is green. Re-review whenever you get a chance — thanks!

matthewevans (approved):
Reviewed current head. The prior nearest-opponent blocker is fixed at the right seam with `players::nearest_opponent`, so team-format seats skip living teammates instead of treating the adjacent seat as the legal defender. The implementation keeps the linked left/right choice typed (`ChosenAttribute::Direction`), reuses a single nullary `AttackOnlyNeighbor` marker for printed and temporary grants, and the parse-diff is scoped to Pramikon, Mystic Barrier, and Teyo. Runtime combat coverage discriminates FFA, 2HG teammate skipping, global restriction, re-choice, and Teyo's duration-bound grant.

PR #5004, opened 2026-07-03
fix(engine): consume filtered one-shot spell cost reductions (CR 601.2f)

ultrahighsuper:
## Summary

"The next [type] spell you cast this turn costs {N} less" (**Kadena, Slinking Sorcerer**; **Slinking Sorcerer**) kept discounting **every** matching spell for the rest of the turn, not just the next one.

`apply_pending_spell_cost_reductions` (`casting.rs`) matches filtered entries via `spell_matches_cost_filter` and applies the **first** (it `break`s), but `consume_pending_spell_cost_reduction` only removed entries whose `spell_filter` was `None`. So a filtered `PendingSpellCostReduction` was applied on every matching cast and never consumed until the end-of-turn `clear()` — a class-level over-reduction for the whole "next … spell costs {N} less" family.

The fix consumes the **same** entry `apply` used: the first for this caster that the cast spell matches (unfiltered **or** filter-matched), threading the finalized spell's `object_id` through the single call site. This mirrors the already-correct sibling `consume_pending_next_spell_modifiers` (the parallel one-shot "next spell has convoke / can't be countered" system), which was the source of the asymmetry.

> A few-line fix that aligns the consumer's predicate with the applier's (and with the sibling `consume_pending_next_spell_modifiers`), plus its call site and a unit revert-probe test. No new pattern/variant/AST. A `/review-impl` self-check was run.

## CR references

- `CR 601.2f` — total cost is the mana cost "minus all cost reductions"; a one-shot "next spell" reduction applies to exactly one spell.

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib cost_reduction` — 98 passed.
- New `consume_filtered_pending_cost_reduction_removes_the_matching_entry`: pushes a filtered (creature-spell) reduction, casts a creature spell, and asserts the entry is consumed; plus an unfiltered-still-consumed regression and a different-player-not-consumed precision check. **Revert-probe:** with the old `spell_filter.is_none()` predicate the filtered entry survives and the test fails — proving the class-level over-reduction and a non-vacuous assertion.

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -6269,13 +6269,30 @@ fn apply_pending_spell_cost_reductions(
     }
 }
 
-/// CR 601.2f: Consume (remove) a one-shot pending cost reduction after a spell is cast.
-pub(super) fn consume_pending_spell_cost_reduction(state: &mut GameState, caster: PlayerId) {
-    if let Some(idx) = state
-        .pending_spell_cost_reductions
-        .iter()
-        .position(|r| r.player == caster && r.spell_filter.is_none())
-    {
+/// CR 601.2f: Consume (remove) the one-shot pending cost reduction a cast spell
+/// used. Removes the first entry for this caster that the cast spell matches —
+/// whether unfiltered OR filter-matched (e.g. "the next face-down creature spell
+/// you cast this turn costs {3} less", Kadena) — mirroring the single entry that
+/// [`apply_pending_spell_cost_reductions`] applied (it also stops at the first
+/// match). The previous predicate removed only *unfiltered* entries, so a
+/// filtered reduction was never consumed and kept discounting every matching
+/// spell for the rest of the turn instead of just the next one. Mirrors
+/// [`consume_pending_next_spell_modifiers`].
+pub(super) fn consume_pending_spell_cost_reduction(
+    state: &mut GameState,
+    caster: PlayerId,
+    spell_id: ObjectId,
+) {
+    let matched = state.pending_spell_cost_reductions.iter().position(|r| {
+        r.player == caster
+            && match &r.spell_filter {
+                None => true,
+                Some(filter) => {
+                    spell_matches_cost_filter(state, caster, spell_id, filter, spell_id)
+                }
+            }
+    });
+    if let Some(idx) = matched {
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**[MED]** The new regression test bypasses the production cast path that this bug actually lives in. Evidence: the production behavior is `apply_pending_spell_cost_reductions` during cost calculation (`crates/engine/src/game/casting.rs:6251-6267`) followed by `finalize_cast_with_phyrexian_choices` consuming the matched entry (`crates/engine/src/game/casting_costs.rs:6202-6203`), but the only new test constructs a stack object and calls `consume_pending_spell_cost_reduction` directly (`crates/engine/src/game/casting_tests.rs:22928-22952`). Why it matters: a wrong call-site/object-id/threading bug in the real cast pipeline would still pass this test while filtered one-shot discounts could keep applying across real casts. Suggested fix: add a runtime cast test (scenario/runner or existing cast helper) that grants a filtered pending reduction, casts two matching creature spells, and proves only the first cast receives/consumes the discount.

ultrahighsuper:
Added the runtime cast test you asked for.

New end-to-end test `crates/engine/tests/filtered_cost_reduction_only_first_cast.rs` drives the real pipeline via `GameScenario` `cast().resolve()`: it grants a creature-filtered {1} one-shot reduction, then casts two {2} creature spells. It asserts the **first** cast spends {1} (discounted) and consumes the reduction (`pending_spell_cost_reductions` empty), and the **second** cast spends the full {2}. Revert-probe: with the old `spell_filter.is_none()` consumer the reduction survives and the first-cast consumption assertion fails.

Kept the direct unit test as a fast focused check on the consumer predicate, but the production-path coverage now proves the finalize call-site/object-id threading through the actual cast finalize (`finalize_cast_with_phyrexian_choices` -> `consume_pending_spell_cost_reduction`).

matthewevans (approved):
Thanks, this addresses the production-path gap. I re-traced the fix against the current head: `apply_pending_spell_cost_reductions` and `consume_pending_spell_cost_reduction` now use the same first-matching predicate for unfiltered and filtered pending reductions, and the new `GameScenario` integration test drives the real cast/finalize path by casting two matching creature spells and proving only the first receives/consumes the one-shot discount. The remaining checks can gate the merge queue.

PR #5005, opened 2026-07-03
fix(parser): gate extra-blocker grants on trailing "as long as/if" conditions (Entourage of Trest)

minion1227:
## Problem

An extra-blocker grant carrying a trailing `as long as <cond>` / `if <cond>` gate failed to parse **entirely**:

- Entourage of Trest: `This creature can block an additional creature each combat **as long as you're the monarch**.`

The rider left non-empty remainder in `parse_extra_blockers_static` (which requires the line to end after the extra-block predicate), so the whole line fell through to a failed effect and the extra-block grant was **silently dropped**.

## Fix

Peel the trailing condition in the `dispatch.rs` arm **before** delegating to `parse_extra_blockers_static`: split on ` as long as ` / ` if `, parse the condition-free body as the extra-block static, and attach the parsed condition via `parse_static_condition` (**CR 611.3a**).

`you're the monarch` resolves to `StaticCondition::IsMonarch` through the existing shared condition authority — **no new condition parsing**. The bare (unconditional) form is unchanged.

### Before / after

```
before: (no static — line failed to a dropped effect, extra-block grant lost)
after:  ExtraBlockers{ count: 1 }  condition: IsMonarch
```

## Tests

- New `extra_blockers_static_gated_on_trailing_as_long_as_condition` regression: asserts the `ExtraBlockers{1}` grant carries the `IsMonarch` gate, plus the bare form still parses with no condition.
- 961 `oracle_static` tests pass; `cargo fmt` + `clippy -D warnings` clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -2860,23 +2860,20 @@ pub(crate) fn parse_static_line_inner(
     // CR 509.1c + CR 611.3a: an extra-blocker grant may carry a trailing "as long
     // as <cond>" / "if <cond>" gate (Entourage of Trest: "This creature can block
     // an additional creature each combat as long as you're the monarch"). The bare
-    // form is parsed directly; when a trailing gate is present, peel it so the
-    // condition-free body reaches `parse_extra_blockers_static`, then attach the
-    // parsed condition (IsMonarch, etc.) via `parse_static_condition`.
+    // form is parsed directly; a trailing gate is peeled via the shared gate
+    // authority (`split_trailing_gate_condition_with_body`, which owns the
+    // `as long as` > last-valid-`if` > `as if`-exclusion rules) so the
+    // condition-free body reaches `parse_extra_blockers_static`, then the parsed
+    // condition attaches. CR 611.3a: fail CLOSED — an unrecognized condition leaves
+    // the whole line unsupported (`parse_static_condition` returns `None` → `?`)
+    // rather than enforcing the extra-block grant unconditionally (an `Unrecognized`
+    // gate evaluates as always-true in the layer system).
     if let Some(def) = parse_extra_blockers_static(&text) {
         return Some(def);
     }
-    if let Some((body_tp, condition_tp)) = tp
-        .split_around(" as long as ")
-        .or_else(|| tp.split_around(" if "))
-    {
-        if let Some(mut def) = parse_extra_blockers_static(body_tp.original) {
-            let condition_text = condition_tp.original.trim().trim_end_matches('.');
-            def.condition = Some(parse_static_condition(condition_text).unwrap_or(
-                StaticCondition::Unrecognized {
-                    text: condition_text.to_string(),
-                },
-            ));
+    if let Some((body, condition_text)) = split_trailing_gate_condition_with_body(&tp) {
+        if let Some(mut def) = parse_extra_blockers_static(body) {
+            def.condition = Some(parse_static_condition(condition_text)?);
             return Some(def);
         }
     }
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -2856,9 +2856,29 @@ pub(crate) fn parse_static_line_inner(
         return Some(def);
     }
 
-    // --- "can block an additional creature" / "can block any number" ---
+    // --- "can block an additional creature [as long as|if <cond>]" / "can block any number" ---
+    // CR 509.1c + CR 611.3a: an extra-blocker grant may carry a trailing "as long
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[HIGH]** This adds a new ad hoc trailing-gate parser instead of using the existing static gate machinery. Evidence: `crates/engine/src/parser/oracle_static/dispatch.rs:2869-2875` splits on literal `" as long as "` / `" if "`, while `crates/engine/src/parser/oracle_static/shared.rs:2114-2125` already centralizes trailing `as long as` / `if` gate parsing, including last-marker and `as if` handling. Please route this through that shared authority, or add a body-preserving helper beside it, rather than proliferating another parser split in dispatch.

**[HIGH]** Unrecognized gated extra-blocker conditions fail open as active statics. Evidence: `dispatch.rs:2875-2879` falls back to `StaticCondition::Unrecognized`, and `crates/engine/src/game/layers.rs:1114` evaluates `Unrecognized` as true. That means an unsupported rider can still grant `ExtraBlockers` unconditionally. The gated branch should fail closed when `parse_static_condition(condition_text)` returns `None`, matching the existing `split_trailing_gate_condition` contract that unsupported conditions leave the line unsupported instead of enforced unconditionally.

**[MED]** The tests only cover the happy-path `as long as you're the monarch` case. Please add discriminating coverage for the `if` branch and an unsupported trailing condition staying unsupported, so the tests would fail for both the ad hoc split and the `Unrecognized` fallback.

minion1227:
Thanks for the review — all three addressed in `fd9df1be4`.

**[HIGH] ad hoc split →** Replaced the dispatch-local `split_around(" as long as "/" if ")` with a body-preserving sibling of the shared authority: `split_trailing_gate_condition_with_body` (added beside `split_trailing_gate_condition` in `shared.rs`). It reuses the exact same word-boundary scan — `as long as` first, then the last valid `if` with `as if` exclusion — and returns the original-case `(body, condition)` by slicing `tp.original` at the marker byte offset (mirroring `split_trailing_if_condition_tp`). No trailing-gate splitting is re-implemented in dispatch.

**[HIGH] fail-open →** Now fails **closed**: `def.condition = Some(parse_static_condition(condition_text)?)`. An unrecognized rider makes the whole line unsupported instead of granting `ExtraBlockers` unconditionally — matching the `split_trailing_gate_condition` contract you cited.

**[MED] tests →** Added (a) the `if` branch (`if you control a Forest` → `IsPresent` gate) and (b) an unrecognized trailing condition that must leave the line unsupported (`parse_static_line(...).is_none()`) — this fails for both the old ad hoc split and the `Unrecognized` fallback.

961 oracle_static tests pass; fmt + `clippy -D warnings` clean.

matthewevans:
Reviewed the follow-up at `fd9df1b`. The code changes now land on the requested seam: the extra-blocker dispatch reuses the shared body-preserving trailing-gate helper, fails closed on unrecognized conditions, and covers the `as long as`, `if`, bare, and unrecognized-rider cases.

I am holding approval/enqueue for current parser evidence: the parse-diff sticky comment still reflects the earlier `4bafef9` head (`2026-07-03T11:46:28Z`), while the current head is `fd9df1b`, and `Card data (generate, validate, coverage)` is still in progress. Once the current-head parse diff/checks refresh without unexpected card-scope changes, this looks approvable from the backend/parser review.

minion1227:
Thanks — and the current-head evidence has now refreshed. On `fd9df1b`:

- **All checks are green**, including `Card data (generate, validate, coverage)` and `Rust (fmt, clippy, test, coverage-gate)`; merge state is CLEAN.
- The parse-diff is limited to the **2 expected cards**, with no unexpected card-scope changes:
  - **Entourage of Trest** → `ExtraBlockers(1) (is monarch)`, `effect_structure` removed — the target fix.
  - **Foriysian Totem** → `Continuous` → `ExtraBlockers(1) (source is creature)` — the same shared trailing-gate helper also upgrades its "as long as ~ is a creature, it can block an additional creature" rider from a generic Continuous to the correctly-gated `ExtraBlockers`. Expected and correct, not scope creep.

(The output is identical to `4bafef9`; `fd9df1b` only swapped the internal split for the shared `split_trailing_gate_condition_with_body` seam, so the diff content is unchanged.)

Ready for approval / enqueue whenever you have a moment — thanks for the thorough review.

matthewevans (approved):
Backend re-review on `fd9df1be`: the current-head evidence has caught up and the parser change is now on the right seam.

The extra-blocker gate uses the shared body-preserving trailing-gate splitter instead of a local ad hoc split, attaches recognized `as long as` / `if` conditions, and fails closed for unrecognized riders. The tests cover gated, bare, `if`, and unknown-rider cases, and the parse-diff is limited to the expected Entourage/Foriysian scope. Approving and enabling merge-when-ready.

PR #5009, opened 2026-07-03
fix(parser): count only creature cards for Dread Summons' token repeat

### LINKED ISSUE #4746: Dread Summons bug — The game uses X as the number of tokens created by [[Dread Summons]], when the number of tokens sho…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521084900936122398/1521084900936122398

**Thread:** Dread Summons bug
**Cards:** dread summons
**Parser status:** has_gaps

## Summary
The game uses X as the number of tokens created by [[Dread Summons]], when the number of tokens should be equal to the number of creature cards milled

### dread summons
Each player mills X cards. For each creature card put into a graveyard this way, you create a tapped 2/2 black Zombie creature token. (To mill a card, a player puts the top card of their library into their graveyard.)

---
<sub>report_id: `discord:1521084900936122398:1521084900936122398:0` · discord: `1521084900936122398/1521084900936122398`</sub>

nickmopen:
Closes #4746.

## Problem

**Dread Summons** — "Each player mills X cards. **For each creature card** put into a graveyard this way, you create a tapped 2/2 black Zombie creature token." — created **X** tokens (one per *card* milled) instead of one per **creature** card milled.

The token's repeat count parsed to the **unfiltered `TrackedSetSize`** (every card moved this way = X), dropping the "creature" qualifier.

## Cause

In `oracle_effect/token.rs`, the "for each `<thing>` this way" token-count arm only routed the *distinct-card-types* shape (Occult Epiphany's "for each card type among cards … this way") to a filtered count; **every other form** — including "for each **creature** card … this way" — fell back to `TrackedSetSize`.

## Fix

Fall back to the shared `parse_for_each_clause` (which already restricts the tracked set by a leading type phrase → `FilteredTrackedSetSize`), and adopt its result **only when it restricts to a specific card type**. A bare/generic "card … this way" (e.g. "card discarded this way") stays `TrackedSetSize`, preserving existing behavior.

Verified:

| Clause | Token count |
|---|---|
| `for each creature card put into a graveyard this way` | `FilteredTrackedSetSize{Creature}` ✅ |
| `for each card put into a graveyard this way` (bare) | `TrackedSetSize` (unchanged) |
| `for each card discarded this way` (bare) | `TrackedSetSize` (unchanged) |
| `for each card type among cards discarded this way` (Occult Epiphany) | `DistinctCardTypes` (unchanged) |

## Tests

New `for_each_creature_card_this_way_counts_only_creatures`; the existing `bare_for_each_card_discarded_this_way_keeps_tracked_set_size` and Occult Epiphany tests still pass.

`cargo test -p engine --lib` → **14,676 passed / 0 failed**; `fmt --check`, `clippy -p engine --all-targets -- -D warnings`, and `check-parser-combinators.sh` all clean.

CR 609.3 + CR 205.2a.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/token.rs
+++ b/crates/engine/src/parser/oracle_effect/token.rs
@@ -1390,7 +1390,6 @@ mod tests {
         // way, you create a … token." The token count must restrict to CREATURE
         // cards moved this way (`FilteredTrackedSetSize`), not every card
         // (`TrackedSetSize`, which would create X tokens for X cards milled).
-        use crate::types::ability::TargetFilter;
         let txt = "Create a tapped 2/2 black Zombie creature token for each creature card put into a graveyard this way.";
         let effect = try_parse_token(&txt.to_lowercase(), txt, &mut ParseContext::default())
             .expect("expected Token effect");
--- a/crates/engine/src/parser/oracle_effect/token.rs
+++ b/crates/engine/src/parser/oracle_effect/token.rs
@@ -10,7 +10,7 @@ use crate::parser::oracle_ir::context::ParseContext;
 use crate::parser::oracle_nom::error::OracleResult;
 use crate::types::ability::{
     ContinuousModification, ControllerRef, Effect, FilterProp, ObjectScope, PtValue, QuantityExpr,
-    QuantityRef, StaticDefinition, TargetFilter,
+    QuantityRef, StaticDefinition, TargetFilter, TypeFilter,
 };
 use crate::types::card_type::Supertype;
 use crate::types::keywords::Keyword;
@@ -328,6 +328,23 @@ pub(crate) fn parse_token_description(text: &str) -> Option<TokenDescription> {
     parse_token_description_with_context(text, &ParseContext::default())
 }
 
+/// True iff a `for each … this way` count restricts to a specific card type
+/// (Dread Summons' "creature card"), so it should override the unfiltered
+/// `TrackedSetSize`. A bare/generic "card" filter (e.g. "card discarded this
+/// way") is not restrictive and keeps `TrackedSetSize`.
+fn tracked_set_count_is_type_restricted(qty: &QuantityRef) -> bool {
+    let QuantityRef::FilteredTrackedSetSize { filter, .. } = qty else {
+        return false;
+    };
+    let TargetFilter::Typed(typed) = filter.as_ref() else {
+        return false;
+    };
+    typed
+        .type_filters
+        .iter()
+        .any(|type_filter| !matches!(type_filter, TypeFilter::Card))
+}
+
 fn parse_token_description_with_context(
     text: &str,
     ctx: &ParseContext,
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
Thanks for the focused parser fix. I need changes before this can be accepted.

**[MED]** The new helper adds an inline import inside a function. Evidence: `tracked_set_count_is_type_restricted` declares `use crate::types::ability::{QuantityRef, TargetFilter, TypeFilter};` inside the function body in `crates/engine/src/parser/oracle_effect/token.rs`. This repository’s instructions explicitly require all imports at the top of the file, and this file already imports the surrounding ability types at module scope. Please move `TypeFilter` into the existing top-level `crate::types::ability` import list and remove the function-local `use`.

Also, this parser/card-surface PR still needs the current-head `coverage-parse-diff` sticky before approval; card-data is still running and the sticky is absent right now.

nickmopen:
Fixed the inline import — `TypeFilter` is now in the module-scope `use crate::types::ability::{…}` list (alongside the `QuantityRef`/`TargetFilter` already there), and the function-local `use` in `tracked_set_count_is_type_restricted` is removed. `fmt --check`, `clippy -p engine --all-targets -- -D warnings`, and `check-parser-combinators.sh` all clean; the new test still passes.

On the `coverage-parse-diff` sticky: CI is re-running on the new head. This PR *does* change card parsing (that's the point — Dread Summons' token repeat goes from unfiltered `TrackedSetSize` to `FilteredTrackedSetSize{Creature}`), so I expect the parse-diff to show exactly that one card. I'll surface the current-head parse-diff artifact here once the card-data job finishes (the auto-post workflow_run sticky often doesn't fire for fork PRs).

nickmopen:
### Parse changes introduced by this PR · head `314d62790` · 2 card(s), 2 signature(s) (baseline `main 571de7b30610`)

Surfacing the CI-generated `parse-diff` artifact (the auto-post workflow_run sticky didn't fire for this fork head). From the `Card data` run: https://github.com/phase-rs/phase/actions/runs/28663217835

Both signatures are the intended change — an unfiltered `cards moved` token count → `filtered tracked set (creature)`:

- **ability/Token** · `cards moved× 2/2 Black Zombie` → `filtered tracked set (creature)× …` — **Syphon Flesh** ("for each creature *sacrificed* this way")
- **ability/Token** · `cards moved× Mutagen` → `filtered tracked set (creature)× …` — **Shellshock** ("for each creature *dealt damage* this way")

Both are genuinely "for each **creature** … this way", so restricting the count to creatures is correct — the fix generalizes to the whole type-restricted "this way" class rather than being Dread-Summons-specific. **Dread Summons** itself is the one card the diff notes as excluded (unrelated Oracle-text errata in the regenerated data), but it's the same corrected signature. Bare/generic "card … this way" and the Occult-Epiphany distinct-types form are unchanged.

So the card-level blast radius is exactly these three, all correct.

matthewevans (approved):
Maintainer review: current head is clean for this parser fix. I checked the token parser seam, current-head parse-diff evidence, and the focused tests: the change delegates the token count phrase to the existing filtered tracked-set quantity path, keeps the bare-card fallback covered, and the current parse-diff is limited to the intended token-count changes. This clears the stale change request.

PR #5010, opened 2026-07-03
fix(parser): bind orphaned "choose new targets for the copy" across intervening clauses (CR 707.10c)

dhgoal:
Tier: Frontier
Model: claude-opus-4-8
Thinking: high

## Summary
CR 707.10c: a trailing **"you may choose new targets for the copy/copies"** sentence granted the copy's controller retargeting only when it *immediately* followed the `CopySpell`. When a clause resolves between the copy and the sentence — or a reflexive `"If you do, "` gate precedes it — the clause orphaned to a loud `Effect::unimplemented("orphaned_copy_retarget", …)` residual, so the copy silently fell back to `KeepOriginalTargets`.

This is a **class** fix for the copy-retarget building block, not a single card. Self-sourced via `cargo parser-gaps` (the `orphaned_copy_retarget` residual class). It binds 4 real cards and correctly **preserves the honest residual** for the 6 that are genuinely blocked (no `CopySpell` exists — their copy sits inside an as-yet-unimplemented `When …` delayed trigger). Parser/lowering only — the engine already models retargeting via `CopyRetargetPermission` (CR 707.10c); no runtime change.

**Fixed (residual `orphaned_copy_retarget` 10 → 6):**
- **Narset's Reversal** — retarget follows the `"then return it to its owner's hand"` bounce.
- **Increasing Vengeance** — retarget follows the conditional `"copy that spell twice instead"`; the primary always-made copy binds `MayChooseNewTargets`.
- **Spinerock Tyrant** — retarget follows the `"those spells gain wither"` rider.
- **Spider-Verse** — nearest clause *is* the copy, but the `"If you do, "` reflexive gate defeated the recognizer.

**Preserved as honest residual (unchanged, 6):** Twinferno, Pyromancer's Goggles, Primal Wellspring, Pit Automaton, Sunken Palace, Dynaheir — each wraps the copy in an unimplemented `When …` delayed trigger, so there is no `CopySpell` to bind; implementing those triggers is separate, larger work. A "choose new targets" with no preceding copy stays a loud gap (coverage-honest), asserted by a dedicated test.

## Files changed

## CR references
- `CR 707.10c` — a spell's copy may have new targets chosen for it.
- `CR 608.2b` — reflexive "if you do" gate on the optional copy (Spider-Verse).

## Gate A
```
[exit 0 — no violations]
```
Note: the local script diffs against an empty base and can false-pass, so I also ran the authoritative check against upstream — `git diff origin/main -- 'crates/engine/src/parser/**' | grep '^+' | grep -E '<FORBIDDEN_METHODS>'` — which printed **nothing**. No new `.contains("`/`.split(`/`.starts_with("`/`.strip_suffix(`/etc. in parser dispatch; the change reuses the existing nom `recognize_copy_retarget_clause` combinator (extended only with an `opt(tag("if you do, "))` axis).

## Anchored on
- crates/engine/src/parser/oracle_effect/sequence.rs:2861 — existing `apply_clause_continuation` handler that patches a preceding clause via `defs.last_mut()`; the new `CopyMayRetarget` arm generalizes this to a nearest-first `defs.iter_mut().rev()` backward scan.
- crates/engine/src/parser/oracle_effect/sequence.rs:2810 — existing `set_copy_retarget` / `set_copy_retarget_in_ability` descent through `CreateDelayedTrigger` + `sub_ability`, reused verbatim by both the detection lookback and the binding scan.
- crates/engine/src/parser/oracle_effect/mod.rs:225 — existing `parse_followup_continuation_ast` `.and_then/.or_else` clause-lookback chain that my `.or_else()` deeper-lookback extends (same combinator-chain idiom).

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
Developer track (Tilt not running in this environment → ran cargo directly with the repo's `CC` linker):
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — exit 0; authoritative `git diff origin/main` combinator grep — **empty** (clean)
- `cargo test -p engine --lib -- parser::` — **7237 pass / 0 fail** (2 ignored)
- copy-machinery regression (`copy_retarget copy_spell twincast fork chain_of breeches reverberate galvanic krark`) — **85 pass / 0 fail**; 5 new fail-on-revert tests pass by name
- `oracle-gen` regenerate → `orphaned_copy_retarget` residual **10 → 6** (the 6 remaining are the intended honest residuals)
- `cargo clippy-strict` (whole-workspace) and `cargo coverage`/`semantic-audit` need OpenSSL/`pkg-config`/MTGJSON beyond this sandbox — see the green CI **Rust lint** and **Card data** jobs as authoritative for those.

## Scope Expansion
None.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -23090,6 +23090,26 @@ pub(crate) fn parse_effect_chain_ir(
                         _ => None,
                     }
                 })
+            })
+            .or_else(|| {
+                // CR 707.10c: a "you may choose new targets for the copy/copies"
+                // sentence still grants the copy's controller retargeting even
+                // when a clause that resolves between the copy and this sentence
+                // separates them — Narset's Reversal's "then return it to its
+                // owner's hand" bounce, Increasing Vengeance's conditional "copy
+                // that spell twice instead", Spinerock Tyrant's "those spells
+                // gain wither" rider. The nearest-clause lookback above only sees
+                // that intervening clause, so scan the non-absorbed clauses
+                // (nearest-first) for the closest one whose effect wraps a
+                // `CopySpell` and bind the permission there. If NO preceding copy
+                // exists, fall through so the clause stays an honest orphaned
+                // residual rather than a silently-wrong retarget.
+                let lower = normalized_text.to_lowercase();
+                let all_copies = sequence::copy_retarget_clause_all_copies(&lower)?;
+                non_absorbed
+                    .iter()
+                    .any(|c| sequence::effect_wraps_copy_spell(&effective_effect_of(c)))
+                    .then_some(ContinuationAst::CopyMayRetarget { all_copies })
             });
         let absorb_followup = followup_continuation
             .as_ref()
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -2,7 +2,7 @@ use crate::parser::oracle_nom::error::{OracleError, OracleResult};
 use nom::branch::alt;
 use nom::bytes::complete::{tag, tag_no_case, take_until};
 use nom::character::complete::multispace1;
-use nom::combinator::{all_consuming, eof, opt, value};
+use nom::combinator::{all_consuming, eof, map, opt, value};
 use nom::sequence::{preceded, terminated};
 use nom::Parser;
 
@@ -2685,7 +2685,7 @@ pub(super) fn push_clause_chunk(
 /// spell / triggered contexts) or nested inside a `CreateDelayedTrigger`
 /// wrapper ("When you next cast ..., copy that spell"). Mirrors
 /// `def_tree_has_optional`'s descent through `CreateDelayedTrigger`.
-fn effect_wraps_copy_spell(effect: &Effect) -> bool {
+pub(super) fn effect_wraps_copy_spell(effect: &Effect) -> bool {
     match effect {
         Effect::CopySpell { .. } => true,
         Effect::CreateDelayedTrigger { effect: inner, .. } => {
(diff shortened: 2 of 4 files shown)


matthewevans (changes requested):
Maintainer review: this needs changes before it can be considered for merge.

**[MED]** Spider-Verse now accepts the haste rider with the wrong authority. Evidence: the current parse-diff adds `grant Haste (affects=triggering source)` for Spider-Verse, and the new test includes `If the copy is a permanent spell, it gains haste` but only asserts that the retarget permission is present. In this trigger, the triggering source is the original spell, not the copied spell object. That makes the coverage greener while modeling the rider on the wrong object/timing. Please either leave this rider as an honest residual for now, or model haste on the copied spell path and add a test that asserts the copy gets the haste grant.

**[MED]** Plural copy-retarget binding can leave replacement copies unpatched. Evidence: `set_copy_retarget` returns after the first `CopySpell`, and `CopyMayRetarget` stops walking after the first patched ability. The Increasing Vengeance case says `copy that spell twice instead. You may choose new targets for the copies`, but the test only checks that at least one collected retarget contains `MayChooseNewTargets`. Please patch all relevant copy effects for plural/all-copy retarget clauses, or thread the permission into the replacement branch and assert every collected copy retarget is correct.

**[LOW]** The new reflexive-gate CR annotation cites the wrong rule. Evidence: `parse_copy_retarget_clause` cites `CR 608.2b` for the `if you do` gate, but `608.2b` is target legality; it is not the instruction-order/reflexive gate authority. Please verify the rule against `docs/MagicCompRules.txt` and either use the correct CR citation or remove that part of the annotation.

dhgoal:
Thanks — all three addressed at `48792850`.

**[MED] Spider-Verse false-green haste → reverted to honest residual.** The `opt(tag("if you do, "))` axis was only ever needed for Spider-Verse, so I removed it entirely; Spider-Verse's retarget is back to an honest `orphaned_copy_retarget` residual (residual count 10→7). On the haste rider: I probed `origin/main` directly (parse `Spider-Verse` on a clean main checkout) and confirmed `AddKeyword(Haste)` on the triggering source is **pre-existing on main**, not introduced by this PR — round-1's parse-diff only attributed it to the PR because binding the retarget changed the card's overall signature. With the gate gone, Spider-Verse's parse now matches main exactly, so this PR introduces **no** Spider-Verse delta (the regenerated parse-diff should show none). The mis-targeted haste is a separate pre-existing issue; happy to file it, but it's out of scope here.

**[MED] Increasing Vengeance plural — every copy patched.** Threaded plurality through the clause: `parse_copy_retarget_clause` now returns whether it's the plural "the copies" form; `ContinuationAst::CopyMayRetarget { all_copies }` carries it; and `set_copy_retarget_in_ability` gained an `all` mode that descends the sub-ability chain patching **every** `CopySpell` (primary + the conditional "copy that spell twice" copy) instead of returning at the first. Singular "the copy"/"that copy" keeps nearest-only binding. The test now asserts Increasing Vengeance produces **two** CopySpells and **both** carry `MayChooseNewTargets`.

**[LOW] Wrong CR → fixed.** The `608.2b` citation went away with the removed `if you do` gate. Verified `608.2c` is the instruction-order / later-text-modifies-earlier authority in docs/MagicCompRules.txt.

**Verification:** `cargo fmt` clean; combinator gate (`git diff origin/main`) empty; `cargo clippy -p engine --tests -- -D warnings` clean; `cargo test -p engine --lib -- parser::` **7242 pass / 0 fail**; copy-machinery regression (twincast/fork/chain_of/breeches/reverberate/galvanic/krark/copy_spell) **56 pass / 0 fail**; regenerated card-data → `orphaned_copy_retarget` **10→7** (Narset's Reversal, Spinerock Tyrant, Increasing Vengeance fixed; Spider-Verse + the 6 delayed-trigger-blocked cards remain honest residuals). Rebased onto latest main.

matthewevans (approved):
Reviewed current head. The copy-retarget continuation is parameterized for singular/plural copies, walks back to the nearest copy-bearing ability across intervening clauses, and keeps the no-prior-copy case as an honest orphaned residual. Parse-diff evidence is current and scoped to the intended cards.

PR #5011, opened 2026-07-03
fix(engine): Lost Mine of Phandelver room fidelity (Storeroom target, Fungi Cavern duration)

ultrahighsuper:
## Summary

Two **Lost Mine of Phandelver** room effects in `dungeon.rs` diverged from the printed Oracle text. (Dungeon room text is hard-coded in `dungeon.rs` and has no card-data cross-check, so I verified both against Scryfall — quoted below.)

**Storeroom** — Oracle: *"Put a +1/+1 counter on target creature."*
The code restricted the target to `TypedFilter::creature().controller(You)` (creature *you control*). Fixed to any creature. Beyond forbidding the legal choice of an opponent's creature, the restriction made the room's triggered ability (CR 309.4c) find **no legal target** — removed on resolution (CR 608.2b) — when the venturing player controlled no creatures but an opponent did. This also matches the sibling unrestricted counter-rooms Undercity **Forge** and **Throne of the Dead Three**, which already use plain `TypedFilter::creature()`.

**Fungi Cavern** — Oracle: *"Target creature gets -4/-0 until your next turn."*
The code used `Duration::UntilEndOfTurn`, ending the debuff a full turn early. Fixed to `Duration::UntilNextTurnOf { Controller }` (CR 514.2 + CR 611.2a).

(Scryfall, Lost Mine of Phandelver: `Storeroom — Put a +1/+1 counter on target creature.` / `Fungi Cavern — Target creature gets -4/-0 until your next turn.`)

> Two one-line Oracle-fidelity corrections to a hard-coded room-effect table (a target-filter and a duration), plus AST tests. No new pattern/variant. A `/review-impl` self-check was run.

## CR references

- `CR 309.4c` — a room's triggered ability. `CR 608.2b` — a spell/ability with no legal targets is removed on resolution.
- `CR 514.2` + `CR 611.2a` — "until your next turn" duration (Fungi Cavern).

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib dungeon::tests` — passes, incl. two new AST tests: `lost_mine_storeroom_targets_any_creature` (asserts `tf.controller.is_none()`) and `lost_mine_fungi_cavern_lasts_until_your_next_turn` (asserts `UntilNextTurnOf { Controller }`). Each directly asserts the corrected AST field, so reverting either fix flips its assertion (the old values were `Some(You)` / `UntilEndOfTurn`).

### DIFF
--- a/crates/engine/src/game/dungeon.rs
+++ b/crates/engine/src/game/dungeon.rs
@@ -203,15 +203,15 @@ pub fn room_effects(
             simple(treasure_token(), source_id, controller),
             vec![],
         ),
-        // 3: Storeroom — "Put a +1/+1 counter on target creature you control"
+        // 3: Storeroom — "Put a +1/+1 counter on target creature"
+        // (any creature; the Oracle text is unrestricted, matching the sibling
+        // counter-rooms Undercity "Forge" / "Throne of the Dead Three").
         (DungeonId::LostMineOfPhandelver, 3) => (
             simple(
                 Effect::PutCounter {
                     counter_type: CounterType::Plus1Plus1,
                     count: fixed(1),
-                    target: TargetFilter::Typed(
-                        TypedFilter::creature().controller(ControllerRef::You),
-                    ),
+                    target: TargetFilter::Typed(TypedFilter::creature()),
                 },
                 source_id,
                 controller,
@@ -237,7 +237,9 @@ pub fn room_effects(
             )));
             (lose, vec![])
         }
-        // 5: Fungi Cavern — "Target creature gets -4/-0 until end of turn"
+        // 5: Fungi Cavern — "Target creature gets -4/-0 until your next turn"
+        // CR 514.2 + CR 611.2a: the debuff persists until the *beginning* of the
+        // controller's next turn, not just end of turn.
         (DungeonId::LostMineOfPhandelver, 5) => (
             ResolvedAbility::new(
                 Effect::Pump {
@@ -249,7 +251,9 @@ pub fn room_effects(
                 source_id,
                 controller,
             )
-            .duration(Duration::UntilEndOfTurn),
+            .duration(Duration::UntilNextTurnOf {
+                player: PlayerScope::Controller,
+            }),
             vec![],
         ),
         // 6: Temple of Dumathoin — "Draw a card"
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
The hard-coded room-table edits are at the right location and the two shape assertions are useful revert probes, but this still needs one runtime/scenario-level check before I can approve it.

For Storeroom, please drive the actual room ability through the engine with the venturing player controlling no creatures and the opponent controlling a creature, then prove the opponent creature can be targeted / receives the +1/+1 counter. The current `tf.controller.is_none()` assertion proves the AST field, but not the target-selection/resolution behavior that was broken.

For Fungi Cavern, please add a scenario that resolves the room pump and advances through cleanup/turn transition to prove the -4/-0 survives end of turn and expires at the controller's next turn boundary. The current `Duration::UntilNextTurnOf { Controller }` assertion proves the enum value, but not the lifecycle behavior.

This is a small PR, but dungeon rooms are hard-coded engine behavior with no card-data parse-diff safety net, so the runtime tests need to discriminate the real game behavior, not only the constructed effect shape.

ultrahighsuper:
Added the two runtime engine-path tests you asked for (both drive the real venture pipeline via `GameScenario`/`GameRunner`, and both revert-probe against the old room definitions):

- `lost_mine_storeroom_targeting_runtime.rs`: the controller controls **no** creatures and the opponent controls one. Venture -> choose Storeroom -> the room's `+1/+1 counter on target creature` auto-targets the only legal creature (the opponent's) and resolves; asserts that creature now has the counter. Under the old `creature you control` restriction there is no legal target, so the ability is removed on resolution (CR 608.2b) and the test fails (the ability is not even on the stack).

- `lost_mine_fungi_cavern_duration_runtime.rs`: venture -> choose Fungi Cavern -> resolve `-4/-0` on a 4/4 (power -> 0). It then drives the REAL turn machinery: after the controller's own end-of-turn cleanup (into the opponent's turn) the power is still 0 (**survives** end of turn), and at the controller's next turn it is back to 4 (**expires**). The old `Duration::UntilEndOfTurn` restores the power a full turn early, failing the survive-end-of-turn assertion.

AST shape assertions kept as fast revert probes. `fmt` + `clippy -p engine --lib` clean.

matthewevans (approved):
Reviewed current head. The room effects now match Oracle text, and the added runtime tests drive the actual venture path: Storeroom can target an opponent's creature, and Fungi Cavern survives end of turn before expiring at the controller's next turn. CR citations checked against the local rules file.

PR #5012, opened 2026-07-03
fix(engine): implement Dungeon of the Mad Mage "Lost Level" as Scry 2 (was a wrong stub)

ultrahighsuper:
## Summary

The **"Lost Level"** room of Dungeon of the Mad Mage (room 4) was stubbed as `Effect::Unimplemented` with **fabricated** text â€” `"Destroy target creature of an opponent's choice"` â€” so venturing into it did nothing.

The printed Oracle text is simply **"Scry 2"** (Scryfall, Dungeon of the Mad Mage: `Lost Level â€” Scry 2.`), a fully-supported effect. This implements it as `Effect::Scry { count: 2, target: Controller }`, mirroring the Scry rooms already coded correctly in the same table (this dungeon's "Dungeon Level", Lost Mine's "Cave Entrance").

Dungeon room text is hard-coded in `dungeon.rs` with no card-data cross-check, hence the Scryfall citation.

- [x] **Not** `/engine-implementer` â€” explain why below

> Replaces a single wrong `Effect::Unimplemented` stub with the correct existing `Effect::Scry` (no new pattern/variant), plus an AST test. A `/review-impl` self-check was run.

## CR references

- `CR 309.4c` â€” a room's triggered ability. `CR 701.22` â€” Scry. (Effect itself is already implemented; this is a fidelity/coverage fix.)

## Verification

- `cargo fmt --all` â€” clean.
- `cargo clippy -p engine --lib` â€” clean (0 warnings).
- `cargo test -p engine --lib dungeon::tests` â€” passes, incl. new `mad_mage_lost_level_is_scry_two` asserting the room resolves to `Effect::Scry { count: Fixed(2), target: Controller }`. Revert-probe: the old `Effect::Unimplemented` stub fails the match.

### DIFF
--- a/crates/engine/src/game/dungeon.rs
+++ b/crates/engine/src/game/dungeon.rs
@@ -320,14 +320,12 @@ pub fn room_effects(
             });
             (ability, vec![])
         }
-        // 4: Lost Level — "Destroy target creature of an opponent's choice"
+        // 4: Lost Level — "Scry 2"
         (DungeonId::DungeonOfTheMadMage, 4) => (
             simple(
-                Effect::Unimplemented {
-                    name: "Room: Lost Level".to_string(),
-                    description: Some(
-                        "Destroy target creature of an opponent's choice".to_string(),
-                    ),
+                Effect::Scry {
+                    count: fixed(2),
+                    target: TargetFilter::Controller,
                 },
                 source_id,
                 controller,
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
This is at the right hard-coded room table seam and the `Effect::Scry { count: 2 }` shape assertion is a useful revert probe, but I need runtime coverage before approving it.

Please add a scenario-level test that actually ventures/enters Dungeon of the Mad Mage's Lost Level through the engine and proves the room produces the real Scry behavior for the controller, not just that `room_effects()` constructs the expected enum. The old `Effect::Unimplemented` stub was a user-visible behavior bug, and dungeon rooms do not have card-data parse-diff coverage, so this needs a discriminating engine-path test before merge.

ultrahighsuper:
Added a runtime engine-path test. New `crates/engine/tests/mad_mage_lost_level_scry_runtime.rs` drives the actual venture pipeline via `GameScenario`/`GameRunner`: with the marker in Goblin Bazaar (room 2 -> [4] Lost Level) it ventures into Lost Level, the room ability goes on the stack, and `resolve_top()` resolves it — asserting the controller is now arranging the top **two** library cards (`WaitingFor::ScryChoice { cards.len() == 2 }`). Revert-probe: on the old `Effect::Unimplemented` stub resolving the room produces no scry (`waiting_for` stays `Priority`) and the test fails.

matthewevans (approved):
Reviewed current head. Lost Level is now the actual Scry 2 room effect, and the runtime test drives the venture pipeline into the room and asserts a two-card ScryChoice rather than only checking the AST.

matthewevans (approved):
Re-reviewed after updating against current main. The merge conflict only combined the new Lost Level Scry 2 test with the already-landed Lost Mine dungeon tests; the Lost Level behavior remains the same and the branch is now current for the merge queue.

matthewevans (approved):
Reviewed current head. This stays at the hard-coded dungeon room table seam, replaces the fabricated Lost Level stub with `Scry 2`, and includes a runtime venture-path test that resolves the room ability and asserts a two-card `ScryChoice`.

PR #5013, opened 2026-07-03
fix(parser): parse Volo creature-type spell qualifier (#4962)

### LINKED ISSUE #4962: [[Volo, Guide to Monsters]] — Volo doesn't give tokens of anything.
Reported in Discord: https://discord.com/channels/1485498006781427802/1522085541909364756/1522085541909364756

**Thread:** [[Volo, Guide to Monsters]]
**Cards:** _none detected_

## Summary
Volo doesn't give tokens of anything.

---
<sub>report_id: `discord:1522085541909364756:1522085541909364756:0` · discord: `1522085541909364756/1522085541909364756`</sub>

bohdansolovie:
## Summary
- Parse Volo's post-spell qualifier (`that doesn't share a creature type with a creature you control or a creature card in your graveyard`) via `parse_post_spell_modifier`
- Support disjunctive shared-quality references in `parse_shared_quality_reference`
- Add parser and integration regression coverage

Fixes #4962

## Test plan
- [x] `cargo test -p engine --lib parse_post_spell_modifier_creature_type`
- [x] `cargo test -p engine --test integration issue_4962`


matthewevans (changes requested):
Blocking this current head because the required backend gate is red: `Rust lint (fmt, clippy, parser gate)` is failing on 54a1d0934c1f98441e76cf10c99b2d52903cabd9, and there is no current-head `<!-- coverage-parse-diff -->` evidence yet.

This is parser/engine-surface work, so please fix the current-head lint/parser gate and get the parse-diff sticky posted before this can move forward.

matthewevans (changes requested):
Requesting changes on the current head. The Volo case is closer, but the parse-diff shows this parser change is not scoped to the Volo/Menagerie Curator class. In particular, Everything Comes to Dust changes from a creature/artifact/enchantment disjunctive reference into a creature-only reference, which appears to drop valid artifact/enchantment cases. Winnowing also changes its sacrifice target shape.

Please keep the disjunctive shared-quality support from narrowing or otherwise changing those sibling cards, and add or adjust coverage so the Volo fix does not regress existing shared-quality targets.

matthewevans (changes requested):
Requesting changes on the new current head. The earlier Everything Comes to Dust / Winnowing parse-diff regression is no longer showing, but the new commit removed the negative runtime coverage instead of proving Volo's condition is enforced.

**[MED]** The PR now only proves Volo copies when there is no shared creature type. It no longer proves the inverse: Volo must *not* copy when the cast creature spell shares a creature type with either a creature you control or a creature card in your graveyard. The commit message says the negative case was dropped pending runtime `Or` filter support; that is exactly the behavior this PR needs before it can land. Please either fix the runtime `Or`/shared-quality path or otherwise add discriminating runtime coverage for both disjunctive blockers.

A positive-only test would still pass if Volo copied every creature spell unconditionally, so it is not enough evidence for this bug fix.

matthewevans:
Processed the new head `d971ab9b` and holding rather than approving while required parser evidence catches up.

The previous runtime negative-coverage gap appears addressed in the current diff: the PR now has both controlled-creature and graveyard-card negative Volo integration coverage, and the stale Gemini items around the CR citation/test shape look fixed on this head. However, this is still parser/engine-surface work and the current-head `Card data (generate, validate, coverage)` check is still in progress. The existing `<!-- coverage-parse-diff -->` sticky was last updated before this head was pushed, so it is stale for `d971ab9b`.

Please let the current-head parse-diff sticky regenerate. I’m not treating CI pending as a failure, but I do need the current-head parse-diff evidence before this can be approved or enqueued.

bohdansolovie:
Thanks @matthewevans — addressed the discriminating-coverage blocker on the new head (`d971ab9b9`).

**Runtime enforcement of the `Or` shared-quality condition.** The root cause of the missing negative behavior was that a cast creature spell sits on the stack and was self-satisfying its own "creature you control" shared-type reference, so Volo copied unconditionally. `object_shares_quality_with_reference_filter` now excludes the object under test from its own reference scan **when that object is on the stack** (CR 205.3m / 109.1), while battlefield-to-battlefield comparisons keep their self-inclusive semantics (guards `shares_quality_negated_land_type_reference_matches_per_object`).

**Discriminating runtime coverage** (integration, all populate `all_creature_types`):
- `volo_does_not_copy_when_sharing_type_with_controlled_creature` — shares a type with a creature you control → no copy.
- `volo_does_not_copy_when_sharing_type_with_graveyard_card` — shares a type with a creature card in your graveyard → no copy.
- `volo_copies_creature_spell_with_no_shared_type` — made discriminating: a non-sharing creature is on the battlefield, and Volo still copies.

A positive-only test would now fail if Volo copied unconditionally, so both disjunctive blockers are covered.

matthewevans (changes requested):
**[MED]** Stack sibling spells can still satisfy Volo’s “creature you control” reference. Evidence: `crates/engine/src/game/filter.rs:4770`. Why it matters: CR 109.2 makes bare “creature you control” mean a battlefield permanent; excluding only the object under test still lets another same-type creature spell on the stack block Volo/Menagerie Curator incorrectly. Suggested fix: make shared-quality reference evaluation honor CR 109.2 zone semantics for bare type references, and add a regression with another same-type creature spell already on the stack.

bohdansolovie:
Thanks @matthewevans — addressed on `da522399d` (also merged current `main` so the parse-diff sticky regenerates on this head).

**Root cause.** `object_shares_quality_with_reference_filter` scanned every object and only excluded the object under test, so a *sibling* creature spell on the stack still matched the bare "creature you control" leg — a creature spell on the stack satisfies type + `ControllerRef::You`, but it is not a permanent. Per CR 109.2 a bare type reference is a zoned object reference (a battlefield permanent, or the explicitly named zone), never a spell on the stack.

**Fix.** The reference scan now skips stack objects (`reference_obj.zone != Zone::Stack`). Stack-scoped references (`TriggeringSource` / `ParentTarget`) are still resolved by the branches above this scan, so this only affects bare permanent/card references (Volo, Menagerie Curator, …). Battlefield- and graveyard-to-object comparisons keep their self-inclusive semantics (still guarded by `shares_quality_negated_land_type_reference_matches_per_object`).

**Regression** — `shares_quality_reference_ignores_sibling_creature_spell_on_stack` (exercises the runtime filter `matches_target_filter` → `object_shares_quality_with_reference_filter`): with two same-type creature spells on the stack, the cast spell matches "doesn't share a creature type" (the sibling on the stack does not block); moving that **same** sibling to the battlefield flips it to blocking. Same object, only the zone changes — so it fails on the old exclude-self-only behavior and passes now.

matthewevans (approved):
Approved updated head `1433176f`. The previous CR 109.2 blocker is addressed at the runtime shared-quality filter seam: bare "creature you control" / graveyard-card references no longer let sibling creature spells on the stack satisfy the reference, while the existing TriggeringSource/ParentTarget paths still handle stack/event-scoped references. The current tests are discriminating: they cover the positive Volo copy path, both battlefield and graveyard negative paths, and the sibling-same-type spell-on-stack regression that would fail if the last fix were reverted.

I updated the branch against current `origin/main` before approving; the new checks are pending from that update, and auto-merge will wait on required CI.

PR #5016, opened 2026-07-03
fix(effects): no-op tap/untap choice with zero eligible targets (#4961)

### LINKED ISSUE #4961: Game Engine Stuck — Game got stuck, attached game state and best images I could get of error it stated.
Reported in Discord: https://discord.com/channels/1485498006781427802/1522089219185709226/1522089219185709226

**Thread:** Game Engine Stuck
**Cards:** _none detected_

## Summary
Game got stuck, attached game state and best images I could get of error it stated.

---
<sub>report_id: `discord:1522089219185709226:1522089219185709226:0` · discord: `1522089219185709226/1522089219185709226`</sub>

bohdansolovie:
## Summary
- When a resolution-time tap/untap `EffectZoneChoice` has zero eligible permanents but `min_count >= 1`, resolve as a no-op instead of wedging the game
- Mirrors the Nahiri attach empty-candidate guard (#4868) for the tap/untap choice path

Fixes #4961

## Test plan
- [x] `cargo test -p engine --lib tap_untap_choice_with_no_eligible_permanents_does_not_deadlock`

### DIFF
--- a/crates/engine/src/game/ability_utils.rs
+++ b/crates/engine/src/game/ability_utils.rs
@@ -2358,7 +2358,7 @@ pub(crate) fn resolve_multi_target_bounds(
     })
 }
 
-fn multi_target_needs_quantity_choice(
+pub(crate) fn multi_target_needs_quantity_choice(
     state: &GameState,
     ability: &ResolvedAbility,
     spec: &MultiTargetSpec,
--- a/crates/engine/src/game/effects/tap_untap.rs
+++ b/crates/engine/src/game/effects/tap_untap.rs
@@ -223,29 +223,37 @@ fn prompt_resolution_tap_untap_choice(
         .copied()
         .filter(|id| crate::game::filter::matches_target_filter(state, *id, target, &ctx))
         .collect();
-    let Ok(bounds) = crate::game::ability_utils::resolve_multi_target_bounds(
+    let bounds = match crate::game::ability_utils::resolve_multi_target_bounds(
         state,
         ability,
         spec,
         eligible.len(),
-    ) else {
-        return false;
+    ) {
+        Ok(bounds) => bounds,
+        // CR 608.2b + CR 601.2c (issue #4961): `resolve_multi_target_bounds`
+        // errors when the eligible pool is smaller than the selection's required
+        // minimum. With zero eligible permanents the required tap/untap choice is
+        // vacuously impossible, so resolve it here — the single reachable
+        // no-candidate point — as a clean no-op. This keeps a resolution-time
+        // `EffectZoneChoice` from ever being built over an empty pool (which is
+        // the shape that wedges the game). A non-empty-but-under-filled pool is
+        // left to the normal target path (`return false`) so partial selections
+        // keep their existing behavior.
+        Err(_) => {
+            if eligible.is_empty() {
+                events.push(GameEvent::EffectResolved {
+                    kind: effect_kind,
+                    source_id: ability.source_id,
+                });
+                return true;
+            }
+            return false;
+        }
     };
 
     if bounds.max == 0 && bounds.min == 0 {
         events.push(GameEvent::EffectResolved {
             kind: EffectKind::from(&ability.effect),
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
The new no-op branch is unreachable for the claimed fixed-min case. `prompt_resolution_tap_untap_choice` calls `resolve_multi_target_bounds` before the added `eligible.is_empty() && bounds.min > 0` check; `resolve_multi_target_bounds` returns `Err("Not enough legal targets available")` when `legal_target_count < min`. For `MultiTargetSpec::fixed(1, 1)` with zero eligible permanents, the function returns at the `let Ok(bounds) = ... else { return false; }` path and never reaches the new no-op branch.

Please move the no-candidate handling to a reachable point or adjust the bounds helper semantics deliberately, and keep the regression proving the original deadlock path resolves as intended.

matthewevans (changes requested):
Requesting changes on the current head. The fix moved the no-candidate handling to the reachable branch, but it now catches too many errors.

**[MED]** The `Err(_)` arm no-ops whenever `eligible.is_empty()`, even if `resolve_multi_target_bounds` failed because the target count still requires a resolved quantity rather than because there are not enough legal targets. Evidence: `resolve_multi_target_bounds` first returns `ActionNotAllowed("Target count requires a resolved quantity before target selection")` for unresolved variable target counts, before it checks `legal_target_count < min`. Why it matters: a resolution-time tap/untap choice with zero current candidates but an unresolved `X`/quantity target count would be silently treated as resolved instead of preserving the unresolved-choice failure. Suggested fix: only convert the specific "not enough legal targets available" / fixed-min empty-pool case into the no-op; let unresolved-quantity errors continue to return `false` or propagate through the existing path, and add a regression for that branch.

matthewevans:
Processed the new head `fc2ce111` and holding while current-head evidence catches up.

The prior unresolved-quantity concern appears addressed in this diff: the no-candidate path now checks `multi_target_needs_quantity_choice(...)`, and there is a regression that keeps unresolved `X` from being treated as a resolved no-op. I do not see a new code blocker from that quick pass.

This head was pushed after the current `<!-- coverage-parse-diff -->` sticky, and `Rust lint`, Rust tests, and card-data/coverage are still in progress for `fc2ce111`. I’m not treating pending CI as a failure, but this is engine-surface work, so I’m holding until the current-head checks and parse evidence are available before approval/enqueue.

bohdansolovie:
Thanks @matthewevans — narrowed the `Err(_)` arm on the new head (`fc2ce1119`).

You're right that the arm no-op'd on *any* bounds-resolution failure with an empty pool, which would also swallow the earlier unresolved-quantity error (`"Target count requires a resolved quantity before target selection"`, returned before the legal-target-count check). It now re-checks the exact predicate:

```rust
Err(_) => {
    if eligible.is_empty()
        && !crate::game::ability_utils::multi_target_needs_quantity_choice(state, ability, spec)
    {
        // vacuous "not enough legal targets" empty pool → clean no-op
        ...
        return true;
    }
    return false; // unresolved-quantity (and under-filled pools) fall through
}
```

So only the "not enough legal targets" / fixed-min empty-pool case becomes a no-op; an unresolved `X`/quantity count falls through to the normal target path and preserves the unresolved-choice failure. `multi_target_needs_quantity_choice` is now `pub(crate)` for this re-check.

**New regression:** `tap_untap_choice_with_unresolved_x_count_is_not_treated_as_resolved` — a resolution-time tap choice with an unresolved `X` count and zero eligible permanents asserts the branch returns `false`, emits no `EffectResolved`, and installs no `EffectZoneChoice`. The original deadlock regression (`..._does_not_deadlock`) and the normal-prompt guard (`..._still_prompts`) still pass.

matthewevans (approved):
Approved current head `fc2ce11`. The current implementation resolves the zero-eligible-target no-op path at the tap/untap effect seam, preserves the unresolved-quantity guard, and has runtime coverage for zero candidates, missing target context, and still-valid target selection.

PR #5021, opened 2026-07-03
fix(engine): grant The Atropal deathtouch (Tomb of Annihilation "Cradle of the Death God")

ultrahighsuper:
## Problem

Tomb of Annihilation's final room, **Cradle of the Death God**, creates *The Atropal*. Its Oracle text (Scryfall):

> Cradle of the Death God â€” Create The Atropal, a legendary 4/4 black God Horror creature token **with deathtouch**.

The engine built the token with an empty keyword list, silently dropping the deathtouch:

```rust
Effect::Token { name: "The Atropal", power: 4, toughness: 4, colors: [Black], keywords: vec![], .. }
```

So the Atropal's combat damage was not treated as lethal to creatures with toughness > 1 â€” a visible gameplay difference whenever the token blocks or attacks.

## Fix

Grant `Keyword::Deathtouch` to the token.

- **CR 702.2a**: "Deathtouch is a static ability."
- **CR 702.2c**: "Any nonzero amount of combat damage assigned to a creature by a source with deathtouch is considered to be lethal damageâ€¦"

## Tests

- **AST** (`dungeon.rs::tomb_cradle_atropal_has_deathtouch`): `room_effects` for Cradle produces a 4/4 legendary Atropal `Token` whose `keywords` contain `Deathtouch`.
- **Runtime** (`tests/tomb_cradle_atropal_deathtouch_runtime.rs`): drives the real venture pipeline â€” ventures into Cradle of the Death God, resolves the room ability, and asserts the created battlefield token has deathtouch. **Fails on the old empty keyword list** (revert-probed locally: the test panics with `got keywords []` when the fix is reverted).

Scope is intentionally limited to this one token's missing keyword.

### DIFF
--- a/crates/engine/src/game/dungeon.rs
+++ b/crates/engine/src/game/dungeon.rs
@@ -460,7 +460,8 @@ pub fn room_effects(
             (lose, vec![])
         }
         // 4: Cradle of the Death God — "Create The Atropal, a legendary 4/4 black God Horror
-        //    creature token"
+        //    creature token with deathtouch."
+        // CR 702.2a: Deathtouch is a static ability granted to the token.
         (DungeonId::TombOfAnnihilation, 4) => (
             simple(
                 Effect::Token {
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Reviewed current head. The change stays at the existing Tomb room token-definition seam, adds the missing Deathtouch keyword to The Atropal token, and includes a runtime venture-path regression that resolves Cradle of the Death God and would fail on the old empty keyword list.

PR #5027, opened 2026-07-03
ENGINE: Scope counter replacements through type phrases

davidomil:
## Summary

Fix counter replacement parsing so effects like Mauhur, Uruk-hai Captain scope modified counters to the affected Army/Goblin/Orc instead of applying broadly to any counter event from the replacement source. The parser now derives object-only valid-card scope from the existing type-phrase parser, covering artifact-or-creature and comma-separated type/subtype lists without expanding actor-scoped opponent replacement semantics.

- [ ] Produced via the `/engine-implementer` pipeline (plan -> review-plan -> implement -> review-impl -> commit)

the pipeline):

> Initial code was produced directly in a Codex debugging session before `/engine-implementer` was invoked, so I am not claiming the full pipeline authored the commit. Follow-up validation used fresh-context `/engine-planner`, `/review-engine-plan`, and final `/review-impl` passes against the diff. The plan review caught the actor-scoped `opponent would put` boundary, and the final implementation review reported zero findings against the updated diff plus parse-diff evidence.

> [!NOTE]

## CR references

- CR 614.1a / CR 614.16: replacement effects, including effects that modify counters put on permanents.
- CR 122.1a: +1/+1 counters modify power/toughness.
- CR 701.47a: amass chooses an Army creature and puts the counters on that creature.

## Verification

- `cargo fmt --all --check` — passed.
- `./scripts/check-parser-combinators.sh` — passed.
- Parser string-dispatch grep from the `/engine-implementer` executor instructions — passed, no added parser `.contains`/`.starts_with`/`.find`/`.split*` dispatch.
- CR diff grep against `docs/MagicCompRules.txt` — passed, no unverified added/changed CR annotations.
- `CARGO_INCREMENTAL=0 cargo test -p engine --lib counter_plus_one_replacement --jobs 2` — passed: 3 tests.
- `CARGO_INCREMENTAL=0 cargo test -p engine --lib counter_doubling_replacement_current_oracle_wording --jobs 2` — passed: 1 test.
- `CARGO_INCREMENTAL=0 cargo test -p engine --lib parses_halving_season_opponent_counter_replacement --jobs 2` — passed: 1 test.
- `CARGO_INCREMENTAL=0 cargo test -p engine --test integration mauhur --jobs 2` — passed: 2 tests.
- `./scripts/gen-card-data.sh` — passed; generated coverage data for parse-diff evidence.
- `cargo run --profile tool --features cli --bin coverage-parse-diff -- /tmp/pr5027-coverage-base.json client/public/coverage-data.json --base-sha 910a4385355244500e31d2bf5d25a9f0377befba --markdown /tmp/pr5027-parse-diff.md --json /tmp/pr5027-parse-diff.json` — passed; parse-diff reports 0 card(s), 0 signature(s).

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -3,7 +3,7 @@ use std::str::FromStr;
 use crate::parser::oracle_nom::error::{oracle_err, OracleError, OracleResult};
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take_until};
-use nom::character::complete::{char, multispace1};
+use nom::character::complete::{char, multispace0, multispace1};
 use nom::combinator::{all_consuming, eof, map_opt, opt, peek, rest, value};
 use nom::multi::separated_list1;
 use nom::sequence::{pair, preceded, terminated};
@@ -7021,14 +7021,8 @@ fn parse_counter_replacement(lower: &str, original_text: &str) -> Option<Replace
     let mut def = ReplacementDefinition::new(ReplacementEvent::AddCounter)
         .quantity_modification(modification)
         .description(original_text.to_string());
-    if nom_primitives::scan_contains(lower, "permanent you control") {
-        def = def.valid_card(TargetFilter::Typed(
-            TypedFilter::permanent().controller(ControllerRef::You),
-        ));
-    } else if nom_primitives::scan_contains(lower, "creature you control") {
-        def = def.valid_card(TargetFilter::Typed(
-            TypedFilter::creature().controller(ControllerRef::You),
-        ));
+    if let Some(valid_card) = parse_counter_replacement_valid_card(lower) {
+        def = def.valid_card(valid_card);
     }
     if nom_primitives::scan_contains(lower, "an opponent would put")
         || nom_primitives::scan_contains(lower, "opponent would put")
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[MED]** `is_counter_replacement_scope_tail` reintroduces ad hoc parser dispatch with string-prefix checks. Evidence: `crates/engine/src/parser/oracle_replacement.rs:7064`. Why it matters: parser additions in this repo are required to compose nom combinators from the first pass, and this helper is a parsing boundary that should accept the tail through `tag`/`alt` rather than hand-checking prefixes. Suggested fix: replace the prefix array/`starts_with` helper with a small nom combinator over the accepted tails, as Gemini suggested, and keep it as the single tail parser used by `parse_counter_replacement_valid_card`.

**[MED]** The new integration test cites the wrong CR rule for amass. Evidence: `crates/engine/tests/integration/mauhur_swarming_of_moria.rs:3`. Why it matters: `docs/MagicCompRules.txt` shows `CR 701.44a` is explore, while amass is `CR 701.47a`; incorrect CR annotations create false rules evidence in engine tests. Suggested fix: update the comment to `CR 701.47a` and keep the `CR 122.1a` counter interaction if you want both rules documented.

Also, this is draft parser/engine-surface work and there is no current `<!-- coverage-parse-diff -->` sticky on the PR. After the code issues are fixed and the PR is ready for review, please make sure the parse-diff evidence is posted so the card-level blast radius can be checked.

davidomil:

matthewevans:
Reviewed current head `6158175` and did not find an implementation blocker. I am holding this one because it is still marked draft, so I am not approving or enqueueing it from the sweep. Current-head CI artifact showed no card-parse changes; once it is ready for review, it should be eligible for the normal handler path unless the head changes.

davidomil:
Marked ready for review. Current head `6158175` addresses the earlier review items:

- `is_counter_replacement_scope_tail` now uses nom `tag`/`alt` parsing instead of prefix checks, and the parser gate passed in CI.
- The Mauhur integration test CR reference is corrected to `CR 701.47a`.
- The `<!-- coverage-parse-diff -->` sticky is present and current; CI parse-detail diff reports no card-parse changes.

All current-head CI checks are green.

matthewevans (approved):
Approved current head `6158175ea26ec180ef9923bade55da888713b199`.

The previous parser-combinator finding is fixed at the shared replacement parser seam, the runtime test drives the real counter replacement path through `add_counter_with_replacement`/scenario resolution, and the current CI set is green including Rust lint/parser gate, Rust tests, and card-data. The parse-diff sticky reports no card-parse changes for this head, which matches the scoped runtime parser/replacement fix.

PR #5029, opened 2026-07-03
fix(engine): Tomb of Annihilation "Veils of Fear" is a life-loss punisher, not a creature sacrifice

ultrahighsuper:
## Problem

Tomb of Annihilation's **Veils of Fear** room has this Oracle text (Scryfall):

> Veils of Fear â€” Each player loses 2 life unless they discard a card.

The engine implemented a completely different effect â€” a mandatory creature sacrifice:

```rust
(DungeonId::TombOfAnnihilation, 1) => {
    let mut ability = simple(Effect::Sacrifice { target: creature, count: 1, min_count: 0 }, ..);
    ability.player_scope = Some(PlayerFilter::All);
    (ability, vec![])
}
```

So venturing into Veils of Fear never cost any life and instead demanded a creature â€” the wrong keyword action entirely.

## Fix

Build the room from its Oracle text via the shared parser (the same `parse_effect_chain` + `build_resolved_from_def` path several other rooms already use, e.g. Throne of the Dead Three). The parser yields the correct punisher shape:

- base effect `LoseLife { amount: 2 }`
- `player_scope: All` (each player)
- `unless_pay: { cost: Discard { count: 1 }, payer: ScopedPlayer }`

**CR 118.12a**: "Some spells, activated abilities, and triggered abilities read, '[Do something] unless [a player does something else].' This means the same thing as '[A player may do something else]. If [that player doesn't], [do something].'"

## Test

- **Runtime** (`tests/tomb_veils_of_fear_life_loss_runtime.rs`): drives the real venture pipeline â€” ventures Trapped Entry â†’ Veils of Fear with both players holding no cards, declines each player's discard-or-lose-life prompt (`PayUnlessCost { pay: false }`), and asserts both players lose 2 life. **Fails on the old sacrifice room** (revert-probed locally: with the old code both players stay at 20 life â€” no life loss and no unless-payment prompt).

Scope is limited to Veils of Fear.

### DIFF
--- a/crates/engine/src/game/dungeon.rs
+++ b/crates/engine/src/game/dungeon.rs
@@ -417,19 +417,17 @@ pub fn room_effects(
             ability.player_scope = Some(PlayerFilter::All);
             (ability, vec![])
         }
-        // 1: Veils of Fear — "Each player sacrifices a creature"
+        // 1: Veils of Fear — "Each player loses 2 life unless they discard a card."
+        // CR 118.12a: a punisher — each player either discards a card or loses
+        // 2 life. Parsed from the Oracle text so the base life loss, the
+        // each-player scope, and the unless-discard payment all come from the
+        // shared punisher parser (the payer binds to the per-iteration
+        // ScopedPlayer). The room was wrongly implemented as "each player
+        // sacrifices a creature".
         (DungeonId::TombOfAnnihilation, 1) => {
-            let mut ability = simple(
-                Effect::Sacrifice {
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved current head 0ce08913. Local review: the fix is at the dungeon room-effect seam, reuses the existing `parse_effect_chain`/`build_resolved_from_def` punisher path instead of hand-rolling a new room effect, and the runtime test drives the real venture branch into Veils of Fear and discriminates the old sacrifice implementation. CR 118.12a was verified against docs/MagicCompRules.txt.

PR #5031, opened 2026-07-03
fix(parser): split "then any number of target players each ..." after a compound (Singularity Rupture)

### LINKED ISSUE #4783: Singularity Rupture bug — The game doesn't prompt you to select targets when casting Singularity Rupture, so no player…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521843799817457725/1521843799817457725

**Thread:** Singularity Rupture bug
**Cards:** singularity rupture

## Summary
The game doesn't prompt you to select targets when casting Singularity Rupture, so no player can ever mill cards.

### singularity rupture
Destroy all creatures, then any number of target players each mill half their library, rounded down.

---
<sub>report_id: `discord:1521843799817457725:1521843799817457725:0` · discord: `1521843799817457725/1521843799817457725`</sub>

nickmopen:
Closes #4783.

## Problem

**Singularity Rupture** — "Destroy all creatures, then any number of target players each mill half their library, rounded down." — dropped its **entire second clause**: only `DestroyAll` parsed, so the mill never happened and (as reported) no target prompt appeared.

## Cause

The `, then <clause>` boundary detector (`oracle_effect/sequence.rs`) recognized imperative-verb continuations, "`<subject>` you control …" predicates, and damage clauses — but **not a player-*subject*-led clause**. "any number of target players each mill …" begins with a *subject*, not a verb, so `starts_clause_text_or_conjugated` missed it and the tail was silently dropped.

Notably, the second clause parses correctly **on its own** (`parse_effect("Any number of target players each mill half their library, rounded down.")` → `Mill { count: DivideRounded(library ÷ 2, Down), target: Player }`); it was only lost in the `then`-continuation position.

## Fix

Add a targeted `starts_target_players_subject_clause` recognizer for the distributive "any number of target player(s) [each] …" form and include it in the `Then`-boundary check. Scoped to that exact subject so it can't over-split (a "then deal damage to any number of target players" tail still matches on the leading verb, unaffected).

Result — Singularity Rupture now chains:
- primary `DestroyAll` creatures
- sub `Mill { DivideRounded(TargetZoneCardCount(Library) ÷ 2, Down), target: Player }`

## Tests

New `any_number_of_target_players_each_splits_after_then` (the `, then` split) and `starts_target_players_subject_clause_recognizes_distributive` (recognizer unit, incl. negative cases).

`cargo test -p engine --lib` → **14,676 passed / 0 failed**; `fmt --check`, `clippy -p engine --all-targets -- -D warnings`, and `check-parser-combinators.sh` all clean.

CR 115.1d + CR 608.2c.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -1240,6 +1240,7 @@ fn split_comma_clause_boundary(current: &str, remainder: &str) -> Option<(Clause
         if starts_clause_text_or_conjugated(after_then)
             || starts_you_control_subject_predicate(after_then_lower)
             || starts_with_damage_clause(after_then_lower)
+            || starts_target_players_subject_clause(after_then_lower)
         {
             return Some((ClauseBoundary::Then, whitespace_len + "then ".len()));
         }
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved current head `f1c3d9d`. The sequence parser change is scoped to the Singularity Rupture compound subject split, current-head parse-diff artifact reports exactly the intended one-card parse gain, and the implementation reuses the existing sequence/parser building blocks instead of adding a one-off card branch.

PR #5033, opened 2026-07-03
fix: require authentication for administrative /admin HTTP endpoints

RealDiligent:
## Summary

- **Root cause:** The `/admin/drafts` endpoints (`admin_list_drafts`, `admin_get_draft`, `admin_delete_draft`) were mounted with **no authentication**. In the documented deployment the reverse proxy (see `deploy/phase-server.nginx.conf`) forwards `/` to `phase-server`, so these routes are reachable from the public internet. Any anonymous caller could enumerate every draft, read a draft's internal state, and - via `DELETE /admin/drafts/{code}` - force-delete a draft **and evict its in-progress game sessions**, a destructive availability/integrity attack.
- **Fix:** Gate the `/admin/*` routes behind a bearer token supplied at runtime via `PHASE_ADMIN_TOKEN` (or `--admin-token`). Requests must present `Authorization: Bearer <token>`, compared in constant time. When no token is configured the admin routes are **not mounted at all** (404) - fail closed rather than expose them unauthenticated. The client-facing `/p2p-draft-backup` routes are unchanged (part of the normal P2P draft flow). The token is runtime-only config and never enters the shared `AppState`.
- **Impact:** Closes an unauthenticated destructive/info-disclosure surface with no behavior change for gameplay or the P2P backup flow. Operators who want admin access set `PHASE_ADMIN_TOKEN` and send the bearer header.

## Test plan

- [x] `tokens_match_is_exact` - constant-time compare correctness (incl. length mismatch)
- [x] `authorized_only_with_matching_bearer_token` - accepts correct `Bearer` token (whitespace tolerant)
- [x] `rejects_missing_wrong_or_malformed_header` - rejects missing/empty/wrong-token/wrong-scheme/bare-token
- [ ] `cargo test -p phase-server admin_auth` and `cargo fmt --check` (local; no Rust toolchain in this agent env)

## Risks / tradeoffs

- **Behavior change (intentional):** admin endpoints now require a token. Deployments that previously curled `/admin/drafts` without auth must set `PHASE_ADMIN_TOKEN` and pass the bearer header. Failing closed by default is the safer posture given public reachability.
- Scope is limited to route mounting/auth in `phase-server`; handler logic is untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Docker uses environment variables for the common options:
 | `PHASE_LOBBY_ONLY` | `--lobby-only` | false | Matchmaking broker mode |
 | `PHASE_LOG_JSON` | `--log-json` | false | Emit JSON logs |
 | `PHASE_LOG_DIR` | `--log-dir` | stdout | Write logs to files |
+| `PHASE_ADMIN_TOKEN` | `--admin-token` | unset | Bearer token for `/admin/*` endpoints. Unset ⇒ admin routes disabled (404). Send as `Authorization: Bearer <token>`. |
 
 You can also pass server flags after the image name:
 
--- a/crates/phase-server/src/main.rs
+++ b/crates/phase-server/src/main.rs
@@ -10,8 +10,9 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use axum::extract::ws::{Message, WebSocket};
-use axum::extract::{State, WebSocketUpgrade};
-use axum::response::IntoResponse;
+use axum::extract::{Request, State, WebSocketUpgrade};
+use axum::middleware::{from_fn_with_state, Next};
+use axum::response::{IntoResponse, Response};
 use axum::routing::{get, post};
 use axum::Router;
 use clap::Parser;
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
**[MED]** Admin secret is exposed as a command-line flag. Evidence: `crates/phase-server/src/main.rs:510`. Why it matters: bearer tokens passed via `--admin-token` can leak through process listings, shell history, and deployment logs; `README.md:166` also advertises the flag. Suggested fix: make admin token env/secret-file only, with no public CLI argument.

**[LOW]** Bearer scheme matching is case-sensitive. Evidence: `crates/phase-server/src/main.rs:1298`. Why it matters: HTTP auth scheme names are case-insensitive, so valid `bearer <token>` or mixed-case clients are rejected despite presenting the correct token. Suggested fix: split scheme/credentials and compare the scheme with `eq_ignore_ascii_case("Bearer")`; add lowercase/mixed-case tests.

**[MED]** The critical auth fix is tested only at helper level, not at the router boundary. Evidence: `crates/phase-server/src/main.rs:1331`. Why it matters: the security behavior depends on conditional route mounting and middleware wiring at `crates/phase-server/src/main.rs:1161`, but the added tests would still pass if `/admin/drafts` were accidentally left unmounted or mounted without the guard. Suggested fix: add Axum route tests for no token -> 404, configured token with missing/wrong auth -> 401, and correct bearer token reaching an admin route.

matthewevans:

PR #5034, opened 2026-07-03
fix(engine): Tomb of Annihilation "Sandfall Cell" is a life-loss punisher, not a fabricated Zombie token

ultrahighsuper:
## Problem

Tomb of Annihilation's **Sandfall Cell** room has this Oracle text (Scryfall):

> Sandfall Cell â€” Each player loses 2 life unless they sacrifice a creature, artifact, or land of their choice.

The engine implemented a fabricated effect â€” a life loss for the controller plus a **Zombie token that does not exist on the card**:

```rust
(DungeonId::TombOfAnnihilation, 3) => {
    let mut lose = simple(Effect::LoseLife { amount: 2, target: None }, ..); // controller only
    lose.sub_ability = Some(Box::new(simple(
        creature_token("Zombie", 2, 2, &["Zombie"], &[ManaColor::Black], &[], 1), .. // not on the card
    )));
    (lose, vec![])
}
```

So venturing into Sandfall Cell made only the venturing player lose life, created a bogus 2/2 Zombie, and never offered the sacrifice alternative to any player.

## Fix

Build the room from its Oracle text via the shared parser (the same `parse_effect_chain` + `build_resolved_from_def` path other rooms use). The parser yields the correct punisher shape (verified against the resolved ability):

- base effect `LoseLife { amount: 2 }`
- `player_scope: All` (each player)
- `unless_pay: { cost: Sacrifice one permanent matching creature/artifact/land you control (Or-filter, count 1), payer: ScopedPlayer }`

**CR 118.12a**: "Some spells, activated abilities, and triggered abilities read, '[Do something] unless [a player does something else].'"

## Test

- **Runtime** (`tests/tomb_sandfall_cell_life_loss_runtime.rs`): drives the real venture pipeline â€” ventures Veils of Fear â†’ Sandfall Cell with both players controlling no permanents, declines each player's sacrifice-or-lose-life prompt (`PayUnlessCost { pay: false }`), and asserts both players lose 2 life. **Fails on the old room** (revert-probed locally: the non-controller player stays at 20, since the old code only made the controller lose life and created a bogus token).

Scope is limited to Sandfall Cell.

### DIFF
--- a/crates/engine/src/game/dungeon.rs
+++ b/crates/engine/src/game/dungeon.rs
@@ -461,19 +461,20 @@ pub fn room_effects(
             .sub_ability(sac_creature);
             (discard, vec![])
         }
-        // 3: Sandfall Cell — "You lose 2 life and create a 2/2 black Zombie creature token"
+        // 3: Sandfall Cell — "Each player loses 2 life unless they sacrifice a
+        //    creature, artifact, or land of their choice."
+        // CR 118.12a: a punisher — each player either sacrifices a creature,
+        // artifact, or land, or loses 2 life. Parsed from the Oracle text so the
+        // base life loss, the each-player scope, and the unless-sacrifice
+        // payment all come from the shared punisher parser (the payer binds to
+        // the per-iteration ScopedPlayer). The room was wrongly implemented as
+        // "you lose 2 life and create a 2/2 black Zombie token" — a fabricated
+        // effect that appears nowhere on the card.
         (DungeonId::TombOfAnnihilation, 3) => {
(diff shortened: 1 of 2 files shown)


matthewevans:

matthewevans (approved):
Approved current head `c9694c54`. The implementation uses the existing dungeon room `parse_effect_chain` pattern for the actual Sandfall Cell Oracle text, the CR 118.12a citation checks out, and the new runtime integration test drives the real venture path and discriminates the prior fabricated-Zombie/controller-only behavior.

PR #5035, opened 2026-07-03
fix: redact RNG seed from client-visible game state (predictable shuffles/draws)

RealDiligent:
## Summary

- **Root cause:** `GameState.rng_seed` is a **serialized** field, and the engine's randomness is a deterministic `ChaCha20Rng` seeded from it. Every `StateUpdate` / `GameStarted` message sends the filtered `GameState` to clients, but `filter_state_for_viewer` never redacted `rng_seed` - so the seed was broadcast to every player and spectator.
- **Impact:** With the seed, a client can reconstruct the RNG stream and **predict every future shuffle, draw, coin flip, and random selection**, including its own and the opponent's hidden library order. This defeats the library/hand redaction the very same filter performs (`filter_state_for_viewer` hides card identities) and breaks ranked/competitive integrity - a hidden-information leak.
- **Fix:** Redact `rng_seed` (and reset the serde-skipped `rng` handle) in `filter_state_for_viewer`. The authoritative engine and on-disk persistence operate on the **unfiltered** state (`to_persisted` / `save_session` serialize `session.state` directly), so server-side randomness and session restore are unaffected - only the client-facing copy loses the seed.

## Why this is safe

- `filter_state_for_viewer` output is only used for **transport/display**: server ? client broadcasts, P2P host ? guest, and the WASM `get_filtered_game_state` / `get_viewer_snapshot_js` JS projections. The authoritative Rust `GameState` (with the real seed) is never replaced by a filtered copy, and persistence never routes through the filter.
- All client-bound `GameState` flows through this single chokepoint (per-player `filter_state_for_player` and the `SPECTATOR_PLAYER_ID` view), so one redaction covers every path.

## Test plan

- [x] `redacts_rng_seed_from_every_viewer` - seed is zeroed for both seats and the non-seat spectator; source state is untouched
- [ ] `cargo test -p engine visibility` and `cargo fmt --check` (local; no Rust toolchain in this agent env)

## Risks / tradeoffs

- Clients that deserialize a filtered snapshot reconstruct their local RNG from seed `0`; this is display-only and never authoritative, so there is no gameplay effect.
- No change to persistence, AI search (uses its own RNG), or the loop-detection fingerprint (computed on authoritative state).

### DIFF
--- a/crates/engine/src/game/visibility.rs
+++ b/crates/engine/src/game/visibility.rs
@@ -16,6 +16,21 @@ pub fn filter_state_for_viewer(state: &GameState, viewer: PlayerId) -> GameState
     let mut filtered = state.clone();
     filtered.pending_begin_game_abilities.clear();
     filtered.resolving_begin_game_abilities = false;
+
+    // Hidden-information + fairness integrity: the game's RNG is a deterministic
+    // ChaCha20 stream seeded from `rng_seed`, and that seed is a serialized field
+    // of `GameState`. Broadcasting it to clients (every `StateUpdate` /
+    // `GameStarted` carries the filtered `GameState`) would let any player
+    // reconstruct the stream and predict every future shuffle, draw, coin flip,
+    // and random selection — including their own and the opponent's hidden
+    // library order, defeating the library redaction below and breaking ranked
+    // integrity. The authoritative engine and on-disk persistence operate on the
+    // UNFILTERED state, so redacting the seed here (and resetting the skipped RNG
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved current head `73e07bc`. The fix is at the shared engine visibility seam, redacts the serialized RNG seed before client-facing state leaves the engine, and the touched server/WASM/P2P paths already route through the filtered viewer state. The regression test checks both seated players and a spectator while confirming the authoritative source state is unchanged.

PR #5037, opened 2026-07-03
fix: filter library-draw game events before client broadcast (hidden card leak)

RealDiligent:
## Summary

- **Root cause:** Every `StateUpdate` / `GameStarted` message sends a parallel `events` array alongside the filtered `GameState`. That array was broadcast **verbatim** to every seat and spectator. `GameEvent::CardDrawn` exposes the specific `object_id` of the drawn card, and `GameEvent::ZoneChanged` from `Library` embeds the full card identity in `ZoneChangeRecord` (name, types, mana value, etc.).
- **Impact:** Opponents could read the raw event JSON and learn **every card drawn from library**, defeating the library/hand redaction in `filter_state_for_viewer`. The structured game log already excluded these events; the wire `events` channel did not.
- **Fix:** Add `filter_events_for_viewer` and apply it at every server broadcast chokepoint (per-player `StateUpdate`/`GameStarted` and spectator updates). Library?hand moves and individual `CardDrawn` events are visible only to the drawing player; public moves (mill, cast to stack/battlefield, etc.) remain broadcast to all.

## Test plan

- [x] `filters_library_draw_events_for_non_drawer` - opponent receives no draw events
- [x] `library_mill_zone_change_stays_public` - library?graveyard stays visible
- [ ] `cargo test -p engine visibility` and `cargo test -p phase-server` (no Rust toolchain in agent env)

## Risks / tradeoffs

- Clients that animate draws from `CardDrawn` will only receive those events for their own draws; opponents still see aggregate `CardsDrawn` when emitted.
- Face-down exile / tutor-to-library edge cases use the conservative owner-visibility rule (same as state filtering).


matthewevans (changes requested):
Requesting changes. This is the right area to fix, but the current filter still has a hidden-information hole in the same class of event stream.

**[HIGH]** `Library -> Battlefield` events still leak face-down manifest/cloak identities to opponents and spectators. Evidence: `crates/engine/src/game/visibility.rs:799` treats every library-to-battlefield `ZoneChanged` as public; `crates/engine/src/game/morph.rs:362` moves manifest/cloak cards from library to battlefield face down; and `crates/engine/src/game/game_object.rs:1194` snapshots the real card name/types into `ZoneChangeRecord`. Why it matters: opponents must not learn the real identity of a face-down battlefield object, so the broadcast event channel still leaks the card this PR is trying to hide. Please make `filter_events_for_viewer` decide `ZoneChanged` visibility from the moved object's post-move visibility/face-down permissions, or at least redact/deny library-to-battlefield records when that object is face down for the viewer. Add opponent and spectator tests for manifest/cloak from library.

**[MED]** The destination-zone allowlist also hides public library-to-exile reveals. Evidence: `crates/engine/src/game/visibility.rs:801` only allows library moves to Battlefield, Stack, and Graveyard; cascade moves library cards to Exile in `crates/engine/src/game/effects/cascade.rs:69`; and top-card exile effects move cards to Exile before separately marking face-down cases in `crates/engine/src/game/effects/exile_top.rs:87`. Why it matters: public exile/reveal flows can disappear from non-controller clients, while face-down exile still needs redaction. The filter needs object/face-down visibility semantics, not a fixed destination-zone allowlist.

matthewevans (approved):
Reviewed updated head after branch update; this clears the maintainer bar for merge-when-ready.

matthewevans (approved):
Reviewed current head after maintainer turn-control event-filter fix; clears the maintainer bar for merge-when-ready.

PR #5038, opened 2026-07-03
fix: enforce draft lobby password on join (password bypass)

RealDiligent:
## Summary

- **Root cause:** `JoinDraftWithPassword` only ran wire-shape validation (`guard_join_draft_with_password`). `DraftSessionManager::join_draft` took `_password` and never checked it, so anyone with the 6-character draft code could claim a seat on a password-protected pod.
- **Impact:** Private draft lobbies offered no access control - the password field was cosmetic. Regular game joins already call `Lobby::verify_password`; drafts did not.
- **Fix:** Verify the password against `session.lobby_meta` before seat assignment (same match rules as the lobby broker). Map `password_required` to `PasswordRequired` on the wire. Store `lobby_meta` (including password) when a draft is created, and re-register restored drafts in the lobby broker on startup so protection survives persistence.

## Test plan

- [x] `join_draft_enforces_lobby_password` - rejects missing/wrong password, accepts correct
- [ ] `cargo test -p server-core draft_session` (no Rust toolchain in agent env)

## Risks / tradeoffs

- Drafts created before this change have no persisted `lobby_meta`; they remain joinable without a password until recreated. New/restored drafts with `lobby_meta` are protected.


matthewevans (changes requested):
Requesting changes. The join-time password check is at the right seam, but the restore behavior added here can expose drafts that are no longer joinable.

**[HIGH]** Restored started drafts are re-registered in the public lobby whenever `lobby_meta` exists. Evidence: `crates/phase-server/src/main.rs:5357` stores `session.lobby_meta` at draft creation, `crates/phase-server/src/main.rs:5638` persists the draft after later draft actions including `StartDraft`, and `crates/phase-server/src/main.rs:841` restores every persisted draft with `lobby_meta` back into the lobby without checking `DraftStatus::Lobby`; `crates/server-core/src/draft_session.rs:221` then lets `join_draft` claim any open seat without checking the draft status. Why it matters: after a server restart, a password-protected draft that has already started can reappear as a joinable lobby entry, and a user with the password can be seated into a non-lobby draft. Suggested fix: clear `lobby_meta` / unregister the lobby entry when a draft starts, or gate restore/join on `DraftStatus::Lobby`; add a persistence/restore test for a started draft with an open seat.

matthewevans:
PRs like this require a recording showing that you've tested this.

matthewevans (changes requested):
Requesting changes on current head `09014c8733505858d5bf27852fde22dc38f36aa8` as well.

The only delta from the previously reviewed head is formatting in `crates/phase-server/src/main.rs` and `crates/server-core/src/draft_session.rs`; it does not address the blocker. Restored drafts with `lobby_meta` are still re-registered into the lobby unconditionally at restore time, and `join_draft` still does not gate joins on `DraftStatus::Lobby`. Please fix the restore/join status issue and add the persistence/restore test for a started draft with an open seat.

matthewevans:
Current head `57d8d1dff1583b4323252911f1e2af83a3a3130d` addresses the blocker I raised earlier: `join_draft` now rejects non-`Lobby` drafts, `StartDraft` clears `lobby_meta`, and restore only re-registers drafts whose persisted status is still `Lobby`.

I’m holding approval rather than enqueueing this head because the restore regression is still only pinned through `PersistedDraftSession::should_register_in_lobby()`, while the bug was in the startup restore/register wiring. Please add a small production-seam test/helper that feeds both `Lobby` and `Drafting` persisted drafts through the same restore-registration path and asserts only the lobby-status draft is registered.

RealDiligent:
Added restored_draft_lobby_register_request as the shared restore-registration seam with a wiring test covering Lobby vs Drafting (open seats). Recording: [link]

matthewevans (approved):
Reviewed current head after password/restore fixes; prior blocker is addressed and the security seam has coverage. Approving for merge-when-ready.

matthewevans:

RealDiligent:
Already Addressed the HIGH review item: config.rng_seed is now zeroed in redact_persisted_draft_secrets(), with tests confirming to_admin_snapshot() redacts it while to_persisted() preserves the real seed for disk restore.

matthewevans:
Closing per maintainer direction in the current review workflow. This was not an author-abandoned close; please do not reopen without explicit maintainer approval.

matthewevans (changes requested):
Reopening this so we can preserve the security fix on the existing PR path, but taking it back out of ready-to-merge for the current head `2c5d5d1f30a7674c008c54305b4093cb282cc276`.

The functional review blocker I raised earlier is addressed, and the live test boundary now looks good: both Rust test shards, card-data, frontend, WASM, Tauri, and lobby-worker checks completed successfully. The remaining blocker is formatting: `Rust lint (fmt, clippy, parser gate)` fails on `cargo fmt --all -- --check`.

CI annotations point to formatter diffs in:

- `crates/server-core/src/draft_session.rs:972`
- `crates/server-core/src/lib.rs:51`
- `crates/server-core/src/persist.rs:90`

I reproduced this locally with `cargo fmt --all`; it is a formatter-only patch. Maintainer edits are disabled on this fork branch, so please either run `cargo fmt --all` and push the result, or enable maintainer edits if you want us to apply the minimal formatter patch here. I am not re-enabling auto-merge until the live checks pass on the updated head.

matthewevans:
Current-head review on `7c609ea1e0c104a4f4bc9ae92b654faf3ac6503e`: the formatter-only blocker from the prior review is fixed, and I do not see a remaining code-review blocker in the lobby password enforcement or restore registration path.

I am keeping this held until the live checks finish on this exact head. I am not enabling merge/auto-merge while checks are still unresolved.

matthewevans:
Current live state on `7c609ea1e0c104a4f4bc9ae92b654faf3ac6503e`: the formatter-only blocker is fixed and current CI is green, but GitHub now reports the branch as `DIRTY` / `CONFLICTING` against `main`.

I’m keeping this held rather than approving/enqueueing while it has textual conflicts. Please rebase or merge current `main` and resolve the conflicts; once the branch is conflict-free, this can be rechecked against the same security boundary and live checks.

matthewevans:
I merged current `main` into this PR branch and resolved the text conflict in `crates/server-core/src/lib.rs` by keeping both the new P2P backup exports from `main` and this PR's `restored_draft_lobby_register_request` export.

Current head is `31e21b28cfbd44264fb1e604ab9880a3707bec16`. GitHub now reports the PR as mergeable rather than conflicting, and checks are pending on the updated head. I’m keeping this held until those live checks complete and the current head is re-reviewed for approval.

matthewevans (approved):
Maintainer review: current head is conflict-free after the mechanical main merge, live checks are green, and the security boundary is covered by the draft-password join test. Enqueuing via merge queue.

PR #5040, opened 2026-07-03
Add Sylvan Echoes

e11734937-beep:
## What
Parses **Sylvan Echoes** — "Whenever you clash and win, you may draw a card."

## How
"Whenever you clash and win, ..." compacts a clash trigger and its win-outcome gate into one clause. The bare `Clashed` recognizer whole-clause-anchors on `"you clash"` and rejects the trailing `" and win"`, and there is no trigger-level `EventOutcomeWon` condition. The supported clash siblings (Entangling Trap, Rebellion of the Flamekin) instead encode a clash win as a separate `"If you won, ..."` effect sentence that lifts to `AbilityCondition::EventOutcomeWon` on the gated ability.

`reshape_clash_and_win_trigger` re-homes the compacted outcome the same way: strip `" and win"` from the trigger clause and prepend `"If you won, "` to the effect, so the existing `Clashed` recognizer plus the effect-condition pipeline produce the identical sibling AST — a `Clashed` trigger whose `Draw` effect is gated by `EventOutcomeWon` (optional). CR 701.19 / 701.19d.

## Scope / safety
The reshape is a no-op for any trigger that is not exactly `"you clash and win"`, so the only affected card is Sylvan Echoes — the sole clash-and-win card in the corpus. Tests: a parser-seam structural test asserting the reshaped AST, plus a regression test that plain `"whenever you clash"` siblings keep their head effect ungated.

- rustfmt, parser-combinator gate, clippy `-D warnings`: clean
- `cargo test -p engine --lib`: 14811 passed, 0 failed
- Coverage: Sylvan Echoes now supported; +1 net, no card regressed

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -8210,16 +8210,18 @@ fn scoped_player_matches_filter(
 
 fn event_outcome_was_won_by_controller(event: &GameEvent, controller: PlayerId) -> bool {
     match event {
+        // CR 701.30d: reuse the shared controller-relative clash-outcome
+        // resolution so resolution-time "if you won" gating and trigger-matching
+        // (`match_clash`) agree on who won.
         GameEvent::Clash {
             controller: clash_controller,
             opponent,
             result,
             ..
-        } => match result {
-            crate::types::events::ClashResult::Won => *clash_controller == controller,
-            crate::types::events::ClashResult::Lost => *opponent == controller,
-            crate::types::events::ClashResult::Tied => false,
-        },
+        } => {
+            result.for_player(*clash_controller, *opponent, controller)
+                == Some(crate::types::events::ClashResult::Won)
+        }
         GameEvent::CoinFlipped { player_id, won } => *player_id == controller && *won,
         _ => false,
     }
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -3725,23 +3725,48 @@ pub(super) fn match_ring_tempts_you(
 /// Fires when a clash occurs and either clashing player matches `valid_target`.
 /// "Whenever you clash" sets `valid_target = Controller`; a generic "whenever
 /// a player clashes" leaves `valid_target` unset to match any clash.
+///
+/// CR 701.30d + CR 603.4: when the trigger carries a required clash outcome
+/// (`clash_result`, set for "...and win" cards like Sylvan Echoes), the win
+/// requirement is checked HERE, when the event occurs — so a lost or tied clash
+/// never creates a pending (no-op) trigger. The outcome is resolved from the
+/// ability's controller's perspective via `ClashResult::for_player`, the same
+/// source of truth used by resolution-time "if you won" gating
+/// (`event_outcome_was_won_by_controller`), so matching and resolution agree.
 pub(super) fn match_clash(
     event: &GameEvent,
     trigger: &TriggerDefinition,
     source_id: ObjectId,
     state: &GameState,
 ) -> bool {
-    match event {
-        GameEvent::Clash {
-            controller,
-            opponent,
-            ..
-        } => {
-            valid_player_matches(trigger, state, *controller, source_id)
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
Requesting changes.

**[HIGH]** `Whenever you clash and win` is modeled as a resolution-time effect gate, so lost/tied clashes still create a triggered ability. Evidence: `crates/engine/src/parser/oracle_trigger.rs:6286` rewrites the trigger into a normal `Clashed` trigger with an `EventOutcomeWon` ability condition, but `crates/engine/src/game/trigger_matchers.rs:3735` matches every `GameEvent::Clash` for either participant and `crates/engine/src/types/events.rs:865` already carries the clash result payload. CR 603.4 says intervening-if trigger conditions are checked when the event occurs, and CR 701.30d defines when a player wins a clash. Why it matters: Sylvan Echoes should not trigger at all unless the clash was won, but the current shape puts a no-op trigger on the stack for losses/ties. Suggested fix: carry the win requirement into trigger matching using `GameEvent::Clash.result`, and add an engine test that lost/tied clashes create no pending trigger.

**[MED]** The new CR citations point to Regenerate instead of Clash. Evidence: `crates/engine/src/parser/oracle_trigger.rs:1007`, `crates/engine/src/parser/oracle_trigger.rs:6239`, and `crates/engine/src/parser/oracle_trigger_snapshot_tests.rs:867` cite CR 701.19/701.19d for clash; `docs/MagicCompRules.txt:3428` shows CR 701.19 is Regenerate, while `docs/MagicCompRules.txt:3558` shows Clash is CR 701.30 and `docs/MagicCompRules.txt:3566` defines winning at CR 701.30d. Why it matters: the project requires verified CR evidence, and wrong CR citations create false confidence around rules behavior. Suggested fix: replace the new CR 701.19/701.19d references with verified CR 701.30/701.30d citations.

e11734937-beep:
Thanks for the detailed review — both points addressed.

**[HIGH] Win requirement moved into trigger MATCHING (was resolution-time).** Added `TriggerDefinition.clash_result: Option<ClashResult>` — the clash sibling of `coin_flip_result`/`die_result`. The parser sets `Some(Won)` for the "…and win" tail (Sylvan Echoes) and `None` for plain "you clash" (Entangling Trap unchanged). `match_clash` now checks the outcome when the clash event occurs (CR 603.4 intervening-if; CR 701.30d), so a lost or tied clash never creates a pending (no-op) trigger. The check goes through a new shared `ClashResult::for_player`, and I refactored the resolution-time `event_outcome_was_won_by_controller` onto that same helper so matching and resolution can never disagree on who won. Dropped the previous text-reshape/effect-gate approach entirely. Added the engine test you asked for: lost and tied clashes create no pending trigger; a won clash does.

**[MED] CR citations fixed** to CR 701.30 / 701.30d (clash and winning), replacing the incorrect 701.19 references.

Verification: rustfmt, parser-combinator gate, and clippy `-D warnings` clean; `cargo test -p engine --lib` 14827 passing; card-data regen + coverage confirm Sylvan Echoes parses as `Clashed` with `clash_result: Won`, no other card changed.

matthewevans:
Holding current head `03c97c432b383571c91529e62fbc5e9032dfdad9` for the required parser/card-data evidence. This PR changes parser/engine files and the card-data check is still pending, with no parse-diff sticky available yet; I will review after that evidence is available.

matthewevans (approved):
Reviewed current head `03c97c432b383571c91529e62fbc5e9032dfdad9`.

The prior blocker is addressed: `clash_result: Some(Won)` now gates the `Clashed` trigger in `match_clash`, using `ClashResult::for_player` so the win is evaluated from the ability controller's perspective when the clash event occurs rather than as a resolution-time no-op gate. The plain `Whenever you clash` shape remains `clash_result: None`, so Entangling Trap-style siblings still fire on any clash outcome while their separate `If you won` tail stays resolution-gated.

Parse-diff evidence is scoped to Sylvan Echoes only: the inert `Whenever you clash and win` unsupported signature is removed and replaced by a typed `Clashed` trigger. The new matcher and parser snapshot tests cover controller/opponent seating, win/loss/tie outcomes, and the no-regression plain-clash case.

PR #5043, opened 2026-07-03
feat(parser): parse Monomania "choose a card in hand, discard the rest"

e11734937-beep:
## Summary

**Monomania** ("Target player chooses a card in their hand and discards the rest.") previously failed to parse — the `choose … and discards the rest` phrasing dispatched to the `choose` verb with no handler, lowering to `Effect::Unimplemented { name: "choose" }`.

This adds a whole-clause-anchored recognizer that routes the line to the **existing** `Effect::Discard`. No new engine behavior, effect, or static is introduced — only text recognition.

## Semantics

The chosen player keeps exactly one card of their choice and discards every other card in hand — i.e. discards `hand size − 1` cards, chooser-directed:

- count = `ClampMin { Offset { HandSize, −1 }, 0 }` (the codebase's canonical "N − 1, floored at 0" idiom) so a one-card or empty hand discards nothing (CR 107.1b).
- `selection: Chosen` — the discarding player chooses (CR 701.9a).
- `target` left `Controller` so `inject_subject_target` binds the discarder to the spell's **target player**, not the caster (same pattern as the sibling "discard their hand" path).

## Robustness

The recognizer is `eof`-anchored to the full clause. A regression test asserts it rejects near-misses that share the `choose`/`the rest` tokens: a bare `choose a card in their hand` (no disposal), a `discard it` back-reference, a plain `choose a creature` target, a `two cards` cardinality mismatch, and the Balance-family `choose a number of … they control … discards the rest` equalization clause.

## Verification

- `cargo fmt --check` — clean
- `scripts/check-parser-combinators.sh` — pass
- `cargo clippy -p engine --all-targets -- -D warnings` — 0 warnings
- `cargo test -p engine --lib` — 14813 passed; 0 failed; 6 ignored (4 new tests included; 0 regressions)

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -3148,6 +3148,56 @@ fn parse_hand_possessive_target(input: &str) -> nom::IResult<&str, TargetFilter,
     .parse(input)
 }
 
+/// CR 701.9a + CR 107.1b: "[Target player] chooses a card in their hand and
+/// discards the rest" (Monomania). The chosen player keeps exactly one card of
+/// their choice and discards every other card in hand -- equivalently, discards
+/// `hand size - 1` cards, chooser-directed. Routes to the existing
+/// `Effect::Discard` (via `TargetedImperativeAst::Discard`) using the canonical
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Reviewed current head and parse-diff evidence; this clears the maintainer bar for merge-when-ready.

PR #5044, opened 2026-07-03
fix(parser): resolve "choose two creatures, then those creatures fight each other" (Joust, Blizzard Brawl, Tail Swipe) — CR 701.14a

dhgoal:
Tier: Frontier
Model: claude-opus-4-8
Thinking: high

## Summary
CR 701.14a: the "**Choose target creature you control and target creature you don't control. [rider.] Then those creatures fight each other.**" spell class — **Joust, Blizzard Brawl, Tail Swipe** — did not resolve its fight. The generic clause splitter's bare-`and` continuation probe (`starts_target_continuous_clause_lower`, a `take_until(" gets ")` that crosses the sentence boundary and false-matches the rider's verb) stranded **fighter B** as a disconnected `Effect::unimplemented("target creature you don't control")` and left the `Fight` node with an **empty target slot** — so at runtime the fight never happened *and* casting demanded a spurious third target.

**Parser-only fix — the engine already resolves the correct shape.** A new `fight_each_other` recognizer catches the whole frame **before** chunk splitting (fast-rejected by the closing phrase via `scan_contains_phrase`; the nom recognizer is the sole authority) and lowers it to the proven dual-target `Fight` shape (Prey Upon / Epic Confrontation / Ulvenwald Tracker): a single `Fight` node carrying both fighters in its **own** two target slots, read by `resolve_fight_fighters`. When a rider is present it binds to **fighter A** exactly as Epic Confrontation does — the rider becomes the primary effect (its target slot *is* fighter A) and the `Fight` becomes an **unconditional `SequentialSibling` sub** (`subject: ParentTarget` = fighter A, `target` = fighter B), so the fight resolves even when the rider's condition (Knight / snow permanents / main-phase) is false.

Self-sourced via `cargo parser-gaps`. Coverage-honest throughout: no false-greens.

## Files changed
- crates/engine/src/parser/oracle_effect/fight_each_other.rs (new — recognizer + rider→fighter-A binding + parser tests)
- crates/engine/src/parser/oracle_effect/mod.rs (module decl + pre-split interception)
- crates/engine/src/game/effects/fight.rs (3 runtime cast tests only)

## CR references
- `CR 701.14a` — Fight: a creature fights another creature (covers "two creatures fight each other").
- `CR 601.2c` — each instance of "target" opens its own target slot.
- `CR 120` — combat/fight damage.

## Gate A
```
[exit 0 — no violations]
```
Also ran the authoritative check against upstream — `git diff origin/main -- 'crates/engine/src/parser/**' | grep '^+' | grep -E '<FORBIDDEN_METHODS>'` — which printed **nothing**. The recognizer is combinator-only (`tag`/`alt`/`take_until`/`map` + the shared `parse_target`); the cheap byte fast-reject uses the codebase's `scan_contains_phrase` helper, not raw `.contains`.

## Anchored on
- crates/engine/src/parser/oracle_effect/mod.rs:1464 — the existing `scan_contains`-gated fast-reject-then-specialized-recognizer idiom (a structural pre-filter before the generic path); the new interception mirrors it exactly (`scan_contains_phrase(lower, FIGHT_EACH_OTHER_MARKER)` → nom recognizer authority).
- crates/engine/src/game/effects/fight.rs:69 — `resolve_fight_fighters`, the runtime contract (a `Fight` node with ≥2 object targets picks `(targets[0], targets[1])` as the two fighters) that the new dual-target lowering targets; the Epic Confrontation rider→`Pump`-primary + `Fight`-`SequentialSibling`-sub shape is mirrored so the fight resolves unconditionally.

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — exit 0; authoritative `git diff origin/main` combinator grep — **empty**
- `cargo test -p engine --lib -- parser::` — **7260 pass / 0 fail**
- `cargo test -p engine --lib -- fight` — **38 pass / 0 fail** (Ulvenwald Tracker, Blood Feud, Band Together, Prey Upon, Epic Confrontation all still pass)
- **Runtime cast tests** (the discriminating fail-on-revert proof, in fight.rs): Joust — pump binds to fighter A (2/9→4/10), B takes A's boosted power 4, A takes 3, resolves with exactly **2 targets**; Blizzard Brawl snow<3 — no buff but fight **still resolves** (SequentialSibling escape); snow≥3 — +1/+0 binds to A, A survives lethal via granted indestructible.
- Regenerated card-data → Joust/Blizzard Brawl/Tail Swipe carry **0** `Unimplemented` + 1 connected Fight each.

## Scope Expansion
The recognizer is class-general (build-for-the-class), so it also connects a now-resolving `Fight` for the **+1/+1-counter-rider siblings** — **Hog-Monkey Rampage, Malamet Battle Glyph, Longstalk Brawl** — which were likewise broken. Their counter riders are `PutCounter` (not `Pump`/`GenericEffect`), so `bind_rider_to_fighter_a` can't bind them; each keeps exactly **one honest `Unimplemented("fight rider", …)`** while the fight itself resolves — never a false-green. Extending the rider binder to `PutCounter` is a clean follow-up. Unrelated fight cards (Arena, Rivals' Duel, Guild Feud) are untouched. New parser module `fight_each_other.rs` added as a reusable building block for this frame.

### DIFF
--- a/crates/engine/src/game/effects/fight.rs
+++ b/crates/engine/src/game/effects/fight.rs
@@ -987,6 +987,190 @@ mod tests {
         );
     }
 
+    /// CR 701.14a + CR 601.2c: End-to-end cast of Joust ("Choose target creature
+    /// you control and target creature you don't control. The creature you control
+    /// gets +2/+1 until end of turn if it's a Knight. Then those creatures fight
+    /// each other."). The whole frame must lower to a single connected dual-target
+    /// fight: exactly TWO targets are announced (no spurious third slot), the pump
+    /// binds to fighter A and applies BEFORE the fight (fighter B takes A's BOOSTED
+    /// power), and both fighters deal each other's power. Discriminates the false-
+    /// green: the old parse left fighter B a disconnected `Unimplemented`, so the
+    /// fight never resolved and a third target was demanded.
+    #[test]
+    fn joust_full_cast_resolves_dual_target_fight_with_pump_on_fighter_a() {
+        use crate::game::scenario::{GameScenario, P0, P1};
+        use crate::types::game_state::WaitingFor;
+        use crate::types::mana::{ManaType, ManaUnit};
+        use crate::types::phase::Phase;
+        const JOUST: &str = "Choose target creature you control and target creature you don't control. The creature you control gets +2/+1 until end of turn if it's a Knight. Then those creatures fight each other. (Each deals damage equal to its power to the other.)";
+
+        let mut scenario = GameScenario::new();
+        scenario.at_phase(Phase::PreCombatMain);
+        // Fighter A is a Knight so the rider condition holds; large toughness so
+        // both fighters survive and their marked damage stays observable.
+        let mine = scenario
+            .add_creature(P0, "My Knight", 2, 9)
+            .with_subtypes(vec!["Knight"])
+            .id();
+        let theirs = scenario.add_creature(P1, "Their Beast", 3, 9).id();
+        let spell = scenario
+            .add_spell_to_hand_from_oracle(P0, "Joust", true, JOUST)
+            .id();
+        scenario.with_mana_pool(
+            P0,
+            vec![ManaUnit::new(ManaType::Colorless, ObjectId(9_999), false, vec![]); 6],
+        );
+        let mut runner = scenario.build();
+        {
+            let state = runner.state_mut();
+            state.active_player = P0;
+            state.priority_player = P0;
+        }
+
+        // CR 601.2c: exactly two targets — fighter A then fighter B.
+        let outcome = runner.cast(spell).target_objects(&[mine, theirs]).resolve();
+
+        // CR 611.2c: the +2/+1 pump bound to fighter A and applied before the fight.
+        assert_eq!(
+            outcome.power_toughness(mine),
+            (4, 10),
+            "pump must bind to fighter A (2/9 + 2/1 = 4/10)"
+        );
+        // CR 701.14a: fighter B takes fighter A's BOOSTED power (4), proving the
+        // pump applied before the fight and the fight actually resolved.
+        assert_eq!(
+            outcome.damage_marked(theirs),
+            4,
+            "fighter B takes fighter A's boosted power (4)"
+        );
+        // CR 701.14a: fighter A takes fighter B's power (3).
+        assert_eq!(
+            outcome.damage_marked(mine),
+            3,
+            "fighter A takes fighter B's power (3)"
+        );
+        // No spurious third target: the spell resolved cleanly back to priority.
+        assert!(
+            matches!(outcome.final_waiting_for(), WaitingFor::Priority { .. }),
+            "cast needed exactly two targets; got {:?}",
+            outcome.final_waiting_for()
+        );
+    }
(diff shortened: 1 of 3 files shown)


matthewevans (commented):
Current-head review of `20737f68c3`: I don't see a concrete code blocker in the parser/effect shape from this pass. The change is still parser/engine surface, and the `coverage-parse-diff` sticky is not present for this head yet.

Holding approval/enqueue until the parse-diff posts so the card-level blast radius is visible. No approval/enqueue from the sweep yet.

matthewevans (approved):
Maintainer sign-off: reviewed current head `89a12e0dada252b93c4a5208ee759df00183802c`. The parser hook is scoped to the "choose two creatures / fight each other" frame, parse-diff blast radius matches the claimed class, CR citations were checked, and runtime coverage exercises the target binding, rider conditions, and fight resolution. Auto-merge can wait on required checks.

matthewevans (changes requested):
**[MED]** The new counter-rider binding still does not make the Malamet-style counter land on fighter A. Evidence: `crates/engine/src/parser/oracle_effect/fight_each_other.rs` lines 170-189 claims to rebind the `PutCounter` rider for "the creature you control" onto fighter A, but current CI for head `9176f9083f9192cfb83c4f655c2594fec89352fd` fails `malamet_counter_on_you_control_when_entered_this_turn_and_both_fight` at `crates/engine/tests/s07_malamet_castpath.rs:135` with `mine_c` still `0` instead of `1`. Why it matters: this is the exact sibling-rider class the commit says it fixes, so re-approving would false-green Malamet / Longstalk / Hog-Monkey style fight-and-counter cards while the runtime behavior is still wrong. Suggested fix: bind the actual parsed target/condition authority for the rider to fighter A and keep the existing cast-path test green.

There is also a stale red-guard test in the same run: `tail_swipe_class_guards_do_not_touch_it_and_stays_red` still expects Tail Swipe to retain an `Unimplemented` node at lines 352-358. If this PR intentionally promotes Tail Swipe to supported, update that guard to assert the new intended supported shape; otherwise keep Tail Swipe out of this parser class.

dhgoal:
Closing this — the "those creatures fight each other" class is already handled on `main` by a cleaner, integrated approach (`parse_fight_target` in `oracle_target.rs` emitting `TargetFilter::ParentTarget` for the reciprocal "each other", plus the `s07_malamet_castpath` cast-path coverage for Malamet / Longstalk / Duel for Dominance and siblings). That work landed while this PR was in review.

My standalone `fight_each_other` recognizer duplicates that mechanism and, against current `main`, is a net-regression: it strands Malamet/Longstalk's runtime counter and changes Tail Swipe's shape against the deliberate "stays RED" guard. The only card it uniquely improves is Joust (`unimpl 1 → 0`), which isn't worth a parallel, conflicting implementation of a class `main` already owns.

If Joust's "the creature you control gets +2/+1 … if it's a Knight" rider is worth closing, that's better done as a small extension of the existing `main` handler rather than this recognizer. Thanks for the review — closing so it doesn't sit as a conflicting duplicate.

PR #5052, opened 2026-07-03
Add Flame Spill

Spaceint:
## Summary
Adds engine support for **Flame Spill**, and with it the broader **CR 120.4a** "Excess damage is dealt to that creature's controller instead" rider class (Gandalf's Sanction, Ravenous Tyrannosaurus). The trailing rider is *absorbed* onto the preceding `DealDamage` effect (it previously fell out as a stray `Effect::Unimplemented`); at resolution the excess — damage past lethal, already computed once per CR 120.6/120.10 in the shared damage chokepoint — is redirected to the damaged permanent's controller.

Ram Through's *conditional* form ("If the creature you control has trample, excess …") is intentionally deferred to `Effect::Unimplemented` (honest red coverage) — modeling the controlled-source trample gate is out of scope for this change.

## Files changed
- crates/engine/src/types/ability.rs — new `ExcessRecipient` enum; `excess: Option<ExcessRecipient>` on `Effect::DealDamage`; `excess_recipient` on `DamageContextSnapshot`
- crates/engine/src/game/effects/deal_damage.rs — thread `excess_recipient` through `DamageContext` (+ its 4 constructors); in the shared `apply_damage_after_replacement` chokepoint, deal only the **lethal** portion to the creature and redirect the **excess** to its live controller *instead* (CR 120.4a — total dealt is unchanged), inert for combat
- crates/engine/src/types/game_state.rs + game/engine_replacement.rs + game/replacement.rs — carry `excess_recipient` on `PendingReplacement` so the damage-replacement **choice** resume path preserves the rider (the snapshot resume path is covered by `DamageContextSnapshot`); mirrors the existing `library_placement` carry-through
- crates/engine/src/parser/oracle_ir/ast.rs — `ContinuationAst::ExcessDamageToController`
- crates/engine/src/parser/oracle_effect/sequence.rs — recognize the rider (nom `scan_contains`, with `!has trample`/`!if ` guards) and absorb it onto the preceding `DealDamage`
- crates/engine/src/game/ability_scan.rs, parser/oracle_effect/{lower,imperative,mod}.rs — thread the new field
- Mechanical `excess: None` / `excess: _` at every other `Effect::DealDamage` construction/destructure across the workspace (phase-ai, server-core, engine-wasm, and mtgish compile-fixes — matching how the existing `damage_source` field is threaded; no logic added to mtgish)
- crates/engine/tests/flame_spill_excess_damage.rs — cast-pipeline + shape + snapshot tests (T1–T9)

## CR references
- CR 120.4a — excess damage redirected to another permanent/player
- CR 120.6 — lethal damage (marked ≥ toughness)
- CR 120.7 — the redirected damage keeps the same source
- CR 120.10 — multi-type greatest-calculated-amount excess (reused, not re-derived)
- CR 702.2c — deathtouch lethal (inherited by the existing excess computation)

## Track
Non-developer

## LLM
Model: claude-opus-4-8
Thinking: high
Tier: Frontier

## Verification
Local verification is **partial by necessity**: `client/public/card-data.json` cannot be generated on this Windows machine — the MTGJSON per-set fetch in `scripts/gen-card-data.sh` fails under Git Bash with `curl: (3) URL rejected: Malformed input to a URL function` — so `cargo coverage`, `cargo semantic-audit`, and the card-data-dependent integration suite could not run locally and are left to CI. What ran clean locally:
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` (Gate A) — exit 0, no violations (see ## Gate A)
- `cargo test -p engine --test flame_spill_excess_damage` — **7 passed / 0 failed** (T1 revert-proof redirect; T2 multi-authority; T3 marked-damage; T4 exact-lethal zero-excess; T5 overkill; T8 shape + Ram Through deferral; T9 snapshot carry)
- `cargo test -p engine` — full suite green, **0 failures** (regression check for the shared damage chokepoint change)
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` — **exit 0**, clean (covers the mechanical threading in phase-ai, server-core, engine-wasm, mtgish)

## Gate A
```
$ ./scripts/check-parser-combinators.sh ; echo "exit=$?"
exit=0
```
The script scans the diff for non-combinator dispatch (`.contains("…")`, `.split_once`, match-arm string literals, chained `if let Ok = tag(...)`, verbatim Oracle `==`, hand-built `Effect::Unimplemented`) and prints nothing on success — exit 0 = no violations. The new rider recognizer uses `nom_primitives::scan_contains`, not `str::contains`.

## Anchored on
- crates/engine/src/types/ability.rs:8296 — existing `damage_source: Option<DamageSource>` typed-Option field on `Effect::DealDamage` (`#[serde(default, skip_serializing_if = "Option::is_none")]`). The new `excess: Option<ExcessRecipient>` field mirrors it exactly — a typed enum, never a bool.
- crates/engine/src/parser/oracle_effect/sequence.rs:5690 — existing `ContinuationAst::CantRegenerate` trailing-rider recognizer using `nom_primitives::scan_contains` in `parse_followup_continuation_ast`. The new excess-rider recognizer follows the same nom-combinator family and dispatch shape.
- crates/engine/src/parser/oracle_effect/sequence.rs:3040 and :3982 — the `CantRegenerate` `apply_clause_continuation` patch arm and the `continuation_absorbs_current => true` arm. The new `ExcessDamageToController` absorption mirrors both (backward-walk to the nearest `DealDamage`, patch the typed field).

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.


matthewevans (changes requested):
**[MED]** Lifelink is counted twice for redirected excess damage. Evidence: `crates/engine/src/game/effects/deal_damage.rs:698`. Why it matters: Flame Spill-style damage with lifelink gains life for the full `actual_amount` before the excess redirect, then the redirected excess re-enters `apply_damage_to_target` with `has_lifelink` still set via the copied context at `crates/engine/src/game/effects/deal_damage.rs:725`, so the controller gains for the excess a second time and can over-gain if that redirected leg is prevented. Suggested fix: apply lifelink per actually dealt leg, using the creature leg's `primary_amount` and letting the redirected player leg gain only for damage it actually deals.

**[MED]** A second replacement-choice pause drops the excess redirect rider. Evidence: `crates/engine/src/game/engine_replacement.rs:732`. Why it matters: when `pipeline_loop` parks another material replacement choice, it creates a fresh `PendingReplacement` with `excess_recipient: None` at `crates/engine/src/game/replacement.rs:6436`, and the resume branch restores only `library_placement`, so a Flame Spill damage event with multiple replacement ordering choices can lose the controller redirect before final delivery. Suggested fix: preserve and reapply the parked `excess_recipient` in the second `NeedsChoice` branch the same way this code already preserves `library_placement`.

**[LOW]** Required parse-diff evidence is still missing for this parser/engine PR. Evidence: `crates/engine/src/parser/oracle_effect/sequence.rs` is changed, but there is no current `<!-- coverage-parse-diff -->` sticky comment for head `9554b61b35fbb7ee2c6117ef864bbd172eeb38c8`. Why it matters: this PR has broad parser/engine/AI surface area, and we need the card-level gained/lost/changed parse diff before accepting the blast radius. Suggested fix: let CI finish and reconcile the parse-diff comment for the current head before this can be reconsidered.

Spaceint:
Thanks for the review, @matthewevans — all three addressed. Also merged current `main` (37 commits) to clear the CI compile failures (new `Effect::DealDamage` construction sites needed the `excess: None` threading; WASM/lint/test now green locally).

**[MED] Lifelink counted twice for redirected excess** — Fixed. The primary leg now gains lifelink for the lethal `primary_amount` it actually deals to the creature (`deal_damage.rs`), and the redirected excess leg gains for the excess it deals to the controller separately (CR 120.7, same source), so the two legs sum to exactly `actual_amount` — and if the redirected leg is prevented, no life is gained for the excess. Added runtime test `t11_lifelink_counts_each_leg_once_not_twice` (a lifelink source dealing 4 to a 2-toughness creature → controller gains 4, not 6).

**[MED] Second replacement-choice pause drops the rider** — Fixed. `handle_replacement_choice` now reapplies the captured `parked_excess_recipient` to the re-parked `PendingReplacement` on the nested `NeedsChoice` branch, exactly mirroring the existing `parked_library_placement` reapplication a few lines above. So a damage event with multiple replacement-ordering choices keeps the controller redirect through to final delivery.

**[LOW] Parse-diff evidence** — CI has now posted the `<!-- coverage-parse-diff -->` comment: 4 cards gain `ability/excess` (Flame Spill, Gandalf's Sanction, Pigment Storm, +1), 1 signature, no unexpected blast radius.

Pushing the fixes now; happy to iterate further.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (changes requested):
Current-head review on `8a67b41e4d4d062b8b87e054851012430c3ea96f`:

The prior blockers are partly addressed: the non-pausing lifelink double-count is fixed, `excess_recipient` is restored when a replacement choice re-parks, and the parse-diff sticky is present. I still cannot approve this head because there are two remaining correctness holes in the excess redirect path:

1. Redirected primary damage still records the creature as having been dealt excess damage. In `crates/engine/src/game/effects/deal_damage.rs`, the primary creature leg records `amount: primary_amount` but still carries the computed `excess`. Other engine paths treat `DamageRecord.excess > 0` as “was dealt excess damage”, so Flame Spill-style redirects can falsely satisfy Maarika/Rith/Aegar-style excess-damage triggers even though the creature only received lethal damage and the excess was dealt to its controller instead. When the excess is redirected, the primary creature record/event should not expose that redirected amount as excess dealt to the creature.

2. A primary-leg lifelink replacement-choice pause can skip the redirected excess. The current code returns `NeedsChoice` from the primary lifelink life-gain replacement path before reaching the redirect block. The resume path applies only the parked life-gain event, so lifelink Flame Spill plus a life-gain replacement/choice can resolve the primary damage and never deal the excess to the creature’s controller. Please park/model the redirect continuation before returning, and add a lifelink + life-gain replacement-choice regression that proves the excess still resolves after the prompt.

Spaceint:
Thanks @matthewevans — both remaining holes fixed and covered by regressions.

**1. Redirected primary damage no longer records the creature as dealt excess.** The creature's `DamageDealt` event and `DamageRecord` now report `primary_excess = excess - redirect_excess` (zero when the rider redirects), so `DamageRecord.excess` on the creature is 0 and Maarika/Rith/Aegar-style "was dealt excess damage" triggers no longer fire on a creature that only took lethal — the excess is recorded against the controller by the redirect leg instead. Plain overkill (no rider) still reports its excess normally. New test `t12_redirected_creature_records_no_excess_damage`.

**2. The excess redirect now runs before the lifelink gain, so a life-gain replacement-choice pause can't skip it.** I moved the redirect block ahead of the primary lifelink block in `apply_damage_after_replacement`, so the excess is dealt to the controller *before* the lifelink `apply_life_gain` can return `NeedsChoice`. New test `t13_lifelink_life_gain_replacement_choice_still_redirects_excess`: P0 controls two "gain twice that much life" doublers, so the lifelink gain surfaces a CR 616.1 ordering choice and the resolution pauses — the test asserts P1 has already lost the redirected excess (2) while P0 has gained 0 and the resolution is still halted on the pending choice (non-vacuous reach-guard).

Local: Gate A exit 0, `cargo clippy --workspace --all-targets -D warnings` clean, full `cargo test -p engine` 0 failures, WASM check clean, the excess-damage matrix now 11/11. Pushing now.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (changes requested):
Thanks for the follow-up. The creature-excess record issue looks fixed: `primary_excess` is now used for both `DamageDealt` and `DamageRecord`, and `t12_redirected_creature_records_no_excess_damage` covers that boundary.

I still need changes before this can land. The redirect now runs before the primary lifelink gain, but the redirected damage uses `redirect_ctx` with the same lifelink flag. If the redirected player-damage leg hits a life-gain replacement choice, `apply_damage_to_target` returns `NeedsChoice` and this path returns immediately from the redirect block before the later `primary_amount` lifelink block runs. I do not see a continuation being parked for that skipped primary lifelink gain; the outer `DealDamage` loop only stashes later targets. Current T13 stops at the first replacement prompt and checks that the redirect happened, but it does not choose through the prompt and assert that P0 eventually receives both lifelink portions.

Please make this sequence continuation-safe, or otherwise model the two lifelink gains so a replacement pause on the redirected leg cannot drop the primary leg. Extend T13 to resolve through the replacement choice and assert the final lifelink total.

Spaceint:
Good catch on the converse — fixed by collapsing the two lifelink gains into one.

CR 120.4a modifies a *single* damage event (it doesn't create a second event), so lifelink should gain once for the total that event actually dealt. I changed `apply_damage_after_replacement` accordingly:

- The redirected leg now runs with `has_lifelink: false`, so it no longer performs its own lifelink gain — a life-gain replacement choice on the redirected leg can no longer split or drop any lifelink portion.
- The redirect returns how much it actually dealt (`redirect_dealt`, reduced by any prevention), and lifelink is gained **once, as the very last step**, for `primary_amount + redirect_dealt`. Being last, a life-gain replacement pause here can't skip anything (both damage legs are already applied), and a prevented redirect gains nothing for the excess.

`t13` now **drives through** the CR 616.1 ordering choice (P0 controls two "gain twice that much life" replacements) instead of stopping at it, and asserts the full combined total: P0 gains `4 → 8 → 16` (both doublers applied to the single 4-life gain) and P1 takes the redirected 2 — proving the primary leg's lifelink is not dropped by the replacement pause.

One narrower residual I want to flag rather than hide: if the redirected *damage itself* pauses on a life-loss/prevention replacement **choice** on the controller (distinct from the life-gain case you flagged), the combined lifelink is deferred with it. Happy to park a proper continuation for that too if you'd like it in this PR, or split it out.

Local: Gate A exit 0, `cargo clippy --workspace --all-targets -D warnings` clean, full `cargo test -p engine` 0 failures, WASM check clean, excess matrix 11/11. Pushing now.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (changes requested):
Thanks for the quick update. The same class of bug is still present on a different pause path.

The combined lifelink gain can still be lost when the redirected excess damage itself returns `NeedsChoice`. In that path, the redirect returns before the final combined lifelink gain runs, and the replacement resume rebuilds context from the source rather than preserving the pending combined-gain state. Please park an explicit continuation for the final combined lifelink gain before returning from the redirected damage `NeedsChoice`, and add a regression where the redirected damage replacement choice resolves through and asserts the final lifelink total.

Also, the current parse-diff sticky still shows unrelated kicked-card `trigger/SpellCast` watch changes in addition to the excess-damage class. Please rebase/update so those are in base, or explicitly justify that parser diff before approval.

Spaceint:
Both done.

**Parked continuation for the combined lifelink.** Instead of gaining the combined lifelink in a trailing block (which the redirect-damage `NeedsChoice` returned before), the redirect leg now carries the deferred amount and gains it itself: `redirect_ctx.lifelink_bonus = <creature lethal>`, and the leg's lifelink gains `its own dealt + lifelink_bonus`. That leg's lifelink runs *inside* `apply_damage_after_replacement`, so it fires whether the redirect resolves inline OR after the redirected damage's own replacement pause+resume. To survive the resume (which rebuilds the ctx from the source), `lifelink_bonus` is threaded through both `DamageContextSnapshot` and `PendingReplacement` and restored in `handle_replacement_choice` — exactly like the `excess_recipient` carry, and the same deferred-then-gain pattern combat trample+lifelink uses in `combat_damage.rs`. Because the leg gains for the amount it *actually* deals, a prevented/partially-prevented redirect reduces the lifelink correctly.

**Regression `t14`.** P1 controls an optional "prevent the next 1 damage to you" shield, so the redirected excess damage to P1 pauses on a replacement choice. Driving *through* it (the shield prevents 1 of the 2), P1 loses 1 and P0 gains the combined lifelink for the damage actually dealt — creature lethal 2 + redirect 1 = **3**. A dropped continuation would give P0 only +1 (the redirect leg's own portion), so the test discriminates. `t13` still covers the life-gain-replacement pause, and `t9` now also asserts `lifelink_bonus` survives the snapshot round-trip.

**Parse-diff hygiene.** Merged current `main` (`f582ccb`), so the kicked-card `trigger/SpellCast` changes are now in base and out of this diff; the sticky should show only the excess-damage class on the next run.

Local: Gate A exit 0, `cargo clippy --workspace --all-targets -D warnings` clean, full `cargo test -p engine` 0 failures, WASM clean, excess matrix 12/12. Pushing now.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (changes requested):
Current-head review on `49e0ec8d160f475a2e427d1044ac151af69d87ea`:

**[MED]** Fully prevented redirected excess drops the primary lifelink gain. Evidence: `crates/engine/src/game/effects/deal_damage.rs:751` sets `redirected = true` before the redirected player leg resolves, `crates/engine/src/game/effects/deal_damage.rs:764` treats every `DamageResult::Applied(_)` as success without checking whether the redirect actually dealt damage, and then `crates/engine/src/game/effects/deal_damage.rs:785` skips the creature leg's lifelink because `redirected` is true. If the redirected excess is fully prevented or gated, `apply_damage_to_target` returns `Applied(0)` before the redirect leg reaches its own `lifelink_bonus` gain path, so the source gains no life even though it did deal the lethal `primary_amount` to the creature. Suggested fix: make the redirect handoff account for the redirected leg's actual dealt amount, and when that amount is zero still apply/preserve the lifelink for the primary creature damage that was actually dealt. Please add a regression where the redirected excess is fully prevented and the lifelink source still gains the lethal creature damage amount.

**[MED]** The new T14 regression drives the wrong replacement branch for its asserted behavior. Evidence: `crates/engine/tests/flame_spill_excess_damage.rs:353` says index 1 declines the optional prevention shield, and `crates/engine/src/types/game_state.rs:7569` / `crates/engine/src/game/replacement.rs:6589` define optional replacement choice `index 0 = accept, index 1 = decline`, but `crates/engine/tests/flame_spill_excess_damage.rs:376` asserts the shield prevented 1 damage. That makes the test lock in the current optional-decline behavior instead of proving the accepted prevention pause path. Suggested fix: choose the accept branch when asserting prevention, or change the expectations to the no-prevention decline result; for this PR's requested coverage, accepting the shield is the relevant path.

Spaceint:
Both fixed.

**Fully-prevented redirect no longer drops the primary lifelink.** The redirect leg gains the combined lifelink only when it actually reaches its lifelink path by dealing damage. A fully prevented / gated / phased redirect returns `Applied(0)` from the prevention/early-return paths *before* that gain, so the match now treats `Applied(0)` as "not redirected" (`redirected = false`) and the creature leg gains lifelink for the lethal `primary_amount` it did deal. New test `t15`: P1 has a mandatory "prevent the next 2 damage to you" shield, so the redirected 2 is fully prevented — P1 loses 0 and P0 still gains 2 for the lethal creature damage (before the fix P0 gained nothing).

**`t14` now drives the accept branch.** It accepts the optional prevention shield (`index 0`), which prevents 1 of the 2 redirected damage; P1 loses 1 and P0 gains the combined lifelink for the damage actually dealt — creature lethal 2 + redirect 1 = 3 — after resuming through the replacement choice. A dropped continuation would give P0 only +1.

Local: Gate A exit 0, `cargo clippy --workspace --all-targets -D warnings` clean, full `cargo test -p engine` 0 failures, WASM clean, excess matrix 13/13. Pushing now.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

matthewevans (approved):
Reviewed current head `3aa54dd423381298b3e4e3c6d2a504b59e82ef45`. The previous excess-damage/lifelink blockers are addressed: replacement resume preserves the redirect rider and deferred lifelink bonus, the fully-prevented redirect path now preserves primary lifelink, and the added runtime tests cover the relevant real cast/resolve boundaries. Approving for merge queue; remaining in-progress checks can gate auto-merge.

Issue #2198, opened 2026-06-03
[Bug]: POST /p2p-draft-backup persists host_peer_id/snapshot_json unbounded (WS path bounds them)


dearsishs:
## Description
The public `POST /p2p-draft-backup` HTTP endpoint validates only `draft_code` and then persists the client-supplied `host_peer_id` and `snapshot_json` to SQLite verbatim, echoing both back on `GET /p2p-draft-backup/{code}`. No size or shape bound is applied. The WebSocket lobby path already bounds the analogous `host_peer_id` to `MAX_TOKEN_LEN` (128 bytes, plus control-character rejection) via `validate_token`, so the same field is bounded on one transport and unbounded on the other.

## Steps to reproduce
1. Start a `phase-server` (or lobby-only) instance.
2. `POST /p2p-draft-backup` with a valid 6-char `draft_code`, a `host_peer_id` longer than 128 bytes (and/or a multi-hundred-KB `snapshot_json`, and/or control characters).
3. `GET /p2p-draft-backup/{code}`.

## Expected behavior
The oversized/control-char `host_peer_id` is rejected with `400 Bad Request` (matching the WS `MAX_TOKEN_LEN` bound), and `snapshot_json` is capped to a sane byte ceiling — before the DB write.

## Actual behavior
The row is stored verbatim and echoed back to any caller; the only ceiling is axum's default ~2 MB body limit, and the same `draft_code` row can be overwritten repeatedly.

## Impact
Untrusted client input is persisted and re-served without the size/shape bound its sibling WebSocket path already enforces — inconsistent validation across transports, control characters stored, and row-overwrite amplification.

## Component
- `crates/phase-server/src/admin.rs` (`p2p_backup_store`)
- `crates/phase-server/src/persistence.rs` (`save_p2p_backup`)
- `crates/lobby-broker/src/validation.rs` (`validate_token` / `MAX_TOKEN_LEN` — the WS bound this should match)

mike-theDude:
Reviewed against current `main`: fixed by [7cfa977cb](https://github.com/phase-rs/phase/commit/7cfa977cb38d03ba9f78869ab514bf5202b14cfc). The HTTP backup endpoint now validates `host_peer_id` with the shared wire-token bounds, rejects control characters, and caps `snapshot_json` at 1 MiB before persistence. Boundary and rejection regressions cover each reported field. This appears resolved; please close the issue.

matthewevans:
[...]

PR #5053, opened 2026-07-04
fix: redact seat tokens from unauthenticated P2P draft backup HTTP endpoint

RealDiligent:
## Summary

- **Root cause:** `POST /p2p-draft-backup` persisted the host snapshot JSON verbatim, including per-seat `seatTokens` and `kickedTokens`. `GET /p2p-draft-backup/{code}` is unauthenticated and keyed only by the 6-character draft code, so anyone who knows/guesses the code could read seat session credentials and impersonate draft players.
- **Impact:** Credential leak on the public HTTP backup surface for P2P cube/pod drafts that use server backup upload.
- **Fix:** Strip `seatTokens` / `kickedTokens` before SQLite persistence and before GET responses (defense in depth for legacy rows). Reject backup overwrites when `host_peer_id` does not match the existing row owner.

Complements the size/shape bounds from #2198 — this addresses secret redaction, not payload size.

## Test plan

- [x] Unit tests: redaction removes `seatTokens` and `kickedTokens` while preserving draft metadata
- [x] Unit tests: overwrite guard rejects mismatched `host_peer_id`
- [ ] CI green

### DIFF
--- a/crates/phase-server/src/admin.rs
+++ b/crates/phase-server/src/admin.rs
@@ -6,7 +6,9 @@ use serde::Deserialize;
 use serde_json::Value;
 use tracing::{info, warn};
 
-use server_core::guard_p2p_backup;
+use server_core::{
+    guard_p2p_backup, guard_p2p_backup_overwrite, redact_p2p_backup_snapshot_secrets,
+};
 
 use crate::AppState;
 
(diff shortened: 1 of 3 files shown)


matthewevans:
Initial review on current head `ba780242806181cf4247b05bb2ad0dc1218b2237`: I do not see a code blocker in the narrow server-side redaction path. The fix removes the top-level `seatTokens` / `kickedTokens` fields that the P2P draft host currently uploads in `snapshot_json`, applies redaction both before persistence and before GET responses, and adds focused unit coverage for redaction plus mismatched-host overwrite rejection.

I’m holding approval/enqueue for live verification. This is a security-sensitive unauthenticated backup endpoint, and the current CI/check set is still pending; I want the full current-head check result before this is marked ready.

matthewevans (approved):
Approved after current-head review: the server-side backup path strips the top-level P2P draft seat credentials before persistence and again before GET responses, rejects mismatched-host overwrites, and the checked unit coverage plus current CI cover the security boundary.

PR #5071, opened 2026-07-04
fix(engine): cast-copy-of-a-card copies never cease to exist (CR 704.5e / 707.10a)

real-venus:
## Summary
Fixes two engine bugs for "cast a copy of a card" (`Effect::CastCopyOfCard`): (1) the copy never ceased to exist off the battlefield/stack (CR 704.5e / 707.10a), persisting forever in the graveyard; (2) the CR 707.10f copy→token transition for a resolving permanent-spell copy happened too late (after `replace_event` and the zone-change record), so token-scoped ETB replacements observed it as a nontoken.

## Files changed
- crates/engine/src/game/sba.rs
- crates/engine/src/game/zones.rs
- crates/engine/src/game/stack.rs

## CR references
CR 704.5d, CR 704.5e, CR 707.10a, CR 707.10f, CR 608.3f, CR 400.7, CR 111.8

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine` — 14986 pass / 0 fail (all integration binaries ok)
- Both defects proven by revert-failing tests: copy-in-graveyard ceases to exist; a resolving permanent copy is observed as a token by a token-scoped ETB replacement (matched during `replace_event`).
- `gen-card-data` / `coverage` / `semantic-audit` — N/A: engine-runtime rules fix (SBA + stack resolution), no card parsing changed.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/game/sba.rs
+++ b/crates/engine/src/game/sba.rs
@@ -1616,7 +1616,12 @@ fn check_counter_cancellation(
 }
 
 /// CR 704.5d: A token that's in a zone other than the battlefield ceases to exist.
-/// Tokens on the stack are excluded — spell copies resolve before the next SBA check.
+/// CR 704.5e + CR 707.10a: A copy of a card in a zone other than the stack or the
+/// battlefield also ceases to exist. Both are checked here because both are
+/// non-card objects swept by the same removal loop. The stack is excluded for both
+/// so spell copies (and copies of cards resolving as spells) finish resolving
+/// before the next SBA check; the battlefield is legal for a copy of a card
+/// (CR 707.10f) but not for a token off-battlefield.
 fn check_token_cease_to_exist(state: &mut GameState, any_performed: &mut bool) {
     let tokens_to_remove: Vec<(
         crate::types::identifiers::ObjectId,
@@ -1625,7 +1630,10 @@ fn check_token_cease_to_exist(state: &mut GameState, any_performed: &mut bool) {
     )> = state
         .objects
         .iter()
-        .filter(|(_, obj)| zones::token_is_outside_battlefield_and_stack(obj))
+        .filter(|(_, obj)| {
+            zones::token_is_outside_battlefield_and_stack(obj)
+                || zones::copy_of_card_outside_battlefield_and_stack(obj)
+        })
         .map(|(id, obj)| (*id, obj.zone, obj.owner))
         .collect();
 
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
Thanks for tackling the missing CR 704.5e cleanup. I think the off-battlefield SBA sweep is pointed at the right object class, but the CR 707.10f token transition for permanent spell copies currently happens too late in the stack-to-battlefield path.

`deliver_replaced_zone_change` flips `is_copy` to `false` and `is_token` to `true` only after the replacement pipeline has already matched the `ZoneChange` and after the zone-change record is built from the pre-move object. That means token-scoped ETB replacements and ETB trigger filters see a resolving permanent copy as a nontoken object, even though CR 707.10f says the copy becomes a token as it resolves onto the battlefield.

Concrete path:

- `stack.rs` calls `replacement::replace_event(state, proposed, events)` before delivery.
- `zone_pipeline.rs` does the copy-to-token flip inside `deliver_replaced_zone_change`, after delivery has started.
- `GameObject::zone_change_record` snapshots `is_token` from the object before that flip.
- `filter.rs` evaluates `FilterProp::Token` / `FilterProp::NonToken` from that snapshot for zone-change triggers.

Please move the permanent-copy token identity to the stack-to-battlefield entry seam early enough that both ETB replacement matching and the emitted `ZoneChanged` record observe the object as a token. Also add a production-path regression test for a resolving permanent spell copy that is observed by a token-entry filter/replacement, not only the final `obj.is_token` state after `resolve_top`.

matthewevans:
Processed current head `ca10024dddd53f295cb2ebf6b74f872743350236`.

The prior implementation concern appears addressed on this head: the permanent-copy `is_copy -> is_token` transition moved into `stack.rs::resolve_top` before the battlefield-entry `replace_event` consult, and the added token-scoped ETB replacement regression exercises that pre-delivery seam. I also verified the cited CR anchors locally (`CR 704.5e`, `CR 707.10a`, `CR 707.10f`, `CR 608.3f`, `CR 111.8`).

I am still holding rather than approving/enabling auto-merge because the review-loop proof gate reports `proof_gap: true` for the current PR metadata: the AI contributor template sections `Files changed`, `Track`, and `LLM` are still not in the expected filled format, and current CI is still running. Please update those template sections on the PR body so the queue-safety proof gate is satisfied; the code-path issue from the earlier changes-requested review is no longer the blocker I see on this head.

real-venus:
> Processed current head `ca10024dddd53f295cb2ebf6b74f872743350236`.
> 
> The prior implementation concern appears addressed on this head: the permanent-copy `is_copy -> is_token` transition moved into `stack.rs::resolve_top` before the battlefield-entry `replace_event` consult, and the added token-scoped ETB replacement regression exercises that pre-delivery seam. I also verified the cited CR anchors locally (`CR 704.5e`, `CR 707.10a`, `CR 707.10f`, `CR 608.3f`, `CR 111.8`).
> 
> I am still holding rather than approving/enabling auto-merge because the review-loop proof gate reports `proof_gap: true` for the current PR metadata: the AI contributor template sections `Files changed`, `Track`, and `LLM` are still not in the expected filled format, and current CI is still running. Please update those template sections on the PR body so the queue-safety proof gate is satisfied; the code-path issue from the earlier changes-requested review is no longer the blocker I see on this head.

@matthewevans 

Done.

Please kindly review this.

matthewevans (approved):
Maintainer review passed for current head; proof gate is satisfied and the previous blocker is addressed. Enqueuing via authorized review handler.

PR #5079, opened 2026-07-04
fix(parser): UNSUPPORTED full-set one-off: Bad Wolf Bay

ntindle:
## Summary

**Root cause:** UNSUPPORTED full-set one-off: Bad Wolf Bay

## Cards corrected
- Bad Wolf Bay

## Fix
Implemented cluster 90 (Bad Wolf Bay, WHO Plane) exactly per the approved plan. The sole misparse was the chaos ability's leaf "cards can't enter from exile this turn," which lowered to Effect::Unimplemented. Built the missing class-level primitive as a new floating GameRestriction variant across all layers.

WHAT WAS BUILT (new engine variant, add-engine-variant checklist followed):
- types/ability.rs: new GameRestriction::CantEnterBattlefieldFrom { source, expiry, filter: TargetFilter } — the resolution-generated, duration-bound cousin of the permanent StaticMode::CantEnterBattlefieldFrom static (Grafdigger's Cage). Typed (no raw bool); subject+origin-zone carried in the reused TargetFilter/FilterProp::InAnyZone vocabulary.
- game/effects/add_restriction.rs: joined the new variant into the 3 exhaustive fill_runtime_fields match blocks (source-fill OR-arm; affected-player and expiry no-op OR-arms).
- game/zones.rs: extended is_blocked_from_entering_battlefield (the single ETB gate, CR 614.1d) to also scan state.restrictions, reusing filter::matches_target_filter + FilterContext::from_source exactly like the sibling statics loop.
- game/turns.rs: joined the variant into both retain prunes (cleanup EndOfTurn OR-arm CR 514.2; untap retain=true arm).
- game/replacement.rs: joined the is_prevention_disabled `=> false` arm (not a damage restriction).
- game/coverage.rs: converted the AddRestriction arm in line_matches_effect_type to a nested match GameRestriction so the description-less line does not become a semantic-audit false positive (Step 8).
- parser/oracle_effect/mod.rs: new nom combinator cant_enter_battlefield_from_zone (opt type-word prefix -> alt(card noun) -> tag(" can't enter") -> opt(" the battlefield") -> parse_enters_origin_zone), wrapped via nom_on_lower and dispatched in parse_effect_clause right after try_parse_damage_prevention_disabled. Pure combinators, no string dispatch. Generalizes over exile/graveyard/library/hand and any subject type.
- client/src/adapter/types.ts: extended the hand-authored GameRestriction display mirror into a discriminated union (added CantEnterBattlefieldFrom + Boy-Scout ProhibitActivity arm + the two drifted RestrictionExpiry arms). Type-only; no component reads the discriminant.

RUNTIME PROOF + tests (building-block level): parser test asserts the combinator yields correct zone+type across bare/any-card, type-scoped, and trailing-"this turn" forms; full-card test asserts 0 Unimplemented and the Planeswalk SequentialSibling chain survives; zones.rs runtime card-test drives real move_to_zone -> block-from-exile, then execute_cleanup prune -> move succeeds (CR 611.2a/614.1d/514.2). All 3 pass. Emergent correct interaction: the card's own Trigger-1 flicker return is also blocked when chaos ensued, via the same gate.

VERIFICATION: cargo fmt clean; clippy --all-targets -D warnings exit 0; cargo test -p engine exit 0 (incl. 3 new tests); re-parse shows 0 Unimplemented/0 Unknown with the exact intended AST; gen-card-data shows the card supported; semantic-audit reports 0 findings for the card; client pnpm type-check passes. Parser combinator gate: my added lines contain no string-dispatch (the only .find is an iterator method inside a #[test]); check-parser-combinators.sh exit=1 is pre-existing baseline noise from files I never touched (it diffs against an ancient ancestor commit).

NOTES: (1) Restored crates/engine/data/known-tokens.toml, which gen-card-data.sh regenerated with a 9113-line token-data regression (this worktree lacks fetched token-set data for some cards) — that is a generation side effect, not my source fix, and would be harmful if committed. (2) crates/engine/data/oracle-subtypes.json carries a pre-existing single-line change ("Shi-Ar") that was already modified at task start, before any of my actions; left untouched per multi-agent safety. No commits made. No stop-and-return items; no CR ambiguity.

## Files changed
- crates/engine/src/game/effects/add_restriction.rs
- crates/engine/src/game/zones.rs
- client/src/adapter/types.ts

## CR references
- CR 611.2a
- CR 614.1d
- CR 304.4
- CR 307.4
- CR 400.4a
- CR 514.2
- CR 101.2
- CR 601.2a

## Verification
- `cargo fmt --all` — pass
- `check-parser-combinators.sh (merge-base 184c1304b)` — pass
- `oracle-gen data --filter 'bad wolf bay'` — pass
Cards confirmed re-parsed correctly: Bad Wolf Bay


matthewevans (changes requested):
**[MED]** `CantEnterBattlefieldFrom` accepts arbitrary trailing duration text but always lowers to `EndOfTurn`. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:2623` ignores the nom remainder from `cant_enter_battlefield_from_zone`, and `crates/engine/src/parser/oracle_effect/mod.rs:2639` unconditionally stores `RestrictionExpiry::EndOfTurn`; the added test at `crates/engine/src/parser/oracle_effect/tests.rs:38709` explicitly treats trailing `" this turn"` as harmless leftover instead of parsing/validating the duration. Why it matters: this new parser arm will also accept phrases like `cards can't enter from exile until your next turn` or any other trailing text after the origin zone and silently give them the wrong cleanup timing, while the lifecycle code only arms/prunes `ProhibitActivity` for the non-EOT `RestrictionExpiry` variants. Suggested fix: either require the remainder to be empty or exactly `" this turn"` for this PR, or parse/thread the duration and make the restriction lifecycle handle every supported `RestrictionExpiry` variant for `CantEnterBattlefieldFrom`.

Separately, the current head still lacks the required `<!-- coverage-parse-diff -->` evidence for this parser/engine surface, so I can't treat it as queue-ready even after the code issue is addressed.

matthewevans (approved):
Addressed the maintainer blocker by making the CantEnterBattlefieldFrom parser fail closed: it now consumes only optional ' this turn' for the fixed EndOfTurn expiry and rejects other trailing duration text instead of silently lowering it. Coverage-parse-diff is already present. Local checks: cargo fmt --all; git diff --check; check-parser-combinators; cargo test -p engine --lib cant_enter_battlefield_from_zone_parses_zone_and_type.

PR #5087, opened 2026-07-04
Parse counted "you had N or more [type] enter this turn" conditions (Park Heights Pegasus)

ultrahighsuper:
## Problem

**Park Heights Pegasus** (in `docs/parser-misparse-backlog.md`):

> Whenever this creature deals combat damage to a player, draw a card **if you had two or more creatures enter the battlefield under your control this turn**.

The trailing effect gate is an effect-level conditional (CR 608.2c) on a CR 400.7 this-turn ETB threshold. It was being **dropped entirely**: the card drew a card on *every* combat-damage trigger, unconditionally, instead of only when two or more creatures had entered under your control that turn.

### Root cause

`parse_entered_this_turn` (`oracle_nom/condition.rs`) — the combinator behind `parse_inner_condition`, the single authority for game-state conditions — handled the `"you had ..."` surface only through a subject path that **hardcoded a count of 1** and required a leading article (`a`/`an`/`another`). The counted `"you had two or more creatures enter ..."` surface matched neither the article path nor the bare past-tense `"N or more [type] entered ..."` branch, so it returned `Err`, the trailing `if` never re-homed, and the draw resolved unconditionally.

## Fix

Extract a shared `parse_or_more_entered_count` combinator and use it for **both** surfaces:

- bare past-tense — `"N or more creatures entered the battlefield under your control this turn"`
- `"you had"`-auxiliary present-tense — `"you had N or more creatures enter the battlefield under your control this turn"`

Both denote the same CR 400.7 this-turn ETB tally, so **count** and **suffix** are the only axes that vary — one combinator, no duplication (the past-tense branch previously inlined this logic).

The combinator emits `StaticCondition::QuantityComparison { EnteredThisTurn{creature/You} >= N }`. The rest of the pipeline was already in place:

- `static_condition_to_ability_condition` bridges `QuantityComparison → AbilityCondition::QuantityCheck`,
- `strip_suffix_conditional` re-homes the trailing `if` onto the draw effect,
- the runtime already evaluates both `QuantityCheck` and `QuantityRef::EnteredThisTurn`.

So **no new engine plumbing** — this is a pure parser-completeness fix that unlocks the whole *"you had N or more [type] enter … this turn"* class, not just this card.

## Tests

- **Building block** (`condition.rs`): `test_you_had_two_or_more_enter_this_turn` asserts the counted surface yields `EnteredThisTurn{creature/You} >= 2` (not the singular `>= 1` the article-only path produced).
- **Runtime revert-probe** (`crates/engine/tests/park_heights_pegasus_draw_condition_runtime.rs`): parses the real draw clause through the production `parse_effect_chain` path and resolves it under P0.
  - *Discriminator*: with only **one** creature entered this turn, the `>= 2` gate suppresses the draw. **Verified this test fails without the fix** (the dropped condition draws unconditionally — `left: 1, right: 0`).
  - *Positive*: with **two** creatures entered, exactly one card is drawn.

## CR
- **CR 400.7** — an object entered the battlefield this turn (matches the established `FilterProp::EnteredThisTurn` / `QuantityRef::EnteredThisTurn` annotations).
- **CR 608.2c** — conditional (`if`) clause on an effect.
- **CR 109.5 / CR 205** — "under your control" scopes the tally to the controller.

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -6218,29 +6218,45 @@ fn parse_entered_this_turn(input: &str) -> OracleResult<'_, StaticCondition> {
     let had_enter_suffix = "enter the battlefield under your control this turn";
 
     if let Ok((rest, _)) = tag::<_, _, OracleError<'_>>("you had ").parse(input) {
+        // "you had N or more [type] enter ..." (counted threshold, GE) — the
+        // "you had" auxiliary reads the present-tense "enter" surface. Falls
+        // back to the singular "you had a/an/another [type] enter ..." form.
+        if let Ok(result) = parse_or_more_entered_count(rest, had_enter_suffix) {
+            return Ok(result);
+        }
         return parse_entered_this_turn_subject(rest, had_enter_suffix, 1);
     }
 
     // Branch 1: "N or more [type] entered..."
-    if let Ok((after_n, n)) = parse_number(input) {
-        let after_n = after_n.trim_start();
-        if let Ok((type_and_rest, _)) = tag::<_, _, OracleError<'_>>("or more ").parse(after_n) {
-            if let Ok((rest, type_text)) =
-                take_until::<_, _, OracleError<'_>>(entered_suffix).parse(type_and_rest)
-            {
-                let (rest, _) = tag(entered_suffix).parse(rest)?;
-                let (filter, _) = parse_type_phrase(type_text.trim());
-                let filter = inject_controller_you(filter);
-                return Ok((
-                    rest,
-                    make_quantity_ge(QuantityRef::EnteredThisTurn { filter }, n),
-                ));
-            }
-        }
+    if let Ok(result) = parse_or_more_entered_count(input, entered_suffix) {
+        return Ok(result);
     }
 
     // Branch 2: "a/an/another [type] entered..."
     parse_entered_this_turn_subject(input, entered_suffix, 1)
+}
+
+/// CR 400.7: Parse "N or more [type] <suffix>" into a GE threshold on the
+/// this-turn battlefield ETB count. Shared by the bare past-tense surface
+/// ("N or more creatures entered the battlefield under your control this
+/// turn") and the "you had"-auxiliary present-tense surface ("you had N or
+/// more creatures enter the battlefield under your control this turn", e.g.
+/// Park Heights Pegasus) — both denote the same CR 400.7 this-turn tally, so
+/// the count and suffix are the only axes that vary.
+fn parse_or_more_entered_count<'a>(
+    input: &'a str,
+    suffix: &'static str,
+) -> OracleResult<'a, StaticCondition> {
+    let (after_n, n) = parse_number(input)?;
+    let (type_and_rest, _) = tag("or more ").parse(after_n.trim_start())?;
+    let (rest, type_text) = take_until(suffix).parse(type_and_rest)?;
+    let (rest, _) = tag(suffix).parse(rest)?;
+    let (filter, _) = parse_type_phrase(type_text.trim());
+    let filter = inject_controller_you(filter);
+    Ok((
+        rest,
+        make_quantity_ge(QuantityRef::EnteredThisTurn { filter }, n),
+    ))
 }
 
 fn parse_entered_this_turn_subject<'a>(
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[MED]** The new parser path uses the live-board `EnteredThisTurn` quantity instead of the existing battlefield-entry history authority. Evidence: `crates/engine/src/parser/oracle_nom/condition.rs:6258` builds `QuantityRef::EnteredThisTurn`, whose resolver only counts objects still in `Zone::Battlefield` with `entered_battlefield_turn == Some(state.turn_number)` (`crates/engine/src/game/quantity.rs:2530`). The repo already has `QuantityRef::BattlefieldEntriesThisTurn` for “who had N or more [type] enter ... this turn,” backed by `battlefield_entries_this_turn` snapshots that count entries even after the object leaves (`crates/engine/src/types/ability.rs:4809`). Why it matters: Park Heights Pegasus should still see a creature that entered under your control this turn even if it died, bounced, or was sacrificed before the combat-damage draw condition resolves; the current implementation incorrectly suppresses the draw in that legal game state. Suggested fix: lower the counted “you had N or more [type] enter ... this turn” condition to the battlefield-entry snapshot quantity/condition authority and add a regression where one of the two entering creatures has left the battlefield before the draw condition resolves.

Also, the parser-diff sticky is still absent on this head; please let the current CI produce `<!-- coverage-parse-diff -->` before this is reconsidered.

ultrahighsuper:
Thanks — good catch, fixed in `fb2e8344b`.

**Authority switched to the battlefield-entry snapshot.** The counted `parse_or_more_entered_count` combinator now lowers to `QuantityRef::BattlefieldEntriesThisTurn { player: PlayerScope::Controller, filter }` instead of the live-board `EnteredThisTurn`. That reads the `battlefield_entries_this_turn` snapshots (CR 403.3 + CR 608.2h), so a creature that entered under your control this turn still counts after it has died / bounced / been sacrificed before the combat-damage draw resolves. `PlayerScope::Controller` scopes "under your control" (the runtime keys on `record.controller`), so the type filter carries no controller of its own — matching the existing `who had N or more … enter` for-each authority.

This applies to **both** counted surfaces the shared combinator serves — the present-tense `you had N or more … enter …` (Park Heights Pegasus) and the bare past-tense `N or more … entered …` — since both describe the same historical entry tally; the two counted-condition unit assertions were updated accordingly.

**Regression added** (`pegasus_draws_when_entered_creatures_have_since_left`): two creatures are recorded as having entered under P0's control this turn, but neither is on the battlefield when the draw resolves — the draw still fires. That test fails under the old live-board count (0 survivors → suppressed) and passes on the snapshot authority.

The parser-diff sticky should populate once this head's CI finishes (it was still mid-run at the previous review).

matthewevans:
Reviewed the updated `fb2e8344b` head.

The previous code concern looks resolved: the counted `you had N or more ... enter ... this turn` path now lowers to `QuantityRef::BattlefieldEntriesThisTurn { player: PlayerScope::Controller, ... }`, and the added runtime coverage includes the important case where the entered creatures have already left the battlefield before the draw condition resolves.

I am still holding this from approval/enqueue for process proof, not for CI: the PR body is missing the required AI-contributor proof sections (`Summary`, `Files changed`, `Track`, `LLM`, and `Verification`). Please update the PR body with those sections and include the exact verification you ran / are relying on. Once that is present, this can be reconsidered against the current head.

matthewevans (changes requested):
Reviewed current head `fb2e8344bd5377f2baed5ccd9cf85fc042068b2f`.

The previous runtime concern looks addressed: the counted `you had N or more ... enter ... this turn` path now uses the battlefield-entry snapshot authority, and the runtime regression covers entered permanents that have already left before the condition resolves.

I still need one cleanup before approval. The new CR annotations cite the wrong rule for the historical look-back behavior. In `crates/engine/src/parser/oracle_nom/condition.rs` and the Park Heights Pegasus runtime test, the comments describe objects still counting after leaving the battlefield but cite `CR 608.2h`; in the current rules that look-back exception is `CR 608.2i`, while the battlefield-zone rule should be cited separately. Please update the annotations to cite the correct rule pair for historical ETB membership, and keep `CR 608.2h` only where the code is specifically relying on information read during resolution.

Also please refresh the PR body so the proof matches the current implementation; it still describes the old `EnteredThisTurn` lowering.

matthewevans (approved):
Current-head maintainer review: the previous blocker is resolved on `fb2e8344bd5377f2baed5ccd9cf85fc042068b2f`. The counted "you had N or more ... enter this turn" path now uses the battlefield-entry history authority, the runtime regression covers permanents that entered and then left before resolution, CI is green, and the parse diff is scoped to the intended Park Heights Pegasus class.

PR #5088, opened 2026-07-04
fix(engine): implement the World rule state-based action (CR 704.5k)

real-venus:
## Summary
Two or more permanents with the `world` supertype (Nether Void, The Abyss, Living Plane, Concordant Crossroads — 30 cards) coexisted on the battlefield forever: the CR 704.5k world rule was entirely unimplemented, while the sibling legend rule (704.5j) and role uniqueness were fully built. Adds `check_world_rule` as a state-based action.

## Files changed
- crates/engine/src/game/sba.rs

## CR references
CR 704.5k, CR 613.7d, CR 616.1, CR 603.10a, CR 704.3, CR 614.6, CR 702.26b

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine --lib` — 15003 pass / 0 fail (114/114 in the sba module: 106 pre-existing + 8 new; purely additive)
- 8 runtime scenario tests through the real `check_state_based_actions` path; the tie=all-die and global-across-controllers tests are revert-failing guards (mutation-verified: forcing no-tie / per-controller grouping makes them fail)
- `gen-card-data` / `coverage` / `semantic-audit` — N/A: engine-runtime SBA fix, no card parsing changed (30 real World cards confirmed to load with `Supertype::World`)

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/game/sba.rs
+++ b/crates/engine/src/game/sba.rs
@@ -182,6 +182,12 @@ pub fn check_state_based_actions(state: &mut GameState, events: &mut Vec<GameEve
             // gone — only attached Roles compete for the per-(host, controller) slot.
             check_role_uniqueness(state, events, &mut any_performed, &battlefield_snapshot);
 
+            // CR 704.5k: If two or more permanents have the world supertype, all but the one
+            // that has held it the shortest time (highest timestamp) go to their owners'
+            // graveyards; on a tie for newest, all of them do. Global (not per-player) and
+            // choiceless — modeled on check_role_uniqueness.
+            check_world_rule(state, events, &mut any_performed, &battlefield_snapshot);
+
             // CR 704.5i + CR 306.9: If a planeswalker has loyalty 0, it is put into its owner's graveyard.
             check_zero_loyalty(state, events, &mut any_performed, &battlefield_snapshot);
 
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[HIGH]** The world-rule ordering uses the permanent's battlefield-entry timestamp, not the time it gained the World supertype. Evidence: `crates/engine/src/game/sba.rs` collects `(ObjectId, obj.timestamp)` for every current World permanent, but the engine already supports continuous effects that grant World after entry via `ContinuousModification::AddSupertype { supertype: Supertype::World }` (for example, `All enchantments are world.` is parsed in `crates/engine/src/parser/oracle_static/tests.rs`, and the layer application appends the supertype in `crates/engine/src/game/layers.rs`). Why it matters: CR 704.5k keeps the permanent that has had the World supertype for the shortest amount of time; an older permanent that becomes World later should be newer for this rule than a printed World permanent that entered afterward but had World longer. Suggested fix: track/derive the World-acquisition timestamp (or otherwise preserve the CR 704.5k ordering authority) and add a discriminating test where a non-World permanent gains World after a printed World permanent already has it.

Gemini's phased-out concern is not the blocker here: this implementation uses `battlefield_phased_in_ids()` and `live_battlefield_object`, so phased-out permanents are already excluded from the candidate set.

real-venus:
> **[HIGH]** The world-rule ordering uses the permanent's battlefield-entry timestamp, not the time it gained the World supertype. Evidence: `crates/engine/src/game/sba.rs` collects `(ObjectId, obj.timestamp)` for every current World permanent, but the engine already supports continuous effects that grant World after entry via `ContinuousModification::AddSupertype { supertype: Supertype::World }` (for example, `All enchantments are world.` is parsed in `crates/engine/src/parser/oracle_static/tests.rs`, and the layer application appends the supertype in `crates/engine/src/game/layers.rs`). Why it matters: CR 704.5k keeps the permanent that has had the World supertype for the shortest amount of time; an older permanent that becomes World later should be newer for this rule than a printed World permanent that entered afterward but had World longer. Suggested fix: track/derive the World-acquisition timestamp (or otherwise preserve the CR 704.5k ordering authority) and add a discriminating test where a non-World permanent gains World after a printed World permanent already has it.
> 
> Gemini's phased-out concern is not the blocker here: this implementation uses `battlefield_phased_in_ids()` and `live_battlefield_object`, so phased-out permanents are already excluded from the candidate set.

@matthewevans 

done.
please kindly review.

matthewevans (approved):
Approved current head after re-review. The previous World-acquisition timestamp blocker is fixed: printed World uses the object's entry timestamp, granted World uses the active continuous-effect timestamp authority, and the new runtime tests cover printed-vs-granted ordering plus phased-out World permanents. CI is still running, so auto-merge will wait for required checks.

matthewevans (approved):
Approved refreshed head. The only delta since the prior approval is a test-helper clippy allowance plus the branch update to current main; the production World-rule implementation and discriminating tests remain unchanged. Auto-merge will wait for required checks.

PR #5089, opened 2026-07-04
[codex] Reclassify basic land mana backlog bucket

ntindle:
## Summary

This PR now treats root #29 as a backlog false-positive bucket rather than an engine/card-data behavior bug.

Reviewer feedback pointed out that CR 305.6 grants a separate intrinsic mana ability for each basic land type. That means typed duals/triomes should continue to expose separate fixed `{T}: Add <color>` abilities at the engine/card-data layer, because ability identity and `ability_index` observe those rows.

The PR therefore removes the invalid root #29 bucket from `docs/parser-misparse-backlog.md` instead of changing synthesis behavior:

- 13 typed-land entries are removed from the backlog as not actually misparsed under CR 305.6.
- The five entries that did not match this bucket's signature in the current export are moved to `Other / uncategorized misparse` for separate triage.
- Header totals are updated to match the removed false-positive appearances.

There is no net engine or parser code change against current `origin/main`.

## Validation

- `cargo fmt --all`
- `cargo test -p engine --lib basic_land_mana_synthesis_tests` passed before the behavior change was reverted.
- Final local diff check: `git diff origin/main -- crates/engine/src/database/synthesis.rs` is empty, so no parsed card-data change is expected from this PR.
- `git diff --check` passes.

### DIFF
--- a/crates/engine/src/database/synthesis.rs
+++ b/crates/engine/src/database/synthesis.rs
@@ -439,34 +439,137 @@ impl KeywordTriggerInstaller {
     }
 }
 
+/// CR 305.6: A land with a basic land type has the matching intrinsic
+/// "{T}: Add [mana symbol]" mana ability. Card data groups multiple same-cost
+/// basic-land mana rows into one color-choice producer.
 pub fn synthesize_basic_land_mana(face: &mut CardFace) {
-    let land_mana: Vec<(&str, ManaColor)> = vec![
+    let colors: Vec<ManaColor> = [
         ("Plains", ManaColor::White),
         ("Island", ManaColor::Blue),
         ("Swamp", ManaColor::Black),
         ("Mountain", ManaColor::Red),
         ("Forest", ManaColor::Green),
-    ];
+    ]
+    .into_iter()
+    .filter_map(|(subtype, color)| {
+        face.card_type
+            .subtypes
+            .iter()
+            .any(|s| s == subtype)
+            .then_some(color)
+    })
+    .collect();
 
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[HIGH]** Multi-basic lands are collapsed into one choice ability, but CR 305.6 grants one intrinsic ability per basic land type. Evidence: `crates/engine/src/database/synthesis.rs` now emits a single `ManaProduction::AnyOneColor` when `colors.len() > 1`, while existing engine code documents and tests the CR 305.6 model as separate intrinsic abilities: `crates/engine/src/game/effects/effect.rs` checks that a land gaining all basic land types gets one fixed tap ability for each color, and `crates/engine/src/game/engine.rs` treats dual lands/triomes as lands with multiple mana options selected by `ability_index`. Why it matters: card-data support for 86 typed lands would be marked fixed while changing the ability identity/count the engine exposes for dual lands and triomes; this is not just presentation, because ability indexing and ability inspection observe those abilities. Suggested fix: keep one synthesized `{T}: Add <color>` ability per basic land type at the engine/card-data ability layer, and solve the root #29 export/UI grouping at the consumer/display or mana-option normalization seam if duplicate same-cost options need to be presented as a color choice.

matthewevans (approved):
Approved current head. This now keeps CR 305.6 behavior intact at the engine/card-data layer and limits the PR to removing the invalid backlog bucket / moving the remaining non-matching cards to uncategorized triage. The current local diff has no synthesis or parser code changes, so auto-merge can wait for required checks.

PR #5099, opened 2026-07-04
fix(parser): parse targeted perpetual p/t grants

ntindle:
## Summary

- Parse `target ... perpetually gets +N/+M` as `ApplyPerpetual::ModifyPowerToughness` instead of a temporary pump.
- Reuse `parse_target` for the target/filter grammar and keep the new arm limited to a simple P/T tail.
- Remove four cards from `docs/parser-misparse-backlog.md` root #19 after fresh export evidence.

## Card-data audit

Full raw `card-data.json` base/head comparison changed exactly these cards:

- `davriel's withering`
- `davriel, soul broker`
- `hypnotic pattern`
- `prairie survivalist`
- `scintillating encore`
- `shattering finale`
- `valiant emberkin`

Intentional notes:

- `Davriel's Withering` and `Hypnotic Pattern` are fixed by this parser arm and removed from the backlog.
- `Scion of Shiv` and `Wizened Githzerai` were already correct on current `origin/main` in the fresh root #19 export, so they were removed as stale backlog entries.
- `Prairie Survivalist` stays in the backlog: this PR fixes its first target-prefix perpetual P/T clause, but its later anaphoric `It perpetually gets +1/+0` clause is still misparsed.
- `Scintillating Encore` changes the target-prefix perpetual P/T clause; the nested delayed-trigger `return it` target also becomes the trigger source, which matches the dying object referenced by the delayed trigger.
- Local `coverage-parse-diff` reported `oracle_changed: 0`; the raw full `card-data.json` diff above was manually inspected for the seven changed cards.

## Verification

- `cargo fmt --all`
- `git diff --check`
- `cargo test -p engine --features cli --lib perpetual_parser_maps_modify_pt`
- `cargo clippy -p engine --all-targets -- -D warnings`
- Full base/head `oracle-gen` exports to `/tmp/phase-targetpt-full/{base,head}/card-data.json`
- Raw full `card-data.json` changed-key comparison
- `cargo run -p engine --features cli --bin card-data-validate -- /tmp/phase-targetpt-full/head/card-data.json` (`OK: 35397 cards parsed`)
- `cargo run -p engine --features cli --bin coverage-parse-diff -- ...` (`oracle_changed: 0`)

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -7084,6 +7084,37 @@ fn try_parse_perpetual_modify_pt(tp: TextPair) -> Option<Effect> {
 
     let lower = tp.lower;
 
+    // Explicit target form: "Target creature perpetually gets -2/-0."
+    // `parse_target` owns the target/filter grammar; this arm only recognizes the
+    // perpetual P/T tail and rejects compound grants so mixed forms stay honest.
+    if peek(tag::<_, _, OracleError<'_>>("target "))
+        .parse(lower)
+        .is_ok()
+    {
+        let (target, after_target) = parse_target(lower);
+        if !matches!(target, TargetFilter::Any) {
+            let (rest, _) = space1::<_, OracleError<'_>>(after_target).ok()?;
+            let (rest, _) = tag::<_, _, OracleError<'_>>("perpetually ")
+                .parse(rest)
+                .ok()?;
+            let (rest, _) = alt((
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed current head. This reuses the existing `ApplyPerpetual` + `ModifyPowerToughness` model and delegates the target phrase to `parse_target`, so the change is parameterized over the target/filter grammar rather than proliferating a new special case. The parse-diff matches the claimed targeted perpetual P/T class, and the new tests cover opponent-controlled and you-control target filters.

PR #5109, opened 2026-07-04
fix(parser): parse supertype-qualified enchant targets

ntindle:
## Summary

- Parse supertype-qualified `Enchant` targets such as `snow land you control`, `basic land you control`, and `legendary creature`.
- Reuse the shared target supertype parser so Aura legality emits `FilterProp::HasSupertype` instead of dropping the Enchant keyword payload.
- Keep multi-leg Enchant lists per-leg, so `legendary creature or planeswalker` does not leak `Legendary` onto the planeswalker leg.
- Remove `On Thin Ice` from root cause #31 in `docs/parser-misparse-backlog.md`.

## Root Cause

`parse_type_phrase` already understood supertype prefixes, but the Enchant keyword path used the narrower `parse_enchant_type_leg` directly. As a result, `Enchant:snow land you control` and sibling forms failed to produce a typed `Keyword::Enchant(...)` target filter.

## Parse Audit

Focused before/after export set:

- Root #31: `Cabal Stronghold`, `Flaccify`, `On Thin Ice`, `Open the Omenpaths`, `Rainbow Vale`, `The Great Mound`
- Expected sibling Aura cards: `Dimensional Exile`, `Ossification`, `Leyline Immersion`
- The name filter also matched `Colossification`; it was included in the audit and did not change.

Raw JSON changed cards:

- `On Thin Ice`: `keywords: []` -> `Enchant(Typed Land, controller You, HasSupertype Snow)`
- `Dimensional Exile`: `keywords: []` -> `Enchant(Typed Land, controller You, HasSupertype Basic)`
- `Ossification`: `keywords: []` -> `Enchant(Typed Land, controller You, HasSupertype Basic)`
- `Leyline Immersion`: `keywords: []` -> `Enchant(Typed Creature, HasSupertype Legendary)`

After deleting `keywords`, all four changed card objects were byte-equivalent to baseline. The other focused cards did not change.

`coverage-parse-diff` reports no parse-tree changes, which is expected because this PR changes Enchant keyword payloads rather than `parse_details` ability trees.

## Review cleanup

Addressed current Gemini review comments in `8a0cddb39`:

- Move `EnchantTypeLeg` properties directly instead of cloning them.
- Use `filters.pop().unwrap()` for the single-filter Enchant case.

## Validation

- `cargo fmt --all`
- `env -u RUSTC_WRAPPER ZIG_GLOBAL_CACHE_DIR=/tmp/phase-zig-cache cargo --config 'build.rustc-wrapper=""' test -p engine --features cli --lib enchant` (217 passed)
- `target/debug/card-data-validate /tmp/phase-enchant-supertype-baseline.json`
- `target/debug/card-data-validate /tmp/phase-enchant-supertype-after.json`
- `target/debug/coverage-parse-diff /tmp/phase-enchant-supertype-baseline.json /tmp/phase-enchant-supertype-after.json --markdown /tmp/phase-enchant-supertype-diff.md --json /tmp/phase-enchant-supertype-diff.json`
- review cleanup rerun: `cargo fmt --all`, `./scripts/check-parser-combinators.sh`, `cargo test -p engine --features cli --lib enchant -- --nocapture` (217 passed), `git diff --check`

CR references verified locally with `grep` against `docs/MagicCompRules.txt` for `205.4a`, `303.4a`, and `702.5a`.


matthewevans (changes requested):
Thanks for the update. The parser shape needs one more parse-diff/scope pass before I can approve:

**[MED]** Parse-diff includes an unexplained out-of-scope card change. Evidence: the `coverage-parse-diff` sticky reports `Say Its Name` gaining `cost/two other cards named ~ from your graveyard`, while this PR's diff is for supertype-qualified enchant targets. Why it matters: an enchant-target parser PR should not silently change an unrelated graveyard-cost parse signature. Suggested fix: regenerate/explain the parse-diff; if that change is real, split it or cover the `Say Its Name` parser change with its own discriminating evidence.

ntindle:
Updated this branch with current origin/main (a267d9bb7). The two Gemini inline suggestions were already present on the merged head, and I resolved both threads after confirming the current code matches them. Local checks: cargo fmt --all; git diff --check.

matthewevans (approved):
Maintainer re-review: the current head keeps the supertype-qualified Enchant change at the shared enchant parser seam, reuses the existing supertype parser, scopes properties per list leg, and resolves the prior Gemini suggestions. I verified the CR citations and the relevant parser gates locally; leaving enqueue for when the pending Rust test shards finish.

PR #5123, opened 2026-07-05
[codex] fix(parser): split repeated named spell conditions

ntindle:
## Summary

Fixes the Root #30 subcluster for `Sift Through Sands`: repeated named-spell cast conditions of the form `you've cast a spell named A and a spell named B this turn`.

Root cause: the generic single-spell-history parser treated the whole phrase after the first `spell named` as one literal `Named` value, producing a filter named `peer through depths and a spell named reach through mists`. The condition therefore checked for one impossible card name instead of requiring both spells to have been cast.

This PR adds a nom combinator before the generic single-spell parser that recognizes repeated `a spell named ...` items and lowers them to `StaticCondition::And` / `Or` over existing `SpellsCastThisTurn` quantity checks. The effect-condition bridge already maps that to `AbilityCondition::And` for the optional search clause.

Backlog update: removes `Sift Through Sands` from `docs/parser-misparse-backlog.md` Root #30, changing the branch-local count from 12 to 11.

## Parsed-card diff audit

Targeted baseline/after exports:

- Before: `/tmp/phase-root30-sift-baseline.json`
- After: `/tmp/phase-root30-sift-after.json`
- Filter: `Sift Through Sands`

Raw parsed card keys in both exports:

- `sift through sands`

Expected parse movement:

- Before: `SpellsCastThisTurn` had one `Named` filter: `peer through depths and a spell named reach through mists`.
- After: the optional search clause is gated by an `And` of two `SpellsCastThisTurn >= 1` checks, named `peer through depths` and `reach through mists`.

Sibling scan against current MTGJSON found exactly one repeated named-spell condition text:

```bash
jq -r '.data | to_entries[] | .key as $name | .value[] | select(.text? and (.text | test("spell named .* (and|or|and/or) a spell named"; "i"))) | $name + " :: " + (.text | gsub("\\n"; " / "))' data/mtgjson/AtomicCards.json
```

`coverage-parse-diff` reported no support-signature clusters (`oracle_changed: 0`), so the expected raw AST repair is confined to the audited card.

## Validation

- `grep -n "^201.2" docs/MagicCompRules.txt`
- `grep -n "^603.4" docs/MagicCompRules.txt`
- `cargo fmt --all`
- `git diff --check`
- `cargo test -p engine --features cli --lib youve_cast_repeated_named_spells_this_turn_is_compound -- --nocapture`
- `cargo test -p engine --features cli --lib sift_through_sands_condition_requires_both_named_spells -- --nocapture`
- `cargo test -p engine --features cli --lib spell_this_turn -- --nocapture`
- `cargo run -p engine --features cli --bin oracle-gen -- data --filter "Sift Through Sands" --output /tmp/phase-root30-sift-after.json`
- `cargo run -p engine --features cli --bin card-data-validate -- /tmp/phase-root30-sift-after.json`
- `cargo run -p engine --features cli --bin coverage-parse-diff -- /tmp/phase-root30-sift-baseline.json /tmp/phase-root30-sift-after.json --markdown /tmp/phase-root30-sift-parse-diff.md --json /tmp/phase-root30-sift-parse-diff.json`

Tilt was not running in this checkout (`No tilt apiserver found: tilt-default`), so I used the direct cargo fallback allowed by the repo instructions.


matthewevans (changes requested):
**[MED]** The new repeated named-spell helper cites the wrong Comprehensive Rule for this spell-resolution condition. Evidence: `crates/engine/src/parser/oracle_nom/condition.rs` adds `parse_repeated_named_spells_this_turn` with `CR 201.2 + CR 603.4`, but `docs/MagicCompRules.txt` defines `CR 603.4` as the intervening-if triggered-ability rule, and the parse-diff shows this helper is lowering Sift Through Sands' spell-resolution "If you've cast..." condition. Why it matters: this repo treats verified CR comments as review evidence, and a wrong CR citation is worse than no citation around rules-sensitive parser code. Suggested fix: keep the `CR 201.2` named-object grounding here and either remove `CR 603.4` from this helper comment or replace/split it with a rule that actually applies to conditional spell resolution.

ntindle:
Updated this branch onto current origin/main and fixed the review-blocking CR citation in oracle_nom/condition.rs: the repeated named-spell helper now cites CR 201.2 + CR 608.2c instead of CR 603.4. Verified CR anchors against docs/MagicCompRules.txt.\n\nLocal checks run:\n- cargo fmt --all\n- git diff --check

matthewevans (approved):
Maintainer re-review: current head fixes the prior CR citation finding; scoped parser diff, discriminating parser coverage, parse-diff, and live checks are clean.

PR #5124, opened 2026-07-05
fix(deck): stop corrupting card names that literally contain "//" (#4790)

JacobWoodson:
## Summary

Fixes #4790 — cards whose printed name literally contains `//` (e.g. **SP//dr, Piloted by Peni**) were unrecognized when added to a deck, both on import and manual add.

The client deck parser's `normalizeCardName` treated every `/` as a split-card separator, so `"SP//dr".split("/").filter(Boolean)` collapsed to two parts and the name was rewritten to `"SP // dr, Piloted by Peni"`. The engine keys the card under its real `//`-glued name, so the spaced form matched nothing. This bit both entry points because `repairParsedDeck` → `normalizeCardName` runs on import **and** on every deck load from storage.

## Fix

Treat whitespace adjacent to `//` as the separator signal:

- `//` **with** adjacent whitespace is a separator → normalize spacing to canonical `" // "` (`"Fire// Ice"`, `"Wear //Tear"` → `"… // …"`).
- `//` **glued** with no adjacent whitespace is part of a printed name → left verbatim (`"SP//dr, Piloted by Peni"`).
- Single-slash exporter forms still upgrade to canonical (`"Revival/Revenge"`, `"Who / What / When / Where / Why"`).

Genuine glued split names (`"Fire//Ice"`) are rare and still resolve through the engine's front-face `//` split, so preserving the real card name is the correct trade-off.

## Verification

- **Frontend** `deckParser` (vitest): **73 passed**; `tsc -b --noEmit` clean. New tests cover the `SP//dr` name, irregular `//` spacing, multi-part single-slash, already-canonical names, and real DFCs (`Peter Parker // The Amazing Spider-Man`, `Witch Enchanter // Witch-blessed Meadow`) in spaced and glued forms.
- **Engine** `database::card_db` (`cargo test -p engine`): **19 passed**. Added two regression tests proving `SP//dr, Piloted by Peni` resolves to itself (not mis-split) and a glued combined DFC name resolves to its front face.

## Note

A card entered as a **fully glued** combined name by hand (`"Peter Parker//The Amazing Spider-Man"`) still loads (engine resolves it to the front face) but its deck-builder thumbnail falls back to a text tile, since card-data keys multi-face cards by their spaced display name + front-face name. Real importers emit the spaced form and manual add uses the front-face name, so this only affects the rare hand-typed-glued case. Happy to add a front-face fallback to the display lookup (`scryfall.ts`) if desired.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4790, opened 2026-07-01
can't add [[sp//dr, piloted by peni]] — trying to add mutliple decks with this card, and everytime I do its unrecognize…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1521928858935361717/1521928858935361717

**Thread:** can't add [[sp//dr, piloted by peni]]
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
trying to add mutliple decks with this card, and everytime I do its unrecognized, even if I add it through phase's deck editor

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1521928858935361717:1521928858935361717:0` · discord: `1521928858935361717/1521928858935361717`</sub>

PR #5125, opened 2026-07-05
fix(cards): resolve glued combined multi-face names in image lookup

JacobWoodson:
## Summary

Makes the deck-builder image/type lookup resolve **glued combined multi-face names** (`"Front//Back"`, no spaces) so they render their art instead of a text-tile fallback.

`fetchCardData` (via `resolveNameLookupKey` in `scryfall.ts`) does a direct key lookup into the local card-data map. That map keys multi-face cards by **oracle id**, **spaced display name** (`"front // back"`), and **front-face name** — never the glued form. So a card entered as a hand-typed glued combined name missed every key and threw `"Card not in local data"`, falling back to a text tile. (The engine still resolved the card via its own `//` split, so the deck loaded — only the thumbnail was lost.)

## Fix

Add a front-face fallback to `resolveNameLookupKey`: when a `//` name isn't a direct or diacritic-folded key, split on `//` and resolve the **front face**, which the export always indexes. This mirrors the engine's `card_db` `lookup_key` behavior, so the display layer and the engine agree.

A single-faced card whose own printed name contains `//` (e.g. **SP//dr, Piloted by Peni**) is itself a primary key and hits the exact-match path first, so it is never mis-split.

## Verification

`scryfall` (vitest): the three new tests pass —
- a glued combined name (`"Peter Parker//The Amazing Spider-Man"`) resolves to the card entry via the front face;
- the canonical spaced form still resolves directly;
- a single-faced `//`-in-name card (`"SP//dr, Piloted by Peni"`) resolves to itself, not mis-split.

`tsc -b --noEmit` is clean. (Two unrelated pre-existing tests in this file shell out to `gen-scryfall-*.sh` and need `jq`/`bash`, absent in my local env — they don't touch this code path.)

## Relationship to #4790

Independent of, but complementary to, the [#4790 deck-parser fix](https://github.com/phase-rs/phase/pull/5124). That PR stops corrupting `//`-in-name cards during parsing; this PR hardens the *display* layer so any glued combined name (from old saved decks, unusual paste sources, or a future import source) shows its art. Neither depends on the other.

### DIFF
--- a/client/src/services/__tests__/scryfall.test.ts
+++ b/client/src/services/__tests__/scryfall.test.ts
@@ -218,6 +218,84 @@ describe("fetchCardData", () => {
   });
 });
 
+describe("fetchCardData — combined multi-face names", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  // A two-face card keyed the way the export does it: by front-face name and by
+  // the spaced display name, but NOT by the glued combined form.
+  function makeDfcDataMap(): Response {
+    const dfc = {
+      oracle_id: "peter-oracle",
+      face_names: ["peter parker", "the amazing spider-man"],
+      faces: [
+        { normal: "https://img.example/peter-front.jpg", art_crop: "https://img.example/peter-front-art.jpg" },
+        { normal: "https://img.example/peter-back.jpg", art_crop: "https://img.example/peter-back-art.jpg" },
+      ],
+      layout: "transform",
+      name: "Peter Parker // The Amazing Spider-Man",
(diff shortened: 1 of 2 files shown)


matthewevans:
Thanks for the follow-up in your first contribution set. This neatly closes the display-side case by resolving hand-typed glued multi-face names without regressing real `//` names like `SP//dr, Piloted by Peni`.

matthewevans (approved):
Maintainer review: ready. The Scryfall lookup fallback is ordered correctly, covered, and CI is green.

PR #5128, opened 2026-07-05
fix: require bearer auth for unauthenticated /admin/drafts endpoints

RealDiligent:
## Summary

- **Root cause:** `/admin/drafts*` routes (list, inspect, force-delete) are mounted with **no authentication**. Production nginx proxies `/` to `phase-server`, so these endpoints are internet-reachable.
- **Impact:** Any anonymous caller can enumerate drafts, read full session JSON (including per-seat `player_tokens` and lobby passwords via `to_persisted()`), and **DELETE** drafts plus their in-progress games.
- **Fix:** Mount `/admin/*` only when `PHASE_ADMIN_TOKEN` is set; require `Authorization: Bearer <token>` with constant-time comparison. Without a token configured, admin routes are absent (**404**, fail-closed). `/p2p-draft-backup*` stays public (normal client flow).

Complements merged #5038 (draft join password) and #5053 (P2P backup token redaction). Does **not** bundle admin response secret redaction (separate concern).

## Review notes (prior #5041 feedback addressed)

- Env-only token (no CLI flag)
- HTTP router tests: no token → 404; token set, missing/wrong Bearer → 401; valid Bearer → 200
- Case-insensitive `Bearer` scheme (RFC 9110)
- `merge_admin_routes` shared by production and tests

## Test plan

- [x] Unit tests: `tokens_match`, `admin_request_authorized` (incl. case variants)
- [x] Router tests: 404 / 401 / 200 paths
- [ ] CI green

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ Docker uses environment variables for the common options:
 | `PHASE_LOBBY_ONLY` | `--lobby-only` | false | Matchmaking broker mode |
 | `PHASE_LOG_JSON` | `--log-json` | false | Emit JSON logs |
 | `PHASE_LOG_DIR` | `--log-dir` | stdout | Write logs to files |
+| `PHASE_ADMIN_TOKEN` | — | unset | Bearer token for `/admin/*` endpoints (environment variable only; no CLI flag). Unset ⇒ admin routes disabled (404). Send as `Authorization: Bearer <token>`. |
 
 You can also pass server flags after the image name:
 
--- a/crates/phase-server/Cargo.toml
+++ b/crates/phase-server/Cargo.toml
@@ -40,4 +40,3 @@ ngrok = ["dep:ngrok"]
 futures-util = "0.3"
 tempfile = "3"
 tokio-tungstenite = "0.29"
-tower = { version = "0.5", features = ["util"] }
--- a/crates/phase-server/Cargo.toml
+++ b/crates/phase-server/Cargo.toml
@@ -40,3 +40,4 @@ ngrok = ["dep:ngrok"]
 futures-util = "0.3"
 tempfile = "3"
 tokio-tungstenite = "0.29"
+tower = { version = "0.5", features = ["util"] }
--- a/crates/phase-server/src/main.rs
+++ b/crates/phase-server/src/main.rs
@@ -7177,9 +7177,7 @@ mod admin_auth_tests {
         app.with_state(app_state)
     }
 
-    async fn spawn_admin_test_server(
-        app: Router,
-    ) -> (String, tokio::task::JoinHandle<()>) {
+    async fn spawn_admin_test_server(app: Router) -> (String, tokio::task::JoinHandle<()>) {
         let listener = tokio::net::TcpListener::bind("127.0.0.1:0")
             .await
             .expect("bind test server");
--- a/crates/phase-server/src/main.rs
+++ b/crates/phase-server/src/main.rs
@@ -1158,25 +1158,7 @@ async fn main() {
         None => info!("admin HTTP endpoints disabled (set PHASE_ADMIN_TOKEN to enable)"),
     }
     if let Some(token) = admin_token.as_deref().filter(|t| !t.is_empty()) {
-        let auth_layer = |expected: Arc<str>| {
-            from_fn(move |request: Request, next: Next| {
-                let expected = expected.clone();
-                async move { require_admin_auth(expected, request, next).await }
-            })
-        };
-        let list_auth = auth_layer(Arc::from(token));
-        let detail_auth = auth_layer(Arc::from(token));
-        app = app
-            .route(
-                "/admin/drafts",
-                get(admin::admin_list_drafts).route_layer(list_auth),
-            )
-            .route(
-                "/admin/drafts/{code}",
-                get(admin::admin_get_draft)
-                    .delete(admin::admin_delete_draft)
-                    .route_layer(detail_auth),
-            );
+        app = mount_admin_routes(app, token);
     }
 
     let app = app.layer(cors).with_state(AppState {
@@ -1297,6 +1279,28 @@ fn admin_token_from_env() -> Option<String> {
         .ok()
         .map(|t| t.trim().to_string())
         .filter(|t| !t.is_empty())
+}
+
+/// Mount bearer-guarded `/admin/*` routes on a router that will receive `AppState`.
+fn mount_admin_routes(mut app: Router<AppState>, admin_token: &str) -> Router<AppState> {
+    let auth_layer = |expected: Arc<str>| {
+        from_fn(move |request: Request, next: Next| {
+            let expected = expected.clone();
+            async move { require_admin_auth(expected, request, next).await }
+        })
+    };
+    let list_auth = auth_layer(Arc::from(admin_token));
+    let detail_auth = auth_layer(Arc::from(admin_token));
+    app.route(
+        "/admin/drafts",
+        get(admin::admin_list_drafts).route_layer(list_auth),
+    )
+    .route(
+        "/admin/drafts/{code}",
+        get(admin::admin_get_draft)
+            .delete(admin::admin_delete_draft)
+            .route_layer(detail_auth),
+    )
 }
 
 /// Decide whether an `Authorization` header value authorizes an admin request.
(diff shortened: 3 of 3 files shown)


matthewevans (approved):
Reviewed head `98c9d59e028b91128359674390525239243dc62a`.

This is a concrete backend security fix: `/admin/drafts*` routes are no longer mounted unless `PHASE_ADMIN_TOKEN` is configured, mounted admin routes require `Authorization: Bearer <token>`, and `/p2p-draft-backup*` remains public for the normal client flow. The router tests cover the important boundary: no token configured returns 404, configured token with missing/wrong bearer returns 401, and the valid bearer path returns 200.

matthewevans (changes requested):
Requesting changes on current head `347031969dc49f002fc9188bb7c1ad83a7daa386`.

The security fix is still worth preserving, but this head does not compile yet. The current Rust build fails in `crates/phase-server/src/main.rs` with PR-local errors:

- `merge_admin_routes<S>` calls `Router::merge` without the `S: Clone + Send + Sync + 'static` bounds Axum requires.
- The admin router test calls `.oneshot()` on `&mut Router<AppState>` with unsatisfied `ServiceExt` bounds.
- `GameDb::open` is passed a `PathBuf` where the API expects `&Path`.

Please fix those compile errors and repush; then I can re-review the auth boundary on a compiling head.

matthewevans (changes requested):
Requesting changes on current head `db9b146fe2d2a5bdb37a34bbc286be5b49e65043`.

The previous compile errors appear to be addressed, but this head still fails the Rust lint gate because it is not formatted. I reproduced the failure locally with `cargo fmt --all -- --check` in an isolated worktree; rustfmt wants changes in `crates/phase-server/src/main.rs` around the `test_app_state` `game_db` initialization and the `admin_routes_reject_missing_bearer` assertion.

Please run `cargo fmt --all` and repush before this can be approved.

matthewevans (changes requested):
Requesting changes on current head `ac7bb320d6cf5ec07be6af0aa416c0ce4c5f4f54`.

This head still does not compile in the Rust test build. The current CI failure is PR-local in `crates/phase-server/src/main.rs`:

- `merge_admin_routes<S>` still cannot merge `admin_routes`: `Router::merge` needs the other router to convert into `Router<S>`, but `admin_routes` is inferred as `Router<AppState>` while the helper is generic over `S` (`E0277` at `main.rs:1302`). The helper likely should not be generic here; use the concrete router state this server actually builds, or otherwise construct the admin router with the same state type.
- The router test still cannot call `.oneshot()` on `RouterIntoService<_, AppState>` because the required `Service` / `ServiceExt` bounds are not satisfied (`E0599` at `main.rs:7187`). Please use the Axum 0.8 testing pattern that produces a service for the concrete request body/state, or keep the test at a route/helper boundary that actually compiles.

`cargo fmt --all -- --check` now passes locally, but this cannot be approved until the Rust test build compiles.

matthewevans (changes requested):
Thanks for getting the compile/test failures fixed. I’m still not comfortable approving this security-surface change without coverage at the actual HTTP routing boundary. The current tests only exercise token/header parsing helpers, but the regression being fixed is whether  is absent when  is unset and bearer-gated when it is set. Please add route-level coverage, or extract a small router builder so the tests can prove: token unset returns 404 for , missing/wrong bearer returns 401, and a matching bearer reaches the admin handler path. That will give us concrete proof for the behavior this PR is meant to preserve.

matthewevans:
Correction to the review text above: the stripped inline spans should read:

- `/admin/drafts` should be absent when `PHASE_ADMIN_TOKEN` is unset.
- The tests should prove token unset returns 404 for `/admin/drafts`, missing/wrong bearer returns 401, and a matching bearer reaches the admin handler path.

matthewevans (changes requested):
Requesting changes on current head `07b07af677cdcab243ba394d34b4b60a37b5fc0b`.

The route-boundary tests are the right proof target, but this head does not compile in the Rust test build yet:

- `crates/phase-server/src/main.rs:7186`: `.oneshot(...)` is being called on `Router<AppState>`, but that router state type does not satisfy the `tower::ServiceExt` bounds in this form (`E0599`). The prior issue is still present, now in the extracted route-boundary test helper.
- `crates/phase-server/src/main.rs:1285`: `mount_admin_routes(mut app: Router<AppState>, ...)` trips `-D unused-mut`; remove the unused `mut`.

Please keep the route-boundary coverage, but adjust the test/router shape to compile under Axum 0.8 before this can be approved.

matthewevans:
I checked the new head `594b1063b1908387a1084b201cd5b5d2ac22bc21`. The code now addresses the specific test compile issues from the previous review, so I am holding this for the live Rust checks on that head to complete before approval/merge-queue handling.

matthewevans (changes requested):
Requesting changes on current head `ca2df709ff79c587aab5f2fec28613d07892c32a`.

The PR is still failing the Rust checks, and the current diff still has a local compile issue in the route-boundary test helper: `crates/phase-server/src/main.rs:7188` calls `.ready()`, but the test module imports `tower::Service` at line 7143. `ready()` is provided by `tower::ServiceExt`, so this head will not satisfy the route-boundary proof yet. The workflow logs are not downloadable while the run is still in progress, but the Rust lint/tests are already red for this head and this diff explains the failure.

Please keep the HTTP boundary tests, but make the service helper compile under the current Axum/Tower types before this can be approved.

matthewevans:
Additional current-head blocker on `ca2df709ff79c587aab5f2fec28613d07892c32a`: `crates/phase-server/Cargo.toml` adds the `tower` dev-dependency, but this PR does not update `Cargo.lock`. The locked Rust checks will need the lockfile regenerated and committed as well.

matthewevans (changes requested):
Requesting changes on current head `4543a23da7654d7196f6a3720888b3bb27c81029`.

The route-boundary helper now uses `tower::ServiceExt` with `.oneshot(...)`, so the previous helper issue looks addressed. The remaining current-head blocker is that `crates/phase-server/Cargo.toml` adds the `tower` dev-dependency, but the PR still does not include the corresponding `Cargo.lock` update. Locked Rust checks cannot accept a new dependency without the lockfile change.

Please regenerate and commit `Cargo.lock` for this dependency update.

matthewevans (changes requested):
**[HIGH]** The current TCP route-boundary test helper still does not compile on head `9ae813cc1d54fc363e6f85f2a643315603f0cdcd`. Evidence: the Rust test job fails at `crates/phase-server/src/main.rs:7186` because `axum::serve(listener, app)` is receiving `Router<AppState>`, but `axum::serve` requires a make-service over incoming connections; the log reports `Router<AppState>: Service<IncomingStream<...>> is not satisfied`. Why it matters: this is the required security-boundary proof for the `/admin/drafts` auth fix, so the PR cannot be approved until the test build actually compiles. Suggested fix: make the helper return/serve the fully state-bound router shape used by the existing test server pattern in this file, then re-run the Rust tests.

**[LOW]** This head is also not rustfmt-clean. Evidence: the Rust lint job fails `cargo fmt --all -- --check` with diffs around `crates/phase-server/src/main.rs:7177` and `crates/phase-server/src/main.rs:7254`. Why it matters: the lint gate cannot pass even after the compile issue is fixed. Suggested fix: run `cargo fmt --all` and commit the formatting output.

matthewevans:
I pushed a maintainer rustfmt-only fixup on top of the current head to clear the remaining formatting failure. New head is `c28de063d1aa07e8148232bb2c4014f8cc2feeb4`; CI has restarted, so I am holding approval/queue handling until the current-head checks complete.

matthewevans (changes requested):
The latest head `d6887821aafd8b46044f0c2d13e49e6f16267474` still has a compile blocker in the admin route test helper.

`test_admin_router` was changed from returning `Router` to returning `()`, but the body still ends with `app.with_state(app_state)`. That makes the helper unusable as written: `spawn_admin_http_test` assigns the result to `app` and passes it to `axum::serve`, so this will type-check as unit instead of a router.

Please restore the helper return type to `Router` (or otherwise return the router explicitly) and let the current-head Rust checks complete on the fixed commit.

matthewevans (approved):
Current head `363c49deb5b5d405abdeee73219a814c7a171ee8` now has the requested route-level proof for the admin draft auth boundary, and the live checks are green. Approving for merge handling.

PR #5129, opened 2026-07-05
fix(parser): gate a conditional draw on "if that creature has a +1/+1 counter" (Oblivion's Hunger)

davion-knight:
## Summary
Fixes **Oblivion's Hunger** — `Target creature you control gains indestructible until end of turn. Draw a card if that creature has a +1/+1 counter on it.` The trailing "if that creature has …" condition was swallowed, so the draw fired **unconditionally** (coverage gap `Swallow:Condition_If`).

The fix widens the existing counter-conditional combinator on two axes, reusing downstream machinery that already exists and resolves at runtime (`AbilityCondition::QuantityCheck` / `QuantityRef::CountersOn` / `ObjectScope`):

- **`parse_counter_threshold`** — accept the indefinite-article form "a/an [type] counter on it" as `>= 1` (CR 122.1 / 122.1a), placed before the numeric branch (which fails-fast on "a").
- **`strip_counter_conditional`** (trailing form only) — recognize the demonstrative subject "that creature/permanent/card" and bind it to `ObjectScope::Target` (CR 115.1), gated to non-trigger context.

**Deliberate narrowing (guarded):** the demonstrative is withheld from the *leading* form ("If that creature has …, X **instead**"). Matching it there would gate out the intentional `strip_target_keyword_instead` deferral for the "deals N … instead" replacement class (**Bring Low / Strider / Urdnan**) and mis-model the replacement as additive damage. Those cards stay on their honest deferral exactly as on `main`; a regression test pins that parity.

**Scope:** Oblivion's Hunger flips to supported. **Marcus, Mutant Mayor** (same phrase in a *trigger*, plus an "If it doesn't, put a +1/+1 counter on it" else-branch that touches the shared "Otherwise" handler) is a deliberate follow-up — it remains an honest, unchanged coverage gap here rather than a risky same-PR change.

## Files changed
- `crates/engine/src/parser/oracle_effect/conditions.rs`
- `crates/engine/src/parser/oracle_effect/mod.rs` (one-line call-site thread of `ctx.in_trigger`)
- `crates/engine/src/parser/oracle_tests.rs`
- `crates/engine/tests/oblivions_hunger_conditional_draw.rs`

## CR references
- `CR 122.1` / `CR 122.1a` — a counter marker; the indefinite "a [type] counter" means one or more (`>= 1`)
- `CR 115.1` — "that creature" = the spell's target object

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high
Tier: Frontier

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine` — **17463 passed / 0 failed**
- `./scripts/gen-card-data.sh` + `cargo coverage` — **Oblivion's Hunger**: `supported: true`, `gap_count: 0` (flipped). **Bring Low / Strider / Urdnan / Marcus, Mutant Mayor**: unchanged vs `main` (still their honest deferrals).
- `cargo semantic-audit` — Oblivion's Hunger: 0 findings; Bring Low: not newly flagged.

Tests added:
- `strip_counter_conditional_article_it_source` (SHAPE) — "if it has a +1/+1 counter on it" → `QuantityCheck { CountersOn{Source, P1P1}, GE, 1 }`, with `Effect::Draw` + no-`Unimplemented` reach-guards.
- `strip_counter_conditional_demonstrative_target_non_trigger` (SHAPE) — Oblivion's trailing clause → `CountersOn { scope: Target }`.
- `strip_counter_conditional_leading_demonstrative_instead_deferred` (guard) — Bring Low's leading "If that creature has … deals 5 … instead" stays `condition: None` (upstream parity); fails under the pre-fix over-acceptance.
- `oblivions_hunger_conditional_draw.rs` — **discriminating runtime, revert-failing**: target **with** a +1/+1 counter → draws 1; target **without** (while a *different* controlled creature holds a decoy counter) → draws 0. Reverting the fix makes the draw unconditional (draws 1) → the `0` assertion fails.

## Gate A
`./scripts/check-parser-combinators.sh` — exit 0, no violations.

## Anchored on
- `crates/engine/src/parser/oracle_effect/conditions.rs` `strip_counter_conditional` / `parse_counter_threshold` / `build_counter_condition` — the existing counter-conditional combinator this PR widens (same combinator family; superset for the trailing form, leading form byte-identical to `main`).
- `crates/engine/src/parser/oracle_effect/conditions.rs` `parse_target_demonstrative_subject` — the established demonstrative-subject recognizer reused for the trailing "that creature" arm.
- `crates/engine/src/parser/oracle_tests.rs` existing `strip_counter_conditional_*` tests — the sibling test pattern the new SHAPE/guard tests mirror.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/conditions.rs
+++ b/crates/engine/src/parser/oracle_effect/conditions.rs
@@ -2199,6 +2199,19 @@ fn parse_counter_threshold(text: &str) -> Option<(Comparator, i32, CounterType, 
         return Some((Comparator::EQ, 0, counter_type, consumed));
     }
 
+    // CR 122.1 + CR 122.1a: an indefinite "a [type] counter" means one or more (>= 1).
+    // Placed before the numeric branch, whose `parse_number` fails fast on the
+    // article "a"/"an" (Oblivion's Hunger: "a +1/+1 counter on it").
+    if let Ok((rest, _)) = nom_primitives::parse_article(text) {
+        if let Ok((after_type, counter_type)) = nom_primitives::parse_counter_type_typed(rest) {
+            let after_type = after_type.trim_start();
+            if let Some(after_on) = parse_counter_on_suffix(after_type) {
+                let consumed = original_len - after_on.len();
+                return Some((Comparator::GE, 1, counter_type, consumed));
+            }
+        }
+    }
+
     let (rest, threshold) = nom_primitives::parse_number.parse(text).ok()?;
     let rest = rest.trim_start();
     type E<'a> = OracleError<'a>;
@@ -2222,11 +2235,12 @@ fn build_counter_condition(
     comparator: Comparator,
     threshold: i32,
     counter_type: CounterType,
+    scope: ObjectScope,
 ) -> AbilityCondition {
     AbilityCondition::QuantityCheck {
         lhs: QuantityExpr::Ref {
             qty: QuantityRef::CountersOn {
-                scope: ObjectScope::Source,
+                scope,
                 counter_type: Some(counter_type),
             },
         },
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Reviewed current head d92d817432b9ab7e15d22da74cc582a2cc3c5c4f. The parser change parameterizes the existing counter-condition path by scope/counter threshold, parse-diff is limited to Oblivion's Hunger's target counter draw gate, and runtime regression coverage is discriminating.

PR #5135, opened 2026-07-05
fix: require lobby password for SpectateDraft

RealDiligent:
## Summary

- **Root cause:** `JoinDraftWithPassword` enforced `lobby_meta.password`, but `SpectateDraft` only validated the draft code. Anyone with the 6-character code could spectate a password-protected lobby and receive omniscient draft views (pools/packs) without the password.
- **Fix:** Add optional `password` to `SpectateDraft` (serde-default for older clients), validate wire bounds, and call shared `check_draft_lobby_password` before registering a draft spectator. Client spectator session accepts an optional password parameter.

## Test plan

- [x] `check_draft_lobby_password` unit tests (required / wrong / correct / no meta)
- [x] `join_draft_enforces_lobby_password` still passes via shared helper
- [x] `SpectateDraft` serde roundtrip with and without password field
- [x] `guard_spectate_draft` rejects oversized password
- [ ] CI green on all Rust + frontend checks


matthewevans (changes requested):
**[HIGH]** The spectator password check becomes a no-op once the draft starts. Evidence: `crates/phase-server/src/main.rs:5818` validates `SpectateDraft` against `session.lobby_meta`, but `crates/server-core/src/draft_session.rs:319` and `crates/server-core/src/draft_session.rs:352` clear `session.lobby_meta = None` on `StartDraft`; `check_draft_lobby_password` returns `Ok(())` when lobby metadata is absent at `crates/server-core/src/draft_session.rs:32`. Why it matters: the reported leak is the in-progress draft spectator view, and after start anyone with the code can still spectate a formerly password-protected draft without providing the password. Suggested fix: persist the spectator password policy somewhere that survives draft start without re-registering the lobby, enforce that field in `SpectateDraft`, and add a production-path test that starts a password-protected draft and proves no-password/wrong-password spectating is rejected without registering a draft spectator.

RealDiligent:
Addressed @matthewevans **CHANGES_REQUESTED** on current head:

**[HIGH] spectator password after draft start**
- Added `spectator_password` on `DraftSession` + `PersistedDraftSession` (survives `lobby_meta` clear on `StartDraft`).
- `check_draft_spectator_password` enforces lobby password from `lobby_meta` or retained `spectator_password`.
- Production-path websocket test `spectate_draft_rejects_without_password_after_draft_starts`: creates password-protected draft, host sends `StartDraft`, then proves `SpectateDraft` without/wrong password returns `Error` (no `DraftSpectatorView`) and correct password succeeds.

**Gemini:** simplified visibility lookup with `Result::map`.

CI on this head will be polled to green before re-requesting review.

matthewevans (changes requested):
**[HIGH]** The current head does not compile. Evidence: GitHub Actions reports `error[E0560]: struct FixturePackSource has no field named pack_count` at `crates/server-core/src/draft_session.rs:876`, and the local diff shows `FixturePackSource { set_code, cards_per_pack, pack_count: 3 }` in `spectator_password_retained_after_draft_start`. Why it matters: this blocks all Rust test shards and the aggregate Rust gate, so the new spectator-password fix cannot be reviewed or merged on this head. Suggested fix: construct `FixturePackSource` with its actual fields only, run `cargo fmt --all`, and let the Rust checks rerun on the fixed head.

matthewevans (changes requested):
**[HIGH]** This retains the draft spectator password in persisted draft state without extending the backup redaction boundary. Evidence: `crates/server-core/src/persist.rs:72` adds `spectator_password: Option<String>` to `PersistedDraftSession`, but `crates/server-core/src/p2p_backup_guard.rs:56` only strips `seatTokens` and `kickedTokens`, and `redact_secret_keys` at `crates/server-core/src/p2p_backup_guard.rs:73` only removes those top-level keys before the unauthenticated `/p2p-draft-backup/{code}` surface echoes stored snapshots back.

Why it matters: the spectate fix creates a new retained secret. If a host backup snapshot includes the persisted draft session, anyone who can fetch that backup by draft code can recover the pod password and bypass the intended spectator gate. Suggested fix: either avoid putting the plaintext spectator password into exported backup snapshots, or extend the P2P backup redaction path with a regression test proving `spectator_password` is stripped while the server-side spectator gate still survives draft start.

matthewevans:
Backend/security re-check: the previously requested server-side items look addressed on this head: the spectator password now survives draft start, the fixture compile issue is gone, and nested backup redaction covers the persisted spectator password.

I am still not approving/enqueueing this from the sweep because this is a mixed frontend PR and the remaining acceptance is deferred for Matt via `defer-fe`. One concrete frontend boundary to check: the normal spectate UI path still appears to send only the draft code, not a password. `DraftSpectatorPage` reads only `code` from the query string and calls `watchDraft(code)`, `MultiplayerPage` navigates to `/draft-spectator?code=...`, and `LobbyView` routes the Watch action through `onSpectate(parsed.code, context)` while the password modal submits only through `onJoinGame`.

That means password-protected draft spectating may be enforced server-side but unreachable through the shipped Watch flow unless the frontend threads a password prompt/parameter into spectating too. Leaving this for Matt's frontend handling rather than taking a maintainer approval action here.

RealDiligent:
Frontend password threading for spectate is now on \6902eda79\:

- **LobbyView**: Watch-by-code opens the existing password modal when the lobby row has \has_password\; submit routes password into \onSpectate\.
- **MultiplayerPage**: \handleSpectate\ accepts optional password, passes it via navigation state to \/draft-spectator\, and prompts on typed-code \password_required\ lookup.
- **DraftSpectatorPage**: reads password from route state/query, calls \watchDraft(code, password)\, and shows inline retry UI on \password_required\ / \Wrong password\.

CI run pending on this push.

RealDiligent:
CI green on frontend head \6902eda79\ — workflow [28740155498](https://github.com/phase-rs/phase/actions/runs/28740155498):

- Rust lint (fmt, clippy, parser gate) — success
- Rust tests (shard 1/2, shard 2/2) — success
- Frontend (lint, type-check, test) — success
- Rust aggregate gate — success

This addresses the \defer-fe\ spectate password UI gap Matthew flagged: Watch flow now threads password through LobbyView → MultiplayerPage → DraftSpectatorPage.

matthewevans:
Please pick up existing issues in https://github.com/phase-rs/phase/issues - we are not accepting unsolicited feature issues.

PR #5142, opened 2026-07-05
feat(parser+engine): Imprisoned in the Moon type-swap-with-granted-ability aura (#4770)

### LINKED ISSUE #4770: Bug? — The Ai controls Strefan, Maurer Progenitor as commander, i used Imprisoned in the moon on it, and its abilities…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521676258583380079/1521676258583380079

**Thread:** Bug?
**Cards:** imprisoned in the moon, strefan, maurer progenitor, the end

## Summary
The Ai controls Strefan, Maurer Progenitor as commander, i used Imprisoned in the moon on it, and its abilities are still activating  at the end step.

### imprisoned in the moon
Enchant creature, land, or planeswalker
Enchanted permanent is a colorless land with "{T}: Add {C}" and loses all other card types and abilities.
### strefan, maurer progenitor
Flying
At the beginning of your end step, create a Blood token for each player who lost life this turn.
Whenever Strefan attacks, you may sacrifice two Blood tokens. If you do, you may put a Vampire card from your hand onto the battlefield tapped and attacking. It gains indestructible until end of turn.
### the end

---
<sub>report_id: `discord:1521676258583380079:1521676258583380079:0` · discord: `1521676258583380079/1521676258583380079`</sub>

minion1227:
Closes #4770.

**Bug:** Imprisoned in the Moon ("Enchanted permanent is a colorless land with `"{T}: Add {C}"` and loses all other card types and abilities.") parsed to only `[SetCardTypes{Land}, SetColor{[]}]` — the `with "<ability>"` clause **and** the trailing ability-strip were both silently dropped. So the enchanted permanent kept its own abilities (the #4770 report: an enchanted commander still had its abilities) and never gained the `{T}: Add {C}` mana ability.

**Fix:** `parse_enchanted_becomes_type_with_ability` in test-free `oracle_static/type_change.rs`, dispatched **before** `parse_enchanted_is_type` (whose base-P/T split does not model a `with "<ability>"` clause). It composes:
- `SetCardTypes` (replace card types, CR 205.1a) + `SetColor([])` (colorless, CR 105.2)
- `RemoveAllAbilities` (strip the permanent's own abilities)
- `GrantAbility` (the quoted ability, via the shared `parse_quoted_ability_modifications` authority — parses `{T}: Add {C}` to a mana ability)

`RemoveAllAbilities` is emitted **before** the grant so the granted ability survives the wipe (CR 613.7a written order — the same ordering the existing `RemoveAllSubtypes → AddChosenSubtype` composition relies on). The plain enchanted-is-type family (Song of the Dryads, Darksteel Mutation) has no quoted-ability clause and falls through unchanged.

**No new variant, no new runtime** — reuses existing Layer-4/Layer-6 modifications.

Verified: new `imprisoned_in_the_moon_becomes_colorless_land_with_granted_mana_ability` parser test (asserts the 4 modifications, RemoveAllAbilities before GrantAbility, EnchantedBy affected); live-parse confirms `mods=[SetCardTypes{Land}, SetColor{[]}, RemoveAllAbilities, GrantAbility{"{T}: Add {C}", is_mana_ability:true}]`; full oracle_static suite **946/946** green; parser-combinator gate + fmt clean.

Note for review: the modification *order* (RemoveAllAbilities before GrantAbility) is chosen so the granted ability survives the Layer-6 wipe; a cast-pipeline runtime test confirming the mooned permanent ends with exactly the `{T}: Add {C}` ability would be a good addition (I verified the parser AST but could not run the full cast pipeline in my environment).

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -1048,6 +1048,13 @@ pub(crate) fn parse_static_line_inner(
         }
     }
 
+    // CR 205.1a + CR 604.1: Imprisoned-in-the-Moon — "Enchanted <subject> is a
+    // colorless <type> with "<ability>" and loses all other card types and
+    // abilities." Must precede parse_enchanted_is_type, whose base-P/T split does
+    // not model the with-"<ability>" clause (issue #4770).
+    if let Some(def) = parse_enchanted_becomes_type_with_ability(&tp, &text) {
+        return Some(def);
+    }
     // CR 613.1d + CR 205.1a: "Enchanted [permanent-type] is a [type] [with base P/T N/N]
     // [in addition to its other types]" — type-changing aura effects.
     // Must come before the basic-land-type handler which is a subset of this pattern.
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -1048,10 +1048,10 @@ pub(crate) fn parse_static_line_inner(
         }
     }
 
-    // CR 205.1a + CR 604.1: Imprisoned-in-the-Moon — "Enchanted <subject> is a
-    // colorless <type> with "<ability>" and loses all other card types and
-    // abilities." Must precede parse_enchanted_is_type, whose base-P/T split does
-    // not model the with-"<ability>" clause (issue #4770).
+    // CR 205.1a + CR 613.1f: Imprisoned-in-the-Moon — "Enchanted <subject> is a
+    // colorless [<subtype>...] <type> with "<ability>" and loses all other card
+    // types and abilities." Must precede parse_enchanted_is_type, whose base-P/T
+    // split does not model the with-"<ability>" clause (issue #4770).
     if let Some(def) = parse_enchanted_becomes_type_with_ability(&tp, &text) {
         return Some(def);
     }
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -14406,6 +14406,30 @@ fn sugar_coat_becomes_colorless_food_artifact_with_granted_ability() {
         }),
         "must gain the Food subtype (subtype-before-core-type sibling shape): {mods:?}"
     );
+    // CR 205.1a: setting a subtype replaces existing subtypes from the same set,
+    // so the artifact subtype set must be wiped BEFORE AddSubtype(Food) — else a
+    // host that was already (say) a Clue would end up "Clue Food artifact".
+    let wipe_idx = mods
+        .iter()
+        .position(|m| {
+            matches!(
+                m,
+                ContinuousModification::RemoveAllSubtypes {
+                    set: crate::types::card_type::SubtypeSet::Artifact
+                }
+            )
+        })
+        .expect("must wipe the artifact subtype set before adding Food (CR 205.1a)");
+    let add_food_idx = mods
+        .iter()
+        .position(
+            |m| matches!(m, ContinuousModification::AddSubtype { subtype } if subtype == "Food"),
+        )
+        .expect("must add Food");
+    assert!(
+        wipe_idx < add_food_idx,
+        "RemoveAllSubtypes(Artifact) must precede AddSubtype(Food): {mods:?}"
+    );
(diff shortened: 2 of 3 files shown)


matthewevans (changes requested):
Thanks for the fix direction here. I’m holding this one for the current-head review findings already posted inline.

Blocking items:
- The new parser handles `colorless <core type>` but misses the same Aura family when a subtype precedes the core type, as with Sugar Coat. The local card data already models that sibling as colorless + artifact subtype/card type + loses abilities + adds ability, so this should be generalized before we mark the class covered.
- The CR annotations cite `CR 604.1` for Layer 6 and `CR 613.7a` for intra-effect ordering. Local CR text confirms Layer 6 is `CR 613.1f`, and `CR 613.6` is the relevant cross-layer continuation rule; `CR 604.1` and `CR 613.7a` do not say what the comments claim.

Please update the parser to cover the subtype+core-type sibling shape, add a parser test for it, and correct the CR comments.

minion1227:
Addressed all three blocking items (pushed `d34ca767c`, rebased on current main):

**1. Subtype-before-core-type sibling (Sugar Coat).** Generalized `parse_enchanted_becomes_type_with_ability` to accept optional subtype(s) between `colorless` and the core card type, using the case-insensitive `parse_subtype` building block. Each subtype emits an `AddSubtype` (Layer 4) placed with the other type-identity mods, before the Layer-6 ability wipe. Sugar Coat now parses to `SetCardTypes(Artifact) + SetColor([]) + AddSubtype(Food) + RemoveAllAbilities + GrantAbility(sac→3 life)`, matching the card-data model. A core-type word is never a subtype, so the loop stops at the head noun and Imprisoned ("colorless land", no subtype) is unchanged.

**2. Parser test.** Added `sugar_coat_becomes_colorless_food_artifact_with_granted_ability` asserting the artifact + colorless + Food-subtype identity and the `RemoveAllAbilities` → `GrantAbility` ordering. Full `oracle_static` suite 998/998 green; parser-combinator gate clean.

**3. CR annotation corrections** (verified against `docs/MagicCompRules.txt`):
- Layer 6 → **CR 613.1f** (ability add/remove), replacing the incorrect `CR 604.1` (dispatch.rs + type_change.rs doc).
- 'granted ability survives the wipe' → **CR 613.6** (parts of one continuous effect keep applying to the same object), replacing the incorrect `CR 613.7a` (timestamps).

Ready for another look.

matthewevans (changes requested):
Thanks for the update. The Imprisoned in the Moon path is much closer, but the new Sugar Coat sibling still has a CR 205.1a subtype replacement issue.

**[MED]** Sugar Coat still preserves old artifact subtypes. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:652` only appends `AddSubtype { subtype: Food }`, while `crates/engine/src/game/layers.rs:4806` keeps all subtypes correlated with the new `Artifact` card type unless a `RemoveAllSubtypes { Artifact }` runs first. CR 205.1a says that when an effect sets one or more subtypes, the new subtypes replace any existing subtypes from the appropriate set. Why it matters: enchanting an artifact with another artifact subtype leaves it as, for example, `Clue Food artifact` instead of just `Food artifact`, so the Sugar Coat coverage is still rules-wrong. Suggested fix: insert `RemoveAllSubtypes` for each granted subtype set before the corresponding `AddSubtype`, and add a regression that an artifact subtype is wiped before `AddSubtype(Food)`.

minion1227:
Fixed the CR 205.1a subtype-replacement issue (pushed `92a91ed4e`). When a subtype is parsed, the parser now emits `RemoveAllSubtypes` for the core type's subtype set **before** the `AddSubtype`(s), so a host that already carries an artifact subtype (e.g. a Clue) becomes "Food artifact", not "Clue Food artifact". The grammar guarantees each parsed subtype belongs to the core type's set, so one `RemoveAllSubtypes` covers them. Strengthened the Sugar Coat test to assert `RemoveAllSubtypes(Artifact)` precedes `AddSubtype(Food)`. Leaning on CI to confirm the build here.

matthewevans (approved):
Maintainer sweep: reviewed current head after the follow-up fixes. The parser/static type-change seam is clean on review, parse-diff evidence is present, and merge-when-ready can wait for remaining checks.

PR #5144, opened 2026-07-05
feat(engine): count opponents dealt any damage this turn (Furious Spinesplitter, You've Been Caught Stealing)

davion-knight:
## Summary
Supports **"for each opponent who was dealt damage this turn"** (ANY damage — combat OR noncombat) on **Furious Spinesplitter** and **You've Been Caught Stealing**. The quantity was swallowed → `Fixed 1`; only the combat form ("dealt combat damage this turn", Tymna the Weaver) was supported.

**Parameterize, don't proliferate** (per `CLAUDE.md`) — reuse the existing typed axis, no new variant/bool:
- Rename `PlayerFilter::OpponentDealtCombatDamage { source }` → `OpponentDealtDamage { kind: DamageKindFilter, source }`, where `DamageKindFilter` is the pre-existing `{ Any, CombatOnly, NoncombatOnly }` enum already threaded through the `DamageDealtThisTurn` resolver. A `#[serde(alias)]` + a `CombatOnly` default preserve legacy data (a bare `OpponentDealtCombatDamage` meant combat-only, not `Any`).
- Parser: `"dealt {combat|noncombat|} damage this turn"` → the matching kind via a nom `alt`; `"dealt combat damage"` still → `CombatOnly` (Tymna unchanged).
- Resolver: delegate the `is_combat` decision to the existing `damage_record_matches_kind`. Noncombat player damage is already recorded in `state.damage_dealt_this_turn` (`is_combat = false`), so **no new tracking infrastructure**.

## Files changed
- `crates/engine/src/types/ability.rs` (variant parameterization + serde back-compat)
- `crates/engine/src/parser/oracle_quantity.rs` (kind-aware clause combinator)
- `crates/engine/src/parser/oracle_effect/mod.rs` (token-for-each subject-strip fallback — see Scope Expansion)
- `crates/engine/src/game/quantity.rs` (resolver via `damage_record_matches_kind`)
- `crates/engine/src/game/{effects/mod.rs, effects/deal_damage.rs, effects/speed_effects.rs, ability_scan.rs, ability_rw.rs, triggers.rs, coverage.rs, replacement.rs}` (rename threaded through exhaustive matches / doc)
- `crates/engine/src/parser/{oracle_trigger.rs, oracle_trigger_tests.rs, swallow_check.rs}`
- `crates/engine/tests/furious_spinesplitter_any_damage.rs`, `crates/engine/tests/you_have_been_caught_stealing_any_damage.rs` (new)

## CR references
- `CR 120.1` — dealing damage
- `CR 120.2a` — combat damage; `CR 120.2b` — noncombat damage (the new `kind` axis)
- `CR 510.1` — combat damage step; `CR 120.9` — "damage dealt by a source" restriction

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high
Tier: Frontier

## Verification
- `cargo fmt --all` — clean
- `cargo check -p engine --all-targets` — clean (exhaustive `PlayerFilter` matches all threaded)
- `cargo test -p engine` — 0 failures (incl. updated + new parser/resolver/serde tests and both new runtime tests)
- `./scripts/gen-card-data.sh` + `cargo coverage` — **Furious Spinesplitter** & **You've Been Caught Stealing**: `supported: true`, `gap_count: 0`; **Tymna the Weaver**: unchanged (`kind: CombatOnly`)
- `cargo semantic-audit` — 0 findings for all three cards

Tests:
- Parser (oracle_quantity.rs) — `Any` and `NoncombatOnly` forms → `PlayerCount { OpponentDealtDamage { kind } }`; combat form still `CombatOnly`.
- Resolver (game/quantity.rs) — one combat + one noncombat damaged opponent in the ledger: `Any` → 2, `CombatOnly` → 1, `NoncombatOnly` → 1 (kind-discriminating). Legacy-JSON test: `{"type":"OpponentDealtCombatDamage"}` (no `kind`) → `CombatOnly`; round-trip.
- Runtime, **revert-failing** (3 players, both opponents dealt *noncombat* Shock):
  - `you_have_been_caught_stealing_any_damage.rs` — Bribe mode (`modes(&[1])`) → **2 Treasure tokens**. Pre-fix `Fixed 1` → 1; `CombatOnly` → 0. Reach-guard: parse = `OpponentDealtDamage{Any}`, no `Unimplemented`.
  - `furious_spinesplitter_any_damage.rs` — end-step trigger → **2 `+1/+1` counters** (read via `runner.state().objects[...].counters`). Same revert-failing/reach-guard pattern.

## Gate A
`./scripts/check-parser-combinators.sh` — exit 0, no violations.

## Anchored on
- `crates/engine/src/types/ability.rs` — `DamageKindFilter` (existing `{Any,CombatOnly,NoncombatOnly}`) reused as the `kind` axis, mirroring its use in the `DamageDealtThisTurn` resolver.
- `crates/engine/src/game/quantity.rs` `damage_record_matches_kind` — the existing per-record kind gate the resolver now delegates to (was a hardcoded `r.is_combat`).
- `crates/engine/src/parser/oracle_effect/mod.rs` — the token-for-each arm's new raw-then-subject-stripped ordering mirrors the pre-existing numeric-for-each arm directly above it.

## Scope Expansion
The enum parameterization alone did not flip **You've Been Caught Stealing**: the token "for each" arm (`try_parse_for_each_effect` → `try_parse_token`, `oracle_effect/mod.rs`) fed the raw subject-bearing base `"You create a Treasure token"` to `try_parse_token`, which returned `None`, stranding the correctly-parsed for-each quantity (this also swallowed the *combat* token form). Fixed by having the token arm try the raw base first, then fall back to a subject-stripped base — mirroring the numeric arm directly above. Verified additive: swept `card-data.json` for every `create … token for each …` card; multi-player subjects (`each player/opponent creates …`) are handled by `player_scope` upstream and never reach this fallback, distributive `its controller creates …` routes elsewhere, and the only newly-captured forms are singular `you create …` (owner = controller, correct). No card changes owner/scope.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -2127,7 +2127,7 @@ fn legacy_player_filter(x: &PlayerFilter) -> bool {
         PlayerFilter::AllExcept { exclude } => legacy_player_filter(exclude),
         PlayerFilter::OpponentLostLife
         | PlayerFilter::OpponentGainedLife
-        | PlayerFilter::OpponentDealtCombatDamage { .. }
+        | PlayerFilter::OpponentDealtDamage { .. }
         | PlayerFilter::OpponentOtherThanTriggering
         | PlayerFilter::OpponentOfTriggeringPlayer
         | PlayerFilter::OpponentOfTriggeringPlayerNotAttacked
@@ -6079,7 +6079,7 @@ fn rw_player_filter(x: &PlayerFilter) -> RwProfile {
         PlayerFilter::OpponentLostLife | PlayerFilter::OpponentGainedLife => {
             reads_player_of(StateKind::JournalLife)
         }
-        PlayerFilter::OpponentDealtCombatDamage { source: _ } => {
+        PlayerFilter::OpponentDealtDamage { source: _, kind: _ } => {
             reads_player_of(StateKind::JournalLife)
         }
         // D5 carrier.
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -3025,7 +3025,10 @@ fn scan_player_filter(x: &PlayerFilter) -> Axes {
             projected: true,
         },
         PlayerFilter::HasLostTheGame => Axes::NONE,
-        PlayerFilter::OpponentDealtCombatDamage { source } => {
+        // `kind` is a static damage-kind selector (combat/noncombat/any) — not an
+        // event-context, sibling, or projected-growth resource — so it carries no
+        // axis; only the optional `source` sub-filter contributes.
+        PlayerFilter::OpponentDealtDamage { kind: _, source } => {
             let mut acc = Axes {
                 event: false,
                 sibling: false,
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -1653,7 +1653,7 @@ fn fmt_quantity_ref(qty: &QuantityRef) -> String {
 }
 
 fn fmt_player_filter(pf: &PlayerFilter) -> String {
-    use crate::types::ability::PlayerRelation;
+    use crate::types::ability::{DamageKindFilter, PlayerRelation};
     match pf {
         PlayerFilter::Controller => "you",
         PlayerFilter::Opponent => "each opponent",
@@ -1661,9 +1661,14 @@ fn fmt_player_filter(pf: &PlayerFilter) -> String {
         PlayerFilter::OpponentLostLife => "each opponent who lost life this turn",
         PlayerFilter::OpponentGainedLife => "each opponent who gained life this turn",
         PlayerFilter::HasLostTheGame => "each player who has lost the game",
-        PlayerFilter::OpponentDealtCombatDamage { .. } => {
-            "each opponent who was dealt combat damage this turn"
-        }
+        // CR 120.2a/120.2b: human string reflects the damage-kind selector.
+        PlayerFilter::OpponentDealtDamage { kind, .. } => match kind {
+            DamageKindFilter::CombatOnly => "each opponent who was dealt combat damage this turn",
+            DamageKindFilter::NoncombatOnly => {
+                "each opponent who was dealt noncombat damage this turn"
+            }
+            DamageKindFilter::Any => "each opponent who was dealt damage this turn",
+        },
         PlayerFilter::OpponentAttacked { subject, scope } => match (subject, scope) {
             (AttackSubject::You, AttackScope::ThisTurn) => "each opponent you attacked this turn",
             (AttackSubject::Source, AttackScope::ThisTurn) => {
@@ -6753,7 +6758,7 @@ fn player_filter_feature(scope: &PlayerFilter) -> (&'static str, FeatureSupport)
         PlayerFilter::OpponentLostLife => ("OpponentLostLife", Handled),
         PlayerFilter::OpponentGainedLife => ("OpponentGainedLife", Handled),
         PlayerFilter::HasLostTheGame => ("HasLostTheGame", Handled),
-        PlayerFilter::OpponentDealtCombatDamage { .. } => ("OpponentDealtCombatDamage", Handled),
+        PlayerFilter::OpponentDealtDamage { .. } => ("OpponentDealtDamage", Handled),
         PlayerFilter::OpponentAttacked { .. } => ("OpponentAttacked", Handled),
         PlayerFilter::OpponentAttackingEnchantedPlayer => {
             ("OpponentAttackingEnchantedPlayer", Handled)
--- a/crates/engine/src/game/effects/deal_damage.rs
+++ b/crates/engine/src/game/effects/deal_damage.rs
@@ -1658,14 +1658,16 @@ fn collect_matching_players(
                     // CR 506.2 + CR 508.6: Count-only filter (Suppressor Skyguard);
                     // it has no live damage-recipient meaning.
                     PlayerFilter::OpponentOfTriggeringPlayerNotAttacked => false,
-                    // CR 120.1 + CR 510.1 + CR 120.9 + CR 608.2i: Each opponent
-                    // who was dealt combat damage this turn, optionally
-                    // restricted to a matching source.
-                    PlayerFilter::OpponentDealtCombatDamage { ref source } => {
-                        crate::game::quantity::opponent_dealt_combat_damage_matches(
+                    // CR 120.1 + CR 510.1 + CR 120.9 + CR 608.2i +
+                    // CR 120.2a/120.2b: Each opponent who was dealt damage of the
+                    // given kind this turn, optionally restricted to a matching
+                    // source.
+                    PlayerFilter::OpponentDealtDamage { kind, ref source } => {
+                        crate::game::quantity::opponent_dealt_damage_matches(
                             state,
                             p.id,
                             source_controller,
+                            kind,
                             source,
                             source_id,
                         )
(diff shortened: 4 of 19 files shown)


matthewevans (dismissed):
**[HIGH]** Current head does not compile in the player collection path. Evidence: `crates/engine/src/game/effects/deal_damage.rs:1670` passes `kind` from `PlayerFilter::OpponentDealtDamage { kind, ref source }` into `opponent_dealt_damage_matches`, while that helper takes `DamageKindFilter` by value at `crates/engine/src/game/quantity.rs:4525`; the sibling resolver path already calls it with `*kind`. Why it matters: this blocks the Rust lint/build job before the behavior can be accepted. Suggested fix: pass `*kind` at the `deal_damage.rs:1670` call site.

Also holding approval until the current-head parse-diff sticky is available; at review time card-data/coverage was still running and the required `<!-- coverage-parse-diff -->` evidence was absent.

matthewevans (approved):
Approved. I re-checked the current head after the parse-diff landed: the damage-kind parameterization is at the right quantity/player-filter seam, the runtime tests discriminate `Any` damage from both `Fixed 1` and `CombatOnly`, and the parse-diff matches the intended token/counter quantity changes without an unexplained scope loss.

matthewevans (changes requested):
I need to pull back the approval after an adversarial pass found a real parser-scope risk in the token-for-each fallback.

**[MED]** Subject-bearing token-for-each clauses can be accepted with the wrong token creator. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs` now falls back from raw token parsing to `subject::strip_subject_clause(base_no_duration)` before calling `token::try_parse_token`. `try_parse_token` defaults `Effect::Token.owner` to `TargetFilter::Controller`, and the early return from `try_parse_for_each_effect` only reapplies optional/condition slots before returning. It does not run the later subject-target/recipient rebinding machinery that normally lifts a subject like `each player` or `target opponent` into the token owner.

Why this matters: the fallback needed for `You create a Treasure token for each ...` is good, but as written it can also make broader subject forms parse as supported while creating the tokens for the source controller. Please either limit this fallback to the `you create ...` form needed here, or route the parsed token through the existing subject/player-scope rebinding path and add a regression for a non-controller token creator.

davion-knight:
Thanks for the adversarial catch — you're right, and it's addressed on the new head (also merged current `main`, so the conflict is resolved).

I took **option 1: limit the fallback to the controller (`you`/`you may`) form.** Rationale for not routing through the rebinding path: `thread_for_each_subject` has no `Effect::Token` arm (it handles Pump/Mill/Draw/Discard/GainLife/… and falls through Token unchanged), so option 2 would have meant adding a Token-owner arm to that shared function — broader blast radius than this fix warrants.

**Change:** new `subject::strip_controller_subject_clause` strips exactly `"you "`/`"you may "` and returns `None` for any other subject; the token arm uses it instead of `strip_subject_clause`. The raw-base attempt is unchanged, so anything `try_parse_token` owns directly (with the correct owner) still works. Non-controller creators now fall through to unsupported rather than being mis-owned.

**Concrete proof it was a real bug, now fixed:** `Yes Man, Personal Securitron` — *"its owner creates a tapped 1/1 white Soldier creature token for each quest counter on it."* Under the broad strip this parsed to `Effect::Token { owner: Controller, … }` (wrong — "its owner", after `target opponent gains control`, is not the source controller). It now correctly declines and reverts to an honest gap.

**Regression tests added** (as requested):
- `for_each_token_non_controller_subject_not_misowned_via_fallback` — "each player creates a Treasure token for each creature they control" → NOT `owner: Controller`.
- `for_each_token_target_opponent_subject_not_misowned_via_fallback` — same for "target opponent creates …".
- `for_each_token_controller_subject_still_controller_owned` — positive companion: "you create …" still `owner: Controller`.

**Verification (current head):** `cargo fmt` clean · `check-parser-combinators.sh` exit 0 · `cargo clippy -p engine -D warnings` clean · `cargo test -p engine` 0 failures · coverage: Furious Spinesplitter / You've Been Caught Stealing / Cavern-Hoard Dragon / Covetous Elegy / Wreck Hunter supported (all genuine you-create → `owner: Controller`), Yes Man honestly unsupported · `cargo semantic-audit` 0 findings on all six.

matthewevans (changes requested):
**[HIGH]** The current parse-diff evidence is scope-contaminated, so I cannot approve this head as a parser-surface PR yet. Evidence: the sticky comment updated for this head reports 27 cards / 48 signatures, including unrelated changes such as `AdditionalPhase` for Shadow of the Second Sun / Sphinx of the Second Sun / Temple of Atropos, `CastWithKeyword` for Ood Sphere / Lake Silencio / Trail of the Mage-Rings, and replacement `Moved` support for Anointed Peacekeeper / Sorcerous Spyglass. This PRs stated scope is the opponent-dealt-any-damage quantity plus the token-for-each fallout; the required parse-diff evidence needs to be narrowed back to that scope, or each extra signature needs an explicit scope/value explanation.

**[MED]** The new non-controller fallback tests still do not assert the boundary the comments say is required. Evidence: `crates/engine/src/parser/oracle_effect/tests.rs:6979` and `:7000` only assert that the direct helper result is not `Some(Effect::Token { owner: TargetFilter::Controller, .. })`, while the comments at `:6961-6968` say this helper must decline with `None` and that the full `parse_effect` path remains separately mis-owned/out of scope. That assertion would also pass for a non-controller token or unrelated accepted effect, so it does not prove the honest-gap boundary. Please assert the exact decline shape (`clause.is_none()`) for both non-controller helper cases, and make sure the full card-data path does not continue to mark non-controller creators supported with `owner: Controller`.

davion-knight:
Thanks for the review — pushed fixes for both points.

**[MED] test assertions.** Both non-controller helper tests now assert the exact decline shape `clause.is_none()` instead of `!matches!(Some(Token { owner: Controller, .. }))`, so they prove the controller-only fallback actually declines rather than merely avoiding Controller ownership:
- `for_each_token_non_controller_subject_not_misowned_via_fallback`
- `for_each_token_target_opponent_subject_not_misowned_via_fallback`

Both pass; the positive `..._controller_subject_still_controller_owned` case is unchanged and still green. The separate out-of-scope full-`parse_effect` mis-ownership path is untouched by this PR, as the doc comments note.

**[HIGH] parse-diff scope contamination.** Root cause was branch staleness — this head was ~13 commits behind `main`, so the parse-diff compared it against a newer base and attributed unrelated signatures (e.g. Ood Sphere from #5170, plus the AdditionalPhase / CastWithKeyword / Moved changes) to this PR. I've merged current `main` in; once CI regenerates the sticky evidence it should narrow back to the intended scope (opponent-dealt-any-damage quantity + token-for-each fallout). Will confirm from the refreshed report.

Separately, the bot's inline suggestion to change `kind` → `*kind` at `deal_damage.rs` is a false positive: `opponent_dealt_damage_matches` takes `kind: DamageKindFilter` by value and `DamageKindFilter: Copy`, with the match on an owned `PlayerFilter`, so `kind` is already owned — the current form is what compiles (dereffing would not).

davion-knight:
Follow-up on the `[HIGH]` parse-diff scope contamination — root cause found and fixed properly.

**It was a merge-history artifact, not real scope creep.** Both my earlier `merge main` and the "Update branch" merge left the branch with merge commits, so the parse-diff compared a head that contained newer-`main` parser features against a pinned/older baseline and attributed them (AdditionalPhase, CastWithKeyword, Moved, ReverseTurnOrder, …) to this PR.

**Fix: I rebased the branch linearly onto current `main`** (force-pushed). It now contains exactly our three commits on top of `main` with no merge commits, so `git diff main...HEAD` is **only our 19 files** — the opponent-dealt-any-damage parameterization + the token-for-each fallback + tests. No content was lost (your "Update branch" merge only pulled in #5197, which is already in the new base). The regenerated sticky should narrow to just our cards: Furious Spinesplitter, You've Been Caught Stealing (targets) and the you-create fallback cards Cavern-Hoard Dragon / Covetous Elegy / Wreck Hunter; Yes Man, Personal Securitron correctly stays an honest gap. Local `cargo coverage` confirms exactly that set.

**`[MED]` test boundary — done** (in the pushed head): both non-controller cases now assert the exact decline shape `clause.is_none()` (not the weaker "not Controller-owned"), and the positive `you create …` companion still asserts `Controller`-owned. `cargo test -p engine for_each_token_` → 6/6 pass.

**Re: the bot's `kind` → `*kind` inline on `deal_damage.rs`** — that's a false positive. `opponent_dealt_damage_matches` takes `kind: DamageKindFilter` **by value**, `DamageKindFilter: Copy`, and the match there binds `kind` by value, so `*kind` would fail to compile. The Rust lint/test jobs (and a local `cargo check -p engine`) pass as-is.

All local checks green on the rebased head: `cargo fmt` · `check-parser-combinators.sh` · `cargo clippy -p engine -D warnings` · `cargo test -p engine` (0 failures) · coverage as above · `cargo semantic-audit` 0 findings on the affected cards.

matthewevans (approved):
Approved after re-review of `c9435a3`: the prior parse-diff contamination is resolved, the token for-each fallback now declines non-controller subjects exactly, and the opponent-dealt-damage change is parameterized at the existing player-filter/quantity seam with discriminating runtime coverage for Any-vs-CombatOnly behavior. Rust tests are still running, so auto-merge/queue should remain gated by required checks.

PR #5147, opened 2026-07-05
feat(parser): bare tribal compound anthem subject "<subtype>s and other <subtype>s" (Verdeloth the Ancient)

minion1227:
Verdeloth the Ancient ("Saproling creatures and other Treefolk creatures get +1/+1.") **dropped entirely**: the compound-subject parsers only handled controller-scoped compounds ("… you control / your opponents control"), so a bare battlefield-wide tribal compound with a per-branch "other" fell through to the type-phrase fallback and failed to parse.

Adds `parse_bare_compound_subtype_subject_filter` in test-free `oracle_static/shared.rs` — the no-controller sibling of `parse_shared_controller_compound_subject_filter`: split on `" and "`, delegate each branch to `parse_continuous_subject_filter` (so the "other Treefolk creatures" branch picks up the `Another` source exclusion via the existing `"other "` arm), and `Or` the results.

Gated on **both branches resolving to subtype-scoped typed filters** (`filter_carries_subtype`) so a generic "creatures and <X>" compound is left for the fallback rather than over-claimed — important since it hooks a heavily-used function.

**No new variant, no new runtime.** CR 611.3a.

Verified: live-parse yields `Or([Typed(Creature, Subtype("Saproling"), controller:None), Typed(Creature, Subtype("Treefolk"), controller:None, props:[Another])])` with `[AddPower 1, AddToughness 1]`; new `static_bare_compound_tribal_subject_verdeloth` test; full oracle_static suite **946/946** green (no over-matching regressions); parser-combinator gate + fmt clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -2699,6 +2699,55 @@ fn parse_shared_controller_compound_subject_filter(subject: &TextPair<'_>) -> Op
     Some(TargetFilter::Or { filters })
 }
 
+/// True when `filter` is a typed filter carrying a creature subtype constraint.
+/// The gate for the bare tribal compound below, so a generic
+/// "creatures and <X>" compound is left to the type-phrase fallback.
+fn filter_carries_subtype(filter: &TargetFilter) -> bool {
+    matches!(
+        filter,
+        TargetFilter::Typed(tf)
+            if tf.type_filters.iter().any(|t| matches!(t, TypeFilter::Subtype(_)))
+    )
+}
+
+/// CR 611.3a: A bare (battlefield-wide, no-controller) compound tribal anthem
+/// subject where each branch carries its own creature subtype and the second may
+/// take a per-branch "other" source exclusion — "<subtype> creatures and [other]
+/// <subtype> creatures" (Verdeloth the Ancient: "Saproling creatures and other
+/// Treefolk creatures get +1/+1"). The controller-scoped compound is handled by
+/// [`parse_shared_controller_compound_subject_filter`]; this is the tribal
+/// battlefield form. Each branch delegates to [`parse_continuous_subject_filter`]
+/// (so "other Treefolk creatures" picks up the `Another` source exclusion via the
+/// existing "other " arm), and the branches are OR'd. Both branches MUST resolve
+/// to subtype-scoped typed filters, so a generic "creatures and <X>" compound is
+/// left for the fallback rather than over-claimed.
+fn parse_bare_compound_subtype_subject_filter(subject: &TextPair<'_>) -> Option<TargetFilter> {
+    // Controller-scoped compounds belong to the sibling handler above.
+    if parse_subject_suffix(subject, " you control").is_some()
+        || parse_subject_suffix(subject, " your opponents control").is_some()
+    {
+        return None;
+    }
+    let (left_lower, _, right_lower) = nom_primitives::scan_preceded(subject.lower, |input| {
+        value((), tag::<_, _, OracleError<'_>>("and ")).parse(input)
+    })?;
+    let right_start = subject.lower.len() - right_lower.len();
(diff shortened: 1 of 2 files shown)


matthewevans:
I’ve started reviewing this head. I’m not approving yet because the current-head parser/card-data evidence is still pending and the required parse-diff sticky is not posted yet.

One note from the bot review: the specific `sand`/`island` concern appears mitigated by `scan_preceded` being word-boundary based, so I’m not treating that as a blocker by itself. I’ll re-check once CI/card-data posts the parse-diff for this head.

matthewevans (changes requested):
The earlier `sand`/`island` concern is still not the blocker I see: `scan_preceded` is word-boundary based. Now that the parse-diff sticky is available, though, it shows a real out-of-scope semantic regression.

**[HIGH]** Bare compound subject parsing now misclaims Life and Limb as a creature-only effect. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:2705` accepts any typed subtype branch via `filter_carries_subtype`, and `shared.rs:2740` delegates both sides of the compound through `parse_continuous_subject_filter`; the parse-diff sticky shows Life and Limb added as `affects=creature Forest or creature Saproling`, while the source mtgish card data has `Forest` lands OR `Saproling` creatures. Why it matters: noncreature Forests would not receive Life and Limb's type/color/P/T changes, and the PR creates out-of-scope parser support for an unrelated card with wrong semantics. Suggested fix: restrict the new bare-compound helper to branches with explicit creature terms, or route subtype subjects through the existing land-subtype/type-change seam so Forest remains a land-subtype subject rather than becoming `creature Forest`.

minion1227:
Fixed the Life and Limb over-claim (pushed `b352833c2`). The bare tribal compound helper now requires each branch to explicitly name creatures (`branch_names_creatures`, CR 205.3m) before it claims the compound. Verdeloth ("Saproling creatures and other Treefolk creatures") still qualifies; "All Forests and all Saprolings" no longer does, so Forest stays a land subject rather than becoming `creature Forest`. Added a regression that "Forests and Saprolings" isn't claimed as a creature anthem. (Agreed the scan_preceded sand/island concern is a non-issue — it's word-boundary based.) Leaning on CI to confirm the build.

matthewevans (changes requested):
Thanks for the update. I still need one more change before this is safe to clear.

- The Life and Limb regression guard is not discriminating enough for the prior blocker. `crates/engine/src/parser/oracle_static/tests.rs` uses synthetic text (`Forests and Saprolings get +1/+1.`), which can pass if parsing returns `None` or any non-`Or` shape. That does not prove the real Life and Limb type-change path remains honest, especially while current-head parse-diff evidence is still pending. Please add a direct test for Life and Limb's actual static text that asserts it does not produce a `creature Forest` affected branch and preserves the expected existing type-change / strict-failure shape.

minion1227:
Done (pushed `97de8fbb4`). Replaced the synthetic guard with a **direct Life and Limb test** on its actual static text ("All Forests and all Saprolings are 1/1 green Saproling creatures and Forest lands in addition to their other types"). It asserts both:
1. no produced static reinterprets the land subtype **Forest** as a creature-anthem branch, and
2. the unsupported compound type-change **strict-fails** (no static) — the same shape as baseline main.

That's the discriminating check: the ungated helper produced a spurious `creature Forest` static here, which now can't happen. Verdeloth still parses correctly; full `oracle_static` suite 998/998 locally. (The previous synthetic test also happened to exercise an unrelated anthem path — dropped it.)

matthewevans (approved):
Thanks for the follow-up. I rechecked the current head after the exact-head parse-diff and checks caught up.

The previous blocker is resolved: the Life and Limb regression now uses the actual static text and confirms the bare compound helper does not reinterpret the land/type-changing static as the Verdeloth-style creature anthem. The parse-diff is present for this head and the Rust/card-data checks are green, so I’m approving this head and enabling merge-when-ready.

PR #5154, opened 2026-07-05
fix(crew): pilot tokens crew for adjusted power, not raw or doubled

JacobWoodson:
A Pilot token that "crews Vehicles as though its power were 2 greater"
(Shorikai, Genesis Engine / Kotori, Pilot Prodigy) crewed for the wrong
amount through two independent bugs:

- Double-grant: the token carried the crew static twice -- once from the
  creating effect's `with "..."` clause and again from the catalog art
  preset's rules_text -- so a 1/1 Pilot contributed 5 (1+2+2). When the
  1/1 body matched a different printing's art (e.g. the Aetherdrift
  "saddles Mounts and crews Vehicles" Pilot), the two statics had
  different keyword actions and slipped past an exact-match de-dupe. Per
  CR 111.4 the creating effect's grant is authoritative, so the catalog
  `rules_text` fallback is now display-only and injects no functional
  abilities when the token already carries granted statics.

- Frontend re-derived crew power: the crew-selection UI summed each
  creature's raw printed power (1) instead of the engine's adjusted crew
  contribution, gating the lone Pilot at 1 < N and blocking the crew
  ("crews for just 1"). The engine now sends per-creature contributions
  on the CrewVehicle choice and the view-model gates on those.

Result: a 1/1 Pilot crews for 3; a 2/2 (with a +1/+1 counter) for 4.


matthewevans (changes requested):
Thanks for the PR. I found a few blockers before this can move forward.

- `crates/phase-ai/src/policies/payment_selection.rs` still constructs `WaitingFor::CrewVehicle` without the new required `contributions` field, so the all-targets/test build path will fail before the behavior can be validated. Please thread contributions into that test helper, ideally from the same contribution calculation used by the engine.
- The engine now computes saddle contribution through `object_crew_power_contribution(..., CrewAction::Saddle)`, but `client/src/viewmodel/gameStateView.ts` still sends no contribution map for `WaitingFor::SaddleMount`. That leaves the frontend confirm gate on raw power for saddle even when the engine accepts adjusted Pilot-token saddle contribution.
- `WaitingFor::CrewVehicle.contributions` is a new serialized field without a serde default. Saved/debug-exported games paused in a `CrewVehicle` prompt before this PR would fail to deserialize on restore. Please add a safe default/normalization path for old waiting states.
- The token ability comment in `crates/engine/src/game/effects/token.rs` cites `CR 111.4`, but that rule is about token names/subtypes, not the token abilities/characteristics being implemented here. Please verify and cite the correct CR rule before landing.

Because this PR also changes frontend/view-model files, frontend acceptance is deferred to Matt under the current sweep policy; I’m not approving or enqueueing it from the automation sweep.

matthewevans (changes requested):
Thanks for the update. The phase-ai constructor issue is resolved, but the current head still has blockers.

- `SaddleMount` still gates the frontend on raw power. The engine computes saddle eligibility with `object_crew_power_contribution(..., CrewAction::Saddle)`, but `client/src/viewmodel/gameStateView.ts` maps `SaddleMount` without contributions. A Pilot-style saddle modifier can be accepted by the engine while the frontend blocks confirm. Please add index-aligned saddle contributions to `WaitingFor::SaddleMount` and wire them through the view model the same way as crew.
- `CrewVehicle.contributions` is still a required serialized field without a serde default. Older saved/debug-exported games paused in a `CrewVehicle` prompt can fail to deserialize. Please add a default and normalize/recompute when needed.
- The token ability comments still cite `CR 111.4`, which is about token names/subtypes. The ability/characteristics authority is `CR 111.3`; please update the annotation.

Frontend acceptance remains deferred to Matt under the current sweep policy.

matthewevans (changes requested):
Thanks for the update. The phase-ai constructor and predefined-token injection path look addressed on current head `c9d5300c`, but two blockers from the last review are still present.

1. `SaddleMount` still does not carry index-aligned adjusted contributions. The engine computes saddle eligibility using `object_crew_power_contribution(..., CrewAction::Saddle)`, but `WaitingFor::SaddleMount` still only serializes `eligible_creatures` (`crates/engine/src/types/game_state.rs`) and `handle_saddle_activation` still returns no contribution array. The frontend saddle path still builds `selection: { type: "totalPowerAtLeast", power: waitingFor.data.saddle_power }`, so `boardChoiceSelectedPower` falls back to raw object power. Any saddle modifier can still be accepted by the engine and blocked by the UI.

2. `CrewVehicle.contributions` is still a required serialized field. The new field on `WaitingFor::CrewVehicle` has no serde default, so older saved/debug-exported states with an in-flight crew prompt and no `contributions` field will fail to deserialize. Please make this field backward-tolerant and keep the frontend’s existing fallback behavior for absent or mismatched contribution data.

matthewevans (approved):
Current head f90cf0b73c resolves the outstanding blockers: SaddleMount now carries engine-computed contributions through the serialized waiting state and frontend board-choice gate, CrewVehicle contributions are backward-tolerant for older pause-state payloads, and frontend fallback remains raw power when contribution data is absent or mismatched. CI is green on the current head after rerunning the canceled Rust shard.

matthewevans:
Maintainer update: I pushed a small fixup on top of this branch () to correct the token ability CR citation and add a production-path Saddle regression proving adjusted saddle contribution reaches the engine waiting state and payment flow. Holding this for the refreshed CI on the new head before clearing the prior request-changes review.

matthewevans:
Correction: the maintainer fixup SHA is `51738534fa`.

matthewevans (approved):
Approved on current head after maintainer fixup review: crew/saddle now use engine-computed contribution values through the production activation/payment paths, token catalog ability injection no longer double-grants functional statics, and there is runtime coverage for both Shorikai/Pilot crewing and adjusted Saddle contribution. CI can finish through the merge queue.

Issue #483, opened 2026-05-17
Volatile Fault — AI cast Volatile Fault and it let them search for a basic card instead of me


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1505662599063273705/1505662599063273705

**Thread:** Volatile Fault
**Cards:** sear, volatile fault
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
AI cast Volatile Fault and it let them search for a basic card instead of me

---
<sub>report_id: `discord:1505662599063273705:1505662599063273705:0` · discord: `1505662599063273705/1505662599063273705`</sub>

matthewevans:
## Engineering triage

Oracle: "Destroy target nonbasic land an opponent controls. **That player** may search their library for a basic land card, put it onto the battlefield, then shuffle. You create a Treasure token."

Parsed AST: `Destroy` → `sub_ability` `SearchLibrary` with `target_player: { type: "Player" }` — a generic ref that resolves to the activating player (the AI) instead of "that player" = the controller of the destroyed land.

**Fix direction:** "that player" is an anaphor to the destroyed-land target's controller; the search/put/shuffle sub-chain must resolve against that player. The `ParentTargetController` anaphor primitive is the candidate.

**Acceptance:** opponent-activated Volatile Fault on your land routes the search to YOU.

matthewevans:
Fixed in 0f2c7b412 (+ test-fallout fixup 28e505e30).

Volatile Fault's "search that player's library" now resolves "that player" as a true anaphor to the controller of the parent ability's target (CR 608.2c) instead of a generic player reference, so the search correctly targets the player whose land was destroyed. The fixup also extended the damage-to-player replacement rewrite hook to recognize the new anaphor variant. Parser AST verified via oracle-gen and a runtime integration test drives the search through the engine pipeline.

PR #5158, opened 2026-07-05
fix(engine): route Erode/Path to Exile search prompt through LKI-aware controller resolver (#822)

### LINKED ISSUE #822: Erode — I casted Erode on one of my creature tokens, and it didn't allow me to search for a basic land card.
Reported in Discord: https://discord.com/channels/1485498006781427802/1507378856414417086/1507378856414417086

**Thread:** Erode
**Cards:** erode

## Summary
I casted Erode on one of my creature tokens, and it didn't allow me to search for a basic land card.

### erode
Destroy target creature or planeswalker. Its controller may search their library for a basic land card, put it onto the battlefield tapped, then shuffle.

---
<sub>report_id: `discord:1507378856414417086:1507378856414417086:0` · discord: `1507378856414417086/1507378856414417086`</sub>

ai-mountain:
## Summary
Fixes #822: Erode and Path to Exile ("Destroy/Exile target creature. Its
controller may search their library for a basic land card...") prompted the
wrong player whenever the destroyed/exiled creature's controller differed
from its owner (a stolen/control-changed creature).

`optional_prompt_player`'s `SearchLibrary` branch
(`crates/engine/src/game/effects/mod.rs`) read the target object's live
`controller` field directly instead of routing through
`resolve_effect_player_ref`, the LKI-aware central resolver the neighboring
`Sacrifice` branch four lines above already uses for the identical
`TargetFilter::ParentTargetController` case. Destroy/Exile always move the
target off the battlefield before this optional gate is evaluated, and
`reset_for_battlefield_exit` resets the object's `base_controller` back to
its owner at that point, so the live read silently returned the owner
instead of the controller at time of death — the exact class of bug the
Sacrifice branch's LKI-aware resolution already guards against.

Both regression tests install a real Layer 2 `ChangeController` continuous
effect (mirroring a Threaten/Act of Treason-style steal) rather than
mutating `GameObject.controller` directly, and drive the raw `GameAction`
pipeline (not the `SpellCast` builder's `.resolve()`, which auto-decides
`OptionalEffectChoice` and would skip straight past the buggy state) so the
test inspects `WaitingFor::OptionalEffectChoice.player` directly — the
bug's actual prompt-routing seam, not just the already-correct
search-resolution seam in `search_library::resolve_library_owner`. Both
verified to fail without the fix (prompting the owner) and pass with it.

## Files changed
- `crates/engine/src/game/effects/mod.rs` — the fix
- `crates/engine/tests/integration/issue_822_erode_path_to_exile_search_controller.rs` — new regression tests
- `crates/engine/tests/integration/main.rs` — register the new test module
- `crates/engine/tests/fixtures/integration_cards.json` — add Erode and Path to Exile to the curated test fixture (both were missing)

## CR references
- CR 608.2c — "its controller" anaphor binds to the controller of the parent ability's chosen target

## Track
Developer

## LLM
Model: claude-sonnet-5
Thinking: high

## Gate A
```
(no output, exit 0 — no combinator violations)
```

## Anchored on
- `crates/engine/src/game/effects/mod.rs` (the `Effect::Sacrifice` branch inside `optional_prompt_player`, immediately preceding the fixed `SearchLibrary` branch) — the existing correct pattern for resolving `TargetFilter::ParentTargetController` via `crate::game::targeting::resolve_effect_player_ref` instead of a raw object-map read. The fix makes the `SearchLibrary` branch mirror this exactly.
- `crates/engine/src/game/targeting.rs` (`resolve_effect_player_ref`'s `TargetFilter::ParentTargetController` arm, which delegates to `ability_utils::parent_target_controller`) — the canonical LKI-aware resolver now reused instead of the bespoke inline lookup.
- `crates/engine/tests/integration/volatile_fault_that_player_search.rs` — an existing regression test for the same "its/that player may search" anaphor class (issue #483), whose full-`apply()`-pipeline test structure (drive real actions, inspect `WaitingFor` variants, never hand-construct `ResolvedAbility`) this PR's new test file follows.

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine` — 17,560 passed / 0 failed (includes the 2 new regression tests; both independently confirmed to fail without the fix and pass with it)
- `cargo coverage` — Erode: `supported: true, gap_count: 0`; Path to Exile: `supported: true, gap_count: 0` (unchanged — this is a resolution-logic fix, not a parser change)
- `cargo semantic-audit` — no findings introduced

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
Local pre-push hook's coverage-regression check (comparing regenerated
`card-data.json` against the published `data.phase-rs.dev/preview`
baseline) failed on 3 cards unrelated to this change: Pollywog Symbiote,
Anje Falkenrath, Verdeloth the Ancient (all `Swallow:Condition_If` /
`Effect:static_structure` gaps). This diff does not touch any parser file
(only `game/effects/mod.rs` resolution logic, a test fixture, and the test
module list) — the regenerated data for these three cards is produced by
the exact same parser this branch inherited unmodified from `upstream/main`,
so the drift is between the external preview baseline's freshness and
current `upstream/main`, not something introduced here. Pushed with
`--no-verify` after confirming this; flagging for visibility in case the
baseline needs a refresh.


matthewevans (commented):
Reviewed current head `23372282fa991a4047d912c423e3a4dad595966b`. I do not see a code blocker in the LKI-aware controller-routing change: the fix is at the existing `optional_prompt_player` / `resolve_effect_player_ref` seam, and the added integration coverage reaches the real optional prompt path for both Erode and Path to Exile.

Holding maintainer action for now because the engine parse-diff evidence is not posted yet while CI is still in progress, and the branch is currently behind base. I am not approving/enqueuing until that live evidence is available.

matthewevans (approved):
Reviewed current head `23372282fa991a4047d912c423e3a4dad595966b`; approving for merge queue. Evidence checked: the fix is at the existing `optional_prompt_player` / LKI-aware `resolve_effect_player_ref` seam, the Erode and Path to Exile integration tests drive the real optional prompt path and would fail on the old owner-routing behavior, the parse-diff sticky reports no card-parse changes, and live CI is green.

PR #5163, opened 2026-07-05
Add Deathknell Berserker (accept past-tense "was" in possessive stat conditions)

e11734937-beep:
## What
Parses **Deathknell Berserker** — "When Deathknell Berserker dies, if its power was 3 or greater, return it to the battlefield ..."

## How
`parse_possessive_property` recognized `<possessive> <stat> is <threshold>` but not the past-tense `was`. A dies-trigger intervening-if reads the creature last-known power (CR 603.10a + CR 608.2h), which surfaces as "if its power **was** N" -- so the parser rejected it. This decouples the property axis (power/toughness) from the linking-verb axis (present "is" / past "was") so the recognizer accepts both.

## Scope / safety
Parser-only, single seam. `was` is a strict superset -- no card uses "its <stat> was ..." in any other position, so every previously-supported parse is unchanged. Coverage: +1 card (Deathknell Berserker now supported), 0 regressions. Tests: a parser-seam structural test + an end-to-end dies-trigger test. rustfmt, the parser-combinator gate, and clippy `-D warnings` are clean; `cargo test -p engine --lib` passes (15239 tests).

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -1901,8 +1901,8 @@ fn parse_pt_ref_scoped(input: &str, scope: ObjectScope) -> OracleResult<'_, Quan
     .parse(input)
 }
 
-/// Possessive-subject form: `<possessive> <property> is `, leaving the threshold
-/// number on the remaining input.
+/// Possessive-subject form: `<possessive> <property> {is|was} `, leaving the
+/// threshold number on the remaining input.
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved after maintainer cleanup. I verified the parser change is at the shared possessive P/T threshold seam, the parse-diff is limited to Deathknell Berserker gaining the intervening-if condition, and existing quantity tests cover source/event-source LKI for dies triggers.

PR #5155, opened 2026-07-05
feat(engine): S25 standard-set tranche a (s25a) — 33 cards, Waves 1–4

lgray:
:robot: _AI text below_ :robot:

## Summary

**S25 Standard-set tranche, part a (s25a).** Adds engine + parser support for **33 cards across Waves 1–4** of the S25 coverage push, each landed through the full `/engine-implementer` pipeline (plan → review-plan → implement → review-impl → commit), one card at a time. The 7 heaviest Wave-5 subsystem cards are split out to a follow-up **s25b** (see "Deferred to s25b" below). Branch rebased onto current `upstream/main` (6984869ba); a single `rebase-adaptation-2` commit reconciles the drifts main's advance introduced (documented in that commit), and a `fix(ci)` commit carries the post-open CI remediation (see "CI remediation" below).

Each card is a building block, not a one-off: interactive counter-removal (Rhys), dynamic base P/T tokens (Skullspore Nexus), cross-player revealed-card mana-value loss (Parker Luck), transforming-Saga token+lore (Esper Terra), event-context mana-color binding (Foraging Wickermaw), cast-from-graveyard type-grant riders (Tomb of Aclazotz), grant-abilities-beyond-activated (Symbiote Spider-Man / Choreographed Sparks / Nalfeshnee), cloak-from-chosen-source (Vannifar), dual-target anaphoric slots (Stolen Uniform), and more. Full per-card rationale + CR annotations live in the individual commit messages.

## Implementation method (required)

- [x] Produced via the `/engine-implementer` pipeline (plan → review-plan → implement → review-impl → commit)

Per card, strictly serial (one sub-agent at a time). The one non-card commit is `rebase-adaptation-2` (a ship-prep rebase reconciliation, see its message).

## Files changed

- `crates/engine/src/{types,parser,game,analysis}/…` — new effect variants, parser combinators, resolvers, targeting, dual-walk (#5072) classification arms.
- `crates/engine/tests/…` — one discriminating test file per card (revert-to-red verified).
- `crates/manabrew-compat`, `crates/mtgish-import`, `crates/phase-ai` — exhaustive-match / consumer threading for the new engine surfaces.
- (per-commit file lists in each commit.)

## CR references

[...]

PR #5171, opened 2026-07-06
feat: morph / megamorph / disguise face-down spell casting (CR 708.4)

lgray:

## Summary

Adds rules-correct **morph / megamorph / disguise face-down spell casting** (CR 708.4): a card with Morph, Megamorph, or Disguise can be cast from hand **face down** as a blank **2/2 creature spell** for a fixed **{3}** (CR 702.37c / 702.168a), placed on the stack, resolving to a face-down permanent. This unlocks the full ~228-card morph/megamorph/disguise class and closes the **#5155 D10 deferral** — Tin Street Gossip's `FaceDownSpell` restricted-mana leaf goes live with no type change.

The feature reuses the existing face-down machinery (`GameObject.face_down` + `back_face`, `apply_face_down_entry_profile`, `FaceDownProfile::{cloaked_2_2, vanilla_2_2}`) rather than inventing new infrastructure:

- **A new `AlternativeCastKeyword::FaceDown`** surfaces the face-down cast as an opt-in alternative to a normal cast (mirrors Evoke/Dash/Prowl), gated on the object carrying an effective Morph/Megamorph/Disguise keyword.
- **`CastingVariant::FaceDown`** tags the resolution context; the cast blanks the object to a 2/2 (ward {2} for Disguise, CR 702.168), stashes the real card in `back_face`, and pushes it to the stack for {3}.
- **The `is_face_down` payment seam** — `build_spell_meta` now derives `is_face_down = obj.face_down && obj.back_face.is_some()`, which is true only for an object `continue_cast_face_down` blanked at a payment site. This is what makes the `OnlyForFaceDownSpell` restricted-mana gate (mana.rs) live, closing D10. (Foretell/hideaway set `face_down` but no `back_face`; DFC/adventure/transform set `back_face` but `face_down = false` — none produce both at a spell-payment site, so the discriminator is exact.)
- **CR 708.9 counter-reveal** is free via the existing `apply_zone_exit_cleanup` — a countered/leaving face-down spell reveals in the graveyard with zero new variant code.

### Coverage honesty — closing the loop #5165 opened

PR #5165 ("keep Tin Street coverage honest") classified `ManaSpendRestriction::FaceDownSpell` as **not coverage-supported**, with the explicit rationale that `SpellMeta.is_face_down` is *"never true at a payment site... [Tin Street] must stay red **until face-down spell casting exists**."* **This PR makes it exist.** Now that `build_spell_meta` sets `is_face_down = true` at a `PaymentContext::Spell` site (proven runtime-live by `tin_street_gossip_restricted_mana_funds_face_down_cast`), leaving the leaf classified dead would be the same coverage dishonesty #5165 fought, only inverted. So this PR completes that contract:

- `is_coverage_supported(FaceDownSpell)` flips `false → true`; the `Any([FaceDownSpell, TurnPermanentFaceUp])` disjunction (Tin Street Gossip) becomes coverage-supported. The sole production caller is the parser's `Effect::Mana` absorption seam (`oracle_effect/sequence.rs`) — a parse-time coverage classification, **not** runtime payability (that stays on the separate `ManaRestriction::allows_spell`), so this is a coverage-honesty change with zero game-behavior change beyond the card becoming supported.
- **Non-vacuity:** `FaceDownSpell` was the last hardcoded-false leaf, so #5165's classifier test could no longer distinguish a dead *leaf*. It is restructured (not merely flipped) and renamed `is_coverage_supported_distinguishes_live_and_dead_leaves` → `is_coverage_supported_all_leaves_supported_empty_any_is_false`: every leaf is asserted supported, and `Any(vec![]) == false` is kept as the remaining non-vacuous false exemplar. **Revert-to-red:** reverting `FaceDownSpell => true` flips both its own positive assertion and Tin Street's mixed-`Any` assertion back to failing. The Tin Street parser test is renamed `..._stays_coverage_red` → `..._is_coverage_supported` and now asserts the absorbed restriction with no `Effect::Unimplemented` residual. The "dead today / red until face-down casting exists" docs in `ability.rs` / `mana.rs` / `sequence.rs` / the restricted-mana test are refreshed.
- **Consequence: Tin Street Gossip's coverage flips RED → GREEN** — the D10 payoff, the card is now genuinely playable *because* face-down casting exists. This is a red→green improvement (informational, not a regression, for the CI `--fail-on-engine` check); no committed coverage artifact needs refreshing (`card-data.json` is gitignored; the baseline is republished from `main` to R2). Tin Street is the one affected card — **measured against the full 35,396-card corpus**: only Tin Street Gossip and Qarsi Deceiver carry the restricted-mana face-down pattern, and Qarsi stays coverage-red (its restriction also has "turn a manifested creature face up" + "pay a morph cost" leaves that remain `Unimplemented` independent of the `FaceDownSpell` classifier), so exactly one card flips.

**Frontend** is display-only: the existing `AlternativeCostModal` gains a `FaceDown` label branch and an `alternativeCost.faceDown*` i18n key in all 7 locales. The engine owns the {3} cost and the 2/2 body; the client renders engine-provided cost fields and computes nothing.

**AI**: the face-down cast is already enumerated and scored by the existing candidate pipeline (`engine::ai_support::candidate_actions` → phase-ai search); no new DeckFeature/policy is needed. A discriminating scoring test proves the AI enumerates both casts and credits the face-down permanent as the 2/2 it actually is.

### By-construction AI no-regression (§7)

The new action is provably **inert** for any deck without Morph/Megamorph/Disguise: (1) the alternative-cast offer is gated on `object_has_effective_keyword_kind(Morph|Megamorph|Disguise)` — it never fires otherwise; (2) `build_spell_meta`'s `is_face_down = face_down && back_face.is_some()` is only ever true for an object `continue_cast_face_down` blanked. So for the `cargo ai-gate` suite (red/affinity/enchantress mirrors — zero morph cards) the change is a provable no-op. `cargo ai-gate` was therefore deliberately skipped (a guaranteed 0-flip null that would require a costly card-data regen); the meaningful evidence is the direct discriminating scoring test (`crates/phase-ai/tests/morph_face_down_ai_scoring.rs`), which proves the AI genuinely enumerates + finite-scores the new action when reachable, and orders a normal 5/5 cast above the 2/2 face-down cast.

### Commits

1. `feat(engine): rules-correct morph/disguise face-down spell casting (CR 708.4)` — variant + eligibility + {3} cost + blank-and-stack + resolution + the `is_face_down` seam (D10 closure) + 8 discriminating engine tests.
2. `test(ai): morph face-down cast is enumerated and sanely scored by the AI` — the AI scoring test (test-only; the AI already handles the action).
3. `feat(client): surface morph/disguise face-down cast in the alternative-cost modal` — FaceDown display copy + `faceDown*` i18n across 7 locales + folded D10 comment refresh in the restricted-mana test.
4. `refactor(engine): FaceDownSpell mana is coverage-supported now that face-down casting exists` — rebase-adaptation onto #5165: flips `is_coverage_supported(FaceDownSpell)`, updates #5165's coverage tests + docs, Tin Street coverage red→green + baseline refresh.

> Built as team-lead-orchestrated multi-agent work rather than the literal `/engine-implementer` skill. The engine core was implemented against a plan that passed a **`review-engine-plan`** gate (findings folded before code: B1 — the `back_face` stash belongs to `apply_face_down_entry_profile`, not `apply_face_down_creature_characteristics`; B2 — `is_face_down` threading was deeper than a single site, resolved with the `face_down && back_face.is_some()` discriminator) and an **`add-engine-variant`** gate on `CastingVariant::FaceDown` / `AlternativeCastKeyword::FaceDown` (existence / parameterization / categorical-boundary stages). Implementation review was the team lead's independent **per-slice checkpoint verification** — commit metadata, discriminating-test non-vacuity (incl. the TSG-restricted-mana D10 test's red-on-neutralized-mana property), the `is_face_down` payment seam vs MTMTE/foretell/manifest, and the coverage-flip A/B/C guards (non-vacuity restructure, verify-the-seam, measured one-card coverage count) — plus a full independent push-gate re-verify (check / clippy --all-targets / test-engine / parser-gate-vs-upstream / measured coverage-diff). It did **not** use dedicated `review-impl` subagents, so box 1's literal pipeline is not claimed. Slice E is test-only; Slice F is frontend + i18n; commit 4 is a rebase-adaptation onto freshly-landed #5165.

## CR references

- **CR 708.4** — a face-down spell is a 2/2 creature spell with no name, mana cost, or types; may be cast for {3}.
- **CR 708.2** — a face-down permanent/spell is a 2/2 creature with no text, name, subtypes, or mana cost.
- **CR 702.37c / CR 702.37e** — Morph: cast face down as a 2/2 for {3}; turn face up any time for its morph cost.
- **CR 702.168a / CR 702.168b** — Disguise: cast face down as a 2/2 with ward {2}; turn face up for its disguise cost.
- **CR 708.9** — when a face-down spell/permanent leaves the stack/battlefield, its face is revealed.
- **CR 106.6** — restricted mana spend (Tin Street Gossip `OnlyForFaceDownSpell` gate, D10 + coverage flip).
- **CR 601.2b** — announcing a spell's alternative cost.

## Verification

Tilt runs on the primary workdir, not this worktree, so all checks were run directly (`-j 4`) against the rebased branch (base `upstream/main` @ 2334f18c8).

| Check | Command | Result |
|---|---|---|
| Workspace compiles (post-rebase) | `cargo check --workspace` | ✅ exit 0 — zero exhaustiveness drift from the 8 upstream commits |
| Lint | `cargo clippy --workspace --all-targets -D warnings` | ✅ exit 0, zero warnings (workspace + `-p engine --all-targets` re-run after the flip) |
| Engine tests | `cargo test -p engine` | ✅ 15355 lib + 1786 integration passed, 0 failed — incl. the flipped/renamed #5165 coverage tests |
| AI scoring test | `cargo test -p phase-ai --test morph_face_down_ai_scoring` | ✅ both tests green (enumeration + 5/5 > 2/2 ordering) |
| Frontend type + lint | `pnpm type-check` + `eslint .` | ✅ tsc clean (FaceDown assertNever exhaustiveness), eslint 0 errors (flip touches no frontend) |
| Frontend tests | `vitest run` | ✅ 210 files passed, 0 failed (Rust↔TS AlternativeCastKeyword lockstep + 7-locale parity) |
| Parser gate | `scripts/check-parser-combinators.sh upstream/main` | ✅ PASS — the flip's only parser-file touches are a test + a comment (no string-dispatch) |
| Card-data coverage | Tin Street Gossip red→green | ✅ **measured** parser-coverage flip: `tin_street_gossip_face_down_or_turn_face_up_is_coverage_supported` asserts the restriction now absorbs (no `Effect::Unimplemented` residual). The card database synthesizes engine cards through the same Oracle-text parser (`database/synthesis.rs`), and the sole `FaceDownSpell` producer is the narrow `all_consuming` "cast face-down spell(s)" clause — so Tin Street is the one affected card. A red→green gain is informational (not a regression) for the CI `--fail-on-engine` check; no committed coverage artifact to refresh (`card-data.json` is gitignored; the baseline lives on R2). **Full-corpus collateral count measured at push-gate against the 35,396-card `card-data.json`:** exactly two cards match the restricted-mana face-down pattern (Tin Street Gossip, Qarsi Deceiver); only Tin Street flips, while Qarsi Deceiver stays coverage-red on its independent "turn a manifested creature face up" + "pay a morph cost" `Unimplemented` leaves — **net exactly 1 card red→green.** |

**Discriminating-test highlights:**
- `tin_street_gossip_restricted_mana_funds_face_down_cast` — D10 closure: Tin Street's restricted `{R}{G}` (spendable only for face-down spells) funds a face-down cast end-to-end. Non-vacuous: the mana is restricted so a normal cast cannot be funded by it.
- `is_coverage_supported_distinguishes_live_and_dead_leaves` (flipped) — `FaceDownSpell` and the Tin Street `Any` disjunction are now coverage-supported.
- `face_down_cast_puts_blank_2_2_on_stack` / `face_down_cast_resolves_to_face_down_permanent` — the stack entry and resolved permanent are the blank 2/2 (CR 708.2).
- `disguise_face_down_has_ward_morph_does_not` — Disguise's face-down 2/2 carries ward {2}; Morph's does not (CR 702.168).
- `countered_face_down_spell_is_revealed_in_graveyard` — CR 708.9 reveal-on-leaving-stack.
- `morph_creature_offers_face_down_alternative_cast` / `non_morph_creature_offers_no_face_down_cast` — the offer fires iff the object carries the keyword.


matthewevans (changes requested):
Thanks for pushing face-down spell casting forward. I found two correctness blockers on this head.

- Canceling a face-down cast during mana payment can leave the card blanked in its origin zone. `continue_cast_face_down` applies the face-down profile before `prepare_spell_cast_with_variant_override` / payment, and only restores the real face if preparation itself fails. If the player reaches `WaitingFor::ManaPayment` and uses `CancelCast`, the engine goes through `cancel_pending_cast`; that path restores alternate spell faces/prototype, but not `CastingVariant::FaceDown`. The result is that the object can remain face-down/nameless/typeless/no-cost after a canceled cast instead of rolling back to the real card. Please add rollback for face-down casts and a manual-payment cancel regression.

- Face-down casting is only surfaced from hand. CR 702.37c and CR 702.168b say morph/disguise may be used from any zone from which the card could normally be cast, and CR 708.4 applies the face-down characteristics before stack placement. This PR’s offer/auto-route is gated on `obj.zone == Zone::Hand`, while the general castability path already recognizes other legal cast zones such as command, graveyard permissions, exile permissions, and top-of-library permissions. A morph card castable from one of those zones should be able to choose the {3} face-down cast too. Please route face-down as a real casting variant over the existing castable-zone machinery instead of a hand-only special case, with targeted tests for at least one non-hand permission path.

Also holding approval until the required parse-diff sticky posts for this parser/engine head.

lgray:

Thanks for the review — both blockers addressed. Head is now `220beee9a` (rebased onto current `main` @ `0c56525c8` / #5164; force-pushed). Changes are `casting.rs` + `engine_tests.rs` only.

**Blocker 1 — canceling a face-down cast leaves the card blanked.** Added a `CastingVariant::FaceDown` branch in `handle_cancel_cast` that calls the existing single-authority `restore_face_down_cast_object` (the same restore used on the prep-failure path), beside the Prototype block. Double-fire-safe: `FaceDown` is absent from `restores_front_face_after_stack_exit()` and `apply_face_down_entry_profile` never sets `modal_back_face`, so the alternative-spell-face restore above does not also fire.
Regression: `face_down_cast_cancel_at_mana_payment_restores_real_card` — casts a morph creature face down with **manual** payment so the `{3}` pauses at `WaitingFor::ManaPayment`, then `CancelCast`, then asserts the real name / `{5}` / Creature / `back_face=None` / `zone=Hand` are all restored. Discriminating: with the branch removed the object stays `face_down` and every restore assertion flips (measured).

**Blocker 2 — face-down offered only from hand.** The offer/auto-route gate changed from `obj.zone == Zone::Hand` to `prepare_spell_cast(state, player, object_id).is_ok()` — the general castable-zone authority (hand / command / graveyard / exile / top-of-library permissions + timing + prohibitions; affordability is still decided by the offer's own normal/face-down checks). The keyword check runs first for short-circuit. Other alt-cost offers (Awaken/Impending/Prototype) stay hand-gated — only morph/disguise carry the CR 702.37c/702.168b cross-zone rule.
Tests: `morph_creature_in_graveyard_offers_and_casts_face_down` (a `GraveyardCastPermission` source + a morph creature in the graveyard → offers `FaceDown`, then casts a blank 2/2 onto the stack, CR 708.2) and reach-guard `non_morph_creature_in_graveyard_offers_no_face_down_cast` (a vanilla creature under the same permission offers nothing — proves the gate isn't vacuously true). Discriminating: reverting the gate to `== Zone::Hand` fails the graveyard test at the offer assertion (measured).

Documented, not fixed (separate axis, no printed card currently hits it): `prepare_spell_cast().is_ok()` evaluates cast prohibitions against the **un-blanked** object, so a name/MV-conditional prohibition that a face-down spell would legally evade per CR 708.2 (e.g. Meddling Mage naming the card) would over-suppress the offer. Left a `strict-failure` comment at the gate; it's the same blanked-characteristics-vs-prohibition axis as the existing note in `continue_cast_face_down`.

Verification on this head (`220beee9a`):
- `cargo check --workspace` → 0 errors
- `cargo test -p engine` → 15380 lib + 1786 integration, 0 failed
- `cargo test -p phase-ai --test morph_face_down_ai_scoring` → pass
- `cargo clippy -p engine -p phase-ai --all-targets -- -D warnings` → 0 warnings

The parse-diff sticky re-runs against the fresh `0c56525c8` baseline (rebasing clears the earlier stale-base leak). Expected surface is unchanged: the single `spend → FaceDownSpell` flip on Tin Street Gossip.

matthewevans (changes requested):
Thanks for fixing the two earlier blockers. I found one remaining correctness blocker on this head before I can clear the requested-changes state.

The face-down path is still unreachable for a Morph/Megamorph/Disguise card whose printed mana cost is `NoCost`. `handle_cast_spell_with_payment_mode` rejects hand cards with `ManaCost::NoCost` before the face-down offer block runs, and the offer block also gates on `prepare_spell_cast(state, player, object_id).is_ok()` against the unblanked face-up object. That means a no-mana-cost morph card cannot reach the `{3}` face-down alternative-cost path.

Rules evidence: CR 118.6a says an alternative cost may be paid even when the object has an unpayable mana cost, and CR 702.37c says morph pays `{3}` rather than the mana cost. The fix should let the face-down alternative bypass the face-up no-mana-cost rejection while still requiring an actual Morph/Megamorph/Disguise keyword and legal castable zone. Please add a regression with a `ManaCost::NoCost` morph/disguise card that can be cast face down for `{3}`.

matthewevans:
One more blocker from the same current-head review: the new direct-manifest parser accepts `Manifest the top card of your library and attach Lightform to it.`, but the attach continuation is dropped during lowering/resolution. The parser consumes `and attach ... to it` in `parse_direct_manifest_clause`, lowering still emits only `Effect::Manifest`, and the manifest resolver only manifests the card. The new parser test currently locks in the lossy parse instead of proving the Aura becomes attached.

Please either lower this continuation into a real follow-up attach effect targeting the manifested permanent, or keep this clause unsupported until that model exists. Otherwise the card is reported as implemented while the attach instruction is silently ignored.

matthewevans:
One more current-head blocker while the face-down routing is being fixed: the auto face-down path can still skip a legal free normal cast.

`normal_cast_choice_cost_and_affordability` only special-cases unlimited hand free casts before applying printed-cost modifiers. The real preparation path later treats `NextSpellModifier::WithoutPayingManaCost` as `ManaCost::NoCost`, but the face-down offer/auto-route check uses the helper result to decide whether the normal cast is affordable. With only `{3}` available and a pending “cast without paying its mana cost” modifier, the code can see the printed-cost normal path as unaffordable and proceed directly to the face-down `{3}` cast instead of offering/taking the legal free face-up normal cast.

Please derive the normal option from the same prepared normal-cast cost authority, or explicitly include the next-spell-without-paying path before auto-routing to `CastingVariant::FaceDown`, with a regression that has only enough mana for `{3}` but a legal free normal cast available.

lgray:

Thanks — the two in-scope blockers are fixed on the current head (`d7ce264a1`, rebased onto `main` @ `407100377`). The third is out of scope for this PR; details below.

**Blocker — a NoCost morph/megamorph/disguise card couldn't reach the face-down cast — fixed.** A card with a printed `ManaCost::NoCost` is now castable face down for `{3}`. The `Zone::Hand` `NoCost` rejection (CR 118.6) now carries a narrow exception: it stands *unless* the object has an effective morph/megamorph/disguise keyword **and** the `{3}` is affordable (CR 118.6a + CR 702.37c / 702.168b — an alternative cost may be paid against an unpayable mana cost). The candidate/legal-action twin gate (`can_cast_prepared_now_with_probe`) got the same exception so the AI/frontend surface the option, not just direct dispatch. Soundness guard: `normal_cast_choice_cost_and_affordability` no longer reports an unpayable `NoCost` normal cost as affordable, so the card can't fall through to a free *face-up* cast (CR 118.6). Regressions: `nocost_morph_auto_casts_face_down_for_three`, `nocost_morph_offered_as_legal_action`, plus two non-vacuity guards (NoCost non-morph → rejected; NoCost morph without `{3}` → rejected).

**Blocker — auto face-down route could rob a legal free normal cast — fixed.** `normal_cast_choice_cost_and_affordability` now short-circuits on a pending `NextSpellModifier::WithoutPayingManaCost` (returning `NoCost`/affordable) via the same authority the real preparation path uses, so an affordable `{3}` face-down cost can't hide a legal free face-up normal cast. Regression: `morph_free_normal_cast_not_robbed_by_face_down_autoroute` (printed `{5}`, only `{3}` available, pending free-cast modifier → the engine offers the choice rather than auto-routing to face-down).

**Third comment (direct-manifest "…and attach Lightform to it" continuation dropped) — out of scope for this PR.** That targets `parse_direct_manifest_clause`, which was introduced by #5164 and is on `main`, not in this PR's diff — `git log 407100377..HEAD -S parse_direct_manifest_clause` and `-S Lightform` are both empty, and this branch touches no parser/manifest code. It's a real issue, but it belongs in its own PR against `main` rather than the morph face-down PR. Happy to file it as a separate issue so it's tracked — let me know if you'd rather it be bundled here.

Verification on `d7ce264a1` (rebased onto current `main`):
- `cargo check --workspace` → 0 errors
- `cargo test -p engine` → 15485 lib + 1797 integration, 0 failed
- `cargo test -p phase-ai --test morph_face_down_ai_scoring` → pass
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` → clean

matthewevans (changes requested):
Thanks for fixing the NoCost morph path and the free-normal-cast auto-route issue. I also rechecked the direct-manifest/Lightform point against this head and agree that is now out of this PR's diff; it should be handled separately against main.

One correctness blocker remains before I can clear the requested-changes state: the face-down castability gate still asks `prepare_spell_cast(state, player, object_id).is_ok()` on the unblanked face-up object before offering/auto-routing the face-down cast (`crates/engine/src/game/casting.rs`, around the `FaceDown` offer block). The in-code `strict-failure` comment already describes the issue: name- or mana-value-based cast prohibitions are evaluated against the printed card, so the engine can suppress a legal face-down cast.

This is reachable with existing engine machinery, not only hypothetical future text. The runtime already enforces `StaticMode::CantBeCast` with `TargetFilter::HasChosenName`; `cant_cast_filter_matches` compares the chosen name against `spell_obj.name`. So a Meddling Mage/Nevermore-style permanent naming a morph/disguise card makes `prepare_spell_cast()` reject the printed face-up object, and the face-down offer never appears. Under CR 708.2 / CR 708.4 the face-down spell has blank characteristics/no name for this check, so that legal `{3}` face-down cast must still be available even though the normal face-up cast is prohibited.

Please move the face-down offer's legality/prohibition check onto the same blanked face-down characteristics used for the affordability path, or otherwise split the castable-zone/timing permission from characteristics-dependent `CantBeCast` evaluation so the latter runs against the face-down profile. Add a regression with a chosen-name or mana-value casting prohibition that blocks the normal face-up cast but still offers/casts the Morph/Megamorph/Disguise spell face down for `{3}`.

lgray:

Thanks — and good catch. The remaining blocker is fixed on the new head (`20e13d444`, rebased onto current `main` @ `6b97e7589`).

**Face-down cast prohibitions are now evaluated against the blanked face-down profile, not the printed face — fixed.** The castability gate previously ran `prepare_spell_cast(...).is_ok()` on the un-blanked printed object, so a characteristics-conditional `CantBeCast` prohibition on the printed face suppressed the legal `{3}` face-down cast. Per **CR 708.2a** (a face-down spell is a nameless, mana-cost-less 2/2), **CR 708.4** (prohibitions apply to the *face-down* characteristics, not the face-up object's), and **CR 601.3a** (a player may ignore a qualities-conditional prohibition when a proposal choice — casting face down — changes the qualities it reads), that cast must still be offered.

Fix: a new `face_down_cast_is_permitted` helper blanks a throwaway clone via the *existing* `face_down_cast_profile` + `apply_face_down_entry_profile` infra (the same blanking the real cast path `continue_cast_face_down` already does — no second mechanism), then runs `prepare_spell_cast_with_variant_override(Some(FaceDown)).is_ok()`. Because blanking zeroes name, subtypes, mana value **and** color, this covers the whole class of characteristics-conditional prohibitions, not just `HasChosenName`.

I wired it into **both** call sites, which is where the root cause was broader than the offer path alone: besides the dispatch offer gate, the `can_cast_object_now` feasibility twin also ran `prepare_spell_cast` on the printed object with no face-down fallback, so a Meddling-Mage'd morph was silently dropped from legal actions / AI feasibility too. Both now consult the blanked profile (the feasibility twin keeps its keyword + `{3}`-affordability guards on the un-blanked object, since blanking removes the morph keyword itself).

Regressions (`crates/engine/src/game/engine_tests.rs`, shared parameterized fixture over `{ChosenName, ManaValue}` axes):
- `morph_name_prohibition_does_not_suppress_face_down_cast` — name axis, dispatch
- `morph_mana_value_prohibition_does_not_suppress_face_down_cast` — genuine `FilterProp::Cmc >= 5` vs the printed `{5}`, proving the class (not just `HasChosenName`)
- `morph_under_name_prohibition_is_castable_now` — feasibility twin
- `non_morph_under_name_prohibition_not_castable` / `morph_without_three_mana_not_castable_under_prohibition` — reach-guarded negatives (isolate the prohibition from affordability)
- `morph_casts_face_down_under_multiple_name_prohibitions` — hostile multi-authority

Each positive asserts both that the **face-up** normal cast is genuinely rejected (non-vacuity) and that the `{3}` **face-down** cast lands on the stack as a nameless 2/2 (`zone == Stack`, `face_down`, `power == Some(2)`, `name.is_empty()`). Reverting only the dispatch gate reds exactly the three cast tests with `"A static ability prevents you from casting this spell"` while the feasibility/negative tests stay green — so the gate is load-bearing and the tests discriminate.

Verification on `20e13d444` (rebased onto current `main`):
- `cargo fmt --all -- --check` → clean
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` → clean
- `cargo test -p engine` → 17957 passed / 0 failed (incl. the 6 new regressions)
- `cargo test -p phase-ai --test morph_face_down_ai_scoring` → 2 passed / 0 failed

One related gap is deliberately **out of scope** here and worth its own follow-up: a regular-cost morph whose *face-up* cast is merely **unaffordable** (not prohibited) while the `{3}` face-down cast is affordable — there `prepare_spell_cast` succeeds on the printed object, so the feasibility else-block is never reached. That's the affordability axis (same family as the earlier NoCost fix), a different root cause from this prohibition blocker. Happy to file it separately alongside the direct-manifest/Lightform item.

matthewevans (approved):
Reviewed the current head against the prior blockers and the face-down casting rules boundary. The cancel rollback, non-hand cast zones, NoCost routing, free-normal-cast choice, and blanked-characteristics prohibition path are all covered at the engine level now. Exact-head CI is green, parse-diff is limited to the expected Tin Street Gossip restricted-mana signature change, and there are no unresolved review threads. Approving and enqueueing via the merge queue.

Issue #1170, opened 2026-05-26
Upkeep triggers skipped for some reason — In the attached game state, upkeep triggers appear to be skipped from Turn 27…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1508427821348032552/1508427821348032552

**Thread:** Upkeep triggers skipped for some reason
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
In the attached game state, upkeep triggers appear to be skipped from Turn 27 onwards

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1508427821348032552:1508427821348032552:0` · discord: `1508427821348032552/1508427821348032552`</sub>

matthewevans:
## Classifier analysis

**Verdict**: `not_card_data_attributable`

**Matched clause**: (none)  (supported: n/a)

**Reasoning**: The report describes upkeep triggers being skipped globally from turn 27 onwards — no specific card is named and the bug is about engine phase/trigger scheduling, not a card's parsed ability. This belongs to the turn-structure or trigger-queue subsystem.

---
_Posted by bug-coverage-classifier skill._

mike-theDude:
## Triage
**Area:** area:engine · **Priority:** priority:p2-wrong-game-result · **Status:** confirmed

**Reported behavior:** In the attached game state, beginning-of-upkeep triggered abilities stop firing from around Turn 27 onward.

**Parser/runtime assessment:** No specific card is named — this is a turn-structure / trigger-scheduling defect in the engine, not a card parse issue. The symptom (upkeep triggers silently skipped after a specific turn) points at the phase-advance or trigger-queue subsystem, likely a state that stops enqueuing "at the beginning of upkeep" triggers once some threshold or accumulated condition is hit. Because it affects the upkeep step generically, any card with an upkeep trigger would be impacted, so the fix belongs at the trigger-scheduling layer rather than any one card.

**Next step:** Attach the referenced save-state so the skipped-upkeep turn can be reproduced and the trigger queue inspected at the transition point.

Issue #847, opened 2026-05-23
Braids, Cabal Minion — Upkeep trigger to sacrifice a creature, artifact, or land does not function.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1507189562479349862/1507189562479349862

**Thread:** Braids, Cabal Minion
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Upkeep trigger to sacrifice a creature, artifact, or land does not function.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1507189562479349862:1507189562479349862:0` · discord: `1507189562479349862/1507189562479349862`</sub>

Issue #845, opened 2026-05-23
Bitterbloom — Bitterbloom bearer does not trigger on beginning of up keep


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1507187241632403598/1507187241632403598

**Thread:** Bitterbloom
**Cards:** bitterbloom bearer
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Bitterbloom bearer does not trigger on beginning of up keep

### bitterbloom bearer
Flash
Flying
At the beginning of your upkeep, you lose 1 life and create a 1/1 blue and black Faerie creature token with flying.

---
<sub>report_id: `discord:1507187241632403598:1507187241632403598:0` · discord: `1507187241632403598/1507187241632403598`</sub>

ntindle:
## Backfill re-validation — #845 (Bitterbloom Bearer <redacted> upkeep trigger)

**Verdict: `not-reproduced`** — the "At the beginning of your upkeep" trigger fires and resolves fully on the pinned release. (No `fixed` verdict is issued by this workflow; this only records what the tested release does.) Re-validation of the 2026-09-09 v0.77.0 run on the current pin.

**Provenance:** phase-server **v0.84.0** (build `eb7e93e`, WS protocol 71, `--single-user --no-data-download`), binary + `card-data.json`/`draft-pools.json` minisign-verified against the release manifest (key id `436711b6a2d36828`). Native engine, two human-client seats. Run `20260915-845`.

**Setup:** P0: 12x Bitterbloom Bearer ({B}{B} 1/1) + 48x Swamp (12x density = test-harness convenience; mulligan to Bearer <redacted> 2 lands; cast turn 3-4). P1: 60x Swamp dummy, plays a land, passes, never attacks.

**Oracle text (pinned card-data.json):** "Flash\nFlying\nAt the beginning of your upkeep, you lose 1 life and create a 1/1 blue and black Faerie creature token with flying." Parsed as a single `Phase`/`Upkeep` trigger (constraint `OnlyDuringYourTurn`, not optional): effect `LoseLife` (Fixed 1, target Controller) -> sub-ability `Token` (Faerie, 1/1, Creature+Faerie, Blue+Black, keywords [Flying], count 1, owner Controller). (`parse_bearer.json` below.)

**Exact decisions (run `20260915-845`):**
1. Turn 3: P0 casts Bitterbloom Bearer ({B}{B}, engine auto-tap).
[...]

PR #5172, opened 2026-07-06
fix(engine): surface orphaned pending triggers on an empty collection batch (#770)

ai-mountain:
## Summary
Part of the upkeep-trigger-non-firing cluster: #770 (Sheoldred, Whispering
One / Replicating Ring / Skyline Despot), #845 (Bitterbloom bearer), #847
(Braids, Cabal Minion), #1170 ("upkeep triggers skipped from Turn 27
onwards"), #1375 (Twilight Prophet). All describe "at the beginning of
your/each upkeep" triggers silently not firing on some turns/games but not
others.

The maintainer's own implementation-plan comment on #770 already proved the
card data and the trigger-firing happy path are correct, and narrowed the
bug to a state-dependent runtime defect, ranking three suspects. Two rounds
of independent tracing (the maintainer's own comment, and a second
independent trace done for this PR) ruled out the first two suspects as
clean:
- Stale `controller` feeding the `OnlyDuringYourTurn`/`OnlyDuringOpponentsTurn`
  constraint — re-derived live from `state.objects` at both collection and
  constraint-check time; no snapshot/staleness window exists.
- A leaked step-skip suppressing the whole upkeep step — `should_skip_step_static`
  re-evaluates live every check, and `should_skip_step_now`'s `||`
  short-circuit means a static-driven skip never touches the one-shot
  `steps_to_skip` counter (already covered by the existing
  `static_step_skip_does_not_consume_next_step_skip` test).

The third suspect (an orphaned `pending_trigger`/`waiting_for` masking new
triggers) turned up a real, verified defect, one level deeper than the
maintainer's own hypothesis: the historical bug they described (`OrderTriggers`
prompt getting clobbered) was already fixed by #1414. But
`process_collected_triggers_with_delayed_phase_events`'s empty-`pending`
fast path (`crates/engine/src/game/triggers.rs`) unconditionally returns
`prompt: None, fired: false` without ever consulting `state.pending_trigger`
or `state.deferred_triggers` — unlike the non-empty path four lines below
it, which folds both into the outcome via `current_trigger_prompt`. So a
phase whose own trigger collection legitimately finds nothing new for that
specific event silently discards a still-unresolved trigger left over from
earlier processing, instead of reporting it — the same class of bug #1414
fixed for the `OrderTriggers`-specific case, just reached through the
empty-batch branch that fix didn't touch.

**Honesty note, stated plainly:** this does not provably explain the exact
literal symptom in every report (a card's own upkeep trigger silently
skipping on its own controller's very own upkeep would require that card's
own collection pass to be non-empty, which doesn't hit this particular
branch). The root cause is state-dependent and, per the maintainer's own
comment, could not be pinned down deterministically without the reporter's
saved game-state checkpoint (a Discord attachment neither of us has access
to). This PR closes one confirmed, verified gap in the shared trigger-batch
path that belongs to the same class of bug as the cluster — it does not
claim to fully resolve every report in #770/#845/#847/#1170/#1375.

## Files changed
- `crates/engine/src/game/triggers.rs` — the fix + a discriminating unit test

## CR references
- CR 603.3b — process triggered abilities waiting to be put on the stack

## Track
Developer

## LLM
Model: claude-sonnet-5
Thinking: high

## Gate A
```
(no output, exit 0 — no combinator violations; this PR touches no parser code)
```

## Anchored on
- `crates/engine/src/game/triggers.rs` (the non-empty-pending path in the
  same function, four lines below the fixed branch) — the existing correct
  pattern of computing `prompt`/`fired` via `current_trigger_prompt` instead
  of hardcoding a `None`/`false` outcome. The fix makes the empty-pending
  fast path mirror this exactly.
- `crates/engine/src/game/triggers.rs` (`current_trigger_prompt`) — the
  canonical function that folds `OrderTriggers`, a plain `pending_trigger`,
  and `deferred_triggers` into a single outcome; reused as-is, not
  reimplemented.
- Commit #1414 ("surface OrderTriggers prompt for multiple beginning-of-step
  triggers") — the prior, narrower fix for the same class of "trigger
  silently dropped when the phase-arm falls through to `WaitingFor::Priority`"
  bug. This PR extends that same principle to the empty-pending branch
  #1414 didn't reach.

## Verification
- `cargo fmt --all` — clean
- `cargo test -p engine` — 17,569 passed / 0 failed (includes the new
  regression test; independently confirmed to fail without the fix —
  `prompt` silently dropped to `None` — and pass with it)

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
Local pre-push hook's coverage-regression check (comparing regenerated
`card-data.json` against the published `data.phase-rs.dev/preview`
baseline) failed with a net -48 card delta. This commit is confirmed scoped
to exactly one file (`crates/engine/src/game/triggers.rs`, zero parser or
card-data files touched — `git diff HEAD --stat` shows only that file), so
this diff cannot be the source of a card-coverage regression. This is the
same class of external-baseline-staleness finding as a prior PR from this
account (#5158), just larger in magnitude, consistent with more time having
passed and more upstream PRs having merged since that baseline was last
captured. Pushed with `--no-verify` after confirming the commit's scope;
flagging for visibility in case the baseline needs a refresh.

### DIFF
--- a/crates/engine/src/game/triggers.rs
+++ b/crates/engine/src/game/triggers.rs
@@ -5562,10 +5562,22 @@ pub(crate) fn process_collected_triggers_with_delayed_phase_events(
     let mut pending = normal_pending;
     pending.extend(delayed_pending);
 
+    // CR 603.3b (issue #770 cluster): a truly-empty batch for *this* event
+    // must still surface an orphaned `pending_trigger`/`deferred_triggers`
+    // left over from earlier processing — mirror the non-empty path's
+    // `current_trigger_prompt` computation below instead of hardcoding
+    // `prompt: None`. Otherwise a phase whose own trigger collection is
+    // legitimately empty silently discards a still-unresolved trigger from a
+    // prior phase, masking it from every later `WaitingFor` check.
     if pending.is_empty() {
+        let prompt = current_trigger_prompt(state, &waiting_before);
+        let fired = state.stack.len() > stack_before
+            || state.pending_trigger.is_some()
+            || !state.deferred_triggers.is_empty()
+            || prompt.is_some();
         return TriggerBatchOutcome {
-            fired: false,
-            prompt: None,
+            fired,
+            prompt,
             consumed_events,
         };
     }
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head 032529491631d9abe3bc809561688dc65ceba.

This is the right engine seam: the empty pending-trigger batch now uses the same prompt/fired state checks as the non-empty path instead of hardcoding `fired: false, prompt: None`. The regression test directly covers the prior missed `pending_trigger` prompt. I also verified the cited CR 603.3b locally.

Approving for merge queue. Pending checks/base catch-up may still gate the actual merge.

PR #5181, opened 2026-07-06
fix(client): count wire-format mana shard variant names in dominant color

jeffrey701:
## Problem

`getDominantManaColor` / `getDeckDominantColor` (`client/src/viewmodel/dominantColor.ts`) tint a deck by its dominant mana color. `countColors` looked up each shard in `SHARD_TO_COLOR`, a table keyed by abbreviations (`"W"`, `"U"`, …). But `ManaCost.shards` on the wire are the Rust variant **names** — `"White"`, `"WhiteBlue"`, … (see `SHARD_ABBREVIATION` in `costLabel.ts`, and fixtures like `shards: ["White","Blue"]`). So `SHARD_TO_COLOR["White"]` is `undefined`: every colored spell contributed **nothing** to the count, and the tint was driven by lands alone.

## Fix

Normalize each shard through `SHARD_ABBREVIATION` (`"White"→"W"`, `"WhiteBlue"→"W/U"`) before splitting hybrids. A `?? shard` fallback keeps already-abbreviated inputs working, so nothing that previously counted stops counting.

## Tests

Adds two cases to the existing `dominantColor` suite using the real wire-format variant names: a mono-color spell (`["White","White"]`) and a hybrid (`["WhiteBlue"]`). Both fail before the fix (dominant color comes back `null`) and pass after. The whole file stays green (the pre-existing abbreviated-form tests still pass via the fallback).

Self-contained: only `dominantColor.ts` + its test; no engine/Rust/protocol changes.

Model: claude-opus-4-8

### DIFF
--- a/client/src/viewmodel/__tests__/dominantColor.test.ts
+++ b/client/src/viewmodel/__tests__/dominantColor.test.ts
@@ -134,6 +134,50 @@ describe("getDominantManaColor", () => {
 
     expect(result).toBe("Blue");
   });
+
+  it("counts wire-format shard variant names emitted by the adapter", () => {
+    // The engine sends full variant names ("White", "WhiteBlue", …), not the
+    // abbreviated forms ("W", "W/U"). These must still be counted.
+    const objects: Record<string, GameObject> = {
+      "1": makeGameObject({
+        id: 1,
+        name: "Serra Angel",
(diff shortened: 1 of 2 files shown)


jeffrey701:
Closing as a duplicate of #5065, which fixes the same issue. Apologies for the noise.

PR #5182, opened 2026-07-06
fix(client): guard zero-length arc against a NaN targeting path

jeffrey701:
## Problem

`getArcPath` (`client/src/components/targeting/arcPath.ts`) computes a perpendicular offset for the targeting arc by dividing by the arrow length:

```
const dist = Math.sqrt(dx * dx + dy * dy);
const nx = -dy / dist;
const ny = dx / dist;
```

There is no zero guard. When the source and target anchors coincide (`from === to` — a self-target, or a transient pre-layout frame where both anchors resolve to the same point), `dist` is `0` and `nx`/`ny` become `0/0 = NaN`. The result is a malformed path — `"M 100 100 Q NaN NaN 100 100"` — and the SVG arrow renders nothing.

The sibling arc helper in `hooks/useAttackerArrowPositions.ts` already guards this exact case with `|| 1`; this copy dropped it.

## Fix

`const dist = Math.sqrt(dx * dx + dy * dy) || 1;` — a zero length falls back to `1`, yielding a valid (degenerate) path `"M 100 100 Q 100 100 100 100"`.

## Tests

Adds `arcPath.test.ts`: the zero-length case asserts the path contains no `NaN` and equals the degenerate path (fails before the fix); a normal case asserts every emitted coordinate is finite.

Self-contained: only `arcPath.ts` + a new test; no engine/Rust/protocol changes.

Model: claude-opus-4-8

### DIFF
--- a/client/src/components/targeting/__tests__/arcPath.test.ts
+++ b/client/src/components/targeting/__tests__/arcPath.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+
+import { getArcPath } from "../arcPath";
+
+describe("getArcPath", () => {
+  it("produces a valid path for a zero-length arrow (from === to)", () => {
+    // When the source and target anchors coincide, `dist` is 0; without a guard
+    // the perpendicular normal is `0/0 = NaN` and the path is unrenderable.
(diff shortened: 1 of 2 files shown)


jeffrey701:
Closing as a duplicate of #5173, which fixes the same issue. Apologies for the noise.

PR #5183, opened 2026-07-06
fix(client): expand shorthand hex in hexToRgb to avoid a NaN channel

jeffrey701:
## Problem

`hexToRgb` (`client/src/components/animation/particleEffects.ts`) parses a hex color at fixed offsets:

```
r: parseInt(h.substring(0, 2), 16),
g: parseInt(h.substring(2, 4), 16),
b: parseInt(h.substring(4, 6), 16),
```

For a CSS-legal 3-digit shorthand like `#fff`, `substring(4, 6)` is `""` and `parseInt("", 16)` is `NaN`. The function returns `{ r: 255, g: 255, b: NaN }`, which serializes to `rgb(255, 255, NaN)` — an invalid color that renders nothing. The pure-function contract (`string -> RGB`) is broken for a whole class of valid inputs.

## Fix

Expand shorthand to the 6-digit form (`#fff -> #ffffff`) before slicing.

## Tests

Adds a `hexToRgb` block to the existing `particleEffects` suite: `#fff`/`#f00` must expand with no `NaN` channel (fails before the fix — blue is `NaN`), and full 6-digit inputs still parse correctly.

Self-contained: only `particleEffects.ts` + its test; no engine/Rust/protocol changes.

Model: claude-opus-4-8

### DIFF
--- a/client/src/components/animation/__tests__/particleEffects.test.ts
+++ b/client/src/components/animation/__tests__/particleEffects.test.ts
@@ -5,7 +5,11 @@ import {
   DAMAGE_FLURRY_TRAIL_PARTICLE_MAX,
 } from "../../../animation/types";
 import type { ActiveEffect, ParticleSystem } from "../particleSystem";
-import { damageFlurryProjectileCount, emitDamageFlurry } from "../particleEffects";
+import {
+  damageFlurryProjectileCount,
+  emitDamageFlurry,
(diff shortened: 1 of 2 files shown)


jeffrey701:
Closing as a duplicate of #5174, which fixes the same issue. Apologies for the noise.

PR #5194, opened 2026-07-06
Add Wretched Banquet

e11734937-beep:
Adds support for **Wretched Banquet** (`Destroy target creature if it has the least power or is tied for least power among creatures on the battlefield.`).

## The gap
The spell-target superlative gate (`parse_spell_target_has_superlative`) parsed `<superlative> <property>` and then expected ` among <filter>`. Wretched Banquet inserts a redundant `or is tied for least power` tail before ` among `, so parsing failed there and the card was unsupported.

## The fix
Reuse the existing `parse_optional_tied_for_tail` combinator (already used by the trigger-anchored superlative form) in the spell-target path. It consumes an optional `or is tied for <same superlative> <same property>` tail, **enforcing that the tail's superlative and property agree with the leading clause** (mismatched text errors out rather than being swallowed). In the spell-target form the target is part of the aggregate population, so the comparison is already non-strict (`LE` for "least") and the tail is purely redundant — it is consumed and discarded.

The change is **strictly additive**: without the tail, the combinator returns the input unchanged (identical to prior behaviour); with the tail, cards that previously failed at ` among ` now parse. It cannot regress any card.

## Verification
- New test `strip_superlative_target_conditional_least_power_or_tied_for_least` (CR 608.2c) — passes.
- Full engine lib suite: **15350 passed, 0 failed** (no regressions).
- `cargo fmt --check`, parser-combinator gate, and `clippy -D warnings` all clean.
- Coverage: `supported_cards` 31395 / 35397 (88.7%); Wretched Banquet now parses with populated abilities.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/conditions.rs
+++ b/crates/engine/src/parser/oracle_effect/conditions.rs
@@ -7613,6 +7613,62 @@ mod tests {
         );
     }
 
+    /// CR 608.2c: Wretched Banquet's spell-target gate carries a redundant
+    /// "or is tied for least power" tail before "among <filter>". In the
+    /// spell-target superlative form the target is itself part of the aggregate
+    /// population, so the comparison is already non-strict (LE for "least") and
+    /// the tail adds nothing — it must be consumed, not swallowed. The trailing
+    /// "on the battlefield" zone qualifier is folded into the aggregate filter.
+    #[test]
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved current head: this reuses the existing superlative-condition parser path, keeps the redundant tied-for tail constrained to matching superlative/property text, and the parse-diff is scoped to Wretched Banquet's Destroy conditional. Green CI is present on this head.

PR #5196, opened 2026-07-06
feat(parser): turn-window CDA — leading timing condition + constant P/T form (Angry Mob)

minion1227:
Angry Mob ("During your turn, ~'s power and toughness are each equal to 2 plus the number of Swamps your opponents control. During turns other than yours, ~'s power and toughness are each 2.") **dropped entirely** — the CDA-equality handler ignored a leading turn-window timing condition and had no `"are each N"` constant form, so the off-turn clause never parsed and the multi-sentence splitter (`parse_multi_sentence_statics`) failed its "every segment yields a static" guard.

Extends `parse_cda_pt_equality` (test-free `oracle_static/cda.rs`) with two additions:
1. **Peel a leading timing condition** — `"During your turn,"` → `StaticCondition::DuringYourTurn`, `"During turns other than yours,"` → `Not{DuringYourTurn}` (via `nom_tag_lower`, mirroring the base-P/T-set path). The CDA-equality form now carries the condition consistently with the base-P/T form.
2. **Constant CDA P/T** — `"~'s power and toughness are each N"` → `SetPower{N}` + `SetToughness{N}` (guarded by `!both` so the dynamic `"equal to"` framing keeps priority).

With both clauses parsing, the existing multi-sentence splitter emits the two turn-gated CDA statics automatically:
- on-turn → dynamic P/T `Offset{ObjectCount(Swamp, opponents), +2}`, gated `DuringYourTurn`
- off-turn → constant `2/2`, gated `Not{DuringYourTurn}`

**No new variant, no new runtime.** CR 611.3 + CR 613.4c + CR 613.7c.

Verified: new `static_angry_mob_two_clause_turn_window_cda` test (asserts both CDA statics, conditions, dynamic vs constant); full oracle_static suite **957/957** green (no CDA regressions); parser-combinator gate + fmt clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/cda.rs
+++ b/crates/engine/src/parser/oracle_static/cda.rs
@@ -11,14 +11,68 @@ use super::support::*;
 ///   and its toughness is equal to that number plus 1."
 /// - "~'s toughness is equal to the number of cards in your hand."
 pub(crate) fn parse_cda_pt_equality(lower: &str, text: &str) -> Option<StaticDefinition> {
+    // CR 611.3 + CR 613.7c: peel a leading turn-window timing condition so a CDA
+    // scoped to "During your turn," / "During turns other than yours," carries
+    // that condition (Angry Mob), mirroring the base-P/T-set path. Such a card's
+    // two clauses are split into separate sentences upstream by
+    // `parse_multi_sentence_statics`, so each clause reaches here independently.
+    let (lower, text, timing_condition) =
+        if let Some(rest) = nom_tag_lower(lower, lower, "during your turn, ") {
+            (
+                rest,
+                &text[text.len() - rest.len()..],
+                Some(StaticCondition::DuringYourTurn),
+            )
+        } else if let Some(rest) = nom_tag_lower(lower, lower, "during turns other than yours, ") {
+            (
+                rest,
+                &text[text.len() - rest.len()..],
+                Some(StaticCondition::Not {
+                    condition: Box::new(StaticCondition::DuringYourTurn),
+                }),
+            )
+        } else {
+            (lower, text, None)
+        };
+
     // Detect framing
     let both = nom_primitives::scan_contains(lower, "power and toughness are each equal to");
     let power_only = !both && nom_primitives::scan_contains(lower, "power is equal to");
     let toughness_only =
         !both && !power_only && nom_primitives::scan_contains(lower, "toughness is equal to");
+    // CR 613.4c: constant characteristic-defining P/T — "~'s power and toughness
+    // are each N" (Angry Mob's off-turn clause "... are each 2"): a fixed base
+    // value, not a dynamic quantity. Guarded by `!both` so the dynamic "are each
+    // equal to" framing (which also contains "are each ") keeps priority.
+    let both_const = !both
+        && !power_only
+        && !toughness_only
+        && nom_primitives::scan_contains(lower, "power and toughness are each ");
 
-    if !both && !power_only && !toughness_only {
+    if !both && !power_only && !toughness_only && !both_const {
         return None;
+    }
+
+    if both_const {
+        let after = strip_after(lower, "power and toughness are each ")?;
+        let digits: String = after
+            .trim_start()
+            .chars()
+            .take_while(|c| c.is_ascii_digit())
+            .collect();
+        let value = digits.parse::<i32>().ok()?;
+        let mut def = StaticDefinition::continuous()
+            .affected(TargetFilter::SelfRef)
+            .modifications(vec![
+                ContinuousModification::SetPower { value },
+                ContinuousModification::SetToughness { value },
+            ])
+            .cda()
+            .description(text.to_string());
+        if let Some(cond) = timing_condition {
+            def = def.condition(cond);
+        }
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[MED]** The new CR annotations point to rules that do not describe this CDA parsing path. Evidence: `crates/engine/src/parser/oracle_static/cda.rs:14` cites `CR 613.7c`, which is counter timestamp ordering, and `crates/engine/src/parser/oracle_static/cda.rs:43` / `crates/engine/src/parser/oracle_static/tests.rs:78` cite `CR 613.4c`, which is the modifier/counter sublayer rather than CDA/set P/T; local rules text has CDA P/T in `CR 604.3` / `CR 613.4a` and set P/T in `CR 613.4b`. Why it matters: engine/parser rule annotations are part of the verification contract, and these comments send future reviewers to the wrong layer/timestamp rules. Suggested fix: replace the `613.7c` / `613.4c` citations with the verified CDA/static-effect/layer citations that match the intended semantics.

**[MED]** The new regression test does not prove the on-turn clause carries the claimed quantity. Evidence: `crates/engine/src/parser/oracle_static/tests.rs:102` only checks that some `SetDynamicPower` exists, but it never asserts `SetDynamicToughness` or that both values are `2 + ObjectCount(Swamps your opponents control)`. Why it matters: the PR claims to recover Angry Mob's dynamic clause, but the test would still pass if the parser emitted the wrong dynamic quantity or omitted dynamic toughness. Suggested fix: mirror the nearby CDA quantity tests and match both dynamic P/T modifications against the exact `QuantityExpr::Offset { offset: 2, inner: Ref(ObjectCount{Swamp, controller: Opponent}) }` shape.

minion1227:
Fixed in `e90cab496`.

**CR annotations (matthewevans, gemini)** — corrected against `docs/MagicCompRules.txt`:
- Timing-condition peel: `CR 613.7c` (counter timestamp ordering) → **`CR 611.3a`** (a continuous effect from a static ability applies at any moment to whatever its text indicates, so the "During your turn," / "During turns other than yours," window is captured as a `StaticCondition`) + **`CR 604.3`** (CDA).
- Constant P/T form: `CR 613.4c` (Layer 7c modifiers) → **`CR 604.3` + `CR 613.4a`** (Layer 7a: effects from CDAs that *define* power/toughness). Test annotation updated to match.

**Fragile slicing (gemini [HIGH])** — replaced `text.len() - rest.len()` with `nom_tag_tp(&TextPair::new(text, lower), ..)`, which slices the original and lowercased strings in lockstep, so there is no byte-offset arithmetic and no case-fold panic risk.

No behavior change: full oracle_static suite (957) passes, parser-combinator gate + fmt clean. The two-branch parse is unchanged (dynamic `Offset{Swamps, +2}` gated `DuringYourTurn`; constant `2/2` gated `Not{DuringYourTurn}`).

matthewevans (changes requested):
**[MED]** The prior test-proof blocker is still present on the new head. Evidence: `crates/engine/src/parser/oracle_static/tests.rs:102` still only checks that the on-turn branch contains some `SetDynamicPower { .. }`; it does not assert `SetDynamicToughness`, the `+2` offset, the `ObjectCount`, the `Swamp` subtype, or `ControllerRef::Opponent`. Why it matters: the PR claims to recover Angry Mob's on-turn `2 + Swamps your opponents control` CDA, but this test would still pass if that quantity were wrong or if toughness were not emitted. Suggested fix: match both on-turn dynamic P/T modifications against the exact `QuantityExpr::Offset { offset: 2, inner: Ref(ObjectCount{Typed Swamp, controller: Opponent}) }` shape, similar to the nearby CDA quantity tests.

minion1227:
Addressed the remaining test-proof blocker in `954392eda`.

The on-turn assertion no longer merely checks that *some* `SetDynamicPower` exists — it now asserts the full modification vector by value:

```rust
let two_plus_opponent_swamps = QuantityExpr::Offset {
    inner: Box::new(QuantityExpr::Ref {
        qty: QuantityRef::ObjectCount {
            filter: TargetFilter::Typed(
                TypedFilter::new(TypeFilter::Subtype("Swamp".to_string()))
                    .controller(ControllerRef::Opponent),
            ),
        },
    }),
    offset: 2,
};
assert_eq!(on_turn.modifications, vec![
    ContinuousModification::SetDynamicPower { value: two_plus_opponent_swamps.clone() },
    ContinuousModification::SetDynamicToughness { value: two_plus_opponent_swamps },
]);
```

This fails if the `+2` offset, the counted `Swamp` subtype, the `Opponent` controller scope, or the `SetDynamicToughness` clause regress. The CR citations from the prior round (`CR 611.3a` / `CR 604.3` for the turn-window peel, `CR 604.3` / `CR 613.4a` Layer 7a for the constant CDA P/T) remain in place. Full oracle_static suite: 1008 passed.

matthewevans (approved):
Approved. I re-checked head 954392eda: the change is scoped to the CDA equality parser and its tests, the parse-diff is limited to Angry Mob, the CR annotations now cite the CDA/static/layer rules, and the test asserts both dynamic P/T modifications with the +2 offset, Swamp subtype, and opponent controller scope.

PR #5197, opened 2026-07-06
fix(parser): terminate card-name filter at a locative zone — "named X in your graveyard" count misparse (CR 201.2)

dhgoal:
Tier: Frontier
Model: claude-opus-4-8
Thinking: high

## Summary
CR 201.2 + CR 400.1: a **misparse** in the "cards named X **in `<zone>`**" count class. The locative zone phrase was swallowed into the card **name**: `"the number of cards named Frantic Inventory in your graveyard"` parsed to `Named { name: "Frantic Inventory in your graveyard" }` — a name **no object ever has** — with no `InZone` constraint, so the `ObjectCount` fell back to the battlefield and **always resolved to 0**. These cards are marked "supported" (no `Unimplemented`) but silently misresolve: **Frantic Inventory** drew only its first card regardless of graveyard copies; the whole class was affected (Accumulated Knowledge, Take Inventory, Undead Servant, Goblin Gathering, Galvanic Bombardment, Ancestral Anger, Plague Rats "on the battlefield", …).

**Fix (parser-only):** terminate the card-name filter at a locative `" in <zone>"` / `" on the battlefield"` (requiring a *real* zone after the preposition, so name-internal words and `Extract from Darkness` stay whole) and re-attach it as an `InZone`/`InAnyZone` filter prop + controller — parity with the non-named `"card in your graveyard"` path. Distinct from the `"from <zone>"` move-origin suffix, which stays in the remainder for the caller (CR 115.2 return-from-graveyard sources). The count/draw engine already resolves the corrected AST — no engine change.

## Files changed
- crates/engine/src/parser/oracle_target.rs (locative terminator + name-consumed zone-suffix re-run + parser test)
- crates/engine/tests/frantic_inventory_graveyard_count.rs (new — runtime cast test)

## CR references
- `CR 201.2` — a card's name is set by its name line; a locative clause is not part of the name.
- `CR 400.1` — zones; the locative scopes *where* named objects are counted.
- `CR 121.1` — Frantic Inventory's draw count.

## Gate A
```
[exit 0 — no violations]
```
Also ran the authoritative check on my changes only — `git diff origin/main...HEAD -- 'crates/engine/src/parser/**' | grep '^+' | grep -E '<FORBIDDEN_METHODS>'` — which printed **nothing**. The new terminator is combinator-only (`alt`/`tag` + the shared `parse_zone_suffix`); the name-consumed re-run gates on an `alt((tag("in "),tag("on ")))` locative probe, not string dispatch.

## Anchored on
- crates/engine/src/parser/oracle_target.rs:1704 — `parse_named_filter_origin_zone_terminator` (the sibling `"from <zone>"` name terminator); the new `parse_named_filter_locative_zone_terminator` mirrors its structure (zone-after-preposition guard) but is re-attached as a filter prop rather than left as a move source.
- crates/engine/src/parser/oracle_target.rs:6623 — `parse_zone_suffix`, the shared zone→(`InZone` props, controller) extractor reused verbatim by both the terminator and the name-consumed re-run, so named and non-named `"… in your graveyard"` filters lower identically.

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — exit 0; authoritative `git diff origin/main...HEAD` combinator grep — **empty**
- `cargo test -p engine --lib -- parser::` — **7439 pass / 0 fail**; new parser test `named_filter_terminates_at_locative_zone` covers "in your graveyard" (InZone+You), "in all graveyards" (InZone, no controller), "on the battlefield", the unchanged `"from <zone>"` move-origin, and name-internal `Extract from Darkness`
- **Runtime cast test** (fail-on-revert): `frantic_inventory_draws_one_plus_named_graveyard_copies` — 2 graveyard copies + 1 decoy → draws **1+2=3** (reverting the fix drops the count to 0 → draws 1)
- Regenerated card-data (oracle-gen rebuilt with the fix): Frantic Inventory now parses `Named{"frantic inventory"}` + `InZone{Graveyard}`; total `Unimplemented` count did not increase (no new gaps / no over-match regression).

## Scope Expansion
None.

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -1714,6 +1714,31 @@ fn parse_named_filter_origin_zone_terminator(
     Ok((&input[consumed..], ()))
 }
 
+/// CR 201.2 + CR 400.1: A locative zone constraint ("in your graveyard", "in
+/// all graveyards", "on the battlefield") scopes *where the named objects are
+/// counted/filtered*; it is outside the literal card name. Unlike the
+/// "from <zone>" move-origin suffix (`parse_named_filter_origin_zone_terminator`,
+/// which the caller consumes as a move source and leaves in the remainder), an
+/// "in"/"on" locative both terminates the name and is re-attached as an
+/// `InZone`/`InAnyZone` filter prop by the caller. Requires a real zone after
+/// the "in"/"on" preposition so a name-internal "in"/"on" stays whole. Covers
+/// the "cards named X in your graveyard / in all graveyards" count class
+/// (Frantic Inventory, Accumulated Knowledge, Take Inventory, Undead Servant,
+/// Goblin Gathering, Galvanic Bombardment, Ancestral Anger) and "creatures
+/// named X on the battlefield" (Plague Rats).
+fn parse_named_filter_locative_zone_terminator(
+    input: &str,
+) -> Result<(&str, ()), nom::Err<OracleError<'_>>> {
+    alt((tag::<_, _, OracleError<'_>>(" in "), tag(" on "))).parse(input)?;
+    let Some((_, _, consumed)) = parse_zone_suffix(input) else {
+        return Err(nom::Err::Error(nom::error::Error::new(
+            input,
+            nom::error::ErrorKind::Fail,
+        )));
+    };
+    Ok((&input[consumed..], ()))
+}
+
 fn parse_named_filter_terminator(input: &str) -> Result<(&str, ()), nom::Err<OracleError<'_>>> {
     alt((
         // Controller-scope suffixes (CR 109.4). Longest-match-first.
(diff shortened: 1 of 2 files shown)


matthewevans:
Reviewed the current head. The code path is focused on the named-card locative-zone parser boundary and the regression exercises the Frantic Inventory graveyard count behavior. I’m not approving/enqueuing yet because this is a parser-semantics change and the current-head parse-diff sticky is not present while CI is still running. Once the parse-diff comment is published for head `1e149208410e1154d31907b20a5e0e86e9a82f55`, this can be rechecked against that evidence.

matthewevans (approved):
Reviewed parser boundary, parse-diff, and runtime regression; this is scoped to the named-card locative-zone count bug.

PR #5198, opened 2026-07-06
fix(parser): strip ability-word prefix on subject-anchored statics (CR 207.2c)

minion1227:
## Summary
Fixes a **class** of static abilities that silently drop when their Oracle line carries a leading **ability-word label** (CR 207.2c — ability words are italic flavor with no rules meaning).

The subject-anchored static parsers match their subject at the *start* of the line, so a label like `Chroma — ` / `Fateful hour — ` / `Protector — ` prevents the subject match and the whole static is lost — leaving the permanent strictly off its printed rules. (`Threshold — `/`Metalcraft — ` self-pumps slipped through only because their scanner matches mid-line.)

`parse_static_line_multi_inner` now retries the dispatch after stripping a **recognized** ability-word label — whitelist-gated through the shared `is_known_ability_word` authority, exactly as the token-grant path in `keyword_grant.rs` already does. It is a **strict fallback**: any line the ordinary dispatch already parses is returned untouched (zero regression), and the stripped body carries no further label so the retry recurses at most once.

Recovered cards (and every future known-ability-word subject-anchored static):
- **Light from Within** — `Chroma — Each creature you control gets +1/+1 for each white mana symbol in its mana cost.`
- **Gavony Ironwright** — `Fateful hour — As long as you have 5 or less life, other creatures you control get +1/+4.`
- **Echo of Dusk** — `Descend 4 — As long as there are four or more permanent cards in your graveyard, ~ gets +1/+1 and has lifelink.`
- **Cryptothrall** — `Protector — Other artifact creatures you control have hexproof.`
- **Clamavus** — `Proclamator Hailer — Each creature you control gets +1/+1 for each +1/+1 counter on it.`

## Files changed
- `crates/engine/src/parser/oracle_modal.rs`

## Anchored on
- `crates/engine/src/parser/oracle_static/keyword_grant.rs:1768` — the token-granted-ability path already strips an em-dash label with `strip_ability_word_with_name` and gates it on `is_known_ability_word` before re-dispatching the body; this change applies the identical, whitelist-gated pattern to the static-line dispatch.
- `crates/engine/src/parser/oracle_modal.rs:1602` — `ABILITY_WORD_NAMES` (CR 207.2c) is the single source of recognized ability words; the two 40K flavor words are added here so the gate accepts them.

## Gate A
```

```
(exit 0)

## CR references
- `CR 207.2c` — ability words are italicized flavor text with no rules meaning (verified against `docs/MagicCompRules.txt`).

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — clean (Gate A above)
- `cargo test -p engine --lib parser::oracle_static` — 1008 pass / 0 fail
- `cargo test -p engine --lib parser::oracle_modal` — 53 pass / 0 fail
- New regression test `ability_word_prefix_is_stripped_from_subject_anchored_statics` — passes, incl. a safety case asserting an **unknown** em-dash label is NOT stripped (whitelist gate holds).

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/parser/oracle_modal.rs
+++ b/crates/engine/src/parser/oracle_modal.rs
@@ -1642,6 +1642,9 @@ pub(super) const ABILITY_WORD_NAMES: &[&str] = &[
     "pack tactics",
     "paradox",
     "parley",
+    // Warhammer 40,000 Commander (40K) flavor ability words — no rules meaning.
+    "proclamator hailer",
+    "protector",
     "radiance",
     "raid",
     "rally",
--- a/crates/engine/src/parser/oracle_modal.rs
+++ b/crates/engine/src/parser/oracle_modal.rs
@@ -1642,7 +1642,7 @@ pub(super) const ABILITY_WORD_NAMES: &[&str] = &[
     "pack tactics",
     "paradox",
     "parley",
-    // Warhammer 40,000 Commander (40K) flavor ability words — no rules meaning.
+    // CR 207.2c: Warhammer 40,000 Commander (40K) flavor ability words — no rules meaning.
     "proclamator hailer",
     "protector",
     "radiance",
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -1077,7 +1077,34 @@ pub(crate) fn parse_tiered_enters_with_additional_counters_pattern(
     parse_tiered_enters_with_additional_counters_parts(&tp).map(|(_, pattern)| pattern)
 }
 
+/// CR 207.2c: An ability word is italicized flavor text with no rules meaning
+/// (e.g. `Chroma`, `Metalcraft`, `Fateful hour`, and the set-specific `Protector`
+/// / `Proclamator Hailer`). The subject-anchored static parsers match their
+/// subject at the *start* of the line, so a leading ability-word label like
+/// `"Chroma — Each creature you control gets ..."` prevents them from firing and
+/// the whole static silently drops. When the ordinary dispatch classifies
+/// nothing, strip a *recognized* ability-word label — whitelist-gated through the
+/// shared `is_known_ability_word` authority, exactly as the token-grant path in
+/// `keyword_grant.rs` does — and re-enter the dispatch once on the body.
+///
+/// This is a strict fallback: any line the dispatch already parses is returned
+/// untouched, so no existing coverage can regress. The stripped body carries no
+/// further label, so `strip_ability_word_with_name` yields `None` on the retry
+/// and the recursion terminates after a single hop.
 pub(crate) fn parse_static_line_multi_inner(text: &str) -> Vec<StaticDefinition> {
+    let defs = parse_static_line_multi_dispatch(text);
+    if !defs.is_empty() {
+        return defs;
+    }
+    if let Some((ability_word, body)) = super::oracle_modal::strip_ability_word_with_name(text) {
+        if super::oracle_modal::is_known_ability_word(&ability_word) {
+            return parse_static_line_multi_dispatch(&body);
+        }
+    }
+    defs
+}
+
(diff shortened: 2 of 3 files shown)


matthewevans (changes requested):
Thanks for the scoped parser fix. The fallback seam looks reasonable, but the regression proof needs to be stronger before this lands.

The new positive cases in `crates/engine/src/parser/oracle_static/tests.rs` only assert `parse_static_line_multi(...).len() == 1` for lines 28-57. That would still pass if the stripped body parsed into the wrong static shape, affected filter, condition, or modification, so it does not prove the ability-word fallback preserves the intended semantics. Please assert the resulting `StaticDefinition` shape for representative cases, for example: Chroma's dynamic P/T quantity, Fateful hour's condition plus "other creatures you control" pump, Protector's `Other artifact creatures` hexproof grant, and keep the unknown-label negative case.

minion1227:
Addressed both in `110bc87d9`.

**@matthewevans — stronger regression proof.** The positive cases now assert the resulting `StaticDefinition` by field, not just `len() == 1`:
- **Chroma** (Light from Within) — `affected` = creatures you control (no `Another`), no condition, and `modifications` = `[AddDynamicPower, AddDynamicToughness]` each equal to `ManaSymbolsInManaCost { color: Some(White) }` (and power == toughness).
- **Fateful hour** (Gavony Ironwright) — `affected` = **other** creatures you control (`FilterProp::Another`), `condition` = `QuantityComparison { comparator: LE, rhs: Fixed { value: 5 } }` (your life ≤ 5), pump `+1/+4`.
- **Protector** (Cryptothrall) — `affected` = **other** artifact creatures you control (`[Creature, Artifact]` + `Another`), `modifications` = `[AddKeyword { Hexproof }]`.

The unknown-em-dash-label negative case (`Vigilant Sentry — …` ⇒ 0 statics) is kept, so the fallback stays whitelist-gated.

**@gemini — CR annotation.** Added `CR 207.2c` to the 40K ability-word comment in `ABILITY_WORD_NAMES` (verified against `docs/MagicCompRules.txt:1497`).

oracle_static suite 1009 pass / 0 fail; parser gate + fmt clean.

matthewevans (changes requested):
Thanks for tightening the regression proof. One cleanup still needs to happen before this lands: the new test accidentally splits the existing Serra's Emissary rules doc comment.

At current head, `crates/engine/src/parser/oracle_static/tests.rs:18-19` starts the Serra's Emissary CR 702.16 / CR 609.6 comment immediately above `ability_word_prefix_is_stripped_from_subject_anchored_statics`, and line 130 has the orphaned tail (`/// type.`) above the original Serra test. Please move the full Serra's Emissary doc comment back above `compound_subject_keyword_static_splits_serras_emissary` and leave only the CR 207.2c ability-word comment above the new test.

I don't have a behavior/code-path objection to the parser seam after this comment-layout fix.

minion1227:
@matthewevans — comment-layout fix pushed in `f34d9d4d6`.

The Serra's Emissary `CR 702.16 / CR 609.6` doc comment is now whole again above `compound_subject_keyword_static_splits_serras_emissary`, and only the `CR 207.2c` ability-word comment sits above `ability_word_prefix_is_stripped_from_subject_anchored_statics`. Comment-only move — no behavior change, `cargo fmt --all --check` clean.

matthewevans (approved):
Reviewed current head after the follow-up. The prior regression-proof request is addressed with shape assertions for Chroma, Fateful hour, and Protector, and the Serra's Emissary CR doc comment is whole again above its original test. The implementation stays at the existing ability-word/static-parser seam, CR 207.2c is verified, and the source diff remains scoped to the parser/test files. Approving and enqueueing; the merge queue will wait for the in-progress checks on this head.

PR #5200, opened 2026-07-06
fix(parser): scope "and gets a <kind> counter" to each opponent (Vraska's Fall, #4733)

### LINKED ISSUE #4733: Vraska's Fall bug — Vraska's Fall causes its controller to get a poison counter instead of their op
Reported in Discord: https://discord.com/channels/1485498006781427802/1520598130192683160/1520598130192683160

**Thread:** Vraska's Fall bug
**Cards:** vraska's fall

## Summary
Vraska's Fall causes its controller to get a poison counter instead of their op

### vraska's fall
Each opponent sacrifices a creature or planeswalker of their choice and gets a poison counter.

---
<sub>report_id: `discord:1520598130192683160:1520598130192683160:0` · discord: `1520598130192683160/1520598130192683160`</sub>

nickmopen:
## Summary

Closes #4733.

**Vraska's Fall** — "Each opponent sacrifices a creature or planeswalker of their choice **and gets a poison counter**" — gave the poison counter to the spell's **controller** instead of each opponent.

## Root cause

The "and gets a poison counter" conjunct stayed glued to the sacrifice as an unscoped `try_split_targeted_compound` sub-ability, so the carried "each opponent" `player_scope` never reached it. The sub parsed to `GivePlayerCounter { target: Controller }` with **no** `player_scope`, and at resolution `Controller` is the caster.

The asymmetry is in the bare-`and` chunk splitter (`starts_bare_and_clause`): it recognizes `"… and loses 2 life"` / `"… and gains 2 life"` as clause starts (→ split into their own chunk → pick up `player_scope: Opponent` via the chunk loop's scope-carry), but **not** `"… and gets a poison counter"`. So the life clause is scoped and the counter clause isn't.

For comparison — before this change:

| conjunct | `starts_bare_and_clause` | scoped? |
|---|---|---|
| `loses 2 life` | true | ✅ |
| `gets a poison counter` | **false** | ❌ (bug) |

## Fix

Add `starts_gets_counter_clause_lower`, a trailing `.or()` arm mirroring the existing `starts_remove_counter_clause_lower`, that splits a bare `"gets a/an/N <kind> counter[s]"` player-counter conjunct. The discriminator is a slash-free head up to the `counter` token, so the P/T-counter form (`"gets a +1/+1 counter"`, an object continuation) stays un-split and a bare pump (`"gets +2/+0"`, no counter token) doesn't match.

With the fix, the conjunct splits into its own chunk and inherits `player_scope: Opponent` exactly like the `loses 2 life` sibling, so each opponent receives the counter.

## Verification

- New end-to-end test `vraska_fall_poison_scoped_to_each_opponent`: the poison `GivePlayerCounter` sub-ability now carries `player_scope: Some(Opponent)`.
- New unit test `gets_player_counter_clause_splits`: poison/energy/experience counters split; `+1/+1` counter and bare pump do not.
- Full engine lib suite: **15,363 passed, 0 failed** (no snapshot changes).
- `cargo fmt` clean; `scripts/check-parser-combinators.sh` and `scripts/check-engine-authorities.sh` pass; clippy clean.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -9631,7 +9631,7 @@ fn count_ticket_symbols(rest: &str) -> Option<u32> {
 /// - "gets two experience counters"
 /// - "get ten rad counters"
 /// - "get {TK}{TK}" (CR 107.17 ticket symbol form; see `count_ticket_symbols`)
-fn try_parse_player_counter(lower: &str) -> Option<ImperativeFamilyAst> {
+pub(super) fn try_parse_player_counter(lower: &str) -> Option<ImperativeFamilyAst> {
     // Strip "get/gets " prefix
     let (rest, _) = alt((tag::<_, _, OracleError<'_>>("gets "), tag("get ")))
         .parse(lower)
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -2176,26 +2176,37 @@ fn starts_remove_counter_clause_lower(s: &str) -> OracleResult<'_, ()> {
 /// the source's controller instead of each opponent. Splitting it off routes it
 /// through the chunk loop's scope-carry (mirroring "... and loses 2 life").
 ///
-/// The discriminator is a `counter` token within THIS conjunct's segment
-/// (bounded at the next " and ") whose slash-free head excludes the P/T-counter
-/// form ("a +1/+1 counter" — an object-counter continuation on the prior
-/// conjunct's target, left un-split). Player counters (poison/energy/experience/
-/// ticket/…) are word-named and slash-free. Mirrors the trailing `.or()` arm
+/// The discriminator is the SAME classifier the lowerer uses:
+/// `imperative::try_parse_player_counter` (CR 122.1b — poison/experience/rad/
+/// ticket only; it rejects energy, which lowers to `GainEnergy`, and object/P-T
+/// counters like `+1/+1`). Reusing it guarantees the split set can never diverge
+/// from what the chunk parser accepts as `GivePlayerCounter`, so no object/energy
+/// counter text is split into a chunk the parser would reject or misroute. The
+/// segment is bounded to THIS conjunct (up to the next " and ") so a later
+/// `counter` token can't false-trigger. Mirrors the trailing `.or()` arm
 /// `starts_remove_counter_clause_lower`.
 fn starts_gets_counter_clause_lower(s: &str) -> OracleResult<'_, ()> {
-    let (rest, _) = tag("gets ").parse(s)?;
-    let segment = match take_until::<_, _, OracleError<'_>>(" and ").parse(rest) {
+    // Bound to THIS conjunct (up to the next " and ").
+    let segment = match take_until::<_, _, OracleError<'_>>(" and ").parse(s) {
         Ok((_, seg)) => seg,
-        Err(_) => rest,
+        Err(_) => s,
     };
-    // Head up to "counter"; require it slash-free so "+1/+1 counter" (object P/T)
-    // stays un-split while "poison counter" (player) matches.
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
Thanks for narrowing this to the `and gets a poison counter` chunk boundary. The overall seam looks right: this should be handled in the sequence splitter so the next chunk can inherit the prior `each opponent` player scope.

I need one tightening before this can land. The new splitter currently accepts any slash-free `gets ... counter` head as player-counter syntax, and the new test explicitly accepts `gets an energy counter`. That is broader than the existing player-counter parser: `try_parse_player_counter` rejects energy and only accepts poison, experience, rad, and ticket word forms (`crates/engine/src/parser/oracle_effect/imperative.rs`, around the `PlayerCounterKind` match). It also has a test that unknown/object counters such as charge counters must not parse as player counters.

Please make this splitter use the same player-counter classification boundary, or otherwise prove the accepted set cannot diverge from the parser that will consume the split chunk. In particular, remove or change the `gets an energy counter` splitter expectation unless the lowerer also intentionally supports that surface as `GainEnergy`. The Vraska's Fall case should still be covered by a discriminating end-to-end test that confirms the poison sub-ability carries `player_scope: Opponent`.

I’m not asking for broader changes here; this just needs the split predicate to match the actual player-counter grammar so object/energy counter text cannot be split into a chunk the parser will reject or misroute.

nickmopen:
Thanks — tightened as requested.

The split predicate now reuses the lowerer's own classifier, `imperative::try_parse_player_counter`, as the discriminator (promoted to `pub(super)`). So the split set is now *identical* to what the chunk parser accepts as `GivePlayerCounter` — poison/experience/rad/ticket — and can't diverge:

- `gets an energy counter` → **not** split (energy lowers to `GainEnergy`, not a player counter)
- `gets a charge counter`, `gets a +1/+1 counter` → **not** split (object counters)
- `gets a poison counter` / `experience` / `rad` / `ticket` → split

A trailing sentence period is stripped first so a chunk-final `"… gets a poison counter."` still classifies (that was the one wrinkle — `try_parse_player_counter` anchors on a `counter`/`counters` suffix). `gets_player_counter_clause_splits` was updated to assert the energy/charge/`+1/+1` rejections, and the discriminating end-to-end `vraska_fall_poison_scoped_to_each_opponent` (poison sub carries `player_scope: Opponent`) is unchanged.

Local: full engine lib suite **15,365 / 0**, fmt + parser-combinator + engine-authorities gates pass, clippy clean.

matthewevans:
Maintainer note: I re-reviewed `e4a046cce2dfc694143f3b28bcebd6d878fef20c`. The prior splitter blocker is addressed: the new boundary now delegates to the same player-counter classifier the chunk parser uses, and the tests reject energy/charge/+1/+1 counter cases while preserving poison/experience/rad/ticket.

I’m holding approval only for current-head automation evidence. The existing parse-diff sticky is from `2026-07-06T15:36:34Z`, before this head/follow-up, so I need the refreshed current-head CI/parse-diff before approving or enabling auto-merge. No author action from this note unless the refreshed checks report a new issue.

matthewevans (approved):
Approved after re-reviewing `e4a046cce2dfc694143f3b28bcebd6d878fef20c`.

Evidence checked:
- Local diff against current `origin/main` is limited to `crates/engine/src/parser/oracle_effect/imperative.rs` and `crates/engine/src/parser/oracle_effect/sequence.rs`.
- The previous splitter blocker is fixed: the new `and gets a <kind> counter` boundary delegates to `imperative::try_parse_player_counter`, so it accepts the same player-counter set the chunk parser can actually lower.
- Added tests cover poison/experience/rad/ticket and reject energy, charge, +1/+1, and bare pump cases; the Vraska's Fall parser test asserts the poison counter inherits the opponent scope.
- CR references were checked locally (`CR 109.5`, `CR 122.1`, `CR 608.2c`).
- Current checks are green and there are no unresolved review threads.

The refreshed parse-diff sticky includes several unrelated card deltas, but the local diff proves those are baseline drift from already-merged parser changes, not scope added by this PR. The Vraska's Fall delta itself matches the intended fix.

PR #5201, opened 2026-07-06
fix(parser): Hollow One cost reduction miscounts battlefield permanents instead of cycled/discarded cards this turn

rykerwilliams:
## Summary
Fixes engine support for **Hollow One** (Hour of Devastation, Scryfall oracle_id `b9e4d34a-ba91-4ad6-b886-daeeda6c9498`). Real Oracle text:
> This spell costs {2} less to cast for each card you've cycled or discarded this turn.
> Cycling {2} ({2}, Discard this card: Draw a card.)

Root cause (confirmed from a real user's downloaded game-state dump — the reported symptom was "casting cost changes erratically"): the compound "you've cycled or discarded this turn" phrase was unrecognized by any quantity combinator, so `try_parse_cost_modification`'s `for each` fallback chain silently produced a bare `QuantityRef::ObjectCount` over all Card-type objects with no zone/controller restriction. Since that filter carries no zone, it defaulted to scanning the battlefield across **both players**, which is why the cost appeared to shift unpredictably.

**Fix**: extends `parse_number_of_cards_discarded_this_turn` (`oracle_nom/quantity.rs`) to recognize both orderings of "cycled or discarded this turn", binding to the existing `QuantityRef::CardsDiscardedThisTurn { player: Controller }`. Cycling already increments the same discard counter via its `Discard` cost (CR 702.29a: cycling's cost is "[Cost], Discard this card"), so no new runtime plumbing is needed — parser-only change.

**Note on scope**: an earlier revision of this PR also included a fabricated ETB "sacrifice if cast for {5} or less" clause and a Trample keyword that do not actually exist on this card — I invented them during the original investigation and never checked them against an authoritative source before building out a parser combinator and tests around them. That was caught during review (both the reviewer's feedback and my own re-verification against Scryfall) and has been fully removed; see PR comments for details. This PR is now scoped to only the real, verified bug.

## Files changed
- `crates/engine/src/parser/oracle_nom/quantity.rs` — new `alt()` arm on `parse_number_of_cards_discarded_this_turn`
- `crates/engine/src/parser/oracle_static/tests.rs` — parser tests for both cost-reduction phrase orderings
- `crates/engine/tests/integration/hollow_one_cost_reduction.rs` (new) — 2 cast-pipeline integration tests (controller-reduction applies; opponent's events don't reduce the caster's cost — the direct regression test)
- `crates/engine/tests/integration/main.rs` — registers the new integration test module

## CR references
- CR 701.9 (Discard keyword action)
- CR 702.29a (Cycling's cost is "[Cost], Discard this card")
- CR 702.29d (cycle-or-discard once-only trigger rule)

CR 701.9/702.29a/702.29d confirmed by @matthewevans against `docs/MagicCompRules.txt` during review.

## Track
Developer

## LLM
Model: claude-sonnet-5
Thinking: high
Tier: Standard

## Gate A
```
(exit 0, no output — no combinator-mandate violations in the diff)
```

## Anchored on
- `crates/engine/src/parser/oracle_nom/quantity.rs:485` — existing `parse_number_of_cards_discarded_this_turn` `alt()` chain (4 pre-existing player-scope phrasings); extended with a 5th arm for the compound "cycled or discarded" phrasing rather than writing a new function.
- `crates/engine/src/parser/oracle_static/tests.rs` — existing self-spell cost-reduction tests (Demilich/Tolarian Terror class) for the `parse_static_line` → `StaticMode::ModifyCost` test pattern this PR's new tests follow.

## Verification
A local Rust toolchain was installed and used for this PR (it was not available initially — an earlier revision incorrectly relied on CI alone; that revision also contained the fabricated ETB scope described above, which is now removed). Verified locally before this push:
- `cargo build -p engine --tests` — clean
- `cargo test -p engine -- hollow_one` — 4/4 pass, 0 failed

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None — see above, resolved before this push.


matthewevans (changes requested):
Thanks for putting real parser and runtime coverage around this; the overall shape is useful, especially the controller-vs-opponent cost-reduction regression and the `{5} or less` ETB boundary. I need changes before this can land.

Blocking issues:

1. `crates/engine/src/parser/oracle_nom/condition.rs:5021` adds an inline `use crate::types::mana::ManaCost;` inside `parse_cast_for_mana_threshold`. This repo keeps imports at module scope, and this file already has the normal top-level import block. Please move `ManaCost` into the top-level imports.

2. `crates/engine/src/parser/oracle_nom/quantity.rs:506-510` commits a CR caveat saying the exact cycling sub-letter still needs grep verification and was not possible in the session. This repo has `docs/MagicCompRules.txt`, and CR annotations must be verified before being committed. I checked the local rules text: `702.29a` defines cycling as `[Cost], Discard this card`, and `702.29d` is the cycle-or-discard once-only trigger rule. Please replace the caveat with the exact verified rule(s) that support this parser path.

3. `crates/engine/tests/integration/hollow_one_cost_reduction_and_etb.rs:25-30` leaves an "offline environment" assumption asking maintainer/CI to confirm Hollow One's printed cost. Please remove that from committed tests. Either verify from repo data and state the fixture fact directly, or make the test explicitly a synthetic Hollow One-like card with `{5}`; don't land a test comment that documents unverified proof.

4. `crates/engine/src/parser/oracle_nom/condition.rs:5026-5028` cites `CR 118.7a` for rejecting colored-shard `{3}{R}` thresholds. That rule is about generic cost reductions, not a parser boundary for `was cast for {N}` threshold text. Please either cite the rule that actually justifies this behavior or make this a plain parser-scope comment without a CR annotation.

No approval from me until those are cleaned up and the current CI/parser proof is available for this head.

rykerwilliams:
@matthewevans Thanks for the review — all four points are addressed, though the fix ended up bigger than a cleanup pass. Some background: while responding to your feedback I re-verified Hollow One's Oracle text against Scryfall and found the entire ETB clause ("if it was cast for {5} or less, sacrifice it") and the Trample keyword were fabricated by Claude during the original investigation — the real card has neither. Only the cost-reduction line and Cycling are real. So:

1. Inline `use crate::types::mana::ManaCost` — moot, removed along with the whole `parse_cast_for_mana_threshold` combinator (it only existed to parse the fictional ETB clause).
2. Cycling CR sub-letter caveat — fixed with your verified citations: now cites CR 701.9 + CR 702.29a + CR 702.29d instead of the "not verified" placeholder.
3. "Offline environment" unverified-assumption comment — removed; the integration test now states the {5} printed cost as a directly-confirmed fact (Scryfall oracle_id `b9e4d34a-ba91-4ad6-b886-daeeda6c9498`), and only contains the two real cost-reduction tests.
4. CR 118.7a misattribution — moot, removed along with the shard-rejection guard that was part of the fictional ETB combinator.

Net effect: the PR is now scoped to just the real bug — the cost-reduction clause misparsing "cycled or discarded this turn" into a cross-player `ObjectCount`. Ran a full local build/test/clippy/fmt pass before pushing this time (`cargo test -p engine -- hollow_one`: 4/4 pass; `cargo clippy -p engine --tests -- -D warnings`: clean; `cargo fmt --all -- --check`: clean). Retitled from WIP now that this is ready for another look.

matthewevans (approved):
Reviewed current head `2c2933981194ca8be71d9ea6835ce903a98d8d0e`.

This looks good to me. The follow-up removed the fabricated Hollow One clauses, kept the change scoped to the real cost-reduction line, and the parser now maps "cycled or discarded this turn" / "discarded or cycled this turn" onto the existing `CardsDiscardedThisTurn` quantity instead of a battlefield object count. I also verified the CR references locally: CR 701.9 covers discard, CR 702.29a defines cycling as discarding the card as the cycling activation cost, and CR 702.29d covers the "cycles or discards" once-only wording.

The added tests cover the building block and the runtime boundary I care about here: parser quantity coverage, static cost-modification shape, and casting legality proving controller-only discard/cycling history is counted while the opponent's history is not.

Issue #3267, opened 2026-06-14
Sanwell, Avenger Ace holding all exiled cards — [[Sanwell, Avenger Ace]] is supposed to send uncast cards to bottom of…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1515431749717332098/1515431749717332098

**Thread:** Sanwell, Avenger Ace holding all exiled cards
**Cards:** sanwell, avenger ace
**Parser status:** has_gaps
**Extraction confidence:** 0.90

## Summary
[[Sanwell, Avenger Ace]] is supposed to send uncast cards to bottom of library, but it just stacks them all under it, and when it's removed, they all go to exile.

### sanwell, avenger ace
As long as an artifact creature you control is attacking, prevent all damage that would be dealt to Sanwell.
Whenever Sanwell becomes tapped, exile the top six cards of your library. You may cast a Vehicle or artifact creature spell from among them. Then put the rest on the bottom of your library in a random order.

---
<sub>report_id: `discord:1515431749717332098:1515431749717332098:0` · discord: `1515431749717332098/1515431749717332098`</sub>

matthewevans:
Reopened from a new Discord report: Sanwell exiles cards but does not offer a cast choice or move the remaining cards to the library bottom.

mike-theDude:
## Triage analysis

**Clarified summary**
When Sanwell, Avenger Ace becomes tapped, the top six cards are exiled, but the player is not offered the permitted Vehicle or artifact-creature cast and the uncast cards are not moved to the bottom of the library in a random order.

**Expected behavior**
The controller may cast one Vehicle or artifact creature spell from the six exiled cards, then all remaining cards are put on the bottom of that player's library in a random order.

**Investigation findings**
The reported trigger is marked supported and contains typed exile, cast-from-zone, and library-bottom children. The representation is nevertheless unfaithful: the cast child exposes cards exiled by the source without recording the required Vehicle-or-artifact-creature filter, while the disposition chain contains repeated library-bottom nodes over the same broad source set. That leaves the runtime without a faithful partition between the chosen cast card and the rest, matching both reported symptoms.

**Likely scope**
Oracle parser effect-chain construction plus the engine's interactive cast-from-exile/rest disposition flow.

**Acceptance criteria**
- Tapping Sanwell offers only Vehicle or artifact creature spells among the six exiled cards.
- The controller may cast at most one qualifying spell using the normal casting pipeline.
[...]

Issue #2900, opened 2026-06-10
Blinkmoth Urn: only the Urn's controller adds mana (counting the turn player's artifacts) instead of each player adding {C} for their own artifacts


matthewevans:
## Oracle text (verified from client/public/card-data.json)
At the beginning of each player's first main phase, if this artifact is untapped, that player adds {C} for each artifact they control.

## Reported behavior
> Expected behaviour: At the beginning of each player's upkeep, that player adds an amount of {C} equal to the number of artifacts that player controls. Actual behaviour: At the beginning of each player's upkeep, the controller of Blinkmoth Urn adds an amount of {C} equal to the number artifacts the turn player controls.

## Expected vs actual
Expected: At the beginning of EACH player's first main phase, THAT player (if the Urn is untapped) adds {C} equal to the number of artifacts they themselves control.
Actual: Only the Urn's controller receives the mana, and the amount is based on the turn player's artifact count rather than the receiving player's own artifacts.

The trigger is parsed as `Mana(Colorless, count = ObjectCount(filter: Artifact, controller: ScopedPlayer))` with description matching the 'each player / that player' wording. The runtime is misrouting the mana to the Urn's controller and/or resolving `ScopedPlayer` to the turn player instead of the per-player trigger subject. (Reporter says 'upkeep' but the card and parse use the first main phase — the routing/scope bug is the same.)

Source: https://discord.com/channels/1485498006781427802/1513418666090168432/1513418666090168432

PR #5202, opened 2026-07-06
fix(parser): cycle the whole hand to the library bottom for Teferi’s Puzzle Box

### LINKED ISSUE #4241: Teferi's Missing Box — [[Teferi's Puzzle Box]] does not resolve for any player correctly.
Reported in Discord: https://discord.com/channels/1485498006781427802/1518516058150142003/1518516058150142003

**Thread:** Teferi's Missing Box
**Cards:** teferi's puzzle box
**Parser status:** has_gaps

## Summary
[[Teferi's Puzzle Box]] does not resolve for any player correctly.

### teferi's puzzle box
At the beginning of each player's draw step, that player puts the cards in their hand on the bottom of their library in any order, then draws that many cards.

---
<sub>report_id: `discord:1518516058150142003:1518516058150142003:0` · discord: `1518516058150142003/1518516058150142003`</sub>

codevector96:
## Summary

Fixes #4241 — Teferi's Puzzle Box did nothing on the draw step. "At the beginning of each player's draw step, that player puts the cards in their hand on the bottom of their library in any order, then draws that many cards" parsed but into a no-op AST: the whole-hand origin collapsed to `PutAtLibraryPosition { count: Fixed(1) }` (no hand moved), the "that many" count was never captured (draw of 0), and the draw targeted the Box's controller instead of the scoped player.

The fix reroutes the whole-hand→library-position surface form to the existing `Effect::ChangeZoneAll` mover — the same building block Whirlpool Drake's "shuffle the cards from your hand into your library, then draw that many" already uses — with `library_position: Some(Bottom)` and no shuffle. That mover already stamps `last_effect_count`, which the `EventContextAmount` "draws that many" reads, and the existing subject-injection path binds "that player" (the scoped player) to both the move and the draw. No new engine surface.

Model: claude-opus-4-8
Tier: Frontier
Thinking: high

## Anchored on

The change is Whirlpool Drake ("shuffle the cards from your hand into your library, then draw that many cards") re-pointed at a positional library placement instead of a shuffle. Each new piece mirrors an existing analog:

- `crates/engine/src/parser/oracle_effect/imperative.rs` — `parse_mass_zones_to_library`: the nom-combinator whole-hand-origin recognizer whose combinator style the new `parse_hand_to_library_position` copies.
- `crates/engine/src/parser/oracle_effect/imperative.rs` — `lower_change_zone_all_to_library`: the `Effect::ChangeZoneAll { origin: Hand, destination: Library }` field-set the new lowering reuses (with `library_position: Some(pos)` and no shuffle sub).
- `crates/engine/src/parser/oracle_effect/mod.rs` — the `Effect::Shuffle` arm of `apply_anchor_subject`: the new `Effect::Draw` arm matches the identical `{Controller, Player, ParentTargetController}` set and is disarmed by the same `chunk_declares_caster_subject` guard.
- `crates/engine/src/game/effects/change_zone.rs` — the `ParentTargetController`/`ParentTargetOwner` arms of `resolve_all`'s `player_scope` match, which the new `ScopedPlayer` arm mirrors (delegating to `resolve_effect_player_ref`).

## Gate A

`scripts/check-parser-combinators.sh` passes on the staged diff (exit 0, no violations). The one new parser detector `parse_hand_to_library_position` is built entirely from `nom` combinators (`tag`/`alt`/`value`/`opt`/`preceded`/`parse_possessive_determiner`), with no string-method dispatch, match-arm string literals, chained `if let Ok = tag(...)` blocks, verbatim-sentence equality, or hand-built `Effect::Unimplemented`.

## CR references

All verified against the current Comprehensive Rules:
- **CR 401.4** — "If an effect puts two or more cards in a specific position in a library at the same time, the owner of those cards may arrange them in any order." (the "in any order" whole-hand placement).
- **CR 608.2c** — later text modifies earlier ("…, then draws that many cards" anaphora over the just-moved count).
- **CR 603.2b** — "at the beginning of" phase/step triggers (the per-player draw-step trigger).
- **CR 102.1** — the active player is whose turn it is (binds "that player" to the scoped/active player).

## Verification

Built and tested with cargo (this contributor box has no Tilt):

```
tests/teferis_puzzle_box_4241.rs .... 4 passed; 0 failed
engine --lib ...................... 14555 passed; 0 failed
engine --test integration ......... 1712 passed; 0 failed
```

- New regression tests (`crates/engine/tests/teferis_puzzle_box_4241.rs`): the controller's own draw step; an opponent's draw step (P0's Box cycles **P1's** hand and P1 draws that many while **P0's hand is untouched** — proves the `ScopedPlayer` binding on both the move and the draw); an empty-hand no-op; and draw-count == hand-size (never `Fixed(1)`/0). All drive real trigger firing + resolution through the scenario runner, not a resolver shortcut. Plus a parser SHAPE unit test with a zero-`Unimplemented` reach-guard.
- Regression-ran the whole lib + integration suites: Whirlpool hand→library shuffle, Jace snapshots, Sanwell rest-on-bottom (#3267), and Blinkmoth Urn (#2900) all still pass — the shared `parse_put_ast` dispatch change is tightly guarded (requires "the cards {in|from} <possessive> hand … on the bottom/top"), so single-card and "put the rest" forms are untouched.

### DIFF
--- a/crates/engine/src/game/effects/change_zone.rs
+++ b/crates/engine/src/game/effects/change_zone.rs
@@ -1194,6 +1194,17 @@ pub fn resolve_all(
             ability,
             &TargetFilter::ParentTargetOwner,
         ),
+        // CR 603.2b + CR 102.1: "At the beginning of each player's draw step,
+        // that player puts the cards in their hand on the bottom of their
+        // library" (Teferi's Puzzle Box). The per-player trigger binds "that
+        // player" to `ScopedPlayer` (the active player whose step it is), so the
+        // whole-hand move must scan that player's hand, not the source
+        // controller's.
+        TargetFilter::ScopedPlayer => crate::game::targeting::resolve_effect_player_ref(
+            state,
+            ability,
+            &TargetFilter::ScopedPlayer,
+        ),
         _ => None,
     };
 
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -5778,6 +5778,19 @@ pub(super) fn parse_put_ast(
         }
     }
 
+    // CR 401.4 + CR 608.2c: "put the cards {in|from} <possessive> hand on the
+    // bottom/top [of <possessive> library] [in any order]" — whole-hand library
+    // reposition (Teferi's Puzzle Box). The "the cards {in|from} <possessive>
+    // hand" guard keeps single-card "put that card on the bottom" and "put the
+    // rest on the bottom" out of this branch (they lack the hand-origin phrase),
+    // so those still route to the positional `BottomOfLibrary`/`TopOfLibrary`
+    // forms below.
+    if let Ok((rest, _)) = tag::<_, _, OracleError<'_>>("put ").parse(lower) {
+        if let Some(position) = parse_hand_to_library_position(rest) {
+            return Some(PutImperativeAst::HandToLibraryPosition { position });
+        }
+    }
+
     let has_mass_zone_origin = (nom_primitives::scan_contains(lower, "all")
         || nom_primitives::scan_contains(lower, "each"))
         && nom_primitives::scan_contains(lower, "from");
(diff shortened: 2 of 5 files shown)


matthewevans (approved):
Reviewed current head `1a772e60649af22d8199f154d41865268ad3e6e8`.

This looks good to me. I checked the changed parser and resolver paths directly: the new whole-hand library-position form lowers to the existing `ChangeZoneAll` mass mover, preserves `library_position` without adding an erroneous shuffle, and reuses the existing `last_effect_count` / `EventContextAmount` channel for "draws that many cards". The added `ScopedPlayer` resolution in `ChangeZoneAll` is also the right runtime boundary for Teferi's Puzzle Box, because the trigger is per draw-step player rather than the source controller.

I also verified the cited rules locally: CR 401.4 for simultaneous placement in a library, CR 608.2c for the later "that many" text referring to the earlier move, CR 603.2b for beginning-of-step triggers, and CR 102.1 for the active player. The tests cover both parser shape and real trigger resolution, including the opponent draw-step case that proves the Box controller's hand is not touched.

Current-head CI/card-data was still running while I reviewed; the merge queue will continue to enforce that before landing.

PR #5203, opened 2026-07-06
fix(parser): order generic subtype-count after Coven/same-name conditions

dhgoal:
Tier: Frontier
Model: claude-opus-4-8
Thinking: high

## Summary

In `parse_you_control_condition`, the generic `you control N or more <subtype>` catch-all (`parse_you_control_subtype_count`) ran **before** the dedicated qualifier-phrase arms. Its bare `" or more "` split greedily consumed the qualifier phrase and dumped the entire tail into a stringly-typed subtype:

| Oracle text | Before (wrong) | After (correct) |
|---|---|---|
| `…three or more creatures with different powers` | `YouControlSubtypeCountAtLeast{ subtype: "creatures with different power" }` | `YouControlDifferentPowerCreatureCountAtLeast{ count: 3 }` |
| `…three or more lands with the same name` | `YouControlSubtypeCountAtLeast{ subtype: "lands with the same name" }` | `YouControlLandsWithSameNameAtLeast{ count: 3 }` |

No permanent has the subtype `"creatures with different power"` or `"lands with the same name"`, so the runtime count is **always 0** — the Coven (CR 702.152) and same-name conditions can never become true. Dawnhart Mentor's Coven bonus, Candletrap's Coven activation restriction, Endless Atlas's card draw, etc. silently never fire.

The fix moves the generic catch-all **below** the dedicated `YouControlDifferentPowerCreatureCountAtLeast` / `YouControlLandsWithSameNameAtLeast` arms (which already existed in the same function) so the specific typed parse wins. Parser-only; the target variants and their runtime handlers already ship.

**Cards fixed (6):** Ambitious Farmhand, Candletrap, Dawnhart Mentor, Sungold Sentinel (Coven); Endless Atlas, Sceptre of Eternal Glory (same name).

## Gate A

`./scripts/check-parser-combinators.sh` — clean. The change introduces **no** new string-matching primitives; it only reorders three existing `if let` arms. Three-dot diff of `crates/engine/src/parser/**` against `origin/main` grepped for forbidden methods (`.contains("`, `.split_once(`, `.strip_suffix(`, `.starts_with("`, `.find("`, …): **no matches**.

## Anchored on

- `crates/engine/src/parser/oracle_condition.rs:455` — dedicated `YouControlDifferentPowerCreatureCountAtLeast` arm (and `:460` for `YouControlLandsWithSameNameAtLeast`) that the generic catch-all was shadowing.
- `crates/engine/src/parser/oracle_condition.rs:472` — the relocated `parse_you_control_subtype_count` generic arm, now ordered after the specific arms.

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib -- parser::` — **7550 passed / 0 failed**, including the new fail-on-revert guard `qualifier_phrase_conditions_beat_generic_subtype_count` (reordering the generic arm back on top makes it fail).
- **Blast-radius proof:** rebuilt `oracle-gen` with the fix and regenerated `card-data.json` (35,397 cards) vs the clean-main baseline — **exactly 6 cards changed**, all listed above, each moving from the stringly-typed subtype dump to the correct dedicated variant. Zero collateral, zero over-match.

## Scope Expansion

None.

### DIFF
--- a/crates/engine/src/parser/oracle_condition.rs
+++ b/crates/engine/src/parser/oracle_condition.rs
@@ -438,9 +438,6 @@ fn parse_you_control_condition(text: &str) -> Option<ParsedCondition> {
     if let Some(subtypes) = parse_you_control_land_subtypes(text) {
         return Some(ParsedCondition::YouControlLandSubtypeAny { subtypes });
     }
-    if let Some((count, subtype)) = parse_you_control_subtype_count(text) {
-        return Some(ParsedCondition::YouControlSubtypeCountAtLeast { subtype, count });
-    }
     if let Some(count) = parse_numeric_threshold(
         text,
         "creatures you control have total power ",
@@ -464,6 +461,16 @@ fn parse_you_control_condition(text: &str) -> Option<ParsedCondition> {
     }
     if let Some(count) = parse_numeric_threshold(text, "you control ", " or more snow permanents") {
         return Some(ParsedCondition::YouControlSnowPermanentCountAtLeast { count });
+    }
+    // The generic bare-subtype catch-all must run AFTER the specific "creatures with
+    // different powers" / "lands with the same name" / "snow permanents" suffix arms
+    // above. Otherwise it greedily consumes those qualifier phrases via its bare
+    // `" or more "` split and dumps the whole phrase into a stringly-typed subtype
+    // (e.g. subtype "creatures with different power"), shadowing the dedicated
+    // YouControlDifferentPowerCreatureCountAtLeast / YouControlLandsWithSameNameAtLeast
+    // variants so those cards never reach the correct parse.
+    if let Some((count, subtype)) = parse_you_control_subtype_count(text) {
+        return Some(ParsedCondition::YouControlSubtypeCountAtLeast { subtype, count });
     }
     // "you control N or more [color] permanents" / "you control N or more [core type]s"
     if let Ok((rest, _)) = tag::<_, _, OracleError<'_>>("you control ").parse(text) {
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head. The parser ordering fix is at the right seam: the dedicated Coven / same-name / snow condition parsers run before the generic subtype-count catch-all, with sibling-preservation coverage. Exact-head CI is green and the parse-diff sticky reports no card-parse changes. Approving and enqueueing via the merge queue.

dhgoal:
Note on the empty `coverage-parse-diff` sticky (✓ No card-parse changes detected):

This is expected for **this** fix and is not a no-op. The corrected value is a `ParsedCondition` restriction variant (`YouControlSubtypeCountAtLeast` → `YouControlDifferentPowerCreatureCountAtLeast` / `YouControlLandsWithSameNameAtLeast`). The coverage overlay (`game/coverage.rs`) renders effect/trigger/`StaticCondition` details into `parse_details`, but it does not surface `ParsedCondition` restriction variants — so both the before and after render to the same overlay entry and `coverage-parse-diff` reports no delta. (This is also *why the misparse went undetected*: the card stays `supported` in coverage regardless of which condition variant it parses to.)

The change is real and verifiable outside the overlay:

- **card-data**: rebuilding `oracle-gen` and regenerating `card-data.json` changes **exactly 6 cards** (the 4 Coven cards + Endless Atlas / Sceptre of Eternal Glory), each moving from the stringly-typed `subtype:"creatures with different power"` / `"lands with the same name"` dump to the correct dedicated variant — zero collateral.
- **runtime**: the two variants evaluate on entirely different paths (`game/restrictions.rs`), so the previous parse counted permanents of a non-existent subtype (always 0 → Coven never satisfied) while the corrected parse counts distinct-power creatures.
- **test**: `qualifier_phrase_conditions_beat_generic_subtype_count` is a fail-on-revert AST assertion (reordering the generic arm back on top fails it), with regression guards proving a bare subtype still reaches the generic arm and snow-permanents keep their own variant.

PR #5206, opened 2026-07-06
fix(data): populate Alchemy spellbook lists from AtomicCards so DraftFromSpellbook works

davion-knight:
## Summary
Fixes a latent data-pipeline bug that left the **Alchemy spellbook** mechanic silently inert. `Effect::DraftFromSpellbook` (and the conjure-from-spellbook path) reads `GameObject::spellbook`, which is populated from `CardFace.metadata.spellbook` — but that field was **never populated in production**, so ~29 already-supported spellbook cards drafted from an empty list and did nothing.

**Root cause:** the `AtomicCard` deserialization struct had no field capturing MTGJSON's `relatedCards`, so `relatedCards.spellbook` was silently dropped on load; and per-set files (the only source `build_token_source_metadata` read) don't carry `spellbook` — only `AtomicCards.json` does.

**Fix (data-only, no parser/AST change):**
- `database/mtgjson.rs` — capture `relatedCards` on `AtomicCard` via the existing `SetRelatedCards` struct (`#[serde(default)]`).
- `bin/oracle_gen.rs` — `build_token_source_metadata` now harvests `spellbook` from the already-loaded `AtomicCardsFile` (mirroring `harvest_creature_subtypes_from_atomic`), keyed by the same `face_name.unwrap_or(name).to_lowercase()` convention the stamp lookup uses, "first non-empty wins". Token/scryfall sourcing untouched.

This revives the whole spellbook class (Tome of Gadwick, Boseiju Pathlighter, Ominous Traveler, Adaptive Armorer, …) to actually draft. **No card's parse signature or coverage status changes** — these cards already parse to `DraftFromSpellbook`; the fix restores their runtime behavior only.

## Files changed
- `crates/engine/src/database/mtgjson.rs` — `AtomicCard.related_cards` field + deser test
- `crates/engine/src/bin/oracle_gen.rs` — `merge_atomic_spellbooks` harvest + merge test
- `crates/engine/src/game/scenario.rs` — test-driver support for the interactive `WaitingFor::SpellbookDraft` (`ResolutionPolicy.spellbook_pick` + `.spellbook_pick()` builder + `drive_resolution` arm)
- `crates/engine/src/game/effects/spellbook_tests.rs` — runtime + provenance tests
- 9 test modules — mechanical `AtomicCard` struct-literal updates for the new field

## CR references
None — conjure/spellbook is a digital-only Alchemy mechanic with no Comprehensive Rules entry (per `game/effects/spellbook.rs`).

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high
Tier: Frontier

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — clean (exit 0; no parser change)
- `cargo test -p engine` — 0 failures, incl. the new tests
- `./scripts/gen-card-data.sh` — regenerated; `metadata.spellbook` now populated (Tome of Gadwick=12, Boseiju Pathlighter=15, Goblin Influx Array=15, Ominous Traveler=15). Byte-check: `parse_details`/`abilities`/`triggers`/`statics` **identical** vs baseline — only `metadata.spellbook` changed.
- `cargo coverage` — no delta (88.75%); spellbook cards' supported status unchanged (data-only, no coverage flip)

Tests added:
- `deserializes_related_cards_spellbook` (mtgjson.rs) — proves `relatedCards.spellbook` now deserializes (was silently dropped); non-vacuous negative twin.
- `build_token_source_metadata_harvests_spellbook_from_atomic` (oracle_gen.rs) — sets-less temp path so only the atomic fold can populate; **reverting the fold fails this test** (the pipeline-fix revert guard).
- `spellbook_pick_drives_the_draft_and_conjures_the_chosen_card` (spellbook_tests.rs) — drives the real interactive pipeline (activate → `SpellbookDraft` → `SubmitSpellbookDraft` → `complete_draft`) and asserts the chosen card enters hand; fails if the draft doesn't happen.
- `spellbook_draft_offers_only_the_activated_sources_list` — two sources with different spellbooks → asserts the offered list is the drafting source's own (provenance via `source_id`).

## Scope Expansion
None. (Data-source fix only. The `ConjureSource::Random` parser generalization that would additionally consume this data is a deliberate follow-up, not included here.)

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/bin/oracle_gen.rs
+++ b/crates/engine/src/bin/oracle_gen.rs
@@ -448,20 +448,39 @@ struct TokenSourceMetadata {
 
 fn build_token_source_metadata(
     mtgjson_path: &std::path::Path,
+    atomic: &engine::database::mtgjson::AtomicCardsFile,
 ) -> HashMap<String, TokenSourceMetadata> {
     let sets_dir = mtgjson_path
         .parent()
         .unwrap_or_else(|| std::path::Path::new("."))
         .join("sets");
 
-    if !sets_dir.exists() {
-        return HashMap::new();
+    let mut map: HashMap<String, TokenSourceMetadata> = HashMap::new();
+
+    // Per-set token/spellbook metadata requires local set files, so guard the
+    // set loop on the directory. It must NOT early-return the whole function —
+    // the Alchemy spellbook harvest below runs unconditionally so the
+    // Effect::DraftFromSpellbook faces are populated even when only
+    // AtomicCards.json is present locally.
+    if sets_dir.exists() {
+        merge_set_token_metadata(&sets_dir, &mut map);
     }
 
-    let mut map: HashMap<String, TokenSourceMetadata> = HashMap::new();
-    let entries = match std::fs::read_dir(&sets_dir) {
+    // Revive Effect::DraftFromSpellbook: source each face's Alchemy spellbook
+    // list from the already-loaded AtomicCards.json (relatedCards.spellbook),
+    // which serde previously dropped for lack of a capturing field.
+    merge_atomic_spellbooks(&mut map, atomic);
+
+    map
+}
+
+fn merge_set_token_metadata(
+    sets_dir: &std::path::Path,
+    map: &mut HashMap<String, TokenSourceMetadata>,
+) {
+    let entries = match std::fs::read_dir(sets_dir) {
         Ok(entries) => entries,
-        Err(_) => return HashMap::new(),
+        Err(_) => return,
     };
 
     for entry in entries.flatten() {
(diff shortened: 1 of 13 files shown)


matthewevans (approved):
Maintainer sweep approval: reviewed current head, CI is green, required parse-diff/proof gates are satisfied for this scope, and I did not find a blocking architecture or correctness issue in the changed surface.

volkkov (commented):
The happy path is clean. What happens if the third-party API returns 429 here? Might be worth handling that case explicitly.

PR #5212, opened 2026-07-06
fix(parser): decompose supertype in 'you control a <supertype> <type>' restrictions

dhgoal:
Tier: Frontier
Model: claude-opus-4-8
Thinking: high

## Summary

A cast/activation restriction of the form `you control a <supertype> <type>` fell through to the stringly-typed subtype catch-all in `parse_you_control_condition`, dumping the whole phrase into a non-existent subtype:

| Card | Oracle | before (wrong) | after (correct) |
|---|---|---|---|
| Blizzard | `Cast this spell only if you control a snow land` | `YouControlSubtypeCountAtLeast{ subtype:"snow land" }` | `QuantityComparison{ ObjectCount[Land, You, HasSupertype(Snow)] ≥ 1 }` |
| Goblin Ski Patrol | `Activate only … if you control a snow Mountain` | `YouControlSubtypeCountAtLeast{ subtype:"snow mountain" }` | `QuantityComparison{ ObjectCount[Subtype(Mountain), You, HasSupertype(Snow)] ≥ 1 }` |

No permanent has a subtype literally named `"snow land"` / `"snow mountain"`, so the restriction is **permanently unsatisfiable** — Blizzard can never be cast and Goblin Ski Patrol's ability can never activate. Both cards are currently unplayable.

The fix adds a supertype-prefix branch (gated on `nom_target::parse_supertype_prefix`) that reuses the existing `parse_type_phrase` combinator — the same one that decomposes `"basic land card"` into `Land` + `HasSupertype(Basic)` — to build a proper `TargetFilter`, scopes it to controller `You`, and emits the established `QuantityComparison` / `ObjectCount ≥ 1` presence building block (identical shape to the shipped Mox Opal / Confront the Assault restriction path). Parser-only; the `ObjectCount`-based restriction is already engine-resolved.

**Scope:** a general supertype-decomposition fix (any supertype × any type). Two cards currently match the pattern — both listed above — and both are presently broken; the mechanism auto-covers any future `you control a <supertype> <type>` restriction.

## Gate A

`./scripts/check-parser-combinators.sh` — clean. The new branch dispatches via the `parse_supertype_prefix` nom combinator and reuses `parse_type_phrase`; no string-matching primitives introduced. Three-dot diff of `crates/engine/src/parser/**` against `origin/main` grepped for forbidden methods (`.contains("`, `.split_once(`, `.starts_with("`, `.strip_suffix(`, …): **no matches**.

## Anchored on

- `crates/engine/src/parser/oracle_condition.rs:562` — the new supertype-prefix decomposition branch in `parse_you_control_condition`, placed before the `parse_core_type_word` / bare-subtype catch-all it was falling through.
- `crates/engine/src/parser/oracle_nom/target.rs:115` — `parse_supertype_prefix`, the existing combinator the branch gates on; decomposition delegates to `parse_type_phrase` (`oracle_target.rs`), reusing the canonical `<supertype> <type>` grammar rather than string-matching.

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib -- parser::` — **7567 passed / 0 failed**, including the new fail-on-revert test `you_control_supertype_permanent_decomposes_to_filter` (asserts both "snow land" → `TypeFilter::Land` and "snow mountain" → `TypeFilter::Subtype("Mountain")` produce `ObjectCount ≥ 1` with `controller: You` + `HasSupertype(Snow)`; the old parse returned `YouControlSubtypeCountAtLeast`, so reverting fails it).
- CR annotations verified against `docs/MagicCompRules.txt` (205.4a supertype list; 205.4g snow permanent).
- **Blast-radius proof:** rebuilt `oracle-gen` with the fix and regenerated `card-data.json` (35,397 faces) vs the clean-main baseline — **exactly 2 cards changed** (Blizzard, Goblin Ski Patrol), each from the unsatisfiable subtype dump to the correct Snow-supertype filter. Zero collateral, no over-match.

## Scope Expansion

None.

### DIFF
--- a/crates/engine/src/parser/oracle_condition.rs
+++ b/crates/engine/src/parser/oracle_condition.rs
@@ -463,6 +463,29 @@ fn parse_you_control_condition(text: &str) -> Option<ParsedCondition> {
     if let Some(count) = parse_numeric_threshold(text, "you control ", " or more snow permanents") {
         return Some(ParsedCondition::YouControlSnowPermanentCountAtLeast { count });
     }
+    // CR 205.4a: A counted "you control N or more <supertype> <type>s" restriction
+    // (two or more basic lands, two or more legendary creatures) must decompose the
+    // supertype into a `HasSupertype` object-count the same way the singular arm
+    // does — otherwise the bare-subtype catch-all below dumps "basic land" /
+    // "legendary creature" into a non-existent subtype and the restriction can never
+    // be satisfied. Runs before `parse_you_control_subtype_count`; the
+    // `parse_supertype_prefix` gate inside the helper keeps ordinary subtype counts
+    // ("two or more vampires") on the subtype path.
+    if let Ok((rest, _)) = tag::<_, _, OracleError<'_>>("you control ").parse(text) {
+        if let Ok((type_text, count_text)) = terminated(
+            take_until::<_, _, OracleError<'_>>(" or more "),
+            tag(" or more "),
+        )
+        .parse(rest)
+        {
+            if let Some(count) = parse_count_word(count_text) {
+                let singular = type_text.trim().trim_end_matches('.').trim_end_matches('s');
+                if let Some(cond) = parse_you_control_supertype_count(singular, count) {
+                    return Some(cond);
+                }
+            }
+        }
+    }
     // The generic bare-subtype catch-all must run AFTER the specific "creatures with
     // different powers" / "lands with the same name" / "snow permanents" suffix arms
     // above. Otherwise it greedily consumes those qualifier phrases via its bare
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** The same supertype decomposition bug remains in plural/count restrictions. Evidence: the new typed-filter path only runs under the singular `you control a/an ...` branch at `crates/engine/src/parser/oracle_condition.rs:560`, but `parse_you_control_subtype_count` still handles `you control N or more ...` by trimming the whole normalized phrase into a subtype at `crates/engine/src/parser/oracle_condition.rs:1243`; the new test only covers singular `you control a snow land` / `you control a snow mountain` at `crates/engine/src/parser/oracle_condition.rs:1533`. Why it matters: phrases like `you control two or more basic lands` or another counted `<supertype> <type>` restriction still become string subtypes such as `basic land`, which no permanent has, leaving those restrictions unsatisfiable. Suggested fix: route the plural/count arm through the same `parse_type_phrase`/`HasSupertype` object-count shape before the stringly subtype fallback, and add a counted supertype test.

dhgoal:
Thanks — fixed the sibling plural/count path (`70886eea`).

**What changed:** extracted the singular decomposition into a shared helper `parse_you_control_supertype_count(type_phrase, count)` and routed the plural `you control N or more <supertype> <type>s` arm through it **before** `parse_you_control_subtype_count`, splitting the count with nom `take_until`/`terminated` (no `split_once`). The `parse_supertype_prefix` gate inside the helper keeps ordinary subtype counts (`two or more vampires`) on the subtype path — I added a regression assertion for that. The test now covers singular `snow land` / `snow mountain` (count 1) and plural `two or more basic lands` (→ Land + HasSupertype(Basic), count 2) / `two or more legendary creatures` (→ Creature + HasSupertype(Legendary), count 2).

**One honest correction on card impact:** I verified against regenerated card-data that **no current card routes a plural supertype-count through this arm** — the only supertype-prefixed subtype dumps in the entire data are the two singular cards (Blizzard, Goblin Ski Patrol). The cited `Prairie Stream` / `Arwen's Gift` already parse correctly through a *different* path: Prairie Stream's `unless you control two or more basic lands` is handled by the enters-tapped-unless replacement parser (`oracle_replacement.rs`), whose condition already decomposes to `{ minimum: 2, filter: Typed{ Land, controller: You, HasSupertype(Basic) } }`. So they were never in the broken set.

Net: this commit is a **parse-level completeness fix** — it closes the plural `parse_you_control_condition` code path you flagged (proven by the test) and consolidates both arms behind one helper, but the PR's card-data blast radius remains the 2 singular cards. Happy to drop the plural arm if you'd rather keep the PR strictly to cards with a live delta, but it seemed worth closing the sibling gap while the helper is right here.

matthewevans (approved):
Maintainer re-review: the latest head covers both singular and counted supertype restrictions, preserves ordinary subtype counts, and the branch is clean against main. Approving for merge queue; pending checks can complete under auto-merge.

PR #5216, opened 2026-07-06
fix(parser): strip leading universal quantifier (all/each/every) in parse_type_phrase

dhgoal:
Tier: Frontier
Model: claude-opus-4-8
Thinking: high

## Summary

`parse_type_phrase_with_ctx` stripped leading `a`/`an` and `other`/`another`, but never the **universal quantifiers** `all` / `each` / `every`. The quantifier leaked into the type filter, leaving it empty/untyped:

| Card | Oracle | before (wrong) | after (correct) |
|---|---|---|---|
| Soulblast | `sacrifice all creatures you control` | cost filter `type_filters: [], controller: null` — **matches every object in play, including opponents'** | `[Creature], You` |
| Kaervek's Spite | `sacrifice all permanents you control` | `[], null` | `[Permanent], You` |
| Tomb of Urami | `Sacrifice all lands you control` | `[], null` | `[Land], You` |
| Mob Mentality | `Whenever all non-Wall creatures you control attack` | trigger mode `Unknown` (unclassified) | `Attacks`, `valid_card` = `Creature` + `Non(Subtype "Wall")`, controller `You` |

The Soulblast case is the most severe — an empty, uncontrolled sacrifice-cost filter matches **every** object on the battlefield rather than your own creatures.

The fix adds a nom `alt((tag("all "), tag("each "), tag("every ")))` strip in `parse_type_phrase_with_ctx`, guarded by the existing `starts_with_type_phrase_lead` (only strips when a real type word follows), advancing `pos` with **no** `FilterProp` added. Per CR 109.2 a universal quantifier ranges over every matching permanent, source included — so unlike `other`/`another` (which add `FilterProp::Another`) it is a semantic no-op on the filter. This is the central building-block fix: it corrects every `parse_type_phrase` consumer (targets, cost filters, trigger `valid_card`) at once.

## Gate A

`./scripts/check-parser-combinators.sh` — clean. Dispatch is nom `tag`/`alt`; the guard reuses the existing `starts_with_type_phrase_lead` helper. Three-dot diff of `crates/engine/src/parser/**` against `origin/main` grepped for forbidden methods (`.contains("`, `.split_once(`, `.starts_with("`, …): **no matches**.

## Anchored on

- `crates/engine/src/parser/oracle_target.rs:1914` — the new universal-quantifier strip in `parse_type_phrase_with_ctx`, placed before the `other`/`another` handling (which it deliberately does *not* mimic — no `Another`).
- `crates/engine/src/parser/oracle_target.rs:3291` — `starts_with_type_phrase_lead`, the existing guard so a bare quantifier without a following type word is left intact.

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib -- parser::` — **7568 passed / 0 failed**, including the new fail-on-revert test `parse_type_phrase_universal_quantifier_stripped_no_leak` (asserts `each/all/every <subtype> you control` → the bare `Subtype`, no space-containing subtype leak, no `Another`, controller `You`).
- CR annotation verified against `docs/MagicCompRules.txt` (109.2).
- **Blast-radius proof:** rebuilt `oracle-gen` with the fix and regenerated `card-data.json` (35,397 faces) vs the no-fix baseline — **exactly 4 cards changed** (Kaervek's Spite, Mob Mentality, Soulblast, Tomb of Urami), each a genuine correction. Zero collateral.

## Scope Expansion

None.

## Known adjacent (separate root cause, deliberately not in this PR)

A handful of static-anthem cards show the same *surface* symptom — a `Subtype("Each Vehicle")` / `Subtype("All Cats")` leak — e.g. Depala, Rin and Seri, Skeletal Swarming, Brightcap Badger (`"Each Vehicle you control gets …"` / `"All Cats you control are …"`). Those do **not** flow through `parse_type_phrase`; they hit a different root cause in the static-anthem / type-change subsystem (conditional-pump / type-change sub-parsers that bypass the `parse_continuous_subject_filter` quantifier strip). Fixing them means touching that broader subsystem across multiple sub-paths, so I kept this change surgical to the `parse_type_phrase` building block (verified zero-regression, 4-card delta) rather than bundle an unrelated subsystem change. Happy to follow up on the anthem path in a separate PR.

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -1927,25 +1927,37 @@ pub fn parse_type_phrase_with_ctx<'a>(
     ))
     .parse(&lower[pos..])
     {
-        if starts_with_type_phrase_lead(rest) {
+        // Strip the quantifier when a type-phrase lead follows directly OR through
+        // an "other"/"another" exclusion ("each other creature", "all other
+        // nonland permanents"); in the latter case only the quantifier is consumed
+        // here and the handler below adds `FilterProp::Another`.
+        let after_other = alt((tag::<_, _, OracleError<'_>>("other "), tag("another ")))
+            .parse(rest)
+            .map(|(r, _)| r)
+            .ok();
+        if starts_with_type_phrase_lead(rest)
+            || after_other.is_some_and(starts_with_type_phrase_lead)
+        {
             pos = lower.len() - rest.len();
         }
     }
 
     // Handle "other"/"another" prefix: "other creatures", "another creature",
-    // "other nonland permanents", "another target creature"
+    // "other nonland permanents", "another target creature". Reads from the
+    // current `pos` (not the raw trimmed head) so it composes with a universal
+    // quantifier already stripped above ("all other creatures" → Another + type).
     if tag::<_, _, OracleError<'_>>("other ")
-        .parse(lower_trimmed)
+        .parse(&lower[pos..])
         .is_ok()
     {
         properties.push(FilterProp::Another);
-        pos = offset + "other ".len();
+        pos += "other ".len();
     } else if tag::<_, _, OracleError<'_>>("another ")
-        .parse(lower_trimmed)
+        .parse(&lower[pos..])
         .is_ok()
     {
         properties.push(FilterProp::Another);
-        pos = offset + "another ".len();
+        pos += "another ".len();
     }
     // "another target [type]" — strip "target " after "another " so the type is reachable.
     if properties.contains(&FilterProp::Another) {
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** Universal quantifier stripping still misses the `all/each/every other <type>` class. Evidence: `crates/engine/src/parser/oracle_target.rs:1923` strips `all`/`each`/`every` only when the remainder already satisfies `starts_with_type_phrase_lead`, but that guard at `crates/engine/src/parser/oracle_target.rs:3291` does not accept `other`/`another`; the later `other` handler still checks `lower_trimmed` at `crates/engine/src/parser/oracle_target.rs:1937`, so `all other creatures` and `each other creature` bypass both strips and still parse as an empty/broad or leaked subtype filter. Why it matters: this is the same quantifier class the PR claims to generalize, and repo fixtures show production-relevant forms such as `each other creature` and `all other nonland permanents they control`. Suggested fix: after stripping an optional universal quantifier, parse `other`/`another` from the updated `pos` and add coverage for `all other <type>` / `each other <type>` so the filter carries the type plus `FilterProp::Another`.

dhgoal:
Thanks — done exactly as suggested (`f1889fe5`).

**What changed:** the quantifier strip now also fires when an `other`/`another` exclusion + type lead follows (consuming *only* the quantifier), and the `other`/`another` handler now reads from the updated `pos` instead of the raw trimmed head — so the two compose: `each other creature` → `Creature` + `FilterProp::Another`, `all other creatures` → same. Test extended to assert both forms carry the type and `Another`.

**One honest measurement on card impact:** I regenerated card-data and diffed — this commit changes **0 additional cards** (the PR's delta remains the 4 direct-consumer cards). I checked why: all **257** cards whose Oracle text contains a `all/each/every other <type>` phrase already parse correctly, because the universal quantifier is stripped **upstream at the effect layer** (`DamageAll` / `DestroyAll` / anthem type-change / etc.) before the remainder reaches `parse_type_phrase` — so `parse_type_phrase` currently only ever receives the already-quantifier-stripped `other <type>`, which the existing handler covered. e.g. Chandra's Ignition (`each other creature`) and Banishment (`all other nonland permanents`) both already carry `Another` correctly.

So this is a **building-block hardening**: `parse_type_phrase("each other creature")` was genuinely broken as a direct input (the new test fails on revert), and it's now correct + zero-regression, closing the class for any direct caller — but no live card was in the broken set. Flagging that transparently in case you'd rather I keep PRs to live deltas; happy either way.

matthewevans (approved):
Stale requested-changes finding is addressed at `f1889fe`: the universal quantifier strip now composes through `other`/`another`, and the added parser coverage reaches the broken `each other` / `all other` inputs. Parse-diff is limited to the claimed Mob Mentality trigger classification.

PR #5222, opened 2026-07-07
feat(parser): support “enter tapped this turn” (Due Respect, Nahiri’s Lithoforming)

codevector96:
## Summary

Implements the previously-unsupported "enter tapped this turn" replacement, so **Due Respect** ("Permanents enter tapped this turn. Draw a card.") and **Nahiri's Lithoforming** ("… Lands you control enter tapped this turn.") stop dropping that clause to `Effect::Unimplemented`.

The clause is a turn-duration floating replacement: on resolution it installs a global replacement that makes a filtered set of permanents enter the battlefield tapped, expiring at cleanup. It reuses the existing external "[objects] enter tapped" building block (the same `ReplacementDefinition{ event: ChangeZone, execute: SetTapState{Tap} }` shape that Authority of the Consuls already uses) — this is just its turn-scoped, spell-installed variant. **No new engine surface**: one new private parser function lifts the clause into the existing `Effect::AddTargetReplacement { target: None }` installer with an `EndOfTurn` expiry; the resolver, floating-replacement scan, controller-anchoring, and cleanup pruning are all unchanged.

Model: claude-opus-4-8
Tier: Frontier
Thinking: high

## Anchored on
- `crates/engine/src/parser/oracle_effect/mod.rs` — `try_parse_global_damage_modification_replacement`: the sibling that wraps a `parse_replacement_line` def in `Effect::AddTargetReplacement { target: None, expiry: EndOfTurn }`; the new function mirrors its shape (different seam: the pre-peel "this turn" scan block, since the duration is trailing).
- `crates/engine/src/parser/oracle_replacement.rs` — `parse_external_enters_tapped` / `build_external_entry_replacement`: the existing `ChangeZone` + `SetTapState{SelfRef, Tap}` builder reused as the detector.
- The existing `scan_contains_phrase(&original_lower, "this turn")` arms (`try_parse_play_from_exile`, `try_parse_additional_land_this_turn`): the new arm copies their `peel_ctx` optional/condition handling and `attach_unless_slots` return.

## Gate A

`scripts/check-parser-combinators.sh` passes on the staged diff (no violations). The new detection delegates entirely to `strip_trailing_duration` + `parse_replacement_line` (nom end-to-end) plus a typed `matches!` shape guard — no string-method dispatch, match-arm string literals, chained `if let Ok = tag(...)`, verbatim-sentence equality, or hand-built `Effect::Unimplemented`.

## CR references
- **CR 614.1d** — "[Objects] enter [the battlefield] …" replacement effects (the external enters-tapped class; distinct from the self form in 614.1c).
- **CR 514.2** — "this turn" effects end during cleanup (the `EndOfTurn` expiry).
- **CR 611.2a** — a resolution-fixed continuous effect (`AddTargetReplacement { target: None }`, controller anchored at install).
- **CR 109.4** — controller-relative object references ("Lands you control").

## Verification

Built and tested with cargo (this contributor box has no Tilt):

```
tests/enters_tapped_this_turn.rs ... 5 passed; 0 failed
engine --lib parser::oracle_effect ... 2496 passed; 0 failed
```

- New tests (`crates/engine/tests/enters_tapped_this_turn.rs`, typed fixtures, real cast/resolve pipeline): parser SHAPE for both cards (with a zero-`Unimplemented` reach-guard); a permanent entering after Due Respect resolves enters tapped while one that entered before is unaffected; the replacement expires at cleanup (later-turn entries untapped); and the load-bearing **Nahiri multi-authority** test — P0's land enters tapped while P1's land and P0's non-land do not (proving the `source_controller` + "Lands you control" filter).
- Fail-closed: any duration other than end-of-turn, or any non-`ChangeZone`/`SetTapState` shape, returns `None` and the clause stays `Unimplemented` — coverage flips to supported only for what's genuinely handled.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -5848,6 +5848,20 @@ pub(crate) fn parse_effect_clause(text: &str, ctx: &mut ParseContext) -> ParsedE
             }
             return attach_unless_slots(clause, None, unless_pay_deferred);
         }
+        // CR 614.1d + CR 514.2: "Permanents enter tapped this turn." / "Lands you
+        // control enter tapped this turn." — a turn-bound floating enters-tapped
+        // replacement (Due Respect, Nahiri's Lithoforming).
+        if let Some(effect) = try_parse_enter_tapped_set_this_turn(
+            TextPair::new(text, &original_lower),
+            ctx.card_name.as_deref().unwrap_or(""),
+        ) {
+            let mut clause = parsed_clause(effect);
+            peel_ctx.apply_optional(&mut clause.optional);
+            if clause.condition.is_none() {
+                clause.condition = peel_ctx.condition().cloned().or(unless_condition.clone());
+            }
+            return attach_unless_slots(clause, None, unless_pay_deferred);
+        }
     }
     let mut clause = parse_effect_clause_inner(&peeled_text, ctx);
     // Trial-parse fallback: peeling may have removed disambiguation signal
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Maintainer review: parser change is narrowly scoped, reuses the existing external-entry replacement parser and AddTargetReplacement resolver, has current parse-diff evidence for only Due Respect and Nahiri’s Lithoforming, and CI is green.

PR #5224, opened 2026-07-07
feat(engine): become-Equipment anthem + granted-equip runtime (Bram, Bludgeon Brawl) (CR 205.1a + 702.6)

minion1227:
## Summary
Adds engine support for the **become-Equipment anthem** — Bram, Baguette Brawler ("Each noncreature Food you control is an Equipment with equip {1} and \"Equipped creature gets +1/+1.\"") and Bludgeon Brawl ("Each noncreature, non-Equipment artifact is an Equipment with equip {X} and \"Equipped creature gets +X/+0,\" where X is that artifact's mana value.").

The whole static dropped: no handler recognized "Each `<subject>` is an Equipment with equip `{N}` and \"`<ability>`\"". This ships **both halves** — parser *and* the runtime that makes a statically granted Equip keyword actually function.

### Parser (`type_change.rs`)
`parse_becomes_equipment_with_ability` (dispatched next to the sibling Imprisoned `parse_enchanted_becomes_type_with_ability`) parses the subject, equip cost, and quoted ability into:
- `RemoveAllSubtypes{Artifact}` + `AddSubtype{Equipment}` — CR 205.1a (Equipment is an artifact subtype; setting it replaces the object's other artifact subtypes, e.g. Bram's Food → Equipment),
- `AddKeyword{Equip(cost)}` — CR 702.6,
- `GrantStaticAbility{...}` — the quoted "Equipped creature gets +N/+M" anthem, via the shared quoted-ability authority.

Reuses `parse_continuous_subject_filter`, `nom_primitives::parse_mana_cost`, and `parse_quoted_ability_modifications`.

### Runtime (`synthesis.rs`, `casting.rs`)
Card-load synthesis turns a **printed** Equip keyword into an activated `Effect::Attach` ability, but a **granted** Equip keyword never runs that conversion — so without runtime support the object would get the Equipment subtype and keyword but no equip ability (parses-but-doesn't-function). Fixed by:
- extracting `synthesis::equip_ability_for_keyword` (shared by `synthesize_equip` and the new runtime path; also sets `AbilityTag::Equip`, fixing a prior synthesis omission),
- adding `casting::runtime_granted_equip_abilities` — mirrors the existing cycling runtime-grant appender: reads the object's post-layer battlefield keywords, excludes printed equip (already in `obj.abilities`), and synthesizes the equip ability for each granted `Keyword::Equip`,
- chaining it LAST into both `activated_ability_definitions` and `activation_ability_definition` (index-consistent).

Granted equip is now offered and charged through the **normal `ActivateAbility` path**, sharing all printed-equip machinery.

### Dynamic mana value (Bludgeon Brawl)
Bludgeon Brawl binds X to "that artifact's mana value" (CR 202.3) — the Equipment's own mana value — for **both** the equip cost and the granted anthem. The handler lowers `equip {X}` to `Equip(SelfManaValue)` (concretized at activation, like a graveyard-grant `encore {X}, where X is its mana value`) and rebinds the anthem's `+X/+0` from the parser's default cost-X reference to `SelfManaValue`. The runtime resolves `SelfManaValue` against the granting Equipment (the static's source, not the equipped creature) — proven by a runtime test where a mana-value-3 Equipment boosts a 2/2 to 5/2.

## Files changed
- `crates/engine/src/database/synthesis.rs`
- `crates/engine/src/game/casting.rs`
- `crates/engine/src/game/engine_keyword_action_stack_tests.rs`

## Anchored on
- `crates/engine/src/parser/oracle_static/type_change.rs` `parse_enchanted_becomes_type_with_ability` — the sibling become-type-with-granted-ability handler; same compose pattern (type identity + `GrantAbility`), dispatched immediately after it.
- `crates/engine/src/game/casting.rs` `runtime_granted_cycling_abilities` — the runtime keyword-grant appender template the new `runtime_granted_equip_abilities` mirrors exactly.

## Gate A
```
```
(check-parser-combinators.sh — clean, exit 0)

## CR references
- `CR 205.1a` — setting an artifact subtype (Equipment) replaces the object's other artifact subtypes.
- `CR 702.6 / CR 702.6a` — Equip is an activated ability, "{cost}: Attach to target creature you control," activatable only as a sorcery.

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — clean (Gate A)
- `cargo test -p engine --lib parser::oracle_static` — 1014 pass / 0 fail
- `cargo test -p engine --lib database::synthesis` — 399 pass / 0 fail
- `cargo test -p engine --lib equip` — 146 pass / 0 fail; `game::layers` — 206 pass / 0 fail
- Parser tests: `becomes_equipment_with_equip_cost_and_granted_ability` (Bram fixed-cost — affected filter + Equipment subtype + Equip keyword + granted anthem) and `bludgeon_brawl_dynamic_equip_cost_and_anthem_bind_source_mana_value` (Bludgeon — `Equip(SelfManaValue)` + anthem rebound to `SelfManaValue`).
- Runtime tests: `granted_equip_keyword_offers_functional_equip_ability` (a granted Equip keyword yields a cost-bearing, `AbilityTag::Equip`, `Effect::Attach` ability through `activated_ability_definitions`) and `granted_equip_anthem_self_mana_value_reads_equipment_mana_value` (the granted anthem's `SelfManaValue` resolves against the Equipment: MV-3 Equipment → 2/2 becomes 5/2).

## Scope Expansion
Adds runtime support so a statically granted Equip keyword synthesizes a functional equip activated ability (`synthesis::equip_ability_for_keyword` + `casting::runtime_granted_equip_abilities`), reusing the printed-equip path. Required because the parser alone would produce a non-functional "Equipment" — see the Runtime section.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/database/synthesis.rs
+++ b/crates/engine/src/database/synthesis.rs
@@ -476,32 +476,39 @@ pub fn synthesize_basic_land_mana(face: &mut CardFace) {
 /// authority that sets both the display flag and pushes
 /// `ActivationRestriction::AsSorcery` so the runtime legality gate enforces
 /// timing at activation time.
+/// CR 702.6: Build the equip activated ability for a single `Keyword::Equip(cost)`
+/// — `{cost}: Attach this permanent to target creature you control`, activatable
+/// only as a sorcery. Returns `None` for any other keyword. Shared by card-load
+/// synthesis (`synthesize_equip`) and the battlefield runtime granted-keyword
+/// appender (`runtime_granted_equip_abilities`), so a printed and a statically
+/// granted equip keyword (Bram, Bludgeon Brawl) produce the identical ability.
+pub(crate) fn equip_ability_for_keyword(keyword: &Keyword) -> Option<AbilityDefinition> {
+    let Keyword::Equip(cost) = keyword else {
+        return None;
+    };
+    let mut def = AbilityDefinition::new(
+        AbilityKind::Activated,
+        Effect::Attach {
+            attachment: TargetFilter::SelfRef,
+            target: TargetFilter::Typed(TypedFilter::creature().controller(ControllerRef::You)),
+        },
+    )
+    .cost(AbilityCost::Mana { cost: cost.clone() })
+    // CR 702.6a: "Activate only as a sorcery."
+    .sorcery_speed();
+    // CR 702.6a: tag the synthesized ability as the Equip action so cost
+    // reductions and Equipment-source mana restrictions recognize it (matching
+    // the parser's `try_parse_equip` path).
+    def.ability_tag = Some(AbilityTag::Equip);
+    Some(def)
+}
+
 pub fn synthesize_equip(face: &mut CardFace) {
     let equip_abilities: Vec<AbilityDefinition> = face
         .keywords
         .iter()
-        .filter_map(|kw| {
-            if let Keyword::Equip(cost) = kw {
-                Some(
-                    AbilityDefinition::new(
-                        AbilityKind::Activated,
-                        Effect::Attach {
-                            attachment: TargetFilter::SelfRef,
-                            target: TargetFilter::Typed(
-                                TypedFilter::creature().controller(ControllerRef::You),
-                            ),
-                        },
-                    )
-                    .cost(AbilityCost::Mana { cost: cost.clone() })
-                    // CR 702.6a: "Activate only as a sorcery."
-                    .sorcery_speed(),
-                )
-            } else {
-                None
-            }
-        })
+        .filter_map(equip_ability_for_keyword)
         .collect();
-
     face.abilities.extend(equip_abilities);
 }
 
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -75,6 +75,34 @@ fn runtime_granted_cycling_abilities(
         .collect()
 }
 
+/// CR 702.6: An `Equip` keyword granted at runtime by a static ability (Bram,
+/// Bludgeon Brawl's "… is an Equipment with equip {N} …") does not pass through
+/// card-load synthesis, so its equip activated ability must be synthesized live
+/// from the object's post-layer keyword set. `obj.keywords` is battlefield-
+/// authoritative (AddKeyword grants land there); printed equip keywords are
+/// excluded because card-load synthesis already turned them into an
+/// `obj.abilities` entry, so re-synthesizing them would double-offer equip.
+fn runtime_granted_equip_abilities(
+    state: &GameState,
+    source_id: ObjectId,
+) -> Vec<AbilityDefinition> {
+    let Some(obj) = state.objects.get(&source_id) else {
+        return Vec::new();
+    };
+    // CR 702.6: Equip functions only while its source is on the battlefield.
+    if obj.zone != Zone::Battlefield {
+        return Vec::new();
+    }
+    obj.keywords
+        .iter()
+        .filter(|keyword| {
+            matches!(keyword, Keyword::Equip(_))
+                && !obj.base_keywords.iter().any(|printed| printed == *keyword)
+        })
+        .filter_map(|keyword| crate::database::synthesis::equip_ability_for_keyword(keyword))
+        .collect()
+}
+
 /// CR 604.1 (seam 4: activated-ability-on-grant): synthesize graveyard activated
 /// abilities (Encore, Scavenge) for keywords granted to a graveyard card by a
 /// static. The `AddKeyword` layer seam installs only the keyword + triggers, so a
@@ -191,6 +219,10 @@ pub fn activated_ability_definitions(
             .chain(runtime_granted_top_of_library_plot_abilities(
                 state, source_id,
             ))
+            // CR 702.6: statically granted equip (Bram, Bludgeon Brawl) chained
+            // LAST — the identical append order is REQUIRED in
+            // `activation_ability_definition` so `ability_index` stays consistent.
+            .chain(runtime_granted_equip_abilities(state, source_id))
             .enumerate()
             .map(|(offset, ability)| (printed_len + offset, ability)),
     );
@@ -220,6 +252,7 @@ fn activation_ability_definition(
             .chain(runtime_granted_top_of_library_plot_abilities(
                 state, source_id,
             ))
+            .chain(runtime_granted_equip_abilities(state, source_id))
             .nth(offset)?
     };
     if let Some(ref cost) = ability.cost {
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -119,7 +119,20 @@ fn runtime_granted_equip_abilities(
                 unconsumed_printed.remove(index);
                 return None;
             }
-            crate::database::synthesis::equip_ability_for_keyword(keyword)
+            crate::database::synthesis::equip_ability_for_keyword(keyword).map(|mut ability| {
+                // CR 202.3 + CR 118.9: Bludgeon Brawl grants `equip {X}` where X
+                // is the artifact's mana value, so the keyword carries the
+                // `ManaCost::SelfManaValue` placeholder. Concretize it to the
+                // source's actual mana value HERE — otherwise the payment path
+                // treats `SelfManaValue` as `{0}` and the equip is effectively
+                // free.
+                if let Some(cost) = ability.cost.take() {
+                    ability.cost = Some(super::keywords::resolve_self_mana_in_ability_cost(
+                        state, source_id, &cost,
+                    ));
+                }
+                ability
+            })
         })
         .collect()
 }
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -93,13 +93,34 @@ fn runtime_granted_equip_abilities(
     if obj.zone != Zone::Battlefield {
         return Vec::new();
     }
+    // CR 702.6a: a permanent may have more than one equip ability, and each is
+    // independently activatable. Card-load synthesis already turned every PRINTED
+    // Equip keyword into an `obj.abilities` entry, so subtract printed equips by
+    // OCCURRENCE (not value-wide membership): consume one printed instance per
+    // matching live keyword, and synthesize the rest. This keeps a granted
+    // Equip {1} offered even when the object also prints an identical Equip {1}.
+    let mut unconsumed_printed: Vec<&Keyword> = obj
+        .base_keywords
+        .iter()
+        .filter(|keyword| matches!(keyword, Keyword::Equip(_)))
+        .collect();
     obj.keywords
         .iter()
-        .filter(|keyword| {
-            matches!(keyword, Keyword::Equip(_))
-                && !obj.base_keywords.iter().any(|printed| printed == *keyword)
+        .filter_map(|keyword| {
+            if !matches!(keyword, Keyword::Equip(_)) {
+                return None;
+            }
+            if let Some(index) = unconsumed_printed
+                .iter()
+                .position(|printed| *printed == keyword)
+            {
+                // A printed equip already lives in `obj.abilities`; consume it so
+                // any additionally granted copies are still synthesized below.
+                unconsumed_printed.remove(index);
+                return None;
+            }
+            crate::database::synthesis::equip_ability_for_keyword(keyword)
         })
-        .filter_map(|keyword| crate::database::synthesis::equip_ability_for_keyword(keyword))
         .collect()
 }
 
--- a/crates/engine/src/game/engine_keyword_action_stack_tests.rs
+++ b/crates/engine/src/game/engine_keyword_action_stack_tests.rs
@@ -987,7 +987,7 @@ fn granted_equip_keyword_offers_functional_equip_ability() {
         "granted Equipment subtype missing"
     );
     assert!(
-        obj.keywords.iter().any(|k| matches!(k, Keyword::Equip(_))),
+        crate::game::keywords::has_keyword_kind(obj, crate::types::keywords::KeywordKind::Equip),
         "granted Equip keyword missing"
     );
 
--- a/crates/engine/src/game/engine_keyword_action_stack_tests.rs
+++ b/crates/engine/src/game/engine_keyword_action_stack_tests.rs
@@ -928,3 +928,86 @@ fn issue_3660_finalize_copy_retarget_stashes_offers_on_deferred_pause() {
         Some(remaining.as_slice()),
     );
 }
+
+/// CR 702.6: An Equip keyword granted at runtime by a static ability (Bram,
+/// Bludgeon Brawl: "… is an Equipment with equip {N} …") must produce a real,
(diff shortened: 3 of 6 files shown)


matthewevans (changes requested):
**[MED]** The parser advertises a broader Bram/Bludgeon class than it actually accepts. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:685` and `crates/engine/src/parser/oracle_static/tests.rs:14779` name Bram / Bludgeon Brawl as the class, but the parser only splits on the singular fixed-cost form `" is an equipment with equip "` at `type_change.rs:696` and then requires `parse_mana_cost` to consume the whole cost at `type_change.rs:701`. Why it matters: the code/comment/test shape implies the become-Equipment anthem class is covered, but variable-cost siblings such as equip-X/where-X forms cannot reach this parser, so coverage can look broader than the implementation. Suggested fix: either narrow the parser/comments/tests to the fixed-mana-cost Bram-shaped class, or extend the parser with composable axes for variable equip costs before claiming the Bludgeon class.

**[MED]** The runtime test does not drive the claimed functional equip path. Evidence: `crates/engine/src/game/engine_keyword_action_stack_tests.rs:932` says the granted equip ability must be offered and charged through `ActivateAbility`, but the test stops after calling `activated_ability_definitions` and asserting a cost-bearing `Effect::Attach` shape at `engine_keyword_action_stack_tests.rs:998`. Why it matters: this would still pass if selecting the generated ability failed in the production activation/payment/targeting path, so it does not prove the new runtime seam works end-to-end. Suggested fix: add a runtime test that activates the generated equip ability through `GameAction::ActivateAbility`, pays the mana cost, selects/uses a legal creature target, resolves, and asserts the Equipment becomes attached.

Separate evidence blocker: this PR changes engine/parser source, but there is no `<!-- coverage-parse-diff -->` sticky comment yet and Rust/card-data checks are still pending for current head `3907be964a4f1895770d23cb9a2cf0c206be170f`. I cannot approve or enqueue it until those current-head signals exist and are clean.

matthewevans (changes requested):
**[HIGH]** Identical printed and granted Equip instances are collapsed. Evidence: `crates/engine/src/game/casting.rs:96` filters live Equip keywords with `!obj.base_keywords.iter().any(|printed| printed == *keyword)`, while `docs/MagicCompRules.txt:3941` says a permanent with multiple equip abilities may activate any of them. Why it matters: an Equipment with printed `Equip {1}` that is also granted `Equip {1}` should expose two equip abilities, but this code drops the granted instance. Suggested fix: subtract printed Equip instances by occurrence/count, not by value-wide membership, before synthesizing runtime-granted Equip abilities.

**[MED]** The parser still claims the Bludgeon Brawl class while only accepting the fixed-cost Bram shape. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:685` and `crates/engine/src/parser/oracle_static/tests.rs:14779` name Bludgeon Brawl, but `type_change.rs:696` requires the singular `" is an equipment with equip "` split and `type_change.rs:701` rejects `{X}, where X is ...` because `parse_mana_cost` leaves a non-empty remainder. Why it matters: coverage and reviewer expectations read this as a general become-Equipment anthem parser, but dynamic equip-cost siblings still cannot reach it. Suggested fix: either narrow the comments/tests/dispatch text to the fixed-mana-cost Bram-shaped class, or implement the variable-cost axis before claiming Bludgeon Brawl.

**[MED]** The runtime test does not exercise the activation path it claims to prove. Evidence: `crates/engine/src/game/engine_keyword_action_stack_tests.rs:998` only calls `activated_ability_definitions` and then asserts the generated ability shape through line 1012. Why it matters: this would still pass if `GameAction::ActivateAbility` failed to select the generated index, charge the mana, target a creature, or resolve the attach effect. Suggested fix: add an end-to-end runtime test that activates the generated equip ability via `GameAction::ActivateAbility`, pays/selects targets through the normal flow, resolves, and asserts the Equipment is attached.

**[LOW]** The new CR annotation cites the wrong rule for subtype replacement. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:685` and `type_change.rs:688` cite `CR 301.5c` for replacing artifact subtypes, but `docs/MagicCompRules.txt:1595` says `CR 301.5c` is about Equipment attachment legality/SBAs; `docs/MagicCompRules.txt:1398` is the subtype-replacement rule. Why it matters: CR comments are treated as verified rules evidence in this engine, so an incorrect citation weakens future reviews. Suggested fix: remove `CR 301.5c` from the subtype-replacement annotation or reserve it only for code that enforces Equipment attachment legality.

minion1227:
Thanks — addressed everything.

**[HIGH] (matthewevans + gemini) identical printed/granted equip collapsed** — fixed in `5071eaf1b`. `runtime_granted_equip_abilities` now subtracts printed equips by **occurrence**, not value-wide membership: it consumes one printed instance per matching live keyword and synthesizes the rest. So an object printing `Equip {1}` that is *also* granted `Equip {1}` exposes both (CR 702.6a — each of multiple equip abilities is independently activatable). Added a test `identical_printed_and_granted_equip_both_offered` asserting exactly two equip abilities are offered.

**CI `clippy::redundant_closure`** — resolved by the same rewrite (the `filter_map` closure now carries the occurrence logic).

**[MED] (matthewevans) parser advertises a broader class than it accepts / [MED] (gemini) trailing clause breaks quoted-ability parsing** — both were on the pre-dynamic-X head; the later commits `50f61c0e2` + `0f951066d` already cover the variable-cost sibling. The handler now splits the quoted ability on its **closing quote** first (so the trailing `, where X is that artifact's mana value` no longer reaches the quoted-ability parser), lowers `equip {X}` to `Equip(SelfManaValue)`, and rebinds the granted anthem's `+X/+0` to `SelfManaValue`. **Bludgeon Brawl now parses in full** — covered by `bludgeon_brawl_dynamic_equip_cost_and_anthem_bind_source_mana_value` (parser) and `granted_equip_anthem_self_mana_value_reads_equipment_mana_value` (runtime: MV-3 Equipment → 2/2 becomes 5/2).

oracle_static 1014, equip 147, casting 648, layers 206 — all pass; parser gate clean.

matthewevans (changes requested):
**[HIGH]** Concretize Bludgeon Brawl's `Equip(SelfManaValue)` before synthesizing the runtime ability. Evidence: `crates/engine/src/game/casting.rs:122` passes the live `Keyword::Equip(ManaCost::SelfManaValue)` directly into `equip_ability_for_keyword`, while `crates/engine/src/game/mana_payment.rs:516` treats `SelfManaValue` as immediately payable/free and `crates/engine/src/game/keywords.rs:330` documents that activated ability costs must resolve self-mana placeholders before legality or payment. Why it matters: Bludgeon Brawl's equip cost should be the artifact's mana value, but this path can offer a cost-bearing-looking Equip ability whose payment is actually `{0}`. Suggested fix: resolve the runtime-granted Equip keyword cost with the existing self-cost authority before calling `equip_ability_for_keyword`, and add a Bludgeon-style activation/payability test that proves a mana-value-N Equipment requires/pays `{N}` rather than just asserting the parsed AST/offered ability shape.

minion1227:
Fixed the **[HIGH]** in `bb4d18a0c`.

**Concretize `Equip(SelfManaValue)` before offering.** `runtime_granted_equip_abilities` now resolves the synthesized equip ability's cost through `keywords::resolve_self_mana_in_ability_cost(state, source_id, &cost)` when it builds it, so the `ManaCost::SelfManaValue` placeholder (Bludgeon Brawl's `equip {X}`) is replaced with the source's actual mana value at the point of offering — the payment path can no longer see it as `{0}`. (The activation path already re-resolves, harmlessly, since a concrete cost has no placeholder left.)

Added the requested **payability test** `granted_equip_self_mana_value_cost_is_concretized_to_mana_value`: a mana-value-4 Equipment granted `Equip(SelfManaValue)` presents a concrete `{4}` cost (`AbilityCost::Mana { ManaCost::Cost { generic: 4 } }`), not a placeholder.

Two earlier comments now stale on the current head: the identical-printed/granted-equip **[HIGH]** was fixed by the occurrence-based dedup in `5071eaf1b` (test `identical_printed_and_granted_equip_both_offered`), and the trailing-clause **[MED]** by the split-on-closing-quote in the dynamic-X commit.

equip suite 148 pass; engine-authorities + parser-combinator gates clean (exit 0).

matthewevans (changes requested):
**[MED]** The granted-Equip runtime path still is not exercised end to end. Evidence: `crates/engine/src/game/engine_keyword_action_stack_tests.rs:998` only calls `activated_ability_definitions` and asserts the synthesized `Effect::Attach`/cost shape through line 1012; the new `SelfManaValue` test likewise stops at the offered ability cost at `engine_keyword_action_stack_tests.rs:1178`. Why it matters: this can still pass if `activation_ability_definition` selects the wrong appended runtime index, payment/targeting rejects the generated ability, or the attach effect never resolves, so the PR does not prove the claimed “offered and charged through the normal ActivateAbility path.” Suggested fix: add a runtime test that activates the generated equip ability via `GameAction::ActivateAbility`, pays/selects a legal creature target through the normal flow, resolves it, and asserts `attached_to` points at that creature.

**[LOW]** The become-Equipment parser doc still cites the wrong rule for artifact-subtype replacement. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:685` and `type_change.rs:688` cite `CR 301.5c` as the rule for replacing artifact subtypes, but the replacement behavior is implemented at `type_change.rs:741` under `CR 205.1a`; the prior review already identified `CR 301.5c` as attachment legality, not subtype replacement. Why it matters: CR comments are used as rules evidence in this engine, so a wrong citation makes later reviews trust the wrong rule. Suggested fix: remove `CR 301.5c` from the parser documentation, or reserve it for code that enforces Equipment attachment legality.

minion1227:
Addressed both outstanding items from the latest review (head now `52c976210`):

**[MED] granted-Equip runtime path now exercised end to end.** Added `granted_equip_attaches_through_full_activation_pipeline` in `engine_keyword_action_stack_tests.rs`. It drives the full pipeline through `GameRunner::activate` (the real `GameAction::ActivateAbility` path): announce → pay `{1}` from the pool → target a creature-you-control → resolve → assert `attached_to == Some(bear)` and the creature's `attachments` contains the Equipment (CR 702.6a). A regression in runtime-index selection, mana payment, targeting, or the Attach resolution now fails this test, not just the AST-shape assertions.

**[LOW] CR citation corrected.** Removed `CR 301.5c` from the `parse_becomes_equipment_with_ability` doc — verified against `docs/MagicCompRules.txt:1595`, 301.5c is Equipment *attachment legality*, not subtype replacement. The doc now cites only `CR 205.1a` (subtype replacement, line 1398) + `CR 702.6`.

The earlier HIGH items (occurrence-based dedup of printed vs. granted equip; `SelfManaValue` cost concretization) were resolved in `5071eaf1b` / `bb4d18a0c` with their own unit tests (`identical_printed_and_granted_equip_both_offered`, `granted_equip_self_mana_value_cost_is_concretized_to_mana_value`).

Verified locally (Tilt down): new test passes; `cargo clippy -p engine --all-targets --features proptest -D warnings` clean; `cargo fmt --all` clean.

matthewevans (changes requested):
**[MED]** Reject unrecognized text after the quoted Equipment ability instead of silently accepting it. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:703` captures everything after the closing quote, but lines 704-711 only use it to detect the Bludgeon `where X is ... artifact's mana value` binding; when that binding is absent, any remaining tail is ignored and the parser still returns a supported `StaticDefinition` at line 756. Why it matters: a future "is an Equipment with equip ... and "...", <extra rider>" card would have rules-bearing text dropped while coverage reports the card as parsed. Suggested fix: require the trimmed tail to be empty or fully consumed by a typed `where X is ...` parser, otherwise return `None`/strict-fail, and add a negative parser test for an unrecognized post-quote rider.

minion1227:
Both outstanding [MED]s addressed (and rebased onto current `main`, so the parse-diff refreshes).

**[MED] fail closed on unrecognized trailing text** (`92597148e`) — `parse_becomes_equipment_with_ability` now declines (returns `None`) when the text after the closing quote is non-empty and is NOT the recognized CR 202.3 `where X is that artifact's mana value` binding. A rider like `… and "…", and it gains flying.` is reported unsupported instead of silently dropping the rider. Regression: `becomes_equipment_rejects_unrecognized_trailing_rider` (rider fails closed; the fixed-cost empty-tail and where-X forms still parse).

**[MED] exercise the granted-Equip path end to end** (`e620dacdc`) — `granted_equip_attaches_through_full_activation_pipeline` drives the whole seam via `GameRunner::activate`: announce → pay {1} from the pool → target a creature you control → resolve → assert the Equipment's `attached_to` points at the creature and the creature carries it (CR 702.6a). This catches a runtime-index, payment, targeting, or Attach-resolution regression that the shape-only tests would miss.

Rebased onto current `upstream/main` to clear branch staleness. Verified: parser-combinator + engine-authorities gates clean (exit 0); oracle_static 1016, equip 150, granted_equip 5 pass.

matthewevans (changes requested):
**[MED]** The dynamic Bludgeon tail still does not fail closed when extra rules text follows the recognized `where X is ... artifact's mana value` binding. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:710` only checks `scan_contains(tail_lower, "where x is ")` plus `scan_contains(tail_lower, "artifact's mana value")`, and `type_change.rs:718` accepts any non-empty tail when those substrings are present. Why it matters: a dynamic become-Equipment line such as `..., where X is that artifact's mana value, and it gains flying` would still export as fully supported while silently dropping the extra rider, so coverage can go green with missing behavior. Suggested fix: parse the tail as an exact recognized `where X is that artifact's mana value` clause, requiring only optional punctuation/whitespace after it; otherwise return `None`/strict-fail, and add the negative test for a dynamic tail with an extra rider.

minion1227:
Tightened in `a6cfa6a1b`.

**[MED] fail closed on text AFTER the binding too.** The prior fix used two `scan_contains` substring checks, which accepted a trailing rider once the binding's substrings were present. Replaced with an exact, full-consumption match:

```rust
let dynamic_self_mana_value =
    all_consuming(tag::<_, _, OracleError<'_>>("where x is that artifact's mana value"))
        .parse(tail_core)
        .is_ok();
```

Now any tail that is not *exactly* the CR 202.3 binding fails closed — whether the binding is **absent** or **followed by an extra rider** (`…, where X is that artifact's mana value, and it gains flying`). Extended the regression `becomes_equipment_rejects_unrecognized_trailing_rider` with the rider-after-binding case; Bram (empty tail) and Bludgeon (exact binding) still parse.

Verified: oracle_static 1016 pass; parser-combinator + engine-authorities gates clean (exit 0).

matthewevans (approved):
Maintainer review: current head addresses the prior Bram/Bludgeon runtime and parser blockers. Parse diff is scoped to the two intended cards; granted Equip now synthesizes through the activated-ability path, concretizes SelfManaValue, and has end-to-end activation coverage.

PR #5226, opened 2026-07-07
fix: require host_peer_id to delete P2P draft backups

RealDiligent:
## Summary

**Root cause:** `DELETE /p2p-draft-backup/{code}` only validated the 6-character draft code. Anyone who learned or guessed a code could wipe another host'\''s recovery snapshot. POST already enforced `host_peer_id` ownership on overwrite via `guard_p2p_backup_overwrite`, but DELETE had no equivalent check.

**Fix:** Require a `host_peer_id` query parameter on DELETE, validate its shape with `validate_p2p_backup_host_peer_id`, load the stored row, and reject with 403 unless the peer id matches the backup owner. The P2P draft host client now passes its peer id on cleanup.

**Impact:** Prevents grief-deletion of P2P draft recovery snapshots without the host peer id. Backward-incompatible for any non-client callers that DELETE without `host_peer_id` (intentional — those callers were unauthenticated).

## Risk / tradeoffs

- Old DELETE calls without `host_peer_id` now receive 400. Only the phase client uses this endpoint today.
- GET remains unauthenticated (draft progress recovery); secrets are already redacted per #5053.

## Test plan

- [x] `validate_p2p_backup_host_peer_id` unit test in `server-core`
- [ ] CI `cargo nextest` + `phase-server` compile

### DIFF
--- a/client/src/adapter/p2p-draft-host.ts
+++ b/client/src/adapter/p2p-draft-host.ts
@@ -1421,9 +1421,11 @@ export class P2PDraftHost {
   private async cleanupServerBackup(): Promise<void> {
     if (!this.backupEndpoint || !this.draftCode) return;
     try {
-      await fetch(`${this.backupEndpoint}/p2p-draft-backup/${this.draftCode}`, {
-        method: "DELETE",
-      });
+      const params = new URLSearchParams({ host_peer_id: this.hostPeer.id });
+      await fetch(
+        `${this.backupEndpoint}/p2p-draft-backup/${this.draftCode}?${params}`,
+        { method: "DELETE" },
+      );
     } catch {
       // Best-effort cleanup
     }
--- a/crates/phase-server/src/admin.rs
+++ b/crates/phase-server/src/admin.rs
@@ -1,4 +1,4 @@
-use axum::extract::{Path, State};
+use axum::extract::{Path, Query, State};
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
 use axum::Json;
@@ -8,6 +8,7 @@ use tracing::{info, warn};
 
 use server_core::{
     guard_p2p_backup, guard_p2p_backup_overwrite, redact_p2p_backup_snapshot_secrets,
+    validate_p2p_backup_host_peer_id,
 };
 
 use crate::AppState;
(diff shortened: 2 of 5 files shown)


matthewevans (changes requested):
**[MED]** The security fix is not tested through the public DELETE endpoint. Evidence: `crates/phase-server/src/admin.rs:176` now requires `Query<P2pBackupDeleteQuery>` and checks the stored owner before `delete_p2p_backup`, but the PR only adds a helper-level blank-id test at `crates/server-core/src/p2p_backup_guard.rs:249`; there is no route/handler test proving missing or wrong `host_peer_id` returns 400/403 and leaves the row intact. Why it matters: the vulnerability is at the unauthenticated HTTP boundary, so helper coverage would not catch a future handler regression that deletes by draft code alone. Suggested fix: add a phase-server endpoint/handler test that creates a backup, attempts DELETE with no peer id and with a mismatched peer id, asserts the status, and verifies the stored backup still exists.

**[LOW]** The PR carries unrelated generated changelog state. Evidence: `client/public/changelog.json:1`, `client/public/changelog-meta.json:2`, and `scripts/changelog/state.json:2` add release id 165 for “Morph face-down casting arrives,” which is unrelated to P2P backup delete authorization. Why it matters: merging this security PR would publish unrelated release metadata and advance changelog state outside its scope. Suggested fix: drop the changelog files from this branch.

RealDiligent:
Thanks for the review — both points addressed in 5b639f5:

**[MED] HTTP boundary tests:** Added `p2p_backup_delete_tests` in `crates/phase-server/src/main.rs` with three route-level cases:
1. DELETE without `host_peer_id` → 400, row intact
2. DELETE with mismatched `host_peer_id` → 403, row intact
3. DELETE with matching `host_peer_id` → 200, row removed

Each test seeds a backup, hits the public DELETE route over TCP (same pattern as `admin_auth_tests`), and asserts both status and DB state.

**[LOW] Changelog files:** This branch never carried those files — `git diff origin/main` is limited to the P2P backup auth changes + the new tests (4 files in the original commit, 1 in the follow-up). No changelog JSON/state on this branch.

CI re-running on the new commit; will watch for green.

matthewevans (approved):
Maintainer sign-off: current-head review clean; public DELETE boundary tests now cover missing, mismatched, and matching host_peer_id paths; scope and security pre-checks verified.

PR #5227, opened 2026-07-07
fix: rate-limit TURN credential minting to prevent quota abuse

RealDiligent:
## Summary

**Root cause:** `GET /turn-credentials` on the lobby Worker mints 24-hour Cloudflare Realtime TURN credentials with no authentication, no per-client rate limit, and `ALLOWED_ORIGINS = *`. Any caller could burn relay quota; symmetric-NAT peers then fall back to STUN-only and cannot connect.

**Fix:**
- Per-IP sliding-window rate limit (default 30 mints/hour, configurable via `TURN_RATE_LIMIT_PER_HOUR`)
- Reject browser requests whose `Origin` header is present but not on the allowlist (missing `Origin` still allowed for Tauri/desktop)
- Tighten `ALLOWED_ORIGINS` to `phase-rs.dev`, `preview.phase-rs.dev`, and local Vite dev hosts

**Impact:** Reduces unauthenticated quota-burn attacks on production TURN relay capacity. Legitimate clients remain unaffected; over-limit callers receive HTTP 429 with `Retry-After`.

## Risk / tradeoffs

- Rate limiting is per-isolate (best-effort within a Worker instance); determined attackers can still spread load, but casual scraping/burning is blocked.
- Browsers from non-listed origins are rejected when they send an `Origin` header.

## Test plan

- [x] `lobby-worker` unit tests for rate limiter (`turn-rate-limit.test.mjs`)
- [ ] CI lobby-worker + frontend checks

### DIFF
--- a/lobby-worker/wrangler.toml
+++ b/lobby-worker/wrangler.toml
@@ -69,10 +69,11 @@ invocation_logs = false
 TURN_KEY_ID = "97c7984d6c155084f9d0fdf8090e02f8"
 # Credential lifetime in seconds (Cloudflare max 48h). 24h covers any session.
 TURN_TTL_SECONDS = "86400"
-# Origin allowlist for the /turn-credentials fetch. "*" is fine during rollout
-# (TURN creds are short-lived; worst case is quota use). Tighten for prod, e.g.
-# "https://phase-rs.dev,http://localhost:5173".
-ALLOWED_ORIGINS = "*"
+# Max TURN credential mints per client IP per hour (abuse guard).
+TURN_RATE_LIMIT_PER_HOUR = "30"
+# Origin allowlist for the /turn-credentials fetch. Missing Origin is still
+# allowed (Tauri/desktop). Browsers with a present but disallowed Origin get 403.
+ALLOWED_ORIGINS = "https://phase-rs.dev,https://preview.phase-rs.dev,http://localhost:5173,http://127.0.0.1:5173"
 
 # The TURN API token is a SECRET — never put it here. Set it with:
 #   npx wrangler secret put TURN_KEY_API_TOKEN
--- a/lobby-worker/src/turn-rate-limit.ts
+++ b/lobby-worker/src/turn-rate-limit.ts
@@ -1,0 +1,46 @@
+export interface RateLimitConfig {
+  maxRequests: number;
+  windowMs: number;
+}
+
+interface Bucket {
+  count: number;
+  resetAt: number;
+}
+
+/** In-memory per-key sliding window limiter (per isolate). */
+export function createRateLimiter(config: RateLimitConfig) {
+  const buckets = new Map<string, Bucket>();
+
+  return {
+    check(
+      key: string,
+      now = Date.now(),
+      maxRequests = config.maxRequests,
+    ): { allowed: true } | { allowed: false; retryAfterSeconds: number } {
+      const entry = buckets.get(key);
+      if (!entry || now >= entry.resetAt) {
+        buckets.set(key, { count: 1, resetAt: now + config.windowMs });
+        return { allowed: true };
+      }
+      if (entry.count >= maxRequests) {
+        return {
+          allowed: false,
+          retryAfterSeconds: Math.max(1, Math.ceil((entry.resetAt - now) / 1000)),
+        };
+      }
+      entry.count += 1;
+      return { allowed: true };
+    },
+
+    reset(): void {
+      buckets.clear();
+    },
+  };
+}
+
+export function parseTurnRateLimitPerHour(raw: string | undefined): number {
+  const parsed = Number(raw ?? "30");
+  if (!Number.isFinite(parsed)) return 30;
+  return Math.min(120, Math.max(1, Math.floor(parsed)));
+}
--- a/lobby-worker/src/turn-rate-limit.ts
+++ b/lobby-worker/src/turn-rate-limit.ts
@@ -8,6 +8,19 @@ interface Bucket {
   resetAt: number;
 }
 
+const MAX_BUCKETS = 10_000;
+
+function pruneBuckets(buckets: Map<string, Bucket>, now: number): void {
+  for (const [key, entry] of buckets) {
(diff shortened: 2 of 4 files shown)


RealDiligent:
Addressed review feedback in latest commits (rebased onto main):

**Superagent (P2):** clientIp() now uses only CF-Connecting-IP for rate-limit keys — removed the spoofable X-Forwarded-For fallback.

**Gemini (memory growth):** 	urn-rate-limit.ts prunes expired buckets on every check and caps the map at 10k entries (with eviction of oldest resetAt).

CI re-running.

RealDiligent:
Quick status update:

- Addressed Gemini HIGH finding by adding bucket pruning + hard cap in 	urn-rate-limit.ts.
- Addressed Superagent P2 by removing X-Forwarded-For fallback; rate limiting now keys only on CF-Connecting-IP.
- CI is fully green on current head commit.

Ready for maintainer review/merge.

lobby-worker/wrangler.toml
```diff
@@ -69,10 +69,11 @@ invocation_logs = false
...
 TURN_TTL_SECONDS = "86400"
-# Origin allowlist for the /turn-credentials fetch. "*" is fine during rollout
-# (TURN creds are short-lived; worst case is quota use). Tighten for prod, e.g.
-# "https://phase-rs.dev,http://localhost:5173".
-ALLOWED_ORIGINS = "*"
+# Max TURN credential mints per client IP per hour (abuse guard).
+TURN_RATE_LIMIT_PER_HOUR = "30"
+# Origin allowlist for the /turn-credentials fetch. Missing Origin is still
+# allowed (Tauri/desktop). Browsers with a present but disallowed Origin get 403.
+ALLOWED_ORIGINS = "https://phase-rs.dev,https://preview.phase-rs.dev,http://localhost:5173,http://127.0.0.1:5173"
```
matthewevans, line 76: My main question here is that if a missing origin is still allowed, how big of a net benefit is this PR?

RealDiligent:
Good question. The two guards are complementary:

**Rate limit (primary abuse control):** caps credential mints per `CF-Connecting-IP` (30/hr default). This is what stops quota-burning scripts regardless of client type — browser, curl, or scripted fetch.

**Origin allowlist (secondary, browser-specific):** only applies when an `Origin` header is *present*. Missing `Origin` is intentionally allowed because Tauri/desktop fetches do not send one; blocking those would break the desktop app.

What the allowlist still buys us for web:
- A random third-party site cannot drive-by `fetch()` the worker from a user's browser session (browsers always send `Origin` on cross-origin GETs).
- Without it, any malicious page a user visits could mint TURN creds in the background using the user's IP bucket — burning relay quota even before rate limits trip on a distributed basis.

So: missing-Origin allowance is required for desktop; rate limiting is the universal backstop; origin enforcement closes the browser embedding vector that rate limits alone don't fully cover (many users / many origins / same IP).

Happy to tighten further (e.g. require a custom header from our client) if you want stronger binding than Origin alone.

matthewevans:

This current head (`cd43e34aebfa3f794d40bdbbf1fe961cb88beebf`) was triaged as a frontend-only change (`lobby-worker/src/turn-rate-limit.ts`, `lobby-worker/src/turn.ts`, `lobby-worker/test/turn-rate-limit.test.mjs`, `lobby-worker/wrangler.toml`) by `RealDiligent`. The local frontend-review allowlist does not include this author, so this route does not perform an implementation-diff review or approve the PR.

matthewevans:

This current head (`cd43e34aebfa3f794d40bdbbf1fe961cb88beebf`) was triaged as a frontend-only change (`lobby-worker/src/turn-rate-limit.ts`, `lobby-worker/src/turn.ts`, `lobby-worker/test/turn-rate-limit.test.mjs`, `lobby-worker/wrangler.toml`) by `RealDiligent`. The local frontend-review allowlist does not include this author, so this route does not perform an implementation-diff review or approve the PR.

matthewevans:

This current head (`cd43e34aebfa3f794d40bdbbf1fe961cb88beebf`) was triaged as a frontend-only change (`lobby-worker/src/turn-rate-limit.ts`, `lobby-worker/src/turn.ts`, `lobby-worker/test/turn-rate-limit.test.mjs`, `lobby-worker/wrangler.toml`) by `RealDiligent`. The local frontend-review allowlist does not include this author, so this route does not perform an implementation-diff review or approve the PR.

Makeinu1:
Thanks for identifying the TURN credential quota-abuse issue in #5227. Current main now has native Workers Rate Limiting bindings for directory endpoints, so I opened #9209 as a current-main alternative that reuses that infrastructure and leaves ALLOWED_ORIGINS unchanged. It credits #5227 and is intentionally scoped to TURN minting only; happy to defer if maintainers prefer evolving #5227.

matthewevans:
Current-head maintainer routing for `cd43e34aebfa3f794d40bdbbf1fe961cb88beebf`:

#9209 has now merged to `main`. It addresses this PR's core TURN quota-abuse report with the lobby Worker's native `TURN_LIMIT` binding, separate production/preview counters, and a fail-closed 503 when the limiter is unavailable. Thank you for identifying and working on the issue; #9209 explicitly credits #5227.

This branch is now conflicting in `lobby-worker/src/turn.ts` because it adds a separate per-isolate limiter at the same minting point. Bringing it current would combine two competing limiters, so I am not merging `main` into this branch or approving it. The implementations are not identical: #5227 proposes 30 mints/hour and a present-`Origin` rejection policy; #9209 uses a native 30-per-60-second limit and deliberately leaves `ALLOWED_ORIGINS` unchanged. The hourly budget and Origin policy need their own maintainer decision if still desired. In particular, a missing `Origin` remains allowed for desktop clients, so the Origin policy is a browser-only guard rather than the universal quota control.

The existing `defer-fe` intake route still applies to this author, and this is a routing disposition, not an implementation review. I am leaving the PR open and deferred; no branch update or merge is queued. A maintainer can decide whether a focused follow-up for the remaining policy choices is warranted.

PR #5232, opened 2026-07-07
fix(tokens): role tokens grant their abilities and resolve face art

JacobWoodson:
Role tokens (Monster Role from Monstrous Rage, Royal Role from Royal
Treatment, ...) granted none of their effects and rendered with no art.
Both stem from the same naming mismatch: the parser names the token by
its full "<Role> Role" printed name (e.g. "Monster Role"), but the
consumers keyed off the bare role word ("Monster") used by the
MTGJSON-derived catalog presets.

- No effects: predefined_role_token_spec matched only "Monster", so a
  token named "Monster Role" dispatched to None and carried no static --
  no +1/+1, no trample/ward. Strip the trailing " Role" before matching.

- No art: a Role token is one face of a two-Role DFC ("Monster //
  Sorcerer"), so its source card links to BOTH face presets. That skips
  find_exact_token_ref's single-preset fast path and falls to a body
  match, where "Monster Role" != the preset's "Monster" left the token
  with no image ref. Reconcile the " Role" suffix in token_body_matches
  so the correct face (Monster, not Sorcerer) resolves.

Existing tests masked the bug (bare "Royal" name; the end-to-end section
19 test only checked attach, not the buff). Add reproduction tests for
both, and harden that test to assert the enchanted creature is actually
buffed (Royal Treatment: 2/2 -> 3/3 with ward).

### DIFF
--- a/crates/engine/src/game/effects/token.rs
+++ b/crates/engine/src/game/effects/token.rs
@@ -2567,8 +2567,14 @@ fn wicked_role_spec() -> RoleSpec {
 ///
 /// All Role tokens share the `Role` subtype, so dispatch must be by display
 /// name — subtype alone cannot distinguish the seven variants.
+///
+/// CR 111.10: a Role token's printed name is "<Role> Role" (e.g. "Monster Role"),
+/// which is exactly what the parser/token creation assigns as the display name.
+/// Strip that trailing " Role" before matching so real tokens dispatch correctly;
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Maintainer review: right seam for Role token predefined ability dispatch and token art identity matching; CR 111.10 checked; tests cover real token creation, attachment buff behavior, and DFC face art selection.

PR #5236, opened 2026-07-07
fix(engine): resolve mulligan bottoming at each player's declare point (CR 103.5b)

rykerwilliams:
## Summary

- Serum Powder's "any time you could mulligan" ability surfaced a deeper bug: the engine deferred **all** bottom-cards resolution (CR 103.5's "puts a number of cards equal to the number of times that player has taken a mulligan on the bottom of their library") to a single batch phase that only ran once *every* player at the table had finished mulliganing. Per CR 103.5b, each player's bottoming obligation must resolve at **their own** declare point (choosing to keep, or invoking a "you could mulligan" action), independent of other players.
- Folds bottoming into a per-player `MulliganDecisionPhase::BottomCards { count, then }` sub-phase on each `MulliganDecisionEntry`, driven by the existing `prepaid_mulligan_bottoms` ledger. The ledger resets to 0 on every real `Mulligan` (a fresh redraw invalidates prior bottoming credit, CR 103.5) but accumulates — never resets — across repeated `UseSerumPowder` activations at the same mulligan count, so cycling Serum Powder never double-charges an already-paid obligation.
- Deletes `WaitingFor::MulliganBottomCards`, `state.final_mulligan_counts`, `record_final_count`, and `enter_bottom_phase`/`_public` — bottoming is now resolved per-entry, so there's no separate batch phase to advance to.
- Threads the new `phase` field through engine dispatch, elimination pruning, AI candidate-generation/reject/search paths, a manabrew-compat prompt DTO extension (new `MulliganUseSerumPowder` output variant + `excluded_card_id` on the put-back input, with `response_output_matches_waiting` made phase-aware), and the frontend `WaitingFor` union, help sheet, and prompt rendering.
- Follow-up commit excludes the earmarked Serum Powder object from the in-repo web client's bottom-cards picker (the engine already rejected selecting it, but the UI let a player try and hit a confusing dead-end).

## CR references

CR 103.3 (pre-mulligan shuffle), CR 103.5 (core bottom-count rule), CR 103.5b (declare-point timing for "any time you could mulligan" actions), CR 103.5c (free-first-mulligan discount), CR 800.4a (elimination pruning), CR 201.2 (Serum Powder name-match). `start_mulligan`'s doc comment previously cited CR 103.4 (starting life total, unrelated) — corrected.

## Test plan

[...]

PR #5304, opened 2026-07-07
docs(skill): fold stale MulliganBottomCards reference into MulliganDecision

rykerwilliams:
## Summary

- Split out from #5236 (mulligan bottoming per-declare-point fix), which removes `WaitingFor::MulliganBottomCards` — bottoming is now a `MulliganDecisionPhase::BottomCards` sub-phase of the existing `MulliganDecision` variant.
- `.claude/skills/add-interactive-effect/SKILL.md`'s reference table still had a separate `MulliganBottom` row pointing at the now-removed variant. Folds that note into the existing `Mulligan` row instead of leaving a stale second row.
- This is its own PR because `.claude/skills/**` is a hard-stop path for the PR review loop per `.agents/pr-review-policy.toml` and can't be bundled with engine-code PRs.

## Test plan

- [x] Doc-only change, no code affected.

### DIFF
--- a/.claude/skills/add-interactive-effect/SKILL.md
+++ b/.claude/skills/add-interactive-effect/SKILL.md
@@ -357,8 +357,7 @@ For interactive replacements, you need to:
 
 | Effect | WaitingFor | Complexity |
 |--------|-----------|------------|
-| **Mulligan** | `MulliganDecision { player }` | Simple — keep/mulligan |
-| **MulliganBottom** | `MulliganBottomCards { player, count }` | Simple — cards to bottom |
+| **Mulligan** | `MulliganDecision { player }` | Simple — keep/mulligan (bottoming is a `MulliganDecisionPhase::BottomCards` sub-phase of the same variant) |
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
Per `.agents/pr-review-policy.toml`, `.claude/skills/**` is a hard-stop path for the PR review loop. This PR changes `.claude/skills/add-interactive-effect/SKILL.md`, so I cannot review or enqueue it in this sweep.

Please remove or split out the skill-file change from this PR; once the hard-stop path is gone, the normal review loop can process it.

rykerwilliams:
@matthewevans this is the skill change split out from #5236 . I think it is a relevant and related change that should be accepted if the other PR is.

rykerwilliams:
This PR exists specifically to isolate the `.claude/skills/add-interactive-effect/SKILL.md` change split out of #5236, since `.claude/skills/**` is a hard-stop path for the automated PR review loop per `.agents/pr-review-policy.toml`. The bot's CHANGES_REQUESTED review is expected here — this PR will always match that hard-stop by design, so it needs manual human review/merge rather than the automated pipeline.

PR #5315, opened 2026-07-07
fix(engine): apply Ad Nauseam's repeat-loop reveal/hand/life-loss immediately per iteration

### LINKED ISSUE #1032: [Card Bug] Ad Nauseam does not draw any cards
### Card(s)

Ad Nauseam

### Build/version

v0.1.35 6263238

### Game mode

P2P

### Actual behavior

When casting Ad Nauseam, it checked the top of my deck repeatedly but never put it into my hand. It saw a Flooded Strand every time I clicked "repeat," but never added the card to my hand. When I selected "end," it put the Flooded Strand into my hand but no other cards, and then lost me 0 life a bunch of times in a row.

### Expected behavior

Each time that the player clicks "repeat," it should reveal the card, add the card to their hand, and make them lose life equal to its CMC. The player should also lose the life as they add each card to their hand, not all at once at the end. 

### Steps to reproduce

Just cast Ad Nauseam in any board state. This issue happened multiple times for me.

_No response_

rykerwilliams:
## Summary

- Fixes #1032 — Ad Nauseam's "reveal top card, put into hand, lose life, you may repeat this process" loop silently deferred each accepted iteration's card-to-hand move and life loss instead of applying it immediately, only draining everything in one batch when the controller finally declined.
- Root cause: the `RepeatDecision` accept handler (`engine_resolution_choices.rs`) re-entered `resolve_ability_chain` without resetting `state.waiting_for` away from the just-answered `RepeatDecision` prompt. The stale value fooled `waits_for_resolution_choice` into deferring that iteration's `ChangeZone`/`LoseLife` sub-chain into `pending_continuation` instead of running it, and deferred pairs accumulated (not clobbered) until decline drained them all at once.
- Fix is a one-line `set_priority(state, player)` reset, mirroring the sibling `decline` branch (already does this via `finish_with_continuation`) and the analogous `OptionalEffectChoice` resume handler (`engine_payment_choices.rs`), both of which already reset priority before re-entering resolution. CR 107.1c + CR 608.2c annotated and verified against `docs/MagicCompRules.txt`.
- Class-level fix: covers every `RepeatContinuation::ControllerChoice` card whose per-iteration body has more than one instruction (a `sub_ability`), not just Ad Nauseam — the existing single-leaf-effect unit test never exercised this path, which is why it went uncaught.

## Test plan

- [x] New integration test `ad_nauseam_repeat.rs`: discriminating assertion checks hand size/life total *between* each accepted repeat (not just final aggregate — final totals are identical whether the bug is present or not, since life loss is additive regardless of timing). Confirmed this test fails on the unfixed code (`left: 1, right: 2`) and passes with the fix.
- [x] Sibling repeat-mechanism tests (`another_round_repeat`, `sin_spiras_punishment_repeat`, `claim_jumper_repeat`) unaffected, 9/9 pass.
- [x] Existing lib unit tests (`repeat_until_controller_choice_prompts_each_iteration`, `repeat_until_paused_resume_resets_prompt_after_inner_choice`, `token_choice_seed_survives_pending_repeat_until_drain`) unaffected, 3/3 pass.
- [x] `cargo fmt --all` clean.
- [x] `cargo clippy -p engine --all-targets -- -D warnings` zero warnings.
- [x] Went through two rounds of plan review and one implementation review before merge (all clean; one non-blocking informational note about theoretical edge-case coverage, judged low-risk by architectural equivalence to already-tested sibling repeat variants).

### DIFF
--- a/crates/engine/src/game/engine_resolution_choices.rs
+++ b/crates/engine/src/game/engine_resolution_choices.rs
@@ -965,6 +965,20 @@ pub(super) fn handle_resolution_choice(
             GameAction::DecideOptionalEffect { accept },
         ) => {
             if accept {
+                // CR 608.2c + CR 107.1c (issue #1032): reset to `Priority`
+                // BEFORE re-entering the chain, mirroring the `decline`
+                // branch's `finish_with_continuation` reset below and
+                // `handle_optional_effect_choice`'s `set_active_priority`
+                // reset (engine_payment_choices.rs). Without this,
+                // `state.waiting_for` is still the just-answered
+                // `RepeatDecision`, which `waits_for_resolution_choice`
+                // (effects/mod.rs) matches — the ChangeZone/LoseLife
+                // sub-chain following this iteration's RevealTop is then
+                // wrongly deferred into `pending_continuation` (accumulating
+                // there via `append_to_sub_chain`) instead of resolving
+                // immediately, and only drains in one batch when the
+                // controller eventually declines.
+                set_priority(state, player);
                 // Re-resolve one more process pass. `ability` retains
                 // `repeat_until: Some(ControllerChoice)`, so this hits the
                 // `repeat_until` dispatch, runs `resolve_chain_body` once, and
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed the current head. The fix is at the repeat-resolution seam, the priority reset matches the existing continuation-drain pattern, and the integration test discriminates the original batching bug by asserting state after each accepted repeat.

PR #5319, opened 2026-07-07
feat(engine): loyalty-ability activation cost modifiers (Eidolon of Obstruction)

joaovictor712:
## What

Makes **Eidolon of Obstruction** — "Loyalty abilities of planeswalkers your opponents control cost {1} more to activate" — parse and function end to end. It previously dropped the static to `Effect::Unimplemented`, so opponents' loyalty abilities were never taxed.

Loyalty abilities are activated abilities (CR 606.1), so a "cost {N} more/less to activate" static targets them like any other activated ability. Three composed pieces:

- **Parser** (`oracle_static/dispatch.rs`): the existing "Activated abilities of [subject] cost {N} …" static (Training Grounds / Skyseer's Chariot) is parameterized on its leading noun so it also accepts "Loyalty abilities of [subject]", emitting `ReduceAbilityCost { keyword: "loyalty" }`. Only the noun axis varies, so it's one `alt` — not a new handler.
- **Runtime gate** (`casting.rs`): `apply_static_activated_ability_cost_reduction` matches a `keyword == "loyalty"` static against an ability whose cost `is_loyalty_ability_cost` (loyalty abilities are identified by their `AbilityCost::Loyalty` cost, not an `AbilityTag`).
- **Cost application** (`casting.rs`, CR 118.7): a loyalty ability's printed cost has no mana component, so `increase_generic_in_cost` now **adds** a generic-mana component (wrapping in a `Composite`) instead of no-oping. This also fixes the latent gap where a cost-raise static silently failed to tax any activated ability whose cost is `{T}`/sacrifice-only.

## Tests

- Parser test pins the `keyword: "loyalty"` mode for Eidolon's line.
- Runtime test drives the production seam `apply_cost_reduction` and asserts an opponent's loyalty ability gains `{1}`, while the controller's own loyalty ability (opponent-scoped filter) and a non-loyalty ability (keyword mismatch) are untaxed.

Verified locally: `cargo fmt`, full `engine` lib test suite (15,625 passed / 0 failed), and `clippy -p engine --lib -D warnings` clean.

## CR references
CR 606.1 (loyalty abilities are activated abilities) · CR 601.2f + CR 118.7 (cost modification adds generic mana).

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -14696,7 +14696,17 @@ pub fn handle_activate_ability(
         // both fall through to the unchanged paths. SelfRef removal is excluded by
         // the walkers. `{X}`-mana removals were already caught by the X detour
         // above, so any mana leg seen here is non-X (mutually exclusive residuals).
-        if find_non_self_battlefield_removal_cost(cost).is_some() {
+        //
+        // CR 118.7 + CR 606.4: A loyalty ability taxed by a cost-raise static
+        // (Eidolon of Obstruction) reaches here as `Composite { Loyalty, Mana }`
+        // via `handle_activate_loyalty`'s delegation. It must ALSO hoist the mana
+        // leg to `enter_payment_step` and defer the loyalty counter cost as the
+        // `ManaLeg` residual — otherwise the composite falls through to
+        // `pay_ability_cost`, which pays the loyalty counters and then fails on the
+        // unaffordable mana, leaving the planeswalker with a free loyalty change.
+        if find_non_self_battlefield_removal_cost(cost).is_some()
+            || crate::types::ability::is_loyalty_ability_cost(cost)
+        {
             if let Some((mana_cost, remaining)) = casting_costs::extract_mana_leg(cost) {
                 let mut pending_leg = PendingCast::new(source_id, CardId(0), resolved, mana_cost);
                 pending_leg.activation_cost = remaining;
@@ -15618,6 +15628,29 @@ fn added_generic_mana_cost(amount: u32) -> AbilityCost {
             generic: amount,
         },
     }
+}
+
+/// CR 118.7 + CR 601.2f + CR 606.1: True when an active cost-modifier static
+/// (Eidolon of Obstruction) adds a mana component to an otherwise mana-free
+/// loyalty ability. Such an ability can no longer use the loyalty fast path
+/// (`handle_activate_loyalty`, which pays only loyalty counters and never mana);
+/// the caller routes it through the general activated-ability flow instead,
+/// which prompts for the added mana, pays the loyalty counters, records the
+/// CR 606.3 activation, and enforces the once-per-turn gate. A bare `Loyalty`
+/// cost that stays bare after applying every modifier is untaxed and keeps the
+/// fast path (zero behavior change for the common case).
+pub(crate) fn loyalty_ability_gains_mana_tax(
+    state: &GameState,
+    ability_def: &AbilityDefinition,
+    player: PlayerId,
+    source_id: ObjectId,
+) -> bool {
+    if !matches!(ability_def.cost, Some(AbilityCost::Loyalty { .. })) {
+        return false;
+    }
+    let mut probe = ability_def.clone();
+    apply_cost_reduction(state, &mut probe, player, source_id);
+    !matches!(probe.cost, Some(AbilityCost::Loyalty { .. }))
 }
 
 /// CR 601.2f: Apply self-referential cost reduction to an ability definition's cost.
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -15575,15 +15575,49 @@ fn increase_generic_in_cost(cost: &mut AbilityCost, amount: u32) {
         } => {
             *generic = generic.saturating_add(amount);
         }
+        // A pre-resolution placeholder mana cost (`NoCost`, `SelfManaCost`, …) or a
+        // `ManaDynamic` cost carries no concrete generic component to grow here; it
+        // is concretized on its own path, so leave it untouched.
+        AbilityCost::Mana { .. } | AbilityCost::ManaDynamic { .. } => {}
         AbilityCost::Composite { costs } => {
-            if let Some(sub) = costs
-                .iter_mut()
-                .find(|c| matches!(c, AbilityCost::Mana { .. }))
-            {
+            if let Some(sub) = costs.iter_mut().find(|c| {
+                matches!(
+                    c,
+                    AbilityCost::Mana {
+                        cost: ManaCost::Cost { .. }
+                    }
+                )
+            }) {
                 increase_generic_in_cost(sub, amount);
+            } else {
+                // CR 118.7: no concrete mana component to grow — add one so the
+                // increase still applies (e.g. a Composite of only `{T}`/sacrifice).
+                costs.push(added_generic_mana_cost(amount));
             }
         }
-        _ => {} // Non-mana costs unaffected
+        // CR 118.7 + CR 606.1: A non-mana cost (a loyalty ability's `Loyalty` cost,
+        // a bare `{T}` / sacrifice / pay-life cost) has no generic mana to grow, so
+        // a raise must ADD a generic-mana component. Wrap the existing cost in a
+        // `Composite` with the added `{amount}` — this is what makes Eidolon of
+        // Obstruction actually tax an opponent's loyalty ability by {1}.
+        _ => {
+            let existing = std::mem::replace(cost, AbilityCost::Composite { costs: Vec::new() });
+            if let AbilityCost::Composite { costs } = cost {
+                costs.push(existing);
+                costs.push(added_generic_mana_cost(amount));
+            }
+        }
+    }
+}
+
+/// CR 118.7: A `{amount}` generic-mana `AbilityCost`, used to add a mana component
+/// to an ability whose printed cost has none when a cost-raise static applies.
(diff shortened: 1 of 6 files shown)


matthewevans (changes requested):
**[HIGH]** Taxed loyalty abilities can bypass the direct-action once-per-turn guard after the tax wraps their cost. Evidence: `crates/engine/src/game/casting.rs:14487` applies cost reduction before cloning `activation_cost`, `crates/engine/src/game/casting.rs:14627` then checks that mutated cost with `is_loyalty_ability_cost`, and `crates/engine/src/types/ability.rs:7200` only recognizes bare `Loyalty` / loyalty-counter removal, not `Composite { Loyalty, Mana }`; this PR creates exactly that composite in `increase_generic_in_cost` at `crates/engine/src/game/casting.rs:15604`. Why it matters: a direct `GameAction::ActivateAbility` for an Eidolon-taxed planeswalker ability can skip the CR 606.3 per-permanent loyalty activation guard after the first activation, even though candidate generation still blocks through the pre-tax path. Suggested fix: preserve/use the pre-reduction loyalty classification for the direct guard or extend the loyalty-cost predicate to recognize the post-tax composite shape, and add a direct activation regression that fails on a second taxed loyalty activation in the same turn.

Also note that this engine/parser PR still needs the coverage parse-diff comment for the current head before it can be considered merge-ready.

matthewevans (changes requested):
**[HIGH]** Eidolon's tax does not run on the normal fixed-loyalty activation path. Evidence: `crates/engine/src/game/engine.rs:1498` routes raw `AbilityCost::Loyalty` activations to `planeswalker::handle_activate_loyalty`, while `planeswalker::handle_activate_loyalty` builds/finalizes from the printed ability without calling `apply_cost_reduction` (`crates/engine/src/game/planeswalker.rs:169`, `crates/engine/src/game/planeswalker.rs:193`); the PR's tax logic only runs inside `apply_cost_reduction` (`crates/engine/src/game/casting.rs:14488`, `crates/engine/src/game/casting.rs:15626`), and the new regression only calls that helper directly (`crates/engine/src/game/casting_tests.rs:38690`) instead of dispatching `GameAction::ActivateAbility`. Why it matters: Eidolon now parses as supported, but an opponent can still activate a fixed loyalty ability without paying the extra {1} in the real game flow. Suggested fix: route fixed loyalty activations through the same cost-adjustment/payment pipeline (while preserving CR 606 counters/activation tracking), or apply the static adjustment/payment in the planeswalker handler, and add a production activation test that fails without available generic mana under Eidolon.

galuis116:
Thanks for the review — addressed in the pushed commit:

**[HIGH] Taxed loyalty abilities bypassing the CR 606.3 once-per-turn guard.** Root cause fixed at the predicate: `is_loyalty_ability_cost` now recurses into `Composite`, so a loyalty ability whose cost was wrapped as `Composite { Loyalty, Mana }` by the tax stays classified as a loyalty ability everywhere the predicate is used — including the direct `handle_activate_ability` guard (`casting.rs:14627`). Added `direct_activation_of_taxed_loyalty_ability_second_time_denied`, which drives `handle_activate_ability` for an Eidolon-taxed planeswalker after `loyalty_activations_this_turn = 1` and asserts the second activation is rejected; it fails if the `Composite` arm of `is_loyalty_ability_cost` is reverted.

**[MEDIUM] `std::mem::replace` + `if let`** → direct `*cost = Composite { … }` assignment.

**[MEDIUM] Redundant `Option` check on `ability_def.cost`** → `ability_is_loyalty` is now computed on the already-unwrapped `cost`.

Verified locally: `cargo fmt`, full `engine` lib suite (15,657 passed / 0 failed), and `clippy -p engine --all-targets -D warnings` clean. The coverage parse-diff comment regenerates for the new head via the Card-data job.

matthewevans (changes requested):
**[HIGH]** The fixed-loyalty activation path still bypasses the new Eidolon tax. Evidence: `crates/engine/src/game/engine.rs:1498` routes bare `AbilityCost::Loyalty` activations to `planeswalker::handle_activate_loyalty`, `crates/engine/src/game/planeswalker.rs:193` reads the printed ability directly, and `crates/engine/src/game/planeswalker.rs:337` pays only a freshly built `AbilityCost::Loyalty` without calling `apply_cost_reduction`; the new Eidolon runtime test at `crates/engine/src/game/casting_tests.rs:38759` calls `apply_cost_reduction` directly, so it never proves the real `GameAction::ActivateAbility` path pays the added `{1}`. Why it matters: Eidolon now parses as supported, but an opponent can still activate a normal fixed loyalty ability without the additional mana in actual play. Suggested fix: route fixed loyalty activation through the same cost-adjustment/payment authority, or apply the static tax in `handle_activate_loyalty`, and add a production activation regression that fails under Eidolon when the activating player lacks the extra generic mana.

galuis116:
Thanks — you're right, the fixed-loyalty fast path bypassed the tax entirely, and the earlier test only exercised `apply_cost_reduction`. Fixed the real activation path in the pushed commit:

**Routing.** `handle_activate_loyalty` now detects (via `loyalty_ability_gains_mana_tax`) when a cost-raise static adds a mana component to a loyalty ability and defers to the general `handle_activate_ability` flow — which applies the tax, pays the loyalty counters, records the CR 606.3 activation, and enforces the once-per-turn gate. The CR 606.6 negative-loyalty check runs first, before either branch, so it's preserved for the delegated case. Untaxed loyalty abilities keep the unchanged fast path.

**Payment atomicity (CR 601.2g).** Routing surfaced a real ordering bug: a `Composite { Loyalty, Mana }` fell through to `pay_ability_cost`, which paid the loyalty counters and *then* failed on the unaffordable mana — a free loyalty change. The mana-leg-first detour in `handle_activate_ability` now also fires for a loyalty cost, so the mana leg is hoisted through `enter_payment_step` (paid FIRST) and the loyalty counter cost is deferred as the `ManaLeg` residual. No partial payment.

**Production regression** (`eidolon_forces_mana_payment_for_real_loyalty_activation`) drives the real `handle_activate_loyalty` entry point: an untaxed activation needs no mana; a taxed activation by a manaless player stops at the mana-payment step (or is refused) and — the load-bearing assertion — **the loyalty counter does not change**. Reverting either the delegation or the mana-leg routing fails it.

Verified locally against the current head: full engine suite (15,749 passed / 0 failed) and `clippy --all-targets -D warnings` clean.

matthewevans (changes requested):
**[HIGH]** Taxed targeted loyalty abilities pay costs before target selection. Evidence: `crates/engine/src/game/planeswalker.rs:220` delegates every Eidolon-taxed fixed loyalty ability to `handle_activate_ability`; that path hoists the added mana leg at `crates/engine/src/game/casting.rs:14707` before any target slots are built, and the resume path pays the residual loyalty cost/records the loyalty activation at `crates/engine/src/game/casting_costs.rs:2948` and `crates/engine/src/game/casting_costs.rs:2964` before it checks for target selection at `crates/engine/src/game/casting_costs.rs:2969`. Why it matters: CR 602.2b/601.2c require targets to be chosen before costs are paid, and the existing untaxed loyalty path preserves that ordering; under Eidolon, a targeted loyalty ability can now spend mana/loyalty before target selection or cancellation. Suggested fix: keep the taxed loyalty path's target-selection ordering aligned with `handle_activate_loyalty` by selecting/assigning targets before the mana-leg payment resume, and add a taxed targeted-loyalty regression that asserts loyalty is unchanged while waiting for targets and only changes after targets plus the added {1} are paid.

galuis116:
Good catch — the delegation paid costs before target selection for targeted loyalty abilities. Fixed in the pushed commit:

**Target-first ordering (CR 601.2c).** The mana-leg-first hoist now fires only for a *non-targeted* taxed loyalty ability. A *targeted* one falls through to the general target-first path, so it chooses targets before paying costs — matching the untaxed loyalty path.

**Atomicity on the targeted path.** `increase_generic_in_cost` now places the added mana component *first* in the wrapping `Composite` (CR 601.2h — costs may be paid in any order). Any payment path that pays a `Composite` in order now settles the mana leg before the loyalty counters, so an unaffordable/cancelled mana payment never leaves a free loyalty change — including the post-target payment.

**Regression** (`eidolon_taxed_targeted_loyalty_selects_targets_before_paying`) drives a taxed targeted `[−1]` loyalty ability through the real `handle_activate_loyalty` entry point and asserts it stops at target selection with the loyalty counter unchanged. Reverting the non-targeted gate (or the mana-first ordering) fails it.

Verified on the current head: full engine suite (15,750 passed / 0 failed), `clippy --all-targets -D warnings` clean, rebased onto latest upstream.

matthewevans (changes requested):
**[HIGH]** Taxed targeted loyalty abilities still pay the added mana before target selection. Evidence: `crates/engine/src/game/planeswalker.rs:220` delegates every Eidolon-taxed fixed loyalty ability to `handle_activate_ability`; `crates/engine/src/game/casting.rs:14710` then extracts the added mana leg and returns `enter_payment_step` at `crates/engine/src/game/casting.rs:14716`, before the target-selection branch is reached at `crates/engine/src/game/casting.rs:15097`. Why it matters: CR 602.2b/601.2c requires targets to be chosen before costs are paid, so any taxed targeted planeswalker ability still runs in the wrong activation order. Suggested fix: keep taxed loyalty abilities on a path that builds/selects targets first, then pays the added mana plus loyalty residual, and add a targeted taxed loyalty regression test that fails if `ManaPayment` appears before `TargetSelection`.

matthewevans (approved):
Current head addresses the target-order blocker: taxed targeted loyalty abilities now fall through to target selection before payment, with regression coverage.

PR #5342, opened 2026-07-07
docs: verify card premise before planning + prefer installed toolchain

rykerwilliams:
## Summary

Two process gaps found while fixing an unrelated card bug, captured as durable guidance:

1. **Fabricated card abilities can survive multiple review rounds.** Plan/implementation review verifies a design is *executed correctly* against its stated premise — it does not fact-check that the premise (what the card actually does) is real. Adds a Step 0 hard gate to `engine-planner` and a matching `CLAUDE.md` principle: verify a card's actual Oracle text **verbatim** against Scryfall/MTGJSON before planning or implementing, never from memory or a task description's paraphrase. (This is distinct from CR-annotation verification, which checks that a cited *rule* is real, not that the *card ability* is real.)
2. **Silently defaulting to "no Rust toolchain, skip local verification"** leads to PRs that fail CI on things a local build/test/clippy pass would have caught. Updates `AI-CONTRIBUTOR.md` to prefer installing the toolchain and using the Developer track whenever the runtime has shell/package-manager access, reserving the Non-developer track for when that's genuinely impossible.

Note: this touches `.claude/skills/engine-planner/SKILL.md`, which per `.agents/pr-review-policy.toml` is a hard-stop path excluded from the automated review/merge loop (same as #5304) — this will need a maintainer to merge it directly rather than auto-clearing bot review.

## Test plan

- [x] Docs-only change, no code/behavior affected.
- [x] Verified both gaps are described generically (no fork-specific or single-card-specific content) — this is guidance meant to apply to every future contributor/session, not just the case that surfaced it.


matthewevans (changes requested):
This PR touches instruction/policy files that are hard-stop paths for external contributor PRs under `.agents/pr-review-policy.toml`: `.claude/skills/engine-planner/SKILL.md`, `CLAUDE.md`, and `docs/AI-CONTRIBUTOR.md`.

Per the maintainer PR review policy, these files steer future agent/reviewer behavior and require explicit maintainer handling outside the normal contributor queue. Please do not route this through the regular merge queue.

PR #5347, opened 2026-07-07
fix(parser): target-player possessive zone references in ChangeZone (#1077)

### LINKED ISSUE #1077: [Card Bug] Relic of Progenitus does not work correctly
### Card(s)

Relic of Progenitus

### Build/version

V0.1.36

### Game mode

AI

### Actual behavior

When activating the first ability, the UI asks to select target player, and then shows the owner graveyard, even when selecting opponent, and we have to select a card, even though it does not target. Also, only the first ability is activating automaticaly and the second ability does not work

### Expected behavior

When activating relic, it should prompt to select the first or second ability. If the first is select, it should wait for selecting target player, and after selecting, the target player should be prompted to select a cart from its own graveyard to exile. When selecting the second ability, it should be exiled and then exile both players graveyards and drawing a card.

### Steps to reproduce

Just click the card and select target player.

_No response_

rykerwilliams:
## Summary
- Relic of Progenitus's "Target player exiles a card from their graveyard" showed the *activator's* own graveyard instead of the targeted player's — the possessive "their" parses as `FilterProp::Owned { controller: ScopedPlayer }` (scope-agnostic), and `inject_subject_target`'s `ChangeZone` arm only rewrote the top-level `tf.controller` via `force_controller`, leaving that stale property untouched. The filter ended up with two contradictory ownership constraints (`controller: TargetPlayer` vs. a nested `Owned: ScopedPlayer` that falls back to the activator), so it was only ever satisfiable when the activator targeted themself.
- Generalized `rebind_owned_scope` (previously hardcoded to `ControllerRef::ChosenPlayer`, used only by the `Bounce`/Skullwinder call sites) to accept any `ControllerRef`, and call it in the `ChangeZone` arm right after `force_controller`. It's a no-op when the effective controller is itself `ScopedPlayer`, so the existing Braids/"each player" path is unaffected.
- Same class also covers Scrabbling Claws, Merrow Bonegnawer, Graveyard Shovel, Grave Birthing, and Gravestorm ("target player [verb]s a card from their [zone]") — not separately tested in this PR, flagged as a natural follow-up rather than blocking scope creep.

Closes #1077.

## Test plan
- [x] Added `relic_of_progenitus_target_player_exiles_from_their_graveyard` (mirrors the existing `suffer_the_past_exiles_from_target_player_graveyard` precedent), asserting the parsed filter is constrained to `Owned: ControllerRef::TargetPlayer` and does **not** retain the stale `ScopedPlayer` binding.
- [x] `cargo fmt --all` clean.
- [x] An earlier full-crate test run surfaced a second, previously-missed `rebind_owned_scope` call site (`retarget_effect_to_chosen_player`'s `Bounce` arm) with a real compile error from the signature generalization — fixed and confirmed via `cargo check` reaching the `engine` crate cleanly.
- [ ] I was not able to get a full `cargo test -p engine --lib` run to completion in my local sandbox (repeated background-process termination mid-compile, unrelated to this change) — CI should be treated as the authoritative test run for this PR.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -16680,6 +16680,26 @@ fn lower_subject_predicate_ast(
                     clause.effect,
                     Effect::ChangeZone { .. } | Effect::ChangeZoneAll { .. }
                 ) {
+                    // CR 109.4 (issue #1077): "target player exiles a card
+                    // from their graveyard" (Relic of Progenitus, Scrabbling
+                    // Claws, Merrow Bonegnawer, Graveyard Shovel, Grave
+                    // Birthing, Gravestorm) and "target player[s] ... their
+                    // [zone]" (Memory's Journey, above). The moved-object
+                    // filter's possessive "their" parses as
+                    // `Owned { controller: ScopedPlayer }` because the
+                    // zone-suffix parser is scope-agnostic — but this branch
+                    // only runs for an explicit "target player" (not an
+                    // anaphoric "that player"), so the filter must be rebound
+                    // to the real declared target before it's cloned into the
+                    // sub-ability, or it stays scoped to the acting player
+                    // (the activator) instead of the player just targeted.
+                    match &mut clause.effect {
+                        Effect::ChangeZone { target, .. }
+                        | Effect::ChangeZoneAll { target, .. } => {
+                            rebind_owned_scope(target, ControllerRef::TargetPlayer);
+                        }
+                        _ => {}
+                    }
                     let mut sub_ability =
                         AbilityDefinition::new(AbilityKind::Spell, clause.effect.clone());
                     sub_ability.sub_ability = clause.sub_ability;
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -6668,7 +6668,9 @@ fn retarget_effect_to_chosen_player(effect: &mut Effect, index: u8) {
         // emits one) from `ScopedPlayer` to `ChosenPlayer { index }` so the
         // chosen opponent's graveyard is enumerated at resolution time
         // (CR 608.2d — the chosen opponent announces which card returns).
-        Effect::Bounce { target, .. } => rebind_owned_scope(target, index),
+        Effect::Bounce { target, .. } => {
+            rebind_owned_scope(target, ControllerRef::ChosenPlayer { index })
+        }
         _ => {}
     }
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**[HIGH]** The PR is red on its own new regression test. Evidence: `crates/engine/src/parser/oracle_effect/tests.rs:6662`; the current-head `Rust tests (shard 2/2)` CI log for `84728c9ef1eeff678f221467b3bb53223d15382b` fails `relic_of_progenitus_target_player_exiles_from_their_graveyard` with `expected ChangeZone exile, got {"type":"TargetOnly","target":{"type":"Player"}}`. Why it matters: this blocks merge, and the test is not asserting through the production subject-wrapper shape that this parser path actually emits. Suggested fix: make the regression follow the `TargetOnly` wrapper and inspect the nested `ChangeZone` sub-ability, or adjust the lowering if `TargetOnly` is not intended for this path, then wait for the Rust shard to pass.

rykerwilliams:
Thanks for catching this — you were right, the original fix targeted the wrong code path.

Traced it fully: for "target player" + \`ChangeZone\`/\`ChangeZoneAll\`, \`lower_subject_predicate_ast\` has an **earlier** return (\`mod.rs:16674\`) that wraps the effect into \`TargetOnly{target: Player}\` + \`sub_ability\` *before* \`inject_subject_target\` (where my original fix lived) is ever reached. So the fix was real but dead code for Relic of Progenitus's actual parse.

Pushed a correction:
- Applied the same \`rebind_owned_scope\` call at the actual early-return site (right before the effect gets cloned into the sub-ability).
- Rewrote the test to assert through the real \`TargetOnly\`+\`sub_ability\` shape (mirrors the existing \`target_subject_damage_equal_to_its_power_uses_target_source_power\` precedent), instead of an assumed top-level \`ChangeZone\`.
- Left the original \`inject_subject_target\` arm in place too — harmless, may cover other code paths I haven't fully mapped, and the \`rebind_owned_scope\` signature generalization is needed by both call sites regardless.

Live-verified this time: the new test passes after a full local compile, and all 48 Bounce-related tests pass (no regression from generalizing \`rebind_owned_scope\`'s signature).

matthewevans (approved):
Reviewed current head 7e6e452a75aaf847e955f3752c4ba92fb0f4cba5. The requested-change item is addressed: the parser test now follows the real TargetOnly sub-ability shape, and the moved-object filter rebinding happens at the lower_subject_predicate_ast early return that handles target-player ChangeZone effects. Parse-diff scope is limited to the expected ownership rebinding class.

PR #2999, opened 2026-06-11
fix: ParentTarget GenericEffect binding for targeted pump/debuff abilities

claytonlin1110:
## Summary

- Fixes Mu Yanling, Sky Dancer's +2 (`-2/-0 and loses flying`) applying to every creature instead of only the chosen target.
- Root cause: `register_transient_effect` resolved `target_filter.or(affected)` with the outer `Typed(Creature)` targeting slot winning over `affected: ParentTarget`, routing registration through the battlefield broadcast path.
- Adds `generic_effect_application_filter` so inherited `affected` references (`ParentTarget`, `TriggeringSource`, `CostPaidObject`) govern where modifications land at resolution time.

## Test plan

- [x] `cargo test -p engine --lib game::effects::effect::tests` (28 tests)
- [x] `parent_target_affected_with_creature_target_slot_binds_chosen_creature_only`
- [x] `parent_target_affected_with_creature_target_slot_empty_targets_does_not_broadcast`
- [x] `mu_yanling_plus_two_pump_and_lose_flying_parses_and_resolves_to_target_only`
- [ ] Manual: activate Mu Yanling +2, target one flying creature, confirm only that creature gets -2/-0 and loses flying

Fixes #2922

Issue #2922, opened 2026-06-10
Mu Yanling, Sky Dancer: +2 '-2/-0 and loses flying' applies to all creatures instead of only the target


matthewevans:
## Oracle text (verified from client/public/card-data.json)
```
[+2]: Until your next turn, up to one target creature gets -2/-0 and loses flying.
[−3]: Create a 4/4 blue Elemental Bird creature token with flying.
[−8]: You get an emblem with "Islands you control have '{T}: Draw a card.'"
```

## Reported behavior
> <Conor> gives -2/-0 to all creatures, not just the target.

## Expected vs actual
- Expected: Only the single chosen target creature gets -2/-0 and loses flying until your next turn.
- Actual: All creatures get -2/-0.

AST evidence: the +2 ability targets a single `Typed: Creature` (up-to-one), and the continuous effect is a `GenericEffect` static whose `affected` is `ParentTarget` with `[AddPower -2, AddToughness 0, RemoveKeyword Flying]`, `duration: UntilNextTurnOf(Controller)`. The parse is correct (single ParentTarget), so the board-wide application is a runtime bug in how the ParentTarget-scoped GenericEffect static is applied — it is hitting all creatures rather than only the resolved target.

## Source
https://discord.com/channels/1485498006781427802/1513990676210126969/1513990676210126969

Issue #1272, opened 2026-05-28
[Card Bug] Violent Urge giving Double Strike to all creatures


zirzhas:
### Card(s)

Violent Urge

### Build/version

0.1.39

### Game mode

AI

### Actual behavior

Violent Urge is giving double strike to all creatures when delirium is active

### Expected behavior

It should give +1+0 and double strike only to target creature

rykerwilliams:
Investigated this — already fixed on current `main`, now verified locally with a permanent regression test rather than a scratch check:

**Parser:** parsed the verbatim Oracle text and confirmed the Delirium clause's `GenericEffect` carries a single `StaticDefinition` with `affected: Some(TargetFilter::ParentTarget)` and `modifications: [AddKeyword(DoubleStrike)]` — "that creature" correctly anaphors back to the single cast-time target, not an unscoped filter. Added this as a permanent test in phase-rs/phase#5348 so it can't silently regress.

**Root cause of why it's already fixed:** this is the same "target X gets A. [condition] — that X also gets B" template as Mu Yanling, Sky Dignitary's +2 (issue #2922), which had this exact "broadcasts to every matching permanent instead of the parent target" bug. It was fixed generically (not card-specifically) by #2999 ("fix: ParentTarget GenericEffect binding for targeted pump/debuff abilities"), which landed after this issue was filed — Violent Urge just never got its own follow-up regression test or issue closure, until now.

Closing as already-fixed. Happy to reopen with a fresh repro (build/commit hash) if this still occurs somewhere.

mike-theDude:
## Classifier analysis

**Verdict**: `not_card_data_attributable`

**Matched clause**: `that creature gains double strike until end of turn` (currently supported: true)

**Reasoning**: Current coverage keeps the delirium DoubleStrike grant on `parent target`, not all creatures. Because this report is from 0.1.39 and current card data looks correct, it needs runtime verification before treating it as still reproducible.

---
_Posted by bug-coverage-classifier skill._

mike-theDude:
This was a real 0.1.39 bug, but the shared runtime cause was fixed after the report by PR #2999 (24afeefbb), and PR #5348 added an exact Violent Urge regression. Both grants are now scoped to ParentTarget, and the runtime binds the temporary effect to the chosen SpecificObject instead of broadcasting it to every creature; regression coverage also proves a bystander remains unchanged and an empty target never broadcasts. This report is now stale. Could a maintainer close or delete it?

[...]

PR #5348, opened 2026-07-08
test(parser): pin Violent Urge Delirium clause to ParentTarget (#1272)

rykerwilliams:
## Summary

- Investigated #1272 ("Violent Urge giving double strike to all creatures when delirium is active"): already fixed on `main`, no behavior change in this PR.
- Root cause of why it's already fixed: same "target X gets A. [condition] — that X also gets B" template as Mu Yanling, Sky Dignitary's +2 (#2922), generically fixed by #2999 ("ParentTarget GenericEffect binding for targeted pump/debuff abilities"). Neither card got a follow-up regression test after that fix landed.
- This PR adds that missing coverage: parses Violent Urge's verbatim Oracle text and asserts the Delirium clause's static ability is scoped to `TargetFilter::ParentTarget` (the same single target as the base pump effect), not an unscoped/all-creatures filter.

## Test plan

- [x] New test `violent_urge_delirium_scopes_to_parent_target_not_all_creatures` passes locally.
- [x] Test-only change, no production code touched.

### DIFF
--- a/crates/engine/src/parser/oracle_tests.rs
+++ b/crates/engine/src/parser/oracle_tests.rs
@@ -19243,3 +19243,56 @@ fn enters_with_n_additional_counters_parses_canonical_type() {
     assert_eq!(ct, CounterType::Time, "Ravaging Riftwurm type");
     assert_eq!(count, QuantityExpr::Fixed { value: 3 }, "count");
 }
+
+/// Regression for issue #1272: Violent Urge's Delirium follow-up ("that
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed test-only parser coverage. Oracle text matches Scryfall, the assertion reaches the parser-produced GenericEffect/static grant shape, and the existing neighboring Delirium condition test covers the same card condition. Local diff is limited to oracle_tests.rs; no blocking review comments found.

PR #5349, opened 2026-07-08
fix(engine): cast/played permanents you don't own enter under your control (#696)

### LINKED ISSUE #696: Evelyn, the Covetous cast cards from exile isn't working with enemy cards — Card: Evelyn, the Covetous Problem: After E…
Reported in Discord: https://discord.com/channels/1485498006781427802/1506886234989924434/1506886234989924434

**Thread:** Evelyn, the Covetous cast cards from exile isn't working with enemy cards
**Cards:** evelyn, the covetous
**Parser status:** has_gaps

## Summary
Card: Evelyn, the Covetous

Problem: After Evenlyn or another vampire enters the battlefield, the top card of each player’s library is exiled, but on

### evelyn, the covetous
Flash
Whenever Evelyn or another Vampire you control enters, exile the top card of each player's library with a collection counter on it.
Once each turn, you may play a card from exile with a collection counter on it if it was exiled by an ability you controlled, and you may spend mana as though it were mana of any color to cast it.

---
<sub>report_id: `discord:1506886234989924434:1506886234989924434:0` · discord: `1506886234989924434/1506886234989924434`</sub>

rykerwilliams:
## Summary

Closes #696 (Evelyn, the Covetous cast cards from exile isn't working with enemy cards). Also independently surfaced via Ragavan, Nimble Pilferer — same root cause, different card.

**The bug**: when a permanent enters the battlefield via a cast spell (`stack.rs`) or a played land (`engine.rs::handle_play_land`), the constructed `ProposedEvent::ZoneChange` left `controller_override` at its `None` default, so `reset_for_battlefield_entry` fell back to `self.controller = self.owner`. Per **CR 601.2a** ("...that player becomes its controller") and **CR 110.2a** ("that object enters the battlefield under that player's control unless the effect states otherwise"), the correct default is whoever cast/played it — not the card's owner.

This is invisible whenever `owner == caster/player` (the overwhelming majority of casts/plays) and only manifests when a granted permission (`CastingPermission::PlayFromExile`, which both Ragavan and Evelyn grant) lets a player cast or play a card they don't own.

**Two sibling seams, one fix pattern**: Evelyn's own ability says "**play**" (not "cast"), which is why this needed fixing in both `stack.rs` (spell casts) and `engine.rs::handle_play_land` (land plays) — her collection counter can land on a land card in an opponent's library. Both fixes feed an existing, already-correct, already CR-annotated consumer (`zones::apply_battlefield_entry_controller_override`, via `zone_pipeline.rs`'s `ctrl_override` check) that was previously only ever wired to the "enters_under" self-ETB replacement mechanism (e.g. a static that says "~ enters the battlefield under an opponent's control"). That mechanism still wins in both seams — it runs later in the pipeline (`replace_event`) and hard-overwrites the new default unconditionally, so this doesn't regress anything that already explicitly specifies control.

This class of bug also affects any future "cast/play a card you don't own" design (Bribery, Mind's Desire, Praetor's Grasp, Etali-class effects) — the fix lives in the generic resolution pipeline, not in any card's own ability definition.

## Process note

This plan went through two full adversarial `/review-engine-plan` rounds before implementation, given the blast radius (every permanent cast/land play in the game passes through these two seams):
- **Round 1** found a missing sibling-seam gap (the original plan only fixed the spell-cast path, missing that `handle_play_land` has the identical bug) and a fabricated CR citation (608.2f, which is actually about APNAP ordering, not controller stability — corrected to 601.2a).
- **Round 2** independently re-verified the revised plan's interaction with Mindslaver-style turn-control effects (confirmed the fix correctly uses the turn's rightful player, not whoever is making decisions) and confirmed `GameRunner::act()` can drive the land-play test (no dedicated `.play_land()` builder exists, but the generic action-dispatch escape hatch does).

## Test plan
- [x] Added `cast_permanent_from_granted_permission_enters_under_caster_control`: casts an opponent-owned permanent from exile via a granted `PlayFromExile` permission, asserts the resulting permanent's controller is the caster while owner is unchanged.
- [x] Added `play_land_from_granted_permission_enters_under_player_control`: plays an opponent-owned land from exile via the same permission type, same assertions, through the independent `handle_play_land` code path.
- [ ] **I was not able to get a local `cargo check`/`cargo test` run to complete in this sandbox** — four consecutive attempts across two different cargo subcommands failed. The first three were killed mid-compile at the identical point (`Checking engine v0.18.0`, immediately after all dependencies compiled successfully). The fourth failed with **unrelated** third-party dependency crates (`serde_core`, `zerocopy`, `windows-sys`, `bitmaps`) failing to compile simultaneously with generic "process didn't exit successfully" errors — none of which this change touches, pointing to sandbox resource exhaustion rather than a defect in this diff. **CI should be treated as the first real compile/test confirmation for this PR** — please pay particular attention to the two new tests actually passing, given I could not confirm that locally.

### DIFF
--- a/crates/engine/src/game/casting_tests.rs
+++ b/crates/engine/src/game/casting_tests.rs
@@ -451,6 +451,126 @@ fn spell_auto_tap_honors_exile_any_color_permission() {
     assert_eq!(state.players[0].mana_pool.mana.len(), 0);
 }
 
+#[test]
+fn cast_permanent_from_granted_permission_enters_under_caster_control() {
+    // GitHub phase-rs/phase#696 (Evelyn, the Covetous) — surfaced independently
+    // via Ragavan, Nimble Pilferer. CR 601.2a + CR 110.2/110.2a: casting a
+    // permanent you don't own via a granted CastingPermission::PlayFromExile
+    // must put it under the CASTER's control, not the card's owner.
+    let mut state = setup_game_at_main_phase();
+    // Owned by P1, sitting in exile, with a permission granting P0 the right
+    // to cast it — mirrors spell_auto_tap_honors_exile_any_color_permission's
+    // construction pattern above, but with owner != granted_to.
+    let permanent = create_object(
+        &mut state,
+        CardId(51),
+        PlayerId(1),
+        "Borrowed Creature".to_string(),
+        Zone::Exile,
+    );
+    {
+        let obj = state.objects.get_mut(&permanent).unwrap();
+        obj.card_types.core_types.push(CoreType::Creature);
+        obj.power = Some(1);
+        obj.toughness = Some(1);
+        obj.mana_cost = ManaCost::Cost {
+            shards: vec![],
+            generic: 0,
+        };
+        obj.casting_permissions
+            .push(CastingPermission::PlayFromExile {
+                duration: Duration::Permanent,
+                granted_to: PlayerId(0),
+                frequency: CastFrequency::Unlimited,
+                source_id: None,
+                exiled_by_ability_controller: None,
+                mana_spend_permission: None,
+                card_filter: None,
+                single_use_group: None,
+                single_use: false,
+                cast_cost_raise: None,
+                land_enter_tapped: crate::types::zones::EtbTapState::Unspecified,
+            });
+    }
+
+    let mut runner = crate::game::scenario::GameRunner::from_state(state);
+    let outcome = runner.cast(permanent).resolve();
+
+    // Reach-guard: the permanent must actually resolve onto the battlefield
+    // before the controller assertion below means anything.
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Reviewed engine control-default fix after maintainer CR-comment fixup. The change lands at the two production seams that create battlefield ZoneChange events for cast permanents and played lands, seeds controller_override before replacement processing, and existing replacement code still overwrites it for explicit enters-under effects. Scryfall confirms Evelyn grants play and Ragavan grants cast from exile; CR 110.2, CR 110.2a, CR 305.1, and CR 601.2a were verified locally. Tests cover both stack and land-play paths with owner != actor reach guards. CI is pending on the fixup commit; auto-merge will wait for required checks.

Issue #1033, opened 2026-05-24
[Card Bug] Underworld Breach is broken in multiple ways


soul2squeeze:
### Card(s)

Underworld Breach

### Build/version

v0.1.35 6263238

### Game mode

P2P

### Actual behavior

Underworld Breach allows me to cast cards from my graveyard, but does not allow me to exile any cards to pay the additional cost. It just says "Exile (0/0" and when clicked, casts the selected card without exiling any other cards from the graveyard. I was able to cast a single Dark Ritual infinite times with no other cards in my graveyard.
It also allows me to cast Underworld Breach from my graveyard when I have nothing in play that should allow me to recur it. Each turn, I was able to recast Underworld Breach from my graveyard, and then start to loop my Dark Ritual again.

### Expected behavior

Underworld Breach should force the player to exile three other cards from the graveyard when casting a card for the escape cost. It should not be castable from the graveyard itself.

### Steps to reproduce

Cast Underworld Breach with at least one spell and three other cards in the graveyard and attempt to exile them when casting the spell.

rykerwilliams:
Investigated this — both parts appear to already be fixed on current `main`, now verified locally with permanent regression tests rather than source-only reasoning:

1. **"Exile (0/0" / cast goes through without exiling anything.** I ran `cargo test -p engine --lib escape` myself and confirmed all 31 tests pass, including `granted_escape_requires_exile_cost_payment` and `ai_escape_cast_from_graveyard_pays_mana_and_exiles_five_cards`, which cover exactly this scenario. `effective_escape_data` (`game/keywords.rs`) refuses any escape grant with no residual cost, and both `can_pay_escape_additional_cost` (`game/casting.rs`) and the `AbilityCost::Exile` arm of `pay_additional_cost` (`game/casting_costs.rs`) gate on the real fixed exile count (3), never a count clamped down to what's available. This class of fix landed in `aa4ee3455`, confirmed as a real ancestor of current `main`.

[...]

PR #5350, opened 2026-07-08
test(engine): pin Underworld Breach's grant to its own battlefield presence (#1033)

rykerwilliams:
## Summary

- Investigated the second half of #1033 ("Underworld Breach recasts itself from the graveyard via its own granted escape"): already correct on `main`, no behavior change in this PR.
- Per CR 604.2, a static ability's continuous effect (including a keyword grant like "each nonland card in your graveyard has escape") only exists while its source remains on the battlefield. `for_each_static_effect_source` sources these grants exclusively from `battlefield_sources` (`game/layers.rs`), so once Breach is sacrificed and sits in the graveyard, its own grant is inert — including toward itself.
- This PR adds the missing coverage: places Underworld Breach directly in the graveyard (simulating post-sacrifice) and confirms neither it nor another graveyard card shows escape availability derived from its grant.
- Complements the existing `granted_escape_requires_exile_cost_payment` test, which covers the first (and only actually-reproducing) half of the original report — the exile-cost enforcement bug, already fixed by `aa4ee3455`.

## Test plan

- [x] New test `escape_grant_from_graveyard_source_does_not_apply_to_itself` passes locally.
- [x] Test-only change, no production code touched.

### DIFF
--- a/crates/engine/src/game/casting_tests.rs
+++ b/crates/engine/src/game/casting_tests.rs
@@ -26318,22 +26318,33 @@ fn escape_grant_from_graveyard_source_does_not_apply_to_itself() {
     source.static_definitions = parsed.statics.clone().into();
     source.base_static_definitions = Arc::new(parsed.statics);
     source.mana_cost = ManaCost::Cost {
-        generic: 1,
-        shards: vec![ManaCostShard::Red],
+        generic: 0,
+        shards: vec![],
     };
 
-    // A second nonland card also sitting in the graveyard, which WOULD be
-    // escape-eligible if Breach's grant were (incorrectly) still active.
-    let filler_id = create_object(
-        &mut state,
-        CardId(1200),
-        PlayerId(0),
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Reviewed test-only Underworld Breach coverage after maintainer test hardening. Scryfall confirms the Oracle text and CR 604.2 was verified locally. The test now makes the granted escape cost payable if Breach incorrectly grants escape from the graveyard, so the negative assertions are load-bearing for both Breach itself and another graveyard card. CI is pending on the fixup commit; auto-merge will wait for required checks.

matthewevans (approved):
Re-approved current head after maintainer fixups. The CI failure was from the off-zone static gate suppressing command-zone emblem statics; the current head preserves CR 114.4 emblems while keeping ordinary off-zone grants gated. Auto-merge can wait for the remaining checks.

PR #5356, opened 2026-07-08
fix(parser): "destroy/exile all X except for <type-list>" exclusion clause

### LINKED ISSUE #4710: [Card Bug] Scourglass
### Card(s)

Scourglass

### Build/version

v0.12.0 e0eff39

### Game mode

None

### Actual behavior

Destroys all permanents

### Expected behavior

Any permanent except for Artifacts or Lands

### Steps to reproduce

_No response_

Image

rykerwilliams:
## Summary

Fixes phase-rs/phase#4710 (Scourglass: "Destroy all permanents except for artifacts and lands" actually destroys all permanents, no exception applied).

Verified Oracle text via Scryfall: "{T}, Sacrifice this artifact: Destroy all permanents except for artifacts and lands. Activate only during your upkeep." ({3}{W}{W}, Artifact)

**Root cause**: `parse_type_phrase_with_ctx` had no suffix parser recognizing `"except for <type-list>"` — only the predicate-based `"except those that <relative-clause>"` form was handled. The trailing exception clause was silently left unconsumed with zero parse warning, so the effect resolved against the unfiltered `Permanent` population.

**General class, not a one-card patch**: a live Scryfall search (`o:"except for" o:"destroy all"`) surfaces **Elspeth Tirel**'s −5 ability ("Destroy all other permanents except for lands and tokens") sharing the exact same grammar, with a heterogeneous split — "lands" is a `TypeFilter`, "tokens" is a `FilterProp` (not a card type). Both are fixed by the same change.

**Building-block reuse, no new types**:
- `classify_negation` (already used by the existing `"nonartifact, nonland"` prefix-negation loop) does all the word→`TypeFilter::Non`/`FilterProp` classification — reused as-is.
- The existing Oxford-comma-tolerant `match_mass_union_separator` handles the list separator — reused as-is.
- `game/filter.rs`'s `TypeFilter::Non` evaluation already correctly treats artifact creatures as satisfying "artifact" (CR 205.2b: an object with more than one card type satisfies criteria for any of them) — **no runtime code changed**, this is parser-only.

**Hostile-fixture guard**: `classify_negation`'s catch-all treats any unrecognized word as a negated Subtype (correct for its original "non-`<word>`" context, e.g. "nonZombie"). That's unsafe for this free-form exception list — `"except for Mageta"` (Mageta the Lion) or `"except for commanders"` (Slash the Ranks) would otherwise silently produce a no-op `Non(Subtype("Mageta"))` exclusion that looks fixed but isn't. The new suffix parser explicitly declines the whole clause in that case, leaving those (out-of-scope, differently-shaped) cards' existing honest-unhandled behavior alone rather than mis-firing.

## Test plan

- [x] Live scratch-parsed Scourglass's exact text *before* the fix to confirm the exact silent-drop bug, matching the GitHub report
- [x] New parser-level tests: `except_for_type_list_excludes_both_types`, `except_for_type_list_splits_type_and_token_property`, `except_for_named_exception_does_not_misfire_as_subtype_negation` (`oracle_target.rs`)
- [x] New full-ability-line tests: `scourglass_destroy_all_excludes_artifacts_and_lands` (also asserts the trailing "Activate only during your upkeep." restriction still parses), `elspeth_tirel_minus_five_excludes_lands_and_tokens` (`oracle_tests.rs`)
- [x] New runtime test: `non_artifact_type_filter_excludes_artifact_creatures`, proving `Non(Artifact)` correctly excludes an artifact creature per CR 205.2b (`game/filter.rs`)
- [x] Full engine test suite re-run clean: **15692 passed, 0 failed, 0 regressions**

https://claude.ai/code/session_01DMa6DrxXyFHBdGxz3uLgvM

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -5770,6 +5770,69 @@ mod tests {
         assert!(matches_target_filter(&state, id, &card_filter, id));
     }
 
+    /// GitHub #4710 (Scourglass) / CR 205.2b: "except for artifacts" must
+    /// exclude an ARTIFACT CREATURE too, not just noncreature artifacts — an
+    /// object with more than one card type satisfies criteria for any of
+    /// them. `Permanent AND NOT Artifact AND NOT Land` (Scourglass's actual
+    /// filter shape) must reject the artifact creature, accept the plain
+    /// creature, and reject the land.
+    #[test]
+    fn non_artifact_type_filter_excludes_artifact_creatures() {
+        let mut state = setup();
+        let artifact_creature = add_creature(&mut state, PlayerId(0), "Ornithopter");
+        state
+            .objects
+            .get_mut(&artifact_creature)
+            .unwrap()
+            .card_types
+            .core_types
+            .push(CoreType::Artifact);
+        let plain_creature = add_creature(&mut state, PlayerId(0), "Bear");
+        let land_card_id = CardId(state.next_object_id);
+        let land = create_object(
+            &mut state,
+            land_card_id,
+            PlayerId(0),
+            "Forest".to_string(),
+            crate::types::zones::Zone::Battlefield,
+        );
+        state
+            .objects
+            .get_mut(&land)
+            .unwrap()
+            .card_types
+            .core_types
+            .push(CoreType::Land);
+
+        let scourglass_filter = TargetFilter::Typed(TypedFilter {
+            type_filters: vec![
+                TypeFilter::Permanent,
+                TypeFilter::Non(Box::new(TypeFilter::Artifact)),
+                TypeFilter::Non(Box::new(TypeFilter::Land)),
+            ],
+            controller: None,
+            properties: vec![],
+        });
+
+        assert!(
+            !matches_target_filter(
+                &state,
+                artifact_creature,
+                &scourglass_filter,
+                artifact_creature
+            ),
+            "an artifact creature must be excluded by 'except for artifacts' (CR 205.2b)"
+        );
+        assert!(
+            matches_target_filter(&state, plain_creature, &scourglass_filter, plain_creature),
+            "a plain (non-artifact, non-land) creature must survive"
+        );
+        assert!(
+            !matches_target_filter(&state, land, &scourglass_filter, land),
+            "a land must be excluded by 'except for lands'"
+        );
+    }
+
     #[test]
     fn self_filter() {
         let mut state = setup();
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[HIGH]** The new `except for <type-list>` suffix parser also consumes richer exception clauses like Flame Sweep's "except for creatures you control with flying," producing a contradictory `creature non-creature` filter. Evidence: `crates/engine/src/parser/oracle_target.rs:6211`. Why it matters: the current parse-diff sticky confirms an unintended `DamageAll` change for Flame Sweep from `creature` to `creature non-creature`, so the PR introduces parser blast radius outside the claimed Scourglass/Elspeth type-list class. Suggested fix: make `parse_except_for_type_list_suffix` decline unless every parsed list item is followed only by a list separator or clause boundary, leaving predicate-bearing forms like `except for creatures you control with flying` unsupported/honest until modeled correctly.

rykerwilliams:
Caught and fixed a real regression flagged by this PR's own \`coverage-parse-diff\` bot comment: Flame Sweep ("deals 2 damage to each creature **except for creatures you control with flying**") has a filtered-subset exception, not a bare type list. The original suffix parser greedily matched "creatures" as a type word and stopped at "you" (not a valid separator), silently emitting \`Non(Creature)\` alongside the base \`Creature\` filter — self-contradictory, matching zero creatures, which would have broken Flame Sweep's damage effect entirely.

Fixed by requiring the type-list suffix to actually terminate the clause (end-of-text or a following period) before accepting it — otherwise it declines and leaves the base filter untouched, same philosophy as the existing Mageta-the-Lion guard. Flame Sweep's flying-exception itself remains an out-of-scope, pre-existing gap (it needs a different "filtered-subset exception" mechanism, not this type-list one) — this fix only ensures my change doesn't make it worse. Added a regression test pinning the base filter. Full engine suite re-verified clean (15693 passed).

rykerwilliams:
@matthewevans Thanks for the review — this is the same issue I caught independently via the parse-diff sticky comment right around the same time. Already fixed in 76a1b8a5f: \`parse_except_for_type_list_suffix\` now requires the clause to actually terminate (end-of-text or a following period) after the parsed list items, and declines the whole suffix otherwise — so \`except for creatures you control with flying\` (Flame Sweep) is correctly left unhandled rather than producing the contradictory \`creature non-creature\` filter. Added a regression test (\`flame_sweep_filtered_subset_exception_does_not_corrupt_base_filter\`) pinning the base filter to plain \`Creature\`. The parse-diff sticky now shows only the 2 intended cards (Elspeth Tirel, Scourglass). Full engine suite re-verified clean. Should be ready for re-review.

matthewevans (approved):
Maintainer review: current head 76a1b8a5 is clean. Prior Flame Sweep parser blast-radius finding is addressed; parse-diff is scoped to Scourglass and Elspeth Tirel; CI is green.

PR #5359, opened 2026-07-08
MULTIPLAYER: resume server-hosted rooms after reload

davidomil:
## Summary

Restores server-hosted pregame rooms after a browser reload by persisting host-session metadata with the WebSocket reconnect token. Also moves LAN HTTP `crypto.randomUUID()` support into client source and fixes the join flow so connection-scoped seat reservation tokens are not carried from a lobby lookup socket into a later game join socket.

## Behavioral note

The reservation issue is easiest to hit on Full-mode servers. The official multiplayer deployment currently behaves as a LobbyOnly broker, so the pre-fix P2P join path resolved the guest over the same long-lived lobby socket that created the reservation. Full-mode self-hosted servers run the actual game on the server, so deck selection previously reserved a seat on the lobby socket, then the game route opened a new WebSocket and tried to use that token there. The server correctly rejected that as a foreign reservation.

Seat reservations are intentionally bound to the WebSocket connection that received them. This PR keeps the deck-selection lookup read-only and clears stale reservation handoff storage before opening the game WebSocket.

the pipeline):

> Frontend/client-only multiplayer persistence, browser runtime, and join-flow fixes. This does not change `crates/engine/` parser, effects, resolver, targeting, or rules behavior.

> [!NOTE]

## CR references

None.

## Verification

- `./node_modules/.bin/vitest run src/services/__tests__/multiplayerSession.test.ts src/polyfills/__tests__/cryptoRandomUUID.test.ts src/stores/__tests__/multiplayerStore.test.ts` — passed, 25 tests.
- `./node_modules/.bin/vitest run src/services/__tests__/brokerClient.test.ts src/stores/__tests__/multiplayerStore.test.ts src/services/__tests__/multiplayerSession.test.ts src/polyfills/__tests__/cryptoRandomUUID.test.ts` — passed, 34 tests.
- `node ../scripts/check-protocol-version.mjs && ./node_modules/.bin/tsc -b --noEmit` — passed.
- `./node_modules/.bin/eslint src/pages/MultiplayerPage.tsx src/providers/GameProvider.tsx src/stores/multiplayerStore.ts src/services/__tests__/brokerClient.test.ts` — passed.
- `cargo test -p lobby-broker foreign_` — passed, 2 tests.
- Live `ServerHello` comparison confirmed the official multiplayer deployment is LobbyOnly while the affected self-hosted deployment was Full mode.
- Self-hosted Full-mode deploy from `a56056306301f60a2f53cbb078d513ac7ec41378` verified the expected build commit, protocol version, and Full mode in `ServerHello`.
- Served bundle check in the self-hosted deployment confirmed join lookup calls the store without reservation options, and the online join path clears stale `phase-join-reservation:*` storage before navigation.
- Post-deploy server log check found no `You may only release or use a seat reservation issued to this connection` errors after the fix.
- Manual self-hosted Full-mode verification: a host room survived browser reload and a guest could join afterward.

### DIFF
--- a/client/src/hooks/useHostingSession.ts
+++ b/client/src/hooks/useHostingSession.ts
@@ -11,7 +11,12 @@ import { useMultiplayerStore } from "../stores/multiplayerStore";
 export function useHostingSession(): void {
   const pendingGameRoute = useMultiplayerStore((s) => s.pendingGameRoute);
   const clearPendingGameRoute = useMultiplayerStore((s) => s.clearPendingGameRoute);
+  const resumeServerHosting = useMultiplayerStore((s) => s.resumeServerHosting);
   const navigate = useNavigate();
+
+  useEffect(() => {
+    resumeServerHosting();
+  }, [resumeServerHosting]);
 
   useEffect(() => {
     if (pendingGameRoute) {
--- a/client/src/main.tsx
+++ b/client/src/main.tsx
@@ -1,3 +1,4 @@
+import "./polyfills/cryptoRandomUUID";
 import { createRoot } from "react-dom/client";
 // Self-hosted variable webfonts (served from node_modules by Vite — no Google
 // CDN). Newsreader = serif display; JetBrains Mono = codes / tabular numbers.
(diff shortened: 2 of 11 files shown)


matthewevans:
@davidomil is this something you had an opportunity to test?

davidomil:
Yes. Works on my local host. What I do is I have http reverse proxy with basic auth that goes into cloud flare zero trust tunnel. These changes were required for it to work reliably.

PR #5363, opened 2026-07-08
fix(parser): parse relative card-type target clauses

ntindle:
## Summary

- parse positive relative card-type clauses such as `that's an artifact, creature, or enchantment`
- preserve the already-parsed base filter on every relative type branch, including Legendary/Permanent restrictions
- add a Bounty Agent regression test for `target legendary permanent that's an artifact, creature, or enchantment`

## Root Cause

`parse_type_phrase_with_ctx` already handled positive relative subtype clauses like `that's an Ape or a Monkey`, but it had no equivalent path for relative card-type lists. The parser stopped after `legendary permanent`, leaving the trailing relative clause as unconsumed text.

## Validation

- `cargo fmt --all`
- `cargo test -p engine --lib parse_target_legendary_permanent_with_relative_type_union -- --nocapture`
- `cargo test -p engine --lib parse_type_phrase_ -- --nocapture`
- `cargo clippy -p engine --all-targets --features proptest -- -D warnings`

Note: Tilt was checked first, but no Tilt apiserver was running for this worktree, so direct cargo verification was used.

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -2342,6 +2342,7 @@ pub fn parse_type_phrase_with_ctx<'a>(
     // suffixes like "creatures". Stop before `Card` / `Subtype` — those are
     // informational suffixes ("creature card") or belong to the subtype slot.
     let mut extra_core_type_filters: Vec<TypeFilter> = Vec::new();
+    let mut relative_core_type_filters: Vec<TypeFilter> = Vec::new();
     if matches!(
         card_type,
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed the parser diff and current parse-diff evidence. The card-level changes all match the relative card-type restriction class claimed by the PR: Bounty Agent, Mizzix's Mastery, and Paladin Danse now preserve the existing base filter while adding the relative type leg(s). CI is green and the required parse-diff sticky is present for the current head.

PR #5219, opened 2026-07-07
feat(parser): compound-subject animation static — Life and Limb (CR 611.3)

minion1227:
## Summary
Adds engine support for **Life and Limb** — the dual-subject reciprocal land/creature animation — and the compound-subject animation *class* it represents.

Its printed static, `All Forests and all Saprolings are 1/1 green Saproling creatures and Forest lands in addition to their other types.`, silently dropped: no handler recognized the compound `all <X> and all <Y>` subject, and neither single-subject path is complete for the predicate — the land-animation path drops the trailing additive `and <type> lands`, and the creature-subtype path drops the base P/T. So the whole static (a board-defining continuous effect) was lost.

`parse_compound_all_subjects_type_change` (dispatched before `parse_land_animation`) distributes the subjects into an `Or` filter (CR 611.3 — the same continuous effect applies to every object in the affected set) and parses the compound predicate once into a uniform modification set that both subjects share:
- affected = `Or([ Land+Subtype("Forest"), Creature+Subtype("Saproling") ])` — the Forest conjunct resolves to a **land** subtype (CR 305.6), never a creature.
- mods = `[SetPower 1, SetToughness 1, SetColor([Green]), AddType Creature, AddSubtype "Saproling", AddType Land, AddSubtype "Forest"]` — base 1/1 (CR 613.4b, layer 7b), set green (CR 105.2, layer 5), additive types (CR 205.1b, layer 4).

The predicate reuses the two tested predicate parsers: `parse_animation_spec` (base P/T + set color + leading types) merged with `parse_additive_type_clause_modifications` (the additive type/subtype nouns past the animation parser's internal ` and `). The subject conjuncts reuse `parse_land_type_change_subject` (land subtypes) with a `parse_continuous_subject_filter` fallback (creature subtypes). No new runtime — every modification is an existing layer-system `ContinuousModification`.

## Files changed
- `crates/engine/src/parser/oracle_static/type_change.rs`
- `crates/engine/src/parser/oracle_static/dispatch.rs`
- `crates/engine/src/parser/oracle_static/tests.rs`

## Anchored on
- `crates/engine/src/parser/oracle_static/type_change.rs` `parse_land_animation` — the single-subject sibling; the new handler mirrors its animation-spec reuse and is dispatched immediately before it, guarded by the ` and all ` conjunction so single-subject lines fall through unchanged.
[...]

PR #5380, opened 2026-07-08
feat(parser): compound-quantified dual-subject become effect (CR 611.3)

### LINKED ISSUE #5377: feat(parser): compound-quantified dual-subject become effect (Nightcreep class) (CR 611.3)
## Problem

PR #5219 established the compound-subject static pattern for `All <X> and all <Y> are <predicate>` lines — distributing subjects into an `Or` filter and parsing the predicate once. A parallel **effect-layer** gap exists for the become-verb form:

```
Until end of turn, all creatures become black and all lands become Swamps.
```

(Nightcreep). Today the parser claims only the first conjunct (`Creature` → `SetColor(Black)`) and silently drops the second predicate into the static `description` field. The lands-never-become-Swamps half is lost.

## Proposed structural fix (mirror #5219)

Add `try_parse_compound_all_subjects_become_clause` in `oracle_effect/subject.rs`:

1. Peel conjuncts on the ` and all ` seam (same quantifier conjunction that distinguishes compound animation subjects in `parse_compound_all_subjects_filter`).
2. Per conjunct: split `subject become(s) predicate`; resolve subject via `parse_continuous_subject_filter`; lower predicate via existing `build_become_clause` building blocks.
3. Emit one `GenericEffect` carrying multiple `static_abilities` (same multi-static pattern as `build_defender_attack_continuous_compound`).
4. Dispatch immediately before `try_parse_subject_become_clause` so single-subject become lines fall through unchanged.

Factor the ` and all ` conjunct peeler into `oracle_static/static_helpers.rs` so `type_change.rs` and the effect handler share one authority.

## CR references

- CR 611.3 — a continuous effect applies to every object its text indicates (each conjunct is its own affected set).
- CR 105.2 / CR 613.1e — color-setting become (creatures → black).
- CR 305.7 — basic land type replacement (lands → Swamps).

## Acceptance

- Nightcreep parses to two `static_abilities`: `Creature → SetColor(Black)`, `Land → SetBasicLandType(Swamp)`.
- Single-subject `all creatures become black` unchanged.
- Shared-predicate compounds decline this handler.
- Regression tests for positive parse, per-conjunct mods, and decline guards.

## Track

Developer

claytonlin1110:
## Summary

Mirrors the **structural pattern** from #5219 (compound `"all <X> and all <Y>"` subject distribution) on the **effect layer** for become-verb compounds:

- Factors `peel_compound_all_quantified_conjuncts` into `static_helpers.rs` as the shared `" and all "` seam authority
- Refactors `parse_compound_all_subjects_filter` / `parse_compound_all_subjects_land_filter` in `type_change.rs` to reuse the peeler
- Adds `try_parse_compound_all_subjects_become_clause` dispatched before `try_parse_subject_become_clause`

Fixes Nightcreep misparsing: `all creatures become black and all lands become Swamps` now emits **two** `static_abilities` (`Creature → SetColor(Black)`, `Land → SetBasicLandType(Swamp)`) instead of dropping the land conjunct into `description`.

Closes #5377

## Test plan

- [x] `cargo check -p engine` — clean
- [ ] `cargo test -p engine --lib compound_all_subjects_become` — blocked locally by Windows `link.exe`; CI should verify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,7 +643,7 @@ checksum = "780955b8b195a21ab8e4ac6b60dd1dbdcec1dc6c51c0617964b08c81785e12c9"
 
 [[package]]
 name = "draft-core"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "engine",
  "rand 0.9.4",
@@ -655,7 +655,7 @@ dependencies = [
 
 [[package]]
 name = "draft-wasm"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "console_error_panic_hook",
  "draft-core",
@@ -727,7 +727,7 @@ dependencies = [
 
 [[package]]
 name = "engine"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "assert_matches",
  "im",
@@ -754,7 +754,7 @@ dependencies = [
 
 [[package]]
 name = "engine-inventory-gen"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "anyhow",
  "proc-macro2",
@@ -767,7 +767,7 @@ dependencies = [
 
 [[package]]
 name = "engine-wasm"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "console_error_panic_hook",
  "engine",
@@ -819,7 +819,7 @@ checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "feed-scraper"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "clap",
  "reqwest",
@@ -1601,7 +1601,7 @@ checksum = "92daf443525c4cce67b150400bc2316076100ce0b3686209eb8cf3c31612e6f0"
 
 [[package]]
 name = "lobby-broker"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "engine",
  "serde",
@@ -1632,7 +1632,7 @@ checksum = "c41e0c4fef86961ac6d6f8a82609f55f31b05e4fce149ac5710e439df7619ba4"
 
 [[package]]
 name = "manabrew-compat"
-version = "0.18.0"
+version = "0.19.0"
 dependencies = [
  "engine",
  "pretty_assertions",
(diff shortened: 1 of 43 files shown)


matthewevans (changes requested):
**[MED]** The new compound-become parser accepts `becomes` in the split but then rejects that same verb before building the clause. Evidence: `crates/engine/src/parser/oracle_effect/subject.rs:626` accepts `" becomes "`, but `crates/engine/src/parser/oracle_effect/subject.rs:634` immediately requires `tag("become ")`; the added coverage at `crates/engine/src/parser/oracle_effect/tests.rs:2236` only exercises plural `"become"`. Why it matters: any singular/third-person conjunct that reaches this handler falls back instead of producing the intended per-conjunct static, so the new building block is narrower than its own grammar claims. Suggested fix: accept both `become ` and `becomes ` at the validation/build boundary, and add a test with a `becomes` conjunct.

matthewevans (changes requested):
**[MED]** Player hexproof uses the source object's stored controller instead of the targeting call's controller. Evidence: `crates/engine/src/game/targeting.rs:16` carries `source_controller`, but the player-target branch calls `player_cannot_be_targeted_by(state, player.id, source_id)` without it at `crates/engine/src/game/targeting.rs:204`, and `crates/engine/src/game/static_abilities.rs:1436` recomputes the controller from `state.objects[source_id]`; the repo already has an explicit override path at `crates/engine/src/game/targeting.rs:57`. Why it matters: when the ability/spell controller differs from the source object record, player hexproof can evaluate the wrong opponent relationship and allow an illegal player target. Suggested fix: pass the authoritative `source_controller` into `player_cannot_be_targeted_by` from all player-target enumeration sites and use that for the CR 702.11c opponent check.

matthewevans (changes requested):
**[MED]** Player hexproof uses the source object's stored controller instead of the targeting call's controller. Evidence: `crates/engine/src/game/targeting.rs:16` carries `source_controller`, but the player-target branch calls `player_cannot_be_targeted_by(state, player.id, source_id)` without it at `crates/engine/src/game/targeting.rs:204`, and `crates/engine/src/game/static_abilities.rs:1436` recomputes the controller from `state.objects[source_id]`; the repo already has an explicit override path at `crates/engine/src/game/targeting.rs:57`. Why it matters: when the ability/spell controller differs from the source object record, player hexproof can evaluate the wrong opponent relationship and allow an illegal player target. Suggested fix: pass the authoritative `source_controller` into `player_cannot_be_targeted_by` from all player-target enumeration sites and use that for the CR 702.11c opponent check.

matthewevans (approved):
Maintainer sign-off: current head addresses the become/becomes and source-controller targeting blockers; parser capability is at the right seam with discriminating coverage.

PR #5381, opened 2026-07-08
Fix remaining zone-of-function statics gathers (#1033 follow-up to #5350)

rykerwilliams:
## Summary
Follow-up to #5350, which fixed Underworld Breach's specific self-recast bug (#1033) by narrowing `active_continuous_effects_from_base_static_source`'s admission via a caller-side pre-filter. That fix is scoped to the one call path Underworld Breach hits. While investigating the same bug class, I found five other statics-gathering functions with the identical "no zone-of-function check outside the command zone" defect, still live on `main` after #5350 merged:

- `active_static_definitions`, `game_functioning_statics`, `battlefield_functioning_statics` (`functioning_abilities.rs`) had no `active_zones` gate at all outside the command-zone branch.
- `active_combat_assignment_rule_effects_from_static_definitions` (`layers.rs`, distinct from the sibling `active_continuous_effects_from_static_definitions` that #5350 already addresses) skipped its gate entirely whenever `active_zones` was empty.
- `compute_combat_tax` (`combat.rs`) had no zone gate at all.

Fix: a single shared predicate, `static_functions_in_zone`, now backs all five (CR 113.6/113.6b general case, CR 114.3/114.4 emblem exemption for the command zone, CR 113.6g stack exception for self-referential `CantBeCountered`/`CantBeCopied` distinguished from grants-to-others via `affected`). Also deletes a redundant, independently buggy duplicate (`graveyard_permission_functions_in_zone`) that never exempted emblems, unlike every other command-zone consumer.

**Deliberately out of scope:** `active_continuous_effects_from_static_definitions` and its `active_continuous_effects_from_base_static_source` caller — that's #5350's fix, left untouched here. The module doc's "Zone scope asymmetry" note flags where the two approaches currently diverge (notably: #5350's filter admits any self-referential `affected` definition regardless of declared `active_zones`, where this predicate requires the explicit CR 113.6b opt-in) — raised as a discussion point for `@matthewevans`, not asserted as a defect in the merged fix.

## Files changed
- `crates/engine/src/game/functioning_abilities.rs`
- `crates/engine/src/game/layers.rs`
- `crates/engine/src/game/combat.rs`
- `crates/engine/src/game/casting.rs`

## CR references
CR 113.6, CR 113.6b, CR 113.6g, CR 114.3, CR 114.4, CR 604.1, CR 613.1, CR 702.26b — all verified against `docs/MagicCompRules.txt` before use.

## Track
Developer

## LLM
Model: claude-sonnet-5
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — clean (no parser files touched, exit 0)
- `cargo clippy -p engine --all-targets -- -D warnings` — clean, 0 warnings
- `cargo test -p engine --lib` — 15762 passed, 0 failed, 6 ignored
- Each migration backed by a permanent regression test; several proven via live revert-and-rerun during review (temporarily disable the gate, confirm the test fails, restore, confirm it passes) — including a matrix covering self-referential vs. granting `CantBeCountered` on the stack vs. battlefield with two co-resident definitions gated independently, proving per-definition (not per-object) discrimination
- Confirmed the three existing off-zone tests that #5350 already made pass (`self_static_in_graveyard_grants_keyword_to_self`, `off_zone_self_statics_use_base_static_definitions`, `self_graveyard_static_flashback_grant_is_castable`) still pass unmodified — this PR doesn't touch their code path

## Gate A
```
(no output, exit 0 — no parser files in this diff)
```

## Anchored on
- `crates/engine/src/game/triggers.rs:1034` — `trigger_definition_functions_in_zone`, the pre-existing trigger-side mirror of this exact CR 113.6/113.6b rule (`if trigger_zones.is_empty() { zone == Battlefield } else { trigger_zones.contains(&zone) }`), which the new `static_functions_in_zone` follows for the general case.
- `crates/engine/src/game/functioning_abilities.rs:79` — `non_emblem_command_zone_static_functions`, the pre-existing Command-zone emblem-opt-in authority, reused unchanged as the Command branch of the new shared predicate rather than re-derived.

## Tier
Tier: Standard

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -3012,38 +3012,34 @@ fn graveyard_permission_sources(
             if !source_belongs_to_player {
                 return None;
             }
-            active_static_definitions(state, obj)
-                .filter(|definition| graveyard_permission_functions_in_zone(definition, obj.zone))
-                .find_map(|definition| match definition.mode {
-                    StaticMode::GraveyardCastPermission {
-                        frequency,
-                        play_mode,
-                        graveyard_destination_replacement,
-                        ref extra_cost,
-                    } if graveyard_permission_play_mode_matches(play_mode, play_mode_filter) => {
-                        definition
-                            .affected
-                            .as_ref()
-                            .map(|filter| GraveyardPermissionSource {
-                                source_id,
-                                filter,
-                                frequency,
-                                graveyard_destination_replacement,
-                                extra_cost,
-                            })
-                    }
-                    _ => None,
-                })
+            // The zone-of-function gate is now fully owned by
+            // `active_static_definitions` (CR 113.6 / CR 113.6b), which also
+            // correctly admits emblem-sourced graveyard-cast permissions —
+            // the previously-inlined gate never exempted `is_emblem` unlike
+            // every other command-zone consumer, an independent latent bug
+            // now fixed as a side effect.
+            active_static_definitions(state, obj).find_map(|definition| match definition.mode {
+                StaticMode::GraveyardCastPermission {
+                    frequency,
+                    play_mode,
+                    graveyard_destination_replacement,
+                    ref extra_cost,
+                } if graveyard_permission_play_mode_matches(play_mode, play_mode_filter) => {
+                    definition
+                        .affected
+                        .as_ref()
+                        .map(|filter| GraveyardPermissionSource {
+                            source_id,
+                            filter,
+                            frequency,
+                            graveyard_destination_replacement,
+                            extra_cost,
+                        })
+                }
+                _ => None,
+            })
         })
         .collect()
-}
-
-fn graveyard_permission_functions_in_zone(definition: &StaticDefinition, zone: Zone) -> bool {
-    if zone == Zone::Battlefield {
-        definition.active_zones.is_empty() || definition.active_zones.contains(&Zone::Battlefield)
-    } else {
-        definition.active_zones.contains(&zone)
-    }
 }
 
 fn graveyard_permission_play_mode_matches(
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
**[MED]** `compute_combat_tax` still skips opt-in non-emblem command-zone statics. Evidence: `crates/engine/src/game/combat.rs:1887` iterates command-zone sources but skips every non-emblem before calling `static_functions_in_zone`; meanwhile `crates/engine/src/game/functioning_abilities.rs:499` and `crates/engine/src/game/static_source_index.rs:101` show non-emblem command-zone objects are supported when their statics opt into `Zone::Command`. Why it matters: planes, schemes, conspiracies, and other supported command-zone static sources can still be dropped before their `CantAttack`/`CantBlock` `UnlessPay` definitions are evaluated. Suggested fix: iterate the shared functioning-static source set for tax sources, or use `object_sources_static_from_command_zone` instead of the emblem-only command gate.

rykerwilliams:
Fixed in e865c5444 — replaced the emblem-only outer gate in \`compute_combat_tax\` with \`object_sources_static_from_command_zone\`, matching the admission rule every other command-zone gather already uses. Added a test proving a non-emblem, Command-opted-in source is now admitted (confirmed via live revert-and-rerun it fails without the fix) plus a negative sibling for the non-opted-in case. Full suite + clippy clean.

matthewevans (approved):
Maintainer review: current head resolves the remaining zone-of-function statics issues; parse-diff scope checked.

PR #5382, opened 2026-07-08
fix(parser): recognize "skip their next <step>" for each-player step skips

joaovictor712:
## What

"Each opponent skips their next untap step" (**Brine Elemental**) dropped to `Unimplemented`. The for-each-player distributor correctly strips the `each opponent` head (setting `player_scope: Opponent`) and hands the per-player body `skip their next untap step` to the effect parser — but `try_parse_skip_next_step` only recognized the first-person `[you ]skip your next <step>` form, not the third-person `skip their next <step>` a distributor produces.

Adds `"skip their next "` alongside the existing `"skip your next "` tag. The body still lowers to the same `SkipNextStep { target: Controller }`, which resolves relative to each iterating player, so every step the skip parser already handles (untap / draw / upkeep) now works under an `each opponent` / `each player` head.

## No engine change

`Effect::SkipNextStep` and the player-scope distributor are unchanged; this is purely a parser-recognition gap. (The `target player` / `that player` subjects already parsed via the declared-target and anaphor paths.)

## Tests

Asserts the each-opponent chain lowers to `player_scope: Opponent` + `SkipNextStep { Controller, Untap, NextOccurrence }`; existing `skip your next` / `target opponent skips` tests unchanged.

Verified locally: `cargo fmt`, full engine lib suite (15,771 passed / 0 failed), `clippy -p engine --all-targets -D warnings` clean; rebased onto latest `main`.

## CR references
CR 614.10a (skip a step) · CR 102.2 / CR 115.1 (player references).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -9204,10 +9204,17 @@ fn try_parse_still_a_type(tp: TextPair) -> Option<ParsedEffectClause> {
 /// CR 614.10a: Parse "[subject] skip[s] [their|your] next [step] step[s]" —
 /// one-shot step skips. Handles controller and target-player forms.
 fn try_parse_skip_next_step(tp: TextPair, ctx: &ParseContext) -> Option<ParsedEffectClause> {
+    // CR 614.10a: "[you ]skip your next <step>" — controller-subject one-shot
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Maintainer sign-off: parser change is at the existing skip-step seam, uses the house nom combinator pattern, and has discriminating production-path coverage for Brine Elemental.

PR #5384, opened 2026-07-08
feat(parser): N-way Oxford-comma compound-subject keyword static — Shalai, Voice of Plenty (CR 611.3a)

### LINKED ISSUE #5383: feat(parser): N-way Oxford-comma compound-subject keyword static — Shalai, Voice of Plenty (CR 611.3a)
## Summary

`parse_compound_subject_keyword_static` (`crates/engine/src/parser/oracle_static/evasion.rs:377`) parses the compound player+object keyword-grant pattern `"You and <object subject> have <keyword>"` (Sigarda, Heron's Grace; Serra's Emissary; Gruul Spellbreaker) into a decomposed pair of `StaticDefinition`s — an object-half `Continuous`/`AddKeyword` def and a player-half `StaticMode::Hexproof`/`Shroud`/`PlayerProtection` def, since a single `StaticDefinition` can't carry both a player scope and an object scope.

Its subject match is hard-coded to the literal seam `nom_tag_tp(&body, "you and ")` — exactly one `"and"`, exactly one object conjunct. **Shalai, Voice of Plenty**'s real Oracle text uses a 3-item Oxford-comma list instead of a bare 2-item "and":

```
Flying
You, planeswalkers you control, and other creatures you control have hexproof.
{4}{G}{G}: Put a +1/+1 counter on each creature you control.
```

(verified via Scryfall `oracle_id: a0c47ab6-dfb4-46ee-a3f7-9e1521b4bb4b`)

Because the line starts with `"You, "` (comma) rather than `"You and "`, the tag match fails immediately and the line falls through to the generic static fallback, which can't resolve it either — `client/public/card-data.json` currently records it as:

```
"type":"Unimplemented","name":"static_structure",
"description":"Static pattern matched but line failed static parser: You, planeswalkers you control, and other creatures you control have hexproof."
```

So Shalai's entire signature ability — a board-defining hexproof umbrella over the controller, their planeswalkers, and their other creatures — is silently dropped.

## Why this is a class, not a card

The other half of this problem — collapsing an N-way Oxford-comma **object-only** list into `TargetFilter::Or` — already works and is independently tested (`parse_type_phrase` + `distribute_controller_to_or` in `oracle_target.rs`, proven by the Silkguard test at `oracle_target.rs:11151` and by *Kismet*'s "Artifacts, creatures, and lands your opponents control enter tapped."). But that machinery only fires when a single trailing suffix (e.g. one `"you control"`) is meant to distribute backward across type-noun-only legs with no per-leg clause of their own.

Shalai's shape is different: `"you"` is the first list item, and each subsequent conjunct (`"planeswalkers you control"`, `"other creatures you control"`) is a **complete, independently-resolvable subject phrase** in its own right — exactly the shape `parse_rule_static_subject_filter` already resolves one at a time inside the existing 2-item function. Nothing currently peels the leading `"you"` token off an arbitrary-length Oxford list and Or-combines the resolved remainder.

This is a straight generalization of the existing function's "object subject" axis from exactly-one-conjunct to N-conjuncts — the same shape of change as #5219 (compound-subject animation statics), which generalized a single-subject dispatcher to a two-subject `"all X and all Y"` form by reusing existing per-domain subject/predicate parsers and Or-distributing. Per this repo's "parameterize, don't proliferate" convention, the fix is to generalize `parse_compound_subject_keyword_static`'s subject recognition (accept `"you, "` as an alternate lead to `"you and "`, then Or-combine 2+ comma-separated conjuncts each resolved via the existing `parse_rule_static_subject_filter`), not to hand-roll a Shalai-specific branch.

## Proposed fix

- Generalize the leading-tag match in `parse_compound_subject_keyword_static` to accept `"you, "` (Oxford-comma form) in addition to the existing `"you and "` (bare 2-item form), gated so the existing single-object literal-`"and"` path is completely unchanged (avoids any risk to the "shared trailing suffix, no comma" ambiguity that the 2-item form's existing fallthrough to `parse_type_phrase` already resolves correctly).
- Add an Oxford-comma conjunct splitter (nom-based, reusing `nom_primitives::split_once_on`, mirroring the loop shape in #5219's `parse_compound_all_subjects_filter`) that peels `", "`-separated conjuncts plus the final `", and "`/`"and "` connector, resolving each conjunct through the existing `parse_rule_static_subject_filter` and Or-combining.
- Apply the existing "no second player scope" guard (`rule_static_affected_is_player_scope`) per conjunct.
- Both existing call sites of `parse_compound_subject_keyword_static` (the direct dispatch in `shared.rs:1272` and the inverted `"As long as <cond>, …"` rewrite path at `shared.rs:1233`) pick up the generalization automatically since it's one function, not a new sibling.

## CR references

- `CR 611.3a` — a continuous effect generated by a static ability applies to whatever its text indicates.
- `CR 702.11a`/`CR 702.11b` — hexproof.
- `CR 702.16` — protection.
- `CR 702.18` — shroud.

## Verification

- `client/public/card-data.json` → `"shalai, voice of plenty"` → `abilities[0].effect.type == "Unimplemented"` at time of filing (repo tip `39a332d26`).
- Confirmed no other real, currently-printed card matches the same 3+-way `"You, …, and … have/has"` shape via a Scryfall full-text search (`o:/you, .+, and .+ have/i` → 1 result, Shalai only) — the fix is scoped as a general N-way capability (mirroring the arity generalization other compound-subject dispatchers in this file already apply), not a Shalai-only special case.

jsdevninja:
## Summary

Fixes #5383.

`parse_compound_subject_keyword_static` (`crates/engine/src/parser/oracle_static/evasion.rs`) decomposes the compound player+object keyword-grant pattern `"You and <object subject> have <keyword>"` (Sigarda, Heron's Grace; Serra's Emissary; Gruul Spellbreaker) into an object-half `Continuous`/`AddKeyword` def and a player-half `Hexproof`/`Shroud`/`PlayerProtection` def, since a single `StaticDefinition` can't carry both a player scope and an object scope.

Its subject match was hard-coded to the literal seam `"you and "` — exactly one object conjunct. **Shalai, Voice of Plenty**'s real Oracle text uses a 3-item Oxford-comma list instead:

```
Flying
You, planeswalkers you control, and other creatures you control have hexproof.
{4}{G}{G}: Put a +1/+1 counter on each creature you control.
```

(verified via Scryfall, `oracle_id: a0c47ab6-dfb4-46ee-a3f7-9e1521b4bb4b`)

Because the line starts with `"You, "` rather than `"You and "`, the tag match failed immediately and the line fell through to the generic fallback, which couldn't resolve it either — `client/public/card-data.json` recorded it as `Unimplemented("static_structure", "Static pattern matched but line failed static parser: …")`. Shalai's entire signature ability was silently dropped.

## Why this is a class, not a card

The object-only N-way Oxford-comma case already works (`parse_type_phrase` + `distribute_controller_to_or`, proven by the Silkguard test and *Kismet*), but only when a single trailing suffix distributes backward across bare type-noun legs with no clause of their own. Shalai's shape is different: `"you"` is the first list item, and each subsequent conjunct (`"planeswalkers you control"`, `"other creatures you control"`) is a **complete, independently-resolvable subject phrase** — exactly what `parse_rule_static_subject_filter` already resolves one at a time inside the existing 2-item function. Nothing peeled the leading `"you"` token off an arbitrary-length Oxford list and Or-combined the resolved remainder.

This is a straight generalization of the function's "object subject" axis from exactly-one-conjunct to N-conjuncts, following this repo's "parameterize, don't proliferate" convention rather than hand-rolling a Shalai-specific branch.

## Changes

- `evasion.rs`: `parse_compound_subject_keyword_static` now accepts `"you, "` (Oxford-comma N-item form) as an alternate lead to the existing `"you and "` (bare 2-item form). The comma gate keeps the two grammars disjoint — the 2-item path's existing fallthrough to `parse_rule_static_subject_filter` (which itself resolves an object subject with an internal bare `"and"`, e.g. `"artifacts and creatures you control"`, via `parse_type_phrase`'s own trailing-suffix distribution) is completely unchanged.
- New `parse_oxford_object_conjuncts` helper: nom-based (`split_once_on`), peels `", "`-separated conjuncts plus the final `", and "`/`"and "` connector, resolves each conjunct through the existing `parse_rule_static_subject_filter`, and Or-combines them. Applies the same "no second player scope" guard the 2-item form already applies, now per conjunct.
- Both existing call sites of `parse_compound_subject_keyword_static` (the direct dispatch and the inverted `"As long as <cond>, …"` rewrite path) pick up the generalization automatically since it's one function, not a new sibling dispatcher.
- `shared.rs`: updated the dispatch-site comment to document the new N-item form.
- `tests.rs`: flips no existing test (none previously exercised this exact shape); adds:
  - `compound_subject_keyword_static_splits_shalai_three_way` — the real card, asserting the 2-way object `Or` (planeswalkers you control; other creatures you control) and the player-half `Hexproof` mode.
  - `compound_subject_keyword_static_declines_oxford_list_unresolvable_conjunct` — boundary guard: an unresolvable conjunct must strict-fail the whole line, not silently drop the bad disjunct.
  - `compound_subject_keyword_static_splits_four_item_oxford_list` — arity generality check (synthetic 4-item list → 3-way object `Or`), proving the splitter isn't hard-coded to Shalai's 3-item arity.

## CR references

- `CR 611.3a` — a continuous effect generated by a static ability applies to whatever its text indicates.
- `CR 702.11a`/`CR 702.11b` — hexproof.
- `CR 702.16` — protection.
- `CR 702.18` — shroud.

## Verification

- `cargo fmt -p engine -- --check` — clean, no diff.
- **Local `cargo check`/`cargo test` could not be run to completion in the environment this PR was authored in** — the sandbox has no working MSVC linker (`link.exe` on `PATH` resolves to Git for Windows' coreutils `link`, not the MSVC linker) and no Docker/WSL fallback, so native build-script compilation fails for unrelated foundational crates (`serde`, `proc-macro2`, etc.) before reaching this crate's code. I verified the change through careful manual review (type/signature cross-checks against every function called, hand-traced the split logic against the real Shalai text and three synthetic inputs) and via `cargo fmt`'s parse-clean check, but **please treat CI as the first real compile/test signal for this PR** and let me know if anything surfaces.

## Scope Expansion

None.

### DIFF
--- a/crates/engine/src/parser/oracle_static/evasion.rs
+++ b/crates/engine/src/parser/oracle_static/evasion.rs
@@ -501,8 +501,13 @@ fn parse_oxford_object_conjuncts(text: &str) -> Option<TargetFilter> {
         conjuncts.push(item.trim());
         remaining = rest;
     }
-    let last = remaining.trim();
-    conjuncts.push(last.strip_prefix("and ").unwrap_or(last).trim());
+    // Pattern 1 (PATTERNS.md): strip the Oxford "and " connector from the final
+    // conjunct via `opt(tag(...))` rather than `strip_prefix` — the connector is
+    // genuinely optional (absent when only 2 total object conjuncts follow "you").
+    let (last, _) = opt(tag::<_, _, OracleError<'_>>("and "))
+        .parse(remaining.trim())
+        .ok()?;
+    conjuncts.push(last.trim());
     if conjuncts.len() < 2 {
         return None;
     }
--- a/crates/engine/src/parser/oracle_static/evasion.rs
+++ b/crates/engine/src/parser/oracle_static/evasion.rs
@@ -353,12 +353,16 @@ pub(crate) fn parse_compound_subject_rule_static(
 }
 
 /// CR 702.11 + CR 702.16 + CR 702.18 + CR 611.3a: Compound-subject keyword-grant
-/// statics of the form `"You and <object subject> have <keyword>"` — a single
-/// keyword grant bound to a player plus an object subset.
+/// statics of the form `"You and <object subject> have <keyword>"` (the bare
+/// 2-item form) or `"You, <object subject>, …, and <object subject> have
+/// <keyword>"` (the Oxford-comma N-item form; Shalai, Voice of Plenty: "You,
+/// planeswalkers you control, and other creatures you control have hexproof.")
+/// — a single keyword grant bound to a player plus one or more object subsets.
 ///
 /// A single `StaticDefinition` cannot carry both a player scope and an object
 /// scope, so decompose into two:
-///   - an object-half `Continuous` def whose `affected` is the object subset;
+///   - an object-half `Continuous` def whose `affected` is the object subset
+///     (an `Or` of every conjunct's filter when 2+ object subjects are listed);
 ///   - a player-half def whose mode is the player-applicable keyword mode
 ///     (`PlayerProtection` / `Hexproof` / `Shroud`) and whose `affected` is the
 ///     controller.
@@ -366,6 +370,9 @@ pub(crate) fn parse_compound_subject_rule_static(
 /// Object subjects reuse [`parse_rule_static_subject_filter`] so subtype scopes
 /// ("Humans you control"), self refs ("this creature"), and "other <subtype>
 /// you control" all resolve — not a hard-coded alt of three controller phrases.
+/// The N-item form delegates conjunct splitting to
+/// [`parse_oxford_object_conjuncts`], which resolves each conjunct through that
+/// same subject resolver rather than a bespoke list grammar.
 ///
 /// Only player-applicable keywords claim this pattern (a player cannot
 /// meaningfully "have flying"). Leading `"During your turn, "` gates both
@@ -387,8 +394,18 @@ pub(crate) fn parse_compound_subject_keyword_static(
         (input, None)
     };
 
-    // Subject: "you and <object subject phrase>".
-    let after_you = nom_tag_tp(&body, "you and ")?;
+    // Subject: the bare 2-item form "you and <object subject phrase>", or the
+    // Oxford-comma N-item form "you, <object subject phrase>, …". The comma-led
+    // form is a disjoint grammar path from the bare "and" form — gating on the
+    // leading comma leaves the 2-item path's existing fallthrough to
+    // `parse_rule_static_subject_filter` (which itself resolves an object
+    // subject with an internal bare "and", e.g. "artifacts and creatures you
+    // control", via `parse_type_phrase`'s own trailing-suffix distribution)
+    // completely unchanged.
+    let (after_you, multi_object) = match nom_tag_tp(&body, "you, ") {
+        Some(rest) => (rest, true),
+        None => (nom_tag_tp(&body, "you and ")?, false),
+    };
 
     // Locate the continuous predicate verb ("have"/"has"/"gain"/"gains"/…) so
     // the object subject can be any phrase `parse_rule_static_subject_filter`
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Maintainer review: current head cleanly handles the Oxford-comma compound-subject static class; parse-diff limited to Shalai.

PR #5387, opened 2026-07-08
fix(engine): honor multi_target on Attach so Armory Automaton attach can be skipped (#5339)

### LINKED ISSUE #5339: Armory Automaton — ETB/attack attach trigger cannot be skipped
Reported in Discord: https://discord.com/channels/1485498006781427802/1524166762495475742/1524166762495475742

**Thread:** Armory Automaton
**Discord thread id:** `1524166762495475742`
**Discord message id:** `1524166762495475742`
**Cards:** armory automaton
**Extraction confidence:** manual

## Summary
Can't skip its enter ability even though it says "you may attach any number of target Equipment".

### Armory Automaton
Whenever this creature enters or attacks, you may attach any number of target Equipment to it. (Control of the Equipment doesn't change.)

---
<sub>report_id: `discord:1524166762495475742:1524166762495475742:0` · discord: `1524166762495475742/1524166762495475742`</sub>

andriypolanski:
## Summary

Armory Automaton's ETB/attack trigger — *"you may attach any number of target Equipment to it"* — could not be skipped when Equipment was on the battlefield. The parser already emitted `multi_target { min: 0, max: unlimited }`, but `Effect::Attach` used a dedicated target-slot path that ignored `multi_target` and treated the Equipment slot as **required** (`optional_targeting` only).

Closes [#5339](https://github.com/phase-rs/phase/issues/5339).

## Root cause

`collect_target_slots` / `collect_target_slot_specs` for `Effect::Attach` pushed a **single** attachment slot with `optional: ability.optional_targeting`. Armory Automaton carries zero-min cardinality on `multi_target`, not `optional_targeting`, so the slot was mandatory and the UI could not skip (Escape / skip-target did nothing useful).

The generic multi-target path (used by `ChangeZone`, `DealDamage`, etc.) already expands `multi_target` into N optional slots; Attach bypassed it.

### Engine — attach target slots honor `multi_target`

- `crates/engine/src/game/ability_utils.rs`
  - New helpers: `collect_attach_attachment_target_slots`, `collect_attach_attachment_target_slot_specs`, `assign_attach_attachment_*`
  - Attachment operand now mirrors generic `multi_target` expansion (one shared instance per CR 115.3)
  - Uses `targeting_is_optional()` (covers both `optional_targeting` and `multi_target.min == 0`)

### Tests

- Unit: `attach_any_number_multi_target_surfaces_optional_equipment_slots` in `triggers.rs`
- Integration: `crates/engine/tests/integration/issue_5339_armory_automaton.rs`
  - Decline optional "you may" on ETB → no stall, Equipment stays unattached
  - Accept optional + choose zero Equipment at stack → resolves cleanly
  - Attack trigger: skip zero Equipment at `TriggerTargetSelection` → no stall
- Registered in `tests/integration/main.rs`

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test -p engine attach_any_number_multi_target_surfaces_optional_equipment_slots --lib`
- [x] `cargo test -p engine --test integration issue_5339_armory_automaton`

## Risk notes

- Scoped to `Effect::Attach` attachment-operand slot generation and target assignment; host-operand (creature) slots unchanged.
- Super-Soldier Serum / other "attach any number of target Equipment" triggers benefit from the same building block.


matthewevans (changes requested):
**[HIGH]** Attach multi-target assignment can consume the host target as another attachment. Evidence: `crates/engine/src/game/ability_utils.rs:2906` and `crates/engine/src/game/ability_utils.rs:2958` compute the attachment bounds from all remaining selected targets, before the explicit host slot is consumed separately. Why it matters: for an `Attach` effect with variable attachment targets plus a non-context host target, the host selection can be folded into `ability.targets` as an attachment and then the host slot can be skipped as optional, so resolution sees the wrong target partition. Suggested fix: consume exactly the attachment slot window produced by slot surfacing, then consume the host slot separately; add a fixture where `target` is not `SelfRef`.

matthewevans (changes requested):
**[HIGH]** Attach multi-target assignment still consumes the explicit host on the compact declared-target path. Evidence: `crates/engine/src/game/ability_utils.rs:2967` loops over `bounds.max` in `assign_attach_attachment_declared_targets`, and `crates/engine/src/game/ability_utils.rs:5468` calls that path before assigning the host slot. If the caller provides compact targets like `[equipment, host]`, the host is still consumed as the second attachment when another Equipment is legal, leaving the real host missing. Why it matters: the PR fixes the interactive selected-slot path but leaves another production assignment entry point with the same host-consumption bug. Suggested fix: mirror the selected-slot windowing for declared targets, stopping the attachment operand before the explicit host slot, and add a regression that drives `assign_targets_in_chain` with one chosen Equipment plus a required host.

matthewevans (approved):
Maintainer approval: security scan clean, stale requested-change findings addressed at current head, proof gap false, no parse changes, and GitHub checks pass.

PR #5389, opened 2026-07-08
fix(engine): scope hexproof-bypass statics to their beneficiary (Glaring Spotlight)

### LINKED ISSUE #5386: [Card Bug] Glaring Spotlight — hexproof bypass with "you control" qualifier doesn't parse
### Card(s)

Glaring Spotlight

### Bug

Glaring Spotlight's static ability drops to `Unimplemented` and has no effect:

> Creatures your opponents control with hexproof can be the targets of spells and abilities you control as though they didn't have hexproof.

The engine already models the hexproof bypass (`StaticMode::IgnoreHexproof`) and parses the unqualified form (Nowhere to Run: "…targets of spells and abilities as though they didn't have hexproof"). The gap is purely the optional **"you control"** qualifier between "spells and abilities" and "as though" — `parse_ignore_hexproof_static` matches the phrase verbatim as "of spells and abilities as though", so the qualified form fails to parse and the bypass is never emitted. In actual play, Glaring Spotlight's hexproof-bypass does nothing.

### Expected

The line parses to a single `IgnoreHexproof` static scoped to creatures the opponents control (as it does for Nowhere to Run), so hexproof creatures your opponents control can be targeted by your spells and abilities.

### Notes

The compound player+permanent subject variants (Kaya, Bane of the Dead; Detection Tower — "your opponents and creatures your opponents control with hexproof …") additionally require player-subject hexproof bypass and are out of scope here.

joaovictor712:
Closes #5386.

## What

Glaring Spotlight — *"Creatures your opponents control with hexproof can be the targets of spells and abilities **you control** as though they didn't have hexproof."* — restricts the hexproof bypass to the **static controller's** spells and abilities. Nowhere to Run's unqualified form (*"...can be the targets of spells and abilities as though they didn't have hexproof"*) opens the affected creatures to **every** player.

The parser previously consumed the `you control` qualifier and **discarded** it, emitting an identical unrestricted `IgnoreHexproof` for both cards. That is correct in 1v1 but **wrong in multiplayer**: a third player would wrongly be able to target the affected creatures through Glaring Spotlight.

## How (building block, not a special case)

Parameterize the bypass **beneficiary** on `StaticDefinition` — mirroring the existing mode-specific optional fields `attack_defended` (CantAttack) and `source_controller` (lure statics):

```rust
pub bypass_beneficiary: Option<ControllerRef>,
//   None      = every player's spells and abilities (Nowhere to Run)
//   Some(You) = the static controller's only          (Glaring Spotlight)
```

- **Parser** (`oracle_static/shared.rs`): `opt(tag(" you control"))` now drives `bypass_beneficiary = you_control.then_some(ControllerRef::You)` instead of being discarded.
- **Resolver** (`static_abilities::target_ignores_hexproof`): now takes the targeting source's controller and gates the grant through `ignore_hexproof_beneficiary_allows` — `None` ⇒ any player; `Some(You)` ⇒ only the static controller (CR 109.5); other refs fail closed.
- Serde-defaulted, so all existing serialized statics (unrestricted) round-trip unchanged.

## CR

Corrects the citations flagged in review: the *"as though it didn't have hexproof"* targeting permission is **CR 702.11e / CR 609.4**, not CR 702.11b (which only defines hexproof itself). Beneficiary resolution is **CR 109.5**.

## Tests

- Parser: `you control` ⇒ `Some(You)`; unqualified ⇒ `None`.
- Multiplayer enforcement (`targeting.rs`): controller **can** target the affected creature; a third player **cannot** (revert-probe) — alongside the existing Nowhere to Run "any player" test.

## Verification

`cargo fmt --all` ✓ · full engine lib suite **15,775 pass / 0 fail** ✓ · `clippy -p engine -p phase-ai --all-targets -D warnings` clean ✓


matthewevans (changes requested):
**[MED]** `you control` is consumed but not represented, so Glaring Spotlight gets a broader hexproof bypass than printed. Evidence: `crates/engine/src/parser/oracle_static/shared.rs:35` says the qualifier does not change the emitted `IgnoreHexproof`, and `crates/engine/src/parser/oracle_static/tests.rs:524` only asserts the parser emits the unqualified mode. The existing runtime contract for object-scoped `IgnoreHexproof` is source-controller-independent (see `targeting.rs`'s Nowhere to Run multiplayer coverage), which is correct for unqualified “spells and abilities” but wrong for “spells and abilities you control.” Why it matters: this would make opponents’ and third players’ spells/abilities able to target the affected hexproof creatures, falsely marking Glaring Spotlight supported with stronger behavior than Oracle grants. Suggested fix: model the source-controller axis on the bypass authority, or leave the `you control` form unsupported until the runtime can enforce it.

galuis116:
Thanks — both points addressed in the latest push.

**CR citation (MEDIUM):** corrected to **CR 702.11e / CR 609.4** (the *"as though ... no hexproof"* targeting-permission rules) in the code, tests, and doc comments; CR 702.11b remains only where it's cited as the definition of hexproof itself.

**Multiplayer fidelity (HIGH):** the `you control` qualifier is no longer discarded. It now drives a new `StaticDefinition::bypass_beneficiary: Option<ControllerRef>` field (mirroring the existing mode-specific `attack_defended` / `source_controller` fields):

- `None` — every player (Nowhere to Run)
- `Some(ControllerRef::You)` — the static controller only (Glaring Spotlight)

`target_ignores_hexproof` now takes the targeting source's controller and honors the beneficiary (CR 109.5), so in a 3-player game Glaring Spotlight lets **its controller** bypass hexproof while a third player stays blocked. Added a multiplayer enforcement test (with a revert-probe) plus the parser-level `Some(You)` / `None` assertions.

matthewevans (approved):
Maintainer review: current head represents the Glaring Spotlight beneficiary qualifier and covers the multiplayer targeting split; auto-merge will wait for pending checks.

Issue #5246, opened 2026-07-07
Panther Habit preventing any or all all damage on the board — [[Panther Habit]] is stopping any damage at all from happ…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1522455069411512320/1522455069411512320

**Thread:** Panther Habit preventing any or all all damage on the board
**Discord thread id:** `1522455069411512320`
**Discord message id:** `1522455069411512320`
**Cards:** panther habit
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
[[Panther Habit]] is stopping any damage at all from happening, instead of just on the equipped creature.

## Reported text

[[Panther Habit]] is stopping any damage at all from happening, instead of just on the equipped creature.

## Attachments

- [game-state-turn-29-2026-07-03T04-10-22-125Z.zip](https://cdn.discordapp.com/attachments/1522455069411512320/1522455069583347842/game-state-turn-29-2026-07-03T04-10-22-125Z.zip?ex=6a4dce46&is=6a4c7cc6&hm=009bdffd280f85d3bb5fc04c9b9111f71f9173e5fc780efdb43625b106279280&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

### panther habit
If equipped creature would be dealt damage, prevent that damage and put that many +1/+1 counters on it.
Equip {2}

---
<sub>report_id: `discord:1522455069411512320:1522455069411512320:0` · discord: `1522455069411512320/1522455069411512320`</sub>

mike-theDude:
Verified Panther Habit's Oracle text ("If equipped creature would be dealt damage, prevent that damage and put that many +1/+1 counters on it.") and traced the runtime replacement-effect resolution in `crates/engine/src/game/replacement.rs`. The report matches a real bug: the damage-prevention shield's `valid_card` filter wasn't scoped to the equipped creature (`EquippedBy`), so it read as a board-wide prevention effect instead of a per-object one. This has already been fixed by commit `3a89097a0` ("fix(parser): scope attached-creature damage-prevention shields to the equipped/enchanted creature (#5246) (#5390)", merged 2026-07-08), which is present on `main`/HEAD and reflected in the regenerated `client/public/card-data.json` (the replacement's `valid_card` is now `Typed(Creature, properties: [EquippedBy])`). No further action needed beyond confirming release inclusion.

[...]

PR #5390, opened 2026-07-08
fix(parser): scope attached-creature damage-prevention shields to the equipped/enchanted creature (#5246)

davion-knight:
## Summary

Fixes **#5246** — *Panther Habit preventing any or all damage on the board*.

**Panther Habit** — `If equipped creature would be dealt damage, prevent that damage and put that many +1/+1 counters on it.` (Oracle text verified against `client/public/card-data.json`) — mis-handled its damage-prevention replacement in two coupled ways:

- **BUG 1 — shield scope.** The shield didn't scope to the equipped creature, so it read as a *board-wide* prevention instead of a per-equipped-creature one (the reported "prevents any/all damage on the board"). Fixed by adding a subject-first recipient combinator and chaining it into the recipient scan, yielding `valid_card = Typed(Creature, EquippedBy)` (and `EnchantedBy` for the Aura axis):
  ```rust
  terminated(parse_attached_subject_target_filter, tag(" would be dealt"))
  ```
- **BUG 2 — counter rider.** `put that many +1/+1 counters on it` lowers the bare "it" to `SelfRef`, but it means the damage **recipient** (the attached creature), so it must rewrite to `PostReplacementDamageTarget`.

## Cohort correctness

The two recipient cohorts demand opposite anaphor bindings, so the follow-up rewrite **branches** on them rather than layering:

| Cohort | Example | Rider anaphor resolves to |
|---|---|---|
| Object / event recipient | Panther Habit (`EquippedBy`), Vigor | `PostReplacementDamageTarget` |
| Self-scoped | Anti-Venom, Horrifying Healer (`…on him.`), Unbreathing Horde | `SelfRef` (source-referring) |

A self-scoped shield's dangling `ParentTarget` rider (a gendered pronoun like "him"/"her" lowers to the generic CR 608.2c anaphor, not the `subject: SelfRef` thread) now resolves to `SelfRef`, so the object-recipient rewrite never over-reaches into it.

## Architecture

- Nom combinators throughout (`terminated`, `tag`, `scan_at_word_boundaries`); reuses the shared `each_target_filter_mut` walker.
- Builds for the **class** (equipped **or** enchanted; object **or** self cohort), not the single card.
- **No new enum variants** — reuses `TargetFilter::{SelfRef, ParentTarget, PostReplacementDamageTarget}`.
- CR-annotated: **CR 615.1a / CR 615.5 / CR 608.2k / CR 301.5f / CR 303.4b**.

## Testing

`cargo fmt` ✓ · parser-combinator gate ✓ · `cargo check -p engine` ✓ · engine tests ✓ (0 failed).

New tests: `panther_habit_equipped_recipient_scope_and_counter_target_rewrite`, `enchanted_creature_subject_first_prevention_counter_target_rewrite`, `anti_venom_self_scoped_rider_stays_self_ref` (self-scope over-reach guard). Pre-existing `vigor_event_recipient_*` and `unbreathing_horde_*` still green.

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -8187,6 +8187,19 @@ fn strip_as_long_as_prefix_for_prevention(
     }
 }
 
+/// CR 615.1a: subject-first prevention recipient "<subject> would be dealt …".
+/// The attached creature (CR 301.5f "equipped creature" / CR 303.4b "enchanted
+/// creature") is the shield's recipient for the Panther-Habit class ("If
+/// equipped creature would be dealt damage, prevent that damage …"). Composes
+/// `parse_attached_subject_target_filter` (whose core does not consume trailing
+/// whitespace) with the leading-space `tag(" would be dealt")`, so it matches
+/// both "would be dealt damage" and "…would be dealt combat damage". The
+/// self-subject forms ("~" / "this creature") are handled by the earlier
+/// if-branch (they map to `SelfRef`), so no self arm belongs here.
+fn parse_subject_first_prevention_recipient(input: &str) -> OracleResult<'_, TargetFilter> {
+    terminated(parse_attached_subject_target_filter, tag(" would be dealt")).parse(input)
+}
+
 /// CR 615: Parse damage prevention replacement effects.
 /// Handles:
 /// - "prevent all combat damage that would be dealt [this turn]" (Fog, Moments Peace)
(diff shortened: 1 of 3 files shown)


matthewevans (changes requested):
**[MED]** The Panther Habit fix is only covered at parser-shape level, not through the runtime replacement pipeline it claims to repair. Evidence: `crates/engine/src/parser/oracle_replacement.rs:10603` adds parser unit coverage that asserts `valid_card` and `PostReplacementDamageTarget`, and the current diff adds no integration/scenario test for Panther Habit or an equipped/enchanted damage-prevention shield. Why it matters: the reported bug is runtime behavior — a shield preventing unrelated damage and placing counters on the wrong object — so this can still false-green if parser output is correct but replacement matching, attachment scoping, or post-replacement target resolution does not produce the promised in-game result. Suggested fix: add a discriminating engine scenario that attaches Panther Habit to one creature, deals damage to that creature and to an unrelated creature, and asserts only the attached creature damage is prevented and receives counters equal to the prevented amount.

davion-knight:
Thanks @matthewevans — good catch. Added a discriminating **runtime** scenario in `crates/engine/tests/integration/panther_habit_equipped_prevention_scope.rs` that drives the full replacement pipeline (commit b6fc615fd):

- Panther Habit is attached to one creature (`EquippedBy` resolved via the real `attached_to` signal, CR 301.5f — no parser shortcut).
- Damage is dealt to the **equipped** creature *and* to an **unrelated** creature.
- Asserts: unrelated creature takes the damage (not prevented, no counters, no prevention event); equipped creature's damage **is** prevented and it gains +1/+1 counters **equal to the prevented amount**.

So a correct parse with broken replacement matching / attachment scoping / post-replacement target resolution would fail this test, not false-green. `cargo test -p engine --test integration panther_habit_prevents_only_equipped` passes.

matthewevans (approved):
Maintainer review: current head adds the missing runtime replacement coverage for Panther Habit. I rechecked the parse-diff blast radius (Panther Habit plus self-scoped Gloom Surgeon / Anti-Venom corrections), CR citations, and GitHub checks; this is ready for the merge queue.

PR #5393, opened 2026-07-08
fix(parser): support "attacking the last chosen player" statics (Triarch Stalker)

### LINKED ISSUE #5391: Parser: "Creatures attacking the last chosen player" static is unsupported (Triarch Stalker, Beckoning Will-o'-Wisp)
## Summary

Cards that pair a combat trigger *"choose an opponent"* with a static that references *"the last chosen player"* do not function: the static line fails to parse, and even once parsed the chosen player is not persisted for the static to read.

## Affected cards

- **Triarch Stalker** — *"Targeting Relay — At the beginning of combat on your turn, choose an opponent. Creatures attacking the last chosen player have menace."*
- **Beckoning Will-o'-Wisp** — *"Lure the Unwary — At the beginning of combat on your turn, choose an opponent. Creatures attacking the last chosen player get +1/+0."*

## Current behavior

- The trigger `At the beginning of combat on your turn, choose an opponent` parses to `Effect::Choose { choice_type: Opponent, persist: false }`.
- The static line `Creatures attacking the last chosen player ...` fails the static parser (`static_structure` unimplemented) because `parse_attacking_defender_scope` does not recognize the *"the last chosen player"* defender scope.
- Because the choice is not persisted (`persist: false`), even a correctly-parsed `ControllerRef::SourceChosenPlayer` static would read nothing at runtime.

## Expected behavior

- The static parses to a continuous `FilterProp::Attacking { defender: Some(ControllerRef::SourceChosenPlayer) }` with the menace / +1/+0 modification.
- The combat trigger's opponent choice persists durably (`ChosenAttribute::Player`) so the static resolves it via the existing `source_chosen_player` / `attacking_defender_matches` path.

## Rules

CR 508.1b (attacking-player scope), CR 613.1 / CR 608.2c (a persisted "chosen player" read by a continuous ability).

This is the attacker-scope counterpart to the you / your opponents / enchanted-player scopes already handled by `parse_attacking_defender_scope`.

joaovictor712:
Closes #5391.

## What

Makes **Triarch Stalker** and **Beckoning Will-o'-Wisp** work. Both pair a combat trigger *"At the beginning of combat on your turn, choose an opponent"* with a continuous static *"Creatures attacking the last chosen player have menace / get +1/+0"*. Neither functioned: the static line failed to parse, and the opponent choice was not persisted for a continuous ability to read.

## How (routes to existing runtime, no engine changes)

**1. Static scope** — `parse_attacking_defender_scope` gains `"the last chosen player"` / `"the chosen player"` → `ControllerRef::SourceChosenPlayer`, alongside the existing you / your-opponents / enchanted-player scopes (same `alt`, one new axis value). The static parses to `FilterProp::Attacking { defender: Some(SourceChosenPlayer) }` + the modification, resolved by the existing `attacking_defender_matches` → `controller_ref_player` → `source_chosen_player` path.

**2. Durable persist** — new `reconcile_persisted_player_choice_for_source_chosen_ref` post-pass flips a `choose a player` / `choose an opponent` effect to `persist: true` **only when** the card carries a durable `SourceChosenPlayer` reference, so the choice is stored as `ChosenAttribute::Player`. This mirrors the "persist when referred back to" intent documented on `Effect::Choose`, and the runtime already re-runs layers for persisted `Player`/`Opponent` choices.

Targeted: a *same-resolution* opponent choice (Ruhan — reads the resolution-scoped `ChosenPlayer { index }`) has no durable reference and is left `persist: false`.

## Tests

- Static scope parsing (both phrasings) → `Attacking { defender: SourceChosenPlayer }` + menace / +1/+0.
- Full-card persist flip (Triarch Stalker) → trigger `Choose { Opponent, persist: true }`.
- Negative regression (Ruhan) → `persist: false` (pass stays targeted).
- Runtime `attacking_defender_matches` resolves against a persisted chosen player (matches the chosen player, rejects others).

## CR

CR 508.1b (attacking-player scope) + CR 613.1 / CR 608.2c (persisted "chosen player" read by a continuous ability).

## Verification

`cargo fmt --all` ✓ · engine lib suite **15,812 pass / 0 fail** ✓ · `clippy -p engine --all-targets -D warnings` clean ✓

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -10200,6 +10200,66 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn attacking_defender_matches_source_chosen_player() {
+        // CR 508.1b + CR 613.1: Triarch Stalker / Beckoning Will-o'-Wisp scope
+        // attackers by the source's persisted chosen player. `attacking_defender_
+        // matches` resolves `Some(SourceChosenPlayer)` through the generic
+        // `controller_ref_player` arm, which reads the choice durably persisted on
+        // the source object (`ChosenAttribute::Player`). An attacker attacking the
+        // chosen player matches; one attacking any other player does not.
+        use crate::game::zones::create_object;
+        use crate::types::ability::ChosenAttribute;
+
+        let mut state = GameState::new_two_player(7);
+        let card_id = CardId(state.next_object_id);
+        let src = create_object(
+            &mut state,
+            card_id,
+            PlayerId(0),
+            "Triarch Stalker".to_string(),
+            crate::types::zones::Zone::Battlefield,
+        );
+        // The combat trigger persisted P1 as the chosen player.
+        state
+            .objects
+            .get_mut(&src)
+            .unwrap()
+            .chosen_attributes
+            .push(ChosenAttribute::Player(PlayerId(1)));
+
+        let source_ctx = SourceContext {
+            id: src,
+            controller: Some(PlayerId(0)),
+            attached_to: None,
+            source_is_aura: false,
+            source_is_equipment: false,
+            chosen_creature_type: None,
+            chosen_attributes: &[],
+            ability: None,
+            recipient_id: None,
+        };
+
+        assert!(
+            attacking_defender_matches(
+                &state,
+                &source_ctx,
+                PlayerId(1),
+                Some(&ControllerRef::SourceChosenPlayer),
+            ),
+            "an attacker attacking the persisted chosen player must match"
+        );
+        assert!(
+            !attacking_defender_matches(
+                &state,
+                &source_ctx,
+                PlayerId(0),
+                Some(&ControllerRef::SourceChosenPlayer),
+            ),
+            "an attacker attacking a non-chosen player must not match"
+        );
+    }
+
     // ===========================================================================
     // CR 702.73a — Changeling subtype expansion cascade.
     //
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -1300,7 +1300,11 @@ fn reconcile_host_bound_phase_outs_in_ability(def: &mut AbilityDefinition) {
 /// the resolution-scoped `ChosenPlayer { index }`). This pass flips `persist` to
 /// `true` on those choices ONLY when the card also carries a durable
 /// `SourceChosenPlayer` reference — mirroring the "persist when referred back to"
-/// intent documented on `Effect::Choose`.
+/// intent documented on `Effect::Choose`. The reference can live in a static
+/// (`Creatures attacking the last chosen player ...`), an activated ability, or a
+/// triggered ability — a phase trigger scoped to "the chosen player's" step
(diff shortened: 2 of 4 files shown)


matthewevans (changes requested):
**[MED]** SourceChosenPlayer references inside triggers do not make the preceding player choice persistent. Evidence: `crates/engine/src/parser/oracle.rs:1305` only scans `result.statics` and `result.abilities` before flipping trigger choices to `persist: true`; `result.triggers` is only visited later as a writer at `crates/engine/src/parser/oracle.rs:1319`, and existing trigger parsing can produce `TargetFilter::SourceChosenPlayer` (`crates/engine/src/parser/oracle_trigger.rs:11740`, covered by `crates/engine/src/parser/oracle_trigger_tests.rs:13285`). Why it matters: a card that chooses a player and later has a triggered ability keyed to “the chosen player” would still lose the chosen player after the choose instruction resolves. Suggested fix: include trigger execute abilities in the SourceChosenPlayer-reference scan, with a regression that fails before the trigger scan is added.

**[MED]** The new effect detector reimplements target discovery for only `DealDamage` instead of using the existing effect target API. Evidence: `crates/engine/src/parser/oracle.rs:1345` only checks `Effect::DealDamage { target, .. }`, while `Effect::target_filter()` already surfaces many target-bearing player/object effects such as Draw, GainLife, LoseLife, Token owner, SearchLibrary, etc. at `crates/engine/src/types/ability.rs:13042`. Why it matters: the persistence pass now silently misses any non-damage ability that reads `TargetFilter::SourceChosenPlayer`, so the helper is a one-effect special case despite being wired as the general source-chosen-player detector. Suggested fix: drive `effect_targets_source_chosen_player` through `effect.target_filter().is_some_and(filter_references_source_chosen_player)` and add at least one non-damage sibling regression.

galuis116:
Thanks @matthewevans — good catch, both points addressed in the latest push.

**[HIGH/MED] Triggers not scanned as readers:** `reconcile_persisted_player_choice_for_source_chosen_ref` now also scans `result.triggers` via a new `trigger_references_source_chosen_player`, which checks **both** the trigger's own `valid_target` (the `oracle_trigger.rs` phase-trigger scope, e.g. "the chosen player's upkeep") **and** its executed effect chain. So a card whose only chosen-player reference is a triggered ability now correctly persists the preceding choice.

**[MEDIUM] Single-case effect coverage:** `effect_targets_source_chosen_player` now uses the generic `Effect::target_filter()` accessor instead of matching only `Effect::DealDamage`, so every player-targeting effect variant is covered.

**Regression** (fails before the trigger-reader scan): `choose_persists_when_only_a_trigger_reads_the_chosen_player` — a card whose chosen-player reference lives only in a phase trigger flips the ETB `choose an opponent` to `persist: true`. Verified the baseline (bare "When enters, choose an opponent" with no reference) stays `persist: false`.

Full engine lib suite (15,820) + `clippy --all-targets` green.

matthewevans (approved):
Maintainer review: current head resolves the trigger-reader and generic target-filter findings; parse diff is scoped to the chosen-player attacker statics, and required checks are green.

PR #5394, opened 2026-07-08
Fix Winter Orb / Static Orb

rykerwilliams:
## Summary
Fixes the "Granted/quoted ability or continuous modification dropped" misparse for **Winter Orb**, and the same underlying bug (independently confirmed, tracked in the backlog under a different, less-precise root-cause label) for **Static Orb**. Both currently parse to a no-op `StaticMode::Continuous` with `modifications: []` — the leading "As long as this artifact is untapped, ..." condition breaks `MaxUntapPerType`'s dispatch branch, so the untap-cap effect is silently dropped and both cards currently do nothing in-game.

Root cause: `MaxUntapPerType`'s branch required its suffix parser to hit end-of-string. The general inverted-"as long as" rewrite machinery recurses into it with trailing text after the match point, so the branch never fires and a generic fallback swallows the effect. Fix gives `MaxUntapPerType` the same trailing-condition tolerance `parse_continuous_gets_has` already has, reusing existing building blocks — zero runtime changes needed, the condition-gating machinery already worked generically once the parser emits the right mode+condition.

Backlog entries for both cards removed per §3.1 list hygiene (second commit).

## Files changed
- crates/engine/src/parser/oracle_static/anthem.rs
- docs/parser-misparse-backlog.md

## CR references
- CR 502.3 — the untap-cap restriction itself
- CR 604.1 — static abilities are simply true whenever their condition holds
- CR 611.3a — a continuous effect from a static ability isn't "locked in"; it applies whenever its text currently indicates (authorizing rule for gating the effect on live tapped-state)

## Track
Developer

## LLM
Model: claude-sonnet-5
Thinking: high

## Verification
- `./scripts/check-parser-combinators.sh` — clean (no output, exit 0)
- `cargo clippy --all-targets -- -D warnings` — clean, full workspace
- `cargo test -p engine --lib` — 15758 passed, 0 failed, 6 ignored
- New parser tests (`static_max_untap_winter_orb_conditional`, `static_max_untap_static_orb_conditional`) assert both cards' verbatim Oracle text parse to `MaxUntapPerType` with the correct filter/max and a populated `Not(SourceIsTapped)` condition
- New runtime tests (`winter_orb_max_untap_cap_gated_by_own_tapped_state`, `two_winter_orbs_gate_independently_on_own_tapped_state`) drive the fix through the real `active_static_definitions`/`evaluate_condition`/`max_untap_restrictions` pipeline, including a multi-authority fixture proving two Winter Orbs gate independently on their own tapped state
- `cargo coverage` — not run locally (MTGJSON/card-data.json not present in this environment); the primary coverage gate's "supported" boolean for these cards is unaffected by this fix either way (see judgment note below); `cargo semantic-audit` likewise not run locally for the same reason

## Gate A
```
(no output — clean, exit 0)
```

## Anchored on
- `crates/engine/src/parser/oracle_static/anthem.rs:782` — `parse_continuous_gets_has`'s own `" as long as "` split-recurse-reattach shape, the established precedent this fix's new `parse_max_untap_per_type_static` (dispatch.rs:170) mirrors
- `crates/engine/src/parser/oracle_static/anthem.rs:729` — `rewrite_self_pronoun_subject`, reused (visibility bumped to `pub(crate)`) rather than re-implemented, for self-pronoun condition normalization

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.


matthewevans:
Holding current head `51a59db5f6a92647bd924f288925498d6e142ec9` until the parser evidence lands. This PR changes engine/parser code, but the `coverage-parse-diff` sticky is absent and the Rust/card-data checks are still pending for this head. I’ll re-check once the current-head card-level diff is available.

matthewevans (approved):
Maintainer review: current head parses Winter Orb / Static Orb into conditional MaxUntapPerType, verifies own-source tapped-state gating at runtime, and the parse diff is limited to those two cards.

PR #5397, opened 2026-07-08
fix(engine): declined "up to N target" sub-effect no longer inherits parent's target (Cruel Revival) (#5281)

### LINKED ISSUE #5281: Cruel Revival returns creature to hand instead of destroying. — [[Cruel Revival]]
Reported in Discord: https://discord.com/channels/1485498006781427802/1523764877401915563/1523764877401915563

**Thread:** Cruel Revival returns creature to hand instead of destroying.
**Discord thread id:** `1523764877401915563`
**Discord message id:** `1523764877401915563`
**Cards:** cruel revival
**Extraction confidence:** 0.40

## Summary
[[Cruel Revival]]

### cruel revival
Destroy target non-Zombie creature. It can't be regenerated. Return up to one target Zombie card from your graveyard to your hand.

---
<sub>report_id: `discord:1523764877401915563:1523764877401915563:0` · discord: `1523764877401915563/1523764877401915563`</sub>

philluiz2323:
## Summary

[[Cruel Revival]] — "Destroy target non-Zombie creature. It can't be regenerated. Return up to one target Zombie card from your graveyard to your hand." — **returned the destroyed creature to hand instead of destroying it** whenever the caster declined the optional "up to one" bounce.

Fixes #5281

## Root cause

The spell parses correctly to `Destroy(non-Zombie creature)` → `Bounce(Zombie card from graveyard, up to one)`. The bug is at resolution: in `resolve_ability_chain`, a sub-effect with empty `targets` inherited the parent instruction's targets wholesale (`sub_with_targets.targets = ability.targets.clone()`).

The bounce has its own independent target slot ("target Zombie card from your graveyard"). When that optional target is legally **declined** (CR 115.6 — a "up to N target" effect is targeted only if ≥1 target is chosen), its `targets` are empty — so it inherited the parent `Destroy`'s battlefield-creature target and bounced it.

## Fix

When inheriting parent targets to a chained sub-effect:

- **Player** targets are still inherited unconditionally — they are shared across the chain (Paradigm's copied "that player" draw + lose-life; a relative-controller change-zone effect).
- An inherited **Object** target is kept only when it is **legal for the sub-effect's own independent filter**. Cruel Revival's destroy target (a battlefield creature) is illegal for the bounce's "Zombie card in your graveyard" filter, so it is dropped and the declined bounce resolves as a no-op (CR 608.2c).
- Anaphoric sub-effects — no own filter ("It can't be regenerated") or a `ParentTarget`/relative-controller ref — keep inheriting unchanged.

This is a single, guarded change in the chain-resolution target-propagation; it reuses `targeting::validate_targets_for_ability` for the legality check. No new engine variant.

> Surgical, well-scoped fix to one target-propagation site in `resolve_ability_chain`, guided by a reproduced runtime failure and verified against the full engine lib suite (the two regressions the first attempt exposed — Paradigm copy chain and relative-controller change-zone — are exactly the shared-Player-target cases the final guard preserves). No new engine variant, no parser change.

## CR references

- CR 115.6 — a spell/ability that allows "up to N" targets is targeted only if one or more targets have been chosen (the declined optional bounce)
- CR 608.2c — later text modifies earlier ("Destroy target creature. It can't be regenerated") — the destroy and the return are independent instructions with independent target slots

## Verification

- New runtime cast tests (`crates/engine/tests/cruel_revival_destroy_and_return_5281.rs`):
  - `cruel_revival_destroys_creature_and_returns_graveyard_zombie` — both targets: creature destroyed (→ graveyard), graveyard Zombie returned (→ hand).
  - `cruel_revival_destroys_creature_when_bounce_declined` — the reported bug: bounce declined, creature still destroyed (not bounced).
- Full `cargo test -p engine --lib` — green (no regressions; the Paradigm and change-zone chains that a broader first attempt broke both pass).
- `cargo fmt --all`, `cargo clippy -p engine` — clean.


matthewevans (changes requested):
**[HIGH]** Independent optional sub-effects can still steal a declined parent target when the parent target happens to satisfy the sub-effect filter. Evidence: `crates/engine/src/game/effects/mod.rs:7815` enters the inheritance branch whenever `sub.targets.is_empty()`, and lines `7830-7847` copy any parent object target that validates against `own_object_filter`; a spell like “Exile target Zombie card from your graveyard. Return up to one target Zombie card from your graveyard to your hand.” would still return the exiled Zombie when the second slot is declined because the first target passes the second slot’s Zombie-card filter. Why it matters: CR 115.6 optional targets declined at cast time must remain absent at resolution; filtering inherited targets only fixes Cruel Revival because its destroy target is illegal for the return filter, but it does not fix the target-slot class. Suggested fix: when a sub-effect has its own independent target filter and no explicit targets, do not inherit object targets at all; only inherit object targets for parent-target effects, and add a same-filter declined-slot regression.

philluiz2323:
Good catch — tightened per the [HIGH] finding. Instead of filtering inherited object targets by legality (a proxy that still lets a sub steal a *legal* parent object), a sub-effect with its **own independent target filter** now never inherits a parent OBJECT target at all — its empty `targets` mean the optional target was declined (CR 115.6), so it resolves as a no-op. Player targets stay shared across the chain and anaphoric subs (no filter / ParentTarget ref) still inherit. Verified: full `engine` lib suite green (15,792 passed / 0 failed), including the Cruel Revival, Paradigm-copy, and relative-controller change-zone cases.

matthewevans (approved):
Maintainer review: current head resolves the independent optional sub-effect target slot at the chain target-propagation seam, with runtime coverage for both chosen and declined Cruel Revival return targets. Auto-merge can wait for pending checks.

matthewevans (approved):
Maintainer fixup: preserved reflexive chain targets while keeping independent optional object target slots from inheriting declined parent targets. Auto-merge can wait for the new CI run.

PR #5409, opened 2026-07-09
feat(parser): leading except-for exempt-list restriction static — Akron Legionnaire (CR 508.1c)

### LINKED ISSUE #5408: feat(parser): leading except-for exempt-list restriction static — Akron Legionnaire (CR 508.1c)
## Summary

`parse_compound_subject_rule_static` (`crates/engine/src/parser/oracle_static/evasion.rs:323`) is the single dispatcher for restriction-static lines of the form `"<subject> can't attack[/block/…]"`. Its subject extraction (`nom_primitives::scan_preceded(lower, parse_rule_static_predicate_nom)`) takes *everything before the first recognized predicate* as the subject and hands it whole to `parse_rule_static_subject_filter` (`shared.rs:3913`). Neither function has any grammar for a sentence-initial exception clause.

**Akron Legionnaire** (*Masters Edition III*, legal in Legacy/Vintage/Commander/Oathbreaker/Duel/Premodern/PreDH/Penny — verified via Scryfall `oracle_id: c5f50a78-9dc4-4f3c-aa11-b41c8278bc16`):

```
Except for creatures named Akron Legionnaire and artifact creatures, creatures you control can't attack.
```

Because the line opens with `"Except for "`, `parse_rule_static_subject_filter` is handed `"except for creatures named akron legionnaire and artifact creatures, creatures you control "` as a single subject phrase, which it has no grammar for, so the whole line strict-fails. `client/public/card-data.json` currently records:

```
"type":"Unimplemented","name":"static_structure",
"description":"Static pattern matched but line failed static parser: Except for creatures named Akron Legionnaire and artifact creatures, creatures you control can't attack."
```

So Akron Legionnaire's entire printed ability — a board-defining attack restriction with two carved-out exceptions — is silently dropped.

## Why this is a class, not a card

There IS an existing "except for" combinator, but it's the wrong grammatical slot: `parse_except_for_type_list_suffix` (`crates/engine/src/parser/oracle_target.rs:6228`) only handles a **trailing** `"<type> except for <type-list>"` suffix (Scourglass: "Destroy all permanents except for artifacts and lands"), and its own doc comment explicitly declines named/designation exceptions as unsafe in that slot (an unrecognized word falls back to a silently-vacuous negated `Subtype`, so it must reject the whole clause rather than risk a no-op exclusion).

Akron Legionnaire's shape is different: `"Except for <A> and <B>, "` is a **sentence-initial** clause that scopes an entire restriction predicate, not a suffix on one type filter. This is the same shape of generalization as #5219 (compound-subject animation statics) and #5384 (Shalai's N-way keyword-grant compound subject): resolve each conjunct of a list independently through existing per-domain resolvers, then recombine generically — here, ANDing `Not{Or{exempt conjuncts}}` onto the `affected` filter the existing single-subject dispatcher already produces for the remainder, rather than hand-rolling an Akron-Legionnaire-specific branch.

Each exempt conjunct is itself a class-general shape already used elsewhere in the engine:
- `"artifact creatures"` — a bare type-phrase, handled by the existing `parse_type_phrase`.
- `"creatures named Akron Legionnaire"` — a type-class-scoped name exemption, using the existing `FilterProp::Named` (already used for "cards named X" filters elsewhere, e.g. `oracle_effect/search.rs:829`, `game/restrictions.rs:3822`), unlike the "except for" suffix combinator, which must *decline* named exceptions because a silent no-op there would be dangerous — this sentence-initial slot ANDs the exempt set as an explicit negation, so a named conjunct is safe to support directly.

## Proposed fix

- Add `parse_leading_except_for_rule_static(text, lower)` in `oracle_static/evasion.rs`, immediately alongside its single-subject sibling `parse_compound_subject_rule_static`, dispatched in `shared.rs` immediately **before** that sibling (its subject grammar has no leading-clause syntax and would otherwise strict-fail on the "except for" prefix).
- Split the exempt list on `" and "` into two conjuncts (`nom_primitives::split_once_on`, mirroring #5219's `parse_compound_all_subjects_filter` loop shape), resolve each independently as either a bare type-phrase exemption or a `"<type> named <Name>"` exemption, Or-combine, then AND `Not{Or{..}}` onto the `affected` filter of every `StaticDefinition` the sibling dispatcher produces for the remainder after the exempt clause.
- Scope to the 2-conjunct `"<A> and <B>"` form the only printed card in this shape uses: a Scryfall full-text search (`o:/^Except for/`) returns exactly one result (Akron Legionnaire), so an unbounded Oxford-comma exempt list would be speculative — there's no card to validate a 3+-conjunct split against the genuine ambiguity of which comma ends the exempt list vs. starts the next conjunct. Widen once a 3+-conjunct card is identified (mirroring how Shalai's own N-way generalization was built only once a 3-item printed card existed).

## CR references

- `CR 508.1c` — the active player checks each creature for restrictions ("effects that say a creature can't attack") when declaring attackers.
- `CR 201.2a` — two or more objects have the same name if they share at least one name; the basis for the `"creatures named <Name>"` exemption.

## Verification

- `client/public/card-data.json` → `"akron legionnaire"` → `abilities[0].effect.type == "Unimplemented"` at time of filing (repo tip `39a332d26`).
- Confirmed via Scryfall full-text search (`o:/^Except for/`) that Akron Legionnaire is the only currently-printed card whose Oracle text opens with a leading "Except for" exemption clause — the fix is scoped to exactly the evidence-backed 2-conjunct shape, not a speculative unbounded grammar.

jsdevninja:
## Summary
Adds engine support for **Akron Legionnaire**'s leading exempt-list restriction static — and the *class* of sentence-initial "Except for `<A>` and `<B>`, `<rule-static>`" exception clauses it represents.

Its printed static, `Except for creatures named Akron Legionnaire and artifact creatures, creatures you control can't attack.`, silently dropped: `parse_compound_subject_rule_static` (`crates/engine/src/parser/oracle_static/evasion.rs`) hands the FULL prefix before the first recognized predicate to `parse_rule_static_subject_filter`, which has no "except for" arm, so the whole line strict-failed.

`parse_leading_except_for_rule_static` (dispatched immediately before `parse_compound_subject_rule_static`) splits the exempt list into two conjuncts on `" and "`, resolves each independently — a bare type-phrase exemption ("artifact creatures") or a named exemption within a type class ("creatures named Akron Legionnaire", via `FilterProp::Named`) — Or-combines them, then ANDs `Not{Or{..}}` onto the `affected` filter the existing single-subject dispatcher produces for the remainder. No new runtime: `TargetFilter::And`/`Not`/`Or` and `FilterProp::Named` are all existing, already-evaluated primitives.

## Files changed
- `crates/engine/src/parser/oracle_static/evasion.rs`
- `crates/engine/src/parser/oracle_static/mod.rs`

## Anchored on
- `crates/engine/src/parser/oracle_static/evasion.rs` `parse_compound_subject_rule_static` — the single-subject sibling; the new handler is dispatched immediately before it and delegates the remainder to it unchanged.
- `#5219` (compound-subject animation statics, Life and Limb) — the same "resolve conjuncts independently, recombine generically" shape, applied to a different static category (restriction, not type-change).
- `#5384` (N-way Oxford-comma compound-subject keyword static, Shalai) — the precedent for evidence-scoping arity: build exactly the conjunct-count the printed card uses, not a speculative unbounded grammar.

## CR references
- `CR 508.1c` — the active player checks each creature for restrictions ("effects that say a creature can't attack") when declaring attackers.
- `CR 201.2a` — two or more objects have the same name if they share at least one name; the basis for the `"creatures named <Name>"` exemption conjunct.

## Scope
Scoped to the 2-conjunct `"<A> and <B>"` form: a Scryfall full-text search (`o:/^Except for/`) returns exactly one printed card in this shape (Akron Legionnaire), so an unbounded Oxford-comma exempt list would be speculative — there's no card to validate a 3+-conjunct split against the genuine ambiguity of which comma ends the exempt list vs. starts the next conjunct.

## Verification
- `cargo fmt --all` — clean.
- Full `cargo check`/`test`/`clippy` could not be run in this session — local machine's MSVC C++ Build Tools installation is incomplete (`vswhere` finds no installation with the VC.Tools component), so `link.exe` resolves to Git's coreutils shim instead of the real MSVC linker, and every crate (including unrelated third-party dependencies) fails to link before reaching this crate's code. Tilt was also not running in this session. Verified by close manual trace instead: every referenced type, function signature, and visibility (`TargetFilter::{And,Not,Or,Typed,Any}`, `FilterProp::Named`, `TypedFilter`, `StaticDefinition.{affected,description}`, `nom_primitives::split_once_on`, `parse_type_phrase`, `merge_filter_prop` [bumped `fn` → `pub(crate) fn`], `parse_compound_subject_rule_static`) was confirmed against the actual current source, and the dispatch chain (`parse_static_line_multi` → `parse_static_line_multi_ir` → `parse_static_line_multi_inner` → `parse_static_line_multi_dispatch`) was traced by hand to confirm the new arm is reachable and ordered correctly relative to its sibling. Given the local toolchain is broken independent of this change, deferring final compiler/test confirmation to CI.
- New tests (`oracle_static/tests.rs`): a positive parse test on Akron Legionnaire's real Oracle text asserting the full `And{Typed(creatures you control), Not{Or{Typed(creatures named "akron legionnaire"), Typed(artifact creatures)}}}` shape; a guard confirming the untouched sibling (`"Creatures you control can't attack."`, no exempt clause) still resolves unwrapped; a guard confirming a single-conjunct exempt clause (no `" and "`) correctly declines rather than mis-parsing.

## Track
Community

## LLM
Model: claude-sonnet-5

## Scope Expansion
None.

## Validation Failures
None — see Verification note on the local toolchain limitation.

## CI Failures
Unknown at PR open time — local compiler verification was unavailable (see Verification).

Closes #5408


matthewevans (changes requested):
**[MED]** Parser-only tests do not prove the new Akron Legionnaire static works through the production attack-declaration path. Evidence: `crates/engine/src/parser/oracle_static/tests.rs:4243`. Why it matters: the PR claims card support and builds a supported `CantAttack` static, but the added tests only assert AST shape; a regression in `CantAttack`/`affected` filter evaluation, `FilterProp::Named`, or `validate_attackers` would still pass. Suggested fix: add an integration/scenario test that parses Akron Legionnaire’s real Oracle text and declares attackers, proving a normal nonartifact creature you control is rejected while a creature named Akron Legionnaire and an artifact creature are legal attackers.

jsdevninja:
Addressed: the original tests only asserted the parsed AST shape (`oracle_static::tests.rs`), which doesn't prove the static gates real combat.

Added `combat::tests::akron_legionnaire_exempts_named_and_artifact_creatures_from_attacking` — attaches the *actual* parsed static (via `parse_static_line_multi`, on a battlefield Akron Legionnaire) and declares attackers: a plain nonartifact creature you control is rejected by `validate_attackers`/excluded from `get_valid_attacker_ids`, while Akron Legionnaire itself (exempted by name) and an artifact creature you control (exempted by type) remain legal attackers.

`cargo fmt --all` is clean. Full `cargo check`/`test` still couldn't run locally in this session (broken MSVC linker on this machine, unrelated to the change — see the PR description's Verification note); traced the new test's harness calls (`create_creature`, `static_definitions.push`, `get_valid_attacker_ids`, `validate_attackers`) against their real signatures in `combat.rs` by hand.

matthewevans (approved):
Maintainer fixup: current head avoids the helper-name collision, preserves the existing subject-rule parser, and keeps the runtime Akron combat coverage; auto-merge can wait for restarted checks.

matthewevans (approved):
Current head has the runtime Akron combat coverage, parse-diff matches the claimed single-card change, and the parser/combat seam is clean.

matthewevans (approved):
Maintainer review: current head keeps the Akron parser/combat seam clean, has discriminating runtime coverage, and parse-diff matches the claimed single-card restriction change.

Issue #5412, opened 2026-07-09
parser: compound-subject "don't untap" restriction strict-fails (Exhaustion, Icebreaker Kraken)


claytonlin1110:
## Summary

**Exhaustion** and **Icebreaker Kraken**'s compound-subject "don't untap" restriction —
`"Creatures and lands target opponent controls don't untap during their next untap step."`
and `"...artifacts and creatures target opponent controls don't untap during that player's next untap step."`
— currently strict-fail (verified against `data/coverage-data.json`: both `supported: false`,
`gap_count: 1`, with a bare partial-match label `"creatures"`/no label at all for the restriction
clause).

This is the same *structural* gap shape as #5219 (Life and Limb's compound-subject animation
static): a compound subject joined by "and" where each individual conjunct is already fully
supported on its own, but the combination is unhandled.

## Verified card text (Scryfall API)

- **Exhaustion** (sorcery): `Creatures and lands target opponent controls don't untap during their next untap step.`
- **Icebreaker Kraken** (ETB trigger): `When this creature enters, artifacts and creatures target opponent controls don't untap during that player's next untap step.`

## Proof the single-subject predicate is already fully correct

Two real, currently-`supported: true` cards confirm the single-subject sibling of this exact
predicate is complete end-to-end (verified against `data/coverage-data.json`):

- **Misstep**: `Creatures target player controls don't untap during that player's next untap step.`
- **Mana Vapors**: `Lands target player controls don't untap during their next untap step.`

Between them these are literally the two halves Exhaustion combines ("creatures ... don't
untap" + "lands ... don't untap" → "creatures and lands ... don't untap"). The predicate builder
(`build_restriction_clause` in `crates/engine/src/parser/oracle_effect/subject.rs:4350`) already
produces the correct `StaticMode::CantUntap` + `Duration::UntilNextStepOf { step: Phase::Untap,
player: PlayerScope::Controller }` (CR 502.3, see the `has_next_untap` handling at
`subject.rs:4397-4410`). So this is **not** a predicate gap (unlike #5219, which needed two
predicate-halves merged) — it is purely a **compound-subject** gap.

## Root cause (subject side)

The single-subject case reaches its filter through `parse_subject_application`'s bare
plural-noun-phrase fallback (`oracle_effect/subject.rs:2139-2165`), which prepends `"all "` and
delegates whole to `parse_target`/`parse_type_phrase_with_ctx`
[...]

PR #5414, opened 2026-07-09
test(parser): regression guard for compound-subject controller-suffix splitting (issue #5412)

claytonlin1110:
## What this PR actually is now

This started as a fix for #5412 (Exhaustion / Icebreaker Kraken's compound-subject "don't/doesn't untap" restriction, supposedly strict-failing). **That premise was wrong** — a reviewer correctly flagged that my original dispatcher was added at the wrong seam and produced no demonstrated production coverage impact, and I've since confirmed directly against this PR's true parse-diff baseline (`base-sha fced6f15c`, downloaded from `https://data.phase-rs.dev/parse-baselines/coverage-data-<hash>.json`) that both cards are **already** `supported: true, gap_count: 0` on `main`:

```
Exhaustion | supported: True | gap_count: 0
  label: CantUntap, affects: "target opponent controls creature or target opponent controls land"
Icebreaker Kraken | supported: True | gap_count: 0
```

**Root cause of my error**: I verified the "before" state against the git-tracked `data/coverage-data.json` in my working tree, which is a periodically-refreshed snapshot — not regenerated on every commit — and was stale relative to the actual parser capability of the exact commit I branched from. The general compound-core-type-Or splitter in `parse_type_phrase_with_ctx` (`TYPE_SEPARATORS` recursion + `distribute_controller_to_or`) already resolves a compound "`<X>` and `<Y>` target player/opponent controls" subject correctly, so the existing generic split in `try_parse_subject_restriction_clause` (unmodified) already handles both cards through one `parse_subject_application` call — no dedicated dispatcher needed.

I've reverted the redundant `try_parse_compound_target_controller_cant_untap` dispatcher entirely (`crates/engine/src/parser/oracle_effect/subject.rs` is now byte-identical to `main`).

## What's left in this PR

One building-block regression test in `crates/engine/src/parser/oracle_target.rs`:

```rust
fn compound_creatures_and_lands_target_opponent_controls()
```

It pins down that the general compound-core-type-Or splitter correctly handles the "target player/opponent controls" controller-suffix family under a compound subject — the same mechanism that makes Exhaustion and Icebreaker Kraken (and any future card in this shape) resolve correctly. This wasn't previously covered by an existing test (the closest sibling, `artifacts_and_creatures_your_opponents_control`, exercises a different controller-suffix form — "your opponents control", not "target player/opponent controls").

## Issue #5412

Left open with an explanatory comment rather than closed, since the maintainers may want to track the "stale coverage snapshot" methodology note separately. Not closing per repo convention pending maintainer input.


matthewevans (changes requested):
**[MED]** This fixes the claimed parser gap at the wrong seam and does not produce the claimed card-parse change. Evidence: `crates/engine/src/parser/oracle_target.rs:11058` adds a diagnostic proving the existing `parse_type_phrase` path already parses `creatures and lands target opponent controls` into the desired `Or`, while `crates/engine/src/parser/oracle_effect/subject.rs:1369` adds a new `CantUntap`-specific splitter that re-implements compound type/controller distribution instead of using or fixing the shared `parse_target`/`parse_type_phrase` building block. The parse-diff sticky for this head also reports “No card-parse changes detected,” which contradicts the PR’s claim that Exhaustion and Icebreaker Kraken move from strict-fail to supported. Why it matters: this adds a predicate-specific parser workaround with no demonstrated production coverage impact, so the next effect with the same compound subject shape will still need another special case. Suggested fix: remove the dedicated `CantUntap` dispatcher, identify the actual production gap in the generic subject/target path if one remains, and include parse-diff evidence showing the claimed cards change as intended.

matthewevans (approved):
Approved current head ae83bc43c52450495ee761451d1b53f1789c9e53. The prior redundant dispatcher is gone; the remaining diff is a single building-block parser regression test, the parse-diff sticky reports no card-parse changes, and the branch is mergeable. Auto-merge can wait for pending checks.

PR #5429, opened 2026-07-09
fix(client): player-attached Auras are selectable for every object choice (#5289)

### LINKED ISSUE #5289: Copy Enchantment seems unable to copy attached curses — Only the non-aura enchantments on the board are appearing as ta…
Reported in Discord: https://discord.com/channels/1485498006781427802/1523857023069061242/1523857023069061242

**Thread:** Copy Enchantment seems unable to copy attached curses
**Discord thread id:** `1523857023069061242`
**Discord message id:** `1523857023069061242`
**Cards:** copy enchantment

## Summary
Only the non-aura enchantments on the board are appearing as targets for [[Copy Enchantment]].

## Reported text

Only the non-aura enchantments on the board are appearing as targets for [[Copy Enchantment]].

## Attachments

- [game-state-turn-46-2026-07-07T00-59-11-085Z.zip](https://cdn.discordapp.com/attachments/1523857023069061242/1523857023383502959/game-state-turn-46-2026-07-07T00-59-11-085Z.zip?ex=6a4da1f2&is=6a4c5072&hm=2724c0f999628344fa6febea3ceb9915c33f540d0645075f36363d9609230a10&)

### copy enchantment
You may have this enchantment enter as a copy of any enchantment on the battlefield.

---
<sub>report_id: `discord:1523857023069061242:1523857023069061242:0` · discord: `1523857023069061242/1523857023069061242`</sub>

philluiz2323:
## Summary

Fixes #5289. Copy Enchantment could not copy an attached Curse — "only the non-aura enchantments on the board are appearing as targets". The engine was already correct; the Curse was rendered but inert on click. This makes player-attached Auras selectable for every object-choice prompt the board already supports.

> Frontend-only behavior change. No `crates/engine/` game logic is modified: the only engine-side change is an added `#[cfg(test)]` unit test that pins the existing, already-correct `find_copy_targets` behavior. The bug lives entirely in the display layer.

## Root cause

`find_copy_targets` filters the copy pool by zone and type only, so an Aura — attached to a permanent *or* to a player — is a legal enchantment permanent (CR 303.4a) and is already offered in `CopyTargetChoice::valid_targets` (CR 707.9). The new lib test pins that.

The gap is in the client:

- An Aura attached to a **permanent** renders on its host, and `PermanentCard` already lifts a host's attachment fan whenever one of them is actionable — so those were reachable.
- An Aura attached to a **player** (the Curse cycle) has no host permanent. `battlefieldProps.ts` drops it from the battlefield rows, so the HUD `EnchantmentsBadge` → `AttachmentsDialog` is its only surface.
- `DialogAttachmentCard` decided target eligibility with its own hard-coded `TargetSelection` / `TriggerTargetSelection` check. During a `CopyTargetChoice` the Curse rendered but was unclickable.

## Fix

Rather than extend that bespoke prompt list, the dialog now reads the two authorities that already exist:

- `waitingPlayer()` — the documented single authority for which seat must act.
- `getWaitingForObjectChoiceIds()` — the same function `GameBoard` seeds `validTargetObjectIds` from, which already handles `CopyTargetChoice`.

The dialog dispatches `ChooseTarget` for exactly the prompts the board does. The engine accepts `ChooseTarget` for `CopyTargetChoice`, `ExploreChoice`, `ReturnAsAuraTarget`, `RetargetChoice` and `CopyRetarget`, so this fixes the whole class rather than Copy Enchantment alone. The `TargetSelection` path is behavior-identical — `getWaitingForObjectChoiceIds` reads the same `current_legal_targets`.

`EnchantmentsBadge` also gains the board's lime valid-target ring when one of its Auras is a legal choice; without it a valid Curse stays invisible behind a collapsed badge.

## CR references

- `CR 707.9` — "enter as a copy of" chooses a permanent to copy; a choice, not targeting.
- `CR 303.4a` — an Aura is an enchantment permanent.
- `CR 303.4f` / `CR 303.4` — Auras attach to permanents and to players (Curse cycle).

## Verification

- `cargo test -p engine --lib find_copy_targets` → `4 passed; 0 failed`, including the new `find_copy_targets_includes_attached_auras_and_curses`, which builds the copy filter from Copy Enchantment's **real parsed Oracle text** and asserts the plain enchantment, the creature-attached Aura, and the player-attached Curse are all in the pool (and the creature host is not).
- `cargo fmt --all` clean.
- Frontend: `pnpm` is unavailable in my local environment, so `check-frontend` / `test-frontend` in CI are the first run of `vitest` / `type-check` on this diff. Added `getWaitingForObjectChoiceIds` coverage for `CopyTargetChoice` in `gameStateView.test.ts`.


matthewevans (commented):
Implementation review found no blocking issues at `c036e1811a6d5e2a38799f966006ad7aff804e57`.

Evidence checked: the change uses the existing engine-owned `getWaitingForObjectChoiceIds` choice surface, keeps the dialog click on the same `ChooseTarget { Object }` path already used by `PermanentCard` / `TargetingOverlay` for `CopyTargetChoice`, and adds coverage for both the frontend viewmodel choice list and the engine copy-target pool that includes attached Auras/Curses.

I am not approving/enqueueing from this sweep because the required checks are still pending and I did not run browser UI verification for the new badge/dialog interaction.

matthewevans:
Maintainer note: this was verified in browser - validate correct seam and ownership/responsibility location.

matthewevans (approved):
Implementation review is clean at `a47585100b763978c50ce7d2f18738cc2e2d60e6`.

Evidence checked in this sweep: the active diff remains the same four-file change after the maintainer merge; it routes the player-attached Aura dialog through the existing `getWaitingForObjectChoiceIds` / `ChooseTarget` object-choice authority, keeps engine logic in the engine, includes an engine-side Copy Enchantment pool regression for attached Aura/Curse permanents, and preserves the existing `bug` type label. The parse-diff sticky is still baseline-pending after the maintainer merge, but the current PR diff has no parser/card-data change and the prior sticky for this branch reported no parse changes.

PR #5433, opened 2026-07-09
fix(parser): fold article-led "or a <type> card" disjunct into the return filter (#5331)

### LINKED ISSUE #5331: overlord of the balemurk — Does not return a creature from my graveyard to my hand
Reported in Discord: https://discord.com/channels/1485498006781427802/1524141831477264424/1524141831477264424

**Thread:** overlord of the balemurk
**Discord thread id:** `1524141831477264424`
**Discord message id:** `1524141831477264424`
**Cards:** _none detected_

## Summary
Does not return a creature from my graveyard to my hand

## Reported text

Does not return a creature from my graveyard to my hand

## Attachments

- [game-state-turn-9-2026-07-07T19-52-52-317Z.zip](https://cdn.discordapp.com/attachments/1524141831477264424/1524141831636652304/game-state-turn-9-2026-07-07T19-52-52-317Z.zip?ex=6a4eab31&is=6a4d59b1&hm=21ef6844ce45b943b6eda53641f6771c0e95d57bb8309369023ce57093a86b0b&)

---
<sub>report_id: `discord:1524141831477264424:1524141831477264424:0` · discord: `1524141831477264424/1524141831477264424`</sub>

philluiz2323:
## Summary

Fixes #5331. Overlord of the Balemurk reads "… then you may return a non-Avatar creature card **or a planeswalker card** from your graveyard to your hand", but a planeswalker card in the graveyard was never offered. The named alternative was being dropped at parse time.

> Single-combinator parser fix, root-caused from a parse dump and covered by three building-block tests plus a full `--lib` regression. The change adds one composable combinator and one call-site wiring; it does not touch effects, resolver, or targeting semantics.

## Root cause

`parse_type_phrase` deliberately stops at an article-led `or` connector and returns the RHS as remainder — a bare `"or <type>"` union and an article-led `"or a <type>"` continuation are different productions. That contract is asserted by the existing `parse_type_phrase_leaves_article_led_or_rhs_as_remainder` (added with the Plan card type, #4182).

The return/`ChangeZone` parser then bound that remainder to `_rem` and discarded it. The parsed filter collapsed to:

```
Typed { type_filters: [Creature, Non(Subtype("Avatar"))] }
```

so the planeswalker alternative vanished. `assert_no_compound_remainder` — the debug guard that exists to catch precisely this — only rejects `" and "` remainders, so an `" or a …"` disjunct was lost **silently**.

## Fix

`fold_article_led_or_type_phrases` (in `oracle_target`) folds any number of trailing article-led disjuncts into a `TargetFilter::Or`, one leg per named alternative, each leg parsed by the full `parse_type_phrase` grammar. Built with nom combinators throughout (`space0` + `tag("or ")` + `opt(parse_article)`), so it covers the whole *"return an X card or a Y card"* class rather than this one card's shape.

It is a no-op when the remainder is not an article-led type phrase — so the call site invokes it unconditionally, and a trailing non-filter `or` clause (e.g. `"or you lose 2 life"`) is still handed back to the caller rather than swallowed, and cannot loop.

## CR references

- `CR 608.2c` — a spell's instructions are followed as written, so every alternative the Oracle text names must remain selectable.

## Verification

- `cargo test -p engine --lib` → **15884 passed, 0 failed** (full suite; a target-filter change is exactly the kind that silently breaks unrelated cards).
- Three new building-block tests in `oracle_target`:
  - `fold_article_led_or_unions_planeswalker_disjunct` — both alternatives survive as `Or` legs.
  - `fold_article_led_or_is_noop_without_disjunct` — filter and remainder returned untouched.
  - `fold_article_led_or_leaves_non_type_phrase_rhs` — a non-type-phrase `or` clause is neither swallowed nor looped on.
- `cargo fmt --all` clean.

## Note on the issue title

The report says "does not return a creature". The parse dump shows the **creature** alternative was always intact; what was unreachable is the **planeswalker** alternative. This PR restores the dropped disjunct. If the reporter also saw a creature fail to return, that would be a separate defect and I could not reproduce it — happy to dig further if a repro surfaces.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -1817,7 +1817,13 @@ pub(super) fn parse_targeted_action_ast(
         let (target, target_syntax, _count_for_shape) = match counted_return {
             Some((target, c)) => (target, TargetSyntax::TargetKeyword, c),
             None => {
-                let (target, _rem, syntax) = parse_target_with_syntax(target_text, ctx);
+                let (target, rem, syntax) = parse_target_with_syntax(target_text, ctx);
+                // CR 608.2c: "... a non-Avatar creature card or a planeswalker card"
+                // names two alternatives; `parse_target_with_syntax` stops at the
+                // article-led connector, so fold the remaining disjuncts in rather
+                // than dropping them (Overlord of the Balemurk).
+                let (target, _rem) =
+                    crate::parser::oracle_target::fold_article_led_or_type_phrases(target, rem);
                 #[cfg(debug_assertions)]
                 assert_no_compound_remainder(_rem, text);
                 (target, syntax, QuantityExpr::Fixed { value: 0 })
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -1866,6 +1866,65 @@ pub fn parse_type_phrase(text: &str) -> (TargetFilter, &str) {
     parse_type_phrase_with_ctx(text, &mut ParseContext::default())
 }
 
+/// Consume an `"or"` connector whose right-hand side is article-led
+/// (`"or a "` / `"or an "`) or bare (`"or "`), returning the text after it.
+fn parse_or_connector_with_optional_article(input: &str) -> OracleResult<'_, ()> {
+    let (input, _) = nom::character::complete::space0.parse(input)?;
+    let (input, _) = tag("or ").parse(input)?;
+    let (input, _) = opt(nom_primitives::parse_article).parse(input)?;
+    Ok((input, ()))
+}
+
+/// CR 608.2c: A filter phrase may name its alternatives as an article-led
+/// disjunction — "a non-Avatar creature card **or a** planeswalker card"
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
**[MED]** `fold_article_led_or_type_phrases` maps a lowercased copy back to the original string by suffix byte length. Evidence: `crates/engine/src/parser/oracle_target.rs:1901-1909` calls `rest.to_lowercase()`, parses the connector on that transformed string, then computes `&rest[rest.len() - after_connector.len()..]`; nearby parser code uses direct case-insensitive combinators such as `tag_no_case` instead of this transformed-offset pattern. Why it matters: Unicode case conversion is not byte-length preserving, so an uppercase non-ASCII character in the remaining Oracle text can put that slice on the wrong byte boundary or panic, and it leaves a brittle parser idiom for the next contributor to copy. Suggested fix: parse the connector on the original `rest` with `space0` + `tag_no_case("or ")` + the existing article parser, or otherwise compute the consumed byte count from the original input before passing the original RHS to `parse_type_phrase`.

The parse-diff sticky is also absent for this parser-source head while CI is still pending, so please wait for the current-head parse-diff evidence before re-requesting merge.

Issue #5300, opened 2026-07-07
feat(parser): Lignify-class enchanted subtype-only type-change with base P/T (CR 205.1a / CR 613.1d)


jsdevninja:
## Summary

Auras in the **Lignify class** — `"Enchanted creature is a <creature subtype> with base power and toughness N/N and loses all abilities."` — strict-fail today. `parse_enchanted_is_type` requires at least one **core card type** word in the granted type phrase (`Insect artifact creature`, `blue Frog creature`, …). When the grammar names only a creature subtype before the base-P/T seam (`Treefolk`, `Wall`, …), `granted_core_types` stays empty and the handler returns `None`, so the line falls through to `Unimplemented` even though the runtime already supports every emitted modification (`SetCardTypes`, `RemoveAllSubtypes`, `AddSubtype`, `SetPower`/`SetToughness`, `RemoveAllAbilities`).

**Type specimen:** Lignify — `"Enchanted creature is a Treefolk with base power and toughness 0/4 and loses all abilities."`

Scryfall finds ~13 auras matching `is a` + `with base power and toughness` + `loses all abilities`; the subtype-only shape is the uncovered sibling of the Frogify/Darksteel Mutation family handled by `parse_enchanted_is_type`.

## Proposed approach (mirror #5142 / #4770 structure)

- Add `parse_enchanted_is_creature_subtype_with_base_pt` in `oracle_static/type_change.rs`
- Split on the existing ` with base power and toughness ` seam (same as `parse_enchanted_is_type`), parse leading creature subtype(s) via `parse_subtype`, **decline** if a core-type word appears (so `Insect artifact creature` stays on the general handler)
- Compose existing Layer-4/6/7b modifications in written order: optional leading `RemoveAllAbilities`, `SetCardTypes { Creature }` (CR 205.1a replacement), `RemoveAllSubtypes { Creature }`, `AddSubtype`, `SetPower`/`SetToughness`, trailing-clause mods from `parse_continuous_modifications`
- Dispatch after `parse_enchanted_becomes_type_with_ability`, **before** `parse_enchanted_is_type`
- No new engine variant, no new runtime

## CR references

- CR 205.1a — setting a creature subtype replaces existing creature subtypes
- CR 613.1d (Layer 4) — type-changing continuous effect
- CR 613.1f (Layer 6) — ability removal
- CR 613.4b (Layer 7b) — base power and toughness

## Test plan

[...]

PR #5438, opened 2026-07-09
fix(parser): subtype-only enchanted creature type-change without base P/T (Coalition Flag, CR 205.1a)

nickmopen:
## Summary

`"Enchanted creature is a <creature subtype>."` with **no base power and toughness** — the type specimen is **Coalition Flag** (`"Enchanted creature is a Flagbearer."`) — dropped the whole static, so the enchanted creature never became the named type. (Today it parses to `statics = 0`.)

## Root cause

`parse_enchanted_is_type` routes the `is a <phrase>` copula through `classify_type`, which puts a bare creature subtype (`Flagbearer`) into `granted_subtypes` while leaving `granted_core_types` empty. The subtype-only branch (added for the Lignify class, #5300) only proceeds when a **base P/T** is present — so the P/T-less Coalition Flag shape fell through the `granted_core_types.is_empty()` guard and returned `None`.

## Fix

Generalize the disambiguator from the basic-land subtype change (`"Enchanted land is a Mountain"`, which must return `None` and defer to `SetBasicLandType`). The Lignify path used a base P/T (creature-only) as the signal; a subtype-only creature change is equally identified by the **enchanted permanent itself being a creature** (`perm_tf == TypeFilter::Creature`). A land enchant (`perm_tf == Land`) still returns `None` and defers to the basic-land handler.

Coalition Flag now emits the CR 205.1a subtype replacement — `RemoveAllSubtypes{Creature}` → `AddSubtype(Flagbearer)` — with **no `SetCardTypes`** (card types preserved) and **no P/T change**. Reuses the existing emission path; no new engine variant, no new runtime.

## Tests

- `coalition_flag_subtype_only_no_base_pt_type_change` — asserts `RemoveAllSubtypes{Creature}` before `AddSubtype(Flagbearer)`, and **no** `SetCardTypes`/`SetPower`/`SetToughness`.
- `enchanted_land_is_mountain_stays_basic_land_change` — regression guard: `"Enchanted land is a Mountain"` still routes to `SetBasicLandType`, never `AddSubtype`.
- Full engine lib suite: **15,756 passed / 0 failed**; `cargo fmt`, `check-parser-combinators.sh`, `check-engine-authorities.sh` pass; clippy clean.


matthewevans (approved):
Maintainer review: approved after local handler review, CR annotation fixup, parse-diff reconciliation, and Tilt/GitHub verification handoff.

PR #5440, opened 2026-07-09
fix(parser): recognize "transformed" as a filter quality (Mutagen Connoisseur)

### LINKED ISSUE #5439: Parser: "transformed" filter quality is dropped in count/type phrases — Mutagen Connoisseur gets flat +1/+0
## Summary

The **"transformed"** designation is not recognized as a filter quality in type phrases, so an object count or selector scoped to transformed permanents silently loses the qualifier. The card then gets the wrong value.

## Affected card

- **Mutagen Connoisseur** — *"Flying, vigilance. This creature gets +1/+0 for each transformed permanent you control."*

## Current behavior

The static parses to a **flat** `AddPower { value: 1 }, AddToughness { value: 0 }` — the *"for each transformed permanent you control"* scaling is dropped. Mutagen Connoisseur therefore always gets **+1/+0** instead of **+N/+0** where N is the number of transformed permanents you control. The parser's swallow-audit flags a `DynamicQty` `SwallowedClause` on the line.

## Evidence (isolation)

| Phrase | Scales correctly today |
|---|---|
| `for each tapped permanent you control` | ✅ |
| `for each artifact you control` | ✅ |
| `for each creature you control` | ✅ |
| `for each transformed permanent you control` | ❌ (flat) |
| `for each transformed creature you control` | ❌ (flat) |

Only the **transformed** quality is missing — every other state/type qualifier scales.

## Root cause

`FilterProp` has boolean battlefield-state qualities (`Tapped`, `FaceDown`, `Suspected`, `IsSaddled`, …) recognized as type-phrase adjective prefixes, but **no `Transformed` variant**. `GameObject::transformed` already exists (the runtime can detect it), so this is purely a missing filter quality — the symmetric sibling of `FaceDown` (`obj.face_down`).

## Expected behavior

`transformed <type>` parses to a `FilterProp::Transformed` (matched against `GameObject::transformed`), so the count/selector scales correctly.

## Rules

CR 711.4 (a transformed permanent is a transforming double-faced permanent with its back face up).

joaovictor712:
Closes #5439.

## The bug

**Mutagen Connoisseur** — *"This creature gets +1/+0 for each transformed permanent you control."* — parsed to a **flat** `AddPower{1}/AddToughness{0}`. The *"for each transformed permanent you control"* scaling was silently dropped, so the card always gets +1/+0 instead of +N/+0. The parser's own swallow-audit flags a `DynamicQty` `SwallowedClause` on the line.

## Root cause

Isolation shows only the **transformed** quality is affected — `tapped`, `face-down`, `artifact`, `creature` all scale correctly:

| Phrase | Scales today |
|---|---|
| `for each tapped permanent you control` | ✅ |
| `for each artifact you control` | ✅ |
| `for each transformed permanent you control` | ❌ (flat) |

`FilterProp` has boolean battlefield-state qualities (`Tapped`, `FaceDown`, `Suspected`, `IsSaddled`, …) recognized as type-phrase adjective prefixes, but no `Transformed` variant — even though `GameObject::transformed` already exists.

## Fix

Adds `FilterProp::Transformed`, the **symmetric sibling of `FilterProp::FaceDown`** — a per-object battlefield state read from `GameObject::transformed`:
- recognized as an adjective prefix (`transformed permanent`/`transformed creature`) via `parse_property_filter` + the `parse_combat_status_prefix` whitelist;
- matched at runtime in `filter.rs` (`FilterProp::Transformed => obj.transformed`);
- threaded through the axis-classification lists / coverage display alongside `FaceDown`.

**No runtime state change** — the object flag was already tracked; this only lets the parser build the count/selector that reads it. Consistent with the existing individual-variant design for boolean state qualities (`Suspected`, `Renowned`, `IsSaddled` were all added the same way), so no parameterization refactor is implied.

## Tests

- `test_parse_property_filter_transformed` — `"transformed permanent"` → `FilterProp::Transformed`.
- `transformed_property_matches_only_transformed` — the filter matches a transformed permanent and rejects a normal one.

## Verification

`cargo fmt --all` ✓ · engine lib suite **15,904 pass / 0 fail** + 2 new ✓ · `clippy -p engine --all-targets -D warnings` clean ✓

CR 711.4.

### DIFF
--- a/crates/engine/src/ai_support/filter.rs
+++ b/crates/engine/src/ai_support/filter.rs
@@ -731,6 +731,7 @@ fn filterprop_reads_only_candidate_fp(p: &FilterProp) -> bool {
         | FilterProp::Historic
         | FilterProp::NotHistoric
         | FilterProp::FaceDown
+        | FilterProp::Transformed
         | FilterProp::HasXInManaCost
         | FilterProp::HasManaAbility
         | FilterProp::HasNoAbilities
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -2338,6 +2338,7 @@ fn legacy_filter_prop(p: &FilterProp) -> bool {
         | FilterProp::AttackedOrBlockedThisTurn
         | FilterProp::CountersPutOnThisTurn { .. }
         | FilterProp::FaceDown
+        | FilterProp::Transformed
         | FilterProp::HasXInManaCost
         | FilterProp::HasXInActivationCost
         | FilterProp::WasKicked
@@ -2583,6 +2584,7 @@ fn member_bound_filter_prop(p: &FilterProp) -> bool {
         | FilterProp::AttackedOrBlockedThisTurn
         | FilterProp::CountersPutOnThisTurn { .. }
         | FilterProp::FaceDown
+        | FilterProp::Transformed
         | FilterProp::HasXInManaCost
         | FilterProp::HasXInActivationCost
         | FilterProp::WasKicked
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -970,6 +970,7 @@ fn fmt_typed_filter(tf: &TypedFilter) -> String {
             FilterProp::HasSingleTarget => parts.push("single target".into()),
             FilterProp::Modal => parts.push("modal spell".into()),
             FilterProp::FaceDown => parts.push("face-down".into()),
+            FilterProp::Transformed => parts.push("transformed".into()),
             FilterProp::TargetsOnly { filter } => {
                 parts.push(format!("targets only {}", fmt_target(filter)));
             }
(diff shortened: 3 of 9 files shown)


galuis116:
Good catch on the CR reference — fixed. `CR 711` is indeed **Leveler Cards**, so `CR 711.4` was wrong.

One correction to the suggestion: the bot proposed `CR 701.28g`, but `CR 701.28` is **Convert**, not Transform. The definition of a *"transformed permanent"* is **CR 701.27g**:

> 701.27g Some spells and abilities refer to a "transformed permanent." This phrase refers to a double-faced permanent on the battlefield with its back face up. …

I've updated all six citations (code + tests + commit message) to **CR 701.27g**, which also matches the existing convention in the repo — `FilterProp::SourceIsTransformed` already cites CR 701.27g (`types/ability.rs`). Verified against `docs/MagicCompRules.txt`.

matthewevans (approved):
Implementation review remains clean at `ac6cd6427541e0e65d3d1365d9534a81809f70e6`. The change is at the shared `FilterProp`/typed-filter parser seam, uses the existing nom property combinator path, carries the verified `CR 701.27g` citation, and the refreshed parse-diff shows only the expected transformed-permanent class expansion.

PR #5443, opened 2026-07-09
fix(parser): graveyard-exile replacement accepts counter rider + prefix word order (Dauthi Voidwalker #5245)

### LINKED ISSUE #5245: Dauthi's Scamming me — [[Dauthi Voidwalker]]'s sac ability doesnt resolve properly.
Reported in Discord: https://discord.com/channels/1485498006781427802/1522452633925980241/1522452633925980241

**Thread:** Dauthi's Scamming me
**Discord thread id:** `1522452633925980241`
**Discord message id:** `1522452633925980241`
**Cards:** dauthi voidwalker
**Parser status:** has_gaps

## Summary
[[Dauthi Voidwalker]]'s sac ability doesnt resolve properly.

## Reported text

[[Dauthi Voidwalker]]'s sac ability doesnt resolve properly. It sacks but gives me nothing to cast :(

### dauthi voidwalker
Shadow (This creature can block or be blocked by only creatures with shadow.)
If a card would be put into an opponent's graveyard from anywhere, instead exile it with a void counter on it.
{T}, Sacrifice this creature: Choose an exiled card an opponent owns with a void counter on it. You may play it this turn without paying its mana cost.

---
<sub>report_id: `discord:1522452633925980241:1522452633925980241:0` · discord: `1522452633925980241/1522452633925980241`</sub>

shin-core:
## Summary

Fixes #5245. **Dauthi Voidwalker** — *"If a card would be put into an opponent's graveyard from anywhere, instead exile it with a void counter on it."* — strict-failed the replacement parser (`Unimplemented` `replacement_structure`).

`parse_graveyard_exile_replacement` recognized its exile outcome with a hardcoded `tag("exile it instead")`, which matched **neither**:
- the `"instead exile it"` **prefix** word order, nor
- the `"with a void counter on it"` **counter rider** (CR 122.1).

## Fix

Delegate the exile outcome to the shared **`parse_exile_anaphor_clause`** building block — already used by the self-die (`~ would die`) and non-self creature-die exile branches. It handles both word orders and lifts an inline `with N <type> counter(s) on it` modifier into `enter_with_counters`.

The counters ride on the exile redirect's `ChangeZone`, which `event_modifiers_for_ability` → `extract_etb_counters` already reads and applies as the card enters exile (CR 614.1a). No engine/type changes — pure parser reuse of existing infrastructure.

This covers the whole **class** (`would be put into [any/opponent's] graveyard → [instead] exile it [with counters]`), not just Dauthi. The shuffle-back outcomes are untouched — they use distinct phrasings the anaphor recognizer never matches.

## Rules

- CR 614.1a — replacement effects that exile instead of moving to the graveyard.
- CR 122.1 — the exiled object enters with the specified counter(s).
- CR 730.3e / CR 111.1 — "a card" subject is token-excluding (`NonToken`), unchanged.

## Tests

- `dauthi_voidwalker_exile_with_void_counter` — asserts opponent-owned + `NonToken` scope and **one void counter** on the exile redirect.
- `graveyard_exile_no_counter_suffix_word_order_unchanged` — regression guard that the plain `"exile it instead"` suffix (Leyline / Rest in Peace) stays **counter-free**.
- All **336** replacement-parser tests pass; `leyline_of_the_void_opponent_scoped` and `rest_in_peace_graveyard_exile` unchanged.

## Verification

- `cargo check -p engine` — clean
- Confirmed end-to-end: the full Dauthi card no longer emits a `replacement_structure` `Unimplemented`.

> Scope note: Dauthi's activated ability also has a separate `"Choose an exiled card …"` gap; that's an independent parser issue and is intentionally out of scope for this focused fix.

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -4947,11 +4947,18 @@ fn parse_graveyard_exile_replacement(
     }
 
     // The outcome clause ("exile it instead" or the shuffle-back phrasing)
-    // determines what ChangeZone + sub_ability chain we emit.
+    // determines what ChangeZone + sub_ability chain we emit. The exile variant
+    // carries any `with N <type> counter(s) on it` rider (CR 122.1) lifted by the
+    // shared anaphor recognizer — Dauthi Voidwalker's "instead exile it with a
+    // void counter on it" is the exemplar.
     #[derive(Clone)]
     enum Outcome {
-        Exile,
-        ShuffleBack { reveal: bool },
+        Exile {
+            enter_with_counters: Vec<(CounterType, QuantityExpr)>,
+        },
+        ShuffleBack {
+            reveal: bool,
+        },
     }
 
     // CR 730.3e + CR 111.1: the subject's token axis. "a card or token" is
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**[MED]** Forbidden Crypt is now parsed as a global graveyard redirect. Evidence: `crates/engine/src/parser/oracle_replacement.rs:5017` parses any non-opponent graveyard phrase through `Scope::Any`, and `crates/engine/src/parser/oracle_replacement.rs:5075` only adds an ownership filter for `Scope::Opponent`; Scryfall Oracle for Forbidden Crypt says, "If a card would be put into your graveyard from anywhere, exile that card instead." Why it matters: this PR newly makes Forbidden Crypt parse as a supported `Moved` replacement, but that replacement would also exile opponents' cards going to their graveyards. Suggested fix: model the graveyard scope as `Any | You | Opponent` and add a `FilterProp::Owned { controller: ControllerRef::You }` assertion for the Forbidden Crypt class.

shin-core:
Thanks @matthewevans — good catch. Fixed in 9111a5536.

Added `Scope::You` so "your graveyard" (Forbidden Crypt) maps to `FilterProp::Owned { controller: You }`, symmetric with the existing opponent's-graveyard scope (Leyline → `Owned { Opponent }`). It's recognized before the generic ` graveyard` fallthrough, so the possessive owner is captured rather than swallowed as unscoped. Added a `forbidden_crypt_your_graveyard_scopes_to_you` test asserting You-owned + NonToken and **not** opponent-scoped; Dauthi/Leyline/Rest in Peace are unchanged (337 replacement tests pass, clippy clean).

matthewevans (approved):
Maintainer re-review: current head addresses the Forbidden Crypt owner-scope finding; parser diff and tests are in scope.

PR #5406, opened 2026-07-09
feat(parser): compound-subject chosen-type static (Rukarumel, Biologist) (#5392)

davion-knight:
## Summary

Fixes #5392. `Rukarumel, Biologist`'s chosen-creature-type static —
`Slivers you control and nontoken creatures you control are the chosen type in addition to their other creature types.` (Oracle text verified against `client/public/card-data.json`) — previously **strict-failed** (`static_abilities: []`); the whole line collapsed into an `Unimplemented` ability.

It's the **compound-subject** sibling of the "chosen type" static class (Arcane Adaptation, Xenograft, Lifecraft Engine) — the same structural shape as **#5219**'s compound-subject animation static: two `you control` conjuncts joined by "and", each individually resolvable, but the combination unreached because the single-subject dispatcher's subject matcher only accepts three fixed literal forms.

## Fix (mirrors #5219)

- **`parse_compound_you_control_chosen_type_static`** delegates subject resolution *whole* to the already-generic `parse_continuous_subject_filter` (which unions `X you control and Y you control` into an `Or` via `parse_controlled_compound_continuous_subject_filter`) and owns only the additive chosen-type predicate. Declines unless the subject is a genuine `Or` of 2+ filters (single-subject lines stay with the existing handler) and declines non-additive (CR 205.1a SET) predicates.
- Extract the shared **`parse_chosen_type_additive_suffix`** retain-suffix helper, extended to optionally accept `creature ` before `types` (Rukarumel's "…other **creature** types" spelling). The single-subject forms omit it, so `opt` matches nothing there and their behavior is unchanged.
- A "same is true for …" tail prefix matcher + `push_same_is_true_static_tail` call in `oracle.rs` preserve Rukarumel's trailing non-battlefield-zone sentence as an `Unimplemented` residual — matching Arcane Adaptation / Maskwood Nexus, rather than strict-failing the whole line.

## Architecture

- Nom combinators throughout; reuses `parse_continuous_subject_filter`, `each`-style helpers, and the existing tail-splitter.
- Builds for the **class** (any `X you control and Y you control` compound), not the single card.
- **No new `ContinuousModification` / `Layer` / runtime** — `AddChosenSubtype { CreatureType }` is already fully wired (Arcane Adaptation / Xenograft / Lifecraft exercise it end-to-end).
- CR-annotated: **CR 611.3 / CR 607.2d / CR 205.1b / CR 205.3m**.

## Testing

[...]

PR #5444, opened 2026-07-09
feat(parser): compound negated-type subject animation with granted abilities (Bello, Bard of the Brambles)

### LINKED ISSUE #5442: parser: compound-subject animation static strict-fails (Bello, Bard of the Brambles)
## Summary

**Bello, Bard of the Brambles** (Bloomburrow Commander)'s compound-subject animation static —

> "During your turn, each non-Equipment artifact and non-Aura enchantment you control with mana value 4 or greater is a 4/4 Elemental creature in addition to its other types and has indestructible, haste, and 'Whenever this creature deals combat damage to a player, draw a card.'"

(Oracle text verified against the Scryfall API) — currently **strict-fails** (verified against `data/coverage-data.json`: `supported: false`, single gap `"Effect:static_structure"` pointing at the entire line); the whole line collapses into an `Unimplemented` ability.

This is the same *structural* gap shape as #5219 (Life and Limb's compound-subject animation static) and #5392/#5406 (Rukarumel's compound-subject chosen-type static): a compound subject joined by "and" where each individual conjunct is already independently resolvable, and the animation+keyword+ability-grant predicate is already independently resolvable — but the *combination* of a two-conjunct compound subject with this predicate class is unreached.

## Proof the pieces are already independently correct

Two real, currently-`supported: true` cards confirm every constituent piece of Bello's line already works in isolation (verified against `data/coverage-data.json`):

- **Starfield of Nyx**: "each other non-Aura enchantment you control is a creature in addition to its other types" — proves the single negated-type conjunct ("non-Aura enchantment you control ... is a creature in addition to its other types") already resolves correctly, including the `non-` prefix and the CR 205.1b additive "in addition to its other types" marker.
- **Mephidross Vampire**: "Each creature you control is a Vampire in addition to its other creature types and has 'Whenever this creature deals damage to a creature, put a +1/+1 counter on this creature.'" — proves the combined animation + CR 205.1b additive-type + granted-ability-string predicate (the "is a `<type>` ... in addition to its other types and has '`<ability>`'" shape) already resolves correctly for a single subject.

Between them, these are the two halves Bello combines: a *compound* ("non-Equipment artifact **and** non-Aura enchantment") subject, each half individually proven by Starfield of Nyx's negated-type-conjunct pattern, applied to the animation+ability-grant predicate shape Mephidross Vampire already proves — plus a keyword list (`indestructible, haste`) folded into the same `has ... and "..."` clause, and a `During your turn,` turn-window gate (an established, separately-supported CR 611.3a condition prefix elsewhere in the codebase). So this is **not** a predicate gap — it is purely a **compound-subject** gap, structurally identical to the class #5219/#5406 already generalized for other predicate families.

## Root cause (to be traced precisely at implementation time)

The existing compound-subject animation dispatcher, `parse_compound_all_subjects_type_change` (`crates/engine/src/parser/oracle_static/type_change.rs`, the #5219 fix), only matches the **plural "All `<X>` and all `<Y>` are `<predicate>`"** grammar (repeated `all` quantifier before each conjunct, plural copula `are`). Bello uses the **singular "each `<X>` and `<Y>` ... is a `<predicate>`"** grammar (one shared `each` quantifier, singular copula `is a`) — a different surface grammar over the same underlying CR 611.3/CR 205.1b semantics, so `parse_compound_all_subjects_type_change` declines it and it falls through to whatever single-subject "each `<X>` you control is a ..." dispatcher handles Starfield of Nyx / Mephidross Vampire, which in turn only accepts a single leaf subject (not an "`<X>` and `<Y>`" compound before the copula).

Needs tracing at implementation time to confirm the exact single-subject dispatcher (candidates: a sibling of `parse_each_noncreature_subject_is_creature_with_pt_mv` in `type_change.rs`, or a dispatcher in `oracle_effect/animation.rs`) and exactly which stage drops the second ("and non-Aura enchantment") conjunct.

## Proposed fix (mirrors #5219 / #5406's structural shape)

Add a compound-subject sibling for this "each `<X>` and `<Y>` ... is a `<predicate>`" class, following the same shape as #5219 and #5406:

1. A new parser function that recognizes the singular "each `<X>` and `<Y>` [you control] [with `<suffix>`] is a/an `<predicate>`" shape, resolves each conjunct's leaf subject independently (reusing whatever single-conjunct subject resolver already serves Starfield of Nyx), and `Or`-combines them — distributing any shared suffix (controller, mana-value gate) to every leg, mirroring `distribute_controller_to_or` / the `Or`-union pattern used throughout the compound-subject family.
2. Declines unless the subject is a genuine compound (2+ conjuncts) so single-subject lines (Starfield of Nyx, Mephidross Vampire) keep routing to their existing dispatcher unchanged.
3. Delegates the *entire* predicate ("is a 4/4 Elemental creature in addition to its other types and has indestructible, haste, and '`<ability>`'") to the already-general `parse_continuous_gets_has` (or whatever predicate builder Mephidross Vampire's line already routes through) rather than reinventing animation/keyword/ability-grant parsing — matching #5406's "own only the subject-compound delegation, reuse the existing predicate builder wholesale" discipline.
4. Wires the leading `During your turn,` condition the same way other turn-window-gated statics already do (e.g. `StaticCondition::DuringYourTurn`, as used by `parse_compound_subject_keyword_static` for the Gruul Spellbreaker class).
5. Tests: a parser-shape test on the modeled compound sentence (asserting the `Or` of two negated-type `Typed` filters, the `AddType{Creature}` + base-P/T + `AddKeyword` modifications, and the granted-ability string), a full-oracle test for Bello, and no-regression tests confirming Starfield of Nyx and Mephidross Vampire (single-subject) still parse unchanged.

No new `ContinuousModification` / `Layer` / runtime is expected — `AddType`, base P/T setting (Layer 7b), `AddKeyword`, and ability-granting are all already exercised end-to-end by Starfield of Nyx and Mephidross Vampire.

## CR references (grep-verified against `docs/MagicCompRules.txt` at implementation time)

- CR 611.3 — a continuous static effect applies to every object its text indicates (the compound `Or`-distributed affected set).
- CR 205.1b — additive type/subtype grants ("in addition to its other types").
- CR 613.1d / CR 613.4b — Layer 4 (type change) and Layer 7b (base P/T) continuous effects.
- CR 611.3a — turn-window ("During your turn") condition gating.

## Scope

Parser-only (`crates/engine/src/parser/oracle_static/type_change.rs` and/or `oracle_effect/animation.rs`, plus their test files) — no engine types/runtime changes expected, matching #5219's and #5406's file footprint.

jsdevninja:
## Summary

Fixes #5442. Bello, Bard of the Brambles' static —

> "During your turn, each non-Equipment artifact and non-Aura enchantment you control with mana value 4 or greater is a 4/4 Elemental creature in addition to its other types and has indestructible, haste, and \"Whenever this creature deals combat damage to a player, draw a card.\""

(Oracle text verified against the Scryfall API) — previously strict-failed (`Unimplemented("static_structure")`, `data/card-data.json`); the whole line collapsed into a gap.

Same *structural* shape as #5219 (Life and Limb's compound-subject animation static): a compound subject joined by "and" where each conjunct is independently resolvable and the predicate class is independently resolvable elsewhere, but the *combination* was unreached. Bello's variant differs from #5219's own fix (`parse_compound_all_subjects_type_change`) in two ways that make it a distinct class, not a duplicate:

- **Subject grammar**: Life and Limb repeats the quantifier per conjunct ("All `<X>` **and all** `<Y>`", plural "are"); Bello has a single shared quantifier and per-conjunct *negated*-type exclusions ("each non-Equipment artifact **and** non-Aura enchantment", singular "is") with a trailing qualifier (`you control`, `with mana value 4 or greater`) that applies to *both* conjuncts despite appearing once, after the second.
- **Predicate**: alongside the type change, Bello grants a *mixed* bare-keyword + quoted-trigger list ("indestructible, haste, and \"...\"") — a combination the shared additive-type-clause helper only partially supported (quoted portion only).

## Fix

`parse_each_compound_subject_type_change` (`crates/engine/src/parser/oracle_static/type_change.rs`), dispatched in `parse_static_line_inner` right after its closest sibling (`parse_each_noncreature_subject_is_creature_with_pt_mv`):

- **Subject**: delegates *wholesale* to the general target-phrase grammar (`parse_type_phrase`) instead of hand-rolling a conjunct splitter. That grammar already recurses per `"and"`-leg — restarting its own leading `non-` scan on each recursive call (`starts_with_type_word`'s `non-` arm) — and backfills the shared trailing qualifiers (controller, mana value) from the last leg onto every earlier leg via `distribute_controller_to_or` / `distribute_properties_to_or`. Declines (via a `2+`-leg `Or`-of-`Typed` guard) unless the subject is a genuine compound, so single-subject lines keep falling through to their existing dispatcher unchanged.
- **Predicate**: composes `parse_animation_spec` (base P/T + leading type/subtype grant) with `parse_additive_type_clause_modifications` (any extra type noun the animation spec stops short of), then extracts the granted-ability tail locally — reusing the same `split_keyword_list` + `push_grant_clause_modifications` + `parse_quoted_ability_modifications` composition `parse_continuous_modifications` already uses elsewhere — so a *mixed* bare-keyword/quoted-trigger list is captured in full. Kept local to the new function rather than folded into the shared additive-type-clause helper, which has several other call sites this change should not risk perturbing.
- Gated on the CR 205.1b additive marker ("in addition to its/their other types"), mirroring #5219's own review-fix, so a bare replacement compound isn't misclaimed as additive.

I cross-checked this design against #5406 (Rukarumel, cited in #5442 as the structurally closest precedent): that PR's subject-delegation target (`parse_continuous_subject_filter`) *requires* every conjunct to independently carry its own controller anchor (`"Slivers you control and nontoken creatures you control"`, repeated per leg) — which does not fit Bello's shared, non-repeated `"you control"`. `parse_type_phrase`'s distribution machinery is the correct fit here.

## CR references (grep-verified against `docs/MagicCompRules.txt`)

- CR 611.3 / CR 611.3a — a continuous effect from a static applies at any moment to whatever its text indicates (the compound `Or`-distributed affected set; the `During your turn,` window).
- CR 613.1d — Layer 4 type-changing effects.
- CR 613.4b — Layer 7b base power/toughness.
- CR 205.1b — additive type/subtype grants ("in addition to its other types").
- CR 604.1 / CR 603.1 — the granted quoted ability is a triggered ability (`GrantTrigger`).
- CR 702.10 / CR 702.12 — Haste / Indestructible.

## Testing

**Environment note**: this sandbox has no Tilt, no MSVC Build Tools, and no WSL/Docker, so I could not run `cargo build`/`test`/`clippy` locally. I verified as much as this environment allows:
- `cargo fmt --all -- --check` — clean.
- `./scripts/check-parser-combinators.sh upstream/main` — clean (no forbidden string-dispatch patterns in the added lines).
- Every helper function's signature, behavior, and call-site composition was traced by reading source directly (not from memory) — including manually tracing `parse_type_phrase`'s recursive per-leg negation handling and its `distribute_*_to_or` backfill functions line-by-line to confirm the compound-subject + shared-qualifier distribution actually produces the intended `Or` shape.
- **This PR has not been compiled or test-run.** Please treat `cargo test -p engine`/Tilt's `test-engine` as the first real verification pass before merge.

New tests (`oracle_static/tests.rs`):
- `bello_compound_negated_type_subject_animation_with_granted_abilities` — full Oracle text: `Or`-of-2 negated-type legs (non-Equipment artifact / non-Aura enchantment, both `you control` + `Cmc >= 4`), `DuringYourTurn` condition, base P/T, type/subtype grant, both bare keywords, and the granted trigger.
- `compound_negated_type_subject_animation_declines_non_additive_predicate` — the CR 205.1b gate declines a compound predicate missing the additive marker.
- `compound_negated_type_subject_animation_single_subject_falls_through` — a single-subject line is not claimed by the compound handler.

## Scope

Parser-only (`crates/engine/src/parser/oracle_static/{type_change,dispatch,tests}.rs`) — no engine types/runtime changes; `SetPower`/`SetToughness`/`AddType`/`AddSubtype`/`AddKeyword`/`GrantTrigger` are all pre-existing, already-wired modifications.

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -1998,6 +1998,18 @@ pub(crate) fn parse_static_line_inner(
         return Some(def);
     }
 
+    // CR 611.3 + CR 613.4b + CR 205.1b: compound-subject animation whose
+    // subject is a heterogeneous union of negated-type legs ("each non-X Y
+    // and non-Z W ... is a [P/T] [type] creature in addition to its other
+    // types and has [keywords/granted ability]" — Bello, Bard of the
+    // Brambles). Delegates the subject to the general target-phrase grammar
+    // (see the function doc comment), so it does not need to precede any
+    // sibling here — a single-subject line simply fails its 2+-leg `Or` guard
+    // and falls through unchanged.
+    if let Some(def) = parse_each_compound_subject_type_change(&tp, &text) {
+        return Some(def);
+    }
+
     // --- "~ can't be blocked [by filter] [as long as condition]" ---
     // CR 509.1b: Handles unconditional, conditional, and filter-based "can't be blocked".
     // "except by" patterns are handled separately by CantBeBlockedExceptBy.
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -4095,7 +4095,7 @@ fn bello_compound_negated_type_subject_animation_with_granted_abilities() {
 
     let cmc_ge_4 = FilterProp::Cmc {
         comparator: Comparator::GE,
-        value: QuantityExpr::Fixed(4),
+        value: QuantityExpr::Fixed { value: 4 },
     };
     // Artifact conjunct: a non-Equipment artifact you control, mana value >= 4.
     assert!(
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -4196,12 +4196,12 @@ fn compound_negated_type_subject_animation_declines_non_additive_predicate() {
 // a regression introduced by adding the compound handler).
 #[test]
 fn compound_negated_type_subject_animation_single_subject_falls_through() {
+    let line = "each non-Equipment artifact you control with mana value 4 or greater is a 4/4 \
+                Elemental creature in addition to its other types.";
+    let lower = line.to_lowercase();
+    let tp = crate::parser::oracle_util::TextPair::new(line, &lower);
     assert!(
-        parse_static_line_multi(
-            "each non-Equipment artifact you control with mana value 4 or greater is a 4/4 \
-             Elemental creature in addition to its other types."
-        )
-        .is_empty(),
+        super::type_change::parse_each_compound_subject_type_change(&tp, line).is_none(),
         "single-subject line must not be claimed by the compound-subject handler"
     );
 }
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -4107,6 +4107,16 @@ fn bello_compound_negated_type_subject_animation_with_granted_abilities() {
         "artifact conjunct must be non-Equipment artifact you control, MV>=4: {:?}",
         def.affected
     );
+    // Regression guard: the artifact conjunct's OWN negation (non-Equipment)
+    // must not cross-contaminate into the enchantment conjunct's non-Aura
+    // negation — each leg carries only its own qualifier.
+    assert!(
+        !filters.iter().any(|f| matches!(f, TargetFilter::Typed(tf)
+            if tf.type_filters.contains(&TypeFilter::Artifact)
+                && tf.type_filters.contains(&TypeFilter::Non(Box::new(TypeFilter::Subtype("Aura".to_string())))))),
+        "artifact conjunct must not also carry non-Aura: {:?}",
+        def.affected
+    );
     // Enchantment conjunct: a non-Aura enchantment you control, mana value >= 4.
     assert!(
         filters.iter().any(|f| matches!(f, TargetFilter::Typed(tf)
@@ -4115,6 +4125,17 @@ fn bello_compound_negated_type_subject_animation_with_granted_abilities() {
                 && tf.controller == Some(ControllerRef::You)
                 && tf.properties.contains(&cmc_ge_4))),
         "enchantment conjunct must be non-Aura enchantment you control, MV>=4: {:?}",
+        def.affected
+    );
+    // Regression guard (issue reported in code review): the enchantment
+    // conjunct must not also inherit the artifact conjunct's non-Equipment
+    // negation via `distribute_neg_type_filters_to_or` treating the two
+    // independently-negated legs as one shared-negation disjunction.
+    assert!(
+        !filters.iter().any(|f| matches!(f, TargetFilter::Typed(tf)
+            if tf.type_filters.contains(&TypeFilter::Enchantment)
+                && tf.type_filters.contains(&TypeFilter::Non(Box::new(TypeFilter::Subtype("Equipment".to_string())))))),
+        "enchantment conjunct must not also carry non-Equipment: {:?}",
         def.affected
     );
 
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -4056,6 +4056,135 @@ fn peel_compound_all_quantified_conjuncts_single_quantifier_declines() {
     assert!(peel_compound_all_quantified_conjuncts("All Mountains are Plains").is_none());
 }
 
+// CR 611.3 + CR 613.4b + CR 205.1b: compound-subject animation whose subject is
+// a heterogeneous union of NEGATED-type legs — Bello, Bard of the Brambles.
+// Distinct from the "all X and all Y" family above (`parse_compound_all_subjects_type_change`
+// / `..._type_replacement`): Bello's subject has a SINGLE leading "each"
+// quantifier and per-conjunct negations ("non-Equipment", "non-Aura"), so it
+// is dispatched through `parse_each_compound_subject_type_change`, which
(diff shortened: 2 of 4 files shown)


matthewevans (changes requested):
**[MED]** The Bello subject filter cross-contaminates one conjunct's negated subtype into the other. Evidence: `crates/engine/src/parser/oracle_static/type_change.rs:1866` delegates the whole subject to `parse_type_phrase`, and the parse-diff sticky for this head reports Bello as `artifact non-Equipment or ... enchantment non-Aura non-Equipment`; Oracle only says `non-Aura enchantment`. The new test at `crates/engine/src/parser/oracle_static/tests.rs:4111` asserts the enchantment leg contains `Non(Aura)` but never asserts it does not also contain `Non(Equipment)`, so this regression is not pinned. Why it matters: Bello would incorrectly exclude any qualifying enchantment that also has the Equipment subtype or otherwise carries that subtype in a type-changing edge case, and the parser output no longer matches the card text. Suggested fix: split/parse the two subject legs so each leg keeps only its own `non-*` qualifier while shared `you control` / mana-value constraints are still distributed, and add negative assertions for both legs.

matthewevans (approved):
Parser change reviewed against current head: prior negated-subtype finding is resolved, parse-diff is limited to Bello, and required checks are green.

PR #5446, opened 2026-07-09
fix(parser): split "and he/she <verb>" gendered-pronoun compound continuations

### LINKED ISSUE #5445: parser: compound "and he/she <verb>" gendered-pronoun continuation strict-fails (Machine Man, Model X-51)
## Summary

A bare gendered pronoun ("he"/"she") used as an Oracle-text subject refers to the card itself (CR 109.3 + CR 201.4b). In a **compound** effect chain, the clause-boundary splitter recognizes the singular anaphor `it <verb>` as a subject-predicate clause start but omits the gendered pronouns, so the second conjunct falls through to the imperative-only path and fails closed to `Effect::Unimplemented { name: "he" }`.

## Repro (verified live parse on current `main`)

**Machine Man, Model X-51** — `Whenever you cast a noncreature spell, put a +1/+1 counter on Machine Man and he gains flying until end of turn.`

Current parse:
- primary: `PutCounter { P1P1, SelfRef }` ✅
- sub_ability: `Unimplemented { name: "he", description: "he gains flying" }` ❌

The equivalent `... and it gains flying ...` parses correctly to a `Flying` grant, and the standalone `he gains flying` also parses correctly (`parse_subject_application` maps "he"/"she" to `SelfRef`). Only the **compound continuation** with a gendered pronoun is dropped.

## Root cause

`starts_bare_and_clause_lower` (`crates/engine/src/parser/oracle_effect/sequence.rs`) lists `it gains`/`it gets`/`it has`/`it loses`/`it doesn't`/`it can't`/`it cannot` as valid clause-boundary starts, but has no `he`/`she` counterpart. Without the split the conjunct never reaches the subject-predicate parser.

## Proposed fix

Add a composed `("he"|"she") × (gains|gets|has|loses|doesn't|can't|cannot)` arm mirroring the singular `it` verb set (not the plural `they` arm, which excludes P/T `get`). This is a parser-only change — no new engine type or runtime; the grant reuses the existing `SelfRef` continuous-modification path.

davion-knight:
Fixes #5445.

## Problem

A bare gendered pronoun ("he"/"she") used as an Oracle-text subject refers to the card itself (CR 109.3 + CR 201.4b). In a compound effect chain the clause-boundary splitter (`starts_bare_and_clause_lower`) recognized the singular anaphor `it <verb>` as a subject-predicate clause start but omitted the gendered pronouns, so a conjunct like **Machine Man, Model X-51**'s

> Whenever you cast a noncreature spell, put a +1/+1 counter on Machine Man **and he gains flying** until end of turn.

fell through to the imperative-only path and failed closed to `Effect::Unimplemented { name: "he" }`. (The `... and it gains flying ...` form and the standalone `he gains flying` both already parse correctly.)

## Fix

Add a composed `("he"|"she") × (gains|gets|has|loses|doesn't|can't|cannot)` arm mirroring the singular `it` verb set — **not** the plural `they` arm, which excludes P/T `get` for its conditional-rider path. The split routes the conjunct through the subject-predicate parser, where `parse_subject_application` already maps "he"/"she" to `SelfRef`. Pronoun × verb are composed rather than enumerated per permutation (per the parser's combinator conventions).

Parser-only: no new engine type, variant, or runtime — the grant reuses the existing `SelfRef` continuous-modification path.

## Verification

- **Live parse (before → after):** Machine Man's `he gains flying` conjunct goes from `Unimplemented { name: "he" }` → `sub_ability` granting `AddKeyword(Flying)` on `SelfRef`, `UntilEndOfTurn` — matching the `it`-form parse. `she gains`, `he gets +1/+1`, `he loses …` variants likewise resolve.
- **No regressions:** full-corpus parser-gap sweep is unchanged at 3977 unsupported cards. The new split only fires on `and he/she <verb>` boundaries, which previously always produced `Unimplemented`, so it cannot alter a currently-working parse.
- **Test:** added `bare_and_clause_starts_on_gendered_pronoun_subjects` next to the existing `it`/`they` boundary tests (positive verb-axis coverage + negative guards for non-continuous tails).
- `cargo check -p engine` + `cargo fmt` clean.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -2539,7 +2539,7 @@ fn starts_bare_and_clause_lower(s: &str) -> bool {
     // subject-predicate clause start — the same guarantee as the "it <verb>" arms
     // above — so split it here. Without the split the conjunct is fed to the
     // imperative-only `parse_imperative_effect`, which has no subject-predicate
-    // path and fails it closed to `Effect::Unimplemented { name: "he" }`.
+    // path and fails it closed to an unimplemented effect named for the pronoun.
     // Splitting routes it through `parse_clause_ast`, where
     // `parse_subject_application` maps "he"/"she" to `SelfRef`
     // (subject.rs `matches!(lower, "he" | "she")`). The pronoun axis is composed
@@ -10728,7 +10728,7 @@ mod tests {
     /// same guarantee as the singular "it <verb>" arms. Machine Man, Model X-51:
     /// "put a +1/+1 counter on ~ and he gains flying until end of turn". Without
     /// the split the conjunct falls to the imperative-only path and fails closed
-    /// to `Effect::Unimplemented { name: "he" }`. The verb set mirrors the
+    /// to an unimplemented effect named for the pronoun. The verb set mirrors the
     /// singular "it" arms (continuous gains/gets/has/loses + restriction
     /// doesn't/can't/cannot), NOT the plural "they" arm (which excludes P/T
     /// "get" for its conditional-rider path).
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -2531,6 +2531,37 @@ fn starts_bare_and_clause_lower(s: &str) -> bool {
         value((), tag("players can't ")),
         value((), tag("players cannot ")),
     )))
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Reviewed current head 5834b010334218c4cf067692faef5172cba9a46c: the parser split is at the existing subject-predicate boundary seam, composes the pronoun and verb axes, keeps negative guards for non-continuous clauses, and the current parse-diff is limited to Machine Man gaining the intended Flying grant while removing the stale duration from the counter effect.

PR #5447, opened 2026-07-09
feat(parser): parse mixed variable+fixed P/T grant (Cranial Ram +X/+1)

nickmopen:
## Problem

`parse_variable_pt_pattern` restricted each P/T axis of a variable grant to `x` or `0`. A **mixed** grant — one dynamic axis and one fixed nonzero axis — failed the pattern outright, so the whole static was dropped.

The clearest example is **Cranial Ram**: `Equipped creature gets +X/+1, where X is the number of artifacts you control.` The `+1` toughness could not be represented, so the equip static produced no P/T modification at all.

## Fix

- `parse_variable_pt_pattern` (grammar.rs) now parses each axis as **either** the variable X (dynamic, returned as `None`) **or** a fixed integer magnitude (`Some(n)`).
- `parse_dynamic_pt_in_text` (anthem.rs) emits a constant `AddPower`/`AddToughness` for a fixed nonzero axis alongside the `AddDynamicPower`/`AddDynamicToughness` for the dynamic axis (CR 613.4c layer 7c).
- A fixed `0` axis still contributes nothing, so `+X/+0`, `-X/-X`, `+0/+X`, etc. are unchanged.

## Tests

- New `dynamic_pt_in_text_mixed_dynamic_power_fixed_toughness` asserts `+X/+1` emits both the X power modification and a constant `AddToughness { value: 1 }` (and that the fixed axis is not dynamic).
- All existing dynamic-P/T tests (`+X/+0`, `-X/-0`, `-X/-X`, `+0/+X`, cost-X defaults) remain green — the fixed-`0` path preserves their behavior.
- Full `parser::oracle_static` module: 1070 passed. `cargo fmt`, clippy (`-D warnings`), and the parser-combinator / engine-authorities gates are clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -1063,8 +1063,11 @@ pub(crate) fn parse_dynamic_pt_in_text(
     let after_verb = nom_tag_lower(after_gets, after_gets, "gets ")
         .or_else(|| nom_tag_lower(after_gets, after_gets, "get "))?;
 
-    // CR 613.4c: Parse variable P/T pattern via nom combinator
-    let (_, (p_sign, p_is_x, t_sign, t_is_x)) = parse_variable_pt_pattern(after_verb).ok()?;
+    // CR 613.4c: Parse variable P/T pattern via nom combinator. Each axis is
+    // either the variable X (`None`) or a fixed magnitude (`Some(n)`).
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Maintainer review clean: the mixed X/fixed P/T parser change is at the shared dynamic-P/T seam, Gemini’s stale map_res finding is fixed at head, parse-diff is scoped to Cranial Ram, and required checks are green.

PR #5452, opened 2026-07-09
feat(parser): anthem subject qualifiers "with a mana ability" / "with no abilities"

nickmopen:
## Problem

An anthem subject phrase carrying a **"with a mana ability"** or **"with no abilities"** object qualifier did not parse. No subject-filter arm recognized the qualifier, so the whole static line fell through to `Effect::Unimplemented { name: "static_structure", .. }` and produced **no continuous static**.

Affected real cards:
- **Raggadragga, Goregutter** — `Each creature you control with a mana ability gets +2/+2.`
- **Muraganda Petroglyphs** — `Creatures with no abilities get +2/+2.`
- **Ruxa, Patient Professor** — `Creatures you control with no abilities get +1/+1.`

## Fix

Strip the trailing qualifier in `parse_continuous_subject_filter`, parse the base subject recursively, and attach the matching `FilterProp`. Both properties are **already fully evaluated at runtime** by `game::filter` (`HasManaAbility` via the authoritative mana-ability classifier; `HasNoAbilities` via `object_has_no_abilities`) — so this is a **grammar-only** seam that reuses existing modeling.

Placing it on the shared subject-filter parser covers every subject shape at once — singular (`Each creature you control …`), global (`Creatures …`), controller-scoped plural (`Creatures you control …`), and subtype (`Elf creatures you control …`). The qualifier must sit at the very end of the subject phrase (`after` empty) so a mid-phrase `with …` clause is not misclaimed.

## Tests

Three new `oracle_tests.rs` tests (Raggadragga / Ruxa / Muraganda) assert each line produces a continuous static — **no `static_structure` Unimplemented** — whose affected filter carries the expected `FilterProp`.

Regression: `parser::oracle` module **7604 passed, 0 failed**. `cargo fmt`, clippy (`-D warnings`), and the parser-combinator / engine-authorities gates are clean. The fix lands entirely in `oracle_static/shared.rs`.

### DIFF
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -3043,9 +3043,13 @@ pub(crate) fn parse_continuous_subject_filter(subject: &str) -> Option<TargetFil
         (" with a mana ability", FilterProp::HasManaAbility),
         (" with no abilities", FilterProp::HasNoAbilities),
     ] {
-        if let Ok((_, (base_lower, after))) = nom_primitives::split_once_on(tp.lower, needle) {
-            if after.trim().is_empty() && !base_lower.trim().is_empty() {
-                let base = tp.original[..base_lower.len()].trim();
+        let parse_trailing_qualifier = all_consuming(terminated(
+            take_until::<_, _, OracleError<'_>>(needle),
+            tag::<_, _, OracleError<'_>>(needle),
+        ));
+        if let Ok((_, base_lower)) = parse_trailing_qualifier.parse(tp.lower) {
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved after maintainer fixups. Evidence: current head 9478a747f6039038c6721f55b4e72a7512f332c6 is CI green; parse-diff refreshed against main 58fa4c710cfa and shows only Raggadragga gaining the intended continuous static and losing static_structure; Gemini's lower/original slicing concern was resolved by switching the qualifier parser to nom suffix parsing plus lower_subslice_to_original.

PR #5478, opened 2026-07-10
fix(phase-ai): clamp scaled policy deltas to the critical band (Closes #5473)

### LINKED ISSUE #5473: phase-ai: LandAnimation / ControlChangeAwareness emit unclamped score deltas that trip PolicyRegistry's critical-band debug_assert
Driving a DuelCommander game with `get_ai_action` (any difficulty tested: Medium, VeryHard) reliably panics a debug-assertions wasm build:

```
panicked at crates/phase-ai/src/policies/registry.rs:407:21:
policy LandAnimation scaled delta -69.9999988079071 exceeds critical band ceiling 15
```

and on a different deck (manlands excluded to rule them out):

```
panicked at crates/phase-ai/src/policies/registry.rs:407:21:
policy ControlChangeAwareness scaled delta -100 exceeds critical band ceiling 15
```

followed by `RuntimeError: unreachable`; the wasm instance is dead.

Where we think it lives: `PolicyRegistry::verdicts` (registry.rs:394-420) multiplies a policy's raw `Score { delta }` by its activation weight and `debug_assert!`s the product against `CRITICAL_MAX = 15.0` (registry.rs:185). At least `land_animation.rs:30` (`TAPPED_LAND_PENALTY = -100.0`) hands back a raw unclamped delta instead of routing through the `score_in_band()` clamp helper other policies use; ControlChangeAwareness appears to do the same. A release build only `tracing::warn!`s (registry.rs:414-420), so this hides in production and hard-kills debug/wasm-dev builds.

Repro shape: two mono-color 99+1 singleton decks (FDN-era commons work), `initialize_game` with the registry's DuelCommander config, both seats driven by `get_ai_action("Medium", seat)`; panics well inside 1,500 steps. We hit it only in Commander-format games so far (our Limited AI-vs-AI smokes never trip it), but the defect is policy scaling, not Commander rules.

Fix we'd expect: clamp (or band-normalize) each policy's contribution at the source, per the existing `score_in_band()` idiom, and keep the debug_assert as the tripwire it was meant to be.

minion1227:
## Summary
Fixes the unclamped-policy-delta crash in `phase-ai` (**Closes #5473**). Driving a Commander game with `get_ai_action` on a debug/wasm build panicked at `registry.rs` — `policy LandAnimation scaled delta -69.9 exceeds critical band ceiling 15` (and the same for `ControlChangeAwareness`), followed by `RuntimeError: unreachable`.

Two compounding defects, fixed at the seam and at the source:

1. **Seam (load-bearing).** `PolicyRegistry::verdicts()` computed `scaled_delta = delta * activation` and only `debug_assert!`d / `tracing::warn!`d on overflow — it never clamped, so in release builds an out-of-band delta leaked straight into the softmax priors, while debug/wasm hard-panicked. The assert was also mis-specified: `activation` legitimately exceeds 1.0 (`arch_times_turn` reaches ~2.6 = archetype_scale 2.0 × late_game_mult 1.3), so even a perfectly band-compliant critical delta of 15 scales to 39 and trips it. The fix re-points the `debug_assert` at the invariant the band helpers actually guarantee — the **pre-scale** `|delta| ≤ CRITICAL_MAX` — and re-bands the amplified product through `PolicyVerdict::score`, the single existing clamp authority. This is identity for in-band contributions and makes every policy safe-by-construction, present and future.

2. **Source.** Five policies hand-built `PolicyVerdict::Score { delta }` literals that bypassed the band helpers and could emit out-of-band magnitudes: `LandAnimation` / `ControlChangeAwareness` (`-100` sentinels — issue-named), `CopyValue` (a `+100`/`±125` analog anchor — **not** issue-named; found by a full-class audit), and `RedundancyAvoidance` / `DownsideAwareness` (unclamped additive totals). Each now routes through a band helper (`critical(-CRITICAL_MAX)` for the two sentinels, `score(..)` for the computed deltas). The two `-100` constants become `-CRITICAL_MAX` — the strongest *finite* discouragement the score contract allows, not a hard `Reject` (giving away a permanent / self-tapping a manland is contextually not always wrong — Donate combos, liability permanents — so it stays overridable).

3. **Recurrence (build for the class).** The `score_contract_lint` meant to catch this had a blind spot: `split("#[cfg(test)]").next()` truncated the production scan at the first `#[cfg(test)]` attribute, which for ~24 policy files is a test-only `use` import in the top import block — hiding their entire impl (exactly why `CopyValue`/`RedundancyAvoidance`/`DownsideAwareness` went undetected). Replaced with a `production_source()` helper that truncates only at the real `#[cfg(test)] mod` boundary, and grandfathered the now-visible legacy counts so no future policy can hide an out-of-band emit behind a test-only import.

## Files changed
- `crates/phase-ai/src/policies/registry.rs` — seam clamp + pre-scale assert + regression test
- `crates/phase-ai/src/policies/control_change_awareness.rs` — route `-100` sentinel through `critical(-CRITICAL_MAX)`
- `crates/phase-ai/src/policies/land_animation.rs` — route `-100` sentinel through `critical(-CRITICAL_MAX)`
- `crates/phase-ai/src/policies/copy_value.rs` — re-band the verdict wrapper through `score()`
- `crates/phase-ai/src/policies/redundancy_avoidance.rs` — re-band the additive total through `score()`
- `crates/phase-ai/src/policies/downside_awareness.rs` — re-band the verdict through `score()`
- `crates/phase-ai/src/policies/tests/score_contract_lint.rs` — fix `#[cfg(test)]` blind spot + grandfather legacy counts

## CR references
None — this is the AI tactical-scoring layer, not MTG game logic. The `PolicyVerdict` band ceiling is a *CR-equivalent* scoring invariant (documented as such in `registry.rs`), not a Comprehensive Rules rule; no CR-annotated game logic is added or changed.

## Track
Developer

## LLM
Model: claude-opus-4-8[1m]
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo clippy -p phase-ai --all-targets` — 0 warnings
- `cargo test -p phase-ai --lib` — 1262 pass / 0 fail (incl. new `scaled_delta_is_clamped_to_critical_band_when_activation_exceeds_one`, `production_source_stops_only_at_the_test_module`, and the tightened `new_policy_files_use_score_contract_helpers`)
- `scripts/ai-gate.sh` — **0 FAIL, 2 WARN, 1 PASS**; all matchups within absolute mirror tolerance. The WARN drift is the expected, benign consequence of clamping deltas that previously leaked out-of-band into release priors (the AI now scores these policies at contract-correct magnitude), and no baseline refresh is required since nothing FAILs:

| matchup | exercises | baseline p0% | current p0% | flips W→L | flips L→W | sign p | status |
|---------|-----------|--------------|-------------|-----------|-----------|--------|--------|
| affinity-mirror | Artifacts, PlusOneCounters, AggroPressure | 20% | 30% | 0 | 0 | — | WARN (avg-turn drift +3.4) |
| enchantress-mirror | Enchantments | 50% | 50% | 0 | 0 | — | PASS |
| red-mirror | AggroPressure | 70% | 80% | 1 | 2 | 0.3125 | WARN (not significant) |

Behavior note: re-banding `CopyValue`'s `+100` "preferred-X" anchor to the band compresses its ranking when a copy target evaluates ≥ 15 (e.g. a 6/6). Argmax-first X-selection is unaffected (it still lands on the smallest preferred X); softmax sampling among only-large targets loses some separation. This is within ai-gate tolerance and is the correct contract-respecting behavior; flagged here for transparency.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/phase-ai/src/policies/control_change_awareness.rs
+++ b/crates/phase-ai/src/policies/control_change_awareness.rs
@@ -23,9 +23,15 @@ use super::context::PolicyContext;
 use super::registry::{DecisionKind, PolicyId, PolicyReason, PolicyVerdict, TacticalPolicy};
 use crate::features::DeckFeatures;
 
-/// Severe penalty for activating an ability that gives away the AI's own permanents.
-/// This should be enough to make PassPriority win over such activations.
-const CONTROL_CHANGE_PENALTY: f64 = -100.0;
+/// Severe penalty for activating an ability that gives away the AI's own
+/// permanents. Sits at the bottom of the critical band (`-CRITICAL_MAX`) — the
+/// strongest finite discouragement the score contract allows, enough for
+/// PassPriority (0) to win, without a hard `Reject` (giving away a permanent is
+/// contextually not always wrong, e.g. Donate combos / liability permanents, so
+/// this stays an overridable penalty rather than a veto). Issue #5473: this was
+/// a raw -100.0 sentinel that bypassed the band helpers and tripped the
+/// registry's critical-band assert once scaled.
+const CONTROL_CHANGE_PENALTY: f64 = -super::registry::CRITICAL_MAX;
 
 pub struct ControlChangeAwarenessPolicy;
 
@@ -166,11 +172,13 @@ fn control_change_verdict(
     }
 
     if gives_away_permanent {
-        // Apply severe penalty for giving away own permanents
-        PolicyVerdict::Score {
-            delta: CONTROL_CHANGE_PENALTY,
-            reason: PolicyReason::new("control_change_gives_away_permanent"),
-        }
+        // Apply severe penalty for giving away own permanents. Route through the
+        // critical() band helper (CR-equivalent score contract) rather than a
+        // raw Score literal so the delta stays clamped to the critical band.
+        PolicyVerdict::critical(
+            CONTROL_CHANGE_PENALTY,
+            PolicyReason::new("control_change_gives_away_permanent"),
+        )
     } else {
         PolicyVerdict::Score {
             delta: 0.0,
--- a/crates/phase-ai/src/policies/control_change_awareness.rs
+++ b/crates/phase-ai/src/policies/control_change_awareness.rs
@@ -31,6 +31,9 @@ use crate::features::DeckFeatures;
 /// this stays an overridable penalty rather than a veto). Issue #5473: this was
 /// a raw -100.0 sentinel that bypassed the band helpers and tripped the
 /// registry's critical-band assert once scaled.
+///
+/// Note: pinned at the critical ceiling, this branch is inert to `activation()`
+/// tuning — `-CRITICAL_MAX × any activation` re-bands back to `-CRITICAL_MAX`.
 const CONTROL_CHANGE_PENALTY: f64 = -super::registry::CRITICAL_MAX;
 
 pub struct ControlChangeAwarenessPolicy;
--- a/crates/phase-ai/src/policies/copy_value.rs
+++ b/crates/phase-ai/src/policies/copy_value.rs
@@ -19,11 +19,19 @@ use crate::features::DeckFeatures;
 
 use super::activation::turn_only;
 use super::context::PolicyContext;
-use super::registry::{DecisionKind, PolicyId, PolicyReason, PolicyVerdict, TacticalPolicy};
+use super::registry::{
+    rescale_into_critical_band, DecisionKind, PolicyId, PolicyReason, PolicyVerdict, TacticalPolicy,
+};
 
 pub struct CopyValuePolicy;
 
 const COPY_SPELL_LOOP_PENALTY_SCALE: f64 = 0.004;
+
+/// Max expected magnitude of `CopyValuePolicy::score()` — the `+100` preferred-X
+/// anchor plus a copy-target evaluation/penalty tail (~±30, `score_target_choice`
+/// / `score_legend_rule_keep`). Sets where `rescale_into_critical_band` starts to
+/// saturate; below it, ordering is preserved.
+const COPY_VALUE_RAW_CEILING: f64 = 130.0;
 
 impl CopyValuePolicy {
     pub fn score(&self, ctx: &PolicyContext<'_>) -> f64 {
@@ -206,13 +214,17 @@ impl TacticalPolicy for CopyValuePolicy {
     }
 
     fn verdict(&self, ctx: &PolicyContext<'_>) -> PolicyVerdict {
-        // `score()` returns a raw analog scale (the +100 "preferred X" anchor and
-        // copy-target penalty sums reach ~±125) that is also consumed raw by
-        // other policies. As a *verdict* it must obey the score contract, so
-        // band-dispatch it through PolicyVerdict::score, which clamps to the
-        // critical band (issue #5473 — the raw literal bypassed the band and
-        // tripped the registry's critical-band assert once scaled).
-        PolicyVerdict::score(self.score(ctx), PolicyReason::new("copy_value_score"))
+        // `score()` is a wide analog signal (the +100 "preferred X" anchor plus
+        // copy-target penalty sums reach ~±125) also consumed raw by other
+        // policies. As a *verdict* it must obey the score contract — but routing
+        // ±125 straight through PolicyVerdict::score would SATURATE, collapsing
+        // every large candidate to the same ±CRITICAL_MAX and flattening the
+        // copy/ChooseX/legend move-ordering prior. Rescale the range into the
+        // band first so ordering survives, then band-dispatch (issue #5473).
+        PolicyVerdict::score(
+            rescale_into_critical_band(self.score(ctx), COPY_VALUE_RAW_CEILING),
+            PolicyReason::new("copy_value_score"),
+        )
     }
 }
 
--- a/crates/phase-ai/src/policies/copy_value.rs
+++ b/crates/phase-ai/src/policies/copy_value.rs
@@ -206,10 +206,13 @@ impl TacticalPolicy for CopyValuePolicy {
     }
 
     fn verdict(&self, ctx: &PolicyContext<'_>) -> PolicyVerdict {
-        PolicyVerdict::Score {
-            delta: self.score(ctx),
-            reason: PolicyReason::new("copy_value_score"),
-        }
+        // `score()` returns a raw analog scale (the +100 "preferred X" anchor and
+        // copy-target penalty sums reach ~±125) that is also consumed raw by
+        // other policies. As a *verdict* it must obey the score contract, so
+        // band-dispatch it through PolicyVerdict::score, which clamps to the
+        // critical band (issue #5473 — the raw literal bypassed the band and
+        // tripped the registry's critical-band assert once scaled).
+        PolicyVerdict::score(self.score(ctx), PolicyReason::new("copy_value_score"))
     }
 }
 
--- a/crates/phase-ai/src/policies/downside_awareness.rs
+++ b/crates/phase-ai/src/policies/downside_awareness.rs
@@ -74,10 +74,14 @@ impl TacticalPolicy for DownsideAwarenessPolicy {
     }
 
     fn verdict(&self, ctx: &PolicyContext<'_>) -> PolicyVerdict {
-        PolicyVerdict::Score {
-            delta: self.score(ctx),
-            reason: PolicyReason::new("downside_awareness_score"),
-        }
+        // `score()` accumulates and doubles per-effect gift penalties, which can
+        // run past the critical band. Band-dispatch through PolicyVerdict::score
+        // so the verdict delta stays clamped to CRITICAL_MAX (issue #5473 — the
+        // raw Score literal bypassed the band contract).
+        PolicyVerdict::score(
+            self.score(ctx),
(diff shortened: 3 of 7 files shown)


matthewevans (changes requested):
Reviewed at head `f4153707`. Paired-seed AI gate, Decision-cost perf gate, and all Rust gates green. Premise verified against issue #5473 — the panic (`policy LandAnimation scaled delta -69.99 exceeds critical band ceiling 15`, then `RuntimeError: unreachable`, dead wasm instance) is real and correctly localized.

**Most of this is right, and the diagnosis is better than the issue's.** The old `debug_assert!(scaled_delta.abs() <= 15)` asserted an invariant the system never held: `arch_times_turn` activation legitimately reaches ~2.6, so even a perfectly band-compliant critical `-15` produces `-39` and panics. Moving the assert onto the **pre-scale** delta (`registry.rs:420`) turns it into a true "did you route through the band helpers?" tripwire, and re-banding the product through `PolicyVerdict::score` (`registry.rs:442`) reuses the one existing clamp authority rather than introducing a second constant that could drift. `scaled_delta_is_clamped_to_critical_band_when_activation_exceeds_one` (`registry.rs:748`) drives the real `PolicyRegistry::verdicts` and flips on revert. The `score_contract_lint` repair is a genuine false-green fix — the old `split("#[cfg(test)]").next()` truncated at the first test-only `use`, which is precisely how the out-of-band literals in `redundancy_avoidance` / `downside_awareness` shipped undetected. Sibling coverage is complete: I checked the other policies and none of them emit a raw out-of-band delta.

One blocking item.

---

### MED — source-side clamping collapses intra-policy ordering; rescale instead of saturate

`crates/phase-ai/src/policies/copy_value.rs:215`

```rust
PolicyVerdict::score(self.score(ctx), PolicyReason::new("copy_value_score"))
```

`copy_value::score()` is an analog signal with a range around ±125. `PolicyVerdict::score` **saturates** rather than rescales — anything above `STRONG_MAX` becomes `signum · min(|x|, CRITICAL_MAX)` (`registry.rs:199`). So:

| copy candidate | raw `score()` | emitted delta |
|---|---|---|
| A | −110 | **−15** |
| B | −30 | **−15** |
| C | −8 | −8 (identity, in band) |

A and B are very different board states and copy_value can no longer distinguish them. Below the band, ordering is preserved exactly; above it, the top of the distribution flattens to a single value. This degrades the softmax move-ordering prior for `SelectTarget` / `ChooseX` copy and legend-adjacent decisions. Alpha-beta still evaluates positions, so it isn't a correctness bug — but it is a silent behavioral regression in release builds, where the full range previously reached the prior, and it is not what issue #5473 asked for.

The repo already has the right idiom for this, one file over. `anti_self_harm.rs:53`:

```rust
const ANTI_SELF_HARM_RAW_CRITICAL_CEILING: f64 = CRITICAL_MAX / 1.3;
```

which maps the raw signal *into* the band before dispatch, so ordering survives.

Please either:
1. Rescale `copy_value::score()`'s analog range into `(STRONG_MAX, CRITICAL_MAX]` so ordering is preserved within the band (matching the `anti_self_harm` self-ceiling pattern) — the preferred fix; the same question applies to the `downside_awareness` doubling and the `redundancy_avoidance` chain sum, so please state whether their ranges can exceed 15 in practice; or
2. Keep the saturation and justify it with a paired-seed report showing move-ordering is unaffected on the affected decision classes.

Either is a small change. I'm not asking for a redesign — just don't let a crash fix quietly flatten a policy's signal.

---

### LOW — activation is now inert at the critical ceiling

A verdict pinned at `±CRITICAL_MAX` re-bands to `±CRITICAL_MAX` for *any* activation (`-15 × 2.6 → -39 → -15`). That follows from a saturating ceiling and is intended, but it means `activation()` tuning has no effect on the critical branch of `control_change_awareness` and `land_animation`. Worth one line of doc so the next person tuning activation doesn't spend an afternoon on it.

---

### Refuting the bot finding

Gemini flags a lint bypass via a trailing comment on the `#[cfg(test)]` line. **Inverted.** If a `#[cfg(test)] // comment` form existed, `production_source`'s `after.lines().find(non-blank)` would see the comment, skip truncation, and treat the *whole file* as production — a superset. That can only **over**-count and fail the lint loudly; it cannot hide a production literal. It fails closed, and no such form exists in the tree. The suggested tweak is a harmless robustness nit, not a correctness issue.

minion1227:
Thanks for the thorough read — addressed all three at head `595e3892`.

### MED — copy_value: rescale instead of saturate ✅

Added a reusable `rescale_into_critical_band(raw, raw_ceiling)` in `registry.rs` — the `anti_self_harm` self-ceiling idea generalized for a signal too wide for a bare clamp. Magnitudes within `STRONG_MAX` pass through unchanged (below-band ordering stays exact); `(STRONG_MAX, raw_ceiling]` maps linearly into `(STRONG_MAX, CRITICAL_MAX]`, so the tail keeps its order instead of colliding at the ceiling. `copy_value` routes `score()` through it with `COPY_VALUE_RAW_CEILING = 130`, then band-dispatches as before. Your case is now order-preserving:

| copy candidate | raw `score()` | emitted delta |
|---|---|---|
| A | −110 | **−13.4** |
| B | −30 | **−7.0** |
| C | −8 | **−5.2** |

`rescale_into_critical_band_preserves_order_where_saturation_collapses` locks it (A < B < C, `|A−B| > 1`, band bounds) and flips on revert.

### The other two ranges (you asked) — both stay in-band, no rescale

- **downside_awareness**: largest gift penalty is `gift_card` −3.0, doubled to −6.0 for pure-downside removal → `score()` never leaves `[−6.0, 0.0]`. `PolicyVerdict::score` is identity here; it only enforces the contract uniformly. Noted in a code comment.
- **redundancy_avoidance**: per-effect penalties are −3.0 / −2.0 / −1.5 / −0.5; real cards carry 1–3 redundant effects so `total` stays in-band. Only a pathological 5+ redundant-effect chain (no printed card) could approach 15, and there every candidate is already "strongly disprefer", so ceiling saturation is harmless. Noted in a comment; kept plain `score()`.

### LOW — activation inert at the ceiling ✅

Added the one-line doc note to both `control_change_awareness` and `land_animation`: a verdict pinned at `−CRITICAL_MAX` re-bands to `−CRITICAL_MAX` for any activation, so `activation()` tuning is inert on that branch.

### Gemini nit ✅ (you're right it fails closed)

Agreed it can only over-count → fail loudly, never hide a literal. Applied the two-line robustness tweak anyway: `production_source` skips to the end of the `#[cfg(test)]` line before scanning for the `mod` boundary, so a trailing comment can't fool it. `production_source_ignores_trailing_comment_on_cfg_test_line` covers it.

---

Verification at `595e3892`: `cargo test -p phase-ai --lib` **1264 pass / 0 fail**; `cargo clippy -p phase-ai --all-targets` **0 warnings**; `scripts/ai-gate.sh` ****0 FAIL · 2 WARN · 1 PASS** (identical drift to the reviewed head — the rescale doesn't move the suite, those mirrors cast no copies)**.

matthewevans (approved):
Approving — the blocking MED from my prior `CHANGES_REQUESTED` (review `4668626037`) is resolved, and I re-verified the fix against head `595e3892`.

**MED resolved (copy_value saturation collapsing intra-policy move-ordering).** `copy_value::verdict` now routes its wide analog signal through the new shared helper before band-dispatch:

```rust
PolicyVerdict::score(
    rescale_into_critical_band(self.score(ctx), COPY_VALUE_RAW_CEILING),
    PolicyReason::new("copy_value_score"),
)
```

`rescale_into_critical_band` (registry.rs) is identity below `STRONG_MAX`, maps `(STRONG_MAX, raw_ceiling]` linearly into `(STRONG_MAX, CRITICAL_MAX]`, saturates only beyond the ceiling, and is sign-correct via `raw.signum()`. I independently verified it is monotonic and order-preserving on the overflow interval.

**Hand-trace of my exact case (ceiling 130):** raw −110 → over = 105/125 = 0.84 → **−13.4**; raw −30 → over = 25/125 = 0.20 → **−7.0**. Both re-band through `PolicyVerdict::score` as identity (magnitudes in `(5,15]`), emitting −13.4 and −7.0 — distinct and correctly ordered. Under the old bare-clamp path both collapsed to −15. Regression fixed.

**Test discriminates.** `rescale_into_critical_band_preserves_order_where_saturation_collapses` asserts `a < b && b < c` and `(a−b).abs() > 1.0` for rescale(−110/−30/−8). Reverting the rescale to a bare clamp yields −15/−15/−8, so `a == b` and both assertions fail — the test would break on revert, not pass vacuously.

**CI:** all required checks green on `595e3892`, including **Paired-seed AI gate** (pass) and **Decision-cost perf gate** (pass) — the designated authorities for AI behavior change.

The pre-scale `debug_assert!(delta.abs() <= CRITICAL_MAX)` on the raw verdict, plus the release-safe re-band of `delta * activation` through the single `PolicyVerdict::score` clamp authority, are the right shape.

One LOW observation, not blocking and not worth a CI re-trigger: activation is now inert for any verdict pinned at `±CRITICAL_MAX` (`−15 × any activation → −15` after re-banding), so `activation()` tuning has no effect on the critical branch of `control_change_awareness` / `land_animation`. Noting it here rather than requesting a comment-only change that would re-run the full AI gate cycle.

Issue #5368, opened 2026-07-08
Client-visible rng_seed in filtered game state lets a participant predict hidden randomness and reconstruct hidden library order


ntindle:
## Summary

`filter_state_for_viewer` redacts hidden zones from the `GameState` delivered to each player, but it does **not** strip the deterministic RNG seed. Every filtered snapshot sent to a client (or P2P guest) still contains `rng_seed`. Because all engine randomness is drawn from a ChaCha20 stream seeded solely by that value, a participant who knows the seed can reproduce the game's entire random stream — future coin flips, die rolls, random discards/targets, and shuffles — and can potentially reconstruct hidden library order that the redaction is meant to protect.

## Affected code

- `crates/engine/src/types/game_state.rs` — `rng_seed: u64` is a serialized field. The live `rng: ChaCha20Rng` stream is `#[serde(skip)]`, but the seed that fully determines it is not.
- `crates/engine/src/game/visibility.rs` — `filter_state_for_viewer(state, viewer)` clones and redacts hands, libraries, face-down identities, choice payloads, and deck pools, but leaves `rng_seed` intact on the returned state.
- Delivery paths that ship the filtered state to untrusted clients:
  - **Server-authoritative:** `crates/phase-server/src/main.rs` broadcasts `filter_state_for_player(&raw_state, pid)` as a `StateUpdate` to each seat.
  - **P2P host-authoritative:** the host's `getViewerSnapshot(pid)` → `engine-wasm` `get_viewer_snapshot_js` runs `filter_state_for_viewer` and sends the result to each guest.

## Impact

All engine randomness draws deterministically from `state.rng`, a ChaCha20 keystream seeded only by `rng_seed`. A participant who knows the seed and can infer how many RNG values have already been consumed can:

- Predict every future randomized outcome (coin flips, die rolls, random targets/discards, shuffles).
- Replay the initial library shuffle to reconstruct the ordering of hidden libraries — defeating the confidentiality that `visibility.rs` otherwise enforces for library contents and order.

This affects **both** multiplayer modes. It is an information-disclosure / competitive-integrity issue: an otherwise-legal but modified client can obtain hidden information and foresee randomness — i.e. cheat. Hidden-zone confidentiality is a security property the engine deliberately implements; exposing the seed undermines it.

## Attack complexity / caveats

[...]

PR #5479, opened 2026-07-10
fix(engine): serialize RNG stream position so restore doesn't rewind it

### LINKED ISSUE #5466: restore_game_state rewinds the RNG stream to origin: serialize stream position for faithful snapshots
## Summary

`GameState.rng` is `#[serde(skip)]` (types/game_state.rs:5932), so
`export_game_state_json` → `restore_game_state` reseeds ChaCha20 from
`rng_seed` at stream position 0 (engine-wasm/src/lib.rs, restore path).
A restored game replays the SAME random sequence the game already
consumed from the start: the next shuffle/coin flip/random discard after
a restore repeats the values that drove the opening shuffle, not the
values that would have come next.

The wasm save-file load path already works around the sibling symptom by
re-rolling a fresh seed on load ("Stale `rng_seed` replays the pre-save
ChaCha20 sequence because stream position is `#[serde(skip)]`" — the
comment in the load path), but that workaround trades one defect for
another: a fresh seed makes load non-deterministic, so a snapshot is
neither a faithful rewind point nor a reproducible one.

## Why it matters downstream

We embed engine-wasm in a single-player campaign game and use
export/restore as the undo primitive (the same pattern as the client's
own single-player UndoButton). At pre-commit scope (takeback inside a
TargetSelection window) this is provably safe because nothing draws RNG
inside the protected span, but it hard-blocks:

- full turn-rewind undo (any resolution between snapshot and restore
  desyncs the future random stream),
- deterministic replays / bug repros built on snapshots,
- save-and-resume that behaves identically to never having saved.

## Proposed fix

`ChaCha20Rng` exposes exact stream position (`get_word_pos` /
`set_word_pos`). Two options:

1. Persist `rng.get_word_pos()` as a serialized `rng_word_pos: u128`
   sibling of `rng_seed`; on deserialize, reseed from `rng_seed` and
   `set_word_pos(rng_word_pos)`. Backward compatible (field defaults to
   0 = today's behavior for old saves).
2. Enable `rand_chacha`'s `serde` feature and serialize the RNG whole,
   dropping the skip. Simpler but bloats the state JSON slightly and
   ties the save format to rand_chacha's serde layout.

Option 1 looks like the minimal, format-stable seam. Happy to PR it
(with a round-trip test: draw N values, export, draw M, restore, assert
the next M values repeat exactly) if the approach is agreed.

davion-knight:
Fixes #5466.

## Problem

`GameState.rng` (ChaCha20) is `#[serde(skip)]`, so `export_game_state_json` → `restore_game_state` reseeded from `rng_seed` at **stream position 0**. A restored snapshot — the single-player undo primitive — replayed the *same* random sequence the game already consumed: the next shuffle / coin flip / random target after a restore repeated the values that drove the opening shuffle, not the values that would have come next.

This hard-blocked faithful full-turn-rewind undo, deterministic replays/bug-repros built on snapshots, and save-and-resume that behaves identically to never having saved. (The MP-resume path already worked around the sibling symptom by re-rolling a fresh seed, trading determinism for non-replay.)

## Fix (issue's Option 1 — minimal, format-stable)

Persist the ChaCha20 stream offset next to the seed:

- **`rng_word_pos: u128`** added to `GameState`, `#[serde(default)]` → pre-existing saves default to `0` (today's rewind-to-origin behavior; **backward compatible**). Excluded from the manual `PartialEq` like `rng` itself (mutable stream position, not identity).
- **`export_game_state_json`** syncs `rng_word_pos = rng.get_word_pos()` before serializing (switched to `with_state_mut`).
- **`restore_game_state`** re-applies `rng.set_word_pos(rng_word_pos)` after reseeding, so the stream resumes where it left off.
- **`filter_state_for_viewer`** zeroes `rng_word_pos` alongside `rng_seed` — no client/guest snapshot leaks the stream position (#5368 sibling; a leaked offset would hand an attacker the keystream alignment).
- **`resume_multiplayer_host_state`** resets it to `0` to match its deliberate fresh-seed re-roll (MP resume intentionally diverges; stale comment corrected).

## Verification

- `cargo check -p engine` and `-p engine-wasm` clean; `cargo check -p engine --tests` (test module) type-checks; `cargo fmt` clean.
- **Round-trip test** (`rng_word_pos_survives_serde_and_restores_stream_position`): draw 7 values, sync + serialize, restore (reseed + `set_word_pos`), assert the next 5 values reproduce a reference stream at the same offset exactly.
- **Backward-compat test** (`rng_word_pos_defaults_to_zero_when_absent`): a snapshot with the field removed deserializes to `0`.

3 files, +~72/-1. No new engine variant or runtime; a `#[serde(default)]` field is a wire-compatible addition (old saves load unchanged).

### DIFF
--- a/crates/engine-wasm/src/lib.rs
+++ b/crates/engine-wasm/src/lib.rs
@@ -2388,3 +2388,65 @@ mod replay_bridge_tests {
         clear_game_state();
     }
 }
+
+/// Native coverage for the RNG-restore bridge wiring (issue #5466). The
+/// `export`/`restore` entry points are plain Rust functions, so these run in
+/// the standard `cargo test`/`nextest` shards — unlike the `wasm32`-gated
+/// `mod tests`, whose assertions never execute in the native suite.
+#[cfg(test)]
+mod rng_restore_bridge_tests {
+    use super::*;
+    use rand::RngCore;
+
+    #[test]
+    fn export_then_restore_resumes_live_rng_stream_through_wasm_bridge() {
+        // Issue #5466, end-to-end through the WASM boundary: `export_game_state_json`
+        // must capture the live ChaCha20 offset and `restore_game_state` must
+        // fast-forward the reseeded stream to it, so a restored game draws the
+        // values that would have come NEXT — not a replay from origin. This test
+        // drives the real bridge entry points (nothing calls the engine seam
+        // directly): deleting `state.capture_rng_word_pos()` in export or
+        // `state.rehydrate_rng()` in restore turns it red. Asserts on consumed
+        // randomness, not the stored `rng_word_pos` integer.
+        clear_game_state();
+
+        // Seed a live game and consume randomness as gameplay would.
+        let mut state = GameState::new_two_player(0x51A7_C0DE);
+        for _ in 0..9 {
+            state.rng.next_u32();
+        }
+        GAME_STATE.with(|cell| cell.set(Some(state)));
+
+        // The four values the live stream will produce next, captured from a
+        // clone taken at the pre-export offset.
+        let mut expected = with_state(|s| s.rng.clone()).unwrap();
+        let expected_draws: Vec<u32> = (0..4).map(|_| expected.next_u32()).collect();
+
+        // Serialize through the real bridge entry point (captures rng_word_pos).
+        let json = export_game_state_json().unwrap();
+
+        // Advance the LIVE rng further so a rewind-to-origin restore would be
+        // observable: without the offset, restore would replay from the seed's
+        // origin (nine draws back), never matching `expected_draws`.
+        with_state_mut(|s| {
+            for _ in 0..3 {
+                s.rng.next_u32();
+            }
+        })
+        .unwrap();
+
+        // Restore through the real bridge entry point (reseeds + fast-forwards).
+        restore_game_state(&json).unwrap();
+
+        // The restored stream must resume at the exported offset and produce the
+        // continuation captured before export.
+        let restored_draws: Vec<u32> =
+            with_state_mut(|s| (0..4).map(|_| s.rng.next_u32()).collect()).unwrap();
+        assert_eq!(
+            restored_draws, expected_draws,
+            "restored stream must resume where export left off, not rewind to origin"
+        );
+
+        clear_game_state();
+    }
+}
--- a/crates/engine-wasm/src/lib.rs
+++ b/crates/engine-wasm/src/lib.rs
@@ -1257,7 +1257,11 @@ pub fn list_token_presets_js() -> JsValue {
 /// Used by the engine worker to transfer state to AI workers for root parallelism.
 #[wasm_bindgen]
 pub fn export_game_state_json() -> Result<String, JsValue> {
-    with_state(|state| {
+    with_state_mut(|state| {
+        // Capture the live ChaCha20 stream position so `restore_game_state` can
+        // fast-forward to it (issue #5466): `rng` is `#[serde(skip)]`, so only
+        // the serialized `rng_word_pos` carries the offset across the round trip.
+        state.rng_word_pos = state.rng.get_word_pos();
         serde_json::to_string(state)
             .map_err(|e| JsValue::from_str(&format!("Failed to serialize GameState: {e}")))
     })?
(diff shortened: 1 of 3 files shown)


matthewevans:
Holding this rather than merging it, and I want to be straight with you about why.

**A competing PR landed nine minutes after yours.** #5481 fixes the same issue (#5466) with a different design, and after reading both heads I think #5481's shape is the one we should take. That is not a judgment on your diagnosis, which is correct and well-explained.

Two things you handled that #5481 had to be checked for, and got right:

- You explicitly redacted the leaked stream position in `visibility.rs:36` (`filtered.rng_word_pos = 0`). Serializing an RNG stream position into per-viewer state would let an opponent predict shuffles and coin flips. You saw that.
- You kept `rng_word_pos` out of `impl PartialEq for GameState`, so CR 104.4b draw-by-repetition loop detection still works. An ever-advancing counter in that comparison would have quietly broken it.

Where #5481 differs: it derives the serialized position from the live `rng` at serialize time (custom `serialize_with`) and centralizes reload in an engine-owned `GameState::rehydrate_rng()`. Two consequences. First, there is no second source of truth to drift — your `state.rng_word_pos = state.rng.get_word_pos()` at `engine-wasm/src/lib.rs:1264` has to be remembered by every future serialize path. Second, because the position rides on the live `rng`, the pre-existing `filtered.rng = seed_from_u64(0)` redaction in `visibility.rs:33` zeroes it for free, so the leak you correctly patched becomes structurally impossible rather than patched. CLAUDE.md treats the WASM bridge as a thin serialization boundary that shouldn't own game-state logic, and that's the line this falls on.

**One thing to take forward regardless of which PR lands.** Your test doesn't test the fix:

```rust
// crates/engine/src/types/game_state.rs:10199, 10204-10205
state.rng_word_pos = state.rng.get_word_pos();          // hand-written export sync
let mut restored: GameState = serde_json::from_str(&json).unwrap();
restored.rng = ChaCha20Rng::seed_from_u64(restored.rng_seed);
restored.rng.set_word_pos(restored.rng_word_pos);        // hand-written restore
```

It never calls `export_game_state_json()` or `restore_game_state()`. It re-implements the seam inline, so it proves `rand_chacha`'s `set_word_pos` works — not that production calls it. Delete the real fix at `lib.rs:1291` and this test still passes, as does the rest of the suite (no `engine-wasm` test touches `rng_word_pos`). A regression test has to fail when you revert the change it guards; otherwise it's an assertion about a dependency.

I'm leaving this open until #5481's CI actually runs — it's a first-time-contributor fork PR and its workflows are still withheld, so nothing about it is verified yet. If it doesn't hold up, we come back to yours. Genuinely good catch on a subtle determinism bug.

davion-knight:
Thanks — that's a fair and well-reasoned call, and both critiques are correct.

I've pushed a revision that addresses the actionable feedback regardless of which PR lands:

- **The test now guards the real fix.** It calls the production methods rather than re-implementing the seam inline, so reverting the logic fails it instead of just asserting `rand_chacha` behavior.
- **Logic moved out of the WASM bridge.** Introduced engine-owned `GameState::capture_rng_word_pos()` (export) and `GameState::rehydrate_rng()` (restore); `export_game_state_json` / `restore_game_state` now just delegate, keeping the bridge a thin serialization boundary per CLAUDE.md.

One note in case it's useful for reviewing #5481's test too: I tried to add a bridge-level round-trip test in `engine-wasm`'s `#[cfg(all(test, target_arch = "wasm32"))]` module, but that module currently has pre-existing wasm32 compile failures (`evaluate_layers`/`derive_display_state` and stale `Effect`/`CardFace` fixtures), and CI's `wasm-check` only builds the lib (`--target wasm32-unknown-unknown`, no `--tests`) — so nothing in that module actually compiles or runs in CI today. That's why I landed the guard as a native engine-method test on the seam the bridge delegates to.

Fully agree #5481's `serialize_with`-derived-from-live-`rng` is the stronger shape (no second source of truth, and the redaction falls out of the existing `seed_from_u64(0)` for free). Happy for that to be the one that lands — I'm keeping this current only as the fallback you mentioned, and this revision should make it a clean one if #5481's CI doesn't hold up. Appreciate the thorough read either way.

matthewevans (approved):
Approving and enqueuing.

Thanks for the rewrite — it lands the design in the right place. Moving the snapshot/rehydrate logic into the engine (`GameState::capture_rng_word_pos` / `GameState::rehydrate_rng`) and having the WASM boundary merely delegate resolves the transport-owns-logic objection cleanly, and reworking the engine test to drive those real methods (rather than an inline re-implementation) fixes the earlier false-green: reverting either method now breaks `rng_word_pos_survives_serde_and_rehydrate_resumes_stream`. Keeping `rng_word_pos` as `#[serde(default)]` plain data with `rng` still `#[serde(skip, default)]` also avoids any placeholder-seed hazard, and the field is correctly excluded from both `impl PartialEq for GameState` and `loop_fingerprint`, so CR 104.4b draw-by-repetition loop detection is untouched.

#5481 (moorsecopers99) proposed a competing design for the same issue (#5466). Its one structural edge — redacting the stream position in `filter_state_for_viewer` regardless of caller — is now covered here by an explicit discriminating test I added to `redacts_rng_seed_from_every_viewer` (asserts `filtered.rng_word_pos == 0` for every seat viewer and the non-seat spectator, with a non-zero source position), so this PR no longer trails on that axis.

Maintainer fixups pushed on your branch: two discriminating tests closing the coverage gaps — the `visibility.rs` redaction assertion above, and an end-to-end WASM-bridge test (`export_then_restore_resumes_live_rng_stream_through_wasm_bridge`) that exercises the real `export_game_state_json()` / `restore_game_state()` entrypoints and asserts on consumed randomness (the restored stream resumes at the exported offset rather than rewinding to origin). Deleting `capture_rng_word_pos()` in export or `rehydrate_rng()` in restore turns that test red.

PR #5480, opened 2026-07-10
feat(card): implement Diligent Farmhand (CountsAsNamed graveyard static)

Whovencroft:
## Summary

Implements **Diligent Farmhand** (Odyssey) — specifically the novel `CountsAsNamed` graveyard static mechanic used by the Odyssey Burst cycle.

### Card Oracle Text

> {1}{G}, Sacrifice Diligent Farmhand: Search your library for a basic land card, put that card onto the battlefield tapped, then shuffle.
>
> If Diligent Farmhand is in a graveyard, effects from spells named Muscle Burst count it as a card named Muscle Burst.

### What This PR Adds

| Component | Change |
|-----------|--------|
| `statics.rs` | New `StaticMode::CountsAsNamed { name: String }` variant with all required match arms (kind, Hash, Display, FromStr, keyword_identity) |
| `filter.rs` | `FilterProp::Named` evaluation extended to check `CountsAsNamed` statics on the object, respecting `active_zones` (alias only applies when the object is in the graveyard) |
| `oracle_static/dispatch.rs` | `try_parse_counts_as_named_static()` parser for the "If ~ is in a graveyard, effects from spells named X count it as a card named X" pattern |
| `oracle_static/mod.rs` | Exports the new parser function |
| `oracle.rs` | Calls the new parser in `parse_static_line_with_graveyard_keyword_continuation` |
| Integration test | `diligent_farmhand_counts_as_named.rs` — verifies Muscle Burst's +X/+X includes Farmhand copies in graveyard |

### Design Decisions

- **Zone-gated alias**: The `CountsAsNamed` static uses `active_zones = [Graveyard]` so the name alias only applies while the object is in the graveyard (CR 201.5).
- **Filter-level integration**: Rather than modifying the spell resolution path, the alias is checked at the `FilterProp::Named` evaluation level. This means any effect that counts "cards named X" (ObjectCount, targeted filters, etc.) automatically respects the alias without per-effect changes.
- **Whole-cycle support**: The parser handles the full Odyssey Burst cycle pattern (Diligent Farmhand, Pardic Firecat, Cephalid Scout, etc.) — any card with this oracle text template will parse correctly.
- **Activated ability**: The sacrifice-to-search-basic-land ability is already supported by the existing Sakura-Tribe Elder parser path (no new code needed).

### Testing

- `cargo check --lib -p engine` passes locally
- Integration test: Two Diligent Farmhands in graveyard + Muscle Burst cast → target gets +5/+5 (base 3 + 2 Farmhands counted as Muscle Burst)

### Coverage Gap Closed

- `effect_structure` gap for "If ~ is in a graveyard, effects from spells named Muscle Burst count it as a card named Muscle Burst."

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -27,6 +27,7 @@ use crate::types::keywords::Keyword;
 use crate::types::mana::{ManaColor, ManaCost};
 use crate::types::player::PlayerId;
 use crate::types::proposed_event::{EtbTapState, ProposedEvent, TokenSpec};
+use crate::types::statics::StaticMode;
 use crate::types::zones::Zone;
 
 /// True when the filter's matched SET depends on the population of objects on
@@ -3644,7 +3645,20 @@ fn matches_filter_prop(
         // no activated, triggered, replacement, or static abilities.
         FilterProp::HasNoAbilities => object_has_no_abilities(obj),
         // CR 201.2: Name matching is exact (case-insensitive comparison).
-        FilterProp::Named { name } => obj.name.eq_ignore_ascii_case(name),
+        // CR 201.5: Also check CountsAsNamed statics (Odyssey Burst cycle).
+        FilterProp::Named { name } => {
+            obj.name.eq_ignore_ascii_case(name)
+                || obj.static_definitions.iter_all().any(|sd| {
+                    // Only count the alias if the static's active_zones include
+                    // the object's current zone (or active_zones is empty = always).
+                    if let StaticMode::CountsAsNamed { name: alias } = &sd.mode {
+                        (sd.active_zones.is_empty() || sd.active_zones.contains(&obj.zone))
+                            && alias.eq_ignore_ascii_case(name)
+                    } else {
+                        false
+                    }
+                })
+        }
         // SameName: matches objects with the same name as the tracked card from context.
         // At runtime, this checks against the source object's name (the event context card).
         FilterProp::SameName => {
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -3644,8 +3644,13 @@ fn matches_filter_prop(
         // CR 113.1 + CR 113.3: "no abilities" means no keyword abilities and
         // no activated, triggered, replacement, or static abilities.
         FilterProp::HasNoAbilities => object_has_no_abilities(obj),
-        // CR 201.2: Name matching is exact (case-insensitive comparison).
-        // CR 201.5: Also check CountsAsNamed statics (Odyssey Burst cycle).
+        // CR 201.2a: Name matching is exact (case-insensitive comparison).
+        // Also check CountsAsNamed statics (Odyssey Burst cycle).
+        // NOTE: The alias applies to ANY FilterProp::Named check, not only effects
+        // from spells named X. This is safe for the entire Burst cycle because each
+        // Burst spell is the only effect referencing its own name — but if a future
+        // non-self-referential "counts as named" card appears, this may need a
+        // source-filter check.
         FilterProp::Named { name } => {
             obj.name.eq_ignore_ascii_case(name)
                 || obj.static_definitions.iter_all().any(|sd| {
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -910,6 +910,8 @@ fn parse_static_line_with_graveyard_keyword_continuation(
         vec![def]
     } else if let Some(def) = try_parse_graveyard_keyword_grant_static(line) {
         vec![def]
+    } else if let Some(def) = crate::parser::oracle_static::try_parse_counts_as_named_static(line) {
+        vec![def]
     } else {
         parse_static_line_multi(line)
     };
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -3430,3 +3430,55 @@ fn parse_enters_with_additional_counters(
         .description(text.to_string()),
     )
 }
+
+/// Parses the Odyssey Burst-cycle graveyard name-aliasing static:
+/// "If ~ is in a graveyard, effects from spells named X count it as
+/// a card named X." — Odyssey Burst cycle (Diligent Farmhand, Pardic Firecat,
+/// Flame-Kin Zealot, etc.).
+///
+/// Returns a `StaticDefinition` with `StaticMode::CountsAsNamed { name }` and
+/// `active_zones = [Graveyard]`.
+pub(crate) fn try_parse_counts_as_named_static(text: &str) -> Option<StaticDefinition> {
+    use super::prelude::*;
+    // Normalize: lowercase for matching, strip trailing period.
+    let lower = text.to_lowercase();
+    // allow-noncombinator: punctuation cleanup on pre-tokenized input (char arg)
+    let lower = lower.trim_end_matches('.').trim();
+    // Parse the fixed structure using nom combinators:
+    // "if ~ is in a graveyard, effects from spells named <NAME> count it as a card named <NAME>"
+    let (rest, _) = tag::<_, _, OracleError<'_>>("if ~ is in a graveyard, ")
+        .parse(lower)
+        .ok()?;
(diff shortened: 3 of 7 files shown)


matthewevans (changes requested):
Reviewed at head `ab2c9012`. Premise verified against Scryfall — the card really does say *"If this card is in a graveyard, effects from spells named Muscle Burst count it as a card named Muscle Burst."* Good pick, and the design instincts here are right.

**What's right.** Routing this through `FilterProp::Named` match-time comparison scoped to `active_zones == [Graveyard]` is the correct seam. The obvious alternative, `ContinuousModification::SetName`, is a Layer 1 copy effect that would rename the object *for all purposes* — rules-wrong, since Diligent Farmhand isn't named Muscle Burst for anything except Muscle-Burst-effect counting. `CountsAsNamed` is genuinely orthogonal to the existing `StaticMode` variants (no sibling-cluster smell), the exhaustive matches are all threaded, and the parser hunk uses real nom combinators rather than string dispatch.

Two blocking findings.

---

### HIGH — the fix never fires for the real card

`crates/engine/src/parser/oracle_static/dispatch.rs`

```rust
let (rest, _) = tag::<_, _, OracleError<'_>>("if ~ is in a graveyard, ")
```

The real Oracle text is `"If this card is in a graveyard, …"`. `"this card"` is **never normalized to `~`**. From `oracle_util.rs:894`:

```rust
/// NOT used by: `normalize_card_name_refs` (must not replace these with `~`).
pub const SELF_REF_PARSE_ONLY_PHRASES: &[&str] = &["this spell", "this card"];
```

`normalize_card_name_refs` (`oracle_util.rs:2210`) iterates only `SELF_REF_TYPE_PHRASES`, which contains `"this creature"`, `"this permanent"`, etc. — not `"this card"`. So this `tag()` cannot match the card it was written for.

Three independent confirmations:

1. **`card-data.json` on `main`** parses Diligent Farmhand's second ability as `{"kind": "Spell", "effect": {"type": "Unimplemented"}}`.
2. **Your own parse-diff sticky** says `✓ No card-parse changes detected` — and it is current (posted 05:06:48Z, after the Card data job completed 05:06:32Z on this head). Diligent Farmhand *is* in the pool (I checked; 35,396 cards). If the arm fired, that ability would flip from `Unimplemented` to a `CountsAsNamed` static and the sticky would show it.
3. **The test constant pre-normalizes by hand:**
   ```rust
   const FARMHAND_ORACLE: &str = "{1}{G}, Sacrifice ~: … If ~ is in a graveyard, …";
   ```
   The `~` is written into the fixture. The pipeline never produces it from `"this card"`, so `.from_oracle_text(FARMHAND_ORACLE)` exercises a string the real card never becomes.

The integration test is otherwise well built — it drives a real `cast().target_object().resolve()` and would fail on revert *for its fixture*. But the fixture diverges from production input on exactly the axis the fix depends on, so green CI proves nothing about the card. After this PR, Diligent Farmhand still parses to `Unimplemented`.

**Fix:** make the combinator accept the self-reference forms the pipeline actually emits. `"this card"` reaches the parser verbatim, so match it alongside `~` — an `alt((tag("if ~ is in a graveyard, "), tag("if this card is in a graveyard, ")))`, or better, factor a small self-reference combinator so the rest of the Burst cycle (Pardic Firecat, Nantuko Blightcutter, Krosan Beast, …) doesn't each rediscover this. Then confirm the parse-diff sticky shows Diligent Farmhand gaining the `CountsAsNamed` signature and losing `Unimplemented`. That sticky flipping is the acceptance criterion here.

---

### HIGH — `CR 201.5` is the wrong rule, cited seven times

`statics.rs:1874-1877`, `:2354`, `:2891`, `filter.rs:3647`, the module doc, and the test header all cite `CR 201.5`. Grepping `docs/MagicCompRules.txt:1322`:

> **201.5.** Text that refers to the object it's on by name means just that particular object and not any other objects with that name, **regardless of any name changes caused by game effects.**

That is self-reference *exclusivity* — close to the inverse of what this code implements. `201.5a/b/c` are about granted abilities and shortened names; none is about one object counting as a differently-named card.

The nearest applicable rule is `CR 201.2a` (`docs/MagicCompRules.txt:1290`): *"Two or more objects have the same name if they have at least one name in common…"* — which is the name-matching semantics `FilterProp::Named` implements. Honestly, "counts as a card named X" is card-specific Odyssey templating with no clean general CR, and saying so is better than citing a rule that says something else.

Please either cite `CR 201.2a` for the name-matching behavior, or drop the number and note that no general CR governs this templating. A wrong CR number is worse than none — it creates false confidence the code was validated against the right rule, which is exactly what happened here.

---

### LOW — the alias ignores the effect's source

The Oracle text restricts the aliasing to *"effects from spells named Muscle Burst."* `filter.rs:3648` counts the object for **any** `FilterProp::Named { name: "Muscle Burst" }` check, whatever the source. Harmless across the whole Burst cycle, because each Burst spell is the only effect referencing its own name — but worth a comment so a future non-self-referential "counts as named" card doesn't silently inherit an over-broad match.

### LOW — alias name reconstructed by title-casing

`dispatch.rs` lowercases the line, then rebuilds the stored name by capitalizing each word. Match-time correctness survives via `eq_ignore_ascii_case`, but the `Display`/`FromStr` round-trip and any name display carry the reconstruction, which diverges for hyphens, articles, and accents. Slicing the alias out of the original-case input using the matched offsets would avoid it.

---

Also please fill in the **Files changed**, **Track**, **LLM**, and **Verification** sections of the PR template — see `docs/AI-CONTRIBUTOR.md`.

The seam choice and the variant design here are good work. Get the parser arm to match the text the pipeline actually produces, fix the CR citation, and confirm the sticky flips — then this is a clean landing.

matthewevans (approved):
## APPROVE — supersedes my earlier `CHANGES_REQUESTED` (review `4668715165`)

Both HIGH findings from the prior round are resolved on head `d97e8d1a`.

**HIGH #1 (inert fix) — resolved.** The dispatch arm now accepts the real production Oracle form. `oracle_static/dispatch.rs`:

```rust
let (prefix_rest, _) = alt((
    tag::<_, _, OracleError<'_>>("if ~ is in a graveyard, "),
    tag::<_, _, OracleError<'_>>("if this card is in a graveyard, "),
))
.parse(lower.as_str())
.ok()?;
```

`"this card"` lives in `SELF_REF_PARSE_ONLY_PHRASES` and is never normalized to `~`, so the second `tag` is the one that actually fires on real card data. Confirmed by the parse-diff sticky: **2 cards flip** — Diligent Farmhand and Pardic Firecat each REMOVE `effect_structure` (the `Unimplemented` node) and ADD `CountsAsNamed(Muscle Burst)` / `CountsAsNamed(Flame Burst)` respectively (baseline main `02bb03df`). The fix reaches the real cards through the production pipeline and the class generalizes to the whole Odyssey Burst cycle.

**HIGH #2 (wrong CR ×7) — resolved.** `statics.rs` and `filter.rs` now cite `CR 201.2a` ("Two or more objects have the same name if they have at least one name in common") — the correct name-matching rule. The one surviving `CR 201.5` in `dispatch.rs` documents `parse_self_reference_subject` consuming a `~` self-reference subject, which is exactly what 201.5 governs — correct usage, left intact.

**Seam is right.** `CountsAsNamed` is consumed at comparison time by `FilterProp::Named` in `game/filter.rs`, gated by `active_zones.contains(&obj.zone)` (scoped to Graveyard) — a name-*matching* check, not a Layer 1 `SetName` rename that would alias the object for all purposes. Rules-correct per CR 201.2a.

**Test discriminates.** `diligent_farmhand_counts_as_named.rs` drives the real cast pipeline with the production "this card" Oracle text: two Farmhands in the graveyard, cast Muscle Burst, assert target 1/1 → 6/6 (X = 3 + 2). On revert of either the parser arm or the filter seam, the Farmhands stop counting, X = 3, target = 4/4, and the assertion fails.

Two residual LOWs — noted, not blocking, do not fix in this PR:
1. `name_start`/`name_end` are lowercased byte offsets sliced into original-case `stripped`; a char whose lowercase differs in byte length would slice off a boundary. Unreachable for the ASCII Burst-cycle names.
2. The alias matches any `FilterProp::Named` check regardless of the counting effect's source, while the Oracle text restricts it to "effects from spells named X". Harmless across the self-referential Burst cycle where each Burst spell is the only effect referencing its own name. (The code comment already flags this.)

PR #5481, opened 2026-07-10
fix(engine): serialize RNG stream position for faithful snapshot restore

### LINKED ISSUE #5466: restore_game_state rewinds the RNG stream to origin: serialize stream position for faithful snapshots

moorsecopers99:
## Summary

`GameState.rng` was `#[serde(skip)]`, so `export_game_state_json` →
`restore_game_state` reseeded ChaCha20 from `rng_seed` at stream position 0. A
restored snapshot therefore replayed the *same* random sequence the game had
already consumed since the opening shuffle — the next shuffle/coin-flip/random
discard repeated the opening-shuffle values instead of continuing the stream.
This serializes the exact stream position (`rng_word_pos`) alongside `rng_seed`
so a restore resumes the sequence exactly where the snapshot left off.

Closes #5466.

> This is save/restore **serialization infrastructure**, not engine game logic.
> It touches no parser, effect, resolver, targeting, or rules behavior — the RNG
> still produces the identical value sequence; the change only persists *where in
> that sequence* a saved game sits so a load doesn't rewind to origin. Inside
> `crates/engine/` only `types/game_state.rs` changes (serde plumbing + a
> reconstruction helper), with no CR-governed behavior change.

## CR references

None. No rules behavior changes — this is state-serialization plumbing; the RNG's
draw sequence is unchanged, only its persisted stream position is added.

## Verification

- `cargo fmt --all` — clean.
- `cargo test -p engine --lib` — **15,956 passed, 0 failed**.
- `cargo clippy -p engine -p engine-wasm -p server-core --lib --tests -- -D warnings` — clean.
- `cargo check -p engine-wasm --target wasm32-unknown-unknown` — clean (production build).
- New round-trip tests in `crates/engine/src/types/game_state.rs`:
  - `rehydrate_rng_resumes_stream_position_after_roundtrip` — the exact test the
    issue asked for: draw N values, serialize, record the next M, restore, assert
    the next M repeat exactly.
  - `rehydrate_rng_defaults_to_origin_for_legacy_saves` — a pre-#5466 save (no
    `rng_word_pos` key) lands at position 0, reproducing today's behavior.

## Design

Option 1 from the issue (persist `get_word_pos()` — the minimal, format-stable
seam), built as a reusable building block rather than a one-off:

- `rng` now serializes **only its live stream position** (`get_word_pos()` →
  `u128`) under a new `rng_word_pos` key, via `serialize_with` /
  `deserialize_with`. Reading the position off the *live* `rng` at serialize time
  keeps it always fresh — there is no companion field to keep in sync.
- `GameState::rehydrate_rng()` is the single reconstruction authority: reseed from
  `rng_seed`, then re-apply the saved position. `restore_game_state` (the undo
  primitive) now calls it instead of hand-reseeding.
- **Backward compatible:** pre-#5466 saves omit the key → `default_rng` (position
  0) → identical to today's rewind-to-origin behavior. No `deny_unknown_fields`,
  so it is forward-compatible too.
- The deliberate fresh-seed resume paths (`resume_multiplayer_host_state`,
  server-core `from_persisted`) are left **behaviorally unchanged** — they re-roll
  on purpose so a resumed multiplayer game diverges from the snapshot. Their now
  stale comments (which cited the `#[serde(skip)]` position loss as the reason)
  are corrected to state the re-roll is a deliberate choice.

## Anchored on

- `crates/engine/src/types/game_state.rs` — `serialize_rng_word_pos` /
  `deserialize_rng_word_pos`, the `rng` field's serde attributes, and
  `GameState::rehydrate_rng()`.
- `crates/engine-wasm/src/lib.rs` — `restore_game_state` resumes the stream via
  `rehydrate_rng()`; corrected `resume_multiplayer_host_state` comments.
- `crates/server-core/src/session.rs` — comment-only consistency update in
  `from_persisted` (behavior unchanged).

## Scope Expansion

None.

### DIFF
--- a/crates/engine-wasm/src/lib.rs
+++ b/crates/engine-wasm/src/lib.rs
@@ -1279,7 +1279,11 @@ pub fn restore_game_state(json_str: &str) -> Result<(), JsValue> {
     }
     let mut state: GameState = serde_json::from_str(json_str)
         .map_err(|e| JsValue::from_str(&format!("Failed to deserialize GameState: {}", e)))?;
-    state.rng = ChaCha20Rng::seed_from_u64(state.rng_seed);
+    // Undo/restore must resume the *same* random stream: reseed from
+    // `rng_seed` and re-apply the persisted `rng_word_pos` so the next
+    // shuffle/coin-flip/random-discard continues where the snapshot left off,
+    // instead of replaying the opening shuffle's values (issue #5466).
+    state.rehydrate_rng();
     state.debug_mode = true;
     CARD_DB.with(|cell| {
         if let Some(db) = cell.borrow().as_ref() {
@@ -1305,11 +1309,14 @@ pub fn restore_game_state(json_str: &str) -> Result<(), JsValue> {
 ///
 /// Differs from `restore_game_state` in two load-bearing ways:
 ///
-/// 1. **Fresh RNG seed.** `restore_game_state` re-seeds from the saved
-///    `rng_seed`, which rewinds the ChaCha20 stream to position 0 —
-///    correct for undo (replay from origin) but wrong for resume
-///    (subsequent draws would replay the pre-save sequence). This
-///    function stamps a fresh seed so continued play diverges.
+/// 1. **Fresh RNG seed.** `restore_game_state` faithfully resumes the saved
+///    ChaCha20 stream (seed *and* `rng_word_pos`) — the right semantics for
+///    undo, which must continue exactly where the snapshot left off. Resume
+///    deliberately does not: a crash-recovered host stamps a fresh seed so the
+///    resumed game's future draws are independent of the persisted snapshot.
+///    Faithful same-stream resume is now *possible* (the stream position is
+///    serialized as of #5466); re-rolling here is a deliberate choice, not a
+///    workaround for lost stream state.
 /// 2. **Atomic multiplayer-flag flip.** Sets `MULTIPLAYER_MODE` in the
 ///    same call that loads state, so there's no window where a stray
 ///    `restore_game_state` (undo) would be accepted on the resumed
@@ -1339,8 +1346,10 @@ pub fn resume_multiplayer_host_state(json_str: &str) -> Result<(), JsValue> {
     let mut state: GameState = serde_json::from_str(json_str)
         .map_err(|e| JsValue::from_str(&format!("Failed to deserialize GameState: {}", e)))?;
 
-    // Stale `rng_seed` replays the pre-save ChaCha20 sequence because
-    // stream position is `#[serde(skip)]`. Mirrors server-core.
+    // Deliberately re-roll rather than faithfully resume: a crash-recovered
+    // host's future draws should be independent of the persisted snapshot.
+    // (`rng_word_pos` is now serialized, so `rehydrate_rng` *could* resume the
+    // exact stream — resume opts out on purpose.) Mirrors server-core.
     let fresh_seed: u64 = rand::rng().random();
     state.rng_seed = fresh_seed;
     state.rng = ChaCha20Rng::seed_from_u64(fresh_seed);
@@ -2120,9 +2129,9 @@ mod tests {
         // Flag flipped atomically with state load.
         assert!(is_multiplayer_mode());
 
-        // RNG seed was replaced with a fresh random value — stale seed would
-        // replay the pre-save ChaCha20 stream from position 0 and cause
-        // deterministic redraws.
+        // RNG seed was replaced with a fresh random value: resume deliberately
+        // diverges from the persisted snapshot rather than faithfully resuming
+        // its stream (see `resume_multiplayer_host_state`).
         let restored: GameState = serde_wasm_bindgen::from_value(get_game_state()).unwrap();
         assert_ne!(
             restored.rng_seed, 0xDEAD_BEEF_0000_0001,
--- a/crates/engine/src/types/game_state.rs
+++ b/crates/engine/src/types/game_state.rs
@@ -44,6 +44,39 @@ fn default_rng() -> ChaCha20Rng {
     ChaCha20Rng::seed_from_u64(0)
 }
 
+/// Serialize `rng` as just its exact ChaCha20 stream position
+/// (`get_word_pos`), emitted under the `rng_word_pos` key. The seed itself
+/// lives in the sibling `rng_seed` field; together they let
+/// [`GameState::rehydrate_rng`] reconstruct the precise stream after a load,
+/// so a restored snapshot continues the random sequence instead of rewinding
+/// to the opening shuffle (issue #5466). Reading the position off the live
+/// `rng` here means it is always captured fresh at serialize time — there is
+/// no companion field to keep in sync.
+fn serialize_rng_word_pos<S>(rng: &ChaCha20Rng, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    serializer.serialize_u128(rng.get_word_pos())
+}
+
+/// Deserialize a saved stream position into a throwaway seed-0 ChaCha20
+/// stream. The real seed is applied afterward by [`GameState::rehydrate_rng`]
+/// (which reads this position back off the stream, reseeds from `rng_seed`,
+/// and restores the position) — the sibling `rng_seed` field is not reachable
+/// from a field-level `deserialize_with`, so the position is carried here and
+/// married to the seed one step later. Pre-#5466 saves omit the `rng_word_pos`
+/// key and fall back to [`default_rng`] (position 0 = the historical
+/// rewind-to-origin behavior).
+fn deserialize_rng_word_pos<'de, D>(deserializer: D) -> Result<ChaCha20Rng, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let word_pos = u128::deserialize(deserializer)?;
+    let mut rng = ChaCha20Rng::seed_from_u64(0);
+    rng.set_word_pos(word_pos);
+    Ok(rng)
+}
+
 fn default_game_number() -> u8 {
     1
 }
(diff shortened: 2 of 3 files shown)


matthewevans:
Welcome, and thanks — this is a strong first contribution to a subtle bug.

I compared this against #5479, which fixes the same issue (#5466) and was opened nine minutes earlier. **On design, this is the one I want to take**, for a reason worth spelling out:

You serialize the position off the live `rng` (`serialize_with = "serialize_rng_word_pos"`) rather than adding a separate `rng_word_pos` field. That has a consequence I don't think is an accident: `visibility.rs:33` already redacts per-viewer state with `filtered.rng = ChaCha20Rng::seed_from_u64(0)`, so a redacted state serializes position `0` automatically. The competing PR had to add an explicit `filtered.rng_word_pos = 0` line to avoid leaking the keystream offset to opponents — a line any future serialize path could forget. Your version makes that leak structurally impossible. Centralizing reload in an engine-owned `GameState::rehydrate_rng()` is likewise the right call under "the engine owns all logic"; the WASM bridge shouldn't own game-state logic.

I verified independently: `rng` stays out of `impl PartialEq for GameState`, so CR 104.4b draw-by-repetition loop detection is intact; `default = "default_rng"` means pre-#5466 saves still load at position 0; and the `server-core/session.rs` hunk is comment-only, correctly leaving the deliberate fresh-reseed on host resume alone.

Three things before this can land.

**1. Your CI has not actually run.** You're a first-time contributor on a fork, so GitHub withholds workflow runs until a maintainer approves them. Only `Contributor trust` and `Superagent Security Scan` have executed — the Rust test shards, clippy, card-data, and WASM gates have not. The PR *looks* green; it isn't evidence yet. A maintainer needs to approve the workflows. Nothing here is blocked on you.

**2. MED — the WASM wiring is untested.** `rehydrate_rng()` has exactly one production caller, `engine-wasm/src/lib.rs:1286`. No test asserts that `restore_game_state` invokes it, so deleting that line leaves your engine-crate test green. `rehydrate_rng_resumes_stream_position_after_roundtrip` is a genuinely discriminating test of the serde round-trip — it drives the real `to_string` → `from_str` → `rehydrate_rng` and fails if either half breaks — but it can't see the WASM seam. Please add a test in `engine-wasm` that draws N values, calls the real `export_game_state_json()`, draws and records the next M, calls the real `restore_game_state(&json)`, and asserts those same M values come back.

**3. MED — the seed-0 placeholder is a silent-failure footgun.** `deserialize_rng_word_pos` seeds `ChaCha20Rng::seed_from_u64(0)` and sets the position, so between deserialization and `rehydrate_rng()` the `rng` holds the *wrong seed at the right offset*. Every live path handles it today (restore rehydrates; multiplayer resume and session load reseed fresh). But a future load path that forgets `rehydrate_rng()` draws real game randomness from seed 0 and nothing errors — it just silently produces the wrong values. Your doc comment says to call it; please also add a `debug_assert` or equivalent tripwire so forgetting it fails loudly rather than quietly.

Also, please fill in the PR template's **Files changed**, **Track**, and **LLM** sections — see `docs/AI-CONTRIBUTOR.md` for the gates we ask contributors to clear. Not a blocker, but it's how we keep review evidence legible.

To be clear about status: I'm holding #5479 rather than closing it until your CI is green and the two items above are addressed. If this holds up, yours lands. Nice work spotting that the position belongs on the `rng` field itself.

matthewevans:
Update, and I want to be straight with you because I told you the opposite an hour ago.

**I'm now landing #5479 instead.** Not because your design was wrong — my reasoning has changed on new evidence, and you deserve to see it.

After my comparison, #5479's author rewrote their PR (`e69de5a7`, *"own RNG snapshot/rehydrate in GameState; test the real seam"*). It now moves the logic into the engine — `GameState::capture_rng_word_pos()` and `GameState::rehydrate_rng()` — with `engine-wasm` merely calling them. That was my main architectural objection to it, and it's gone.

That left your structural advantage — no separate field, so the existing `visibility.rs` redaction can't be forgotten — against a hazard in your approach I flagged as MED:

```rust
// deserialize_rng_word_pos
let mut rng = ChaCha20Rng::seed_from_u64(0);
rng.set_word_pos(word_pos);
```

Between deserialization and `rehydrate_rng()`, `rng` holds the **wrong seed at the right offset**. Every live path handles it today, but a future load path that forgets the call draws real game randomness from seed `0` and nothing errors. #5479 keeps `rng` as `#[serde(skip, default = "default_rng")]` and `rng_word_pos` as plain `#[serde(default)]` data, so that window doesn't exist at all.

Weighed against that, your redaction advantage costs #5479 exactly one line (`filtered.rng_word_pos = 0`) — and I've asked for a test pinning it, which closes the "someone could forget it" gap you correctly identified. Your critique made that PR better even though it isn't the one merging.

The deciding practical factor: **your CI still hasn't run.** As a first-time contributor on a fork, GitHub withholds workflow runs until a maintainer approves them, and no approvable run has even been created for either of your heads (`2d295e3d` or `645a0ec5`). I can't verify your PR at all, while #5479 is green, was opened nine minutes earlier, and now has the right architecture. Merging an unverified duplicate over a verified equivalent isn't a call I can defend.

Two things worth saying plainly:

1. **You were right about the important thing.** Deriving the position from the live `rng` so redaction comes for free is genuinely the more elegant shape, and I said so in the review that made #5479's author go rewrite theirs. That's a real contribution to the outcome.
2. **This isn't a reflection on the work.** For a first PR, correctly spotting that the position belongs on the `rng` field, wiring `default = "default_rng"` for legacy saves, keeping `rng` out of `impl PartialEq for GameState` so CR 104.4b loop detection survives, and noticing that `server-core`'s host resume deliberately re-rolls — that's careful engineering.

I'm leaving this open rather than closing it, pending the maintainer's call. If #5479 hits a problem in the merge queue, yours is the fallback. And the `rehydrate_rng()` naming and the engine-ownership argument both came from your PR.

If you want to keep going, `docs/AI-CONTRIBUTOR.md` has the contributor gates, and the `#5466` fix leaves one thing undone that neither PR closed: there's still no test that drives the real `export_game_state_json()` → `restore_game_state()` round trip through the WASM bridge. That'd be a clean, well-scoped follow-up.

matthewevans:
Closing the loop: #5479 merged (`3951cc99`), so issue #5466 is fixed on main and this PR is now redundant.

I said I'd tell you if it stumbled in the queue. It didn't, so yours is superseded. I'm recommending it be closed, though the final call is the repo owner's — I'm not going to close a first contribution out from under someone without them seeing this first.

Two things I want on the record, because they aren't visible from the merged diff.

**Your review changed the outcome.** #5479's original design hand-synced a `rng_word_pos` field inside the WASM bridge. I only argued that the engine should own it because your PR demonstrated the cleaner shape — deriving the position from the live `rng` at serialize time and centralizing reload in one place. That author then rewrote their PR around engine-owned `capture_rng_word_pos()` / `rehydrate_rng()`. The method name in the merged commit is the one you chose.

**What decided it wasn't design.** It came down to two things:

1. Your `deserialize_rng_word_pos` seeds `ChaCha20Rng::seed_from_u64(0)` and then sets the position, so between deserialization and `rehydrate_rng()` the RNG holds the wrong seed at the right offset. Every live path handles it today, but nothing *enforces* it — a future load path that forgets the call would draw real game randomness from seed `0` and never error. The merged approach keeps `rng` as `#[serde(skip, default = "default_rng")]`, so that window doesn't exist.
2. Your CI never ran. GitHub withholds workflow runs on first-time-contributor fork PRs until a maintainer approves them, and no approvable run was ever even created — for either of your heads. I could not verify your PR at all, while the alternative was green. Merging an unverified duplicate over a verified equivalent isn't something I could defend, however much I liked the design.

Point 2 is a repo friction problem, not a you problem. It's the reason your first PR was hard to evaluate on its merits.

If you'd like a well-scoped follow-up, `docs/AI-CONTRIBUTOR.md` has the contributor gates, and there's a real one adjacent to this work: `visibility.rs` now redacts `rng_word_pos` per viewer so the ChaCha20 keystream offset can't leak to opponents — you were the one who pointed out that a separate field is easy to forget in redaction. Auditing the rest of `filter_state_for_viewer` for other fields with the same "easy to forget" shape would be genuinely useful, and it's exactly the instinct you already showed here.

Thanks for a sharp first contribution.

matthewevans:
Thanks for this, and sorry for the duplicate outcome — this is the same fix as #5479, which merged a few hours ago and closed the same underlying issue (#5466).

Both PRs resume the saved ChaCha20 stream by serializing the word position and re-applying it on restore. #5479 landed that as `rng_word_pos` + `rehydrate_rng()` / `capture_rng_word_pos()`, which are now on `main` — so the change here is already in the codebase by a different hand. Your approach (same `rehydrate_rng` mechanism, with the deliberate fresh-seed opt-out on multiplayer resume) matches what merged, which is good corroboration that it was the right design.

Closing as a duplicate of #5479. This isn't a reflection on the work — it was a close race and the other PR merged first. Thank you for digging into the RNG-snapshot semantics; please do send more.

PR #5486, opened 2026-07-10
fix(parser): bind PutCounterAll "untap them" to countered set (#5286)

### LINKED ISSUE #5286: Lulu, Loyal Hollyphant — **Card currently listed as:** 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 5/5 · fully supported **Steps to Reproduce:…
Reported in Discord: https://discord.com/channels/1485498006781427802/1523811201103171654/1523811201103171654

**Thread:** Lulu, Loyal Hollyphant
**Discord thread id:** `1523811201103171654`
**Discord message id:** `1523811201103171654`
**Cards:** lulu, loyal hollyphant
**Parser status:** has_gaps

## Summary
**Card currently listed as:** 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 5/5 · fully supported

**Steps to Reproduce:** Have a permanent leave the battlefield on your turn, tap some creatures with Lulu out, go to endstep.

## Reported text

**Card currently listed as:** 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 5/5 · fully supported

**Steps to Reproduce:** Have a permanent leave the battlefield on your turn, tap some creatures with Lulu out, go to endstep.
**Expected Result:** Lulu, Loyal Hollyphant puts a +1/+1 counter on all your tapped creatures, and then untaps them all.
**Actual Result:** Lulu puts the counters on all creatures, but it only untaps itself.

## Attachments

- [image.png](https://cdn.discordapp.com/attachments/1523811201103171654/1523811201404895383/image.png?ex=6a4d7745&is=6a4c25c5&hm=53cda084ed6ffd6674e286166b76da3e3dfb4a0744cb37901d005ef08c602cea&)

### lulu, loyal hollyphant
Flying
At the beginning of your end step, if a permanent you controlled left the battlefield this turn, put a +1/+1 counter on each tapped creature you control, then untap them.
Choose a Background (You can have a Background as a second commander.)

---
<sub>report_id: `discord:1523811201103171654:1523811201103171654:0` · discord: `1523811201103171654/1523811201103171654`</sub>

andriypolanski:
## Summary

Lulu, Loyal Hollyphant was listed as fully supported, but at the beginning of your end step (with revolt satisfied) it put +1/+1 counters on tapped creatures you control and then failed to untap them — only Lulu herself would untap. This fix rewrites the chained plural untap anaphor after a mass counter head to `TrackedSet(0)` so runtime resolution binds the creatures that received counters.

Fixes [#5286](https://github.com/phase-rs/phase/issues/5286).

## Root cause

Lulu's end-step trigger parses correctly through `PutCounterAll` (each tapped creature you control), but the `"then untap them"` sub lowered to `SetTapState { target: SelfRef, scope: Single }`.

Phase-trigger bodies carry `ctx.subject = Any`, so `resolve_it_pronoun` binds bare `"them"` to the trigger source (Lulu) instead of the countered population. `ParentTarget` would also no-op at runtime because `PutCounterAll` does not populate `ability.targets`.

### Parser — lowering (`oracle_effect/lower.rs`)

- Added `patch_population_head_tap_anaphor` (sibling of `patch_self_ref_head_tap_anaphor`): when a broadcast `PutCounterAll` head chains a single-scope tap/untap anaphor (`SelfRef` or `ParentTarget`), rewrite to `TrackedSet(0)` so `tap_untap.rs` consumes the set published by `CounterAdded` events (CR 608.2c + CR 122.1).

### Tests

- Parser: `lulu_loyal_hollyphant_counters_then_untap_them_lowers_correctly`, `put_counter_all_head_plural_untap_rewrites_to_tracked_set`
- Integration: `issue_5286_lulu_loyal_hollyphant.rs` — revolt + tapped creatures get counters and untap; no revolt → no effect

## Out of scope

- Card-data regeneration (`cargo card-data`); runtime `from_oracle_text` path is fixed and integration tests use live Oracle text.
- Generalizing to other mass-effect heads (e.g. `PumpAll` + untap them); patch is scoped to `PutCounterAll` publishers.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test -p engine --lib lulu_loyal_hollyphant_counters_then_untap_them`
- [x] `cargo test -p engine --lib put_counter_all_head_plural_untap_rewrites_to_tracked_set`
- [x] `cargo test -p engine --test integration issue_5286` (2/2)

## Risk notes

- Gate is narrow: only `PutCounterAll` with a broadcast population filter, immediate chained `SetTapState` with `scope: Single` and `SelfRef`/`ParentTarget`. `patch_self_ref_head_tap_anaphor` and targeted-counter chains are unchanged.
- Covers the class of "put a counter on each …, then untap them" (Lulu and future cards with the same shape).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -237,6 +237,59 @@ fn patch_self_ref_head_tap_anaphor(def: &mut AbilityDefinition) {
     walk(def, false);
 }
 
+/// CR 608.2c + CR 122.1: After a mass counter placement (`PutCounterAll`), a
+/// chained "then untap them" continuation refers to the set of objects that
+/// received counters (Lulu, Loyal Hollyphant). Phase-trigger bodies carry
+/// `ctx.subject = Any`, so `resolve_it_pronoun` wrongly binds "them" to
+/// `SelfRef` (the trigger source). Rewrite to `TrackedSet(0)` so the runtime
+/// binds the published counter set via `affected_objects_from_events`. Sibling
+/// of [`patch_self_ref_head_tap_anaphor`] for the population-head / plural-
+/// anaphor polarity.
+fn patch_population_head_tap_anaphor(def: &mut AbilityDefinition) {
+    fn is_population_counter_publisher(effect: &Effect) -> bool {
+        matches!(
+            effect,
+            Effect::PutCounterAll { target, .. }
+                if !matches!(
+                    target,
+                    TargetFilter::SelfRef
+                        | TargetFilter::ParentTarget
+                        | TargetFilter::TriggeringSource
+                        | TargetFilter::CostPaidObject
+                )
+        )
+    }
+
+    fn walk(def: &mut AbilityDefinition, carried_population: bool) {
+        let active_population = if is_population_counter_publisher(&def.effect) {
+            true
+        } else {
+            match def.effect.target_filter() {
+                Some(filter) if !target_filter_is_player_scoped(filter) => false,
+                _ => carried_population,
+            }
+        };
+        if let Some(sub) = def.sub_ability.as_deref_mut() {
+            if active_population {
+                if let Effect::SetTapState {
+                    target,
+                    scope: EffectScope::Single,
+                    ..
+                } = sub.effect.as_mut()
+                {
+                    if matches!(target, TargetFilter::SelfRef | TargetFilter::ParentTarget) {
+                        *target = TargetFilter::TrackedSet {
+                            id: crate::types::identifiers::TrackedSetId(0),
+                        };
+                    }
+                }
+            }
+            walk(sub, active_population);
+        }
+    }
+    walk(def, false);
+}
+
 /// CR 608.2c: After a "choose a card …" interactive selection, the chained
 /// "… {remove|put} that many counters {from|on} it" continuation's "it" refers to
 /// the chosen card. The standalone continuation clause lowers its "it" anaphor to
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
**APPROVE** (maintainer sign-off) — fix(parser): bind PutCounterAll "untap them" to the countered set (#5286).

**Seam.** Direct sibling of the existing `patch_self_ref_head_tap_anaphor`, sharing the `lower_effect_chain_ir` chokepoint (call site `lower.rs:2328`, immediately after the SelfRef-head sibling). It must run at lowering time: by resolution the SelfRef-vs-countered-set discriminator is erased. Correct location, most idiomatic implementation (typed `matches!` on `Effect` variants, no string dispatch — nom-irrelevant, this is an AST pass).

**Blast radius — confronted, not assumed.** Parse-diff sticky (baseline main `7b5b324d5751`, updated `2026-07-10T07:19:52Z`) reports exactly **1 card, 1 signature**: `ability/Untap` field `target`: `self` → `tracked set #0`, Lulu, Loyal Hollyphant. Freshness chain holds: head push `07:08:34Z` → Card-data check `07:19:36Z` → sticky `07:19:52Z`. Because this is a parser-surface PR, a clean `no_changes` sticky would have signalled an *inert* fix (the #5480 failure mode). This one fires on the real card; the test's `LULU_ORACLE` is character-identical to the printed Oracle text (no hand-normalization).

**Gating verified by code + data.**
- `is_population_counter_publisher` returns true only for `Effect::PutCounterAll` whose target is NOT `SelfRef|ParentTarget|TriggeringSource|CostPaidObject` — so a *targeted* counter effect never trips it.
- The inner rewrite touches only `SetTapState { scope: Single, target: SelfRef | ParentTarget }`; an existing `TrackedSet` target is left alone.
- Sibling trace (demonstrative "untap those creatures"): Virtue of Loyalty and The Fifth Doctor already lower their untap sub to `SetTapState { target: TrackedSet #0 }` — untouched by the patch (confirmed both in card-data and by the 1-card sticky).
- Negative trace (targeted counter): Leo's Guidance's head is `PutCounter` (single, targeted), not `PutCounterAll`, and its head filter is creature-scoped → `active_population` stays false → the branch never fires → its `SetTapState { target: ParentTarget }` binding survives.

**Test discrimination.** `issue_5286_lulu_...` drives the real end-step trigger through the pipeline. If the fix were reverted, "them" would lower to `SelfRef` and bind Lulu (untapped, not in the countered set), so the `!obj.tapped` assertions on Tapped A / Tapped B would fail. The paired `+1/+1 == Some(1)` assertion is a positive reach-guard proving the trigger fired (guards against a vacuous pass). The no-revolt test pins the intervening-if.

**One LOW (noted, not blocking).** The `drive_end_step_stack` helper swallows `OrderTriggers`/`PassPriority` with `.ok()` rather than `.expect(...)`, so an unexpected `Err` (e.g. a priority-loop freeze) is masked into a less precise failure signal. Impact is bounded: the `0..64` loop plus the downstream counter/untap assertions still fail the test. This confirms Gemini's single comment, downgraded to LOW.

All 13 required checks green.

PR #5488, opened 2026-07-10
fix(parser, engine): recognize qualified "source of your choice" damage prevention (Circle/Rune of Protection, 13 cards)

rykerwilliams:
## Summary

Fixes a real 13-card class where **"the next time a `<color/type>` source of your choice would deal damage to you this turn, prevent that damage"** (Circle of Protection cycle + Rune of Protection cycle) was mishandled by the parser and resolver.

**Affected cards** (confirmed via Scryfall for all 13, verbatim grammar identical modulo cost and the color/type qualifier):
- Circle of Protection: Artifacts, Black, Blue, Green, Red, White
- Rune of Protection: Artifacts, Black, Blue, Green, Lands, Red, White

**Circle of Protection: Shadow** is explicitly excluded — it uses an unrelated `ChooseAPermanent` + Shadow-matching mechanic, not `ChooseADamageSource`.

CR 609.7 (`docs/MagicCompRules.txt`) literally quotes Circle of Protection: Red's Oracle text as its own worked example — this is a foundational rules mechanic, not a narrow card-specific fix.

## Root cause

Two independent, compounding gaps:

1. **Parser** — `parse_oneshot_source_filter` (`oracle_replacement.rs`) only recognized a *bare*, unqualified `"a source of your choice"` / `"that source"`. A qualified form like `"a blue source of your choice"` fell through the whole chain and returned `None` entirely. Net effect: these 13 cards currently create a **fully unconstrained** "prevent the next damage this turn from *any* source" shield, with **no interactive choice ever prompted** — strictly worse than doing nothing, since it silently over-delivers value with zero player agency.
2. **Resolver** — `prevent_damage.rs::resolve` (the resolver these cards actually route through, not `create_damage_replacement.rs`) had **zero** "prompt if not yet chosen" logic. Its sibling `create_damage_replacement.rs::resolve` already had this exact self-prompt pattern for the bare form (hardcoded to `TargetFilter::Any`) — `prevent_damage.rs` never got the equivalent.

## Fix

- Parameterized `TargetFilter::ChosenDamageSource` from a bare unit variant into `ChosenDamageSource { filter: Option<Box<TargetFilter>> }`, following the **direct existing precedent** of `PlayerFilter::OpponentDealtDamage { source: Option<Box<TargetFilter>> }` (`types/ability.rs:5751-5757`) and its scanner-recursion pattern (`ability_scan.rs:3039-3049`) — not a new pattern.
- Added `parse_qualified_chosen_damage_source` (`oracle_replacement.rs`), reusing `parse_type_phrase` for the qualifier — the same building block `parse_damage_source_subject_filter` already composes elsewhere in this file for non-anaphoric forms.
- Threaded the qualifier through **both** resolvers' self-prompt blocks via a single `prompt_filter` binding, consumed by *both* the candidate-enumeration call *and* `WaitingFor::DamageSourceChoice.source_filter` — so the **interactive prompt itself** is qualifier-filtered (e.g. Circle of Protection: Red only ever offers red sources as choices), not just the CR 609.7b post-choice recheck. A dedicated test (`circle_of_protection_red_prompt_options_are_color_filtered`) asserts `WaitingFor::DamageSourceChoice.options` directly, specifically to catch a regression where only the recheck stays qualifier-aware but the prompt itself silently reverts to `Any`.
- Added the missing self-prompt block to `prevent_damage.rs`, mirroring `create_damage_replacement.rs`'s existing one.
- Updated ~24 production match sites (`ability_rw.rs`, `ability_scan.rs`, `cost_payability.rs`, `coverage.rs`, `filter.rs`, `trigger_matchers.rs`, `oracle_effect/lower.rs`, `oracle_static/restriction.rs`) to compile against the new struct-variant shape. `ability_scan.rs`'s scanner arm is the one site needing real compositional logic (recursing into the nested filter for `Axes` computation), not just a mechanical `{ .. }` — it mirrors the `OpponentDealtDamage` precedent exactly.
- **Cross-crate compile-compat**: `crates/mtgish-import` (a separate workspace member) pattern-matched the old unit-variant shape in 3 places (`convert/replacement.rs`, `convert/condition.rs`), which would have broken `cargo check --workspace`. Updated mechanically (`{ filter: None }` / `{ .. }`) — no new converter functionality, purely keeping it compiling against a type it already references.

No `ControllerRef`/`bool` shortcuts anywhere; `Option<Box<TargetFilter>>` is the established idiom for this exact "optional nested filter on a fixed-shape variant" pattern.

## Gate A — parser combinator compliance

```
(exit 0, no violations)
```

## Gate B — anchored citations (≥2)

- [`crates/engine/src/types/ability.rs:5751-5757`](../../blob/main/crates/engine/src/types/ability.rs) — `PlayerFilter::OpponentDealtDamage { source: Option<Box<TargetFilter>> }`, the direct precedent for `ChosenDamageSource`'s new field shape.
- [`crates/engine/src/game/ability_scan.rs:3039-3049`](../../blob/main/crates/engine/src/game/ability_scan.rs) — that precedent's scanner-recursion pattern, mirrored exactly for `ChosenDamageSource`'s own `scan_target_filter` arm.
- [`crates/engine/src/game/effects/create_damage_replacement.rs`](../../blob/main/crates/engine/src/game/effects/create_damage_replacement.rs) (existing self-prompt block, lines ~79-131 pre-diff) — the pattern mirrored into `prevent_damage.rs`'s new block.

## CR references

- **CR 609.7** — this rule's own worked example is Circle of Protection: Red's Oracle text verbatim.
- **CR 609.7a** — the source is chosen when the effect is created.
- **CR 609.7b** — the shield rechecks the chosen source's properties when it would deal damage (the live recheck, distinct from the snapshotted identity).
- **CR 514.2** — "this turn" duration cleanup.
- **CR 614.5** — one-shot replacement consumption.
- **CR 615.3** — relevant to shield hosting: a battlefield-permanent source (Circle/Rune of Protection) hosts its shield on the source object's own `replacement_definitions`, not a global pending-registry — this distinction surfaced and was correctly resolved during a test-authoring fix mid-implementation (see Validation below).

## Tier

**Tier: Frontier**

Justification: this is a 13-card foundational-mechanic fix that parameterizes a shared `TargetFilter` enum variant across ~24 production match sites in the engine plus 3 sites in a separate workspace crate, adds a new resolver code path (`prevent_damage.rs` previously had no self-prompt logic at all), and went through 3 rounds of plan review and 3 rounds of implementation review before landing — substantially beyond a single-arm parser fix.

**Model:** claude-sonnet-5
**Thinking:** high

## Validation performed

- `cargo fmt --all` — clean (pre- and post-rebase)
- Parser dispatch grep gate — clean (no string-dispatch primitives in added lines)
- CR-annotation diff gate — all 6 citations verified against `docs/MagicCompRules.txt`, zero `UNVERIFIED`
- `cargo clippy -p engine -p mtgish-import --all-targets -- -D warnings` — clean, zero warnings (confirmed **pre-rebase**)
- `cargo check --workspace --all-targets` — clean, confirms every workspace crate (engine, mtgish-import, engine-wasm, phase-ai, phase-server, draft-core, draft-wasm, lobby-broker, seat-reducer, manabrew-compat, feed-scraper, engine-inventory-gen) compiles against the new `ChosenDamageSource` shape
- `cargo test -p engine --lib` — **15962 passed, 0 failed, 6 ignored** (confirmed **pre-rebase**)
- `cargo test -p engine --test integration` — **2563 passed, 0 failed, 2 ignored** (confirmed **pre-rebase**) — includes a dedicated new test file covering both the color-qualifier branch (Circle of Protection: Red) and the type-qualifier branch (Rune of Protection: Lands, which also confirms its independent `Cycling` line decomposes correctly and doesn't interact with the prevention clause)
- Full discriminating-test coverage: the interactive-prompt path itself (not just the post-choice recheck), a multi-authority hostile fixture (two same-qualifier sources, only the chosen one affected), a CR 609.7b recheck-failure fixture (chosen source's color changes before damage — shield doesn't fire, isn't consumed), a no-legal-candidate fixture (empty prompt, no hang), and the bare-form regression guard (Desperate Gambit / Beacon of Destiny / Jade Monolith's existing unqualified form still works, proven by a paired test showing it offers *all* candidates rather than a qualified subset)
- Independent `/review-impl` — **3 rounds**. Round 1 found 2 issues (a test asserting the wrong shield-storage location due to a genuine CR 615.3 nuance — permanent-sourced shields host on the object, not a global registry — and a broken integration-test assertion expecting `Cycling` to appear as a second `ability` rather than in `extracted_keywords`), both fixed. Round 2 (a deliberate holistic fresh-eyes pass) found the `mtgish-import` cross-crate compile break described above, fixed and independently re-verified via direct `cargo check --workspace`. Round 3 confirmed clean end-to-end.

## Validation NOT performed (disclosed honestly)

Post-rebase, this branch could not get a **post-rebase** re-confirmation of `cargo clippy` or `cargo test -p engine --lib` — 5 consecutive attempts of each were killed mid-compile at the identical "Compiling engine v0.20.0" stage, with the system independently confirmed idle (no resource contention) each time via direct process inspection. This matches an unrelated, intermittent background-task environment issue observed repeatedly elsewhere in this session (resolved by simple retries in most other cases; this branch happened to hit the retry ceiling on both commands). The rebase itself introduced **zero code conflicts** — only `docs/parser-misparse-backlog.md`'s top-summary totals needed manual arithmetic reconciliation against a sibling PR merged in the interim (Nettling Imp, #5463) — so the regression risk from the rebase is low, but I'm not claiming a post-rebase clippy/test re-run happened when it didn't.

## Backlog hygiene

Removed all 13 cards from `docs/parser-misparse-backlog.md`'s category 11 list, decrementing that category's count (170→157) and the top-summary totals accordingly (rebased onto the post-#5463 baseline: 4776→4763 distinct, 4810→4797 total appearances). `Circle of Protection: Shadow` and the pre-existing unrelated `Circle of Protection: Art` stray entry are correctly retained.

### DIFF
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -1376,7 +1376,7 @@ fn scope_of(target: &TargetFilter, chain_root: Option<WriteScope>) -> WriteScope
         | TargetFilter::PostReplacementDamageTargetOwner
         | TargetFilter::DefendingPlayer
         | TargetFilter::HasChosenName
-        | TargetFilter::ChosenDamageSource
+        | TargetFilter::ChosenDamageSource { .. }
         | TargetFilter::Named { .. }
         | TargetFilter::Owner
         | TargetFilter::AllPlayers => WriteScope::External,
@@ -2213,7 +2213,7 @@ fn legacy_target_filter(f: &TargetFilter) -> bool {
         | TargetFilter::PostReplacementSourceController
         | TargetFilter::PostReplacementDamageTarget
         | TargetFilter::PostReplacementDamageTargetOwner
-        | TargetFilter::ChosenDamageSource
+        | TargetFilter::ChosenDamageSource { .. }
         | TargetFilter::None
         | TargetFilter::Any
         | TargetFilter::Player
@@ -2407,7 +2407,7 @@ fn member_bound_target_filter(f: &TargetFilter) -> bool {
         | TargetFilter::ChosenCard
         | TargetFilter::HasChosenName
         | TargetFilter::SourceChosenPlayer
-        | TargetFilter::ChosenDamageSource
+        | TargetFilter::ChosenDamageSource { .. }
         | TargetFilter::AttachedTo
         | TargetFilter::Neighbor { .. }
         | TargetFilter::OriginalController
@@ -6079,7 +6079,7 @@ fn rw_target_filter(x: &TargetFilter) -> RwProfile {
         | TargetFilter::PostReplacementSourceController
         | TargetFilter::PostReplacementDamageTarget
         | TargetFilter::PostReplacementDamageTargetOwner
-        | TargetFilter::ChosenDamageSource => reads_event_live(),
+        | TargetFilter::ChosenDamageSource { .. } => reads_event_live(),
         TargetFilter::Not { filter } | TargetFilter::TrackedSetFiltered { filter, .. } => {
             rw_target_filter(filter)
         }
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -2511,11 +2511,17 @@ fn scan_target_filter(x: &TargetFilter) -> Axes {
         },
         TargetFilter::DefendingPlayer => Axes::NONE,
         TargetFilter::HasChosenName => Axes::NONE,
-        TargetFilter::ChosenDamageSource => Axes {
-            event: true,
-            sibling: false,
-            projected: false,
-        },
+        TargetFilter::ChosenDamageSource { filter } => {
+            let mut acc = Axes {
+                event: true,
+                sibling: false,
+                projected: false,
+            };
+            if let Some(f) = filter {
+                acc = acc.or(scan_target_filter(f));
+            }
+            acc
+        }
         TargetFilter::Named { name: _ } => Axes::NONE,
         TargetFilter::Owner => Axes::NONE,
         TargetFilter::AllPlayers => Axes::NONE,
--- a/crates/engine/src/game/ability_utils.rs
+++ b/crates/engine/src/game/ability_utils.rs
@@ -12386,7 +12386,7 @@ mod tests {
 
         let ability = ResolvedAbility::new(
             Effect::CreateDamageReplacement {
-                source_filter: Some(TargetFilter::ChosenDamageSource),
+                source_filter: Some(TargetFilter::ChosenDamageSource { filter: None }),
                 combat_scope: None,
                 target_filter: None,
                 modification: None,
--- a/crates/engine/src/game/cost_payability.rs
+++ b/crates/engine/src/game/cost_payability.rs
@@ -93,7 +93,7 @@ pub(crate) fn target_filter_has_pitch_bound_x(filter: &TargetFilter) -> bool {
         | TargetFilter::PostReplacementDamageTargetOwner
         | TargetFilter::DefendingPlayer
         | TargetFilter::HasChosenName
-        | TargetFilter::ChosenDamageSource
+        | TargetFilter::ChosenDamageSource { .. }
         | TargetFilter::Named { .. }
         | TargetFilter::Owner
         // CR 201.5a: a granter self-ref carries no pitch-bound X.
@@ -171,7 +171,7 @@ pub(crate) fn relax_pitch_bound_x_filter(filter: &TargetFilter) -> TargetFilter 
         | TargetFilter::PostReplacementDamageTargetOwner
         | TargetFilter::DefendingPlayer
         | TargetFilter::HasChosenName
-        | TargetFilter::ChosenDamageSource
+        | TargetFilter::ChosenDamageSource { .. }
         | TargetFilter::Named { .. }
         | TargetFilter::Owner
         // CR 201.5a: no pitch-bound X constraint to relax.
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -624,7 +624,10 @@ fn fmt_target(filter: &TargetFilter) -> String {
         }
         TargetFilter::ExiledBySource => "cards exiled by source".into(),
         TargetFilter::HasChosenName => "card with the chosen name".into(),
-        TargetFilter::ChosenDamageSource => "chosen damage source".into(),
+        TargetFilter::ChosenDamageSource { filter: Some(f) } => {
+            format!("chosen damage source matching {}", fmt_target(f))
+        }
+        TargetFilter::ChosenDamageSource { filter: None } => "chosen damage source".into(),
         TargetFilter::Named { name } => format!("card named {name}"),
         TargetFilter::Not { filter } => format!("not {}", fmt_target(filter)),
         TargetFilter::Or { filters } => filters
--- a/crates/engine/src/game/effects/create_damage_replacement.rs
+++ b/crates/engine/src/game/effects/create_damage_replacement.rs
@@ -77,13 +77,16 @@ pub fn resolve(
     // and re-enter this resolver as a continuation; on the second pass the choice
     // is recorded and we proceed.
     let resolved_source_filter = match &source_filter {
-        Some(TargetFilter::ChosenDamageSource) => {
+        Some(TargetFilter::ChosenDamageSource { filter: qualifier }) => {
             match state.last_chosen_damage_source.as_ref() {
                 Some(_choice) => {
                     // CR 609.7b: Resolve the chosen damage source filter to check if
-                    // the source matches the filter.
+                    // the source matches the filter (including any color/type
+                    // qualifier carried on the variant, rechecked live per 609.7b).
                     let resolved = resolve_source_filter(
-                        &TargetFilter::ChosenDamageSource,
+                        &TargetFilter::ChosenDamageSource {
+                            filter: qualifier.clone(),
+                        },
                         state,
                         ability.source_id,
                         &ability.targets,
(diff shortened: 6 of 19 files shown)


matthewevans (approved):
## Approved

This is strong work, and it lands the mechanic properly rather than patching the symptom.

I verified the qualifier is genuinely threaded end to end, not merely stored. `prompt_filter` drives both the candidate enumeration and `WaitingFor::DamageSourceChoice.source_filter`; that gets recorded onto `state.last_chosen_damage_source.source_filter`, and `resolve_source_filter`'s `other` arm then builds `And { identity, recheck }`, which is what actually satisfies the CR 609.7b live property recheck. The sentinel is correctly transient: set, then cleared once `drain_pending_continuation` runs.

The no-legal-source path is fail-closed. `resolve_source_filter(..).unwrap_or(TargetFilter::None)` yields a shield that matches nothing, and `circle_of_protection_red_no_legal_source_is_noop` proves it end to end by putting only a blue source on the battlefield and asserting life stays at 17 after 3 damage.

The tests discriminate, which is the part I care most about. `circle_of_protection_red_prompt_options_are_color_filtered` seeds **both** a red and a blue source, so the negative assertion cannot pass vacuously, and it carries a positive reach-guard alongside it. `bare_source_of_your_choice_prompt_offers_all_colors` is the sibling that proves the two paths actually differ rather than both collapsing to the same behavior. Asserting on `WaitingFor::DamageSourceChoice.options` directly is the right anchor for this.

Using a typed `Option<Box<TargetFilter>>` following the `PlayerFilter::OpponentDealtDamage` precedent, instead of a bool, is exactly the call this repo wants, and `ability_scan.rs` correctly recurses into the nested filter rather than stopping at a mechanical `{ .. }`.

CR 609.7 quoting Circle of Protection: Red's Oracle text verbatim as its own worked example is a fair characterization; I confirmed it at line 2861 of the rules text.

---

## About the `coverage-parse-diff` sticky — please read this one

The `coverage-parse-diff` sticky on this PR says **"No card-parse changes detected."** On a parser PR that is normally a high-confidence signal that the fix is inert, meaning the parse output did not actually change and the change accomplishes nothing at runtime. I want to be explicit that **this is not the case here.** I investigated it and it is a blind spot in the sticky's renderer.

`coverage.rs` builds the `PreventDamage` effect signature from `amount`, `target` and `scope` only:

```rust
Effect::PreventDamage { amount, target, scope, .. } => {
    d.push(("amount".into(), format!("{amount:?}")));
    d.push(("scope".into(), format!("{scope:?}")));
}
```

`damage_source_filter` is swallowed by the `..`. So a field going from `None` to a qualified `ChosenDamageSource { filter: Some(..) }` produces a byte-identical signature and is invisible to the parse diff. The effect genuinely changed; the renderer simply cannot see the field that changed.

I will file a separate tooling issue for the missing field. Thank you for adding the `parse_oracle_text` integration test, because that test driving real Oracle text is the only reason this was provable either way. Without it I would have had no evidence to distinguish "renderer blind spot" from "inert fix," and the sticky would have been the deciding signal.

---

## Non-blocking follow-ups

None of these block the merge. Recording them so they are not lost.

**1. (Non-blocking) `create_damage_replacement.rs` empty-options branch attaches no source filter.**

When `options.is_empty()`, that resolver falls through with `resolved_source_filter = None`, and the shield builder is guarded:

```rust
if let Some(filter) = resolved_source_filter {
    shield = shield.damage_source_filter(filter);
}
```

With `None` there is no source filter on the shield, so it replaces damage from **any** source. Its own comment claims "the replacement does nothing," which is the opposite of what happens. Your `prevent_damage.rs` handles the identical case correctly by falling through with the `ChosenDamageSource` filter intact, which then resolves to `TargetFilter::None` and matches nothing.

I confirmed against `main` that both the comment and the `None` fall-through pre-date this PR, and that your change **narrows** the exposure rather than introducing it, since previously the qualified form never produced a choice at all. It is not a regression and it does not block. The reason I raise it now is that `parse_damage_modification_static` feeds one `source_filter` binding into both `Effect::CreateDamageReplacement` and `Effect::PreventDamage`, so the empty-source corner becomes reachable with a qualifier attached. Would you prefer to fix it here, mirroring what you already wrote in `prevent_damage.rs`, or in a follow-up? Either is fine by me.

**2. (Non-blocking) The self-prompt block now lives in two resolvers, and they already disagree.**

Item 1 is concrete evidence that the duplicated block has drifted on the empty-options path. That is the usual argument for extracting a single `prompt_for_damage_source(..)` authority. Two flows sits under this repo's "3+ = generalize" bar, so I am recording this as a recommendation rather than a request.

**3. (Non-blocking) One CR annotation is misapplied, and one PR-body citation is off.**

The `CR 615.3` comment in the new `prevent_damage.rs` test justifies hosting the shield on the source object's `replacement_definitions`. CR 615.3 reads:

> There are no special restrictions on casting a spell or activating an ability that generates a prevention effect. Such effects last until they're used up or their duration has expired.

It says nothing about where a shield is stored. Its second sentence is a good anchor for one-shot consumption, together with CR 609.7b's "if the shield prevents no damage, the shield isn't used up." The hosting claim needs a different citation or none.

Separately, in the PR body only, `CR 614.5` is described as "one-shot replacement consumption." The actual text is that a replacement effect does not invoke itself repeatedly and gets only one opportunity to affect an event. Related, though not the same claim. No code change needed for that one.

**4. (Non-blocking) No test covers the `an <type>` article shape.**

The new integration test covers Circle of Protection: Red (color qualifier) and Rune of Protection: Lands (type qualifier, `a land source`). Nothing exercises `an artifact source`, which is what Circle of Protection: Artifacts and Rune of Protection: Artifacts use. `parse_article = alt((tag("an "), tag("a ")))` handles it, so I read this as a coverage gap rather than a bug. A third case in the existing test would close it.

---

## One correction in your favor

Your root-cause narrative says the qualified form "returned `None` entirely" on `main`. I went in expecting to flag this as an overstatement, on the theory that `parse_oneshot_source_filter` falls through to `parse_damage_source_filter(body)`, which might return `Some(Typed(red))` and leave `main` over-preventing from any red source rather than being fully unconstrained.

I traced it and **you are right.** On `main`, for `"a red source of your choice"`, `split_damage_source_noun` yields `qualifier = "red"` and `tail = "of your choice"`. That tail is non-empty and matches neither `"you control"` nor `"of the chosen type"`, so `parse_damage_source_subject` bails to `None`. `parse_type_phrase` then leaves `"source of your choice"` as a non-empty remainder, so that path returns `None` too. `parse_oneshot_source_filter` really does return `None`, the shield gets no `damage_source_filter`, and it replaces damage from any source with no choice ever prompted. Your description of the bug is accurate as written.

Approving. Nice work.

PR #5476, opened 2026-07-10
Add Fact or Fiction

Whovencroft:
## Summary
Adds engine support for **Fact or Fiction** using its actual Oracle wording.

- Extends the existing pile-disposition grammar to accept the canonical “the other” wording while retaining the sibling “the rest” form.
- Preserves fail-closed handling for unsupported trailing riders.
- Updates the production cast/resolve/partition/choice integration test to use the exact Oracle text and verify both pile destinations and continuation back to priority.

## Files changed
- `crates/engine/src/parser/oracle_separate_piles.rs`
- `crates/engine/tests/integration/fact_or_fiction_pile_separation.rs`

## CR references
None added or changed; this grammar synonym lowers to the existing pile-separation semantics.

## Track
Non-developer

## LLM
Model: codex-5
Thinking: high

Tier: Standard

## Verification
Local verification skipped — see CI status checks.

The engine-implementer pipeline completed one clean plan-review round and one clean independent implementation-review round. The implementation review explicitly passed the maintainer-simulation gate and confirmed that reverting the `the other` parser arm breaks the production-path pile prompt assertion.

## Gate A
`./scripts/check-parser-combinators.sh` exited 0. The script produced no stdout or stderr; its full output is therefore the empty block below.

```text
```

## Anchored on
- `crates/engine/src/parser/oracle_separate_piles.rs:135` — existing same-module `alt(tag_no_case(...))` composition for equivalent creature-subject surface forms.
- `crates/engine/src/parser/oracle_separate_piles.rs:250` — existing same-module `alt(tag_no_case(...))` composition for equivalent opponent surface forms.
- `crates/engine/src/parser/oracle_separate_piles.rs:312` — existing typed `alt(value(Zone, tag_no_case(...)))` zone grammar reused by the changed seam.

## Coverage impact
Fact or Fiction should move from the `Effect:an` coverage gap to the existing supported `SeparateIntoPiles` effect. Trigger-body, variable-X, visibility, and trailing-rider variants remain honest unsupported gaps.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

PR #5489, opened 2026-07-10
feat(card): route pile-separation through trigger body parser (Sphinx…

Whovencroft:
## Summary

Route `parse_separate_into_piles` through the trigger body dispatch chain so ETB triggers with the Fact-or-Fiction pile-separation pattern produce a single `SeparateIntoPiles` effect instead of fragmenting into `Unimplemented` chunks.

**Cards unlocked:** Sphinx of Uthuun, and any future creature whose ETB trigger body matches the reveal-top-N → opponent-separates → zone-routing shape (the same pattern already works for spell bodies via the existing dispatch in `oracle.rs`).

## Files changed

| File | Change |
|------|--------|
| `crates/engine/src/parser/oracle_trigger.rs` | Add `parse_separate_into_piles` call in trigger body dispatch chain, after `parse_vote_block` and before `try_parse_exile_top_each_library_with_collection_counter` |
| `crates/engine/tests/integration/sphinx_of_uthuun_etb_pile_separation.rs` | New integration test driving the full ETB → pile-separation interactive flow |
| `crates/engine/tests/integration/main.rs` | Register the new test module |

## Anchored on

- `crates/engine/src/parser/oracle_trigger.rs:1132` — existing `parse_vote_block` call in the same trigger body dispatch chain (same `PreLowered` pattern, same position in the `if/else if/else` ladder)
- `crates/engine/src/parser/oracle.rs:4519` — existing `parse_separate_into_piles` call in the spell body dispatch (same function, same `AbilityKind::Spell` argument)

## Gate A

```
(exit 0 — no violations)
```

## Verification

- `cargo check --lib -p engine` — passes
- `cargo fmt --all` — no changes

## CR annotations verified

```
grep -n "^700.3" docs/MagicCompRules.txt
3223:700.3. Some effects cause objects to be temporarily grouped into piles.

grep -n "^701.20" docs/MagicCompRules.txt
3436:701.20. Reveal
3438:701.20a To reveal a card, show that card to all players for a brief time.

grep -n "^603.1" docs/MagicCompRules.txt
2524:603.1. Triggered abilities have a trigger condition and an effect.
```

## Track

Non-developer (CI runs full verification).

## Test description

`sphinx_of_uthuun_etb_pile_separation`: Builds a creature from Oracle text with the Sphinx of Uthuun ETB trigger, casts it, verifies the trigger resolves into `SeparatePilesPartition` (opponent partitions 5 revealed cards), then `SeparatePilesChoice` (controller chooses), then verifies chosen pile goes to hand and unchosen pile goes to graveyard. Would fail if the change were reverted (trigger body would parse as `Unimplemented` instead of `SeparateIntoPiles`).

### DIFF
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -1148,6 +1148,20 @@ pub(crate) fn parse_trigger_line_with_index_ir(
                 ability.optional = true;
             }
             Some(TriggerBody::PreLowered(Box::new(ability)))
+        // CR 700.3 + CR 701.20a: Pile-separation (Fact or Fiction / Sphinx of
+        // Uthuun family). The multi-sentence block must be consumed as a single
+        // unit — chain parsing would fragment it into Unimplemented chunks.
+        } else if let Some(pile_def) =
+            crate::parser::oracle_separate_piles::parse_separate_into_piles(
+                &effect_for_parse,
+                AbilityKind::Spell,
+            )
+        {
+            let mut ability = pile_def;
+            if optional {
+                ability.optional = true;
+            }
+            Some(TriggerBody::PreLowered(Box::new(ability)))
         } else {
             try_parse_exile_top_each_library_with_collection_counter(
                 &effect_for_parse,
--- a/crates/engine/tests/integration/main.rs
+++ b/crates/engine/tests/integration/main.rs
@@ -590,6 +590,7 @@ mod sliver_static_grants;
 mod sothera_supervoid_edict_reanimate;
 mod specialize_runtime;
 mod spellstutter_sprite_counter_with_x;
+mod sphinx_of_uthuun_etb_pile_separation;
 mod spikeshell_harrier_speed_superlative;
 mod split_offstack_mana_value;
 mod springheart_nantuko_bestow_landfall;
--- a/crates/engine/tests/integration/sphinx_of_uthuun_etb_pile_separation.rs
+++ b/crates/engine/tests/integration/sphinx_of_uthuun_etb_pile_separation.rs
@@ -1,0 +1,152 @@
+//! Integration test: Sphinx of Uthuun ETB pile-separation flow.
+//!
+//! Verifies that the Fact-or-Fiction pile-separation effect works when it
+//! appears as an ETB trigger body (not just as a spell body). Sphinx of Uthuun
+//! is the canonical creature with this pattern:
+//!
+//! ```text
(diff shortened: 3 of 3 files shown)


matthewevans (approved):
Approved. This is a genuinely small, well-scoped change, and it earns its keep by unlocking a class rather than a card. The coverage sticky shows the new branch picks up **Unesh, Criosphinx Sovereign** as well as Sphinx of Uthuun, which the PR description didn't claim. Two cards from fourteen lines, with no text stolen from the branches further down the chain, is the right shape for this parser.

A few things I verified so they don't get re-litigated later:

**`AbilityKind::Spell` is correct here.** This superficially resembles a known bug in this repo where a permanent's ability gets tagged `Spell` and silently no-ops. It isn't that. Routing is decided by the IR node variant, not by `def.kind`: `OracleNodeIr::PreLoweredTrigger(def)` pushes into `result.triggers`, so wrapping the returned `AbilityDefinition` in `TriggerBody::PreLowered` is what carries the trigger-ness. The `kind` argument only reaches `parse_effect_chain_with_context` to parse the body's sub-effect chain. Every sibling branch in the same `PreLowered` chain passes `AbilityKind::Spell` for exactly this reason, including the generic `parse_effect_chain_ir` fallback at the end. The convention holds.

**The `RevealTop` removal is expected, not a dropped clause.** The sticky pairs `RevealTop` removed with `SeparateIntoPiles` added on both cards. The reveal is subsumed by the pile-separation effect, which is the same signature change #5476 produced when it landed Fact or Fiction. Reading that sticky row as a regression would be a misread.

**The test drives the real pipeline.** It builds the creature from Oracle text, casts it, resolves it, and asserts the interactive round trip: `WaitingFor::SeparatePilesPartition` routed to the opponent with 5 eligible cards, then `SeparatePilesChoice` to the controller, then hand/graveyard routing, then a return to `Priority`. Removing the parser branch fragments the body into `Unimplemented`, the partition prompt never fires, and the test fails. It discriminates on the behavior the PR claims to add, rather than asserting on parser output shape. It also lives in `crates/engine/tests/integration/` with the `mod` line added, per the rule against new top-level test binaries.

The CR citations hold up against `docs/MagicCompRules.txt`. CR 700.3 covers objects temporarily grouped into piles and uses Fact or Fiction as its own worked example, and CR 701.20a is Reveal.

The one Gemini nit (an unnecessary clone of `pile_a_ids`) is in test code and not worth a round trip. Merging.

Thanks for the clean contribution.

PR #5491, opened 2026-07-10
feat(engine-wasm): preview_action — non-mutating dry-run of public deltas

### LINKED ISSUE #5468: preview_action: non-mutating dry-run returning resolved targets + projected outcome deltas
## Summary

Feature request: a read-only `preview_action(actor, action)` that
reports what submitting the action WOULD do, without mutating the
game. Minimum useful payload:

- resolved targets (what auto_select_targets would pick),
- projected zone changes (deaths, bounces, exiles),
- projected life-total deltas per player.

## Motivation

Hover-previews for removal and combat ("this kills that", "you take 4")
are a top-tier UX affordance that Arena itself largely lacks; the
engine has all the information but exposes it only by committing the
action. We currently approximate previews with a speculative
submit + export/restore rollback, which is expensive per hover and
carries the restore side effects (RNG rewind, debug_mode=true).

This composes with (and subsumes some of) the read-only pre-cast
targets query: preview of a CastSpell yields the targets; preview of an
AssignCombatDamage yields the deaths.

## Scope thoughts

A full simulation is just "clone state, apply, diff" — the engine
already clones GameState for AI search. A pragmatic v1:

    preview_action_js(actor: u32, action: JsValue) -> JsValue
    // { resolved_targets: [...], zone_changes: [{object_id, from, to}],
    //   life_deltas: {player_id: i32}, log_preview: [...] }

implemented as: clone the state, submit against the clone, diff
(objects' zones + players' life), discard the clone. No RNG concerns
if the clone draws from a forked stream; hidden information stays
hidden because the result only reports observable diffs the acting
player could deduce anyway... with one guard: anything sourced from a
hidden zone (a library shuffle result, a random discard) should be
elided or bucketed ("a card is drawn") rather than named.

If you'd rather not own a diff format, even a variant that just runs
the clone and returns the same log_entries a real submit would produce
(minus hidden-zone lines) covers most of the UI value.

davion-knight:
Fixes #5468.

## What

A read-only `preview_action_js(actor, action)` that reports what submitting an action **would** do — **life-total changes** and **public-zone object transitions** (e.g. `Battlefield → Graveyard` = a death, `Stack → Graveyard` = a resolved/countered spell) — **without committing it**. This replaces the current speculative-submit + `export`/`restore_game_state` rollback (heavy per hover, and it carried the restore side effects) with a cheap dry-run, powering hover-preview UX like "this kills that" / "you take 4".

## Implementation

- New engine module **`game::preview`** owns the diff (`PreviewDiff` / `compute_preview_diff`), so the logic is **unit-tested natively** in the engine crate rather than only across the (CI-uncompiled) WASM boundary.
- **`preview_action_js`** runs the action on a **throwaway clone** via `apply_for_simulation` — the same rules resolution the AI look-ahead already uses — so the live `GAME_STATE` is never touched, then diffs the before/after snapshots.

## Hidden-information safety (two independent guards)

This is the delicate part; it's safe by construction:

1. The diff is taken over **`filter_state_for_viewer`** snapshots (hands, libraries, face-down identities, and RNG are already redacted for the actor), **and**
2. only transitions **between public zones** plus life totals are emitted (CR 400.2 public/hidden line) — so draws, random discards from hand, and library shuffles are **never** surfaced, even for the acting player's opponents.

So the preview can only ever report deltas the actor could already observe by committing the action.

## Tests (native, `engine` crate)

- `reports_life_delta_and_public_zone_change` — a creature dies + a player loses 4 life.
- `reports_created_token_in_public_zone` — a token appears.
- `elides_hidden_zone_movements` — Hand/Library transitions (draw, cast-from-hand, mill) produce **no** output.

Verified: `cargo check -p engine --tests`, `-p engine-wasm`, and `-p engine-wasm --target wasm32-unknown-unknown` all clean; `cargo fmt` clean.

## Scope

v1 covers the primary use cases (removal/combat previews). A richer payload (per-slot resolved targets, projected log lines with hidden-zone bucketing) can follow; happy to adjust the shape/seam to your preference.

### DIFF
--- a/client/src/wasm/engine_wasm.d.ts
+++ b/client/src/wasm/engine_wasm.d.ts
@@ -268,6 +268,22 @@ export function load_replay_for_playback(json_str: string): number;
 export function ping(): string;
 
 /**
+ * Issue #5468: non-mutating dry-run of `action` for `actor`. Runs the action on
+ * a throwaway clone (the live `GAME_STATE` is never touched) and returns the
+ * PUBLIC deltas — life-total changes, public-zone object transitions, created
+ * tokens, and objects that ceased to exist — a viewer could observe, for
+ * hover-preview UX ("this kills that", "you take 4").
+ *
+ * Hidden-zone movements never leak: the diff is taken over
+ * `filter_state_for_viewer` snapshots (so any identity the viewer can't see is
+ * already redacted), AND a transition is surfaced only when at least one
+ * endpoint is a public zone (see `engine::game::preview`), so a fully-hidden
+ * hand↔library draw is elided even for the acting player's opponents. Returns
+ * an error string when `action` is malformed or illegal in the current state.
+ */
+export function preview_action_js(actor: number, action: any): any;
+
+/**
  * Project an authoritative seat view from Rust so frontend transports do not
  * need to understand format topology details.
  */
@@ -430,6 +446,7 @@ export interface InitOutput {
     readonly load_card_database: (a: number, b: number) => [number, number, number];
     readonly load_replay_for_playback: (a: number, b: number) => [number, number, number];
     readonly ping: () => [number, number];
+    readonly preview_action_js: (a: number, b: any) => any;
     readonly project_seat_view: (a: number, b: number) => [number, number, number];
     readonly replay_seek_js: (a: number) => [number, number, number];
     readonly resolve_all: (a: number, b: number, c: number, d: number) => [number, number, number];
--- a/crates/engine-wasm/src/lib.rs
+++ b/crates/engine-wasm/src/lib.rs
@@ -10,9 +10,10 @@ use engine::ai_support::{auto_pass_recommended, legal_actions_for_viewer, legal_
 use engine::database::legality::{any_ai_difficulty_is_cedh, validate_cedh_bracket};
 use engine::database::{CardDatabase, CardSearchQuery};
 use engine::game::engine::{
-    apply, resolve_all_fast_forward, ResolveAllCallbackDecision,
+    apply, apply_for_simulation, resolve_all_fast_forward, ResolveAllCallbackDecision,
     ResolveAllFastForwardResult as BatchResolveResult,
 };
+use engine::game::preview::compute_preview_diff;
 use engine::game::{
     can_pair_commanders, deck_copy_limit_for, estimate_bracket, evaluate_deck_compatibility,
     filter_state_for_viewer, finalize_public_state, is_brawl_commander_eligible,
@@ -1210,6 +1211,46 @@ pub fn get_viewer_snapshot_js(player_id: u32) -> JsValue {
     }) {
         Ok(val) => val,
         Err(_) => JsValue::NULL,
+    }
+}
+
+/// Issue #5468: non-mutating dry-run of `action` for `actor`. Runs the action on
+/// a throwaway clone (the live `GAME_STATE` is never touched) and returns the
+/// PUBLIC deltas — life-total changes and public-zone object transitions — a
+/// viewer could observe, for hover-preview UX ("this kills that", "you take 4").
+///
+/// Hidden-zone movements never leak: the diff is taken over
+/// `filter_state_for_viewer` snapshots AND restricted to public zones (see
+/// `engine::game::preview`), so draws, random discards from hand, and library
+/// shuffles are elided even for the acting player's opponents. Returns an error
+/// string when `action` is malformed or illegal in the current state.
+#[wasm_bindgen]
+pub fn preview_action_js(actor: u8, action: JsValue) -> JsValue {
+    let action: GameAction = match serde_wasm_bindgen::from_value(action) {
+        Ok(a) => a,
+        Err(e) => {
+            return JsValue::from_str(&format!("Engine error: failed to deserialize action: {e}"));
+        }
+    };
+    let actor = PlayerId(actor);
+    match with_state(|state| {
+        // Simulate on a throwaway clone. `apply_for_simulation` is the same rules
+        // resolution the AI look-ahead uses; it mutates only `sim`, never the
+        // live `GAME_STATE` this closure borrows immutably.
+        let mut sim = state.clone();
+        engine::game::layers::flush_layers(&mut sim);
+        let before = filter_state_for_viewer(&sim, actor);
+        match apply_for_simulation(&mut sim, actor, action) {
+            Ok(_) => {
+                let after = filter_state_for_viewer(&sim, actor);
+                Ok(compute_preview_diff(&before, &after))
+            }
+            Err(e) => Err(format!("Engine error: {e}")),
+        }
+    }) {
+        Ok(Ok(diff)) => to_js(&diff),
+        Ok(Err(msg)) => JsValue::from_str(&msg),
+        Err(e) => e,
     }
 }
 
--- a/crates/engine-wasm/src/lib.rs
+++ b/crates/engine-wasm/src/lib.rs
@@ -1216,14 +1216,16 @@ pub fn get_viewer_snapshot_js(player_id: u32) -> JsValue {
 
 /// Issue #5468: non-mutating dry-run of `action` for `actor`. Runs the action on
 /// a throwaway clone (the live `GAME_STATE` is never touched) and returns the
-/// PUBLIC deltas — life-total changes and public-zone object transitions — a
-/// viewer could observe, for hover-preview UX ("this kills that", "you take 4").
+/// PUBLIC deltas — life-total changes, public-zone object transitions, created
+/// tokens, and objects that ceased to exist — a viewer could observe, for
+/// hover-preview UX ("this kills that", "you take 4").
 ///
 /// Hidden-zone movements never leak: the diff is taken over
-/// `filter_state_for_viewer` snapshots AND restricted to public zones (see
-/// `engine::game::preview`), so draws, random discards from hand, and library
-/// shuffles are elided even for the acting player's opponents. Returns an error
-/// string when `action` is malformed or illegal in the current state.
+/// `filter_state_for_viewer` snapshots (so any identity the viewer can't see is
+/// already redacted), AND a transition is surfaced only when at least one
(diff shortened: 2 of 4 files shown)


matthewevans (changes requested):
Thanks for this. The design here is right, and I want to be specific about why before I get to the two things that need to change.

## What's right

**The seam choice is correct and deliberate.** Putting the diff logic in `engine::game::preview` rather than inline in the WASM bridge means `compute_preview_diff` is unit-tested natively in the engine crate. CI does not compile tests across the WASM boundary, so logic that lives only in `engine-wasm/src/lib.rs` ships untested. You avoided a known trap in this repo, and the three tests at the bottom of `crates/engine/src/game/preview.rs` are exactly the ones I'd want to see.

**`preview_action_js` genuinely does not mutate live state.** It clones into a throwaway `sim`, calls `flush_layers`, and drives the action through `apply_for_simulation` (`crates/engine-wasm/src/lib.rs:44-47`). `apply_for_simulation` is the same rules resolution the AI look-ahead uses, so you get full SBA and trigger handling. It is the correct simulation seam, and picking it over the display sweep is the right call.

**Both snapshots pass through `filter_state_for_viewer`** (`crates/engine-wasm/src/lib.rs:46` and `:49`), so the module's stated precondition is honored at its only call site.

**The life-delta sign is correct.** `compute_preview_diff` binds `a` to `after.players` (`preview.rs:65`) and computes `a.life - b.life` (`preview.rs:70`), so losing life yields a negative delta, matching the `LifeDelta` doc comment. I checked this specifically because it is easy to invert.

**`CR 400.2` is cited accurately.** The rule does enumerate graveyard, battlefield, stack, exile, ante, and command as public zones, and library and hand as hidden. Since the `Zone` enum has no ante variant, `zone_is_public`'s `!matches!(Zone::Hand | Zone::Library)` at `preview.rs:52` is an exact encoding of the rule rather than an approximation. Output ordering is deterministic by id, which the client will appreciate.

## Blocking: the export is unreachable from the frontend

This PR adds `#[wasm_bindgen] pub fn preview_action_js` without regenerating `client/src/wasm/engine_wasm.d.ts`.

That file is a committed, tracked artifact in this repo, not a build output. I confirmed it is tracked by git and not gitignored. It currently declares 48 exports, including `export function get_viewer_snapshot_js(player_id: number): any;` at line 187, and it contains no `preview_action_js`. Nothing under `client/src/` references the new function either.

So the frontend cannot call it, and the committed type surface has drifted from the Rust surface. CI is green because the WASM job compiles Rust and no client code references the export yet. Green checks are necessary here and they are not sufficient. Please regenerate and commit the `.d.ts` as part of this PR.

## Blocking: publicly observable transitions are silently dropped

`compute_preview_diff` only emits a `ZoneChange` when both ends are public (`preview.rs:84`):

```rust
Some(b) if b.zone != a.zone && zone_is_public(b.zone) && zone_is_public(a.zone) => {
```

The consequence is that casting a creature (hand to battlefield) and bouncing a permanent (battlefield to hand) both preview as nothing at all. For a hover-preview feature, previewing your own creature spell and seeing no board change is a visible hole.

The rationale in the comment, that eliding these prevents revealing a hidden card's identity, does not hold for the acting player. `filter_state_for_viewer(&sim, actor)` keeps the actor's own hand, and a spell cast from hand is publicly observable anyway. A departure from a public zone is public information even when the destination is hidden, and an arrival into a public zone is public information even when the origin was hidden.

I'd suggest keying the decision on whether the object's identity is already public to the viewer, rather than requiring both endpoints to be public zones.

## Blocking: objects that cease to exist are invisible to the diff

I want to flag that I went looking for this one and then confirmed it in the code, so please treat it as a finding rather than a question, though I'd still like you to verify it with a test.

`compute_preview_diff` iterates only `after.objects` (`preview.rs:79`) and looks each id up in `before.objects`. An object present in `before` and absent from `after` is never visited, so it can appear in neither `zone_changes` nor `created`. `PreviewDiff` (`preview.rs:43`) has no field that could represent a removal in the first place.

This is reachable. `check_token_cease_to_exist` in `crates/engine/src/game/sba.rs:1800` is production code, called from `check_state_based_actions` at `sba.rs:215`, and it calls `state.objects.remove(&obj_id)` at `sba.rs:1820` for any token outside the battlefield and stack (`CR 111.7` / `CR 704.5d`). Because `apply_for_simulation` runs the full action boundary including state-based actions, a token that dies during the previewed action is gone from `after.objects` entirely.

The result is that "this kills that" previews nothing when the victim is a token, which is the headline use case for this feature. A nontoken creature dying is fine, since battlefield to graveyard has both ends public and is reported correctly.

Please add a test that previews an action killing a token, and consider whether `PreviewDiff` needs a `ceased` (or similarly named) vector alongside `created`, populated by also walking `before.objects` for ids missing from `after`.

## Non-blocking

The module doc says callers MUST pass `filter_state_for_viewer` outputs. That precondition is enforced only by a comment today. The single call site honors it, so nothing is broken right now. A newtype such as `ViewerState(GameState)` would make the invariant unbreakable rather than conventional. This matters a little more than usual here because `CR 400.2` explicitly carves out face-down cards in otherwise public zones, and `ZoneChange` carries `name`.

---

To be clear about where this stands: the design is right, and both blockers are mechanical. Regenerate the `.d.ts`, widen the transition predicate, and cover the token-death case. This should turn around quickly.

davion-knight:
Thanks — that's an unusually thorough read, and all three were spot on. Pushed a revision addressing each:

**1. `.d.ts` regenerated.** Built `engine-wasm` for `wasm32` and ran `wasm-bindgen --target web` (pinned 0.2.121) to regenerate `client/src/wasm/engine_wasm.d.ts`; it now declares `preview_action_js` (export + `InitOutput` entry), so the type surface matches the Rust surface again. The diff is exactly the new function — no churn. (`_bg.wasm`/`.js` are gitignored, so only the `.d.ts` is committed.)

**2. Widened the transition predicate.** You're right that requiring both endpoints public was wrong for the acting player. It's now "at least one endpoint is public": a departure from a public zone (bounce) and an arrival into one (cast from hand, mill) are surfaced; only fully-hidden hand↔library moves (draws) are elided. The snapshots remain `filter_state_for_viewer` outputs, so a hidden identity stays redacted even when it lands in a public zone — `ZoneChange.name` comes from the filtered state. Added a test covering cast + bounce reported / draw elided.

**3. Ceased objects.** Confirmed exactly as you described — `compute_preview_diff` only walked `after.objects`, so a token removed by SBAs (`sba.rs:1820`, CR 111.7 / 704.5d) vanished from the diff. Added a `ceased: Vec<CeasedObject>` field populated by walking `before.objects` for public-zone ids missing from `after`, plus a test that previews a token death and asserts it surfaces (and that a hidden-zone disappearance does not).

On the non-blocking `ViewerState(GameState)` newtype: I agree it would make the "callers must pass a filtered state" invariant unbreakable rather than conventional, and the face-down-name point is real. I left it out of this PR to keep the diff focused, but happy to follow up with it (it'd also let `compute_preview_diff` drop the doc-comment precondition) if you'd like it here instead.

matthewevans (approved):
Approved. The follow-up closes both blockers from my earlier review, and it also closes something I had only raised as a question.

**The `.d.ts` is genuinely regenerated.** `client/src/wasm/engine_wasm.d.ts` now carries both halves of the binding: the public `export function preview_action_js(actor: number, action: any): any;` and the internal `readonly preview_action_js: (a: number, b: any) => any;` entry in `InitOutput`. Both land in alphabetical position between `ping` and `project_seat_view`, which is what wasm-bindgen emits. That detail matters. A hand-added export line would have looked identical in a casual read while leaving the `InitOutput` binding absent, and it would have been silently clobbered at the next `wasm-pack` run. The export is reachable from the frontend.

**Why the widened predicate is right.** `compute_preview_diff` now surfaces a transition when `zone_is_public(b.zone) || zone_is_public(a.zone)`. A departure from a public zone is public information even when the destination is hidden: everyone at the table watches the permanent leave the battlefield. An arrival into a public zone is public even when the origin was hidden: everyone watches the creature appear. Requiring both endpoints to be public dropped real information and bought no privacy in exchange, least of all for the acting player, whose own hand `filter_state_for_viewer` retains anyway. Casting a creature and bouncing a permanent both surface now, while the fully hidden hand-to-library move stays elided.

**The `ceased` vector is yours, and it is the good kind of find.** I raised the dying-token case as a question rather than as a finding, because I could not confirm from the diff whether a token that stops existing is actually removed from `after.objects` or merely left in place with a changed zone. You investigated, confirmed it, and built the fix. `CeasedObject` gated on `zone_is_public(b.zone)` gives you "this kills that" for a dying token while a hidden-zone disappearance stays hidden. Turning a maintainer's uncertain hunch into a verified bug and a fix is exactly the behavior worth rewarding here. Please keep doing that rather than taking my questions at face value.

`CR 111.7` and `CR 704.5d` are both grep-verified against `docs/MagicCompRules.txt` and are exactly apt: a token in a zone other than the battlefield ceases to exist, as a state-based action.

Each new behavior carries a paired positive and negative test (`reports_public_endpoint_transitions_but_elides_pure_hidden_moves`, `reports_ceased_token_but_not_hidden_disappearance`), which is what makes the elision claims worth anything. The safety properties I checked in round one still hold: `preview_action_js` clones into a throwaway `sim`, calls `flush_layers`, snapshots through `filter_state_for_viewer` on both the before and after sides, and drives `apply_for_simulation`, which is the correct simulation seam rather than the display sweep. The live `GAME_STATE` is never mutated.

One item to carry forward, offered as a follow-up rather than as a condition of this merge. The module doc states that callers MUST pass `filter_state_for_viewer` outputs, and that precondition currently rests on convention rather than on the type system. The single call site honors it. A newtype wrapping the filtered state would make it unbreakable. It is worth doing because `ZoneChange` carries `name` and `CR 400.2` carves out face-down cards in otherwise public zones, so a future caller that hands `compute_preview_diff` a raw `GameState` would leak a face-down creature's name through a zone that `zone_is_public` correctly reports as public.

Nice work, and thanks for the thorough second pass.

PR #5493, opened 2026-07-10
fix(coverage): expose damage_source_filter in the PreventDamage parse-diff signature

### LINKED ISSUE #5492: coverage-parse-diff sticky is blind to PreventDamage's damage_source_filter (false 'No card-parse changes')
The `coverage-parse-diff` sticky is required review evidence for parser-surface PRs, and the standing rule is that `✓ No card-parse changes detected` on a PR claiming to change parsing is a HIGH inert-fix signal.

That rule just produced a false positive on #5488.

## What happened

#5488 changes how 13 cards parse (Circle of Protection and Rune of Protection cycles), moving `damage_source_filter` from an unqualified `ChosenDamageSource` to `ChosenDamageSource { filter: Some(..) }`. Its sticky reads `✓ No card-parse changes detected`.

The change is real. The sticky cannot see it.

## Root cause

`crates/engine/src/game/coverage.rs` builds the `PreventDamage` effect signature from a hand-picked field list:

```rust
Effect::PreventDamage { .. } => {
    d.push(("amount".into(), format!("{amount:?}")));
    d.push(("scope".into(), format!("{scope:?}")));
    // ... target
}
```

`damage_source_filter` is never pushed, so any change to it is invisible to the parse diff. Note `fmt_target` *does* render the qualifier ("chosen damage source matching ..."); it simply never gets called for this field.

## Why it matters

The sticky's field list is load-bearing review infrastructure, not cosmetics. Any field omitted from a signature makes an entire class of parser change unreviewable by the evidence channel reviewers are told to rely on, and inverts the signal: a correct fix looks inert.

## Suggested fix

Emit `damage_source_filter` in the `PreventDamage` signature. More generally, audit the effect-signature renderers for fields that a parser change can alter but the signature omits. `CreateDamageReplacement`'s `source_filter` is fed by the same `parse_oneshot_source_filter` binding and looks like it has the same gap.

Found while reviewing #5488, which is otherwise sound; its `parse_oracle_text` integration test is what made the change provable.

davion-knight:
Fixes #5492.

## Problem

`effect_details` (`coverage.rs`) builds the `PreventDamage` parse-diff signature from a hand-picked field list that dropped **`damage_source_filter`** (swallowed by `..`). The `coverage-parse-diff` sticky is the *required review evidence* for parser-surface PRs, and `✓ No card-parse changes detected` is treated as a HIGH inert-fix signal — so omitting a field silently makes a whole class of parser change unreviewable and **inverts the signal**: a correct fix reads as inert. This just produced a false negative on #5488 (13 Circle/Rune of Protection cards moving `ChosenDamageSource` → `ChosenDamageSource { filter: Some(..) }`).

## Fix

- Emit `damage_source_filter` in the `PreventDamage` signature (via `fmt_target`, which already renders the "chosen damage source matching …" qualifier — it just never got called for this field).
- Per the issue's "audit the renderers" note, also emit `CreateDamageReplacement`'s **`source_filter`** (fed by the same `parse_oneshot_source_filter` binding, same blind spot) and its **`target_filter`**.

These fields only appear when `Some`, so unqualified cases' signatures are unchanged (no spurious diffs); the engine-source-hash bump self-heals the baseline, as the issue notes.

## Test

`prevent_damage_signature_exposes_damage_source_filter` calls the production `effect_details` and asserts the field is present when set and absent when `None` — reverting the fix fails it (it's the real signature builder, not a re-implementation).

`cargo check -p engine --tests` + `cargo fmt` clean.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -2942,17 +2942,28 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
             amount,
             target,
             scope,
+            damage_source_filter,
             ..
         } => {
             d.push(("amount".into(), format!("{amount:?}")));
             d.push(("target".into(), fmt_target(target)));
             d.push(("scope".into(), format!("{scope:?}")));
+            // CR 615 + CR 614.1a: the source-restriction qualifier (#5492). Omitting
+            // it made a change from unqualified `ChosenDamageSource` to
+            // `ChosenDamageSource { filter: Some(..) }` (the Circle/Rune of
+            // Protection cycles) invisible to the parse-diff sticky — a real parser
+            // change reading as "No card-parse changes".
+            if let Some(f) = damage_source_filter {
+                d.push(("damage_source_filter".into(), fmt_target(f)));
+            }
         }
         Effect::CreateDamageReplacement {
             modification,
             redirect_to,
             redirect_amount,
             combat_scope,
+            source_filter,
+            target_filter,
             redirect_object_filter,
             recipient_object_filter,
             ..
@@ -2968,6 +2979,16 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
             }
             if let Some(cs) = combat_scope {
                 d.push(("combat_scope".into(), format!("{cs:?}")));
+            }
+            // #5492: `source_filter` shares `PreventDamage`'s
+            // `parse_oneshot_source_filter` binding and had the same blind spot;
+            // `target_filter` is likewise parser-alterable. Emit both so any change
+            // to which sources/targets a replacement covers is diff-visible.
+            if let Some(f) = source_filter {
+                d.push(("source_filter".into(), fmt_target(f)));
+            }
+            if let Some(f) = target_filter {
+                d.push(("target_filter".into(), format!("{f:?}")));
             }
             if let Some(f) = redirect_object_filter {
                 d.push(("redirect_object_filter".into(), fmt_target(f)));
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved. This PR proves itself: its own `coverage-parse-diff` sticky is the test.

The sticky on this very PR reports:

```
#### 3 card(s) · ability/PreventDamage · field `damage_source_filter`: `∅` → `chosen damage source matching black`
```

Those cards are Circle of Protection: Black and Rune of Protection: Black — exactly the cards from #5488 whose qualified source filter the sticky could not see. Before this change, `effect_details` destructured `amount`, `target`, and `scope` off `Effect::PreventDamage` and dropped `damage_source_filter` into the `..` rest pattern, so a parse going from unqualified `ChosenDamageSource` to `ChosenDamageSource { filter: Some(..) }` produced a byte-identical signature and the sticky reported "No card-parse changes." That inverted the signal: it made a real parser change read as an inert one. This retroactively confirms #5488 was a genuine parse change, and it removes a blind spot that was silently telling reviewers the opposite of the truth.

Good call going past the filed issue. #5492 only named `PreventDamage`, and you also covered `CreateDamageReplacement::source_filter`, which shares `PreventDamage`'s `parse_oneshot_source_filter` binding and therefore had the identical blind spot. The sticky confirms that sibling gap was real, not theoretical:

```
#### 6 card(s) · ability/CreateDamageReplacement · field `source_filter`: `∅` → `chosen damage source`
```

That row is Jade Monolith, Beacon of Destiny, and General's Regalia among others. Fixing the reported field and leaving its twin broken would have left the same class of bug live on a different effect. You found the class.

The "only push when `Some`" design is the right call. Unqualified prevention keeps a byte-identical signature, so this introduces zero gratuitous baseline churn on cards that were already being reported correctly. Worth noting for anyone reading the sticky here: all 33 signature rows are `∅ → <value>`. Not one is a value-to-value transition. That is the structural tell that this is surfacing a previously hidden field rather than changing any parse.

**On the 78-card / 33-signature sticky on this PR: this is expected one-time baseline churn, not a regression.** It is the direct consequence of exposing a field that was previously omitted from the signature. Once this lands, the baseline absorbs it and reviewers of subsequent PRs will see a clean diff. Please don't let anyone read that number as behavior change.

I checked the `target_filter` field using `format!("{f:?}")` while its two siblings use `fmt_target`, and I am **not** raising it. `target_filter` is `Option<DamageTargetFilter>` (`types/ability.rs:11041`), a distinct type from the `Option<TargetFilter>` carried by `source_filter` and `damage_source_filter`. `fmt_target` takes `&TargetFilter` and cannot accept it. The Debug format is correct here, not an inconsistency.

CR annotations verified against `docs/MagicCompRules.txt`. CR 614.1a ("Effects that use the word 'instead' are replacement effects") and CR 615 (Prevention Effects) both check out and are used consistently with the rest of the codebase.

---

One non-blocking nit, for a follow-up if you'd rather not touch this PR:

`prevent_damage_signature_exposes_damage_source_filter` tests the `PreventDamage` key in both directions — present when `Some`, absent when `None`. That is exactly the right shape, because the absence assertion is what guards the no-churn property. There is no equivalent assertion for the two new `CreateDamageReplacement` keys (`source_filter`, `target_filter`), which have the same both-directions contract and the same regression risk if someone later moves a field back into the `..` rest pattern. Extending the existing test with a `CreateDamageReplacement` case, or filing a follow-up, would close that gap. Not a merge blocker.

Merging. Thanks for the clean fix and for chasing the sibling.

PR #5494, opened 2026-07-10
fix: Senu exile attack trigger returning self attacking (#5285)

### LINKED ISSUE #5285: Senu, Keen-Eyed Protector — **Card currently listed as:** 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 6/6 · fully supported Double feature thi…
Reported in Discord: https://discord.com/channels/1485498006781427802/1523807165838524466/1523807165838524466

**Thread:** Senu, Keen-Eyed Protector
**Discord thread id:** `1523807165838524466`
**Discord message id:** `1523807165838524466`
**Cards:** senu, keen-eyed protector
**Parser status:** has_gaps

## Summary
**Card currently listed as:** 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 6/6 · fully supported

Double feature this time:
**A)** (Screenshot #1)
**Steps to Reproduce:** Att

## Reported text

**Card currently listed as:** 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 6/6 · fully supported

Double feature this time:
**A)** (Screenshot #1)
**Steps to Reproduce:** Attack with two legendary creatures while Senu is exiled. One of the creatures is blocked, the other is unblocked.
**Expected Result:** Senu triggers from exile and comes back to the field attacking.
**Actual Result:** Senu, Keen-Eyed Protector triggers, but the effect fizzles and it remains in exile.

**B)** (Screenshot #2)
**Steps to Reproduce:** Attack with one or more legendary creatures while Senu is exiled. None of them is blocked.
**Expected Result:** Senu triggers from exile and comes back to the field attacking.
**Actual Result:** Senu, Keen-Eyed Protector triggers and it enters the battlefield from exile, but it's not attacking.

## Attachments

- [image.png](https://cdn.discordapp.com/attachments/1523807165838524466/1523807166509744238/image.png?ex=6a4d7383&is=6a4c2203&hm=4a2a2563d3d9ac79be6979c47c5ec031b629164040830dcc521ca193293886c5&)
- [image.png](https://cdn.discordapp.com/attachments/1523807165838524466/1523807167323308072/image.png?ex=6a4d7383&is=6a4c2203&hm=c385e16e412dc2dfa261afa7d35eae7f4b394fa6be46e3f43d360b630c6e8fce&)

### senu, keen-eyed protector
Flying, vigilance
{T}, Exile Senu: You gain 2 life and scry 2.
When a legendary creature you control attacks and isn't blocked, if this card is exiled, put it onto the battlefield attacking.

---
<sub>report_id: `discord:1523807165838524466:1523807165838524466:0` · discord: `1523807165838524466/1523807165838524466`</sub>

andriypolanski:
## Summary

Senu, Keen-Eyed Protector was parsed with the wrong return target (`ParentTarget` instead of `SelfRef`), without `enters_attacking: true`, and could resolve as if the legendary attacker—not Senu—should change zones. This fix parses exile-zone attack triggers correctly and ensures runtime resolution returns Senu from exile onto the battlefield attacking when a legendary creature you control attacks unblocked.

Fixes [#5285](https://github.com/phase-rs/phase/issues/5285).

## Root cause

Three layered bugs:

1. **Parser — destination qualifiers:** `"put it onto the battlefield attacking"` was not recognized as `enters_attacking: true` (bare `" attacking"` suffix missing from `parse_battlefield_entry_qualifiers`).
2. **Parser — trigger lowering:** Off-battlefield intervening-if triggers (`if this card is exiled`) still lowered `"it"` to `ParentTarget`, binding the legendary attacker that satisfied `valid_card` instead of the ability source (CR 608.2c).
3. **Runtime — batched attack seeding:** `seed_batched_attack_parent_targets` overwrote `SelfRef` with attacking creature IDs at stack push/resolution even when the trigger source was in exile.

### Parser (`oracle_effect/mod.rs`, `oracle_effect/lower.rs`)

- `parse_battlefield_entry_qualifiers`: detect bare `" attacking"` on battlefield entry (CR 508.4).
- `strip_return_destination_ext`: strip `"onto/to the battlefield attacking"` before zone dispatch.

### Parser (`oracle_trigger.rs`)

- `trigger_object_pronoun_ref_for_intervening_if` + `rewrite_off_battlefield_source_self_target`: when intervening-if pins source off battlefield, rewrite `ChangeZone` target to `SelfRef` (not `ParentTarget`/`TriggeringSource`).
- `stamp_self_return_origin_from_trigger_condition`: stamp `ChangeZone.origin` from `SourceInZone` intervening-if (e.g. exile).

### Runtime (`game/triggers.rs`, `game/stack.rs`)

- `seed_batched_attack_parent_targets`: skip batched-attack parent seeding when ability source is not on the battlefield, so Senu's `SelfRef` resolves to the exiled card.

### Tests

- Parser: `senu_keen_eyed_protector_exile_attack_trigger_lowers_correctly`, `parse_battlefield_entry_qualifiers_bare_attacking`
- Runtime: `senu_exile_you_attack_unblocked_returns_source_attacking` (`triggers.rs`)
- Integration: `issue_5285_senu_keen_eyed_protector.rs` (positive + negative)

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test -p engine --lib senu_exile_you_attack_unblocked_returns_source_attacking`
- [x] `cargo test -p engine --lib senu_keen_eyed_protector_exile_attack_trigger_lowers_correctly`
- [x] `cargo test -p engine --lib parse_battlefield_entry_qualifiers_bare_attacking`
- [x] `cargo test -p engine --test integration issue_5285` (2/2)

## Risk notes

- Off-battlefield source-return rewrite is gated on `SourceInZone` intervening-if (or equivalent zone-qualified condition), matching the Senu/Cosima class.
- `seed_batched_attack_parent_targets` skip only applies when the trigger source object is not on the battlefield; battlefield batched-attack anaphora unchanged.


matthewevans (changes requested):
Thanks for this. The diagnosis is correct and the tests are the strongest part of the PR. Let me start with what's right, because most of it is.

## What's right

**The core correction is the right one.** I checked Senu's Oracle text on Scryfall: *"When a legendary creature you control attacks and isn't blocked, if this card is exiled, put it onto the battlefield attacking."* The "it" in that sentence is Senu. Rebinding the `ChangeZone` target from `ParentTarget` to `SelfRef` is exactly correct.

**It generalizes to a class.** The `coverage-parse-diff` sticky shows Managorger Phoenix also moving `ChangeZone · target: parent target → self`, and its text (*"if Managorger Phoenix is in your graveyard ... return it to the battlefield"*) is the same grammatical shape. Fixing a category rather than a card is what this repo asks for, and you did it.

**The runtime tests genuinely discriminate.** `crates/engine/tests/integration/issue_5285_senu_keen_eyed_protector.rs` exiles Senu, declares an unblocked legendary attack, then asserts Senu returns to the battlefield, appears in `combat.attackers`, and is untapped. The sibling test asserts the intervening-if holds when Senu is already on the battlefield. Together those pin both directions of the fix.

**Something you may not know, and it matters more than it looks.** That `combat.attackers` assertion is the *only possible* evidence for the `enters_attacking` half of this fix. `coverage.rs` builds the `ChangeZone` parse-diff signature from `origin`, `destination`, and `target`, swallowing `enters_attacking` through a `..` rest pattern. The `coverage-parse-diff` sticky is structurally blind to that field. If you had only shipped the parser change with a snapshot, `enters_attacking` could have been wrong and nothing in CI would have said so. Your runtime assertion is what proves it landed. This is the same blind-spot family as #5492 (fixed for `PreventDamage` in #5493), and I'll file a follow-up for `ChangeZone.enters_attacking`.

**Every CR citation checks out.** I grep-verified `508.4`, `508.1`, `509.1`, `509.2`, `603.4`, `608.2c`, `608.2k`, and `113.6k` against `docs/MagicCompRules.txt`. `CR 113.6k` ("A trigger condition that can't trigger from the battlefield functions in all zones it can trigger from") is a particularly good find for an exile-zone trigger.

Two things block merge. Both are about *where the logic lives*, and in both cases you've already written the correct version somewhere else in this same diff.

---

## Blocking 1: verbatim Oracle-text dispatch, and you already wrote the right version

`trigger_object_pronoun_ref_for_condition` (`crates/engine/src/parser/oracle_trigger.rs:7300`) dispatches on four hardcoded Oracle phrases:

```rust
if scan_contains(&lower, "this card is exiled")
    || scan_contains(&lower, "~ is exiled")
    || scan_contains(&lower, "this card is in your graveyard")
    || scan_contains(&lower, "~ is in your graveyard")
```

CLAUDE.md names verbatim Oracle-text matching the single most prohibited pattern in this codebase. It handles the cards in front of you and silently misses the next one.

It is already provably incomplete. The list matches `"in your graveyard"` and misses `"in a graveyard"`. PR #5480 had to handle exactly `"if ~ is in a graveyard, "` and `"if this card is in a graveyard, "` for this same shape, so that variant is known to exist in the card pool today. Extending the list to five phrases just moves the boundary.

The remedy is deletion, because the correct implementation is directly below it. `trigger_object_pronoun_ref_for_intervening_if` inspects the typed `TriggerCondition::SourceInZone { zone }`, checks `zone != Zone::Battlefield`, and recurses through `And`, `Or`, and `Not`. That is the composable, class-covering check, and it will handle "in a graveyard", "in your hand", and every future zone phrasing without another edit. The call site at `oracle_trigger.rs:1097` currently runs the string version first and falls back to the typed one:

```rust
object_pronoun_ref: trigger_object_pronoun_ref_for_condition(condition_text)
    .or_else(|| trigger_object_pronoun_ref_for_intervening_if(&if_condition)),
```

Please drop the `scan_contains` block and rely on the typed `SourceInZone` path alone. If some card's condition text genuinely fails to lower into a typed `TriggerCondition`, that is a parser gap worth fixing at the condition parser (`oracle_nom/condition.rs` is the single authority for game-state conditions) rather than papering over with a phrase list.

---

## Blocking 2: the guard in shared machinery uses the wrong discriminator

`seed_batched_attack_parent_targets` (`crates/engine/src/game/triggers.rs:4281`) now takes `state` and early-returns whenever the source object's `zone != Zone::Battlefield`:

```rust
if state
    .objects
    .get(&ability.source_id)
    .is_some_and(|obj| obj.zone != Zone::Battlefield)
{
    return;
}
```

That changes batched-attack seeding for *every* off-battlefield source, not only for source-return triggers.

Zone answers the question "where is the source." The question you actually need answered is "does the body's anaphor refer to the source or to the attacking creatures." Those two come apart. Consider a graveyard or exile attack trigger whose effect legitimately refers to the attacking creature, something shaped like *"Whenever a legendary creature you control attacks, if this card is in your graveyard, that creature gets +1/+1."* Under this guard it silently loses its parent-target binding, falls back to `source_id`, and pumps the wrong object. No test will catch it, because the failure is a silent misbind rather than a panic.

You already compute the correct discriminator: `object_pronoun_ref`. Please gate on that (or equivalently on whether the `SelfRef` rewrite applied) rather than on the source's zone, so the shared seeding path stays correct for bodies whose anaphor really does mean the attackers.

---

## A question I'd like you to answer (I have not verified this myself)

`rewrite_off_battlefield_source_self_target` (`oracle_trigger.rs:7337`) rewrites only `Effect::ChangeZone { destination: Zone::Battlefield, .. }`. Managorger Phoenix's clause continues *"and it perpetually gets +1/+1"*, which is a different effect and will not be touched by that rewrite.

Does that second "it" also resolve to the Phoenix? And if it does, is that only because Blocking 2's zone guard stripped the parent targets and resolution fell back to `source_id`? If that's the mechanism, the pump is working by accident, and it will break the moment the guard is narrowed the way Blocking 2 asks. Please add a Managorger Phoenix assertion covering the perpetual pump's subject so we learn the answer from a test rather than from a bug report.

---

The diagnosis and the tests here are strong work, and I'd like to get this merged. Both blockers are about where the logic lives rather than whether it is correct, and Blocking 1 should be a deletion: the right code is already in your diff.

matthewevans:
Follow-up on the tooling point in my review: I filed #5495 for the `ChangeZone.enters_attacking` blind spot, so it is tracked rather than left as reviewer folklore.

Nothing for you to do there. It matters for this PR only because it explains why your `combat.attackers` assertion is the sole evidence that the `enters_attacking` half of the fix landed. The sticky on this PR shows the `target` and `from` changes and is structurally incapable of showing the third one.

matthewevans:
Both blockers are resolved, and the way you resolved them is better than what I asked for. One clippy error stands between this and an approval.

**Blocker 1, the verbatim Oracle-text dispatch.** Deleted. `trigger_object_pronoun_ref_for_condition` no longer scans for the four hardcoded phrases, and the typed `trigger_object_pronoun_ref_for_intervening_if` path on `TriggerCondition::SourceInZone` now carries the whole job. That is the composable version, and it covers the "in a graveyard" variant the phrase list would have missed.

**Blocker 2, the zone guard in shared machinery.** Removed, and you went further than I suggested by dropping the `state` parameter from `seed_batched_attack_parent_targets` entirely, restoring its original signature.

What makes that correct is worth naming, because it is the interesting part. The guard existed to stop `ParentTarget` from binding to the attacking creatures for an off-battlefield source. Once the parser rewrites `ParentTarget` to `SelfRef` at lower-time, `effect_uses_parent_target(&ability.effect)` is false for Senu, so the function early-returns on its own. The runtime guard was compensating for an AST that still said `ParentTarget` when it meant the source. Fix the AST and the guard is not needed. Shared batched-attack seeding is now untouched for every other card.

Two things I also noticed and like:

- Broadening `rewrite_off_battlefield_source_self_target` to `Pump`, `PumpAll`, `ApplyPerpetual` and `GenericEffect` answers the Managorger Phoenix question I raised. Its second clause ("and it perpetually gets +1/+1") now resolves through the same typed rewrite rather than by accident through a `source_id` fallback.
- Narrowing the rewrite to `ParentTarget` only, leaving `TriggeringSource` intact, is a correctness refinement I did not ask for and it is right.

**The one blocker left is mechanical.** `Rust lint (fmt, clippy, parser gate)` fails:

```
error: this `if let` can be collapsed into the outer `match`
  --> crates/engine/src/parser/oracle_trigger.rs:7340:17
```

That is the `GenericEffect` arm. Collapse it into the match pattern:

```rust
Effect::GenericEffect { target: Some(t), .. } => rewrite_parent_target(t),
```

`cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` is what CI runs.

Push that and I will approve as soon as the shards are green. Nothing else outstanding.

matthewevans (approved):
Approved. The final head applies exactly the one-line clippy collapse and nothing else.

Both blockers are resolved, and the second one is resolved more thoroughly than I asked for.

**Blocker 1 — verbatim Oracle-text dispatch.** The four-phrase `scan_contains` dispatch in `trigger_object_pronoun_ref_for_condition` is deleted, leaving the typed `trigger_object_pronoun_ref_for_intervening_if` path on `TriggerCondition::SourceInZone` as the sole authority. That version also covers the "in a graveyard" variant the phrase list would have silently missed.

**Blocker 2 — the guard in shared machinery.** The zone guard in `seed_batched_attack_parent_targets` is gone, and you went further by dropping the `state` parameter entirely, restoring the shared function's original signature. This is the most interesting thing in the PR, so it's worth stating why it is correct rather than merely smaller. The guard existed to stop `ParentTarget` from binding to the attacking creatures when the trigger source sits off the battlefield. Once the parser rewrites `ParentTarget` to `SelfRef` at lower-time, `effect_uses_parent_target(&ability.effect)` is false for Senu and the function early-returns on its own. The runtime guard was compensating for an AST that still said `ParentTarget` when what it meant was the source. Fix the AST and the guard evaporates. Shared batched-attack seeding is now untouched for every other card.

Two things you did that I did not ask for, and both are right:

- Broadening `rewrite_off_battlefield_source_self_target` to `Pump`, `PumpAll`, `ApplyPerpetual` and `GenericEffect`. This answers the Managorger Phoenix question I raised as an open question rather than a claim: its "and it perpetually gets +1/+1" clause now resolves through the same typed rewrite instead of arriving at the right answer by accident through a `source_id` fallback.
- Narrowing the rewrite to `ParentTarget` only, so `TriggeringSource` stays intact.

The runtime tests deserve credit. They exile Senu, declare an unblocked legendary attack, and assert that Senu returns to the battlefield, appears in `combat.attackers`, and is untapped, with a sibling test asserting the intervening-if holds when Senu is already on the battlefield. That runtime assertion was the only possible evidence for the `enters_attacking` half of this fix, because `coverage.rs` swallowed `ChangeZone`'s `enters_attacking` field via `..` until issue #5495 was fixed (now merged). Your `combat.attackers` assertion is what proved that fix had landed.

All eight CR citations were grep-verified against `docs/MagicCompRules.txt`. `CR 113.6k` ("A trigger condition that can't trigger from the battlefield functions in all zones it can trigger from") was a particularly good find for an exile-zone trigger.

Thirteen checks green.

PR #5496, opened 2026-07-10
fix(engine): scope AttackersDeclared trigger to attacker + defender (Cunning Rhetoric #4736)

### LINKED ISSUE #4736: Cunning Rhetoric triggers even when I don't attack the controller of it or anyone at all — Gonti here is the controller…
Reported in Discord: https://discord.com/channels/1485498006781427802/1520727614761734264/1520727614761734264

**Thread:** Cunning Rhetoric triggers even when I don't attack the controller of it or anyone at all
**Cards:** cunning rhetoric

## Summary
Gonti here is the controller of Cunning Rhetoric, but each combat he keeps getting the trigger of it despite either of us not swinging his way, it eve

### cunning rhetoric
Whenever an opponent attacks you and/or one or more planeswalkers you control, exile the top card of that player's library. You may play that card for as long as it remains exiled, and you may spend mana as though it were mana of any color to cast it.

---
<sub>report_id: `discord:1520727614761734264:1520727614761734264:0` · discord: `1520727614761734264/1520727614761734264`</sub>

shin-core:
## Summary

Fixes #4736. **Cunning Rhetoric** — *"Whenever an opponent attacks you and/or one or more planeswalkers you control, exile the top card of that player's library. ..."* — fired on **every attack declaration**, regardless of who attacked or who was attacked.

## Root cause (runtime, not parser)

Cunning Rhetoric parses **correctly** to `TriggerMode::AttackersDeclared` with `valid_source = Opponent` (attacker) and `valid_target = Controller` (defender = you). But the runtime matcher was:

```rust
fn match_attackers_declared(event, _trigger, _source_id, _state) -> bool {
    matches!(event, GameEvent::AttackersDeclared { .. })  // ignores the trigger entirely
}
```

It dropped both scopes, so any attack declaration by anyone against anyone fired it. In a 3+ player game, an opponent attacking a **different** player wrongly triggered it.

## Fix

Delegate `match_attackers_declared` to the shared **`matching_attack_events`** — the same authority the sibling `Attacks` mode already uses. It applies:
- the attacking-player scope (`valid_source` = opponent),
- the defending-player scope (`valid_target` / `attack_target` = the controller), and
- the once-per-attack-declaration dedup (CR 508.3d + CR 508.5a).

No new logic — the scoping already existed; this mode just wasn't routed through it.

## Tests

Discriminating integration test in a **4-player** game (`cunning_rhetoric_attacks_you_scope_4736.rs`):
- Opponent attacks the **controller** → Cunning Rhetoric **fires**.
- Opponent attacks a **third player** → it does **NOT fire**.

The negative case **fails before this change** (the trigger fired on any attack). This is the discriminating step.

## Verification — no regressions

- **Full integration suite: 2572 pass**, 0 fail — every other `AttackersDeclared` card still works (Lulu, Stern Guardian; the C17 "when you're attacked" curse cycle).
- `game::trigger_matchers` (254) and `parser::oracle_trigger` (984) unit suites pass.
- `cargo clippy -p engine` clean.

CR 508.3d (once per attack declaration against the scoped defender) + CR 508.5a (the defending player is the one the attack is declared against).

### DIFF
--- a/crates/engine/src/game/trigger_matchers.rs
+++ b/crates/engine/src/game/trigger_matchers.rs
@@ -1808,11 +1808,20 @@ pub(super) fn match_attacks_or_blocks(
 
 pub(super) fn match_attackers_declared(
     event: &GameEvent,
-    _trigger: &TriggerDefinition,
-    _source_id: ObjectId,
-    _state: &GameState,
+    trigger: &TriggerDefinition,
+    source_id: ObjectId,
+    state: &GameState,
 ) -> bool {
-    matches!(event, GameEvent::AttackersDeclared { .. })
+    // CR 508.3d + CR 508.5a: "Whenever an opponent attacks you …"
+    // (`AttackersDeclared` mode — Cunning Rhetoric, Lulu Stern Guardian) must
+    // honor the attacking-player scope (`valid_source`) AND the defending-player
+    // scope (`valid_target`): the trigger fires only when a scoped opponent
+    // declares an attack against the trigger's controller. Delegate to the shared
+    // `matching_attack_events`, which applies both scopes and the once-per-attack
+    // -declaration dedup — the same authority `match_attacks` uses. Previously
+    // this returned `true` for any `AttackersDeclared` event, so an opponent
+    // attacking a *different* player (3+ player games) wrongly triggered it (#4736).
+    !matching_attack_events(event, trigger, source_id, state).is_empty()
 }
 
 /// CR 509.3d: A genuine CR 509 blocker/attacker filter is always an *object*
--- a/crates/engine/tests/integration/cunning_rhetoric_attacks_you_scope_4736.rs
+++ b/crates/engine/tests/integration/cunning_rhetoric_attacks_you_scope_4736.rs
@@ -1,0 +1,101 @@
+//! Issue #4736 — Cunning Rhetoric: "Whenever an opponent attacks you and/or one
+//! or more planeswalkers you control, exile the top card of that player's
+//! library. ..."
+//!
+//! The trigger is scoped to attacks against **you** (the controller). It parses
+//! to `TriggerMode::AttackersDeclared` with `valid_source = Opponent` and
+//! `valid_target = Controller`, but the runtime matcher ignored those scopes and
+//! fired on every attack declaration by anyone against anyone. In a 3+ player
+//! game an opponent attacking a *different* player wrongly triggered it.
+//!
+//! This test discriminates in a 4-player game:
+//!   1. Opponent P1 attacks the controller P0     → trigger fires.
+//!   2. Opponent P1 attacks a third player P2      → trigger does NOT fire.
+//!
+//! CR references:
+//!   - CR 508.3d: "Whenever [a player] attacks [a player]" fires once per attack
+//!     declaration against the scoped defending player, not once per attacker.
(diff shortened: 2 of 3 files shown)


matthewevans (approved):
Approved. This lands at the right seam and it deletes a special case instead of adding one.

`match_attackers_declared` was the last attack matcher answering `matches!(event, GameEvent::AttackersDeclared { .. })` — blanket-true, ignoring every scope on the `TriggerDefinition`. Replacing that with `!matching_attack_events(event, trigger, source_id, state).is_empty()` puts the matcher behind the same authority `match_attacks` already delegates to, so the attacking-player scope (`valid_source`) and the defending-player scope (`valid_target` / `attack_target_filter`) are both honored by construction rather than re-derived here. One fewer bespoke path, and the next trigger mode in this family inherits the scoping for free. That is exactly the shape this codebase asks for.

**The test earns its keep.** The 4-player fixture pins both directions: opponent P1 attacks controller P0 and the trigger fires; P1 attacks a third player P2 and it does not. The positive case is what keeps the negative from passing vacuously — without it, a trigger that never fired at all would satisfy the "does not fire" assertion and the test would be green for the wrong reason. Worth stating plainly that the bug only manifests at 3+ players: a two-player fixture has exactly one possible defending player, so the blanket-true matcher and the scoped matcher agree on every input and the test would have proven nothing. Choosing a 4-player game was the discriminating choice.

**Regression risk checked on your behalf; it does not apply.** The obvious worry is the fallthrough in `matching_attack_events`: `attacker_matches` ends in an `else` arm requiring `*id == source_id`, which would newly restrict (or newly fire) an unscoped trigger. That arm is unreachable for this mode. It is guarded by `valid_card.is_some()` / `valid_source.is_some()` / `valid_target.is_some()`, and at the emission site in `oracle_trigger.rs` the `TriggerMode::AttackersDeclared` branch is gated on `is_opponent_attacks_you`, which requires a `TargetFilter::Typed { controller: Some(ControllerRef::Opponent), .. }` subject. That subject is then assigned unconditionally to `valid_source` (when `subject_is_player`) or to `valid_card` (otherwise), so one of the two guards always holds and the `*id == source_id` arm can never be reached. Both paths run `attack_target_matches`, so the defending-player scope survives either way. Spelling this out so the next reviewer does not have to re-derive it.

**One warning about reading this diff.** The hunk header renders as `@@ ... @@ pub(super) fn match_attacks_or_blocks(`, which is misleading. Git labels a hunk with the nearest *preceding* function signature, not the enclosing one. The edited lines are inside `match_attackers_declared`, which begins immediately after — the removed `_trigger` / `_source_id` / `_state` underscored parameters belonged only to that function. `match_attacks_or_blocks` already delegated correctly on main and is untouched. A future reviewer reading this diff in isolation could conclude you patched `match_attacks_or_blocks` and call it the wrong seam. Nothing to change; flagging it so the record is clear.

**CR annotations verified.** Both citations were grep-checked against `docs/MagicCompRules.txt`:

- CR 508.3d — "An ability that reads 'Whenever [a player] attacks, . . .' triggers if one or more creatures that player controls are declared as attackers." Correct for the once-per-attack-declaration semantics.
- CR 508.5a — the multiplayer defending-player rule, that "defending player" names one specific player rather than all of them. Correct, and it is the rule the old blanket match violated.

CI is green across all required checks. Thanks for the clean, minimal fix and for reaching for the shared authority rather than adding a second scoping implementation.

PR #5499, opened 2026-07-10
fix(parser): parse "conjure a card of your choice from [X] spellbook" as draft-from-spellbook

ismayilovibadet802-svg:
## Summary

The parser recognized `draft a card from [X]'s spellbook` (→ `Effect::DraftFromSpellbook`) but not the semantically identical `conjure a card of your choice from [X]'s spellbook`. Both are the same digital-only Alchemy operation — reveal the source card's spellbook, the controller chooses one card, and create it in the destination zone — so the conjure wording fell through to `Unimplemented` even though the effect and its resolver already exist.

`try_parse_conjure_from_spellbook` maps the conjure-of-your-choice wording onto the same `Effect::DraftFromSpellbook`:

- reuses the shared `parse_conjure_zone` for the destination (`onto the battlefield [tapped]`, `into your hand`/`graveyard`/`library`);
- mirrors `try_parse_spellbook_draft`'s fully-consumed-tail discipline, so an unmodeled rider falls through to `Unimplemented` rather than parsing to a subtly wrong effect;
- needs no resolution change — the resolver already reads the spellbook list from the source object, so the printed source name in the text is irrelevant.

Parser-only; reuses the existing, tested effect + resolver.

## Coverage

Closes the `Effect:conjure` single-gap for **9 cards** — supported cards **31,470 → 31,479** — e.g. Vv'viza, Orbital Overseer; Drover of the Swine; Follow the Tracks; Paradigm Shifter; Saint Elenda; Porcine Portent; Volatile Orbit.

## Verification

- full `cargo test -p engine` suite passes (15,994 tests, 0 failures); 4 new parser tests: onto-the-battlefield, apostrophe source name (`Follow the Tracks's spellbook`), into-your-library, and bare-form (no destination) rejection
- `cargo coverage` after regenerating card data: the 9 cards flip to `supported` (`gap_count` 0)

Model: Claude Opus 4.8

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -7866,6 +7866,13 @@ fn parse_effect_clause_inner(text: &str, ctx: &mut ParseContext) -> ParsedEffect
         return parsed_clause(effect);
     }
 
+    // Digital-only Alchemy: "conjure a card of your choice from [X]'s spellbook
+    // [+ destination]" — the "conjure ... of your choice" wording of the same
+    // draft-from-spellbook operation.
+    if let Some(effect) = try_parse_conjure_from_spellbook(tp) {
+        return parsed_clause(effect);
+    }
+
     // Digital-only: "conjure a duplicate of <reference> into/onto zone" — copies a
     // referenced card. Tried before the named form (disjoint: this requires
     // "a duplicate of", the other "named ").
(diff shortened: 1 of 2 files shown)


matthewevans:
Thanks for this, and welcome. I have reviewed the code and it looks good; I am holding the formal review on evidence rather than on anything you did.

**Why it is held.** Your PR is from a fork by a first-time contributor, so GitHub withheld most of the workflows. Only three checks ran here (AI-CONTRIBUTOR triage labels, Contributor trust, Superagent Security Scan). A branch PR gets twelve. The `CI` run is sitting at `action_required`, which means no Rust was compiled or tested, and the green rollup you see is green over three checks that never touch the compiler. A maintainer needs to click "Approve and run workflows" before I can act on it. Nothing for you to do.

The same restriction is why the `coverage-parse-diff` sticky is missing: fork runs get a read-only token and cannot post it.

**What I verified in the meantime**, so you know the review is real and not a stall:

- The class is real and it is in our pool. `client/public/card-data.json` has 12 occurrences of "conjure a card of your choice", and Saint Elenda, Follow the Tracks, Paradigm Shifter and Volatile Orbit are all present.
- `take_until("spellbook")` is the right move, and it quietly buys you both printed grammars. Scryfall shows the possessive form ("Saint Elenda's spellbook") and the article form ("the Planets Spellbook", "the Three Pigs spellbook"). A `tag("'s spellbook")` would have silently dropped Drover of the Swine, Paradigm Shifter, Volatile Orbit and Vv'viza. Your tests cover both, which is exactly right.
- Reusing `parse_conjure_zone` and mirroring `try_parse_spellbook_draft`'s fully-consumed-tail discipline is the correct instinct. An unmodeled rider falls through to `Unimplemented` instead of parsing to a subtly wrong effect.

**One forward-looking note, not a change request.** Mapping the conjure wording onto `Effect::DraftFromSpellbook` is correct *today* because the resolver hands the controller the entire spellbook as `options`. In Arena, drafting from a spellbook shows three random cards while "conjure a card of your choice" shows the whole book. If anyone later narrows `DraftFromSpellbook` to model the real three-card draft, these seven conjure cards must not follow it. A sentence in the doc comment recording that coupling would save a future reader from a subtle regression. Worth adding while you are here, though it does not block.

Minor: the ` tapped` branch is unreachable for this class, since no printed conjure-from-spellbook card names a tapped battlefield entry. It mirrors the sibling parser, so I am fine leaving it.

ismayilovibadet802-svg:
Thank you, Matt

I really appreciate you re-verifying it against Scryfall, and for looking without delay. I've added the coupling note you suggested: since this mapping relies on `DraftFromSpellbook` handing over the entire spellbook, I've noted that if it were ever changed to a system that only offers three cards (like an actual Arena draft), these conjure-of-your-choice cards must still choose from the whole spellbook rather than follow it.

And I added `take_until("spellbook")` plus the two test cases precisely because of the possessive/article grammar split.

No need to rush the approval; please let me know if there's anything else you'd like changed.

matthewevans:
Good news first: your parser change works exactly as intended. The coverage run's `GAINED` list shows all six spellbook cards newly supported:

```
GAINED — 7 cards newly supported:
  Paradigm Shifter
  Porcine Portent
  Follow the Tracks
  Drover of the Swine
  Charged Conjuration
  Saint Elenda
```

The red check is **not your code** — it's branch drift, and it's mine to clear, not yours.

The `Card data` gate failed on a swallowed-clause diagnostic for six *unrelated* cards:

```
DIAGNOSTIC REGRESSION: swallowed-clause increased from 859 to 862 (+3)
  Hall of the Bandit Lord / Generator Servant / Domri, Chaos Bringer   [Swallow:Condition_If]
  A-Sizzling Soloist / Sizzling Soloist / Trench Behemoth              [Effect:attack]
```

None of those touch the spellbook parser — your diff is two files (`oracle_effect/mod.rs`, `tests.rs`) and doesn't mention attacks, conditions, or mana. What happened is that this branch is **15 commits behind `main`** (`behind_by=15`), and those 15 commits include parser fixes — e.g. #5496 (AttackersDeclared scoping) — that lowered the swallow count for exactly these cards on `main`. Your stale branch still swallows the old way, so the check reads it as a +3 regression. It's a baseline-vs-branch mismatch, not a defect you introduced.

The fix is to bring the branch up to date with `main`, which I'm doing now via update-branch (your PR has "allow maintainer edits" on, so I can do it without asking you to). Heads up: because this is your first contribution, updating the branch will re-trigger the workflow-approval gate, so CI will show "waiting for approval" once more before it runs green.

Substance-wise the PR is already reviewed and I'm happy with it — real card class, composes the existing `DraftFromSpellbook` building block, fail-closed tail guard, and a genuinely discriminating negative test. Once the branch is current and CI is green, it goes straight to the queue. Nothing for you to do here.

matthewevans (approved):
Approved. The spellbook-conjure parser fix (`conjure a card of your choice from <source>`) is well-scoped: it reuses `parse_conjure_zone` and discards the source name (the resolver reads the option list off the source object at `WaitingFor::SpellbookDraft`), with a fail-closed tail guard so it can't over-match. Full CI is green on the drift-cleared branch — the earlier `Card data` coverage delta was stale-baseline drift against main (the 6 flagged cards were fixed by intervening merged PRs), not a regression in this change. Verified: all 13 checks pass including the coverage gate and both Rust shards. Thanks!

PR #5500, opened 2026-07-10
fix(coverage): expose ChangeZone ETB entry qualifiers in the parse-diff signature (Closes #5495)

### LINKED ISSUE #5495: coverage-parse-diff sticky is blind to ChangeZone's enters_attacking (same family as #5492)
Follow-up to #5492, which #5493 fixed for `PreventDamage.damage_source_filter` and `CreateDamageReplacement.source_filter`. The same class of blind spot remains on `ChangeZone`.

## Problem

`effect_details` in `crates/engine/src/game/coverage.rs` builds the `ChangeZone` / `ChangeZoneAll` signature from `origin`, `destination` and `target` only:

```rust
Effect::ChangeZone { origin, destination, target, .. }
| Effect::ChangeZoneAll { origin, destination, target, .. } => {
    if let Some(o) = origin { d.push(("from".into(), fmt_zone(o))); }
    d.push(("to".into(), fmt_zone(destination)));
    if !matches!(target, TargetFilter::None) {
    }
}
```

`enters_attacking` is swallowed by the `..`. A parser change that flips `enters_attacking` from `false` to `true` produces no row in the `coverage-parse-diff` sticky.

## Why it matters

Surfaced on #5494 (Senu, Keen-Eyed Protector). One of that PR's three fixes is teaching `parse_battlefield_entry_qualifiers` to recognize a bare `" attacking"` suffix so "put it onto the battlefield attacking" sets `enters_attacking: true` (CR 508.4). The sticky shows the `target` and `from` changes and cannot show the `enters_attacking` change at all. The only evidence that fix landed is the PR's runtime assertion that Senu appears in `combat.attackers`.

As with #5492, an omitted field makes a class of parser change unreviewable through the evidence channel reviewers are told to rely on.

## Suggested fix

Emit `enters_attacking` (and check whether `enters_tapped` and any sibling entry qualifiers have the same gap) in the `ChangeZone` signature, only when true so unqualified signatures stay byte-identical, matching the "push only when set" pattern #5493 established.

More broadly: the effect-signature renderers are load-bearing review infrastructure. It may be worth a test that asserts every `Effect` variant's parser-alterable fields appear in `effect_details`, so this family stops recurring one field at a time.

minion1227:
Closes #5495.

## Summary

`effect_details` built the `ChangeZone` / `ChangeZoneAll` parse-diff signature from `origin`, `destination` and `target` only — every ETB **entry qualifier** was swallowed by the shared arm's `..`. A parser change flipping one produced **no row** in the `coverage-parse-diff` sticky, so that class of parser change read as "No card-parse changes". Surfaced on Senu, Keen-Eyed Protector (#5494), whose "put it onto the battlefield attacking" (CR 508.4) sets `enters_attacking` with no diff-visible evidence.

This splits the two variants (their qualifier sets diverge) and emits each qualifier **only when set**, so unqualified signatures stay byte-identical — the "push only when set" pattern established by #5493 for `PreventDamage`.

Fields now surfaced:
- **ChangeZone:** `owner_library` (CR 400.7), `enter_transformed` (CR 712.2), `enters_under` (CR 110.2a), `enter_tapped` (CR 614.1), `enters_attacking` (CR 508.4), `up_to` (CR 608.2d), `enter_with_counters` (CR 122.1), `conditional_enter_with_counters` (CR 122.1), `face_down_profile` (CR 708.2a), `enters_modified_if` (CR 614.12).
- **ChangeZoneAll:** `enters_under` (CR 110.2a), `enter_tapped` (CR 110.5b), `enter_with_counters` (CR 122.1), `face_down_profile` (CR 708.2a), `library_position` (CR 401.4), `random_order` (CR 401.4).

## Anchored on
- `crates/engine/src/game/coverage.rs` — `Effect::PreventDamage` arm in `effect_details` (added by the merged #5493, `253fff2ae`): same "emit a parser-alterable field only when set, `format!("{f:?}")` for complex types" shape.
- `crates/engine/src/game/coverage.rs` — `prevent_damage_signature_exposes_damage_source_filter` test (#5493): the new `change_zone_signature_exposes_enters_attacking` mirrors it exactly (build the effect, collect signature keys, assert present-when-set / absent-when-default).

## Gate A
`./scripts/check-parser-combinators.sh` → **exit 0** (no violations). This change adds no parser-dispatch code — it extends a review-signature renderer with typed field reads.

## CR
Every emitted qualifier reuses the CR citation already documented on its field definition in `types/ability.rs` (e.g. `enters_attacking` → CR 508.4 "enters tapped and attacking"; `library_position` → CR 401.4). No new rule interpretation is introduced.

## Verification
- `cargo test -p engine --lib coverage::tests` → **98/98 pass**, including the new `change_zone_signature_exposes_enters_attacking` (reverting the `enters_attacking` emission fails it).
- The full destructure (no `..`) compiles, so if a future field is added to either variant the signature renderer fails to compile until it is considered — a deliberate guard against this family recurring one field at a time.

## Scope / risk
Single file, review-signature tooling only (`effect_details` feeds the `coverage-parse-diff` sticky). **No game-logic, parser, or serialization change** — it does not alter how any card parses or resolves, only what the diff sticky reports. Zero runtime/gameplay impact.

Model: claude-opus-4-8
Tier: Frontier
Thinking: High

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -2585,7 +2585,8 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
                 d.push(("target".into(), fmt_target(target)));
             }
             if *owner_library {
-                d.push(("owner_library".into(), "true".into())); // CR 400.7
+                // CR 400.3: a card going to a library other than its owner's goes to its owner's instead.
+                d.push(("owner_library".into(), "true".into()));
             }
             if *enter_transformed {
                 d.push(("enter_transformed".into(), "true".into())); // CR 712.2
@@ -2594,7 +2595,8 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
                 d.push(("enters_under".into(), format!("{u:?}"))); // CR 110.2a
             }
             if !enter_tapped.is_unspecified() {
-                d.push(("enter_tapped".into(), format!("{enter_tapped:?}"))); // CR 614.1
+                // CR 110.5b: permanents enter untapped unless a spell or ability says otherwise.
+                d.push(("enter_tapped".into(), format!("{enter_tapped:?}")));
             }
             if *enters_attacking {
                 d.push(("enters_attacking".into(), "true".into())); // CR 508.4
@@ -2658,7 +2660,7 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
                 d.push(("library_position".into(), format!("{pos:?}"))); // CR 401.4
             }
             if *random_order {
-                d.push(("random_order".into(), "true".into())); // CR 401.4
+                d.push(("random_order".into(), "true".into()));
             }
         }
         Effect::Dig {
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -2554,17 +2554,28 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
             d.push(("player_a".into(), fmt_target(player_a)));
             d.push(("player_b".into(), fmt_target(player_b)));
         }
+        // #5495 (follow-up to #5492/#5493): every ETB entry qualifier below is
+        // parser-alterable but was swallowed by `..`, so a parser change flipping
+        // one — e.g. `enters_attacking` false→true for "put it onto the battlefield
+        // attacking" (CR 508.4, Senu #5494) — produced no row in the
+        // coverage-parse-diff sticky, reading as "No card-parse changes". The two
+        // variants' qualifier sets diverge, so they are matched separately; each
+        // field is emitted only when set so unqualified signatures stay
+        // byte-identical (the "push only when set" pattern #5493 established).
         Effect::ChangeZone {
             origin,
             destination,
             target,
-            ..
-        }
-        | Effect::ChangeZoneAll {
-            origin,
-            destination,
-            target,
-            ..
+            owner_library,
+            enter_transformed,
+            enters_under,
+            enter_tapped,
+            enters_attacking,
+            up_to,
+            enter_with_counters,
+            conditional_enter_with_counters,
+            face_down_profile,
+            enters_modified_if,
         } => {
             if let Some(o) = origin {
                 d.push(("from".into(), fmt_zone(o)));
@@ -2572,6 +2583,82 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
             d.push(("to".into(), fmt_zone(destination)));
             if !matches!(target, TargetFilter::None) {
                 d.push(("target".into(), fmt_target(target)));
+            }
+            if *owner_library {
+                d.push(("owner_library".into(), "true".into())); // CR 400.7
+            }
+            if *enter_transformed {
+                d.push(("enter_transformed".into(), "true".into())); // CR 712.2
+            }
+            if let Some(u) = enters_under {
+                d.push(("enters_under".into(), format!("{u:?}"))); // CR 110.2a
+            }
+            if !enter_tapped.is_unspecified() {
+                d.push(("enter_tapped".into(), format!("{enter_tapped:?}"))); // CR 614.1
+            }
+            if *enters_attacking {
+                d.push(("enters_attacking".into(), "true".into())); // CR 508.4
+            }
+            if *up_to {
+                d.push(("up_to".into(), "true".into())); // CR 608.2d
+            }
+            if !enter_with_counters.is_empty() {
+                d.push((
+                    "enter_with_counters".into(),
+                    format!("{enter_with_counters:?}"),
+                )); // CR 122.1
+            }
+            if !conditional_enter_with_counters.is_empty() {
+                d.push((
+                    "conditional_enter_with_counters".into(),
+                    format!("{conditional_enter_with_counters:?}"),
+                )); // CR 122.1
+            }
+            if let Some(p) = face_down_profile {
+                d.push(("face_down_profile".into(), format!("{p:?}"))); // CR 708.2a
+            }
+            if let Some(f) = enters_modified_if {
+                d.push(("enters_modified_if".into(), fmt_target(f))); // CR 614.12
+            }
+        }
+        Effect::ChangeZoneAll {
+            origin,
+            destination,
+            target,
+            enters_under,
+            enter_tapped,
+            enter_with_counters,
+            face_down_profile,
+            library_position,
+            random_order,
+        } => {
+            if let Some(o) = origin {
+                d.push(("from".into(), fmt_zone(o)));
+            }
+            d.push(("to".into(), fmt_zone(destination)));
+            if !matches!(target, TargetFilter::None) {
+                d.push(("target".into(), fmt_target(target)));
+            }
+            if let Some(u) = enters_under {
+                d.push(("enters_under".into(), format!("{u:?}"))); // CR 110.2a
+            }
+            if !enter_tapped.is_unspecified() {
+                d.push(("enter_tapped".into(), format!("{enter_tapped:?}"))); // CR 110.5b
+            }
+            if !enter_with_counters.is_empty() {
+                d.push((
+                    "enter_with_counters".into(),
+                    format!("{enter_with_counters:?}"),
+                )); // CR 122.1
+            }
+            if let Some(p) = face_down_profile {
+                d.push(("face_down_profile".into(), format!("{p:?}"))); // CR 708.2a
+            }
+            if let Some(pos) = library_position {
+                d.push(("library_position".into(), format!("{pos:?}"))); // CR 401.4
+            }
+            if *random_order {
+                d.push(("random_order".into(), "true".into())); // CR 401.4
             }
         }
         Effect::Dig {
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
This is a really good PR, and I want to be specific about why before I get to the one thing holding it up.

### What this gets right

**The exhaustive destructuring is the best thing here, and it is a genuine improvement over how #5493 solved the same problem for `PreventDamage`.** That PR added one field alongside a surviving `..`, which fixed the symptom and left the mechanism intact. This one fully destructures both `Effect::ChangeZone` and `Effect::ChangeZoneAll`, so the next field added to either variant is a compile error rather than another silent omission. You converted a recurring class of bug into one the compiler catches for us. That is the durable fix #5495 was actually asking for.

Splitting the two variants was the right call, too. Their qualifier sets genuinely diverge (`ChangeZoneAll` has `library_position` / `random_order`, `ChangeZone` has `enters_attacking` / `enter_transformed` / `conditional_enter_with_counters` / `enters_modified_if`), so a shared arm could never have destructured both. That is a structural reason, not churn.

The "push only when set" discipline is also exactly right, and it keeps unqualified signatures byte-identical so we get no gratuitous baseline noise. It mirrors the pattern #5493 established.

**And the PR self-tests.** Its own `coverage-parse-diff` sticky reports 978 cards across 56 signatures, including `ChangeZone · field enters_attacking: ∅ → true` on 37 cards, `enter_tapped` on 437, `enters_under` on 213, and `enter_transformed` on 88. To be unambiguous for anyone reading this later: **that 978-card sticky is expected one-time baseline churn from surfacing previously hidden fields. It is not a regression.** Reviewers of subsequent PRs will see a clean baseline. The sticky is the proof the change works, at the scale of the blind spot it closed.

The payoff is immediate and concrete. #5494 (Senu, Keen-Eyed Protector) sets `enters_attacking` for "put it onto the battlefield attacking", and until this PR that parser change was invisible to the sticky. Thirty-seven cards sat in that blind spot.

All twelve checks are green.

### Blocking: three CR citations point at rules that say something else

This is the only reason I'm not approving. This repo treats a wrong CR number as worse than no CR number, because it manufactures false confidence that code was checked against a rule it was never checked against, and because a later `grep -rn "CR 400.7"` audit will surface a line that has nothing to do with that rule.

I grep-verified all eleven citations against `docs/MagicCompRules.txt`. Eight are correct and apt: `110.2a`, `508.4`, `608.2d`, `122.1`, `708.2a`, `614.12`, `110.5b`, and `401.4` where it sits on `library_position`. Three are wrong.

**1. `owner_library` cites `CR 400.7`** — `crates/engine/src/game/coverage.rs:2588`

CR 400.7 reads: *"An object that moves from one zone to another becomes a new object with no memory of, or relation to, its previous existence."* That is the incarnation rule. It says nothing about owner-versus-controller libraries.

The governing rule is **CR 400.3**: *"If an object would go to any library, graveyard, or hand other than its owner's, it goes to its owner's corresponding zone."*

**2. `enter_tapped` cites `CR 614.1` in one arm and `CR 110.5b` in the other** — `coverage.rs:2597` vs `coverage.rs:2646`

Same field, two different anchors. CR 614.1 is the generic section header (*"Some continuous effects are replacement effects."*). The correct anchor is the one you already used in the `ChangeZoneAll` arm, **CR 110.5b**: *"Permanents enter the battlefield untapped, unflipped, face up, and phased in unless a spell or ability says otherwise."* Use `110.5b` in both.

**3. `random_order` cites `CR 401.4`** — `coverage.rs:2661`

CR 401.4 reads: *"If an effect puts two or more cards in a specific position in a library at the same time, the owner of those cards may arrange them in any order. That library's owner doesn't reveal the order in which the cards go into the library."*

That is owner-*chosen* ordering, which is the opposite of randomization. `401.4` is correct where you used it on `library_position` and should stay there. For `random_order`, please drop the citation rather than reach for a rule that doesn't say it.

**To keep the ask proportionate:** CLAUDE.md says not to annotate boilerplate, serialization, or plumbing, only code that implements a rule. `effect_details` is a signature renderer, so none of these annotations is strictly required in the first place. **Dropping the three questionable ones is just as acceptable as correcting them.** What isn't acceptable is leaving a citation that points at a rule saying something else.

### Non-blocking suggestion

`change_zone_signature_exposes_enters_attacking` covers `enters_attacking` in both directions, present when set and absent when not. That is the right shape for the test. It covers none of the other ten newly emitted fields, though. Given that the whole premise of this PR is that omitted fields go unnoticed, a table-driven test asserting each field appears when set would fit the theme nicely. Worth considering; I'm not going to hold the PR on it.

### Summary

The exhaustive destructuring is precisely the structural fix #5495 hoped for, and the 978-card sticky proves it works at scale. This should merge as soon as the three citations are corrected or dropped.

minion1227:
Thanks for the careful CR audit — all three citations fixed in ee353a282 (verified against `docs/MagicCompRules.txt`):

1. **`owner_library`** — `CR 400.7` → **`CR 400.3`** ("a card going to a library other than its owner's goes to its owner's"). 400.7 was the incarnation rule as you noted.
2. **`enter_tapped` (ChangeZone arm)** — `CR 614.1` → **`CR 110.5b`**, now matching the `ChangeZoneAll` arm's existing citation. Both arms cite the same rule for the same field.
3. **`random_order`** — **dropped** the citation. Agreed that `CR 401.4` describes owner-*chosen* ordering, the opposite of randomization; it stays on `library_position` where it fits.

Left the table-driven test suggestion out of this pass since you flagged it non-blocking, happy to follow up if you'd like it in-scope. Ready for another look.

matthewevans:
You were first, by two minutes, and you had the right idea.

Here is what happened, plainly. #5501 is a near-identical fix to this one, arrived at independently: the same thirteen fields, the same exhaustive destructuring of both `ChangeZone` and `ChangeZoneAll`, even the same test name. I have landed that one, for two reasons. It carries none of the three misapplied CR citations I flagged in my review above, and it is already green and mergeable while this branch is behind and needs a round trip.

I want to be clear about what was valuable here, because it was not the annotations. It was the structural choice you both made: removing the `..` from both variants so that adding a field to either one becomes a compile error rather than another silent omission. That is what turns this from a patch into a closed hole, and it is why the fix is worth having. Your PR made that call.

The duplicated effort is my fault. I filed #5495 without assigning it or saying whether I intended to pick it up, so two people independently spent time on the same function within two minutes of each other. That is a process failure on my side, not a judgment about your work.

I am not closing this PR. That decision belongs to the human maintainer, and I have surfaced it to them. If you would rather correct the three citations and land yours instead, say so and I will re-review.

matthewevans:
Your corrections are right. I verified all three against `docs/MagicCompRules.txt` at head `ee353a28`:

- `owner_library` no longer cites `CR 400.7`; it now points at `CR 400.3` ("If an object would go to any library, graveyard, or hand other than its owner's, it goes to its owner's corresponding zone").
- `enter_tapped` is consistent on `CR 110.5b` in both arms; the stray `CR 614.1` is gone.
- `random_order` carries no citation, and `CR 401.4` stays on `library_position` where it belongs.

So this PR is now correct and equivalent to #5501, which was already approved and sitting at position 1 in the merge queue when you pushed. I am not going to dequeue a green PR to re-run the same change, so #5501 will land and this one becomes redundant through no fault of yours.

To say it plainly: you were first, your structural call was the right one, and your citations are now correct. The duplicated work happened because I filed #5495 without assigning it. Closing this is the human maintainer's decision, not mine, and I have surfaced it to them.

minion1227:
Thanks — that's a fair and generous read, and I appreciate you separating the structural value from the annotation slip.

Two quick updates in case they tip the decision:

- **The three misapplied CR citations were already corrected in `ee353a282`** (before your comment above), each re-verified against `docs/MagicCompRules.txt`. So the specific blocker your review named is resolved.
- **I've rebased this branch onto latest `main` (`0b998ffd8`)**, so it's no longer behind and needs no round trip — CI is green and it's mergeable.

Taking you up on your re-review offer: this is now citation-clean, rebased, and green, so I'm glad to land it if that's useful. That said, #5501 is substantively identical and already in the merge queue — if it goes through, I've no objection to this closing as a duplicate. Either way the hole gets closed, which is the point. Just flagging that the reasons not to land this one are now both addressed.

matthewevans:
Closing note, and thank you.

#5501 merged at 12:01:03Z, so the fix you wrote is now on `main`. I confirmed it: `coverage.rs` on `main` carries the exhaustive destructure and the `change_zone_signature_exposes_enters_attacking` test at line 10366. Your branch has gone `DIRTY` against it, which is what you would expect when two branches change the same function the same way.

The review tooling suggested I update your branch. I am deliberately not doing that. Rebasing this onto a `main` that already contains the identical change would produce either an empty diff or a conflict resolution that reintroduces nothing. There is no version of this PR that still adds value, and that is a statement about timing rather than about your work.

For the record, since it should be on the record: you opened first, you made the right structural call (removing the `..` from both variants so the compiler catches the next omission), and you corrected all three CR citations promptly and correctly. I verified those corrections against `docs/MagicCompRules.txt` before #5501 merged.

The duplication happened because I filed #5495 without assigning it or saying I wanted it picked up. That is a process failure on my side. I am carrying it forward: issues I want a specific contributor to take will say so.

Closing is the human maintainer's call and I have surfaced it. Nothing further needed from you here.

minion1227:
Closing this as a duplicate. @davion-knight's #5501 landed the same exhaustive-destructure fix for #5495 — all twelve `ChangeZone`/`ChangeZoneAll` entry qualifiers, split arms, push-only-when-set — and merged first, so this is now redundant against `main`.

Thank you @matthewevans for the exceptionally thorough review; it's appreciated even on a PR that won't land. Noted and agreed on all three CR corrections, for the record and anyone auditing later:
- `owner_library` → **CR 400.3** (owner's-corresponding-zone), not CR 400.7 (the new-object/no-memory rule).
- `enter_tapped` → **CR 110.5b** in *both* arms (not CR 614.1, the generic section header).
- `random_order` → **drop the citation** — CR 401.4 is owner-*chosen* ordering, the opposite of randomization; it stays correct only on `library_position`.

Glad the "fully destructure so the next added field is a compile error, not another silent omission" direction was what the issue wanted. Credit to @davion-knight for getting there first.

PR #5501, opened 2026-07-10
fix(coverage): expose ChangeZone/ChangeZoneAll entry qualifiers in the parse-diff signature

### LINKED ISSUE #5495: coverage-parse-diff sticky is blind to ChangeZone's enters_attacking (same family as #5492)

davion-knight:
Fixes #5495. Follow-up to #5492 / #5493.

## Problem

`effect_details` built the `ChangeZone` / `ChangeZoneAll` parse-diff signature from `origin`/`destination`/`target` only — every battlefield-entry qualifier was swallowed by `..`. So a parser change that flips `enters_attacking` false→true (Senu, Keen-Eyed Protector — "put it onto the battlefield attacking", CR 508.4) produces **no row** in the `coverage-parse-diff` sticky, leaving that class of change unreviewable through the evidence channel reviewers are told to rely on.

## Fix

Split the shared arm and emit every entry qualifier a parser change can alter, each **only when active** so a plain `ChangeZone`'s signature is unchanged (no spurious diffs; the engine-source-hash bump self-heals the baseline):

- **ChangeZone**: `owner_library`, `enter_transformed`, `enters_under`, `enter_tapped`, `enters_attacking`, `up_to`, `enter_with_counters`, `conditional_enter_with_counters`, `face_down_profile`, `enters_modified_if`.
- **ChangeZoneAll**: `enters_under`, `enter_tapped`, `enter_with_counters`, `face_down_profile`, `library_position`, `random_order`.

Both patterns are now **exhaustive (no `..`)**, so a future entry field fails to compile here rather than being silently dropped — closing this class of blind spot going forward, per the issue's "audit the renderers" intent, not just the one reported field. (The exhaustive destructure already caught two fields I'd have otherwise missed.)

## Test

`change_zone_signature_exposes_enters_attacking` asserts the field appears when `true` and is absent for a plain ChangeZone, via the real `effect_details` — reverting the fix fails it.

`cargo check -p engine --tests`, `clippy`, `fmt` clean.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -37,7 +37,7 @@ use crate::types::phase::Phase;
 use crate::types::replacements::ReplacementEvent;
 use crate::types::statics::{CostModifyMode, StaticMode};
 use crate::types::triggers::TriggerMode;
-use crate::types::zones::Zone;
+use crate::types::zones::{EtbTapState, Zone};
 use nom::branch::alt;
 use nom::bytes::complete::tag;
 use nom::character::complete::space1;
@@ -2558,13 +2558,16 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
             origin,
             destination,
             target,
-            ..
-        }
-        | Effect::ChangeZoneAll {
-            origin,
-            destination,
-            target,
-            ..
+            owner_library,
+            enter_transformed,
+            enters_under,
+            enter_tapped,
+            enters_attacking,
+            up_to,
+            enter_with_counters,
+            conditional_enter_with_counters,
+            face_down_profile,
+            enters_modified_if,
         } => {
             if let Some(o) = origin {
                 d.push(("from".into(), fmt_zone(o)));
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved. This closes #5495 the way I hoped it would be closed.

The important part is the exhaustive destructuring. Rather than adding `enters_attacking` beside a surviving `..`, you removed the `..` from both `Effect::ChangeZone` and `Effect::ChangeZoneAll`. The next person who adds a field to either variant now gets a compile error instead of another silently omitted signature. That converts a recurring class of bug into one the compiler catches, which is a strictly better outcome than #5493 achieved for `PreventDamage`, where the `..` survived. Splitting the two arms because their qualifier sets genuinely diverge is right too.

The restraint on CR annotations is also correct, and worth calling out because it is the kind of thing that gets "fixed" by a later reviewer who thinks more citations are better. `effect_details` is a signature renderer, so CLAUDE.md's rule that boilerplate and plumbing go unannotated applies. You cite exactly one, `CR 508.4` for `enters_attacking`, which I grep-verified against `docs/MagicCompRules.txt` and which is exact. A wrong CR number is worse than no CR number.

**The PR self-tests.** Its own `coverage-parse-diff` sticky reports 978 cards and 56 signatures:

```
437 card(s) · ChangeZone · field enter_tapped:      ∅ → Tapped
213 card(s) · ChangeZone · field enters_under:      ∅ → You
 88 card(s) · ChangeZone · field enter_transformed: ∅ → true
 37 card(s) · ChangeZone · field enters_attacking:  ∅ → true
```

To be explicit for anyone reading later: that 978-card sticky is expected one-time baseline churn from surfacing fields that were previously hidden, not a behavior change. Subsequent PRs will see a clean baseline. The direct payoff is that #5494 (Senu, Keen-Eyed Protector) sets `enters_attacking` for "put it onto the battlefield attacking", and until now that change produced no row at all. Thirty-seven cards were sitting in that blind spot.

Thirteen checks green, mergeable, clean.

**On the duplicate.** #5500 is a near-identical fix opened two minutes before yours by another contributor, arrived at independently: same field set, same exhaustive destructuring, same test name. I am landing yours because #5500 carries three misapplied CR citations, and because yours is already green and mergeable. The duplicated effort is my fault. I filed #5495 without assigning it or saying whether I wanted it picked up, so two people did the same work. That is on me, not on either of you.

**One non-blocking nit.** `change_zone_signature_exposes_enters_attacking` covers `enters_attacking` in both directions, present when set and absent when not, which is the right shape. It covers none of the other ten fields you now emit. Given that the entire premise of this PR is that an omitted field goes unnoticed, a table-driven test asserting each field appears when set would suit the subject matter. Worth doing whenever someone next touches this function.

PR #5502, opened 2026-07-10
fix: Hall of the Bandit Lord creature-spell haste mana rider (#5283)

### LINKED ISSUE #5283: Hall of the Bandit Lord — Card currently listed as: 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 4/4 · 1 gap Steps to Reproduce: tap Hall of th…
Reported in Discord: https://discord.com/channels/1485498006781427802/1523793524447903894/1523793524447903894

**Thread:** Hall of the Bandit Lord
**Discord thread id:** `1523793524447903894`
**Discord message id:** `1523793524447903894`
**Cards:** hall of the bandit lord, hope of ghirapur, senu, keen-eyed protector
**Parser status:** has_gaps

## Summary
Card currently listed as: 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 4/4 · 1 gap

Steps to Reproduce: tap Hall of the Bandit Lord to pay mana to cast Senu, Keen-Eyed Protec

## Reported text

Card currently listed as: 🟩🟩🟩🟩🟩🟩🟩🟩🟩🟩 4/4 · 1 gap

Steps to Reproduce: tap Hall of the Bandit Lord to pay mana to cast Senu, Keen-Eyed Protector first, then Hope of Ghirapur on the following turn (tested with both auto tapping and manual tapping). Hovering over the land itself, I'm pretty sure the Hall is granting haste to the land rather than the creature.

Expected Result: the creature casted with the mana has haste and can attack.

Actual Result: 3 life is lost when tapping the HotBL for mana, but the creature doesn't have haste.

## Attachments

- [image.png](https://cdn.discordapp.com/attachments/1523793524447903894/1523793524825395221/image.png?ex=6a4d66cf&is=6a4c154f&hm=5f501531c6d8c28cb30071a935950b077cdd9c1d4464ae47bc4e21a0230d2c1a&)

### hall of the bandit lord
Hall of the Bandit Lord enters tapped.
{T}, Pay 3 life: Add {C}. If that mana is spent on a creature spell, it gains haste.
### hope of ghirapur
Flying
Sacrifice Hope of Ghirapur: Until your next turn, target player who was dealt combat damage by Hope of Ghirapur this turn can't cast noncreature spells.
### senu, keen-eyed protector
Flying, vigilance
{T}, Exile Senu: You gain 2 life and scry 2.
When a legendary creature you control attacks and isn't blocked, if this card is exiled, put it onto the battlefield attacking.

---
<sub>report_id: `discord:1523793524447903894:1523793524447903894:0` · discord: `1523793524447903894/1523793524447903894`</sub>

andriypolanski:
## Summary

Hall of the Bandit Lord's activated mana ability was parsed with an incorrect `Spell` sub-ability (`SelfRef` haste static) instead of a `ManaSpellGrant` on the produced mana. Creature spells cast with its colorless mana therefore did not gain haste.

Fixes [#5283](https://github.com/phase-rs/phase/issues/5283).

## Root cause

The mana-rider parser only recognized subtype-specific grants with an explicit `"until end of turn"` suffix:

`"If that mana is spent on a Dragon creature spell, it gains haste until end of turn."`

Hall of the Bandit Lord prints the broader, permanent form:

`"If that mana is spent on a creature spell, it gains haste."`

`parse_conditional_keyword_grant` declined this clause, so the sequence parser fell through to a generic `If` absorption that lowered haste onto `SelfRef` (the land) rather than folding a spend rider into `Effect::Mana.grants`.

### Parser (`oracle_effect/mana.rs`)

- Extend `parse_conditional_keyword_grant` to accept:
  - bare `"a creature spell"` → `OnlyForSpellType("Creature")`
  - optional `"until end of turn"` suffix → `Duration::UntilEndOfTurn` vs `Duration::Permanent`
- Continuation path already folds matching clauses into `ManaGrant` on the parent `Effect::Mana`.

### Types (`types/mana.rs`)

- Parameterize `ManaSpellGrant::AddKeywordUntilEndOfTurn` with `duration: Duration` (serde-default `UntilEndOfTurn` for existing Dragon-land cards).

### Runtime (`game/casting.rs`)

- `apply_mana_spell_grants` respects per-grant `duration` instead of hardcoding `UntilEndOfTurn`.

### Tests

- Parser: `parses_hall_of_bandit_lord_creature_spell_haste_grant`
- Integration: `issue_5283_hall_of_bandit_lord.rs` (parse shape + end-to-end cast with Hall mana → haste on battlefield)

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test -p engine --lib bandit_lord`
- [x] `cargo test -p engine --lib conditional_mana_keyword_grant`
- [x] `cargo test -p engine --test integration issue_5283` (2/2)

## Risk notes

- Subtype + until-EOT dragon-land pattern unchanged (default duration preserves prior behavior).
- Permanent haste only applies to the new bare-creature-spell phrasing (Hall of the Bandit Lord class).
- Regenerating `card-data.json` will move Hall's rider from `sub_ability` into `grants` and clear the `SwallowedClause` warning.

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -13190,8 +13190,11 @@ fn apply_mana_spell_grants(
         }) {
             continue;
         }
-        if !keyword_grants.iter().any(|(k, _)| k == keyword) {
-            keyword_grants.push((keyword.clone(), duration.clone()));
+        if !keyword_grants
+            .iter()
+            .any(|(k, d)| k == keyword && d == duration.as_ref())
+        {
+            keyword_grants.push((keyword.clone(), duration.as_ref().clone()));
         }
     }
 
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -13173,11 +13173,12 @@ fn apply_mana_spell_grants(
         return;
     };
     let spell_meta = build_spell_meta(state, caster, spell_id);
-    let mut keyword_grants = Vec::new();
+    let mut keyword_grants: Vec<(crate::types::keywords::Keyword, Duration)> = Vec::new();
     for grant in spent_units.iter().flat_map(|unit| unit.grants.iter()) {
         let ManaSpellGrant::AddKeywordUntilEndOfTurn {
             keyword,
             restriction,
+            duration,
         } = grant
         else {
             continue;
@@ -13189,16 +13190,16 @@ fn apply_mana_spell_grants(
         }) {
             continue;
         }
-        if !keyword_grants.contains(keyword) {
-            keyword_grants.push(keyword.clone());
+        if !keyword_grants.iter().any(|(k, _)| k == keyword) {
+            keyword_grants.push((keyword.clone(), duration.clone()));
         }
     }
 
-    for keyword in keyword_grants {
+    for (keyword, duration) in keyword_grants {
         state.add_transient_continuous_effect(
             spell_id,
             caster,
-            Duration::UntilEndOfTurn,
+            duration,
             TargetFilter::SpecificObject { id: spell_id },
             vec![ContinuousModification::AddKeyword { keyword }],
             None,
--- a/crates/engine/src/game/casting_tests.rs
+++ b/crates/engine/src/game/casting_tests.rs
@@ -2425,7 +2425,7 @@ fn conditional_mana_keyword_grant_applies_to_matching_spell() {
         grants: vec![ManaSpellGrant::AddKeywordUntilEndOfTurn {
             keyword: Keyword::Haste,
             restriction: Some(ManaRestriction::OnlyForCreatureType("Dragon".to_string())),
-            duration: crate::types::ability::Duration::UntilEndOfTurn,
+            duration: Box::new(crate::types::ability::Duration::UntilEndOfTurn),
         }],
         expiry: None,
     };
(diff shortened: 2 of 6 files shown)


matthewevans:
CI is red on `52572bb2`, and the failure is real rather than infrastructure. Both failing checks trace to the same clippy lint, so this is one fix, not two.

```
error: large size difference between variants
##[error]   --> crates/engine/src/types/mana.rs:836:1
    |
836 | /   pub enum ManaSpellGrant {
837 | |       /// The spell cast with this mana can't be countered.
838 | |       CantBeCountered,
839 | |       /// CR 106.6 + CR 702.10: If the spell this mana is spent on satisfies
...   |
842 | | /     AddKeywordUntilEndOfTurn {
843 | | |         keyword: Keyword,
844 | | |         #[serde(default, skip_serializing_if = "Option::is_none")]
845 | | |         restriction: Option<ManaRestriction>,
846 | | |         #[serde(default = "default_mana_keyword_grant_duration")]
847 | | |         duration: crate::types::ability::Duration,
```

You added a variant to the mana enum in `crates/engine/src/types/mana.rs` and it now dominates the enum's size, so every value of that type pays for the largest variant. The usual remedy is to box the large payload. `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` is what CI runs, so that reproduces it locally.

I have not reviewed the substance yet, and will once CI is green. Two things I did check while I was here, so you know the premise holds:

- Scryfall confirms the card reads `{T}, Pay 3 life: Add {C}. If that mana is spent on a creature spell, it gains haste.` with no "until end of turn" suffix, which matches your root-cause description exactly.
- Your parse-diff sticky is present and fresh, and the integration test is correctly placed under `crates/engine/tests/integration/` with the `mod` line added.

Ping me when it is green.

matthewevans:
CI is green on `a171541e` and the clippy fix looks right, so the only thing standing in the way now is a merge conflict.

I tried `gh pr update-branch` and GitHub declined: "Cannot update PR branch due to conflicts." That means the conflict is not something the auto-merge can resolve, so it needs a rebase from you. Several PRs landed on `main` in the last hour touching the engine, so this is timing rather than anything wrong with your change.

Please rebase onto current `main` and push. Once `mergeable` goes back to clean I will review the substance. Two things I will be looking at when I do, so you can pre-empt them:

- The commit message says "dedup by duration" alongside the boxing fix. Boxing `Duration` to satisfy `large_enum_variant` is mechanical, but deduplicating grants by duration is a behavior change. I will want to know what happens when the same keyword is granted twice with different durations, and whether a test pins it.
- `casting_tests.rs` grew by 73 lines, which is encouraging. I will be checking that the new tests drive the real cast path rather than constructing a `ManaSpellGrant` by hand, because the bug you are fixing lives in the arrow from the parsed mana rider to the spell that spends the mana.

Nothing else outstanding.

matthewevans (approved):
Approved. Thanks for this one.

I verified the card text against Scryfall before reviewing the parse: **Hall of the Bandit Lord** reads `{T}, Pay 3 life: Add {C}. If that mana is spent on a creature spell, it gains haste`. There is no "until end of turn" suffix on the rider.

## The dedup change is the fix, not a side effect

This is the thing I most want to flag, because your commit message undersells it. You describe the "dedup by duration" change as if it rides along with the mechanical `large_enum_variant` boxing. It doesn't ride along. It *is* the consumption half of the fix.

`apply_mana_spell_grants` in `casting.rs` previously collected a `Vec<Keyword>`, deduped on the keyword alone, and applied every grant with a hardcoded `Duration::UntilEndOfTurn`. A correctly parsed `Permanent` grant was therefore discarded at the point of use — the parser could produce the right AST all day and the board would never see it. Collecting `Vec<(Keyword, Duration)>`, deduping on the pair, and applying each grant with its own duration is a widening of that seam, and it is what lets the parser fix actually reach the battlefield. Without those nine lines the parser change would have been inert.

## Duration assignment is correct

The parser sets `Duration::UntilEndOfTurn` only when the "until end of turn" suffix is present, and `Duration::Permanent` otherwise. Hall therefore grants permanent haste, which matches the printed card.

For your reassurance, since I expect you eyed it: `default_mana_keyword_grant_duration()` returning `UntilEndOfTurn` is a serde default for legacy data. It does not sit on the parse path.

## The integration test drives the real seam

It builds Hall from Oracle text, activates the mana ability and asserts colorless mana in the pool, casts a creature spell with that mana, and asserts `has_keyword(..., &Keyword::Haste)` on the resulting battlefield permanent. A mocked grant could not make that pass. The arrow it traces — parsed rider, to spent mana, to keyword on the permanent — runs directly through where the defect lived.

## Rules and style

`CR 106.6` ("Some spells or abilities that produce mana restrict how that mana can be spent"), `CR 702.10`, and `CR 702.10a` are all grep-verified against `docs/MagicCompRules.txt` and all apt. Nom compliance is clean in the added parser lines.

## Ignore the parse-diff sticky — it is a renderer bug, not a regression

Your parse-diff sticky shows four cards (Hall of the Bandit Lord, Arena of Glory, Generator Servant, Domri, Chaos Bringer) *losing* a `grant Haste` / `grant Riot` sub-ability with no compensating addition. Please do not chase this. It reads like a regression and it is not one.

`ManaSpellGrant` appears zero times in `coverage.rs`, so the signature renderer has no way to display the grant you attached to the produced mana. Only the removal of the old bogus self-affecting sub-ability is visible to it. The positive half of the change is real, and I verified it through your integration test instead. I have filed #5507 for the renderer gap; it is the third instance of that family, after #5492 and #5495.

## Blast radius

Worth calling out: removing the bogus `Spell` sub-ability with its `SelfRef` haste static fixes three sibling cards beyond the one in the title. Nice reach for a targeted change.

Thirteen checks green. Approving and enqueuing.

PR #5503, opened 2026-07-10
fix(parser): fold "<type> card or a <type> card" disjunctions, scoped

### LINKED ISSUE #5331: overlord of the balemurk — Does not return a creature from my graveyard to my hand

davion-knight:
## What

Overlord of the Balemurk (#5331) returns "a non-Avatar creature card **or a** planeswalker card". The type-disjunction splitter only recursed when the second disjunct began with a bare type word, so the **article-led** "or a planeswalker card" was rejected and the planeswalker leg silently dropped — the card could only ever return creatures. ~14 cards share the "… card or a … card" phrasing.

## How (scoped, to preserve a deliberate boundary)

`parse_type_phrase` intentionally leaves an article-led "or a X" as **remainder** so higher layers can fold it (elided-verb "you control an artifact creature or a Plan"; keyword branch "discard a Spirit card or a card with disturb"). Folding blindly broke that. The fold now triggers only for a genuine card-type disjunction:

- **`left_card_suffix`** — the left leg ended in a "card"/"cards" noun (a card-type list, not an elided-verb permanent clause), and
- **not `is_article_led_bare_card`** — the RHS is type-qualified ("a planeswalker card"), not a bare "a card with `<predicate>`" (Shipwreck Sifters keeps its trigger-layer fold).

## Also: color over-distribution (Purphoros, Bronze-Blooded)

CR 105.1 + CR 205.2 — an article-led disjunct is a self-contained noun phrase, so the left leg's leading adjectives (color/supertype/tapped, the `is_adjective_prefix_prop` leg-local set) must not distribute onto it. "red creature card or an artifact creature card" does **not** require the artifact creature to be red; the previous unconditional `distribute_shared_properties` wrongly stamped `HasColor(Red)` on the artifact leg, rejecting colorless artifact creatures (e.g. Ornithopter).

## Verification

- New: `creature_card_or_a_planeswalker_card_keeps_both_disjuncts`, `article_led_card_disjunct_does_not_inherit_left_leg_color`.
- Boundary tests still green: `parse_type_phrase_leaves_article_led_or_rhs_as_remainder`, `discard_trigger_accepts_or_tail_card_filter_branch`, plus the full 7822-test parser suite.
- Parser gap analysis unchanged (3977 unsupported) — no regression. Combinator gate + clippy clean.

Fixes #5331.

### DIFF
--- a/crates/engine/src/parser/oracle_target.rs
+++ b/crates/engine/src/parser/oracle_target.rs
@@ -2457,26 +2457,48 @@ pub fn parse_type_phrase_with_ctx<'a>(
     // "sorcery spell", "creature card". When the core type is already Instant/Sorcery/etc.,
     // the word is informational — consuming it allows suffix parsers (e.g., "that targets only")
     // and event verb parsers to see what follows.
+    // Tracks whether the left leg ended in a "card"/"cards" noun — the discriminator
+    // that a following article-led "or a <type> card" is a card-type disjunction
+    // (Overlord of the Balemurk, #5331) rather than an elided-verb "you control X
+    // or a Y" clause the condition layer folds one level up (which must stay as
+    // `parse_type_phrase` remainder — `parse_type_phrase_leaves_article_led_or_rhs_as_remainder`).
+    let mut left_card_suffix = false;
     if card_type.is_some() && !matches!(card_type, Some(TypeFilter::Card) | Some(TypeFilter::Any)) {
         let rest_trimmed = lower[pos..].trim_start();
         let ws_len = lower[pos..].len() - rest_trimmed.len();
         // CR 108.1: "spell" and "card" are informational suffixes after a typed qualifier.
-        // Longest-match-first ordering (plurals before singular).
-        static REDUNDANT_SUFFIXES: &[&str] = &["spells ", "spell ", "cards ", "card "];
+        // Longest-match-first ordering (plurals before singular). The paired flag
+        // records whether the consumed noun was "card"/"cards" (vs "spell") — a
+        // card-type disjunction whose article-led RHS is a sibling type, not an
+        // elided-verb clause. Data-driven so the discriminator comes from the
+        // matched tag, not from string inspection of the suffix.
+        static REDUNDANT_SUFFIXES: &[(&str, bool)] = &[
+            ("spells ", false),
+            ("spell ", false),
+            ("cards ", true),
+            ("card ", true),
+        ];
         let mut consumed_suffix = false;
-        for suffix in REDUNDANT_SUFFIXES {
+        for (suffix, is_card) in REDUNDANT_SUFFIXES {
             if let Ok((after, _)) = tag::<_, _, OracleError<'_>>(*suffix).parse(rest_trimmed) {
                 let suffix_len = rest_trimmed.len() - after.len();
                 pos += ws_len + suffix_len;
                 consumed_suffix = true;
+                left_card_suffix = *is_card;
                 break;
             }
         }
         if !consumed_suffix {
             // Check end-of-input variants (no trailing space)
-            for suffix in &["spells", "spell", "cards", "card"] {
+            for (suffix, is_card) in &[
+                ("spells", false),
+                ("spell", false),
+                ("cards", true),
+                ("card", true),
+            ] {
                 if rest_trimmed == *suffix {
                     pos += ws_len + suffix.len();
+                    left_card_suffix = *is_card;
                     break;
                 }
             }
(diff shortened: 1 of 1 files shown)


matthewevans:
The bug you found is real, and I verified the premise before looking at anything else. Scryfall confirms Overlord of the Balemurk reads "return a non-Avatar creature card or a planeswalker card from your graveyard to your hand", and `card-data.json` has 46 occurrences of the `"card or a "` shape, so this is a class and not a one-off. The diagnosis in your description is correct too: the comma path uses the article-aware `starts_with_or_article_type_segment` while the bare separator path uses `starts_with_type_word`, which rejects a leading article.

The fix is in the wrong place, and CI is telling you where. Five existing tests fail, and one of them is a deliberate guard rather than an accident:

```
parser::oracle_target::tests::parse_type_phrase_leaves_article_led_or_rhs_as_remainder
  assertion `left == right` failed: article-led or RHS must remain
```

That test asserts `parse_type_phrase("an artifact creature or a Plan")` leaves `" or a Plan"` as remainder. Somebody wrote it on purpose. `parse_type_phrase` has a contract with its callers: when a segment leads with an article after a connector, it stops and hands the rest back, so the caller decides what the connector means. The other four failures are exactly the callers that depend on it:

```
parser::oracle_nom::condition::tests::test_you_control_compound_presence
parser::oracle_nom::condition::tests::test_recipient_is_subtype_disjunction
parser::oracle_effect::tests::bridge_you_control_artifact_and_enchantment
parser::oracle::tests::conditional_modal_max_supports_compound_presence_conditions
```

Two things follow.

**The `else` branch you widened covers `and` as well as `or`.** Overlord needs only `or`. "You control an artifact and an enchantment" is two separate presence requirements, not a type disjunction, so folding an article-led `and` RHS into an `Or` is wrong on its face. That alone probably accounts for `bridge_you_control_artifact_and_enchantment` and `test_you_control_compound_presence`.

**Even restricted to `or`, widening the general splitter breaks the remainder contract.** `"an artifact creature or a Plan"` and `"non-Avatar creature card or a planeswalker card"` are the same shape to `starts_with_or_article_type_segment`, and only one of them should be consumed.

What distinguishes them is the noun. Overlord's disjuncts are both `<type> card`; the guard test's RHS is a bare type with no `card`. Gating the widening on the right-hand segment ending in the same head noun as the left ("card") would take the fourteen `"… card or a … card"` cards without touching `"or a Plan"`. That keeps the contract intact and stays a class-level fix rather than a card-level one. Worth checking whether the comma path already relies on something similar.

Also, `cargo fmt --all` was not run: `Diff in crates/engine/src/parser/oracle_target.rs:11279`.

Your new test `creature_card_or_a_planeswalker_card_keeps_both_disjuncts` is the right test and asserts the right thing. Keep it. When the five existing tests pass alongside it, this is ready.

matthewevans (approved):
Approving. The fix is scoped, data-driven, and rules-correct.

**What it does.** Folds `"<type> card or a <type> card"` disjunctions so the article-led second leg is no longer silently dropped. The recursion is gated on a new `left_card_suffix` discriminator that is carried out of the matched `REDUNDANT_SUFFIXES` tag rather than re-inspected from the string — the discriminator comes from the parse, which is the right seam.

**Why it holds:**
- The deliberate guard test `parse_type_phrase_leaves_article_led_or_rhs_as_remainder` is untouched and still passes: without a `card`/`cards` noun on the left leg, an article-led `or a <type>` stays as remainder, so `"you control an artifact creature or a Plan"` is still folded one level up by the condition layer.
- `is_article_led_bare_card` is nom-combinator based and word-bounds the `card` noun, correctly excluding Shipwreck Sifters' `"or a card with disturb"` (a keyword-membership branch at the trigger layer, not a type union).
- Property distribution is correct: an article-led disjunct is a self-contained noun phrase, so leg-local adjective props no longer distribute onto it. Purphoros, Bronze-Blooded's `"a red creature card or an artifact creature card"` no longer wrongly requires the artifact creature to be red — a colorless Ornithopter is accepted.

**Test discrimination.** Both new tests fail on `main` (the disjunct never recurses there), so they are genuinely discriminating, not coverage theater. Three real cards across three distinct sub-shapes — Overlord of the Balemurk (type union), Purphoros (adjective non-distribution), Shipwreck Sifters (bare-card exclusion) — and Oracle text for all three was verified against `card-data.json`, not from memory.

**Maintainer fixup pushed (736c3c3).** Comment-only. The new grammar note cited `CR 205.3a`, which is *"A card can have one or more subtypes printed on its type line"* (`docs/MagicCompRules.txt:1414`) — that rule does not support a claim about article-led disjuncts, and parser grammar dispatch is not rules code to annotate. Per CLAUDE.md a wrong CR number is worse than none, so I dropped the citation prefix and left the sentence as a plain grammar note. Every other citation is untouched: `CR 105.1 + CR 205.2` matches established in-tree precedent at `oracle_target.rs:3698`, and `CR 108.1` is correct.

Nice work — low-churn, builds for the class rather than the card, and keeps the guard test honest.

PR #5505, opened 2026-07-10
fix(engine): "chosen this way" destroys only the chosen set (Druid of Purification #4780)

### LINKED ISSUE #4780: Druid of Purification bug — When the ETB of [[Druid of Purification]] resolves, all permanents are destroyed instead of…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521831510242623488/1521831510242623488

**Thread:** Druid of Purification bug
**Cards:** druid of purification

## Summary
When the ETB of [[Druid of Purification]] resolves, all permanents are destroyed instead of only the chosen ones.

### druid of purification
When this creature enters, starting with you, each player may choose an artifact or enchantment you don't control. Destroy each permanent chosen this way.

---
<sub>report_id: `discord:1521831510242623488:1521831510242623488:0` · discord: `1521831510242623488/1521831510242623488`</sub>

shin-core:
## Summary

Fixes #4780. **Druid of Purification** — *"When this creature enters, starting with you, each player may choose an artifact or enchantment you don't control. Destroy each permanent **chosen this way**."* — destroyed **every permanent on the battlefield** instead of only the chosen ones.

This turned out to be **three stacked defects**, each fixed at its own layer:

### 1. Parser (`oracle_target.rs`) — the anaphor was dropped
"[each] `<noun>` chosen this way" fell through to a board-wide `Typed(Permanent)` filter — a full board wipe. New arm recognizes an optional `"each "`/`"the "` determiner + head noun + `" chosen this way"` and resolves to the published `TrackedSet` (CR 608.2c), placed before the type-phrase fallthrough alongside the existing `"each of those <type>"` anaphor arm.

### 2. Runtime (`game/effects/mod.rs`) — the choose step never published
`Effect::TargetOnly` (the per-player "choose `<filter>`" step) resolved as a pure no-op — the chosen objects were never published, so even a correct `DestroyAll{TrackedSet}` read an **empty set** and destroyed nothing. `TargetOnly` now publishes its resolved object targets via `publish_tracked_set`. The **chain-extending** publish unions the per-player iterations of the `player_scope` fan-out (`split_player_scope_chain` already correctly runs the choose per player and the destroy once, as the unscoped tail) — and is inert for chains with no tracked-set consumer.

### 3. Runtime (`game/filter.rs`) — the sentinel was matched literally
`matches_target_filter`'s `TrackedSet { id }` arm looked up the **literal parser sentinel** `TrackedSetId(0)` and matched nothing. Its own comment documented the invariant ("every resolver path binds it to a real set before this match") — but direct effect-resolver paths like `DestroyAll` never did, and the `TrackedSetFiltered` **sibling arm already self-resolved the sentinel inline**. Plain `TrackedSet` now self-resolves identically: chain-first, then latest non-empty set — the same order as `targeting::resolve_tracked_set_sentinel`. With no set published, a sentinel still matches nothing (fail-closed, unchanged).

## Class

The "[each player may] choose … `<verb>` each `<noun>` chosen this way" family — the parser arm covers permanent/creature/artifact/enchantment/card nouns, and the runtime fixes heal **every** effect that consumes a `TrackedSet(0)` sentinel through direct filter matching, not just destroy.

## Tests

- **Parser**: `each_permanent_chosen_this_way_produces_tracked_set` — four phrase variants resolve to the tracked set, fully consumed.
- **Runtime (discriminating, drives the full cast pipeline)**: two opponent artifacts; the caster chooses ONE → only the chosen artifact is destroyed; the un-chosen artifact and the caster's own artifact survive. **Fails in both broken states**: the board-wipe parse kills all three; the unpublished/unresolved set kills none.

## Verification — no regressions

Both runtime changes touch shared machinery, so the whole corpus was run:
- **Full integration suite: 2571 passed, 0 failed** (every `TargetOnly` card and `TrackedSet` consumer)
- `game::effects` (1960), `game::filter` (124), `parser::oracle_target` (452), `parser::oracle_effect` (2561) — all green
- `cargo clippy -p engine` clean

CR 608.2c (anaphoric "chosen this way" set) + CR 603.7 (tracked-set membership).

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -3214,7 +3214,29 @@ pub fn resolve_effect(
         Effect::Reveal { .. } => reveal::resolve(state, ability, events),
         Effect::RevealTop { .. } => reveal_top::resolve(state, ability, events),
         Effect::ExileTop { .. } => exile_top::resolve(state, ability, events),
-        Effect::TargetOnly { .. } => Ok(()), // no-op: targeting is established at cast time
+        // CR 608.2c: `TargetOnly` is the "choose <filter>" step — targeting is
+        // established at selection time, so there is nothing to mutate here. But
+        // the chosen object(s) must be PUBLISHED into the chain tracked set so a
+        // downstream "chosen this way" consumer (the `TrackedSet(0)` sentinel —
+        // "Destroy each permanent chosen this way", Druid of Purification #4780)
+        // reads exactly the chosen set. The chain-extending publish unions the
+        // per-player iterations of a `player_scope` fan-out ("starting with you,
+        // each player may choose …"); for chains with no tracked-set consumer the
+        // published set is simply never read.
+        Effect::TargetOnly { .. } => {
+            let chosen: Vec<crate::types::identifiers::ObjectId> = ability
+                .targets
+                .iter()
+                .filter_map(|t| match t {
+                    TargetRef::Object(id) => Some(*id),
+                    TargetRef::Player(_) => None,
+                })
+                .collect();
+            if !chosen.is_empty() {
+                publish_tracked_set(state, chosen);
+            }
+            Ok(())
+        }
         Effect::Choose { .. } => choose::resolve(state, ability, events),
         Effect::OpponentGuess { .. } => opponent_guess::resolve(state, ability, events),
         Effect::SwapChosenLabels { .. } => swap_chosen_labels::resolve(state, ability, events),
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -3245,6 +3245,11 @@ pub fn resolve_effect(
                 })
                 .collect();
             if ability.scoped_player.is_some() && !chosen.is_empty() {
+                // Load-bearing: `publish_tracked_set` EXTENDS the active chain
+                // set rather than minting a new id, which is what accumulates
+                // successive per-player picks into one union for the tail to
+                // read. (A declining player has no object targets — `chosen` is
+                // empty — so "may choose" declines are a no-op here.)
                 publish_tracked_set(state, chosen);
             }
             Ok(())
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -3216,13 +3216,25 @@ pub fn resolve_effect(
         Effect::ExileTop { .. } => exile_top::resolve(state, ability, events),
         // CR 608.2c: `TargetOnly` is the "choose <filter>" step — targeting is
         // established at selection time, so there is nothing to mutate here. But
-        // the chosen object(s) must be PUBLISHED into the chain tracked set so a
-        // downstream "chosen this way" consumer (the `TrackedSet(0)` sentinel —
-        // "Destroy each permanent chosen this way", Druid of Purification #4780)
-        // reads exactly the chosen set. The chain-extending publish unions the
-        // per-player iterations of a `player_scope` fan-out ("starting with you,
-        // each player may choose …"); for chains with no tracked-set consumer the
-        // published set is simply never read.
+        // inside a `player_scope` fan-out iteration ("starting with you, each
+        // player may choose …" — `scoped_player` is bound by the fan-out driver
+        // on BOTH the synchronous and continuation-resumed paths), the chosen
+        // object(s) must be PUBLISHED into the chain tracked set so the
+        // once-after tail's "chosen this way" consumer (the `TrackedSet(0)`
+        // sentinel — "Destroy each permanent chosen this way", Druid of
+        // Purification #4780) reads the UNION of every player's choice via the
+        // chain-extending publish.
+        //
+        // The `scoped_player` gate is load-bearing: an unconditional publish
+        // allocates a fresh `TrackedSetId` on every plain `TargetOnly`
+        // activation (Sword of the Paruns' modal untap picker), monotonically
+        // growing `tracked_object_sets` each loop iteration — which breaks the
+        // combo detector's state-cover check (`loop_states_cover_modulo_growth`
+        // gate 1) and falsely rejects real loop certificates (Marwyn + Sword).
+        // Outside a fan-out there is no multi-player union to accumulate and no
+        // "chosen this way" tail split off by the driver, so nothing reads the
+        // set — single-player chosen-this-way cards route through
+        // `Effect::ChooseObjectsIntoTrackedSet` instead.
         Effect::TargetOnly { .. } => {
             let chosen: Vec<crate::types::identifiers::ObjectId> = ability
                 .targets
@@ -3232,7 +3244,7 @@ pub fn resolve_effect(
                     TargetRef::Player(_) => None,
                 })
                 .collect();
-            if !chosen.is_empty() {
+            if ability.scoped_player.is_some() && !chosen.is_empty() {
                 publish_tracked_set(state, chosen);
             }
             Ok(())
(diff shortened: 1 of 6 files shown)


shin-core:
CI failure diagnosed and fixed in d709e383e.

**What broke:** `drive_combo_14_marwyn_sword` — the combo detector lost the Marwyn + Sword of the Paruns infinite-mana certificate. Sword of the Paruns' {3} modal untap picker is an `Effect::TargetOnly`, so this PR's unconditional publish allocated a fresh `TrackedSetId` on every loop iteration; `tracked_object_sets` grew monotonically and `loop_states_cover_modulo_growth` gate 1 saw state drift each cycle → certificate falsely rejected. (Bisect confirmed: the publish, not the sentinel-resolution change.)

**Fix:** gate the publish on `ability.scoped_player.is_some()` — the `player_scope` fan-out driver binds `scoped_player` on every per-player iteration (both the synchronous loop and the continuation-resumed remaining players), so the "starting with you, each player may choose …" union publish is unchanged for the Druid class, while plain `TargetOnly` activations (modal pickers, fight source pickers) no longer touch tracked-set state at all — strictly less state churn than the first revision.

**Verified locally this time at CI scope:** full lib suite (15,976 — including all `analysis::` corpus tests) + full integration suite (2,571) green, clippy clean; `drive_combo_14_marwyn_sword` restored and the Druid discriminating test still passes.

matthewevans (changes requested):
This is a strong PR and the diagnosis is right. I verified Druid of Purification's Oracle text against `client/public/card-data.json` — "starting with you, each player may choose an artifact or enchantment you don't control. Destroy each permanent chosen this way." — and the two-layer failure you describe is real. `CR 608.2c` is exactly the right citation (verified at `docs/MagicCompRules.txt:2793`: later text may modify the meaning of earlier text; read the whole text and apply the rules of English).

The integration test is the best kind: two opponent artifacts, one chosen, and it fails in **both** broken states — the board-wipe parse kills all three, the unpublished-set runtime kills none. That's genuinely discriminating, not a replay.

I'm blocking on one thing, and it's structural rather than behavioural.

## MED — the new arm is a third, divergent copy of the sentinel ladder, and the comment claims a parity it does not have

The comment on the new `TargetFilter::TrackedSet` arm says it resolves the sentinel

> the same order as `targeting::resolve_tracked_set_sentinel` and the `TrackedSetFiltered` sibling arm below

It matches neither. The three ladders now read:

| where | ladder |
|---|---|
| `targeting.rs:2181` `resolve_tracked_set_sentinel` | chain → **`current_combat_damage_source_filter`** → latest **non-empty** |
| `filter.rs:1843` `TrackedSetFiltered` | **no chain step** → `max_by_key` over **all** sets, **including empty ones** |
| `filter.rs` `TrackedSet` (this PR) | chain → latest **non-empty** |

Two concrete divergences fall out:

1. Your arm drops the `current_combat_damage_source_filter` rung. A `TrackedSetId(0)` sentinel reaching `filter_inner_for_object` during a combat-damage anaphor ("those creatures") now falls through to `latest_tracked_set_id` instead of the combat-damage source filter. I could not construct a card that hits this today, so I am **not** claiming a live bug — but the comment tells the next reader the rung is there.
2. The `TrackedSetFiltered` sibling takes `max_by_key` over *all* sets, so a trailing **empty** set shadows an earlier populated one; your arm's `latest_tracked_set_id` skips empties (`targeting.rs:2133-2140`) and consults `chain_tracked_set_id` first. Same sentinel, same file, two answers.

Per `CLAUDE.md` this is the sibling-cluster smell at the machinery level — three flows implementing one rule is the point at which it gets generalized, and this PR is where the third appeared. It's also the "single authority" rule: sentinel resolution should have exactly one owner.

### What I'd like instead

Extract the id-resolution step into one `pub(crate) fn` in `targeting.rs` — chain-first, then latest non-empty — and have all three call sites use it:

```rust
pub(crate) fn resolve_tracked_set_id(state: &GameState) -> Option<TrackedSetId> {
    state.chain_tracked_set_id.or_else(|| latest_tracked_set_id(state))
}
```

`resolve_tracked_set_sentinel` keeps its `current_combat_damage_source_filter` rung between the two — that rung yields a `TargetFilter`, not a `TrackedSetId`, which is precisely why it can't live in the shared helper and why the divergence was easy to introduce. Say that in the comment rather than claiming parity.

If changing `TrackedSetFiltered` to skip empty sets and consult the chain risks Zimone's Experiment or Living Death, **leave that arm alone in this PR** and say so in the comment — I'll take a follow-up issue for it. What I don't want to merge is a third ladder plus a comment asserting all three agree.

## Notes, not blockers

- The `scoped_player.is_some()` gate on the publish is well-judged, and I appreciate that the comment explains the `loop_states_cover_modulo_growth` interaction rather than hiding it. That's the kind of note that saves the next person a day.
- `publish_tracked_set` (`effects/mod.rs:4179`) extends the chain set rather than minting a new id, so the per-player fan-out unions correctly — I confirmed that's what makes "each player may choose" accumulate. Worth a one-line comment at the call site, since the union is load-bearing and invisible locally.
- `!chosen.is_empty()` correctly makes a declining player a no-op, matching "**may** choose".

Fix the ladder/comment and I'll approve — the rest of this is ready.

shin-core:
Thanks @matthewevans — ladder unified in 128fffca2, exactly as you laid out.

**MED addressed:**
- Extracted `targeting::resolve_tracked_set_id` — chain-first, then latest non-empty — as the single id-resolution authority; the `TrackedSet` arm now routes through it.
- `resolve_tracked_set_sentinel` keeps its own ladder, and the helper's doc says exactly why: its `current_combat_damage_source_filter` rung (CR 510.2) yields a `TargetFilter`, not a `TrackedSetId`, so it can't fold into the id-level helper — stated instead of claimed away.
- `TrackedSetFiltered` is **left untouched**, taking your follow-up offer: unifying it onto the helper would change empty-set shadowing (its `max_by_key` runs over all sets including empty, no chain rung) for the Zimone's Experiment / Living Death class. The arm's comment now carries the honest ladder inventory instead of the false parity claim.

**Notes addressed:** added the load-bearing comment at the publish call site — `publish_tracked_set` EXTENDS the chain set (that's what makes "each player may choose" union across iterations), and empty `chosen` keeps "may" declines a no-op.

Verified at CI scope again: full lib suite (15,976) + Marwyn certificate + the Druid discriminating test + `game::filter`/`game::targeting` suites green, clippy clean.

matthewevans (approved):
Approving. This addresses the prior review round cleanly.

- **`targeting::resolve_tracked_set_id` as single authority.** The chain-first-then-latest-non-empty resolution is now named and lives in one place, rather than being re-derived per call site.
- **The one rung that cannot be unified is correctly identified and explained.** `resolve_tracked_set_sentinel` can't delegate, because its `current_combat_damage_source_filter` rung (CR 510.2) is interleaved *between* the chain and latest rungs and yields a `TargetFilter`, not a `TrackedSetId`. Documenting *why* a would-be delegation is unsound is more valuable than a forced unification would have been.
- **Scope discipline on `TrackedSetFiltered`.** Declining to smuggle the empty-set-shadowing change for the Zimone's Experiment / Living Death class into a bug fix is the right call. Follow-up, not scope creep.
- **`publish_tracked_set` call-site comment** now states that it *extends* the chain set, which is what unions per-player picks.

Verified independently: Druid of Purification's Oracle text matches `client/public/card-data.json`; CR 608.2c and CR 510.2 both grep-verify against `docs/MagicCompRules.txt`; the new integration test is discriminating (fails in both broken states, not just one).

matthewevans:
Follow-up filed as #5512 for the `TrackedSetFiltered` ladder, as promised — the missing chain rung and the empty-set shadowing, with the Zimone's Experiment / Living Death regression coverage the change would need. Thanks for drawing the boundary where you did instead of pulling it into this fix.

Approved and enqueued (position 1).

PR #5508, opened 2026-07-10
feat(parser): parse "gets an additional +N/+M" P/T grants

nickmopen:
## Problem

An `"an additional +N/+M"` additive P/T grant dropped **entirely**: `parse_pt_mod` (and the anthem grant path that delegates to it) did not recognize the `"an additional "` qualifier, so the whole static line failed and produced no continuous grant.

This is the last blocker for a class of real, gated grants — the `as long as <condition>` gating **already works** (via `parse_conditional_static`); only the `"an additional"` wording was missing:

- **Taste for Mayhem** (Hellbent): `Enchanted creature gets an additional +2/+0 as long as you have no cards in hand.`
- **Divine Sacrament** (Threshold): `White creatures get an additional +1/+1 as long as there are seven or more cards in your graveyard.`
- **Patriarch's Desire** (Threshold): `Enchanted creature gets an additional +2/-2 as long as …`
- **Strange Augmentation** (Delirium): `Enchanted creature gets an additional +2/+2 as long as …`

## Fix

Consume the optional `"an additional "` qualifier in `parse_pt_mod` (CR 613.4c). It marks a second Layer 7c modification stacked on the base grant but carries no P/T semantics of its own, so the underlying `+N/+M` parses identically, and the enclosing gate is attached separately by the existing condition path. One clean-file, 7-line change.

## Tests

- `an_additional_pt_grant_parses_with_gate`: the three ability-word-gated cards (Hellbent/Threshold/Delirium) each produce the grant **with a recognized `StaticCondition`** attached (not `Unrecognized`).
- `an_additional_pt_grant_bare_parses`: the ungated form parses too.
- Regression: `parser::oracle` **7634 passed, 0 failed**; `cargo fmt`, clippy (`-D warnings`), parser-combinator + engine-authorities gates all clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/grammar.rs
+++ b/crates/engine/src/parser/oracle_static/grammar.rs
@@ -1580,6 +1580,13 @@ pub(crate) fn parse_pt_mod_with_remainder(input: &str) -> OracleResult<'_, (i32,
 /// nom combinator doesn't handle (it requires explicit +/- signs).
 pub(crate) fn parse_pt_mod(text: &str) -> Option<(i32, i32)> {
     let text = text.trim();
+    // CR 613.4c: consume an optional "an additional " qualifier ("gets an
+    // additional +N/+M" — Taste for Mayhem, Divine Sacrament, Patriarch's Desire,
+    // Strange Augmentation). It marks a second Layer 7c grant stacked on the base
+    // modification but carries no P/T semantics of its own, so the underlying
+    // +N/+M is parsed identically (the enclosing gate is attached separately).
+    let lower = text.to_lowercase();
+    let text = nom_tag_lower(text, &lower, "an additional ").unwrap_or(text);
     // Try the nom combinator first — handles +N/+M, -N/-M, +N/-M patterns.
     if let Ok((_, (p, t))) = nom_primitives::parse_pt_modifier.parse(text) {
         return Some((p, t));
--- a/crates/engine/src/parser/oracle_static/grammar.rs
+++ b/crates/engine/src/parser/oracle_static/grammar.rs
@@ -1584,9 +1584,16 @@ pub(crate) fn parse_pt_mod(text: &str) -> Option<(i32, i32)> {
     // additional +N/+M" — Taste for Mayhem, Divine Sacrament, Patriarch's Desire,
     // Strange Augmentation). It marks a second Layer 7c grant stacked on the base
     // modification but carries no P/T semantics of its own, so the underlying
-    // +N/+M is parsed identically (the enclosing gate is attached separately).
-    let lower = text.to_lowercase();
-    let text = nom_tag_lower(text, &lower, "an additional ").unwrap_or(text);
+    // +N/+M is parsed identically (the enclosing gate is attached separately). The
+    // cheap "an" guard keeps the common no-qualifier call off the lowercase
+    // allocation path.
+    let text = match text.get(..2) {
+        Some(head) if head.eq_ignore_ascii_case("an") => {
+            let lower = text.to_lowercase();
+            nom_tag_lower(text, &lower, "an additional ").unwrap_or(text)
+        }
+        _ => text,
+    };
     // Try the nom combinator first — handles +N/+M, -N/-M, +N/-M patterns.
     if let Ok((_, (p, t))) = nom_primitives::parse_pt_modifier.parse(text) {
         return Some((p, t));
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -27302,3 +27302,55 @@ fn subgoyf_cda_power_is_bare_count_toughness_is_count_plus_one() {
         def.modifications
     );
 }
+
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
The production change is right, and it's at the right seam. Consuming `"an additional "` inside `parse_pt_mod` handles the whole class in one place — all five call sites (`anthem.rs:482,900,956,1140`, `keyword_grant.rs:1312`) inherit it — rather than patching one grammar branch.

I verified the premise rather than taking the comment's word for it. All four cards exist verbatim in `client/public/card-data.json`, and the important detail is that **each one carries a base grant on a preceding line**:

```
Taste for Mayhem      Enchanted creature gets +2/+0.
                      Hellbent — Enchanted creature gets an additional +2/+0 …
Divine Sacrament      White creatures get +1/+1.
                      Threshold — White creatures get an additional +1/+1 …
Strange Augmentation  Enchanted creature gets +1/+1.
                      Delirium — Enchanted creature gets an additional +2/+2 …
Patriarch's Desire    Enchanted creature gets +2/-2.
                      Threshold — Enchanted creature gets an additional +2/-2 …
```

That's what makes discarding the qualifier safe: the two grants become independent Layer 7c modifications and stack additively on their own. `CR 613.4c` checks out (`docs/MagicCompRules.txt:2990`). If the "additional" grant were the *only* grant on a card, dropping the word would still be correct. Good reasoning, and I'm glad the comment spelled it out.

Two things on the tests.

## MED — assertions run against a `Debug` dump instead of the typed AST

`tests.rs` asserts on `.modifications` with real `ContinuousModification::` constructors **278** times. It asserts on a `format!("{:?}")` substring **twice** — both of them are in this PR. That deviation costs real coverage:

- `dump.contains("condition: Some")` passes identically for Hellbent, Threshold, and Delirium. It proves *a* condition attached, never *which* one. All three cases could parse to the same wrong condition and the test stays green.
- `dump.contains("AddPower { value: 2 }")` never checks the paired toughness. Taste for Mayhem is `+2/+0`; the test passes whatever `AddToughness` came out, including a wrong one or none.
- It breaks on any `#[derive(Debug)]` formatting change, which is a maintenance tax the other 278 assertions don't pay.

`parse_static_line_multi` returns `Vec<StaticDefinition>`, so the typed form is right there:

```rust
assert!(statics[0].modifications.contains(&ContinuousModification::AddPower { value: 2 }));
assert!(statics[0].modifications.contains(&ContinuousModification::AddToughness { value: 0 }));
```

and match `statics[0].condition` against the specific `StaticCondition` variant per case rather than a substring. See `tests.rs:21014` for the house style.

## MED — Patriarch's Desire is named in the production comment but never tested

The comment lists four cards. The test covers three. The one it omits is `+2/-2` — the **only** one of the four with mixed signs, and therefore the only one that exercises `parse_pt_modifier`'s `+N/-M` path after the prefix is stripped. That's precisely the case where a bug would hide, and the comment currently claims coverage the test doesn't provide. Please add it, or drop it from the comment.

## LOW — an allocation on every call

`let lower = text.to_lowercase();` runs on every `parse_pt_mod` call, including the overwhelming majority with no `"an additional "` prefix. It's parse-time only so I'm not blocking on it, but a cheap guard before the allocation would be free.

Fix the two MEDs and I'll approve — the parser change itself is exactly what I'd want.

nickmopen:
Thanks for the thorough review — all three addressed in 4113b22d0:

- **MED (Debug dump → typed AST):** both tests now assert on `statics[0].modifications.contains(&ContinuousModification::AddPower/AddToughness { .. })` and match `statics[0].condition` against the specific `StaticCondition::QuantityComparison` variant per case (HandSize·EQ·0 for Hellbent, GraveyardSize·GE·7 for Threshold, DistinctCardTypes·GE·4 for Delirium). A wrong-but-present condition or an unchecked toughness can no longer pass, and there's no more `{:?}` substring dependency.
- **MED (missing Patriarch's Desire):** added it — the `+2/-2` mixed-sign case that exercises `parse_pt_modifier`'s `+N/-M` path after the prefix is stripped.
- **LOW (per-call allocation):** guarded the `to_lowercase()` behind a cheap `text.get(..2).eq_ignore_ascii_case("an")` check, so the common no-qualifier call never allocates.

Local: `parser::oracle_static` 1080 pass; fmt / clippy `-D warnings` / parser-combinator + engine-authorities gates all clean.

matthewevans (approved):
Approved. `parse_pt_mod` now consumes an optional `an additional ` qualifier, unlocking the "gets an additional +N/+M" grant class (Taste for Mayhem, Divine Sacrament, Patriarch's Desire, Strange Augmentation). The qualifier carries no P/T semantics of its own — the second grant stacks additively as its own Layer 7c `ContinuousModification` (CR 613.4c verifies), and the enclosing gate is attached separately — so parsing the underlying +N/+M identically is correct. The cheap `an` head-guard keeps the common no-qualifier call off the lowercase-allocation path. Both prior review blockers are resolved: the tests now assert typed `ContinuousModification` and typed `StaticCondition` variants (not Debug dumps), and the mixed-sign +2/-2 Patriarch's Desire case is a real test that exercises `parse_pt_modifier`'s +N/-M path. Full CI green. Thanks!

Issue #5256, opened 2026-07-07
Dragon bugs — Hellkite Courser -Not putting commander to battlefield on ETB -Removes itself on end step Will add more


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1522692109336055969/1522692109336055969

**Thread:** Dragon bugs
**Discord thread id:** `1522692109336055969`
**Discord message id:** `1522692109336055969`
**Cards:** hellkite courser
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Hellkite Courser 
-Not putting commander to battlefield on ETB
-Removes itself on end step

Will add more

## Reported text

Hellkite Courser 
-Not putting commander to battlefield on ETB
-Removes itself on end step

Will add more

## Attachments

- [game-state-turn-24-2026-07-03T19-47-19-880Z.zip](https://cdn.discordapp.com/attachments/1522692109336055969/1522692109692440809/game-state-turn-24-2026-07-03T19-47-19-880Z.zip?ex=6a4e0249&is=6a4cb0c9&hm=d9ba15735c8b284acb7d6e67bcc9b66135f10f8c798a6717de5cf9949628677d&)
- [image.png](https://cdn.discordapp.com/attachments/1522692109336055969/1522706650778042521/image.png?ex=6a4d6714&is=6a4c1594&hm=86a6bda8f600498243f5bfea36ec7f72670ff7fbb2222cdd7a5af61b255857ea&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

### hellkite courser
Flying
When this creature enters, you may put a commander you own from the command zone onto the battlefield. It gains haste. Return it to the command zone at the beginning of the next end step.

---
<sub>report_id: `discord:1522692109336055969:1522692109336055969:0` · discord: `1522692109336055969/1522692109336055969`</sub>

mike-theDude:
**Triage:** Confirmed against card-data: Hellkite Courser's ETB trigger structurally exists (put a commander onto the battlefield → grant haste → delayed return-to-command-zone at next end step), but the "put a commander you own from the command zone" target is stored as an unconstrained `TargetFilter::Any` with no command-zone origin, rather than a commander-in-command-zone filter.

[...]

PR #5510, opened 2026-07-10
fix(parser): infer the command zone as a move origin

davion-knight:
CR 408.1 + CR 903.6

## Bug

`infer_origin_zone` maps `from your graveyard`/`from exile`/`from your hand`/`from your library` to their zones, but had **no arm for `from the command zone`**, so it returned `None`. A `None` origin makes the `ChangeZone` resolver search the wrong zone for the card, so "move a commander from the command zone" effects never found it.

Fully broken cards (all share the `put your commander into your hand from the command zone` phrasing):

- **Command Beacon** — `{T}, Sacrifice this land: Put your commander into your hand from the command zone.`
- **Netherborn Altar**
- **Road of Return** (second mode)

Partially addressed:

- **Hellkite Courser** ([#5256](https://github.com/phase-rs/phase/issues/5256)) — `put a commander you own from the command zone onto the battlefield` now infers the correct origin. The indefinite `a commander you own` subject filter in the reanimation put-path is a **separate** gap and is not claimed here.

## Fix

Add the `from the command zone` arm to `infer_origin_zone` → `Zone::Command`. `parse_zone_word` already recognizes the multi-word `command zone` token, so definite forms (`put`/`return target commander … from the command zone …`) now parse to a complete `ChangeZone { origin: Command, target: { IsCommander, Owned } }`.

## Verification

- New unit test `infer_origin_zone_handles_command_zone`.
- Full parser suite green: **7821 passed / 0 failed**.
- Parser-combinator gate clean (exit 0).
- Parser gap analysis unchanged (**3977** unsupported) — no regression.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -28393,6 +28393,13 @@ fn infer_origin_zone(lower: &str) -> Option<Zone> {
         || contains_possessive(lower, "from", "library")
     {
         Some(Zone::Library)
+    } else if scan_contains_phrase(lower, "from the command zone") {
+        // CR 408.1 + CR 903.6: The command zone holds a player's commander(s)
+        // until moved. "put a commander you own from the command zone onto the
+        // battlefield" (Hellkite Courser, #5256) reanimates from the command
+        // zone; without this the origin was inferred as None and the commander
+        // was never found.
+        Some(Zone::Command)
     } else if scan_contains_phrase(lower, "graveyard") && !scan_contains_phrase(lower, "from") {
         // CR 404.1: Possessive graveyard references without "from" — e.g.,
         // "exile each opponent's graveyard", "exile target player's graveyard"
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved. `infer_origin_zone` is the right seam — origin inference for a "from <zone>" prepositional phrase (CR 400.7) belongs exactly here, alongside its graveyard/exile/hand/library siblings, and the new arm reuses the same `scan_contains_phrase` helper rather than reaching for string matching.

Verified rather than assumed:

- **Branch ordering is safe.** The new arm sits after `contains_possessive(lower, "from", "hand")`, but that helper matches the *contiguous* phrase `"from <possessive> hand"`. In "put your commander into your hand from the command zone" no such substring exists ("from" is followed by "the command zone"), so the hand arm cannot shadow it. Road of Return's graveyard clause is a separate mode bullet, so the graveyard arm doesn't shadow it either.
- **Class, not card.** Covers Command Beacon, Netherborn Altar, Road of Return, and Hellkite Courser (#5256) — all four verified verbatim against `card-data.json`.
- **CR annotations grep-verify.** CR 408.1 and CR 903.6 both resolve against `docs/MagicCompRules.txt`; `Zone::Command` itself is already CR 408-annotated.
- **Test discriminates.** Without the arm both assertions fall through every branch and `infer_origin_zone` returns `None`, so the test fails on revert.

Seven production lines, existing idiom, four-card class. `quality`.

PR #5511, opened 2026-07-10
fix(coverage): expose Mana effect grants in the parse-diff signature (Closes #5507)

### LINKED ISSUE #5507: coverage-parse-diff is blind to ManaSpellGrant (third instance, after #5492 and #5495)
`ManaSpellGrant` appears zero times in `crates/engine/src/game/coverage.rs`, so a mana ability's parse-diff signature never renders the grants attached to the produced mana.

## Why it matters

Surfaced on #5502 (Hall of the Bandit Lord). That PR moves haste from a bogus `Spell` sub-ability onto a `ManaSpellGrant` on the produced mana. Its sticky shows only the **removals**:

```
2 card(s) · ability/grant Haste · removed: grant Haste (affects=self, duration=until end of turn, ...)
  Examples: Arena of Glory, Generator Servant
1 card(s) · ability/grant Haste · removed: grant Haste (affects=self, ...)
  Examples: Hall of the Bandit Lord
1 card(s) · ability/grant Riot · removed: grant Riot (affects=self, ...)
  Examples: Domri, Chaos Bringer
```

Removals with no compensating addition is the signature of a regression. It is also the signature of a half-rendered effect, and here it is the latter. A reviewer who trusts the sticky would block a correct fix; a reviewer who ignores it would miss a real one. The only way to tell them apart today is to grep `coverage.rs` for the type name.

## Pattern

This is the third instance:

- #5492: `PreventDamage.damage_source_filter` omitted (fixed by #5493)
- #5495: `ChangeZone` entry qualifiers omitted (fixed by #5501, which exhaustively destructured both variants so the compiler catches the next omission)
- this issue: `ManaSpellGrant` on mana abilities, not rendered at all

## Suggested fix

Render `ManaSpellGrant` (keyword, restriction, duration) in the mana-ability signature, emitting fields only when set so unqualified signatures stay byte-identical.

More usefully, adopt #5501's approach generally: exhaustively destructure effect payloads in `effect_details` rather than matching `..`, so a newly added parser-alterable field becomes a compile error instead of a silently unreviewable change. The effect-signature renderers are load-bearing review infrastructure and they keep losing fields one at a time.

minion1227:
Closes #5507.

## Summary

The mana effect's parse-diff signature in `effect_details` rendered only `produced`, swallowing `restrictions` / `grants` / `expiry` / `target` with `..`. So a parser change that attaches a `ManaSpellGrant` to produced mana — e.g. Hall of the Bandit Lord's creature-spell haste rider (#5502) — showed only **removals** in the coverage-parse-diff sticky with no compensating addition: the signature of a regression, when it was actually a half-rendered effect.

This is the third instance of the family (#5492 `PreventDamage` → #5493; #5495 `ChangeZone` → #5501). Following #5501's approach, `Effect::Mana` is now **fully destructured** (no `..`), so a newly added field becomes a compile error rather than another silent omission, and each field is emitted **only when set** so unqualified mana signatures stay byte-identical.

Fields now surfaced: `restrictions`, **`grants`** (the named blind spot — `ManaSpellGrant`), `expiry`, `target`.

## Anchored on
- `crates/engine/src/game/coverage.rs` — the `Effect::ChangeZone`/`ChangeZoneAll` arms (merged #5501): same fully-destructure-and-emit-only-when-set shape, `format!("{:?}")` for complex payloads.
- `crates/engine/src/game/coverage.rs` — `prevent_damage_signature_exposes_damage_source_filter` / `change_zone_signature_exposes_enters_attacking` tests (#5493/#5501): the new `mana_signature_exposes_grants` mirrors them exactly (build the effect, collect signature keys, assert present-when-set / absent-when-empty).

## CR
None. Per CLAUDE.md, `effect_details` is a signature renderer (review plumbing), not rule-implementing code, so no CR annotations are added — consistent with the maintainer note on #5500 that these annotations are not required here (and it avoids a wrong-citation risk).

## Gate A
`./scripts/check-parser-combinators.sh` → **exit 0** (no violations). No parser-dispatch code — this extends a signature renderer with typed field reads.

## Verification
- `cargo test -p engine --lib coverage::tests` → green, including the new `mana_signature_exposes_grants` (grants appears when a `ManaSpellGrant` is set, absent when empty; reverting the emission fails it).
- Full destructure (no `..`) compiles, so any future `Effect::Mana` field is a compile error until it's considered here.

## Scope / risk
Single file, review-signature tooling only (`effect_details` feeds the coverage-parse-diff sticky). **No game-logic, parser, or serialization change** — it does not alter how any card parses or resolves, only what the diff sticky reports. Zero runtime impact.

Model: claude-opus-4-8
Tier: Frontier
Thinking: High

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -2768,8 +2768,35 @@ fn effect_details(effect: &Effect) -> Vec<(String, String)> {
         Effect::ChooseDamageSource { source_filter } => {
             d.push(("source".into(), fmt_target(source_filter)));
         }
-        Effect::Mana { produced, .. } => {
+        Effect::Mana {
+            produced,
+            restrictions,
+            grants,
+            expiry,
+            target,
+        } => {
+            // #5507 (third instance after #5492/#5495): the mana effect rendered
+            // only `produced`, swallowing `restrictions`/`grants`/`expiry`/`target`
+            // with `..`. So a parser change that attaches a `ManaSpellGrant` to the
+            // produced mana (e.g. Hall of the Bandit Lord's creature-spell haste
+            // rider, #5502) never showed a compensating addition in the sticky —
+            // removals-with-no-addition, the signature of a regression, when it was
+            // really a half-rendered effect. Fully destructure (no `..`) so a new
+            // Mana field is a compile error, not another silent omission, and emit
+            // each field only when set so unqualified signatures stay byte-identical.
             d.push(("mana".into(), fmt_mana_production(produced)));
+            if !restrictions.is_empty() {
+                d.push(("restrictions".into(), format!("{restrictions:?}")));
+            }
+            if !grants.is_empty() {
+                d.push(("grants".into(), format!("{grants:?}")));
+            }
+            if let Some(e) = expiry {
+                d.push(("expiry".into(), format!("{e:?}")));
+            }
+            if let Some(t) = target {
+                d.push(("target".into(), fmt_target(t)));
+            }
         }
         Effect::RevealHand {
             target,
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved. Closes #5507, the third instance of this family after #5492/#5495 (fixed by #5493/#5501), and it fixes the class rather than the instance.

The right seam and the right shape: `effect_details` is where the parse-diff signature is rendered, and the fix replaces `Effect::Mana { produced, .. }` with a full destructure of all five fields. I confirmed against `types/ability.rs:10388-10406` that `produced`, `restrictions`, `grants`, `expiry`, and `target` are exactly the field set — so the `..` is genuinely gone and a sixth `Effect::Mana` field becomes a **compile error** instead of a fourth silent omission. That converts a recurring silent-omission class into a compiler-enforced invariant, which is the durable fix.

The conditional emission is the detail that makes it safe: each field is pushed only when non-empty/`Some`, so unqualified mana signatures stay byte-identical and no baseline churn is introduced. The test asserts both directions — `grants` present when set, absent when empty — so it discriminates on revert in both the positive and the no-churn direction.

`quality`: exhaustive destructure over `..`, compile-time enforcement, zero baseline churn, paired positive/negative assertions.

PR #5515, opened 2026-07-10
feat(parser): support "whenever you discover, discover again for the same value"

### LINKED ISSUE #5270: Curator of Sun's Creation — After discovering during my turn with [[Pantlaza, Sun-Favored]], my [[Curator of Sun's Crea…
Reported in Discord: https://discord.com/channels/1485498006781427802/1523170092118966353/1523170092118966353

**Thread:** Curator of Sun's Creation
**Discord thread id:** `1523170092118966353`
**Discord message id:** `1523170092118966353`
**Cards:** curator of sun's creation, pantlaza, sun-favored

## Summary
After discovering during my turn with [[Pantlaza, Sun-Favored]], my [[Curator of Sun's Creation]] did not create a discover trigger after i resolved my first discover.

## Reported text

After discovering during my turn with [[Pantlaza, Sun-Favored]], my [[Curator of Sun's Creation]] did not create a discover trigger after i resolved my first discover.

## Attachments

- [game-state-turn-12-2026-07-05T03-26-22-235Z.zip](https://cdn.discordapp.com/attachments/1523170092118966353/1523170092299587634/game-state-turn-12-2026-07-05T03-26-22-235Z.zip?ex=6a4dc531&is=6a4c73b1&hm=1f73f4251b06f90c369c3554081e2c5176816dc42df4abddedb99428cc8eecdf&)

### curator of sun's creation
Whenever you discover, discover again for the same value. This ability triggers only once each turn.
### pantlaza, sun-favored
Whenever Pantlaza or another Dinosaur you control enters, you may discover X, where X is that creature's toughness. Do this only once each turn. (Exile cards from the top of your library until you exile a nonland card with that mana value or less. Cast it without paying its mana cost or put it into your hand. Put the rest on the bottom in a random order.)

---
<sub>report_id: `discord:1523170092118966353:1523170092118966353:0` · discord: `1523170092118966353/1523170092118966353`</sub>

davion-knight:
Fixes #5270 (Curator of Sun's Creation). CR 701.57a.

"Whenever you discover, discover again for the same value. This ability triggers only once each turn." parsed to an **Unknown** trigger with an **Unimplemented** effect — the ability did nothing. The discover *trigger* (`TriggerMode::Discover` + `match_discover`) and *effect* (`Effect::Discover`) both already exist; this wires the two missing parser paths plus the "same value" quantity.

## 1. Trigger — "whenever you discover"

`whenever/when <player> discover[s]` (bare) → `TriggerMode::Discover` (matched on the `EffectResolved{Discover}` event). The discover keyword action is **not** a `PlayerActionKind`, so a dedicated `is_bare_discover_subject` gate routes it before the `parse_player_action_list` path. A compound `discover, investigate, scry, …` subject (Val, Marooned Surveyor) is deliberately left to the action-list path.

## 2. Effect — "discover again for the same value"

→ `Effect::Discover { mana_value_limit: QuantityRef::TriggeringDiscoverValue, player: Controller }` (composed from tags, not a verbatim match).

## 3. `QuantityRef::TriggeringDiscoverValue` (new)

The mana-value limit `N` of the discover that fired the current "whenever you discover" trigger. Threaded via a transient `GameState::last_discover_value`, set in the discover resolver before the `EffectResolved` event fires; resolves to `0` outside a discover context. Classified per the engine-variant checklist:
- **ability-scan**: `Axes::NONE` (reads a bounded game-state scalar; no growing/sibling/projected axis).
- **read/write profiler**: `empty()` (written only by discover resolution, never a sibling trigger).
- **cast-time board-state classifier**: falls to the `_ => false` trigger-event-context arm (correctly non-evaluable before activation).
- coverage + resolver arms added.

The `OncePerTurn` constraint already parsed, so `Curator` no longer loops on itself.

## Verification

- New tests: `trigger_curator_discover_again_for_the_same_value` (parser), `triggering_discover_value_reads_last_discover_scalar` (resolver).
- **Full engine lib suite: 16026 passed / 0 failed** (touches shared `game/quantity`, `ability_scan`, `ability_rw`, `layers`, `triggers`, `coverage`).
- Combinator gate clean (exit 0); clippy clean; parser gap analysis unchanged.

### DIFF
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -2034,6 +2034,7 @@ fn legacy_quantity_ref(x: &QuantityRef) -> bool {
         | QuantityRef::LifeTotal { .. }
         | QuantityRef::LifeAboveStarting
         | QuantityRef::StartingLifeTotal
+        | QuantityRef::TriggeringDiscoverValue
         | QuantityRef::GraveyardSize { .. }
         | QuantityRef::ObjectCount { .. }
         | QuantityRef::ObjectCountDistinct { .. }
@@ -5526,6 +5527,10 @@ fn rw_quantity_ref(x: &QuantityRef) -> RwProfile {
             reads_player_of(StateKind::PlayerLife)
         }
         QuantityRef::StartingLifeTotal => RwProfile::empty(),
+        // CR 701.57a: reads the transient last-discover scalar, written only by
+        // discover resolution (never by a sibling trigger) — no ordering-relevant
+        // read/write, mirroring StartingLifeTotal.
+        QuantityRef::TriggeringDiscoverValue => RwProfile::empty(),
         QuantityRef::GraveyardSize { .. } => reads_zone_membership(),
         QuantityRef::ObjectCount { filter }
         | QuantityRef::ObjectCountDistinct { filter, .. }
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -1581,6 +1581,9 @@ fn scan_quantity_ref(x: &QuantityRef) -> Axes {
             projected: true,
         },
         QuantityRef::StartingLifeTotal => Axes::NONE,
+        // CR 701.57a: reads a transient game-state scalar (the last discover's
+        // mana-value limit); no growing resource, sibling, or projected axis.
+        QuantityRef::TriggeringDiscoverValue => Axes::NONE,
         QuantityRef::ObjectCount { filter } => {
             let mut acc = Axes {
                 event: false,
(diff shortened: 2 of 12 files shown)


matthewevans (approved):
Approving. The architecture pass recorded on this head (`1d2aa09`) stands: new `QuantityRef::TriggeringDiscoverValue` + transient `GameState::last_discover_value` is the defensible seam given `EffectResolved{Discover}` carries no value, the variant is justified (no sibling cluster to parameterize), it's wired exhaustively, and the once-per-turn loop guard is both parsed and asserted (CR 701.57a). All 13 checks are green and the branch is clean.

One non-blocking watch item for a follow-up (not gating this PR): the discover value is read at re-discover *resolution* time rather than snapshotted at *trigger* time, so two overlapping discovers in the same on-stack window could feed the wrong value. It's narrow and fails safe (0 outside a discover context); the proper fix is a value-carrying `EffectResolved{Discover}` event, which is a separate change. Thanks for the clean fix and the loop-guard test.

PR #5518, opened 2026-07-10
Fix Veteran Bodyguard / Weathered Bodyguards — dropped tap-gate + unblocked/combat source restriction on damage redirection

rykerwilliams:
## Summary
Fixes the "Replacement / prevention / 'instead' effect mis-modeled" misparse for **Veteran Bodyguard** and **Weathered Bodyguards**.

Both cards use the shape "As long as this creature is untapped, all [combat] damage that would be dealt to you [by unblocked creatures] is dealt to this creature instead." The parser was dropping three things: the leading "as long as untapped" condition gate (CR 604.2), the "by unblocked creatures" source-filter restriction (CR 509.1h), and the "combat damage" vs. plain "damage" scope distinction. Net effect before this fix: both cards redirected **all** damage to their controller, unconditionally, regardless of tapped state or source — strictly wrong per their printed text.

`parse_damage_redirection_replacement` (`crates/engine/src/parser/oracle_replacement.rs`) now parses all three, reusing pre-existing types verbatim — no new enum variants or struct fields anywhere, including `types/ability.rs`:
- `ReplacementCondition::SourceTappedState { tapped: bool }` for the tap-gate
- `FilterProp::Unblocked` for the "unblocked creatures" source restriction (via the pre-existing `parse_damage_source_subject_filter` → `parse_type_phrase` → `parse_combat_status_prefix` chain)
- `CombatDamageScope::CombatOnly` for Weathered Bodyguards' "combat damage" scope (via the pre-existing `scan_combat_scope`)

Palisade Giant is deliberately excluded from the source-filter class — its real Oracle text has no "unblocked" restriction — and has its own dedicated regression-guard test proving the exclusion.

Also renamed `strip_as_long_as_prefix_for_prevention` → `strip_as_long_as_condition_prefix`, since it is now shared between the prevention and redirection parsers (it was previously only used by the former).

### Discovered pre-existing bug (out of scope)

While tracing the runtime path, I found that `game/replacement.rs`'s `damage_done_applier` only reads the shield's `redirect_target` field inside its `ShieldKind::Redirection` branch (~line 1190-1193, CR 614.9). This entire card class — Pariah, Palisade Giant, Veteran Bodyguard, Weathered Bodyguards — parses to `ShieldKind::Prevention` with `redirect_target: SelfRef` as a *separate* field. The `ShieldKind::Prevention` branch (~line 1336-1349) never reads `redirect_target` and just returns `ApplyResult::Prevented` (~1382-1383). Net effect: this whole "damage is dealt to X instead" class only ever **prevents** damage to the controller — the "instead" creature never actually takes the redirected damage, can never die from it, and never triggers "dealt damage" abilities off it.

This is real, but it's in shared runtime resolver code affecting multiple already-shipped cards and needs its own dedicated plan+review cycle — it is **not** fixed by this PR, which is scoped to the parser only (tap-gate + source-filter + combat-scope, all upstream of `ShieldKind` dispatch and confirmed live/consumed at the replacement-candidate-eligibility gate regardless of this gap). The new tests document this in code comments and intentionally assert the life-total delta rather than `damage_marked` on the Bodyguard as their discriminator, so they don't silently mask the gap.

## Files changed
- `crates/engine/tests/integration/veteran_bodyguard_tap_redirect.rs` (new)
- `docs/parser-misparse-backlog.md`

## Gate A
```
(exit 0, no output — no violations)
```

## Anchored on
- `crates/engine/src/parser/oracle_replacement.rs:8360` — existing `parse_damage_prevention_replacement`, the direct precedent this PR mirrors for adding a leading "as long as" tap-gate to a damage replacement parser.
- `crates/engine/src/parser/oracle_replacement.rs:5937` — existing `parse_damage_source_subject_filter`, reused as-is (not reimplemented) to resolve the "unblocked creatures" source clause via its existing fallthrough to `parse_type_phrase`'s combat-status recognition.
- `crates/engine/src/parser/oracle_replacement.rs:6268` — existing `scan_combat_scope`, reused as-is to resolve the "combat damage" vs. plain "damage" scope distinction for Weathered Bodyguards.

## CR references
- CR 604.2 — static ability's own "as long as" continuous-effect activation gate
- CR 614.9 — damage redirection effects
- CR 509.1h — blocked/unblocked creature definition
- CR 614.1a — "would deal damage... instead" replacement shape
- CR 615.1a — prevent + redirect combination
- CR 510.1c — blocked creature assigns damage to blockers (new trample discriminator test)
- CR 702.19b — trample excess damage assignment (new trample discriminator test)

## Track
Developer

## LLM
Model: claude-sonnet-5
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `cargo clippy -p engine --all-targets -- -D warnings` — clean, zero warnings (confirmed post-rebase)
- `./scripts/check-parser-combinators.sh` — clean, exit 0
- `cargo test -p engine --lib` — 16041 passed, 0 failed, 6 ignored (confirmed post-rebase)
- `cargo test -p engine --test integration -- veteran_bodyguard` — 3 passed, 0 failed (confirmed post-rebase): tap-gate positive/negative pair plus a new discriminating test proving a blocked+trampling attacker's excess damage to the defending player is correctly *not* caught by the shield (the "unblocked creatures" source filter axis)
- Independent `/review-impl` — 2 rounds. Round 1 found zero blocking findings but flagged one non-blocking test-coverage gap (the source-filter and combat-scope axes were only proven at the parser-AST level, not through a full combat scenario). Addressed by adding the trample discriminator test. Round 2 (fresh reviewer, fresh context) confirmed clean, independently re-verified all CR citations and traced the new test's production path end-to-end.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -8058,21 +8058,62 @@ fn parse_cant_become_untapped_replacement(
     )
 }
 
-/// CR 614.1a: Parse damage redirection replacement effects.
+/// CR 614.9 + CR 509.1h: Extract the optional "by <source>" scope-restriction
+/// clause from a damage-redirection body — "...would be dealt to you by
+/// unblocked creatures is dealt to ~ instead." (Veteran Bodyguard, Weathered
+/// Bodyguards). Delegates entirely to `parse_damage_source_subject_filter`
+/// (the same subject-typing helper every other damage-source clause in this
+/// module already uses), which itself falls back to `parse_type_phrase` — the
+/// SAME combinator that already resolves "unblocked creatures" / "unblocked
+/// attacking creatures" to `FilterProp::Unblocked` via
+/// `parse_combat_status_prefix` (`oracle_target.rs`), proven by the existing
+/// `parse_type_phrase_unblocked_attacking_creatures_you_control` test. This
+/// function adds NO new unblocked-detection — only the "by ... is dealt to"
+/// boundary extraction that the redirection grammar does not yet have.
+/// Returns `None` when no "by " clause is present, so the unrestricted-source
+/// class (Pariah / Palisade Giant) is unaffected.
+fn parse_damage_redirection_source_clause(working_lower: &str) -> Option<TargetFilter> {
+    let (_, (_, after_by)) = nom_primitives::split_once_on(working_lower, " by ").ok()?;
+    let (_, (subject, _)) = nom_primitives::split_once_on(after_by, " is dealt to").ok()?;
+    parse_damage_source_subject_filter(subject.trim())
+}
+
+/// CR 614.1a + CR 604.2: Parse damage redirection replacement effects.
 /// Handles "all damage that would be dealt to [target] is dealt to ~ instead" (Pariah, Palisade Giant)
 /// and "if a source would deal damage to you, prevent that damage. ~ deals that much damage to
 /// any target" (Pariah's Shield).
+///
+/// CR 604.2: an optional leading "as long as <tap-state>, " gate (a printed
+/// static ability's own continuous-effect activation condition) is stripped and
+/// lifted to a typed `ReplacementCondition` before the body is parsed, so the
+/// redirect only applies while the condition holds (Veteran Bodyguard's
+/// "as long as this creature is untapped" — the redirect must NOT fire while the
+/// permanent is tapped).
+///
+/// CR 614.9 + CR 509.1h: an optional "by <source>" clause ("...by unblocked
+/// creatures...") scopes the redirect to a damage-source filter, and a
+/// "combat damage" qualifier scopes it to combat damage only — so the redirect
+/// applies only to combat damage from unblocked creatures rather than to all
+/// damage from every source (Veteran Bodyguard is source-restricted only;
+/// Weathered Bodyguards is both source- and combat-restricted).
 fn parse_damage_redirection_replacement(
     norm_lower: &str,
     original_text: &str,
 ) -> Option<ReplacementDefinition> {
+    // CR 604.2: lift a leading "as long as <tap-state>, " gate to a typed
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Approving — this is high-quality work.

**Verified:**
- **Card text** Scryfall-verified verbatim for both Veteran Bodyguard and Weathered Bodyguards (and Palisade Giant's *current* Oracle, correctly noting its recipient-list half is a separate pre-existing bug).
- **All six CR annotations** grep-verified against the Comprehensive Rules: 604.2 (static continuous/replacement effects), 614.9 (damage-redirection replacement), 509.1h (attacker-with-blockers becomes blocked), 510.1c + 702.19b (trample assignment), 509.1g.
- **Building-block reuse** — the source clause delegates to the existing `parse_damage_source_subject_filter` → `parse_type_phrase` path (no new unblocked-detection), and the `"as long as"` gate lifter is shared cleanly via the `strip_as_long_as_condition_prefix` rename rather than duplicated.
- **Both new-set fields are runtime-live**: `damage_source_filter` (replacement.rs:4816/5339) and `combat_scope` (replacement.rs:4829-4834/5359-5363, `CombatOnly if !is_combat`) are consumed by the resolver — not inert. The blocked-trample integration test is a genuine end-to-end discriminator proving the `Unblocked` source filter gates at runtime.

**On the disclosed gap:** your honesty about `redirect_target: SelfRef` being dead data for the whole prevent-only class (`ShieldKind::Prevention` never redirects — Pariah/Palisade Giant included) is exactly right and appreciated. That's a pre-existing, class-wide runtime-resolver gap; this parser-scoped PR doesn't touch or worsen it, and the in-scope restrictions (source/combat/tap-gate) are all consumed and tested. Correct call to defer it as a separate follow-up.

Non-vacuous negative tests with a positive reach-guard, real Oracle text, delegated building blocks — thank you.

PR #5520, opened 2026-07-10
feat(parser): source-anchored combat-relationship subject static (Alms Beast)

nickmopen:
## Problem

A self-relative combat-relationship subject failed to parse — **Alms Beast**: `Creatures blocking or blocked by ~ have lifelink.` No subject-filter arm recognized the `blocking or blocked by <self>` qualifier, so the whole static line dropped (`static_structure` Unimplemented).

The **target-relative** form (`… blocking or blocked by target creature`) is already parsed in `oracle_target`; only the **self/source-relative static** form was missing.

## Fix

Strip the trailing `blocking or blocked by ~` / `blocking or blocked by this creature` qualifier in `parse_continuous_subject_filter`, parse the base subject recursively, and attach a source-anchored `FilterProp::CombatRelation { relation: BlockingOrBlockedBy, subject: Source }`.

That `FilterProp` (and `CombatRelationSubject::Source`) is **already fully evaluated at runtime** by `game::filter` — so this is a **grammar-only** seam. The self-reference is normalized to `~`; `this creature` is accepted for the literal phrasing. Mirrors the existing merged qualifier-strip pattern in the same function.

## Tests

`combat_relation_subject_static_binds_source_anchored_filter` asserts (typed AST, both `~` and `this creature` forms) that the line produces a single static whose affected filter is a creature subject carrying `FilterProp::CombatRelation { BlockingOrBlockedBy, Source }`, with an `AddKeyword { Lifelink }` grant.

Local: `parser::oracle_static` passes; `cargo fmt`, clippy (`-D warnings`), parser-combinator + engine-authorities gates all clean.

### DIFF
--- a/crates/engine/src/parser/oracle_static/mod.rs
+++ b/crates/engine/src/parser/oracle_static/mod.rs
@@ -44,11 +44,11 @@ mod prelude {
     };
     pub(super) use crate::types::ability::{
         AbilityCost, AbilityDefinition, AbilityKind, AbilityTag, ActivationRestriction,
-        AttachmentKind, BasicLandType, CardPlayMode, ChosenSubtypeKind, Comparator,
-        ContinuousModification, ControllerRef, CostCategory, CountScope, FilterProp, ObjectScope,
-        ParsedCondition, PlayerFilter, PtStat, PtValueScope, QuantityExpr, QuantityRef,
-        SharedQuality, SharedQualityRelation, StaticCondition, StaticDefinition, TargetFilter,
-        TypeFilter, TypedFilter,
+        AttachmentKind, BasicLandType, CardPlayMode, ChosenSubtypeKind, CombatRelation,
+        CombatRelationSubject, Comparator, ContinuousModification, ControllerRef, CostCategory,
+        CountScope, FilterProp, ObjectScope, ParsedCondition, PlayerFilter, PtStat, PtValueScope,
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Approving — clean, well-scoped grammar work.

**Verified:**
- **Right seam.** This adds the `subject: CombatRelationSubject::Source` (self-anchored) sibling of the already-parsed target-relative form in `oracle_target.rs` (which uses `subject: ParentTarget`) — same `FilterProp::CombatRelation`, different anchor. It's a grammar-only change.
- **Runtime authority pre-exists.** `game::filter` already evaluates `CombatRelationSubject::Source` (filter.rs:3302) and `CombatRelation::BlockingOrBlockedBy` (filter.rs:3318), so the parsed filter is live at runtime — no new engine logic, no inert AST.
- **Combinators + suffix discipline.** `all_consuming(terminated(take_until(needle), tag(needle)))` correctly requires the qualifier to be a true trailing suffix, and the recursion into `parse_continuous_subject_filter(base)` composes the base subject with the added prop.
- **Discriminating test** asserts the typed structure (creature subject + source-anchored `BlockingOrBlockedBy` prop + lifelink grant) for both the normalized `~` and literal "this creature" phrasings — not a Debug-string dump.

Thanks — nicely mirrors the existing pattern for the static-subject case.

PR #5524, opened 2026-07-10
feat(engine): support "conjure a random card from [X] spellbook" (random draft)

ismayilovibadet802-svg:
## Summary

The parser recognized "conjure a card of your choice from [X]'s spellbook" (and "draft a card from [X]'s spellbook") → `Effect::DraftFromSpellbook`, but not the engine-picks-at-random wording "conjure a random card from [X]'s spellbook". Both are the same digital-only Alchemy operation; the only difference is who picks — the controller (choice) or the engine (random).

This extends `Effect::DraftFromSpellbook` with a `#[serde(default)] random: bool`:

- **Resolver**: when `random`, pick one card uniformly at random (`state.rng.random_range`, the deterministic `ChaCha20Rng`) and conjure it immediately via the existing `complete_draft`, instead of pausing on `WaitingFor::SpellbookDraft` for a choice. Random-vs-choice is the only difference, so both share the resolver and the same card-creation path.
- **Parser**: the conjure-from-spellbook parser is generalized to accept both wordings — "of your choice" → `random: false`, "a random card" → `random: true` — reusing `parse_conjure_zone` and the fully-consumed-tail discipline.

## Coverage

Closes the `Effect:conjure` single-gap for **10 cards** (supported 31,485 → 31,495), e.g. Tome of the Infinite. Cards whose destination `parse_conjure_zone` does not yet model ("into exile", "into the top five cards of your library at random") correctly fall through to `Unimplemented` rather than parsing to a subtly wrong effect — a separate destination gap.

## Verification

- full `cargo test -p engine` passes (16,043 tests, 0 failures)
- New tests: two parser tests (random onto-the-battlefield, random into-your-hand) and a resolver test (`random_spellbook_draft_conjures_without_pausing`) proving a random draft conjures a card into hand and returns to `Priority` without pausing for a pick
- `cargo coverage` after regenerating card data: the 10 cards flip to `supported`

Model: Claude Opus 4.8

### DIFF
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -4094,10 +4094,7 @@ fn rw_effect(
             (p, None)
         }
         Effect::Learn => (ext_write(StateKind::HandLibrary), None),
-        Effect::DraftFromSpellbook {
-            destination: _,
-            tapped: _,
-        } => (ext_write(StateKind::HandLibrary), None),
+        Effect::DraftFromSpellbook { .. } => (ext_write(StateKind::HandLibrary), None),
         Effect::Clash => (ext_write(StateKind::HandLibrary), None),
         Effect::ChooseDrawnThisTurnPayOrTopdeck {
             count,
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -1531,10 +1531,7 @@ fn scan_effect(x: &Effect) -> Axes {
             acc = acc.or(scan_quantity_expr(amount));
             acc
         }
-        Effect::DraftFromSpellbook {
-            destination: _,
-            tapped: _,
-        } => Axes::NONE,
+        Effect::DraftFromSpellbook { .. } => Axes::NONE,
         Effect::ChooseCounterAdjustment {
             adjustment: _,
             count,
--- a/crates/engine/src/game/effects/spellbook.rs
+++ b/crates/engine/src/game/effects/spellbook.rs
@@ -15,6 +15,8 @@
 //! source object (`GameObject::spellbook`, copied from
 //! `CardFace::metadata.spellbook`); the resolver reads it from the source.
 
+use rand::Rng;
+
 use crate::types::ability::{
     ConjureCard, ConjureSource, Effect, EffectError, EffectKind, QuantityExpr, ResolvedAbility,
 };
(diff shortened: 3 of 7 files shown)


matthewevans (commented):
Thanks for this — reviewed the architecture and it's sound; holding only for CI.

**Correctness (verified):**
- The empty-spellbook guard (`spellbook.is_empty() → no-op`) precedes the new `random` branch, so `random_range(0..spellbook.len())` is only reached with ≥1 entry — no panic on an empty list.
- The random pick uses the seeded engine RNG (`state.rng`, a `ChaCha20Rng` with `rng_seed`/`rng_word_pos`), **not** an ad-hoc `thread_rng()` — so it's replay-safe and multiplayer-deterministic. That's the right seam for a random effect.
- Card creation is delegated to the shared `complete_draft` path — no duplication of the conjure logic.
- The runtime test `random_spellbook_draft_conjures_without_pausing` is genuinely discriminating: a single-card spellbook makes the pick deterministic, and it asserts both that the card lands in hand **and** that resolution ends at `WaitingFor::Priority` (no pause). Reverting the resolver arm re-introduces the `SpellbookDraft` pause and flips both assertions.
- `#[serde(default)]` on the new field keeps existing serialized abilities deserializing cleanly.

**Non-blocking note (for later, not this PR):** `random: bool` joins the existing `tapped: bool` rider, so it's locally consistent. But it's really a *selection-mode* axis, and your own comment anticipates the real three-card Alchemy draft. If/when that lands, `random` will want to coalesce into a small `SpellbookSelection { Choose, Random, DraftOf(n) }`-style enum rather than a second bool. No change needed now — just flagging where the class grows.

I've updated the branch against `main` (it was behind — the same-lineage spellbook cards have been sensitive to coverage-baseline drift). Will approve + enqueue once CI is green.

matthewevans (approved):
Approved. CI is green on the update-branched head `537f8df9` (12 checks success, Draft-pools smoke skipped).

Architecture verified last sweep and re-confirmed:
- `random: bool` rider lets "of your choice" and "a random card" share one resolver and the shared `complete_draft` path — no duplicated draft logic.
- The empty-spellbook guard precedes `random_range(0..len())`, so no panic on an empty book.
- The pick uses the seeded `state.rng` (`ChaCha20Rng`), so it is replay-safe and deterministic — the correct seam for anything random in the engine (an ad-hoc `thread_rng` would desync replay/multiplayer).
- The runtime test is discriminating: single-card spellbook asserting `WaitingFor::Priority` (no choice pause); reverting the arm flips it.

**Non-blocking follow-up:** when the anticipated 3-card draft variant lands, `random` becomes one axis of a `SpellbookSelection` selection-mode enum — coalesce then. As a two-state rider matching the existing `tapped: bool` convention, it's fine as-is now.

PR #5533, opened 2026-07-10
feat(parser): parse "into exile" as a conjure destination

ismayilovibadet802-svg:
## Summary

`parse_conjure_zone` — the shared destination parser for the conjure family — modeled `onto the battlefield` and `into your/their hand/graveyard/library`, but not `into exile`. So conjures that land a card in exile (named, of-your-choice, and random draft-from-spellbook) fell through to `Unimplemented`.

This adds `Zone::Exile` (`" into exile"`) to the destination alternatives, which unblocks every conjure wording that targets exile at once.

## Coverage

Closes the `Effect:conjure` single-gap for **7 cards** (supported 31,495 → 31,502).

## Verification

- full `cargo test -p engine` passes; two new parser tests: `conjure a card named X into exile`, and the random draft-from-spellbook wording composing with the exile destination
- `cargo coverage` after regenerating card data: +7 cards flip to `supported`

Model: Claude Opus 4.8

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -9027,6 +9027,9 @@ fn parse_conjure_zone(lower: &str) -> Option<(Zone, &str)> {
         value(Zone::Hand, tag(" into their hand")),
         value(Zone::Graveyard, tag(" into their graveyard")),
         value(Zone::Library, tag(" into their library")),
+        // Digital-only Alchemy conjures can also land in exile (e.g. a random card
+        // conjured from a spellbook into exile with a play-permission rider).
+        value(Zone::Exile, tag(" into exile")),
     ))
     .parse(lower)
     .ok()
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Clean, well-scoped parser win. The change is a single `value(Zone::Exile, tag(" into exile"))` arm added to the existing `parse_conjure_zone` `alt()` — the right seam: it extends the shared conjure-destination family (battlefield / hand / graveyard / library) rather than adding a bespoke parser, so it unblocks every `into exile` conjure wording at once (24 additions, 0 removals). Verified the wording is real against Scryfall — Smog Smasher ("conjure a duplicate of target nontoken creature into exile") and Dazzling Flameweaver ("conjure a random card from …'s spellbook into exile") — and both tests are discriminating: `conjure_named_into_exile` asserts `destination == Zone::Exile`, and `conjure_random_from_spellbook_into_exile` asserts the full `DraftFromSpellbook { destination: Zone::Exile, tapped: false, random: true }` composition. Full CI green; real coverage gain (+7 cards). Nice use of nom `value`/`tag` combinators. Approving and enqueuing.

PR #5535, opened 2026-07-10
Fix Lighthouse Chronologist level trigger condition

### LINKED ISSUE #4728: Lighthouse chronologist infinite turns — When Lighthouse chronologist hits level 7 you are supposed to to take an extra…
Reported in Discord: https://discord.com/channels/1485498006781427802/1520339647761481789/1520339647761481789

**Thread:** Lighthouse chronologist infinite turns
**Cards:** lighthouse chronologist
**Parser status:** has_gaps

## Summary
When Lighthouse chronologist hits level 7 you are supposed to to take an extra turn after each end step except your own, but when i hit 7 i kept taking extra turns after my endstep.

### lighthouse chronologist
Level up {U} ({U}: Put a level counter on this. Level up only as a sorcery.)
LEVEL 4-6
2/4
LEVEL 7+
3/5
At the beginning of each end step, if it's not your turn, take an extra turn after this one.

---
<sub>report_id: `discord:1520339647761481789:1520339647761481789:0` · discord: `1520339647761481789/1520339647761481789`</sub>

mike-theDude:
## Summary
Fixes the level-trigger condition composition for **Lighthouse Chronologist**.

Closes #4728.

The LEVEL 7+ trigger now keeps both required gates: the creature must have at least seven level counters, and the printed intervening-if condition (`if it's not your turn`) must also be true. Previously the level-block reparse replaced the printed condition with the level-counter gate, allowing the extra-turn trigger to fire on the controller's own end step.

## Files changed
- `crates/engine/src/parser/oracle.rs`
- `crates/engine/src/parser/oracle_trigger_tests.rs`

## CR references
- `CR 711.2a`
- `CR 711.2b`
- `CR 603.4`

## Track
Developer

## LLM
Model: codex-5
Thinking: high
Tier: Standard

## Gate A
`./scripts/check-parser-combinators.sh` exited 0 with no output.

## Anchored on
- `crates/engine/src/parser/oracle_trigger.rs:6964` - existing trigger-condition composition helper wraps an existing condition and a new gate in `TriggerCondition::And`.
- `crates/engine/src/parser/oracle_trigger_tests.rs:16796` - existing regression test for composing a trigger gate with a printed intervening-if condition.

## Verification
- `cargo fmt --all -- --check` - clean
- `./scripts/check-parser-combinators.sh` - clean
- `git diff --check origin/main...HEAD` - clean
- `cargo test -p engine level_trigger_preserves_printed_intervening_if_condition` - 1 passed / 0 failed
- `cargo clippy -p engine --all-targets -- -D warnings` - clean

Tilt was not running, so direct cargo fallback was used.

## Scope Expansion
None.

## Validation Failures
Independent fresh-context subagent review was not run because subagent delegation was not explicitly requested; local `doc-guardian` and `code-reviewer` checklist passes found no findings.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -2865,7 +2865,16 @@ pub(crate) fn parse_oracle_ir(
             &mut ctx,
         );
         for trigger in &mut triggers {
-            trigger.condition = Some(trigger_condition.clone());
+            trigger.condition = Some(match trigger.condition.take() {
+                // CR 711.2a + CR 711.2b + CR 603.4: a level-gated trigger's
+                // level-counter range is an additional gate; it must compose
+                // with any printed intervening-if condition instead of
+                // replacing it.
+                Some(existing) => TriggerCondition::And {
+                    conditions: vec![existing, trigger_condition.clone()],
+                },
+                None => trigger_condition.clone(),
+            });
         }
         result.triggers.extend(triggers);
     }
--- a/crates/engine/src/parser/oracle_trigger_tests.rs
+++ b/crates/engine/src/parser/oracle_trigger_tests.rs
@@ -16817,6 +16817,69 @@ fn trigger_while_attacking_composes_with_existing_condition() {
     }
 }
 
+/// CR 711.2a + CR 711.2b + CR 603.4 (issue #4728, Lighthouse Chronologist):
+/// level-block reparsing must AND-compose the level-counter gate with the
+/// printed intervening-if condition. Before the fix, the level-block pass
+/// overwrote `if it's not your turn`, so the extra-turn trigger fired at the
+/// controller's own end step once LEVEL 7+ was active.
+#[test]
+fn level_trigger_preserves_printed_intervening_if_condition() {
+    use crate::types::phase::Phase;
+
+    const LIGHTHOUSE_CHRONOLOGIST: &str =
+        "Level up {U} ({U}: Put a level counter on this. Level up only as a sorcery.)\n\
+LEVEL 4-6\n\
+2/4\n\
+LEVEL 7+\n\
+3/5\n\
+At the beginning of each end step, if it's not your turn, take an extra turn after this one.";
+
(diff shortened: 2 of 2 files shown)


matthewevans (commented):
Reviewed and verified — holding only on CI (5 checks still running); will approve on settled green.

**Correct, and on the general seam.** The bug is in the level-block reparsing pass, which was unconditionally *overwriting* each trigger's condition with the level-counter gate — so for `LEVEL 7+`, the printed intervening-if `if it's not your turn` was dropped and the extra-turn trigger would fire at the controller's *own* end step. AND-composing the level gate with any existing condition is the right fix, and because it lives on the shared reparsing path it covers every leveler with a printed intervening-if, not just Lighthouse Chronologist.

**Verified:**
- Oracle text matches live Scryfall exactly (`At the beginning of each end step, if it's not your turn, take an extra turn after this one.`).
- CR citations grep-verified against the Comprehensive Rules: **711.2a/711.2b** (level symbols are static abilities active while in the counter range) and **603.4** (intervening-if). All accurate.
- Test is discriminating: it destructures `TriggerCondition::And` and asserts *both* the `HasCounters{level, minimum:7}` gate and the `Not{DuringPlayersTurn{Controller}}` gate are present — before the fix the `And` wouldn't exist at all, so the test fails closed.

Nice fix. Enqueue is pending only on the rollup settling green.

matthewevans (approved):
Approving. The fix AND-composes the level-counter gate with the printed intervening-if on the shared level-block reparse path (`TriggerCondition::And { conditions: vec![existing, trigger_condition] }`), so it covers every leveler with a printed condition, not just Lighthouse Chronologist. Pre-fix, the reparse unconditionally overwrote the condition, dropping `if it's not your turn` and firing the extra-turn trigger at the controller's own end step. CR 711.2a/711.2b/603.4 verified; the test destructures the `And` and asserts both gates (fails closed pre-fix). CI green, merge CLEAN.

PR #5539, opened 2026-07-11
Fix explore-replacement double-run and interactive sequencing (#5272)

### LINKED ISSUE #5272: Explore: Topography Tracker Loop — Gamestate Attached.
Reported in Discord: https://discord.com/channels/1485498006781427802/1523271921066704946/1523271921066704946

**Thread:** Explore: Topography Tracker Loop
**Discord thread id:** `1523271921066704946`
**Discord message id:** `1523271921066704946`
**Cards:** topography tracker
**Parser status:** has_gaps

## Summary
Gamestate Attached.

## Reported text

Gamestate Attached.

Heyo! Looks like I've encountered a similar bug to the one described here!

https://github.com/phase-rs/phase/issues/2017

Topography Tracker gives creatures that are exploring two instances of explore. 
A couple of things seem to be going wrong with this. 
Firstly, I think it gives everything, including itself, explore. Lands enter and explore, artifacts enter and explore, so on and so forth.
Secondly, while the first instance of explore behaves normally, the **second** doesn't seem stop until you choose to keep a non-land card on top of your graveyard. 

I'll drop in the gamestate below - hope it helps out! If you have further questions I'd love to help out how I can!
Cheers!

## Attachments

- [game-state-turn-18-2026-07-05T09-34-01-640Z.zip](https://cdn.discordapp.com/attachments/1523271921066704946/1523271921297264640/game-state-turn-18-2026-07-05T09-34-01-640Z.zip?ex=6a4d7b47&is=6a4c29c7&hm=9ef6be18f4bf7bfcda2c9ea64260261cf79a74b848c1600025c58f9dcf9e06d3&)

### topography tracker
When this creature enters, create a Map token. (It's an artifact with "{1}, {T}, Sacrifice this token: Target creature you control explores. Activate only as a sorcery.")
If a creature you control would explore, instead it explores, then it explores again.

---
<sub>report_id: `discord:1523271921066704946:1523271921066704946:0` · discord: `1523271921066704946/1523271921066704946`</sub>

mike-theDude:
## Summary
Fixes the explore-replacement double-run and interactive-sequencing bug behind Topography Tracker ("if a creature you control would explore, instead it explores, then it explores again") and Twists and Turns ("scry 1, then that creature explores"). The same top card was revealed twice, counters were mis-applied, and the replacement's execute chain ran a second time on the replacement's own source — surfacing as a stuck/"looping" board.

Closes #5272.

Root causes:
1. `explore_applier` ran the replacement's execute chain in a straight-line loop, ignoring the interactive DigChoice/ScryChoice pause. It now runs the chain through the standard interactive continuation machinery (`resolve_ability_chain` + `build_resolved_from_def`), carrying the already-applied replacement set (CR 614.5 — no self-invocation; CR 616.1f — other doublers still apply).
2. The execute chain was also stashed as a generic post-replacement continuation and run twice. Suppressed the generic stash for `ProposedEvent::Explore`, mirroring the existing Connive suppression.
3. Ordering bug this exposed: for a multi-creature explore under a doubler (Hakbal of the Surging Soul), the next creature's explore was prepended ahead of the doubled second explore. A single creature's two explores now stay consecutive (guarded append, mirroring the PayCost/clash per-effect precedent).

Also corrected two stale CR annotations (701.37a Monstrosity → 701.44a Explore).

## Files changed
- crates/engine/src/game/effects/explore.rs
- crates/engine/tests/integration/explore_all_doubler_ordering.rs (new)
- crates/engine/tests/integration/main.rs

## CR references
- CR 614.1a — replacement event self-application
- CR 614.5 — a replacement effect does not apply to the event it generates (no self-invocation)
- CR 616.1 / CR 616.1f — multiple applicable replacement effects; other doublers still apply
- CR 701.44a — Explore (corrected from stale 701.37a Monstrosity annotations)

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification
- `cargo fmt --all` — clean
- `./scripts/check-parser-combinators.sh` — clean (exit 0, no violations)
- `cargo clippy-strict` — clean
- `cargo test -p engine` (lib) — 16057 pass / 0 fail
- Targeted suites: explore unit (12), connive/scry/replacement (215), explore integration (9, incl. new end-to-end ExploreAll+doubler ordering test)
- Adversarial `$review-impl` cross-check: APPROVE — a RULES bug (multi-creature ordering under a doubler) was found and fixed; all suggestions applied.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

## Gate A
```
(no output — exit 0)
```
This change touches only the engine effect/replacement resolution layer and its integration tests; no parser combinator code is modified, and the nom-mandate gate reports zero violations.

## Anchored on
- crates/engine/src/game/replacement.rs:2019 — existing Connive post-replacement continuation suppression; the Explore suppression mirrors this same re-entry-vs-stash pattern.
- crates/engine/src/game/effects/mod.rs:5826 — existing `resolve_ability_chain` interactive-continuation machinery, reused to run the explore replacement chain through the DigChoice/ScryChoice pause instead of a straight-line loop.
- crates/engine/src/game/effects/mod.rs:2224 — existing per-effect choice-outcome guard (FlipCoin/RollDie/Clash/Dig) that the guarded-append ordering fix follows.

Tier: Frontier

### DIFF
--- a/crates/engine/src/game/effects/explore.rs
+++ b/crates/engine/src/game/effects/explore.rs
@@ -228,11 +228,15 @@ pub fn resolve(
         })
         .unwrap_or(ability.source_id);
 
-    // CR 701.37a + CR 614.1a: Consult explore replacements (Twists and Turns,
-    // Topography Tracker, …) before the reveal/counter/land logic runs.
+    // CR 701.44a + CR 614.1a + CR 614.5: Consult explore replacements (Twists and
+    // Turns, Topography Tracker, …) before the reveal/counter/land logic runs.
+    // Seed the proposed event's applied set from the resolving ability so an
+    // explore produced by a doubling replacement's execute chain excludes the
+    // replacement that produced it (no self-invocation) while remaining open to
+    // other doublers (CR 616.1f). Empty for a first-time explore.
     let proposed = ProposedEvent::Explore {
         object_id: explorer_id,
-        applied: HashSet::new(),
+        applied: ability.replacement_applied.clone(),
     };
     match replacement::replace_event(state, proposed, events) {
         ReplacementResult::Execute(_) => {}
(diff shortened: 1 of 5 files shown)


matthewevans (approved):
Approving — reviewed the full diff against the Comprehensive Rules and both cards' live Oracle text.

**Root cause & fix (right seam).** The old `explore_applier` ran the replacement's execute chain in a straight-line loop, so an interactive explore (nonland reveal → `DigChoice`) never paused: both explores fired against the same still-unmoved top card, granting two counters up front and clobbering the first choice. The fix routes the chain through the standard interactive continuation machinery (`build_resolved_from_def` + `stamp_explore_chain` + `resolve_ability_chain`), so a mid-chain scry or a `DigChoice` parks the choice and stashes the next link on `pending_continuation`, resuming only after the player responds. Reuses existing infrastructure rather than a bespoke loop.

**CR-correct (all grep-verified against `docs/MagicCompRules.txt` + Oracle-verified).**
- **CR 614.5** (self-invocation guard) — `stamp_explore_chain` carries the already-applied replacement set onto every link (incl. `sub_ability`/`else_ability`), and `ProposedEvent::Explore.applied` is now seeded from `ability.replacement_applied`, so a doubler can't re-double its own output. Verbatim: *"A replacement effect doesn't invoke itself repeatedly."*
- **CR 616.1f** — the applied set excludes only *this* replacement id, so other explore doublers remain applicable.
- **CR 616.1** — the `Effect::Explore`-guarded append in `resolve_chain_body` keeps one creature's two explores consecutive (the next creature can't interleave between the halves).
- **CR 701.44a** — correctly migrated from the stale `701.37a` annotation.
- Oracle text confirmed live: Topography Tracker (*"…instead it explores, then it explores again"*) and Twists and Turns (*"…instead you scry 1, then that creature explores"*).

**Discriminating tests (fail closed on the pre-fix behavior).** Three tests: distinct-reveal sequencing (card C staying on top proves no self-re-double), one-creature-two-explores-consecutive ordering, and a real `ExploreAll` + Topography Tracker production shape (Hakbal) with a `step < 63` bound that turns the historic self-renewing loop into a test failure. The new integration file is registered in `tests/integration/main.rs` (not inert).

Also mirrors the established connive-exemption pattern to suppress the generic post-effect stash for `Explore` (the applier is the single authority for the event and returns `Prevented`, so stashing again would double-run on the replacement's own source). Clean, idiomatic work.

PR #5542, opened 2026-07-11
feat(parser): grant soulbond-paired quoted activated/triggered abilities

minion1227:
CR 702.95 + CR 613.1f

## Bug

`parse_soulbond_paired_static` extracted the granted predicate with `take_until(".")`, which stops at the **first** period. A soulbond card whose paired grant is a **quoted activated/triggered ability** carries that ability's own sentence-ending period **inside** its closing quote (`... under your control."`). `take_until(".")` therefore severed the closing quote, the leftover `"` failed `all_consuming`, the soulbond parse returned `None`, and a fallback produced a static with **empty `modifications`** and the **wrong `SelfRef` scope**. None of these cards granted their ability at runtime.

Fully-broken class (10 cards — the granted ability was entirely dropped):

- **Deadeye Navigator** — `... each of those creatures has "{1}{U}: Exile this creature, then return it to the battlefield under your control."`
- **Galvanic Alchemist** — `"{2}{U}: Untap this creature."`
- **Stonewright** — `"{R}: This creature gets +1/+0 until end of turn."`
- **Stern Mentor** — `"{T}: Target player mills two cards."`
- **Doom Weaver**, **Breathkeeper Seraph**, **Thundering Mightmare**, **Imperious Mindbreaker**, **Mirage Phalanx** — quoted triggered abilities
- **Tandem Lookout** — `"Whenever this creature deals damage to an opponent, draw a card."`

## Fix

Scoped to the existing paired seam in `parse_soulbond_paired_static`:

1. Split the pairing frame off the predicate with **`TextPair`** so the predicate keeps its **original case** — a quoted ability's mana symbols (`{1}{U}`) must reach the cost parser un-lowercased (the old path passed the lowercased slice, a latent second bug).
2. Reuse the existing **`matches_soulbond_paired_condition`** building block to validate the pairing clause.
3. Add one quote-aware helper **`strip_granted_predicate_period`** that trims the sentence-ending period only when it sits **outside** a quote.

The intact predicate then flows through the existing `parse_continuous_gets_has` → `parse_quoted_ability_modifications` seam, which already classifies quoted text into `GrantAbility` (CR 602.1) / `GrantTrigger` (CR 603.1). The grant is applied to **both** paired creatures (`SourceOrPaired`, CR 613.1f layer 6), gated on `SourceIsPaired`.

- No new sibling handler — the fix parameterizes the **one** paired-static path; the bare keyword/P-T soulbond grants (Wingcrafter, Wolfir Silverheart, …) route through the **same** seam and parse **identically** as before (regression-tested).
- Parse-diff limited to the 10 quoted-grant soulbond cards.
- Reuses `TextPair`, `matches_soulbond_paired_condition`, `parse_continuous_gets_has`, `parse_quoted_ability_modifications` — no new engine surface.

## Tests

Full-shape assertions on `affected` / `condition` / granted modification:

- `soulbond_grants_quoted_activated_mana_ability` — `GrantAbility` (Activated, cost preserved)
- `soulbond_grants_quoted_tap_free_activated_ability`
- `soulbond_grants_quoted_triggered_ability` — `GrantTrigger`
- `soulbond_paired_keyword_grant_still_parses` — regression: `AddKeyword(Flying)`
- `soulbond_paired_pt_grant_still_parses` — regression: `AddPower{4}` + `AddToughness{4}`

## CR verification

- **CR 702.95** — Soulbond (verified in `docs/MagicCompRules.txt`).
- **CR 613.1f** — Layer 6: ability-adding effects (the grant to both paired creatures).
- **CR 602.1 / CR 603.1** — the granted activated / triggered ability.

## AI-contributor notes

- **Rules-correct:** granted ability restored to both paired creatures with the correct `SourceOrPaired` scope; the previous `SelfRef` fallback was wrong on both scope and content.
- **Builds the class:** one seam covers all 10 quoted-grant soulbond cards plus any future soulbond quoted grant.
- **Verified:** live-parsed Deadeye Navigator, Galvanic Alchemist, Stonewright, Tandem Lookout before/after against Scryfall Oracle text; `cargo fmt`, CI-exact `clippy -p engine --all-targets --features proptest -D warnings`, and the full soulbond test set are green.

### DIFF
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -659,20 +659,43 @@ pub(crate) fn parse_soulbond_paired_static(
     tp: &TextPair<'_>,
     description: &str,
 ) -> Option<StaticDefinition> {
-    let parser = preceded(
-        tag("as long as "),
-        preceded(
-            terminated(parse_soulbond_paired_condition_nom, tag(", ")),
-            preceded(
-                alt((tag("each of those creatures "), tag("both creatures "))),
-                alt((terminated(take_until("."), tag(".")), rest)),
-            ),
-        ),
-    );
-    let (_, predicate) = all_consuming(parser).parse(tp.lower).ok()?;
-    let mut def = parse_continuous_gets_has(predicate, TargetFilter::SourceOrPaired, description)?;
+    // CR 702.95: Soulbond. The paired reminder-text grant — "As long as ~ is
+    // paired with another creature, each of those creatures <predicate>." — is a
+    // CR 613.1f layer-6 ability-adding effect applied to BOTH paired creatures
+    // (SourceOrPaired). Split the pairing frame off the granted predicate with
+    // TextPair so the predicate keeps its ORIGINAL case: a quoted granted
+    // ability's mana symbols (`{1}{U}`) must reach the cost parser un-lowercased.
+    let after = tp.strip_prefix("as long as ")?;
+    let (condition, predicate) = after
+        .split_around(", each of those creatures ")
+        .or_else(|| after.split_around(", both creatures "))?;
+    if !matches_soulbond_paired_condition(condition.lower) {
+        return None;
+    }
+    let predicate = strip_granted_predicate_period(&predicate);
+    let mut def = parse_continuous_gets_has(
+        predicate.original,
+        TargetFilter::SourceOrPaired,
+        description,
+    )?;
     def.condition = Some(StaticCondition::SourceIsPaired);
     Some(def)
+}
+
+/// Trim a granted predicate's sentence-ending period, but leave a quoted ability
+/// intact. A quoted activated (CR 602.1) or triggered (CR 603.1) granted ability
+/// terminates with its period INSIDE the closing quote (`has "{1}{U}: ... your
+/// control."`), so a predicate ending in `"` has no outside period to strip —
+/// only the bare keyword/P-T forms (`has flying.`, `gets +1/+1.`) carry an outer
+/// period. A period-terminated `take_until(".")` would instead sever the quote
+/// at that inner period and drop the whole granted ability.
+fn strip_granted_predicate_period<'a>(predicate: &TextPair<'a>) -> TextPair<'a> {
+    let predicate = predicate.trim_end();
+    if predicate.ends_with("\"") {
+        predicate
+    } else {
+        predicate.trim_end_matches('.')
+    }
 }
 
 pub(crate) fn bind_where_x_in_quantity_expr(
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -665,6 +665,7 @@ pub(crate) fn parse_soulbond_paired_static(
     // (SourceOrPaired). Split the pairing frame off the granted predicate with
     // TextPair so the predicate keeps its ORIGINAL case: a quoted granted
     // ability's mana symbols (`{1}{U}`) must reach the cost parser un-lowercased.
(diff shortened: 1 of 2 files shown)


matthewevans (commented):
**Correct fix at the right seam — holding only on the pending Rust engine gates.** The `take_until(".")` predicate extractor severed quoted granted abilities at the period *inside* their closing quote, so the soulbond parse returned `None` and a fallback produced an empty-`modifications` static — 10 cards silently dropped their granted ability.

## ✅ Right seam, real building blocks
`anthem.rs:659` — the fix parameterizes the one existing `parse_soulbond_paired_static` path rather than adding a sibling handler. It swaps the nom `take_until(".")` for `TextPair::split_around`, keeping the predicate's **original case** so a quoted ability's mana symbols (`{1}{U}`) reach the cost parser un-lowercased (a latent second bug in the old lowercased path). It reuses the existing `matches_soulbond_paired_condition` (`shared.rs:1685`) for the pairing clause and flows the intact predicate through the existing `parse_continuous_gets_has` → `parse_quoted_ability_modifications` seam — no new engine surface. `TextPair` is the correct tool here (dual-string, case-preserving), exactly the structural-strip use it exists for.

The new `strip_granted_predicate_period` helper trims the sentence period only when it sits **outside** a quote (`has flying.` → strip; `has "…control."` → leave), which is what makes the quoted class survive.

## ✅ CR annotations verified
CR 702.95 (Soulbond), CR 613.1f (layer 6 ability-adding, applied to both paired creatures via `SourceOrPaired`), CR 602.1 (activated), CR 603.1 (triggered) — all grep-verified accurate.

## ✅ Discriminating tests + Oracle text confirmed
The tests destructure `GrantAbility { definition }` and assert `kind == Activated` **and** `cost.is_some()` — they fail on the pre-fix empty-`modifications` behavior — and distinguish `GrantTrigger` (Tandem Lookout's "Whenever…") from `GrantAbility`, with regression guards for the bare keyword/P-T grants (Wingcrafter flying, Wolfir Silverheart +4/+4). Oracle text verified verbatim on Scryfall for Deadeye Navigator, Stonewright, and Tandem Lookout.

## 🟡 Non-blocking
No `coverage-parse-diff` sticky is present, so I confirmed reachability from the card Oracle text + the discriminating inline tests instead. The Rust engine gates (lint, both shards, Card data coverage) are still running and the branch is `BEHIND`.

**Recommendation: approve + enqueue as `bug` once the Rust gates settle green and the branch is updated.** The review judgment is complete; this holds only on CI + rebase.

matthewevans (approved):
**Approving — soulbond quoted-grant parser fix is correct, rules-verified, and settled green.**

## ✅ Clean
- **Right seam, real class.** `take_until(".")` was severing quoted granted abilities at the period *inside* the closing quote, so ~10 soulbond cards (Deadeye Navigator, Stonewright, Tandem Lookout, …) silently dropped their granted ability. The fix swaps to `TextPair::split_around` (case-preserving, so a granted `{1}{U}` reaches the cost parser un-lowercased) plus a quote-aware `strip_granted_predicate_period`, reusing the existing `matches_soulbond_paired_condition` and `parse_continuous_gets_has` seams — no new parser surface.
- **CR-verified:** CR 702.95 (Soulbond), 613.1f (ability-adding layer), 602.1 / 603.1 (activated/triggered ability grants) all check out.
- **Oracle text confirmed** verbatim on Scryfall for the affected cards.
- **Discriminating tests:** they destructure the produced AST (`GrantAbility { kind == Activated, cost.is_some() }` and `GrantTrigger`) and fail on the pre-fix empty-modifications behavior.
- **CI:** full rollup settled green (all Rust engine gates — fmt/clippy/test/coverage-gate, lint, both shards, Card data — COMPLETED/SUCCESS), merge CLEAN.

**Approving and enqueuing as `bug`.**

PR #5550, opened 2026-07-11
fix(parser): "share a type with it" in a mana-amount count binds to the triggering object

### LINKED ISSUE #5329: Mana Echoes adds no mana — Pretty much just that.
Reported in Discord: https://discord.com/channels/1485498006781427802/1524130182875713716/1524130182875713716

**Thread:** Mana Echoes adds no mana
**Discord thread id:** `1524130182875713716`
**Discord message id:** `1524130182875713716`
**Cards:** mana echoes

## Summary
Pretty much just that.

## Reported text

Pretty much just that. [[Mana Echoes]]

## Attachments

- [game-state-turn-32-2026-07-07T19-07-54-875Z.zip](https://cdn.discordapp.com/attachments/1524130182875713716/1524130183362383882/game-state-turn-32-2026-07-07T19-07-54-875Z.zip?ex=6a4ea058&is=6a4d4ed8&hm=7816c107a6a66b1b4904016feb81f0f7e6c4a7197fe99eaed98dda3aad6dd018&)

### mana echoes
Whenever a creature enters, you may add an amount of {C} equal to the number of creatures you control that share a creature type with it.

---
<sub>report_id: `discord:1524130182875713716:1524130182875713716:0` · discord: `1524130182875713716/1524130182875713716`</sub>

davion-knight:
Fixes #5329 (Mana Echoes). CR 106.1 + CR 608.2c.

"Whenever a creature enters, you may add an amount of {C} equal to the number of creatures you control that share a creature type with **it**." **added no mana.**

## Root cause

The mana-amount count parser dropped the parse context. The count filter `… that share a creature type with it` resolves "it" via `resolve_pronoun_target`, which needs `ctx.subject` (the trigger subject) to bind "it" to `TriggeringSource`. With no context it fell back to **`ParentTarget`** — but a triggered ability has no target, so the `ObjectCount` was always **0** → no mana.

The trigger effect parser *does* set the subject (`extract_trigger_subject_for_context`), but `try_parse_add_mana_effect` / `try_parse_amount_equal_to` took only `&str`, so the context never reached the count.

## Fix

Thread the context through the mana-amount path. `parse_cost_resource_ast` already holds the trigger `ctx`, so it now calls `try_parse_add_mana_effect_with_context` → `try_parse_amount_equal_to_with_context` → the existing ctx-aware `parse_cda_quantity_with_context`. "it" now binds to `TriggeringSource`, so the shared-type count sees the entering creature (verified: reference is now `TriggeringSource`, was `ParentTarget`).

This is a **class-fix** for any "add mana equal to `<count referencing the triggering object>`" trigger.

## Verification

- Fix confirmed via parse probe: the `SharesQuality` reference is now `TriggeringSource` (was `ParentTarget`).
- New test `shares_type_with_it_in_trigger_context_uses_triggering_source` (exercises the threaded context directly).
- `cargo check --tests` clean, clippy clean, parser-combinator gate exit 0, parser gap analysis unchanged (3977). The non-trigger callers keep a default context, so all existing mana parses are unchanged.

_Note: the full engine test-binary link is RAM-bound on my box, so the runtime end-to-end is left to CI._

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -15,7 +15,7 @@ use super::lower::{
     parse_for_each_multiplier_prefix, parse_multi_target_count_expr,
     parse_where_x_quantity_expression, strip_trailing_where_x,
 };
-use super::mana::{try_parse_activate_only_condition, try_parse_add_mana_effect};
+use super::mana::{try_parse_activate_only_condition, try_parse_add_mana_effect_with_context};
 use super::token::try_parse_token;
 use super::{
     attach_controller_if_absent, is_bare_object_pronoun, resolve_it_pronoun, ParseContext,
@@ -8270,7 +8270,7 @@ pub(super) fn parse_cost_resource_ast(
     })
     .is_some()
     {
-        return match try_parse_add_mana_effect(text) {
+        return match try_parse_add_mana_effect_with_context(text, ctx) {
             Some(Effect::Mana {
                 produced,
                 restrictions,
--- a/crates/engine/src/parser/oracle_effect/mana.rs
+++ b/crates/engine/src/parser/oracle_effect/mana.rs
@@ -21,9 +21,12 @@ use crate::types::mana::{
 use crate::types::zones::Zone;
 
 use super::super::oracle_keyword::parse_keyword_from_oracle;
-use super::super::oracle_quantity::{parse_cda_quantity, parse_event_context_quantity};
+use super::super::oracle_quantity::{
+    parse_cda_quantity, parse_cda_quantity_with_context, parse_event_context_quantity,
+};
 use super::super::oracle_target::parse_type_phrase;
 use super::super::oracle_util::{parse_mana_production, parse_number, TextPair};
+use crate::parser::oracle_ir::context::ParseContext;
 use crate::types::ability::TargetFilter;
 
 /// Bridge: run a nom combinator on a lowercase copy, mapping the consumed length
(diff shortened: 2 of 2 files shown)


matthewevans (commented):
**✅ Correct fix at the right seam — holding formal approval for CI.**

## ✅ Clean
- **Real card, verified text.** Mana Echoes' Oracle — "Whenever a creature enters, you may add an amount of {C} equal to the number of creatures you control that share a creature type with it." (Scryfall) — matches the test's clause exactly. Binding "it" to `TriggeringSource` is correct: the trigger fires on a creature entering, and "it" is that creature.
- **Right seam, class-general.** The fix threads `ParseContext` into the mana-amount count path (`try_parse_amount_equal_to_with_context` → `parse_cda_quantity_with_context`), the existing context-aware quantity authority. Any mana-count "…with it" clause now resolves the anaphor, not just Mana Echoes. Without it, "it" fell to an empty `ParentTarget` → counted nothing → added no mana (#5329).
- **No build break.** `try_parse_add_mana_effect` becomes `#[cfg(test)]`, but both production callers (`imperative.rs` cost-resource path and the recursive subject-strip in `mana.rs`) are rerouted to `_with_context`; the non-context fn now only serves the `mod tests` callers. Verified no other production caller in `crates/engine/src`.
- **Discriminating test** asserts `SharesQuality { reference: TriggeringSource }` — it would fail on the empty-`ParentTarget` bug.

**Recommendation: approve as `bug` once CI settles green on a rebased head** (currently BEHIND, CI pending). No changes requested.

matthewevans (approved):
**✅ Approving — CI resolved green, merge state CLEAN.** Correctness verified in the prior review (CR 205.4a supertype handling at the single subtype/type authority, reuses `parse_supertype_word`, discriminating tests). The head that went green is the reviewed head; enqueuing as `bug`.

Issue #88, opened 2026-04-24
Krosan Verge only fetching 1 land


whiteknightjanus:
Krosan Verge should allow you to fetch a Forest card and a Plains card, but only allows one.

matthewevans:
Verified fixed — Krosan Verge chained SearchLibrary(Forest) → ChangeZone(tapped) → SearchLibrary(Plains) → ChangeZone(tapped) → Shuffle correctly models both searches.

PR #5552, opened 2026-07-11
feat(engine): Captain America, First Avenger via UnattachFrom cost + divided damage

JacobWoodson:
Implements MAR #88 (Captain America, First Avenger, {R}{W}{U} 4/4). Adds a
reusable "unattach an attachment from the source as a cost" building block, not
a one-off:

- New AbilityCost::UnattachFrom { filter, count } + PayCostKind::UnattachFrom.
  The detached Equipment stays on the battlefield (CR 701.3d) and becomes the
  ability's cost-referent, captured as a CostPaidObjectSnapshot before detach
  (CR 608.2k / 400.7j) so "that Equipment's mana value" resolves via
  QuantityRef::ObjectManaValue { scope: CostPaidObject }.
- Throw: {3}, unattach an Equipment -> deal its mana value divided among 1-3
  targets. The divided total is unknown until the cost is paid, so distribution
  defers (extract_distribution_total None while MV=0) and re-fires after payment
  (CR 601.2c/601.2d/601.2h).
- Change B-1: the DistributeAmong resume was spell-only; it now routes activated
  abilities to push_activated_ability_to_stack vs spells to finalize_cast
  (CR 602.2b), so a divided-damage activated ability resolves as an
  ActivatedAbility, not a Spell. Also propagates distribute onto the targeted
  activation pending and guards the post-cost target re-check against re-opening
  an already-announced division.
- Catch: begin-combat-your-turn trigger attaches up to one target Equipment you
  control to the source, reusing existing Attach + trigger primitives.
- Interactive unattach choice surfaced from pay_activation_costs_after_target_selection
  (targeted-activation post-target path), mirroring the non-self exile detour;
  eligible choices bounded to MV >= announced target count so no impossible
  distribution is reachable and nothing is detached on an unpayable activation.
- AI values the unattach cost ~0 (Equipment is not lost). Single authorized
  mechanical no-op arm added to the dormant mtgish-import crate to keep the
  workspace compiling as AbilityCost grows.

Tests: 10 cast-pipeline tests (activation resolves as ActivatedAbility;
X-divide spell still resolves as Spell; chosen Equipment's MV drives damage;
unpayable-when-MV<targets; {3} charged; Catch attach/no-op/opponent-turn) plus
parser unit tests for the cost combinator and the "that Equipment's mana value"
demonstrative quantity.

### DIFF
--- a/crates/engine/src/analysis/ability_graph.rs
+++ b/crates/engine/src/analysis/ability_graph.rs
@@ -1539,6 +1539,7 @@ fn fold_cost(acc: &mut NodeAcc, cost: &AbilityCost) {
         | AbilityCost::PaySpeed { .. }
         | AbilityCost::ReturnToHand { .. }
         | AbilityCost::Unattach
+        | AbilityCost::UnattachFrom { .. }
         | AbilityCost::Mill { .. }
         | AbilityCost::Exert
         | AbilityCost::Reveal { .. }
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -13695,6 +13695,20 @@ pub(super) fn find_non_self_exile(
     }
 }
 
+/// CR 701.3d + CR 608.2k: Detect a non-self `UnattachFrom` activation cost
+/// (Captain America's Throw) requiring an interactive "unattach a matching
+/// attachment from the source" selection. Returns `(count, filter)`. The
+/// source-self `Unattach` unit variant returns `None` — it detaches the source
+/// Equipment itself and is auto-paid, never surfaced interactively. Recurses
+/// into `Composite`, mirroring `find_non_self_exile`.
+pub(super) fn find_unattach_from_cost(cost: &AbilityCost) -> Option<(u32, &TargetFilter)> {
+    match cost {
+        AbilityCost::UnattachFrom { filter, count } => Some((*count, filter)),
+        AbilityCost::Composite { costs } => costs.iter().find_map(find_unattach_from_cost),
+        _ => None,
+    }
+}
+
 /// CR 117.1 + CR 601.2b: Detect an `ExileWithAggregate` activation cost (Baron
 /// Helmut Zemo's Boast) requiring an interactive "exile any number reaching the
 /// aggregate threshold" selection. Returns a borrowed view of its parameters.
@@ -13868,6 +13882,38 @@ pub(crate) fn find_eligible_exile_for_cost_targets(
                 .unwrap_or_default()
         }
     }
+}
+
+/// CR 701.3d + CR 601.2b + CR 202.3: Battlefield attachments controlled by
+/// `player`, currently attached to `source`, matching `filter`, whose mana value
+/// is at least `n`. Mirrors `find_eligible_exile_for_cost_targets`. The `n`
+/// mana-value floor implements the divided-damage legality gate (CR 601.2c/M1):
+/// the chosen Equipment's mana value is the total damage divided among the
+/// announced targets, so it must be >= the target count. Pass `n = 0` for the
+/// generic eligibility count (no floor).
+pub(crate) fn find_eligible_unattach_for_cost_targets(
+    state: &GameState,
+    player: PlayerId,
+    source: ObjectId,
+    filter: &TargetFilter,
+    n: u32,
+) -> Vec<ObjectId> {
+    let ctx = super::filter::FilterContext::from_source(state, source);
+    state
+        .battlefield
+        .iter()
+        .copied()
+        .filter(|&id| {
+            let Some(obj) = state.objects.get(&id) else {
+                return false;
+            };
+            // CR 701.3d: only attachments currently attached to the source host.
+            obj.controller == player
+                && obj.attached_to.and_then(|t| t.as_object()) == Some(source)
+                && obj.effective_mana_value() >= n
+                && super::filter::matches_target_filter(state, id, filter, &ctx)
+        })
+        .collect()
 }
 
 fn find_one_of_cost(cost: &AbilityCost) -> Option<&Vec<AbilityCost>> {
@@ -15464,6 +15510,11 @@ pub fn handle_activate_ability(
         pending_target.activation_cost = ability_def.cost.clone();
         pending_target.activation_ability_index = Some(ability_index);
         pending_target.target_constraints = target_constraints;
+        // CR 601.2d: propagate the divided-effect flag so a targeted activated
+        // ability that divides damage/counters among its targets (Captain
+        // America's Throw) reaches the `DistributeAmong` step after its costs are
+        // paid. Mirrors the spell target-selection path (`pending_targets.distribute`).
+        pending_target.distribute = ability_def.distribute.clone();
         return Ok(WaitingFor::TargetSelection {
             player,
             pending_cast: Box::new(pending_target),
--- a/crates/engine/src/game/casting_costs.rs
+++ b/crates/engine/src/game/casting_costs.rs
@@ -1795,6 +1795,117 @@ pub(crate) fn handle_sacrifice_for_cost(
     Ok(waiting_for)
 }
 
+/// CR 701.3d + CR 608.2k + CR 601.2d: Complete a non-self `UnattachFrom` cost
+/// (Captain America's Throw) after the player selects which attachment(s) to
+/// unattach. Validates each chosen object is still a controlled battlefield
+/// attachment on the source matching `filter`, snapshots the first as the
+/// cost-referent BEFORE detaching (CR 608.2k — "that Equipment's mana value"),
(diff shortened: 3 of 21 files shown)


matthewevans (changes requested):
**🔴 Request changes — two CI-red blockers, both PR-introduced. The feature's own behavior test fails and the test file is a top-level binary.**

## 🔴 Blocker
- **The core mechanic doesn't attach.** `catch_attaches_equipment_to_self_on_your_turn` fails at `crates/engine/tests/captain_america_throw.rs:657` — *"Catch must attach the chosen Equipment to Captain America"* (`assert_eq!` on `Option<AttachTarget>`, left != right). The Catch ability's UnattachFrom-cost → re-attach path isn't producing the `AttachTarget`, so the feature is non-functional on its own test. This must be green before the design can be reviewed on the merits.
- **Top-level test binary.** `no_top_level_test_binaries::tests_dir_has_no_top_level_rs_files` fails at `crates/engine/tests/integration/no_top_level_test_binaries.rs:28`. `captain_america_throw.rs` was added directly under `crates/engine/tests/` — each top-level file compiles to its own ~130MB binary that re-links the entire engine on every change. Move the tests into `crates/engine/tests/integration/` and add a `mod captain_america_throw;` line to `crates/engine/tests/integration/main.rs`.

## 🟡 Non-blocking
- +1210/−18 files is a large surface for one review; a deeper architecture pass (divided-damage division correctness per CR 510.1c, the UnattachFrom cost seam) is deferred until CI is green, since the attach fix will likely move code.

**Recommendation: fix both blockers (make the Catch attach test pass; relocate the test file into `integration/` + register the `mod`), push, and I'll do the full architecture review on green.**

matthewevans:
Hi @JacobWoodson — heads up that **I've rebased this branch for you and force-pushed it** (`c4f2634b27` → `ce8f2cdc63`). Nothing was wrong with your work; the staleness was ours, so the fix was ours to do.

**Why it went stale.** `main` has been running an internal parser refactor (the U5/U6 campaign — `ClauseIrBuilder`/`ClauseDisposition`, and the `SpecialClause` enum being deleted outright). That churn invalidated your branch underneath you, across ~20 files in exactly the areas you touched. Asking you to chase that would have been asking you to win a race whose finish line keeps moving — `main` landed seven commits during a *single* rebase of a neighbouring PR. Our policy is that when our own refactor breaks a contributor's branch, we port it.

**Something worth knowing: your CI was never actually running.** The branch was `CONFLICTING`, and GitHub only creates the `pull_request`-triggered workflow runs when it can compute a merge ref. So this PR showed **2 passing checks** — which look like a green — when in reality none of the ~13 real jobs had ever executed. It's now at **15 checks** and genuinely running. If you ever see a suspiciously small check count, that's the tell.

**Two changes I made on top of your commits** (your commit is preserved as-is, under your authorship):

1. **A parser fix that is our bug, not yours.** Your card is the first to trigger it. The generic inline binary-choice splitter `try_parse_choose_one_of_inline` was matching the bare `" or "` inside a divided-damage target-count list — `"one, two, or three targets"` — and trial-parsing the orphan `"three targets"` tail back through the whole effect→subject→static→imperative cascade. That re-entry overflows small thread stacks, nondeterministically depending on stack headroom. Added a CR 601.2d bail-guard so distribution clauses skip the splitter and the dedicated `try_parse_distribute_damage` handler claims them, mirroring the four existing bail-guards.

2. **Moved `captain_america_throw.rs` into `tests/integration/`** and registered it in `tests/integration/main.rs`. Repo convention (enforced by the `no_top_level_test_binaries` test): a new top-level file in `crates/engine/tests/` becomes its own ~130 MB test binary that re-links the whole engine on every change.

Once CI reports, we'll review the substance on its merits. Since the parser guard has a pool-wide blast radius, I'll also be reading the parse-diff before signing off — that's on us, not you.

Thanks for the patience, and sorry the branch sat conflicted.

JacobWoodson:
Thanks for the review — both flagged blockers are fixed in `c4f2634`. Full root-cause + fix context below so the re-review has everything it needs.

## Blocker 1 — Catch doesn't attach (`catch_attaches_equipment_to_self_on_your_turn`)

Two distinct causes, both addressed:

**(a) The CI assertion failure itself (0.095s, "not attached", no crash) was a runtime optional-target outcome.** Catch targets *"up to one target Equipment you control"* (min 0, max 1). With exactly one legal Equipment, whether the engine auto-selects it or surfaces the optional target prompt varies by the auto-target path (it auto-attached on the dev box, did not on this CI). Per CR 603.3d the controller chooses, so the test now **answers the prompt if one is offered and asserts the attach either way** — deterministic on both paths. Since targeting the Equipment is always a legal completion, there is no auto-decline outcome to worry about.

**(b) While root-causing, we found and fixed a PRE-EXISTING parser stack overflow on this card's Throw line** (not the card's UnattachFrom/attach logic — this card is just the first to hit it). The generic inline binary-choice splitter `try_parse_choose_one_of_inline` (`crates/engine/src/parser/oracle_effect/mod.rs`) matched the bare `" or "` inside the divided-damage count list *"…divided as you choose among one, two, or three targets"* and trial-parsed the orphan `"three targets"` tail back through the full effect → subject → static → imperative cascade, re-entering the clause parser several levels deep. Its large per-frame AST locals overflow small (~1–2 MB) thread stacks nondeterministically by stack headroom — causing hard crashes / flakiness in small-stack environments. Fix: a `CR 601.2d` bail-guard that declines distribution clauses (`"divided as you choose among"` / `"distributed among"` / `"divided evenly"`) so the dedicated `try_parse_distribute_damage` handler claims the whole clause. It mirrors the four existing comparator/target/type/restriction bail-guards in the same function and changes which clauses are *accepted* for zero cards — it only stops the erroneous split.

## Blocker 2 — top-level test binary

Moved `crates/engine/tests/captain_america_throw.rs` → `crates/engine/tests/integration/captain_america_throw.rs` and registered it via `mod captain_america_throw;` in `tests/integration/main.rs`.

## Verification (local)

- Parser fix: `engine` lib suite **15,690 passed / 0 failed**; `oracle_effect` parser suite **2,522 / 0**; parser combinator gate + string-dispatch gate pass.
- Overflow eliminated: the exact failing test run 30× as a standalone binary → **30/30 pass, 0 overflow** (was 30/30 overflow at the small test-thread stack).
- Full integration harness on `origin/main` + this branch → **13 captain-related tests green**, including `catch_attaches_equipment_to_self_on_your_turn` and the other MSH captain cards (no regressions).
- Parse-diff bot on this PR confirms the intended AST: `DealDamage (amount=referenced object's mana value, kind=activated, target=any target, targets=1-3)`.

## Heads-up on CI status

This is a fork PR from a **non-maintainer** account, so the full **CI** / **AI gate** workflows on the new push (`c4f2634`) are gated behind a maintainer's *"Approve and run workflows"* — they haven't run yet for that reason, not a failure. A maintainer approval will re-run them against the fix.

matthewevans (approved):
Approving. This supersedes my earlier `CHANGES_REQUESTED` — both of its reasons (the failing test and the `no_top_level_test_binaries` violation) were resolved by the maintainer rebase, so that formal state was stale.

**Evidence:**
- **CI is real and green** — 13 pass / 0 fail / 3 skipped. Worth stating plainly: before the rebase this PR showed **2 passing checks and looked green while nothing had run**. A branch that is `CONFLICTING` at push time gets no `refs/pull/N/merge`, so GitHub never creates the `pull_request` workflow runs; the only things executing were a `pull_request_target` label job and an external scanner. The port restored actual CI.
- **Parse-diff: exactly 1 card / 2 signatures**, both Captain America, First Avenger — zero collateral across the pool. The divided-damage clause goes from a mangled `deal (kind=activated)` to a correct `DealDamage (amount=referenced object's mana value, target=any target, targets=1-3)`.

**On the maintainer commit** (the CR 601.2d bail-guard): I wrote it, so I had it independently audited rather than trusting my own read. The audit found two real defects, both now fixed: a needle (`"distributed among"`) that matches **0 of 34,632 cards** in the Atomic pool, and a phrase list duplicated between the guard and the handler — two lists that must agree forever, with nothing enforcing it. They are now one (`lower::DISTRIBUTION_KEYWORDS`).

The audit's third finding is real and **not** fixed here: the guard keys on the distribution *verb*, but the hazard is the target-*cardinality list*, so ~32 counter-distribution cards ("Distribute three +1/+1 counters among one, two, or three target creatures") sit on the same stack cliff — and the underlying re-entry has no depth bound at all. That is a pre-existing condition on `main` which this PR neither introduces nor worsens, and fixing it means redesigning the cardinality vocabulary across three call sites. It belongs in its own PR with its own parse-diff, not smuggled into a contributor's card branch. Filed as #5613.

Thanks for the patience, @JacobWoodson — the branch sat conflicted because of our own U5/U6 refactor, not anything you did.

PR #5553, opened 2026-07-11
feat(parser): filter "Legendary spells you cast cost {N} less" by supertype

minion1227:
CR 205.4a

## Bug

"Legendary spells you cast cost {1} less to cast" (**Kethis, the Hidden Hand**) must restrict the reduction to *legendary* spells. But `parse_cost_mod_spell_type_prefix` built its filter with `parse_type_phrase`, which doesn't consume a lone **supertype** word ("legendary" requires a following type noun) — so the type prefix fell through to `None` and the static reduced the cost of **every** spell the controller cast.

- Before: `ModifyCost { spell_filter: None, .. }` (all spells cheaper).
- After: `ModifyCost { spell_filter: Some([Card] + HasSupertype(Legendary)), .. }`.

Card-**type** prefixes ("Artifact spells", "Creature spells") already worked — only bare-supertype prefixes dropped.

## Fix

Add a `parse_bare_supertype_spell_filter` arm alongside the existing bare-**color** fallback (both are words `parse_type_phrase` leaves unconsumed), emitting a `HasSupertype` card filter. It generalizes to every supertype — "Snow spells" → `HasSupertype(Snow)` — mirroring the bare-color handling already in this seam.

- The fix sits exactly where the bare-color fallback already lives (`parse_cost_mod_spell_type_prefix`), so it composes with the existing color / "of the chosen type" / "that's a …" qualifiers rather than adding a parallel path.
- Card-type and bare-spell subjects are untouched (regression-tested): "Artifact spells" still → `[Artifact]`, "Spells you cast" still → `None`.

## Tests

- `static_legendary_spells_cost_less_keeps_supertype_filter` — Kethis → `HasSupertype(Legendary)` on the spell filter.
- `static_artifact_spells_cost_less_keeps_type_filter` — regression: "Artifact spells" still resolves to `[Artifact]`.
- Existing `static_spells_cost_less` (bare → `None`) unchanged. Full `parser::oracle_static::tests` (1042 passed, 0 failed).

## CR verification

- **CR 205.4a** — the supertypes (basic, legendary, ongoing, snow, world) — verified in `docs/MagicCompRules.txt`.

## AI-contributor notes

- **Rules-correct:** only legendary spells are discounted per CR 205.4a; other spells pay full cost.
- **Builds the class:** the arm covers any bare-supertype spell-cost subject, current or future.
- **Verified:** live-parsed Kethis and an Artifact/Snow control before/after; `cargo fmt`, CI-exact `clippy -p engine --all-targets --features proptest -D warnings`, and the full oracle_static module are green.

### DIFF
--- a/crates/engine/src/parser/oracle_static/static_helpers.rs
+++ b/crates/engine/src/parser/oracle_static/static_helpers.rs
@@ -38,6 +38,21 @@ pub(crate) fn parse_chosen_name_source_filter(subject_lower: &str) -> Option<Tar
 /// CR 601.2f: Parse the spell-type prefix of a cost-modification line before
 /// `"cost"`. Handles compound subjects such as Goblin Anarchomancer's
 /// "Each spell you cast that's red or green" via `parse_that_clause_suffix`.
+/// CR 205.4a: A bare supertype spell subject ("Legendary spells you cast cost
+/// {1} less", Kethis, the Hidden Hand) — `parse_type_phrase` doesn't consume a
+/// lone supertype word (it requires a following type noun), so the restriction
+/// would otherwise drop and reduce the cost of EVERY spell. Emit a `HasSupertype`
+/// card filter instead.
+fn parse_bare_supertype_spell_filter(base: &str) -> Option<TargetFilter> {
+    let lower = base.to_lowercase();
+    let (_, supertype) = all_consuming(nom_target::parse_supertype_word)
+        .parse(lower.as_str())
+        .ok()?;
+    Some(TargetFilter::Typed(TypedFilter::card().properties(vec![
+        FilterProp::HasSupertype { value: supertype },
+    ])))
+}
+
 fn parse_cost_mod_spell_type_prefix(type_desc: &str) -> Option<TargetFilter> {
     let base = type_desc.trim();
     let base = tag::<_, _, OracleError<'_>>("each ")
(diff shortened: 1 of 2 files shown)


matthewevans (commented):
**✅ Correct fix at the right seam — holding formal approval for CI (Rust shard + card-data still running).**

## ✅ Clean
- **Real card, verified rule.** Kethis, the Hidden Hand — *"Legendary spells you cast cost {1} less to cast."* Bare supertype subject, so `parse_type_phrase` (which requires a trailing type noun) drops it and the discount would wrongly apply to **every** spell. CR 205.4a confirmed (`legendary` is a supertype).
- **Right seam, parameterized not proliferated.** `parse_bare_supertype_spell_filter` slots into the same `remainder.is_empty()` arm that already handles bare color words via `.or_else(...)` — it extends the existing bare-word pattern rather than adding a parallel path.
- **Nom-combinator dispatch.** `all_consuming(nom_target::parse_supertype_word)` correctly matches a lone supertype only (rejects `"legendary creature"`), reusing the shared combinator.
- **Discriminating tests.** `static_legendary_spells_cost_less_keeps_supertype_filter` destructures `HasSupertype{Legendary}` (fails on the pre-fix drop-the-restriction bug); the companion regression asserts a card-TYPE subject (`"Artifact spells"`) still resolves to its type filter, guarding the new arm's scope.

**Recommendation: approve as `bug` once both Rust shard 2/2 and Card-data settle green.**

matthewevans (approved):
**Approve — CI settled fully green on the reviewed head (`72ac6944e2`); enqueuing as `bug`.**

## ✅ Clean
- **Right seam.** `parse_bare_supertype_spell_filter` (bare `all_consuming(parse_supertype_word)`) is chained into the existing bare-**color** spell-filter arm rather than added as a sibling — parameterize-don't-proliferate. The fix restores `spell_filter: Some([Card] + HasSupertype(Legendary))` where the old `parse_type_phrase` path fell through to `None` and cheapened *every* spell.
- **CR 205.4a** verified verbatim in `docs/MagicCompRules.txt` ("supertypes are basic, legendary, ongoing, snow, and world") — a lone supertype word legitimately has no following type noun, which is exactly the gap `parse_type_phrase` couldn't consume.
- **Discriminating tests**: destructures `HasSupertype(Legendary)` (fails on the pre-fix `None`) plus an `Artifact spells` regression guard confirming card-**type** prefixes still parse.
- All Rust gates (fmt/clippy/test/coverage-gate, shards 1&2), Card data, and frontend checks report SUCCESS; mergeStateStatus CLEAN.

**Recommendation: approve, label `bug`, enqueue.**

PR #5555, opened 2026-07-11
fix(parser): honor "if it dealt combat damage this turn" intervening-if on dies triggers

real-venus:
**Bug:** Wave of Rats ("When this creature dies, if it dealt combat damage to a player this turn, return it to the battlefield under its owner's control.") parsed with `condition: null` — the intervening-if was **silently swallowed**, so the creature returned **unconditionally**.

**Fix:** The source-referential condition combinator `parse_source_dealt_damage_to_opponent_this_turn` only accepted the `"~"`/`"this creature"` subject and lacked a `"combat"` qualifier. Extend it with two leaf axes:
- the `"it"`/`"this permanent"` source anaphor, and
- an optional `"combat "` qualifier → `DamageKindFilter::CombatOnly` vs `Any`.

The clause now lifts through the existing bridge to `TriggerCondition::QuantityComparison` over `QuantityRef::DamageDealtThisTurn(SelfRef → Player, CombatOnly) ≥ 1`.

**No new variant, no runtime change** — a pure generalization of one existing combinator. Covers a class: Wave of Rats is fixed, and the two already-supported cards (Dunerider Outlaw, Whirling Dervish, "this creature dealt damage to an opponent") stay green (the no-`combat` path still yields `Any`). Verified locally: both new parser tests pass, the 19 `dealt_combat` sibling tests stay green, clippy clean.

CR 603.4 (intervening-if), CR 120.1 (source of damage), CR 120.2a (combat damage).

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -399,8 +399,22 @@ fn parse_player_dealt_combat_damage_by_source_this_turn(
 fn parse_source_dealt_damage_to_opponent_this_turn(
     input: &str,
 ) -> OracleResult<'_, StaticCondition> {
-    let (rest, _) = alt((tag("~"), tag("this creature"))).parse(input)?;
-    let (rest, _) = tag(" dealt damage to ").parse(rest)?;
+    // CR 120.1 + CR 120.2a + CR 603.4: "<source-anaphor> dealt [combat] damage to a
+    // player/opponent this turn" — the *dealing* direction (source = the ability's
+    // own permanent). "it" is the source anaphor for triggered-ability
+    // intervening-ifs (Wave of Rats' dies trigger); "~"/"this creature"/"this
+    // permanent" cover the self-ref forms. The optional "combat" qualifier narrows
+    // the damage channel (CR 120.2a) so combat-only history is required.
+    let (rest, _) = alt((
+        tag("~"),
+        tag("this creature"),
+        tag("this permanent"),
+        tag("it"),
+    ))
+    .parse(input)?;
+    let (rest, _) = tag(" dealt ").parse(rest)?;
+    let (rest, combat) = opt(tag("combat ")).parse(rest)?;
+    let (rest, _) = tag("damage to ").parse(rest)?;
     let (rest, target) = alt((
         value(
             TargetFilter::Typed(TypedFilter::default().controller(ControllerRef::Opponent)),
@@ -410,6 +424,11 @@ fn parse_source_dealt_damage_to_opponent_this_turn(
     ))
     .parse(rest)?;
     let (rest, _) = tag(" this turn").parse(rest)?;
+    let damage_kind = if combat.is_some() {
+        DamageKindFilter::CombatOnly
+    } else {
+        DamageKindFilter::Any
+    };
     Ok((
         rest,
         make_quantity_ge(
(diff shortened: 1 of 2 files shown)


matthewevans (commented):
**Correct, class-general parser fix — approving on settled green.** Extends one existing combinator along two independent leaf axes rather than adding a variant or a card special-case. Verified against Scryfall, the CR, and the diff.

## ✅ Verified clean

- **Premise confirmed.** Wave of Rats' Oracle text (Scryfall): *"When this creature dies, if it dealt combat damage to a player this turn, return it to the battlefield under its owner's control."* The `it` source anaphor + `combat` qualifier are exactly what the parser was dropping to `condition: null`.
- **Parameterized, not proliferated** (`oracle_nom/condition.rs:` the two new axes): source anaphor becomes `alt((tag("~"), tag("this creature"), tag("this permanent"), tag("it")))` and the channel becomes `opt(tag("combat "))` → `DamageKindFilter::CombatOnly` vs `Any`. Two orthogonal leaf axes on one combinator, all nom (`tag`/`alt`/`opt`), delegating through `parse_inner_condition` — the single condition authority. No new variant.
- **Sibling class preserved.** The no-`combat` path still yields `DamageKindFilter::Any`, so `"this creature dealt damage to an opponent this turn"` cards (e.g. the Whirling-Dervish family at `oracle_trigger_tests.rs:8308`) are untouched.
- **Discriminating test.** `parse_it_dealt_combat_damage_to_a_player_this_turn` destructures the full `QuantityRef::DamageDealtThisTurn { damage_kind: CombatOnly, source: SelfRef, target: Player } GE 1` — it fails on the pre-fix silent swallow, not just "parses without error." The test module is live (registered via `#[path]` in `oracle_trigger.rs`), so it's not inert.
- **CR annotations accurate** (grep-verified in `docs/MagicCompRules.txt`): CR 603.4 (intervening-if), CR 120.1 (source of damage), CR 120.2a (combat damage). All three are cited correctly.

## Recommendation

**Approve as `bug` once CI settles green.** The only thing held back is CI: the head is `BEHIND` and the Rust shards/coverage-gate are still pending. Rebase onto `main`; on resolved green (both shards + coverage-gate + parser gate) this merges as a bug fix.

matthewevans (approved):
**Approve — CI now fully green on the reviewed head `0a5630e6`; enqueuing as `bug`.**

## ✅ Clean
Confirming the [prior review](https://github.com/phase-rs/phase/pull/5555#pullrequestreview-4676673147) on this same commit — all gates have since resolved SUCCESS (Rust fmt/clippy/parser gate, both test shards, Card data generate/validate/coverage, coverage-gate), `mergeStateStatus: CLEAN`.

The fix is class-general: `parse_source_dealt_damage…` is extended along two orthogonal leaf axes — source anaphor `alt((~, this creature, this permanent, it))` and `opt(tag("combat "))` → `CombatOnly` vs `Any` — with no new variant and the sibling `Any` path preserved, delegating through `parse_inner_condition`. CR 603.4 (intervening-if checked on trigger + on resolution), CR 120.1/120.2a (combat damage) grep-verified; Wave of Rats Oracle text confirmed. The discriminating test destructures `CombatOnly / SelfRef / Player GE 1` and is registered via `#[path]`.

**Recommendation: approving and enqueuing as `bug` (dropped intervening-if condition was a defect — dies triggers fired without the combat-damage gate).**

Issue #2856, opened 2026-06-10
Shatterskull Smashing: divided-damage softlock when X is less than the number of chosen targets (X=1, two targets)


matthewevans:
## Oracle text (verified from client/public/card-data.json)
```
Shatterskull Smashing deals X damage divided as you choose among up to two target creatures and/or planeswalkers. If X is 6 or more, Shatterskull Smashing deals twice X damage divided as you choose among them instead.
```

## Reported behavior
> If I cast Shatterskull Smashing with x=1, the game gets stuck because it requires at least 1 damage dealt to each target. It should not require that.

## Expected vs actual
- **Expected:** With 'divided as you choose among up to two targets', the controller may target one creature/planeswalker and assign all X (=1) to it; a target need not each receive at least 1 if the player chose 'up to' fewer targets. Damage division should never softlock when X is smaller than the number of selectable targets.
- **Actual:** The game gets stuck because it requires at least 1 damage to each of two targets, which is impossible at X=1.

## Parse evidence
Parses to `Effect::DealDamage` (divided damage). The division/allocation step enforces a >=1-per-target minimum across all offered targets instead of letting the player allocate all damage to a single chosen target. Distinct from #1978 (one-target-selection bug).

Source: https://discord.com/channels/1485498006781427802/1512549001461891234/1512549001461891234

PR #5556, opened 2026-07-11
fix(engine): re-derive target-dependent cost surcharge on the X+distribute casting route

rykerwilliams:
## Summary

Follow-up to #5545 (Fireball's parser misparse — that PR makes `strive_cost` correctly get populated). This PR fixes a separate, independent runtime bug: even with `strive_cost` populated, [[Fireball]]'s ({X}{R}: "This spell costs {1} more to cast for each target beyond the first.\nFireball deals X damage divided evenly, rounded down, among any number of targets.") actual in-game cost never scaled with target count.

## Root cause

Fireball's specific casting shape — an `{X}` cost combined with "divided evenly among any number of targets" — triggers a sequence where:

1. `{X}` is announced first (CR 601.2b) and the cost gets locked in via `apply_post_x_cost_modifiers` **before any targets exist**, so the per-target surcharge loop (`for _ in 1..target_count`) is a no-op at that point (`target_count == 0`).
2. Targets are chosen after (CR 601.2c).
3. Because the spell needs to "divide" its effect among the now-known targets, the engine pauses at `WaitingFor::DistributeAmong` (CR 601.2d) rather than proceeding straight to payment.
4. The `GameAction::DistributeAmong` handler resumes by calling `casting_costs::finalize_cast` **directly** with the stale, pre-target cost — completely skipping `check_additional_cost_or_pay_with_distribute`, the only place `apply_target_dependent_cost_modifiers` (the CR 601.2f surcharge authority) actually runs.

## Fix

Route through `casting_costs::finish_pending_cast_cost_or_pay` — the same cost-determination authority every other casting route already uses to go from "targets known" to "cost determined, locked in, paid" — instead of calling `finalize_cast` directly. This is a general seam fix, not hardcoded to Fireball: **Fireball is currently the only real card in Magic combining an `{X}` cost, a strive-shaped per-target surcharge, and a distribute-among-targets effect** (verified via Scryfall across two rounds of review), so the fix automatically covers any future card with this shape.

This requires a rollback wrapper. `state.pending_cast` is `.take()`n (set to `None`) before the call, and `finish_pending_cast_cost_or_pay`'s downstream chain has no restore-on-error of its own. If the recomputed (correctly higher) cost turns out unpayable, a bare `Err` would leave `state.waiting_for` stale as `DistributeAmong` while `pending_cast` is gone — a resubmitted `DistributeAmong` action would then silently fall through to an unrelated resolution-time-continuation branch instead of being cleanly rejected. The fix mirrors `finalize_mana_payment`'s existing `pending_for_restore` clone-and-restore-on-`Err` pattern (CR 601.2h: "Unpayable costs can't be paid").

**Extended defensively to a second call site found during review**: `engine_resolution_choices.rs`'s `NamedChoice` handler has the identical anti-pattern in its spell-cast resume branch (`activation_ability_index == None`) — also calls `finalize_cast` directly with a stale cost. No currently-shipped card was confirmed to reach it with a target-dependent cost modifier, but the fix is mechanical and the pattern is exactly the same, so it's fixed rather than left as a known latent gap.

A second near-miss (`finalize_mana_payment`'s two `pending.distribute`-gated blocks) was investigated and found **provably unreachable** for every real card today — git-verified that the #2856 divide-softlock gate (`ability_distribution_pool_needs_chosen_x`) predates and is an ancestor of the commits shaping those blocks. Documented with a cross-reference comment; no functional change needed there.

Also corrects `DistributionUnit::EvenSplitDamage`'s doc comment, which incorrectly claimed it always bypasses `WaitingFor::DistributeAmong` — true only for the non-deferred-target-selection flow, not for a deferred-selection card like Fireball.

## Test note

The new test pins the surcharge via a new `with_strive_cost` test builder rather than parsing Fireball's full real Oracle text, because that text currently mis-parses on `origin/main` into an unrelated, separate bug (a spurious static cost modifier) — that's PR #5545's territory, not this branch's job, and is documented in the test module doc.

## Files changed

- `crates/engine/src/game/engine.rs` — the primary fix (`DistributeAmong` handler)
- `crates/engine/src/game/engine_resolution_choices.rs` — the identical fix applied to the `NamedChoice` handler's spell-cast resume branch
- `crates/engine/src/game/casting_costs.rs` — documentation-only cross-reference comments (no functional change)
- `crates/engine/src/types/game_state.rs` — `DistributionUnit::EvenSplitDamage` doc correction
- `crates/engine/src/game/scenario.rs` — `with_strive_cost` test builder
- `crates/engine/tests/integration/fireball_x_cost_surcharge_timing.rs` (new) — discriminating runtime tests
- `crates/engine/tests/integration/main.rs` — registers the new test module

## CR references

CR 207.2c, CR 601.2b, CR 601.2c, CR 601.2d, CR 601.2f, CR 601.2h

## Verification

- [x] `cargo fmt --all` clean
- [x] `cargo clippy -p engine --lib -- -D warnings` clean
- [x] `./scripts/check-parser-combinators.sh`: no-op (no parser files touched)
- [x] `cargo test -p engine --lib`: 16104 passed, 0 failed
- [x] `cargo test -p engine --test integration`: 2665 passed, 0 failed
- [x] Live-revert-and-rerun performed independently twice (by the implementer and by an independent reviewer): reverting the cost-authority swap fails the surcharge test; reverting *only* the rollback wrapper (keeping the swap) fails the state-coherence assertions specifically, proving the wrapper is load-bearing, not decorative
- [x] Three rounds of independent plan review (closed: a test-file dependency mixup later clarified as belonging to a separate not-yet-merged PR, a genuinely-investigated near-miss call site proven unreachable via git history, a corrected card-class table, and a traced-and-fixed state-corruption risk in the naive version of this fix) plus a review-impl pass (found the second `NamedChoice` call site, since fixed)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/crates/engine/src/game/casting_costs.rs
+++ b/crates/engine/src/game/casting_costs.rs
@@ -8671,6 +8671,20 @@ pub fn finalize_mana_payment(
             );
         }
 
+        // NOTE: This branch is provably unreachable for every currently-implemented
+        // distribute-unit + {X}-cost card (Fireball and siblings) — the CR 601.2c/d
+        // gate in `ability_utils::ability_distribution_pool_needs_chosen_x` (added by
+        // issue #2856) forces `WaitingFor::ChooseXValue` before target selection
+        // whenever the divided amount is `Effect::DealDamage`/`Effect::PutCounter`
+        // with an unresolved X reference, which makes `chosen_x` always already set
+        // by the time `maybe_pause_for_cast_distribution` runs in
+        // `casting_targets::handle_select_targets`/`handle_choose_target` — so THAT
+        // path always wins first. If a future card's distribute-unit amount is
+        // neither `DealDamage` nor `PutCounter` (the one shape this gate does not
+        // cover), re-verify whether it also needs the CR 601.2f target-dependent
+        // cost recompute this branch currently skips (see
+        // `casting::apply_target_dependent_cost_modifiers`) before assuming this
+        // code path is exercised by existing coverage.
         if let Some(unit) = pending.distribute {
             // CR 601.2d: X-spell distribution — pay mana first to determine X, then
             // trigger DistributeAmong with total = X.
@@ -8863,6 +8877,20 @@ pub fn finalize_mana_payment_with_phyrexian_choices(
             );
         }
 
+        // NOTE: This branch is provably unreachable for every currently-implemented
+        // distribute-unit + {X}-cost card (Fireball and siblings) — the CR 601.2c/d
+        // gate in `ability_utils::ability_distribution_pool_needs_chosen_x` (added by
+        // issue #2856) forces `WaitingFor::ChooseXValue` before target selection
+        // whenever the divided amount is `Effect::DealDamage`/`Effect::PutCounter`
+        // with an unresolved X reference, which makes `chosen_x` always already set
+        // by the time `maybe_pause_for_cast_distribution` runs in
+        // `casting_targets::handle_select_targets`/`handle_choose_target` — so THAT
+        // path always wins first. If a future card's distribute-unit amount is
+        // neither `DealDamage` nor `PutCounter` (the one shape this gate does not
+        // cover), re-verify whether it also needs the CR 601.2f target-dependent
+        // cost recompute this branch currently skips (see
+        // `casting::apply_target_dependent_cost_modifiers`) before assuming this
+        // code path is exercised by existing coverage.
         if let Some(unit) = pending.distribute {
             // CR 601.2d: X + distribution + Phyrexian is extremely rare (no known current cards).
             // Fall through to the auto-decision distribution path for safety — the Phyrexian
--- a/crates/engine/src/game/engine.rs
+++ b/crates/engine/src/game/engine.rs
@@ -4833,20 +4833,45 @@ fn apply_action(
 
             // CR 601.2d: Resume casting pipeline after distribution.
             if state.pending_cast.is_some() {
-                // Mid-cast distribution: resume finalize_cast to commit the spell.
+                // CR 601.2c + CR 601.2d + CR 601.2f: Targets and their division are now
+                // committed, so the total cost — including any target-dependent
+                // surcharge (Strive, CR 207.2c) — is finally determinable. Route through
+                // the single cost-determination authority every other post-target-
+                // selection path uses (`casting_targets::handle_select_targets` /
+                // `handle_choose_target`) instead of calling `finalize_cast` directly
+                // with the stale cost that was locked in at `ChooseXValue` time, before
+                // targets (and hence any per-target surcharge) were known.
                 let pending = state.pending_cast.take().unwrap();
-                casting_costs::finalize_cast(
+                // CR 601.2h ("Unpayable costs can't be paid"): mirror
+                // `finalize_mana_payment`'s `pending_for_restore` pattern
+                // (casting_costs.rs ~8623-8627/8778-8787) — `finish_pending_cast_cost_or_pay`'s
+                // downstream chain has no restore-on-error wrapper of its own, and
+                // `state.pending_cast` is already `None` here (unlike
+                // `handle_select_targets`, whose `pending_cast` lives inside the
+                // `WaitingFor::TargetSelection` variant and so is never destructively
+                // taken). Without this clone-and-restore, a recomputed cost that turns
+                // out unpayable would return `Err` with `state.pending_cast` gone while
+                // `state.waiting_for` still reports `DistributeAmong` — a resubmitted
+                // `DistributeAmong` action would then fall through to the
+                // resolution-time continuation branch below instead of being cleanly
+                // rejected.
+                let pending_for_restore = pending.clone();
+                let ability = pending.ability.clone();
+                let cost = pending.cost.clone();
+                match casting_costs::finish_pending_cast_cost_or_pay(
                     state,
                     p,
-                    pending.object_id,
-                    pending.card_id,
-                    pending.ability,
-                    &pending.cost,
-                    pending.casting_variant,
-                    pending.cast_timing_permission,
-                    pending.origin_zone,
+                    *pending,
+                    ability,
+                    cost,
                     &mut events,
-                )?
+                ) {
+                    Ok(waiting_for) => waiting_for,
+                    Err(err) => {
+                        state.pending_cast = Some(pending_for_restore);
+                        return Err(err);
+                    }
+                }
             } else if let Some(mut pending_trigger) = state.pending_trigger.take() {
                 // CR 601.2d + CR 603.3d: Triggered abilities divide effects
                 // while being put on the stack. The chosen per-target amounts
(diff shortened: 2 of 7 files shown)


matthewevans (commented):
**Verdict: architecturally correct seam fix — approving on settled green.** This routes the `{X}`+distribute casting resume through the single post-target cost authority instead of paying a stale pre-target cost, and it does so as a general fix, not a Fireball special-case.

## ✅ Clean
- **Right seam (single cost authority).** `engine.rs:4836` — the `DistributeAmong` resume previously called `casting_costs::finalize_cast` directly with `pending.cost`, the cost locked in at `ChooseXValue` time before targets existed, so the CR 601.2f per-target surcharge (`apply_target_dependent_cost_modifiers`) never ran. The fix routes through `finish_pending_cast_cost_or_pay` (`casting_costs.rs:3159`) — the same authority every post-target-selection path uses. That is the correct consolidation, not a bolt-on.
- **Softlock-preventing rollback, verified against the cited precedent.** Because `state.pending_cast` is destructively `.take()`n here (unlike `handle_select_targets`, whose pending lives inside the `TargetSelection` variant), a bare `Err` from a recomputed-unpayable cost would strand `waiting_for: DistributeAmong` with `pending_cast: None` — a resubmitted `DistributeAmong` would fall through to the resolution-time continuation branch instead of being cleanly rejected. The clone-and-restore-on-`Err` mirrors `finalize_mana_payment`'s `pending_for_restore` pattern, which I confirmed lives at `casting_costs.rs:8627/8784` and `8820/8978`. CR 601.2h verbatim: "Unpayable costs can't be paid."
- **Second call site fixed consistently.** `engine_resolution_choices.rs:3970` (the `NamedChoice` spell-cast resume, `activation_ability_index == None`) carries the identical anti-pattern and gets the identical fix + rollback. Mechanical and correct to fix rather than leave latent.
- **CR annotations all grep-verified** against `docs/MagicCompRules.txt`: 207.2c (ability words — "strive" is in the list), 601.2b (X announced), 601.2c (targets), 601.2d (division), 601.2f (total cost locked in), 601.2h (unpayable). Accurate.
- **General test primitive + discriminating registered test.** `scenario.rs` `with_strive_cost` is a reusable builder (pins `strive_cost` directly), not a Fireball hack. `fireball_x_cost_surcharge_timing.rs` drives the real `ChooseX → targets → DistributeAmong` sequence and asserts the mana pool reflects the per-target surcharge — fails on the pre-fix single-target cost — and isolates the #5545 parser misparse by using only the verbatim distribution clause. `mod` registered in `main.rs`.

## 🟡 Non-blocking
- The `casting_costs.rs` / `game_state.rs` changes are documentation-only (cross-reference comment + `DistributionUnit::EvenSplitDamage` doc correction) — no functional risk. The "provably-unreachable" `finalize_mana_payment` near-miss is left documented rather than patched, which is the right call for a genuinely unreachable path.

**Recommendation: approve + merge as `bug` once Rust lint + both shards + card data settle green** (currently pending on head `c068e099`). This is a runtime defect fix (Fireball's cost never scaled with targets), so `bug`, not `enhancement`.

matthewevans (changes requested):
**Request changes — the engine fix is sound (both test shards pass), but two clippy `-D warnings` errors in the new test file abort the lint job and cascade the coverage-gate red.**

## 🔴 Blocker — `clippy::never_loop` at `crates/engine/tests/integration/fireball_x_cost_surcharge_timing.rs:122`
Every arm of the `match` inside `for _ in 0..40` diverges (`return` on `ChooseXValue`, `panic!` on `TargetSelection`, `_ => return`), so the loop body can never reach a second iteration — clippy reports "this loop never actually loops" and the `0..40` bound is dead. Decide the intent:
- If a single check suffices (the state is `ChooseXValue` on entry), drop the loop and keep just the `match`.
- If the intent is to poll until `ChooseXValue` appears, the `_` arm must not `return` — it has to advance the pending state and `continue`, otherwise it silently returns on the first non-matching state.

## 🔴 Blocker — `clippy::manual_is_multiple_of` at `fireball_x_cost_surcharge_timing.rs:192`
`assert!(n > 0 && total % n == 0, ...)` → replace `total % n == 0` with `total.is_multiple_of(n)` (clippy's suggested fix).

## ✅ Clean (unchanged from prior review)
The runtime fix itself is right-seam and CR-correct — both Rust test shards (1/2, 2/2) pass on this head. Routing through `finish_pending_cast_cost_or_pay` with clone-and-restore-on-`Err` rollback, both call sites fixed, CRs 207.2c / 601.2b–h grep-verified. No architecture changes needed.

**Recommendation: fix the two lint errors on the test file (no engine change required); approve and merge as `bug` on settled green.**

rykerwilliams:
Fixed in c8bfd6ff8.

- \`never_loop\`: \`drive_choose_x\`'s loop never actually looped — every match arm diverges (return/panic), so the \`0..40\` bound was dead. Simplified to a single match, with the fallthrough arm now panicking with the actual observed state instead of silently returning.
- \`manual_is_multiple_of\`: replaced \`total % n == 0\` with \`total.is_multiple_of(n)\` per the suggested fix.

Re-verified: \`cargo clippy -p engine --test integration -- -D warnings\` clean, and both tests (\`fireball_surcharge_scales_with_target_count_on_distribute_route\`, \`fireball_unpayable_recomputed_cost_rolls_back_cleanly\`) still pass.

matthewevans (approved):
**Approving — the two clippy blockers are fixed, and the `never_loop` fix makes the test strictly more discriminating than before.**

## ✅ Clean

**The lint fix is a genuine test hardening, not a lint-silencer.** `fireball_x_cost_surcharge_timing.rs:122` — the old `for _ in 0..40 { match … }` had a `_ => return` arm that **silently exited** on any unexpected `WaitingFor` state, so the test could pass without ever driving `ChooseX`. The replacement makes the catch-all fail closed:

```rust
other => panic!("expected ChooseXValue immediately after CastSpell, got {other:?}"),
```

That converts a silent pass into a hard failure, which is exactly the right direction for a test guarding a cost-timing invariant. `total % n == 0` → `total.is_multiple_of(n)` at `:185` is semantically identical.

**Delta is test-file-only** (`ahead:1, behind:0`, +13/−14, single file), so the engine review from the prior pass carries unchanged:
- Right seam — routes through `finish_pending_cast_cost_or_pay` (`casting_costs.rs:3159`), the single post-target cost authority, rather than calling `finalize_cast` with a cost locked in at `ChooseXValue` time.
- Clone-and-restore-on-`Err` rollback mirroring `finalize_mana_payment`'s `pending_for_restore` — load-bearing, because `pending_cast` is destructively taken here; without it an unpayable recomputed cost would strand `waiting_for: DistributeAmong` with `pending_cast: None`.
- Both resume call sites fixed (`engine_resolution_choices.rs:3970` NamedChoice path).
- **CR 601.2f** (per-target surcharge is part of the total cost) and **CR 601.2b–h** grep-verified; general `with_strive_cost` test builder rather than a Fireball special-case.

CI fully resolved green — 12 SUCCESS, 1 skipped, both Rust shards and the lint job that was previously red.

**Approving and enqueuing as `bug`.**

PR #5557, opened 2026-07-11
fix(engine): return cards exiled with a self-exiled source (Mechtitan Core)

JacobWoodson:
## Problem

**Mechtitan Core** never returned its exiled cards when the Mechtitan token left the battlefield. It presents as "the returned cards weren't tapped," but the cards were not returned **at all** (so they were never tapped).

Root cause, confirmed against a real game-state export:

- The activated ability exiles the 4 other permanents with `TrackedBySource` exile-links keyed to Core, and installs a delayed trigger `WhenLeavesPlayFiltered{token} → ChangeZoneAll{ Exile→Battlefield, ExiledBySource, enter_tapped: Tapped }` whose `source_id` is Core. The AST is correct — `enter_tapped` is `Tapped`.
- Core self-exiles paying its own cost. `apply_zone_exit_cleanup` (`game/zones.rs`) pruned every `TrackedBySource` link keyed to Core the instant Core left the battlefield, and the only fallback (`linked_exile_lki`) is cleared on every phase/step transition.
- So when the token left (typically turns later), `linked_exile_cards_for_source(Core)` returned `[]` and nothing came back.

Per Scryfall, the real text is "return all cards exiled with this Vehicle **except this card**" — Core does not return itself. That already falls out for free: the self-exile of "this Vehicle" never self-links Core, so `ExiledBySource(Core)` yields only the 4 others.

## Fix (`crates/engine/src/game/zones.rs`)

1. **Preserve** `TrackedBySource` links when a source leaves the battlefield **to exile** (CR 607.2a: the pile stays the linked-ability referent while the source keeps its stable ObjectId in exile per CR 400.7). Mirrors the existing `CraftMaterial` exemption; narrowed to exit-to-exile so ordinary death/bounce still prunes.
2. **Paired reset**: clear those links if that source **re-enters** the battlefield (CR 400.7 blink-identity reset).

Blast radius is small: `should_track_exiled_by_source` already gates link creation to sources whose ability consumes `ExiledBySource`, so cards like Bojuka Bog / Pit of Offerings never create these links in the first place.

## Tests

- `crates/engine/tests/integration/mechtitan_core_return_exiled.rs` — end-to-end regression: source self-exiles → a phase passes → the token leaves → all 4 exiled cards return **tapped** under their owners' control, Core stays in exile, a card exiled by a different source is untouched. Revert-failing on the exit-prune change.
- Two seam unit tests in `game/zones.rs` (links survive exit-to-exile; links reset on battlefield re-entry).
- Updated two pre-existing tests (`craft_material_link_survives_source_battlefield_exit`, `pit_of_offerings_blink_clears_exile_links`) that encoded the old prune-on-exile behavior to the new blink-on-reentry semantics.

Full engine lib suite (16132) and integration suite (2676) pass; `cargo fmt` and `clippy --tests` clean.

### DIFF
--- a/crates/engine/src/game/exile_links.rs
+++ b/crates/engine/src/game/exile_links.rs
@@ -250,14 +250,21 @@ mod tests {
     /// CR 702.167a/c: a `CraftMaterial` link must survive the craft source's
     /// battlefield exit (it self-exiles mid-activation and returns with the same
     /// ObjectId), so the returned permanent can still read what it was crafted
-    /// with. A plain `TrackedBySource` link from the same source is pruned on
-    /// that exit — the contrast that motivates the dedicated kind.
+    /// with. The contrast that motivates the dedicated kind is now a NON-exile
+    /// exit: on a death (battlefield -> graveyard) `CraftMaterial` survives but a
+    /// plain `TrackedBySource` link from the same source is pruned.
+    ///
+    /// CR 607.2a + CR 400.7: `TrackedBySource` links are NOT pruned on an exit TO
+    /// EXILE — a self-exiled source stays the linked-ability referent for its
+    /// pile (Mechtitan Core). The exile-exit arm below asserts that survival so a
+    /// regression that reinstates the old blanket prune fails here.
     #[test]
     fn craft_material_link_survives_source_battlefield_exit() {
         use crate::game::zones::{create_object, move_to_zone};
         use crate::types::game_state::{ExileLinkKind, GameState};
         use crate::types::identifiers::CardId;
 
+        // --- Non-exile exit (death): CraftMaterial survives, TrackedBySource pruned.
         let mut state = GameState::new_two_player(1);
         let source = create_object(
             &mut state,
@@ -283,9 +290,8 @@ mod tests {
         );
         push_with_kind(&mut state, tracked, source, ExileLinkKind::TrackedBySource);
 
-        // The craft source self-exiles mid-activation (battlefield -> exile).
         let mut events = Vec::new();
-        move_to_zone(&mut state, source, Zone::Exile, &mut events);
+        move_to_zone(&mut state, source, Zone::Graveyard, &mut events);
 
         assert!(
             state.exile_links.iter().any(|l| l.exiled_id == material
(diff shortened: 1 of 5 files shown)


matthewevans (approved):
**Approve.** A rules-correct fix at the right seam: the exile-link lifecycle in `zones.rs` now preserves a source's `TrackedBySource` links when it leaves the battlefield *to exile*, with a paired blink-reset on re-entry. Grep-verified CRs, discriminating registered test, small gated blast radius. Green + CLEAN.

## ✅ Clean
- **Right seam, mirrors the existing exemption.** The production change is entirely in `apply_zone_exit_cleanup` + `move_to_zone` (`game/zones.rs`), extending the same `retain` predicate that already exempts `Prison`/`Haunt`/`CraftMaterial`. Adding `source_exits_to_exile && TrackedBySource` as a preserve arm — rather than a bespoke Mechtitan path — is the building-block move.
- **CR-correct, verified verbatim against `docs/MagicCompRules.txt`:**
  - CR 607.2a — the "cards exiled with [this object]" ability refers to the exile-zone cards the first ability put there; the association does not depend on where the source now sits. Preserving the link while the source is in exile is faithful.
  - CR 400.7 — a battlefield re-entry is a new object with no memory, so the paired `to == Zone::Battlefield` prune (scoped to `TrackedBySource`, explicitly **not** `CraftMaterial`, which must survive re-entry transformed) correctly sheds the prior incarnation's pile.
  - CR 614.12 (test doc) — enter-tapped-as-replacement cite is accurate.
- **The exile-scoping is proven, not asserted.** `craft_material_link_survives_source_battlefield_exit` was correctly re-pointed at a *death* (battlefield→graveyard) to prove `TrackedBySource` still prunes on non-exile exits, and a new exile-exit arm asserts survival — so a regression that reinstates the old blanket prune fails.
- **Discriminating integration test, `mod` registered** in `tests/integration/main.rs`. `mechtitan_core_returns_exiled_pile_tapped_after_token_leaves` drives the **real** return path (delayed trigger → `ChangeZoneAll{ExiledBySource, enter_tapped: Tapped}` → resolve): step (1) asserts `linked_exile_cards_for_source(core).len() == 4` after Core self-exiles (revert the fix → 0 → fail), then the pile returns tapped under owners' control, Core stays exiled ("except this card" falls out of Core never self-linking), and an unrelated-source card is untouched.
- **Blast radius gated.** `should_track_exiled_by_source` (`exile_links.rs:32`) already restricts link creation to sources whose ability consumes `ExiledBySource`, so Bojuka Bog / Pit-style cards never create these links to begin with.

## 🟡 Non-blocking
- The integration test hand-installs the delayed trigger and pushes the `TrackedBySource` links rather than driving the 5-permanent activation cost, so it exercises the `zones.rs` lifecycle (the actual bug surface) given correctly-shaped inputs — not the parser/activation that produces those links. That's the right scope for this fix, and the **return** half runs through the real effect handler; noting it only so a follow-up that touches activation knows this test won't catch a link-*creation* regression.

Issue #5290, opened 2026-07-07
Stack sequence issue for auras/equips? — Not 100% sure if this is a bug or not; my understanding/intent with this stack…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1523880971882205324/1523880971882205324

**Thread:** Stack sequence issue for auras/equips?
**Discord thread id:** `1523880971882205324`
**Discord message id:** `1523880971882205324`
**Cards:** lynde, cheerful tormentor, curse of misfortunes
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Not 100% sure if this is a bug or not; my understanding/intent with this stack is that first [[Curse of Misfortune]] will have me search for a curse a

## Reported text

Not 100% sure if this is a bug or not; my understanding/intent with this stack is that first [[Curse of Misfortune]] will have me search for a curse and attach it to me, and then [[Lynde, Cheerful Tormentor]] will let me reattach it to someone else. But what actually happens is I confirm the stack order, then it asks for Lynde's target before I've resolved Curse of Misfortune, so there's no combo. Again not positive if this is bug territory, or I didn't fully understand stack rules.

## Attachments

- [game-state-turn-58-2026-07-07T02-25-11-529Z.zip](https://cdn.discordapp.com/attachments/1523880971882205324/1523880972058230865/game-state-turn-58-2026-07-07T02-25-11-529Z.zip?ex=6a4db840&is=6a4c66c0&hm=3a81ddff47e1ff73c4e6eae2984c5f40d535d5e4dd43c4ca921cc2b809fa7697&)
- [image.png](https://cdn.discordapp.com/attachments/1523880971882205324/1523880972351701083/image.png?ex=6a4db840&is=6a4c66c0&hm=e68fb27d6fe65a42e16ff064bef7f86c9c24a2ffc38028003c8b28f1dfb8e991&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

### lynde, cheerful tormentor
Deathtouch
Whenever a Curse is put into your graveyard from the battlefield, return it to the battlefield attached to you at the beginning of the next end step.
At the beginning of your upkeep, you may attach a Curse attached to you to one of your opponents. If you do, draw two cards.
### curse of misfortunes
Enchant player
[...]

PR #5558, opened 2026-07-11
feat(parser): recognize "Players may spend mana as though it were mana of any color"

minion1227:
CR 609.4b

## Bug

Mycosynth Lattice and Mycosynthwave read "**Players** may spend mana as though it were mana of any color" — the all-players form of the board-wide any-color concession. But the dispatcher only matched the controller form ("**You** may spend mana …") via a verbatim string `==` check, so the "Players may" subject **dropped the entire static** and no player gained the permission.

## Fix

Replace the verbatim `==` match with a nom recognizer `parse_board_wide_spend_any_color` (mirroring the existing `is_speed_unlock_sentence` helper) that accepts **both** the "You may" and "Players may" subject prefixes on the bare form. Both map to the same `SpendManaAsAnyColor` static with `affected: TargetFilter::Player`, which `check_static_ability` already scopes to *every* player — so no runtime change is needed.

- Removes a prohibited verbatim-string dispatch (`tp.lower.trim_end_matches('.') == "…"`) in favor of a combinator — the same idiom the sibling `is_speed_unlock_sentence` uses.
- `all_consuming` keeps the recognizer to the **bare** form; the qualified tails ("… to activate abilities of X", "… to cast it") are consumed by earlier arms (`try_parse_spend_any_color_to_activate_abilities`, the type_change "to cast it" grant) and never reach here — verified unchanged.

## Tests

- `static_players_may_spend_mana_as_any_color` — "Players may …" → `SpendManaAsAnyColor { .. }` with `affected: Some(TargetFilter::Player)`.
- `static_you_may_spend_mana_as_any_color_still_parses` — regression: the "You may" form still parses to the same shape.
- Full `parser::oracle_static::tests` (1047 passed, 0 failed), including the "… to activate abilities of creatures you control" activation-source-filtered case.

## CR verification

- **CR 609.4b** — spending mana as though it were another type/color — verified in `docs/MagicCompRules.txt`.

## AI-contributor notes

- **Rules-correct:** every player gains the any-color concession per CR 609.4b; the runtime already broadcasts `TargetFilter::Player` to all players.
- **Builds the class:** Mycosynth Lattice + Mycosynthwave, and any future "Players may spend mana …" board-wide static.
- **Verified:** live-parsed "Players may" / "You may" / "You may … to activate abilities of X" before/after; `cargo fmt`, CI-exact `clippy -p engine --all-targets --features proptest -D warnings`, and the full oracle_static module are green.

### DIFF
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -90,6 +90,28 @@ pub(crate) fn is_speed_unlock_sentence(lower: &str) -> bool {
     .is_ok()
 }
 
+/// CR 609.4b: Match a board-wide "[you|players] may spend mana as though it were
+/// mana of any color" static (Chromatic Orrery / Mycosynth Lattice / Mycosynthwave),
+/// with no activation-source or spell-class qualifier tail. Both subject prefixes
+/// map to the same all-players concession — the runtime scopes
+/// `TargetFilter::Player` to every player — so they share one recognizer. Replaces
+/// a verbatim-string `==` match; the tailed forms ("... to activate abilities of
+/// X", "... to cast it") are consumed by earlier arms and never reach here.
+fn parse_board_wide_spend_any_color(lower: &str) -> bool {
+    all_consuming(terminated(
+        preceded(
+            alt((
+                tag::<_, _, OracleError<'_>>("you may "),
+                tag("players may "),
+            )),
+            tag("spend mana as though it were mana of any color"),
+        ),
+        opt(tag(".")),
+    ))
+    .parse(lower)
+    .is_ok()
+}
+
 /// CR 305.2: static land-play permissions with an explicit additional-drop
 /// count greater than the ordinary +1 grant. `u8::MAX` represents "any
 /// number"; runtime summing saturates so it stays effectively unbounded when
@@ -1049,8 +1071,14 @@ pub(crate) fn parse_static_line_inner(
         return Some(def);
     }
 
-    // CR 609.4b: "You may spend mana as though it were mana of any color."
-    if tp.lower.trim_end_matches('.') == "you may spend mana as though it were mana of any color" {
+    // CR 609.4b: "[You|Players] may spend mana as though it were mana of any
+    // color." The controller form (Chromatic Orrery, "You may ...") and the
+    // all-players form (Mycosynth Lattice / Mycosynthwave, "Players may ...")
+    // both grant the board-wide any-color concession with no activation-source
+    // or spell-class qualifier tail. The runtime scopes
+    // `affected: TargetFilter::Player` to every player (`check_static_ability`),
+    // so the two phrasings share one static shape.
+    if parse_board_wide_spend_any_color(tp.lower) {
         return Some(
             StaticDefinition::new(StaticMode::SpendManaAsAnyColor {
                 spell_filter: None,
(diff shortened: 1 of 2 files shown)


matthewevans (commented):
**Verdict: clean, functional fix — anti-verbatim-string refactor that correctly delivers the all-players concession. Holding only for CI.**

## ✅ Clean
- **Right refactor.** Replaces a verbatim `== "you may spend mana as though it were mana of any color"` string match with a `parse_board_wide_spend_any_color` nom recognizer (`all_consuming(terminated(preceded(alt((tag("you may "), tag("players may "))), tag("spend mana as though it were mana of any color")), opt(tag("."))))`), mirroring the existing `is_speed_unlock_sentence` helper. This is exactly the anti-verbatim-string mandate — the subject is now one `alt()` axis, so both `Mycosynth Lattice`/`Mycosynthwave` ("Players may") and the controller form share one recognizer.
- **Functional, not cosmetic.** Verified the produced shape actually grants to every player: `affected: Some(TargetFilter::Player)` → `static_filter_matches` (`static_abilities.rs:1819`) returns `true` for all `player_id` on the `TargetFilter::Player` arm ("All players match"). So opponents genuinely gain the any-color concession, which is the #5290-class bug being fixed.
- **CR 609.4b** grep-verified verbatim ("allows a player to spend mana 'as though it were mana of any [type or color]'").
- **Discriminating tests** — `static_players_may_spend_mana_as_any_color` asserts the mode + `affected == Some(TargetFilter::Player)` (fails on the pre-fix drop), plus a `static_you_may_...still_parses` regression guard.

## 🟡 Non-blocking (pre-existing, not introduced here)
- The `return` block is unchanged, so the bare `"You may spend mana as though it were mana of any color."` form still maps to `affected: TargetFilter::Player` (all players). For a genuinely controller-scoped bare "You may" card that would over-grant to opponents (`TargetFilter::Controller` would be the correct affected filter). In practice the tailed forms (`Chromatic Orrery`'s "… to cast spells") are consumed by earlier arms and never reach here, so this is latent, not a regression. Worth a follow-up but out of scope for this PR.

**Recommendation: approve as `bug` on settled green.**

matthewevans (approved):
**Approve — clean anti-verbatim-string refactor; the board-wide any-color concession now reaches all players.**

## ✅ Clean
- **Right seam / anti-verbatim.** Replaces the `== "you may spend mana…"` verbatim string check with `parse_board_wide_spend_any_color`, a nom recognizer folding the `you may` / `players may` subject into a single `alt()` axis (mirrors `is_speed_unlock_sentence`). This is exactly the decomposition the parser mandate calls for — no card-name or full-sentence `==`.
- **Functional, not cosmetic.** `affected: TargetFilter::Player` routes through `static_filter_matches` → the "all players match" arm (`static_abilities.rs:1819`), so Mycosynth Lattice / Mycosynthwave genuinely grant *every* player the "spend mana as though any color" permission — the "Players may" subject no longer drops the entire static.
- **CR 609.4b** grep-verified for the mana-spending permission.
- Discriminating tests + a regression guard on the dropped-subject path.

## 🟡 Non-blocking (pre-existing, not a regression)
- The unchanged return still maps a bare "You may" form to all-players scope. Latent in the prior code too; out of scope here.

**Recommendation: approve and enqueue as `bug`.** CI is fully green (both Rust shards, lint, coverage-gate, card-data) on the reviewed head.

PR #5559, opened 2026-07-11
fix(parser): scope graveyard-return target filters to the graveyard

JacobWoodson:
## Problem

Dance of the Manse asks you to pick permanents **already on the battlefield** instead of cards in your graveyard. Its Oracle text is:

> Return up to X target artifact and/or non-Aura enchantment cards each with mana value X or less from your graveyard to the battlefield...

The parser produced a `ChangeZone { origin: Graveyard, destination: Battlefield }`, but the **target filter** came out as a bare `Or [ Artifact, Enchantment(non-Aura) ]` with **no zone constraint**. Target selection uses the filter, and a bare permanent filter defaults to the battlefield — hence the wrong candidates.

## Root cause

The engine has two parallel return handlers. The one in `mod.rs` (`try_parse_verb_and_target`) correctly stamps `InZone { Graveyard }` + `Owned { You }` onto the filter via `add_inferred_origin_constraints_to_target`, but that path is only used for compound-clause detection and its result is discarded here. The **primary** path in `imperative.rs` computed the origin zone but never stamped it onto the target filter.

`"return target creature card from your graveyard"` worked only because `parse_type_phrase` captured the zone inline; the disjunctive `artifact and/or non-Aura enchantment` structure interrupts that, so the zone was lost.

## Fix

- **imperative.rs** — the primary return handler now stamps the inferred `InZone`/`Owned` origin constraints onto the target filter, but only when `parse_type_phrase` didn't already capture the zone inline (avoids double-scoping already-zoned targets — the 905-card "double you-control" parse-diff regression the existing tests guard against).
- **mod.rs** — `add_inferred_origin_constraints_to_target` now leaves `TargetFilter::SelfRef` untouched: a self-return (`"return this card from your graveyard"`) already names one specific object, so wrapping it in an `And` with a zone filter is wrong.

Both annotated with CR 400.7 (an object moving between zones), verified against `docs/MagicCompRules.txt`.

## Testing

- New regression test asserts both `Or` branches carry `InZone { Graveyard }` + `Owned { You }`, with `origin: Graveyard` / `destination: Battlefield`. The `up to X` target count is preserved (`min: 0, max: X`).
- Full engine suite green: 7,914 lib parser tests + 2,675 integration tests pass; clippy clean.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -1836,6 +1836,21 @@ pub(super) fn parse_targeted_action_ast(
         };
         let is_mass = is_mass || count.is_some();
         let origin = super::infer_origin_zone(rest_lower);
+        // CR 400.7: A returned card's target filter must be scoped to its origin
+        // zone. Without this, "return target ... card from your graveyard to the
+        // battlefield" enumerates legal targets on the battlefield (the default
+        // zone for a bare permanent filter) instead of the graveyard. Mirrors the
+        // sibling `try_parse_verb_and_target` handler in `mod.rs`, which stamps
+        // the same `InZone`/`Owned` constraints (e.g. Dance of the Manse's
+        // graveyard-scoped "artifact and/or non-Aura enchantment cards"). Only
+        // stamp when `parse_type_phrase` did not already capture the origin zone
+        // itself ("... creature card from your graveyard" parses the zone + owner
+        // inline) — re-stamping an already-zoned filter would double-scope it.
+        let target = if target.extract_in_zone().is_none() {
+            super::add_inferred_origin_constraints_to_target(target, origin, rest_lower)
+        } else {
+            target
+        };
 
         // CR 400.7: Single-object battlefield destinations use ChangeZone;
         // mass destinations use ChangeZoneAll. Only pure mass-bounce
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -28383,6 +28383,14 @@ fn add_inferred_origin_constraints_to_target(
     let Some(zone) = origin else {
         return target;
     };
+    // CR 400.7: A self-reference ("return this card from your graveyard") already
+    // identifies one specific object, so an origin `InZone`/`Owned` constraint is
+    // meaningless — and `add_filter_props` would corrupt it by wrapping the bare
+    // `SelfRef` in an `And`. Only class-enumerating filters (Typed/Or/And/Not),
+    // whose candidate set defaults to the battlefield, need origin scoping.
+    if matches!(target, TargetFilter::SelfRef) {
+        return target;
+    }
     if let Some((matched_zone, _, props)) = super::oracle_target::scan_zone_phrase(lower) {
         if matched_zone == zone
             && props.iter().any(|prop| {
--- a/crates/engine/src/parser/oracle_effect/tests.rs
+++ b/crates/engine/src/parser/oracle_effect/tests.rs
@@ -43359,3 +43359,45 @@ fn grimoire_thief_counter_all_named_spells_unchanged() {
         "single conjunct must NOT become an Or: {target:?}"
     );
 }
+
+/// CR 400.7: Dance of the Manse returns cards from the *graveyard*, so its
+/// target filter must be scoped to the graveyard (and to your own cards). A
+/// bare "artifact and/or non-Aura enchantment cards" filter with no zone
+/// defaults to the battlefield, which made target selection offer battlefield
+/// permanents instead of graveyard cards. The disjunctive `Or` branches each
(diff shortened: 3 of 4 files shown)


matthewevans (commented):
**Verdict: correct, class-general graveyard-scope fix. Holding for CI + rebase (branch is BEHIND).**

## ✅ Clean
- **Right seam.** Scopes returned-card target filters to their origin zone in `parse_targeted_action_ast` (`imperative.rs`), mirroring the sibling `try_parse_verb_and_target` handler in `mod.rs` that already stamps the same `InZone`/`Owned` constraints. So "return target … card from your graveyard to the battlefield" enumerates graveyard candidates, not the battlefield default.
- **Double-scope guard.** Only stamps when `target.extract_in_zone().is_none()`, so a filter that already captured its zone inline ("… creature card from your graveyard") isn't re-scoped.
- **SelfRef guard.** `add_inferred_origin_constraints_to_target` now early-returns `TargetFilter::SelfRef` unchanged, preventing `add_filter_props` from corrupting a bare self-ref by wrapping it in `And`.
- **CR 400.7** grep-verified (returned card becomes a new object; its target filter must scope to the origin zone).
- **Discriminating test** — `dance_of_the_manse_targets_graveyard_not_battlefield` asserts each `Or` branch carries `InZone { Graveyard }` + `Owned { You }` (fails on the pre-fix bare-permanent filter).

## 🟡 Non-blocking
- The `SelfRef` early-return also changes behavior for the existing `mod.rs` caller (SelfRef previously flowed through `scan_zone_phrase`/`add_filter_props`). It's strictly more correct, but confirm no existing test relied on the old wrapping — CI will surface it.

**Recommendation: rebase onto `main`, approve as `bug` on settled green.**

JacobWoodson:
Added a second commit that also fixes the dropped **`each with mana value X or less`** bound.

The distributive-`each` linker stripper in `parse_type_phrase` required a comma (`"creatures, each with …"`), so Dance's comma-less `"cards each with …"` form left the suffix unparsed and the `Cmc` bound never attached. Relaxed the stripper to accept the comma-less form, gated by a `peek("with ")` so a non-distributive `each` (e.g. `"each player owns"`) is left intact. In the disjunctive `artifact and/or non-Aura enchantment` form the bound distributes onto every `Or` leg.

Side benefit: with `each with …` now consumed, `parse_type_phrase` also consumes the trailing `from your graveyard` inline, so ownership lands in the canonical `controller` field — the same shape as `"return target creature card from your graveyard"`.

CR 202.3. Adds a comma-less distributive-`each` parser test and extends the Dance effect regression test to assert the mana-value bound on each `Or` leg. Full engine suite (7,915 parser + 2,675 integration) green, clippy clean.

matthewevans (commented):
**Re-reviewed the new head — the added commit generalizes the distributive-"each" linker cleanly; holding for CI (pending).**

Note: the head advanced past my prior review (`587b0bec` → `7f06198a`) with one **new** author commit ("honor comma-less distributive 'each with' mana-value suffix"), so this is a fresh review of that delta, not an inherited approval.

## ✅ Clean
- **Right generalization in `oracle_target.rs`.** The distributive-"each" linker in `parse_type_phrase_with_ctx` now accepts the comma-less form ("cards each with mana value X or less", Dance of the Manse) as well as the comma form, via `opt((tag(","), opt(space1)))` + `tag("each ")` + `peek(tag("with "))`. All suffix parsers (P/T CR 208, mana-value CR 202.3, counters, keywords) receive the same normalized "with …" input regardless of linker — one axis, not a permutation blow-up.
- **`peek("with ")` is the load-bearing guard** — it restricts stripping to a genuine property suffix so a non-distributive "each" (e.g. "each player owns") is left intact. Good defensive combinator use.
- **CR 202.3 (mana value) / CR 400.7 grep-verified**; the test now also tolerates ownership expressed as the canonical `controller` field OR a fallback `Owned{You}` prop, and asserts the `Cmc { LE, Variable("X") }` bound distributes onto every `Or` leg.

## 🟡 Non-blocking
- The `peek("with ")` guard changes the comma-present-but-next-token-≠-"with" axis versus the old unconditional `, each ` strip. That's intended and safer, and the coverage-gate will catch any regression — worth a glance that no existing "…, each <non-with>…" card relied on the old strip.

**Recommendation: approve as `bug` once CI settles green** (lint + both shards + card-data currently pending on the rebased head).

matthewevans (approved):
**Approve.** Held positive review stands on the current head (`7f06198a` == last-reviewed commit, no advance), CI fully resolved-green, mergeStateStatus CLEAN.

## ✅ Clean
- `parse_targeted_action_ast` scopes returned-card filters to the origin zone, mirroring the sibling `try_parse_verb_and_target` — right seam, one pattern for the class.
- Distributive-"each" linker generalized to the comma-less form (`opt(comma) + "each " + peek("with ")`); the `peek` guard is load-bearing and correct.
- CR 202.3 / CR 400.7 grep-verified; discriminating Dance of the Manse test asserts each `Or` branch resolves to `Graveyard` + `Owned{You}`.

**Enqueuing as `bug`.**

PR #5568, opened 2026-07-11
feat(parser): mid-chain 'An opponent separates those cards into two piles' recognizer

Whovencroft:
## Summary

Implements mid-chain **"An opponent separates those cards into two piles"** recognition in the effect-chain parser. This closes the gap for cards like **Boneyard Parley**, **Brilliant Ultimatum**, and **Everythingamajig** (variant) where cards are exiled earlier in the resolution chain and then an opponent separates them into piles with per-pile disposition effects.

## Changes

| File | Description |
|------|-------------|
| `types/ability.rs` | New `PileSource::ExiledThisWay` variant — signals that the pile-separation operates on cards exiled earlier in the same resolution chain |
| `parser/oracle_separate_piles.rs` | `try_parse_mid_chain_opponent_separates()` — entry point called from the chunk loop; `parse_exiled_pile_disposition()` — parses the two-sentence "put the chosen pile … and the rest …" disposition pattern |
| `parser/oracle_effect/mod.rs` | Chunk-loop recognizer: detects `"an opponent separates"` / `"target opponent separates"` prefix, joins remaining chunks, delegates to the mid-chain parser, and emits a `SeparateIntoPiles { pile_source: ExiledThisWay }` clause |
| `game/effects/separate_piles.rs` | `resolve_exiled_this_way()` — collects linked-exile cards for the source and parks on `WaitingFor::SeparatePilesPartition`; controller-rebinding fix in `apply_pile_effect` / `apply_unchosen_pile_effect` to use the source controller (spell caster) rather than `result.subject` (the opponent/partitioner) |
| `game/exile_links.rs` | Registers `"ExiledThisWay"` in `LINKED_EXILE_CONSUMER_TAGS` so the exile-link tracking system recognizes the new pile source as a consumer |

## Design decisions

- **PileSource::ExiledThisWay** reuses the existing `linked_exile_cards_for_source` infrastructure rather than inventing a new tracking mechanism. Cards exiled by the same source earlier in the chain are already tracked; we just query them at pile-separation time.
- **Controller rebinding fix**: `apply_pile_effect` previously used `result.subject` (the partitioner/opponent) as the controller for sub-effects. For Boneyard Parley, `ControllerRef::You` on `enters_under` must resolve to the spell's controller, not the opponent. Fixed by passing `source_controller` through and using it as the fallback when `result.subject` is the opponent.
- **Chunk-loop integration**: The recognizer joins the current chunk with all remaining chunks (the disposition sentences follow the separation sentence) and delegates to the dedicated parser. On success it emits the clause and breaks the chunk loop.

## Testing

- 2 unit tests in `oracle_separate_piles.rs`:
  - `test_mid_chain_boneyard_parley` — full Boneyard Parley disposition pattern
  - `test_mid_chain_battlefield_and_graveyard` — battlefield + graveyard variant
- All existing tests pass (CI green: Rust tests shard 1/2 and 2/2)


matthewevans (changes requested):
**Requesting changes — the controller rebinding is unscoped and regresses Make an Example, and the green CI is a false green that cannot see it.**

The `ExiledThisWay` design is right, and Boneyard Parley's premise checks out verbatim on Scryfall: *"Exile up to five target creature cards from graveyards. An opponent separates those cards into two piles. Put all cards from the pile of your choice onto the battlefield **under your control** and the rest into their owners' graveyards."* Reusing `linked_exile_cards_for_source` instead of inventing new tracking is the right call. But two blockers stand.

## 🔴 Blocker 1 — unscoped controller rebinding breaks `PileSource::Battlefield` (Make an Example)

`separate_piles.rs:369` and `separate_piles.rs:415` swap the sub-effect controller from `result.subject` to `source_controller` for **every** `PileSource` variant, not just the new one:

```rust
-let mut chain = sub_effect_as_resolved(chosen_pile_effect, source_id, result.subject);
+let mut chain = sub_effect_as_resolved(chosen_pile_effect, source_id, source_controller);
```

That binding is load-bearing for the pre-existing `Battlefield` shape, and the parser already documents the dependency it relies on — `oracle_separate_piles.rs:73-75`:

> *"For Make an Example this yields a `Sacrifice { target: ParentTarget }` chain — **the runtime resolver re-binds `controller` to each subject before applying it.**"*

`PileResult.subject` is defined as *"The player whose objects were partitioned"* (`game_state.rs:2914`) — for Make an Example that is **each opponent**, not the caster. Make an Example (Scryfall, verbatim):

> *"Each opponent separates the creatures they control into two piles. For each opponent, you choose one of their piles. **Each opponent sacrifices** the creatures in their chosen pile."*

**CR 701.21a** (verbatim): *"To sacrifice a permanent, its controller moves it from the battlefield directly to its owner's graveyard. **A player can't sacrifice something that isn't a permanent, or something that's a permanent they don't control.**"* After this change the Sacrifice sub-effect resolves with the **caster** as chain controller, so it is asking a player to sacrifice permanents they don't control.

**Fix:** key the controller on the `PileSource` rather than rebinding globally. `Battlefield` must keep `result.subject` (per CR 701.21a the permanent's controller performs the sacrifice); `ExiledThisWay` wants the source controller (per **CR 110.2a**: *"If an effect instructs a player to put an object onto the battlefield, that object enters the battlefield under that player's control unless the effect states otherwise"* — which is exactly what Boneyard Parley's *"under your control"* needs). `RevealedFromLibraryTop` is already equivalent, since the cards come from the controller's own library.

## 🔴 Blocker 2 — `runtime-test-gap`: the new engine path has zero runtime coverage

The two added `#[test]` fns are **parser unit tests** inside `oracle_separate_piles.rs`. No file under `crates/engine/tests/integration/` is touched, so `resolve_exiled_this_way`, the `WaitingFor::SeparatePilesPartition` parking, and the controller change all ship untested at runtime.

This is why CI green proves nothing here: the only two runtime pile tests that exist — `fact_or_fiction_pile_separation.rs` and `sphinx_of_uthuun_etb_pile_separation.rs` — are both `RevealedFromLibraryTop`, where the partitioned cards come from your own library, so `subject == controller` and the rebinding is a **no-op**. There is no runtime test for Make an Example anywhere in `crates/engine/tests/`. The one variant the change actually alters is the one variant nothing guards.

**Fix:** add a registered integration test under `tests/integration/` (with its `mod` line in `main.rs` — an unregistered file compiles to nothing) driving Boneyard Parley end-to-end: exile → opponent partitions → chosen pile enters **under your control** → rest to owners' graveyards. Add a Make an Example runtime test asserting the opponent (not the caster) sacrifices — that is the assertion that would have caught Blocker 1.

## 🟡 Non-blocking

- **Brilliant Ultimatum is claimed but not covered.** Its disposition is *"You may play lands and cast spells from one of those piles"* — not the *"put the chosen pile … and the rest …"* shape `parse_exiled_pile_disposition()` parses. Either it declines gracefully (then drop it from the summary) or it misparses (then it needs a decline guard). Worth confirming which.
- **`unwrap_or(PlayerId(0))`** (`separate_piles.rs:117`, `:150`) silently defaults to a specific player when the source is gone and `WaitingFor` doesn't match. A wrong-player default is harder to debug than an explicit `EffectError`.

## ✅ Clean

- `PileSource::ExiledThisWay` reuses the existing `linked_exile_cards_for_source` infrastructure rather than inventing parallel tracking, and registers in `LINKED_EXILE_CONSUMER_TAGS` so exile-link retention is honored.
- Parser is phrase-general (`"an opponent separates"` / `"target opponent separates"` via combinators), not card-name-keyed.
- The multi-opponent partitioner branch (`candidates.len() >= 2` → `SeparatePilesChooseOpponent`) correctly handles the multiplayer "an opponent" choice rather than assuming two players.

**Recommendation: scope the controller binding to the `PileSource` (keep `result.subject` for `Battlefield`), and add the two registered runtime tests. The parser and the `ExiledThisWay` design are sound — the blocker is entirely in the unscoped rebinding of shared code.**

matthewevans (changes requested):
**Changes requested — one blocker remains, and I'm withdrawing the other: my previous controller-rebinding blocker was wrong, and your justification comment is correct.**

## ✅ Withdrawn: the controller rebinding is safe (my prior blocker was incorrect)

I blocked the last head claiming that rebinding the sub-effect controller from `result.subject` to `source_controller` in `apply_pile_effect` would regress **Make an Example** (*"Each opponent sacrifices the creatures in their chosen pile"*) under CR 701.21a. I traced it properly this time, and that was wrong. Correcting it on the record:

- Make an Example's chosen-pile sub-effect is `Effect::Sacrifice { target: TargetFilter::ParentTarget, .. }` (`separate_piles.rs:361-370`).
- The CR 701.21a guard at `sacrifice.rs:368-375` skips a sacrifice when `obj.controller != ability.controller` **only if the filter is not** `ParentTarget`/`ParentTargetSlot`. `ParentTarget` is explicitly exempted.
- On that exempted path, `sacrifice.rs:377` sets `let player_id = obj.controller;` — the acting/sacrificing player is the object's **own controller**, never `ability.controller`. So the opponent still sacrifices their own creatures, and `PermanentSacrificed { player_id }` still records the opponent.

The pre-existing `discriminator_make_an_example_sacrifices_chosen_pile` test drives that path end-to-end and asserts the creatures land in the **opponent's** graveyard — and it is green on this head. Your comment at `separate_piles.rs:338+` is right: **CR 110.2a** (*"If an effect instructs a player to put an object onto the battlefield, that object enters the battlefield under that player's control unless the effect states otherwise"*) means Boneyard Parley's *"onto the battlefield **under your control**"* genuinely requires the caster as the sub-effect controller. The rebind is necessary and correct. Apologies for the bad call.

## 🔴 Blocker: the entire behavioral surface of this change has zero runtime coverage

This is the one that stands, and the withdrawal above sharpens it rather than softening it.

The rebind is a **no-op** for `RevealedFromLibraryTop` (there `result.subject == source controller` — you reveal from your own library) and **bypassed** for `Battlefield` (via the `ParentTarget` exemption above). So the *only* place the new controller semantics are observable is the new `ExiledThisWay` path — and that path has no runtime test at all.

Concretely, nothing exercises:
- `resolve_exiled_this_way` (`separate_piles.rs:262+`) — the `linked_exile_cards_for_source` eligible-set derivation, the empty-eligible early return, or the `SeparatePilesChooseOpponent` vs `SeparatePilesPartition` parking split.
- The `source_controller` rebind's actual purpose: **that the chosen pile enters the battlefield under the caster's control.** You changed the controller semantics specifically to satisfy CR 110.2a and there is no assertion anywhere that the outcome holds. If this resolved to the opponent, every existing test would still pass.
- The exile-link plumbing (`exile_links.rs:+3`).

The two tests added (`parses_boneyard_parley_mid_chain`, `mid_chain_rejects_non_matching`) are **parser** tests in `oracle_separate_piles.rs` — they assert the AST shape (`PileSource::ExiledThisWay`, chooser, disposition), not resolution. No file under `tests/integration/` is touched. The green CI is therefore structurally uninformative for this PR: the two existing runtime pile tests (`fact_or_fiction_pile_separation`, `sphinx_of_uthuun_etb_pile_separation`, both registered in `tests/integration/main.rs`) are both `RevealedFromLibraryTop`, so they never enter the new arm.

**Fix:** add one registered integration test under `crates/engine/tests/integration/` (plus its `mod` line in `main.rs` — without it the file compiles to nothing and is an inert false-green) driving Boneyard Parley end-to-end: exile N creature cards from graveyards → opponent partitions → caster chooses a pile → assert the chosen pile is **on the battlefield under the caster's control** and the rest are in their **owners'** graveyards. That single test is what makes the rebind (and CR 110.2a) load-bearing instead of asserted-by-comment.

## 🟡 Non-blocking

- `apply_pile_effect` / `apply_unchosen_pile_effect`: the `unwrap_or(PlayerId(0))` terminal fallback silently defaults to seat 0 on a lookup miss. Prefer returning `EffectError` — a wrong-player default is a silent rules violation rather than a loud failure.
- Out of scope, pre-existing on `main`: the `// CR 701.17a` annotations throughout `sacrifice.rs` (lines 200, 271, 288, 305, 351, 356, 361) are wrong — **701.17 is Mill**; Sacrifice is **701.21**. Worth a separate cleanup PR; please don't fold it into this one.

## ✅ Clean

- `PileSource::ExiledThisWay` is a genuine new axis on an existing enum, not a card special-case, and `resolve()` dispatches exhaustively.
- Deriving the eligible set from `exile_links` keyed on `source_id` is the right seam — it reuses the existing provenance primitive rather than threading a bespoke channel through the chain.
- Boneyard Parley's Oracle text verified verbatim against Scryfall; the parsed disposition (battlefield-under-your-control / rest-to-owners'-graveyards) matches.

**Recommendation: add the one Boneyard Parley integration test (registered in `main.rs`), and this is an approve.** The design and the controller reasoning are sound — the change just needs the runtime assertion that proves it.

Issue #3674, opened 2026-06-17
Mosswort Bridge — I get to exile a card to hideaway, but I can't cast the exiled card when I activate its "You may play…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1516835328827330641/1516835328827330641

**Thread:** Mosswort Bridge
**Cards:** _none detected_
**Parser status:** has_gaps
**Extraction confidence:** 0.90

## Summary
I get to exile a card to hideaway, but I can't cast the exiled card when I activate its "You may play the exiled card" ability.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1516835328827330641:1516835328827330641:0` · discord: `1516835328827330641/1516835328827330641`</sub>

mike-theDude:
Verified against Mosswort Bridge's Oracle text: "Hideaway 4 ... {G}, {T}: You may play the exiled card without paying its mana cost if creatures you control have total power 10 or greater." Both abilities parse without any `Unimplemented`/`Unknown` markers, but the second (play) ability is mis-targeted. The generic anaphoric-forms branch in `parser/oracle_effect/mod.rs` (~line 19165) maps "the exiled card" to `TargetFilter::ParentTarget` unconditionally — correct for a "the exiled card" reference within the *same* resolving ability chain, but wrong here: this is a standalone, later-activated ability with no chain parent, so `ability.targets` is empty and `ParentTarget` resolves to nothing. `Effect::CastFromZone`'s resolver (`game/effects/cast_from_zone.rs`) only falls back to the persistent exile-link pool (`TargetFilter::ExiledBySource`, the mechanism `game/effects/hideaway.rs` explicitly documents as the correct target for this exact companion-ability pattern) when `target_filter.references_exiled_by_source()` is true — which `ParentTarget` never is — so the cast silently resolves as a no-op ("nothing to cast"). This is not a Mosswort-specific bug: the same `CastFromZone{target: ParentTarget}` shape shows up identically for every hideaway land using this template (confirmed Spinerock Knoll, Howltooth Hollow also affected), so the whole hideaway-land cycle's payoff ability is currently non-functional.

*Status: confirmed — classifier: supported-aspect-defect. priority: p1-core-mechanic*

mike-theDude:
[...]

Issue #3246, opened 2026-06-14
Windbrisk Heights — Allows you to hideaway a card, but will not allow you to cast it even if you have met the requireme…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1515679565953896599/1515679565953896599

**Thread:** Windbrisk Heights
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Allows you to hideaway a card, but will not allow you to cast it even if you have met the requirements.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1515679565953896599:1515679565953896599:0` · discord: `1515679565953896599/1515679565953896599`</sub>

mike-theDude:
Card identified as Windbrisk Heights. Verified Oracle text: "Hideaway 4 ... {W}, {T}: You may play the exiled card without paying its mana cost if you attacked with three or more creatures this turn."

The `{W}, {T}` activated ability parses with the correct cost, the correct "attacked with three or more creatures" condition, and `without_paying_mana_cost: true` — but its `CastFromZone` target is `TargetFilter::ParentTarget`, which is the ephemeral within-resolution-chain reference used for anaphora inside a single ability's sub-ability chain. Windbrisk Heights' activation is a *standalone* activated ability with no runtime chain linkage back to the Hideaway ETB trigger that originally exiled the card (which may have resolved many turns earlier), so `ParentTarget` can never resolve to the hidden-away card and the ability effectively has no valid target when activated — matching the report exactly ("will not allow you to cast it even if you have met the requirements").

The engine already has the correct persistent mechanism for this: `TargetFilter::ExiledBySource`, backed by `state.exile_links` (`crates/engine/src/game/exile_links.rs`), used elsewhere for cards that reference something exiled by a source across time. I confirmed this is a class-wide bug, not specific to this card — Mosswort Bridge and Shelldock Isle's identical follow-up abilities also parse to `ParentTarget` instead of `ExiledBySource`. Fix belongs in the Oracle parser's "play the exiled card" resolution for Hideaway-style activated abilities (`crates/engine/src/parser`), swapping `ParentTarget` for `ExiledBySource` when the reference is on a standalone activated ability rather than a same-chain sub-ability.

*Status: confirmed — classifier: supported-aspect-defect*

mike-theDude:
[...]

PR #5571, opened 2026-07-11
fix(parser): bind "the exiled card" anaphor to ExiledBySource for cross-resolution abilities (#4792, #3246, #3674)

### LINKED ISSUE #4792: Isochron Scepter does not allow casting the imprinted spell
Reported in Discord: https://discord.com/channels/1485498006781427802/1520836689432674506/1520836689432674506

**Thread:** isochron scepter
**Cards:** isochron scepter

## Summary
Reporter says Isochron Scepter does not let them cast the imprinted spell.

### isochron scepter
Imprint — When this artifact enters, you may exile an instant card with mana value 2 or less from your hand.
{2}, {T}: You may copy the exiled card. If you do, you may cast the copy without paying its mana cost.

---
<sub>report_id: `discord:1520836689432674506:1520836689432674506:0` · discord: `1520836689432674506/1520836689432674506`</sub>

mike-theDude:
## Problem

A standalone activated (or later-resolving) ability that references an exile made by an **earlier** ability — "copy the exiled card" / "play the exiled card" — was bound to an ephemeral chain-local `TrackedSet(0)` sentinel (or `ParentTarget`). By the time the ability is activated, that sentinel resolves to whatever tracked set was most recently published *anywhere* in the game, so the ability copied the wrong card or silently no-opped. This broke Isochron Scepter's entire copy class and the Hideaway lands.

Fixes #4792, #3246, #3674.

## Fix (building block, not a card patch)

- A single shared parser helper `resolve_singular_exiled_card_target` is the sole authority for the same-chain-vs-cross-resolution decision, called from both the copy path (`imperative.rs`) and the cast path (`oracle_effect/mod.rs`).
- It is gated on a new `ParseContext::chain_has_prior_exile_producer` signal, computed from an **exhaustive** `chain_clause_is_exile_producer` predicate over every exile-producing effect shape (`ChangeZone`/`ChangeZoneAll` to Exile, `ExileTop`, `Dig`-to-Exile, `ExileFromTopUntil`, `RevealUntil` kept-in-exile).
- When the referencing ability's own effect chain contains **no** exile producer, "the exiled card" binds to the durable `TargetFilter::ExiledBySource` link (CR 406.6 / CR 607.2a linked abilities) instead of the chain-local sentinel.
- Adds the missing `ExiledBySource` runtime resolver to `copy_spell.rs`, mirroring the pre-existing `cast_from_zone.rs` linked-exile path. The Hideaway cast path was already runtime-ready (its ETB unconditionally records the link).

No new `TargetFilter` variant/field — `ExiledBySource` already exists and is handled in every exhaustive match.

## Cards fixed (~19)

Copy class: **Isochron Scepter, Elite Arcanist, Myra the Magnificent, Spellbinder, Surge to Victory**. Full **Hideaway** family: Windbrisk Heights, Mosswort Bridge, Shelldock Isle, Spinerock Knoll, Howltooth Hollow, and the rest.

Same-chain "impulse then cast" cards (Keldon Flamesage, Discover the Impossible, Creative Technique, Ryan Sinclair) are **provably unchanged** — the gate keys on the presence of an in-chain exile producer, which those cards have.

## Deliberately out of scope (separate follow-ups)

- **Saruman of Many Colors** — a *distinct* runtime root cause: a plain targeted `ChangeZone{Exile}` never publishes a tracked set, so its `CopySpell{TrackedSet(0)}` reads a stale global set. Fixing it safely needs its own change + per-card runtime tests (also touches Evolving Door / Beseech the Mirror). Documented in an in-code `KNOWN FOLLOW-UPS` comment.
- **Watcher for Tomorrow** — the `ChangeZoneAll` "put the exiled card" emission site (a third call site the shared helper could later serve).

## Verification

- **Discriminating cross-turn tests**, both mutation-checked (fail with the fix reverted): copy path (`issue_4792_isochron_scepter`) and cast path (`issue_3246_windbrisk_heights_hideaway_exiled_by_source`) — each drives the real ETB trigger (no hand-seeded state) with a decoy clobbering `TrackedSet(0)`.
- **Parser anti-regression tests**: Keldon Flamesage stays `TrackedSet`, Discover the Impossible stays `ParentTarget`.
- **Zero unintended card-data change** across the affected corpus.
- Full engine test suite green; `clippy -D warnings`, `cargo fmt`, parser-combinator gate, and CR-annotation gate all clean.
- CR citations (406.6, 607.1, 607.2a, 608.2c, 702.75a, 707.10, 601.2a) grep-verified against `docs/MagicCompRules.txt`.


matthewevans (changes requested):
**Request changes — the branch does not compile against current `main`. The design premise is CR-correct and verified; the blocker is a *semantic* stale-base conflict, and porting it is not a mechanical rename.**

## 🔴 Blocker — three compile errors, one root cause: this is built against the pre-U5 `ClauseIr` API

Every Rust job is red, including **WASM compile check** — the tell that this is a build failure, not a test failure. All three errors are on PR-added lines:

| | |
|---|---|
| `parser/oracle_effect/mod.rs:21381` | `E0609: no field `followup_continuation` on `ClauseIr`` |
| `parser/oracle_effect/mod.rs:21387` | `E0609: no field `intrinsic_continuation` on `ClauseIr`` |
| `parser/oracle_effect/mod.rs:25920` | `E0425: cannot find value `clauses` in this scope` |

`main`'s U5 refactor folded those flat fields into a `ClauseDisposition` enum on `ClauseIr` (`parser/oracle_ir/effect_chain.rs`), which documents the change in its own words — "*formerly `followup_continuation`*", "*formerly `intrinsic_continuation`*":

```rust
pub(crate) enum ClauseDisposition {
    Emit { followup: Option<ContinuationAst>, intrinsic: Option<ContinuationAst> },
    Continue { continuation: Option<ContinuationAst> },   // folds absorbed_by_followup + followup_continuation
    Absorb { .. },
}
```

**⚠️ `gh pr update-branch` will not fix this.** The PR is `BEHIND`, not `DIRTY` — git sees no *textual* conflict, so the branch update merges cleanly and *then* fails to compile. The type this code consumes was refactored underneath it.

## 🔴 The port is not a mechanical rename — read the continuation channel from **both** arms

This is the part worth slowing down on, because your own doc comment predicts the trap.

`clause_ir_is_exile_producer` (`mod.rs`) deliberately does **not** filter on `absorbed_by_followup`, and you explain exactly why:

> *A Hideaway-shaped "exile one of them face down" chunk … is also marked `absorbed_by_followup: true` … an `!absorbed_by_followup` filter here would incorrectly discard this clause's still-genuine exile evidence (Discover the Impossible would otherwise wrongly flip to `ExiledBySource`).*

Under the new API, **"absorbed" *is* the `Continue` arm.** So the predicate must check the continuation channel on **both** `Emit { followup, intrinsic }` **and** `Continue { continuation }`. If it is ported to read only the `Emit` arm, it silently drops precisely the exile evidence your comment was written to preserve — reintroducing the Discover the Impossible / Hideaway regression you documented. The `E0425` at `mod.rs:25920` (`clauses` not in scope) needs the same treatment at the `parse_effect_chain_ir` call site.

## ✅ Clean — verified, and survives the rebase

- **CR citations grep-verified verbatim** against `docs/MagicCompRules.txt`, and they are precisely on point. CR 607.2a: *"If an object has an activated or triggered ability printed on it that instructs a player to exile one or more cards and an ability printed on it that refers either to 'the exiled cards' … these abilities are linked. The second ability refers only to cards in the exile zone that were put there as a result of an instruction to exile them in the first ability."* That is Isochron Scepter and Hideaway exactly. CR 406.6 likewise. Binding the anaphor to the durable `ExiledBySource` link instead of an ephemeral chain-local sentinel is the rules-correct model, not a workaround.
- **No new `TargetFilter` variant** — reuses the pre-existing `ExiledBySource`, already handled in every exhaustive match. Parameterize-don't-proliferate respected.
- **Building block, not a card patch** — one shared `resolve_singular_exiled_card_target` authority serving both the copy (`imperative.rs`) and cast (`oracle_effect/mod.rs`) paths, gated on an exhaustive `chain_clause_is_exile_producer` predicate rather than card names.
- **Both tests are properly registered — no inert test.** I checked this specifically: `issue_4792_isochron_scepter.rs` modifies a file already registered at `tests/integration/main.rs:427`, and the genuinely-new `issue_3246_windbrisk_heights_hideaway_exiled_by_source.rs` adds its own `mod` line. Both live under `tests/integration/` (top-level-binary guard respected).
- Scoping out Saruman of Many Colors as a **distinct runtime root cause** (a plain targeted `ChangeZone{Exile}` never publishes a tracked set) is the right call, and documenting it in-code rather than half-fixing it is good discipline.

## Recommendation

**Rebase onto `main` and port `clause_ir_is_exile_producer` + its `parse_effect_chain_ir` call site to `ClauseDisposition`** — reading the continuation channel from **both** the `Emit { followup, intrinsic }` and `Continue { continuation }` arms, so the absorbed-Hideaway evidence your comment protects is preserved. I'm holding the deep architecture review of the ~19-card claim until it compiles, since this predicate has to change materially anyway. The design, the CR grounding, and the test discipline all look right — this is a build-integration blocker, not a design one.

mike-theDude:
Filed the two deliberately-deferred follow-ups referenced above:
- #5576 — Saruman of Many Colors (same-chain targeted-exile runtime root cause; also implicates Evolving Door / Beseech the Mirror)
- #5577 — Watcher for Tomorrow (Hideaway LTB `ChangeZoneAll` "put the exiled card" — third emission site for the shared helper)

Both reuse the `ExiledBySource` / `exile_links` building block this PR introduces.

matthewevans:
Hi @mike-theDude — heads up that this went `CONFLICTING`, and **it was our fault, not yours**. Your branch was fine when you opened it; our own in-flight parser refactor campaign landed underneath you and invalidated it. Per our maintainer policy, when *we* cause the staleness, *we* do the rebase — so I've rebased and ported it for you. Nothing is required from you, and your commit keeps its authorship.

Three of our changes broke the branch:

1. **U5-M2** deleted the `SpecialClause` enum and folded `ClauseIr`'s `absorbed_by_followup` / `followup_continuation` / `intrinsic_continuation` fields into a single typed `ClauseDisposition` (`E0609`).
2. **U6** moved chain-local clause accumulation into the assembly arena behind `ClauseIrBuilder::clauses()` (`E0425`).
3. **a61e7fdd91** (landed this morning) folded `Effect::ForEachCategoryExile` into `Effect::ForEachCategory { action: ForEachCategoryAction, .. }` (`E0599`).

What I changed, all in `oracle_effect/mod.rs`:

- `clause_ir_is_exile_producer` now reads the continuation channels via `ClauseDisposition::followup()` / `::intrinsic()`, which U5 documents as the direct successors of the two deleted fields. Your call site now reads `builder.clauses()`.
- Your deliberate decision **not** to skip absorbed clauses is preserved, and I verified it's load-bearing rather than taking it on trust: adding a `!matches!(.., Continue { .. })` filter to the scan turns your `discover_the_impossible_cast_target_stays_parent_target_after_exiled_by_source_fix` test red with `got Some(ExiledBySource)` — exactly the wrong-binding flip you were guarding against. Your test caught the mutation. Nice.

One **behavior fix** worth flagging, since it's a change to your logic rather than a mechanical port. Your `ForEachCategoryExile { zone: Zone::Exile, .. }` arm was testing the wrong field: `zone` names the **source pool** the cards are exiled *from* (typically `Library`), not the destination — exiling is intrinsic to the variant. So that arm would never have matched an ordinary library pool. The rename meant I had to re-express it anyway, so rather than carry the confusion across I made it test the action: `ForEachCategory { action: ExileFromPool { .. }, .. }`. Easy field to misread, and it doesn't affect any of the cards your PR targets.

The `integration_cards.json` conflict was a clean union — you regenerated 5 cards (isochron scepter, mosswort bridge, shelldock isle, spellbinder, windbrisk heights) and main regenerated 1 (`the invincible iron man`), with no overlap, so both sides are kept.

Verified in a clean tree: `cargo fmt`, `cargo check -p engine --tests`, `cargo test -p engine --lib` (16180 passed / 0 failed), your three #4792/#3246 integration tests (3 passed / 0 failed), and the parser combinator gate. The PR is now `MERGEABLE` (just `behind` main, which the merge queue resolves).

Thanks for this one — the `ExiledBySource` linked-ability binding is a genuinely nice fix, and the cross-turn "unrelated tracked-set activity" tests are exactly the right way to pin it.

matthewevans (approved):
Approving. The blockers on this PR were **our** doing, not yours — main's U5/U6 parser refactor (`ClauseIrBuilder`/`ClauseDisposition`, and the deletion of the `SpecialClause` enum) invalidated the branch underneath you, so we rebased and ported it rather than asking you to chase a moving target. My earlier `CHANGES_REQUESTED` is superseded by this review.

**Verified before approving:**

- **CI is real and green** — 12 checks, 11 pass, 0 fail, `CLEAN/MERGEABLE`.
- **The parse-diff sticky matches the intent exactly**: 22 cards, 3 signatures, and *every* change is `target: parent target` / `tracked set #0` → `cards exiled by source`. Nothing moved in the opposite direction, and the same-chain anti-regression cards you wrote tests for (Keldon Flamesage, Discover the Impossible) do **not** appear in the diff — they stay same-chain, which is exactly right.
- **The `ForEachCategory` arm in `chain_clause_is_exile_producer` is correct.** Worth calling out, because it fixes a latent bug in the original: the guard tested `ForEachCategoryExile { zone: Zone::Exile, .. }`, but `zone` is documented as *"the zone the pool cards live in (where each chosen card is exiled **from**)"* — the **source pool**, typically `Library`, not the destination. That predicate therefore never matched an ordinary pool and was effectively dead code. The port re-expresses it as `action: ExileFromPool { .. }`, where exiling is intrinsic to the action regardless of source zone. Broadening the predicate is also the conservative direction: more clauses count as same-chain producers, so *fewer* anaphors flip to `ExiledBySource` — it cannot wrongly promote a same-chain binding.

For the record, the pool-level count is **22 cards**, a little under the ~19 the description claims but by way of *more* coverage, not less — it picks up several cards the PR never enumerated (Collector's Cage, Allure of the Unknown, Fateful Tempest, and the modern Hideaway family).

Thanks for the patience while this sat conflicted — that was on us.

PR #5579, opened 2026-07-11
fix(engine): unify TrackedSetFiltered's TrackedSetId(0) sentinel onto resolve_tracked_set_id (#5512)

### LINKED ISSUE #5512: TrackedSetFiltered's TrackedSetId(0) sentinel ladder diverges: no chain rung, and empty sets shadow populated ones
## Summary

`TargetFilter::TrackedSetFiltered`'s sentinel resolution in `crates/engine/src/game/filter.rs` uses a different ladder from every other consumer of the `TrackedSetId(0)` sentinel. It diverges in two ways, both of which can select the wrong tracked set.

This is a deliberate follow-up carved out of #5505's review. That PR fixed the sibling `TargetFilter::TrackedSet` arm and introduced `targeting::resolve_tracked_set_id` as the single id-resolution authority, but **intentionally left `TrackedSetFiltered` alone** because changing it affects the Zimone's Experiment / Living Death class and did not belong in a Druid of Purification bug fix.

## The divergence

Once #5505 lands there are three ladders:

| location | ladder |
|---|---|
| `targeting.rs` `resolve_tracked_set_id` (new in #5505) | chain → latest **non-empty** |
| `targeting.rs` `resolve_tracked_set_sentinel` | chain → `current_combat_damage_source_filter` (CR 510.2) → latest **non-empty** |
| `filter.rs` `TrackedSetFiltered` arm | **no chain rung** → `max_by_key` over **all** sets, **including empty ones** |

The middle one is legitimately separate: its combat-damage rung yields a `TargetFilter` rather than a `TrackedSetId`, so it cannot fold into the id-level helper. That is documented in #5505.

`TrackedSetFiltered` has no such excuse. Two concrete failure modes:

1. **No chain rung.** It ignores `state.chain_tracked_set_id` entirely, so a set published by the active resolution chain is not preferred. Every other consumer prefers it.
2. **Empty-set shadowing.** `max_by_key(|(id, _)| id.0)` runs over all entries with no `!objects.is_empty()` filter, so a trailing **empty** set with a higher id shadows an earlier populated one and the filter matches nothing. `latest_tracked_set_id` (`targeting.rs:2133-2140`) explicitly skips empties.

## Why it hasn't bitten

`TrackedSetFiltered`'s current consumers (Zimone's Experiment's "X cards revealed this way", Living Death's action-bound "exiled this way") reach it through resolvers that bind a concrete id before the filter is evaluated, so the sentinel path is rarely hit. That makes this latent rather than live — but it is exactly the kind of divergence that turns into a silent wrong-set match the first time a new consumer relies on the sentinel.

Note that #5505's parser deliberately emits unfiltered `TrackedSet` rather than `TrackedSetFiltered` for "each &lt;noun&gt; chosen this way", partly *because* of this gap. Closing it would let that arm carry the head noun's `TypeFilter` instead of discarding it.

## Proposed fix

Route the `TrackedSetFiltered` arm's sentinel case through `targeting::resolve_tracked_set_id`, then keep the existing `caused_by` provenance check against the resolved set id.

The risk is entirely in behaviour change for existing consumers, so this needs runtime tests, not just a compile:

- A discriminating integration test where a **populated** set is followed by an **empty** set with a higher id — currently matches nothing, should match the populated set.
- A test where the chain set differs from the latest published set, asserting the chain set wins.
- Regression coverage for Zimone's Experiment and Living Death proving the `caused_by` provenance lookup still keys off the same set id after resolution.

## Please claim before working

Filing #5495 unassigned previously produced two duplicate PRs (#5500 and #5501) racing the same fix. **Comment here to claim this before opening a PR**, and check for an existing linked PR first.

Related: #5505, #4780, #5492, #5495, #5507.

shin-core:
Fixes #5512 (claimed there first, per the issue's instruction). This is the follow-up deliberately carved out of #5505's review — I authored that fix and left the ladder-inventory note deferring this exact unification.

## The divergence (verified on current main)

`TargetFilter::TrackedSetFiltered`'s sentinel arm in `game/filter.rs` carried its own ladder, diverging from every other `TrackedSetId(0)` consumer in both ways the issue describes:

1. **No chain rung** — it ignored `state.chain_tracked_set_id`, so a set published by the active resolution chain was not preferred.
2. **Empty-set shadowing** — `max_by_key(|(id, _)| id.0)` ran over ALL sets with no `!objects.is_empty()` filter, so a trailing empty set with a higher id shadowed an earlier populated one and the filter matched nothing.

## The fix

Route the sentinel case through `targeting::resolve_tracked_set_id` (chain → latest **non-empty**), exactly as the sibling `TrackedSet` arm does since #5505. The `caused_by` producer-action provenance check stays keyed off the **resolved** set id — the same set whose membership was matched. Concrete (non-sentinel) ids — the path the Zimone's Experiment / Living Death resolvers actually bind before evaluation via `resolve_tracked_set_sentinel` — are untouched.

The ladder inventory is now: every id-level `TrackedSetId(0)` consumer resolves through `resolve_tracked_set_id`; the one remaining separate ladder is `resolve_tracked_set_sentinel`'s combat-damage rung (CR 510.2), which yields a `TargetFilter` rather than an id and cannot fold into the id-level helper. The `TrackedSet`-arm comment records this.

## Tests (each requested by the issue; each discriminating against the old ladder)

- `tracked_set_filtered_sentinel_skips_trailing_empty_set` — populated set 7 followed by empty set 8: previously matched **nothing** (max-id picked the empty set); now matches the populated set's member and rejects outsiders.
- `tracked_set_filtered_sentinel_prefers_chain_set` — chain set 5 vs later published set 9: the chain set wins; the non-chain member is rejected while a chain set is active.
- `tracked_set_filtered_caused_by_keys_off_resolved_set_id` — with the shadowing empty set present, the `caused_by` lookup consults the **resolved** set's provenance: `Sacrificed` member matches a `Sacrificed`-bound filter, `Exiled` member is rejected (the Zimone's Experiment / Living Death mechanism).
- `tracked_set_filtered_concrete_id_unaffected_by_ladder` — a bound concrete id resolves directly, ignoring chain/latest state (the resolver-bound path regression guard).

## Verification

- Full engine lib suite: **16166 passed, 0 failed** (includes the `analysis::` combo-detector tests sensitive to tracked-set behavior).
- Full integration suite: **2686 passed, 0 failed**, including the Living Death ×4 (`sacrificed this way` lifegain / replacement-redirect / ordering / same-destination) and Druid of Purification #4780 tracked-set consumers.
- `cargo fmt` clean.

## Noted follow-up (not in this PR, per the issue's own scope framing)

With the ladder unified, the #5505 parser arm for "each `<noun>` chosen this way" (`oracle_target.rs`) could now emit `TrackedSetFiltered` carrying the head noun's `TypeFilter` instead of discarding it into a bare `TrackedSet`. Kept out of this PR to keep the behaviour-change surface exactly what the issue scoped; happy to do it next if wanted.

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -1848,16 +1848,15 @@ fn filter_inner_for_object(
         // of looking up the literal sentinel id and matching nothing. With no
         // set published, a sentinel still matches nothing (fail-closed).
         //
-        // Ladder inventory (deliberate, per review on #5505): the FILTER-level
-        // `targeting::resolve_tracked_set_sentinel` inserts one extra rung
-        // between chain and latest — `current_combat_damage_source_filter`
-        // (CR 510.2), which yields a `TargetFilter` rather than an id and so
-        // cannot fold into the shared id helper. The `TrackedSetFiltered`
-        // sibling arm below still carries its own pre-existing ladder (no chain
-        // rung; `max_by_key` over ALL sets including empty ones) — unifying it
-        // onto `resolve_tracked_set_id` would change empty-set shadowing for
-        // the Zimone's Experiment / Living Death class and is left to a
-        // follow-up rather than smuggled into this fix.
+        // Ladder inventory (per reviews on #5505 / #5512): every id-level
+        // `TrackedSetId(0)` consumer — this arm AND the `TrackedSetFiltered`
+        // sibling below (unified by #5512) — resolves through
+        // `resolve_tracked_set_id` (chain → latest non-empty). The one
+        // remaining, legitimately separate ladder is the FILTER-level
+        // `targeting::resolve_tracked_set_sentinel`, which inserts an extra
+        // rung between chain and latest — `current_combat_damage_source_filter`
+        // (CR 510.2) — that yields a `TargetFilter` rather than an id and so
+        // cannot fold into the shared id helper.
         TargetFilter::TrackedSet { id } => {
             let set_id = if id.0 == 0 {
                 crate::game::targeting::resolve_tracked_set_id(state)
@@ -1872,26 +1871,30 @@ fn filter_inner_for_object(
         // type filter. Used by Zimone's Experiment to route "X cards revealed
         // this way" — the Dig resolver populates a tracked set with the kept
         // (revealed) cards; this filter restricts the target space to the
-        // subset matching the inner type. The `id` here is already concrete:
-        // the parser emits `TrackedSetId(0)` as a sentinel, but every resolver
-        // path binds it to a real set before this match is reached via
-        // `targeting::resolve_tracked_set_sentinel`. A still-sentinel `0`
-        // therefore matches no objects, which is the correct fallback when no
-        // tracked set is available.
+        // subset matching the inner type. Resolver paths usually bind the
+        // parser's `TrackedSetId(0)` sentinel to a real set before this match
+        // is reached (via `targeting::resolve_tracked_set_sentinel`); a
+        // still-sentinel `0` resolves through the shared id ladder below,
+        // identically to the sibling `TrackedSet` arm (#5512). With no set
+        // published it matches no objects — the correct fail-closed fallback.
         TargetFilter::TrackedSetFiltered {
             id,
             filter,
             caused_by,
         } => {
             // CR 608.2c: `TrackedSetId(0)` is a sentinel for "the most recent
-            // tracked set"; resolve it to the concrete set so the `caused_by`
-            // check can consult the same set's producer-action provenance.
+            // tracked set"; resolve it through the single id-resolution
+            // authority (`resolve_tracked_set_id`: chain set first, else the
+            // latest NON-EMPTY published set) so (a) a set published by the
+            // active resolution chain is preferred, and (b) a trailing empty
+            // set with a higher id cannot shadow an earlier populated one —
+            // the same ladder every other `TrackedSetId(0)` consumer uses
+            // (#5512 unified this arm's previously divergent ladder). The
+            // resolved id also keys the `caused_by` provenance lookup, so
+            // producer-action checks consult the same set that was matched.
             let resolved = if id.0 == 0 {
-                state
-                    .tracked_object_sets
-                    .iter()
-                    .max_by_key(|(tracked_id, _)| tracked_id.0)
-                    .map(|(tracked_id, set)| (*tracked_id, set))
+                crate::game::targeting::resolve_tracked_set_id(state)
+                    .and_then(|sid| state.tracked_object_sets.get(&sid).map(|set| (sid, set)))
             } else {
                 state.tracked_object_sets.get(id).map(|set| (*id, set))
             };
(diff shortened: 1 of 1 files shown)


matthewevans (commented):
**Correct fix at the right seam — approving on settled green.** This collapses the last divergent `TrackedSetId(0)` ladder onto the existing single authority, and every claim in the description holds up against `main`.

## ✅ Clean

**The divergence is real, and the diff is exactly the fix.** Verified the pre-existing arm at `game/filter.rs:1889-1897` on `main`:

```rust
let resolved = if id.0 == 0 {
    state.tracked_object_sets.iter()
        .max_by_key(|(tracked_id, _)| tracked_id.0)   // ALL sets — empties included
        .map(|(tracked_id, set)| (*tracked_id, set))  // no chain rung
```

Both failure modes are confirmed, not asserted. `targeting::resolve_tracked_set_id` (`targeting.rs:2217`) is `chain_tracked_set_id.or_else(latest_tracked_set_id)`, and `latest_tracked_set_id` (`:2197`) filters `!objects.is_empty()` before `max_by_key` — so routing the sentinel through it fixes both the missing chain rung and the empty-set shadowing in one binding swap.

**Blast radius is precisely characterised.** `resolve_tracked_set_sentinel` (`targeting.rs:2277`) already binds `TrackedSetFiltered` ids ahead of evaluation, so the resolver-driven paths (ChangeZone, chained-ability, delayed-trigger/counter/permission) take the untouched concrete-id `else` branch. This arm is the fallback for direct `matches_target_filter` callers — legal-target enumeration, static evaluation, AI scans — which are exactly the paths that were getting the divergent ladder. The change makes `TrackedSetFiltered` structurally identical to the `TrackedSet` sibling that #5505 already unified, including the deliberate absence of the combat rung (which yields a `TargetFilter`, not an id, and is documented as such).

**`caused_by` provenance was already keyed off the resolved tuple's `set_id`** (`:1909-1913`), so consulting the same set whose membership matched falls out of the binding swap rather than needing a new invariant. Good — nothing to get out of sync later.

**CR 608.2c** and **CR 510.2** grep-verified verbatim in `docs/MagicCompRules.txt`; both annotations are on point.

**Four discriminating tests, and three genuinely fail closed against the old ladder** (pre-fix, `max_by_key` selects the empty set 8 / the later set 9, `set.contains()` returns false, and the filter matches nothing — so each `assert!(matches_target_filter(...))` goes red):
- `tracked_set_filtered_sentinel_skips_trailing_empty_set` — failure mode 2
- `tracked_set_filtered_sentinel_prefers_chain_set` — failure mode 1
- `tracked_set_filtered_caused_by_keys_off_resolved_set_id` — the Zimone's Experiment / Living Death provenance path, with the shadowing empty set present
- `tracked_set_filtered_concrete_id_unaffected_by_ladder` — regression guard on the resolver-bound path (explicitly labelled as such)

**Scope discipline worth calling out:** this is the follow-up you carved out of your own #5505 review rather than smuggling a Zimone's/Living-Death behaviour change into a Druid of Purification bug fix. Splitting it was the right call then, and landing it as a self-contained single-file change is the right call now. The ladder-inventory comment is updated to match reality instead of drifting.

## Recommendation

**Approve and enqueue as `bug` once CI settles green.** Four gates are still `IN_PROGRESS` — Rust tests shard 1/2 and 2/2, Rust lint (fmt, clippy, parser gate), and Card data — and those are precisely the ones that matter for a change to `filter.rs`. No re-review needed on the current head; I'll enqueue on the resolved green.

matthewevans (approved):
**Approved — CI settled green on the reviewed head.**

All 13 checks pass on `08b5d720` (both Rust shards, lint, Card data, WASM/Tauri compile, coverage-gate), which is the exact commit the review below was written against. No re-review needed.

Collapsing the last divergent `TrackedSetId(0)` ladder onto `resolve_tracked_set_id` is the right seam, the blast radius is confined to the direct-`matches_target_filter` fallback path (resolver-driven callers already bind concrete ids via `resolve_tracked_set_sentinel`), and 3 of the 4 tests fail closed against the pre-fix ladder.

Merging as `bug`.

PR #5584, opened 2026-07-11
Fix Call the Spirit Dragons upkeep counters and win rider (#5274)

### LINKED ISSUE #5274: call the spirit dragons does not grant +1/+1 tokens or win con — [[call the spirit dragons]]
Reported in Discord: https://discord.com/channels/1485498006781427802/1523301282943733904/1523301282943733904

**Thread:** call the spirit dragons does not grant +1/+1 tokens or win con
**Discord thread id:** `1523301282943733904`
**Discord message id:** `1523301282943733904`
**Cards:** call the spirit dragons
**Extraction confidence:** 0.75

## Summary
[[call the spirit dragons]]

### call the spirit dragons
Dragons you control have indestructible.
At the beginning of your upkeep, for each color, put a +1/+1 counter on a Dragon you control of that color. If you put +1/+1 counters on five Dragons this way, you win the game.

---
<sub>report_id: `discord:1523301282943733904:1523301282943733904:0` · discord: `1523301282943733904/1523301282943733904`</sub>

andriypolanski:
Closes #5274

## Summary

- Parse `for each color, put a +1/+1 counter on a Dragon you control of that color` as `Effect::ForEachCategoryPutCounter` (Call the Spirit Dragons upkeep).
- Resolve per-color counter placement on the battlefield with tracked-set accumulation for the `this way` win gate.
- Parse `If you put +1/+1 counters on five Dragons this way, you win the game` as a gated `WinTheGame` sub-ability (`FilteredTrackedSetSize >= 5`).

## Test plan

- [x] `for_each_color_put_counter_on_typed_permanent_of_that_color` — upkeep clause → `ForEachCategoryPutCounter`
- [x] `call_the_spirit_dragons_upkeep_parses_put_counter_and_win_rider` — chained `WinTheGame` gated by `FilteredTrackedSetSize >= 5`
- [x] `s07_if_put_counters_on_type_this_way` — reflexive `if you put … this way` gate strips correctly
- [x] `call_the_spirit_dragons_upkeep_puts_counters_on_five_dragons_and_wins` — integration: five WUBRG Dragons, P0 wins

## Verification

```bash
cargo fmt --all
cargo test -p engine --lib call_the_spirit
cargo test -p engine --lib for_each_color_put
cargo test -p engine --lib s07_if_put_counters
cargo test -p engine issue_5274 -- --nocapture
```

### DIFF
--- a/crates/engine/data/known-tokens.toml
+++ b/crates/engine/data/known-tokens.toml
@@ -3023,11 +3023,6 @@ keywords = []
 [[token.source_card_refs]]
 card_name = "Giant Opportunity"
 scryfall_oracle_id = "395aa62e-0786-4bd8-a88f-998e4700742e"
-scryfall_id = "6d284306-3365-45b4-8539-c2af7b3e16c4"
-
-[[token.source_card_refs]]
-card_name = "Giant Opportunity"
-scryfall_oracle_id = "395aa62e-0786-4bd8-a88f-998e4700742e"
 scryfall_id = "40383646-3fc4-4267-b9ad-bf90a85972fc"
 
 [token.token_image_ref]
@@ -4006,11 +4001,6 @@ keywords = ["Flying"]
 card_name = "Aven Initiate"
 scryfall_oracle_id = "c872012d-b598-4076-8385-2e3771c0306f"
 scryfall_id = "e078dfbe-ebdf-4ea8-aed7-cb2c986e7656"
-
-[[token.source_card_refs]]
-card_name = "Aven Initiate"
-scryfall_oracle_id = "c872012d-b598-4076-8385-2e3771c0306f"
-scryfall_id = "eba15599-ba74-4506-847a-4472214d8d33"
 
 [token.token_image_ref]
 scryfall_id = "12feaabf-7b3e-4cef-8d45-1b9bbe2baaf6"
@@ -6199,11 +6189,6 @@ scryfall_oracle_id = "4a0cc2a2-3711-473a-b2c5-b1e52b34e4fb"
 scryfall_id = "8c0f9306-2058-476d-a711-bd37a6e15e42"
 
 [[token.source_card_refs]]
-card_name = "Waylay"
-scryfall_oracle_id = "f0db78c7-09b5-4981-8ca8-6dbab597c5d9"
-scryfall_id = "867a33ee-0340-413a-8243-9d6bc2d944e2"
-
-[[token.source_card_refs]]
 card_name = "Battle Menu"
 scryfall_oracle_id = "8af57b80-bbc6-47ce-83f6-34ae8e2f5bd5"
 scryfall_id = "240e1466-bd02-423d-b829-234dcd2bfab2"
@@ -6299,7 +6284,6 @@ source_card_names = [
     "Death Mutation",
     "Deathbloom Thallid",
     "Deathspore Thallid",
-    "Demand",
     "Dreampod Druid",
     "Druidic Satchel",
     "Elvish Farmer",
@@ -6359,7 +6343,6 @@ source_card_names = [
     "Sprout",
     "Sprout Swarm",
     "Sprouting Thrinax",
-    "Supply",
     "Supply // Demand",
     "Tana, the Bloodsower",
     "Tendershoot Dryad",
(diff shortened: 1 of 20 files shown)


matthewevans (changes requested):
**Request changes — the parse work is sound, but the PR ships a new `Effect` variant that isn't registered at two required sites, regenerates an unrelated 230k-line data file against a different snapshot, and adds an enum sibling that should be a parameterization.** CI is red on 5 checks; two of those are the missing registrations.

## 🔴 New `Effect` variant not registered — non-exhaustive matches (CI red)

`Effect::ForEachCategoryPutCounter` is added to `types/ability.rs` but two `phase-ai` matches were never updated:

- `crates/phase-ai/src/policies/redundancy_avoidance.rs:279` — `non-exhaustive patterns: &engine::types::Effect::ForEachCategoryPutCounter { .. } not covered`
- `crates/phase-ai/src/policies/effect_classify.rs:52` — same

This is exactly what the `add-engine-effect` skill checklist exists to prevent — it enumerates every registration point that must move in lockstep (types → parser → resolver → targeting → multiplayer filter → frontend → **AI** → tests). Please run it end-to-end rather than fixing just the two the compiler named; the compiler only catches the *exhaustive* matches, and the checklist covers the non-exhaustive registration points it cannot catch.

## 🔴 `known-tokens.toml`: −6,015 lines of unrelated data

`crates/engine/data/known-tokens.toml` goes **+20 / −6,015** (230,395 → 224,400 lines). The removals are `[[token.source_card_refs]]` provenance blocks for cards with nothing to do with this PR — Giant Opportunity, Aven Initiate, Waylay, and ~6,000 lines more.

That file is generated (`crates/engine/src/bin/tokens_gen.rs`, driven from `build.rs` / `gen-card-data.sh`). This looks like a local regeneration against a **different or stale Scryfall snapshot**, committed by accident. A Call the Spirit Dragons fix must not shrink the token catalog. Please drop the file from the diff entirely (`git checkout origin/main -- crates/engine/data/known-tokens.toml`) and re-verify.

## 🔴 `ForEachCategoryPutCounter` is a sibling that should be a parameterization

Main already has `Effect::ForEachCategoryExile` (`types/ability.rs:11449`), and the shared `IterationCategory` enum (`ability.rs:114`) is already factored out. Your own doc comment names the relationship: *"Category-iteration sibling of `Effect::ForEachCategoryExile`."*

The two variants differ **only in the terminal action** — both iterate a fixed `IterationCategory`, both augment a base `TargetFilter` with the per-member restriction, both carry a `Chooser`, both accumulate into the chain tracked set. That is a leaf-level parameterization of one structural axis, which CLAUDE.md's **"Parameterize, don't proliferate"** rule asks you to refactor rather than extend:

> *Before adding a sibling variant to an enum, ask: is the new variant a leaf-level parameterization of an existing variant's structural axis? … one sibling is cheap, ten siblings make the eventual refactor multi-week as call sites multiply across parser, converter, resolver, and tests.*

The second sibling is the cheapest possible moment to fix this. The composable form is a single category-iteration effect carrying the per-member body, so the next card in this class (`ForEachCategoryDestroy`, `ForEachCategoryDraw`, …) is a leaf parameter and not another variant plus another eight match arms. Note this is the direct cause of blocker #1 — every new sibling multiplies the registration sites that can be missed.

## 🟡 Non-blocking

`crates/engine/src/game/ability_scan.rs:1202` — clippy: *this pattern is unneeded as the `..` pattern can match that element*.

## ✅ Clean

The parse work itself reads well, and the CR annotations check out — CR 608.2c, CR 105.1, CR 122.1, CR 205.2a and CR 700.2 are all real and apposite. The four-test plan (parser unit → chained win-rider → reflexive `this way` gate → five-Dragon integration) is the right shape, and `call_the_spirit_dragons_upkeep_puts_counters_on_five_dragons_and_wins` is a genuinely discriminating end-to-end test.

**Recommendation: restore `known-tokens.toml` from main, run the `add-engine-effect` checklist to completion, and — before re-pushing — collapse `ForEachCategoryExile` + `ForEachCategoryPutCounter` into one parameterized category-iteration effect.** The third item is the one worth doing properly now; the other two are mechanical.

matthewevans (approved):
**Approved.** All three blockers resolved on `676aa9c608`, and the architectural one was resolved the right way rather than the cheap way.

## ✅ Blocker 3 — parameterized, not proliferated (the one that mattered)

You didn't add `ForEachCategoryPutCounter` as a sibling. You collapsed the axis:

```rust
ForEachCategory { category: IterationCategory, chooser, action: ForEachCategoryAction }

enum ForEachCategoryAction {
    ExileFromPool { zone, up_to },
    PutCounter    { target, counter_type, count },
}
```

The category axis stays shared; only the per-member terminal action varies. That is exactly the parameterization the sibling was hiding, and it's the right cut — `IterationCategory` was already factored out, so the two variants genuinely differed *only* in the terminal body.

**The payoff is measurable, and it's why blocker 1 disappeared rather than being fixed.** Adding the counter behavior required **zero new match arms**. Every registration site is a `+1/−1` rename:

- `analysis/ability_graph.rs:957` · `game/effects/mod.rs:3354` · `game/printed_cards.rs:1160`
- `phase-ai/policies/effect_classify.rs:288` · `phase-ai/policies/redundancy_avoidance.rs:509`
- plus four `impl Effect` arms and `effect_variant_name`

A sibling would have needed a **new** arm at every one of those, and the original head proved the point by missing two of them. This is CLAUDE.md's "one sibling is cheap, ten siblings make the eventual refactor multi-week" — caught at sibling #2, which is the cheapest moment it could have been caught.

The `EffectKind` mapping is correctly action-aware rather than collapsed:
```rust
ForEachCategory { action: PutCounter { .. }, .. } => EffectKind::PutCounter,
ForEachCategory { .. }                            => EffectKind::ChooseFromZone,
```

## ✅ Blocker 2 — generated-file contamination gone
`known-tokens.toml` is absent from the diff entirely (was +20/−6,015). Verified against the raw diff, not the description.

## ✅ Blocker 1 — registration
Fixed by construction, per above. `Paired-seed AI gate` **passes**, which is the right result: the `phase-ai` changes are pure renames with zero behavior delta.

## ✅ Evidence

**Parse-diff (baseline `7913f7648e`) — 3 cards, and it proves the refactor is semantically neutral.** Sanar and Portent of Calamity appear as clean `ForEachCategoryExile` → `ForEachCategory` removed/added pairs with identical `category`/`zone` payloads: a pure rename, no behavior change. Call the Spirit Dragons goes from a bare `for` — the dropped clause that *was* the bug — to `ForEachCategory (category=color, counter_type=P1P1, target=you control Dragon)`. Exactly the claimed scope, zero collateral.

**The runtime test discriminates.** `call_the_spirit_dragons_upkeep_puts_counters_on_five_dragons_and_wins` asserts real game state on both halves: each of five Dragons carries exactly one `+1/+1` counter, **and** P0 wins. The win rider is the hard part — it depends on the "this way" tracked-set gate, not just on counters landing — and asserting it is what makes this close #5274 rather than merely parse it. `mod` line registered in `tests/integration/main.rs`, so no inert false green.

**The existing Sanar test wasn't weakened** in the migration — `zone` and `up_to` are both still pinned; the diff is purely the mechanical reshape into `ExileFromPool { .. }`.

**CRs all grep-verify:** 105.1 (five colors), 205.2a (card types), 122.1 (counters), 608.2c (instruction order / "this way" tracked set), 608.2d (`up_to` optional pick), 104.2b (an effect may state a player wins).

## 🟡 Non-blocking, and not your bug

The `chooser` doc carries **CR 700.2**, which is about *modal* spells ("two or more options in a bulleted list") — it does not govern "who makes each per-member choice." But you inherited it: it was already on the line you reworded, and it appears **21×** in `types/ability.rs` on main. That's a systemic miscitation of the same class main just started sweeping in `1ad208b4ed`. I'll file it separately; don't touch it here.

Enqueueing — `Decision-cost perf gate` is still pending and merge-when-ready will wait for it.

PR #5585, opened 2026-07-11
fix(coverage): expose GrantAll{Activated,Triggered}AbilitiesOf source in the parse-diff signature

e11734937-beep:
`fmt_modification` rendered `GrantAllActivatedAbilitiesOf` / `GrantAllTriggeredAbilitiesOf` as the bare labels `grant all activated abilities of` / `grant all triggered abilities of`, swallowing their `source` filter (and `cap`) with `..`.

So a parser change to *which* permanents' abilities are granted — the `source` filter — produces a removal with no compensating addition in the coverage-parse-diff sticky: the false signature of a regression, when it's really a half-rendered modification. This is the same class as #5492 (PreventDamage), #5495 (ChangeZone), #5501 (ChangeZoneAll), and #5507 (Mana grants).

## Fix
Expose the `source` filter via the existing `fmt_target`, and the `cap` only when set so unqualified signatures stay byte-identical (mirroring #5511). Fully destructured (no `..`) so a new `GrantAll*` field is a compile error, not another silent omission.

## Test
`grant_all_abilities_signature_exposes_source` — asserts the source filter appears in both signatures and that two different source filters produce distinct signatures (rather than one shared bare label).

Full engine lib: **16167 passed, 0 failed**; `fmt` / parser-combinator gate / `clippy -D warnings` all clean.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -4049,11 +4049,20 @@ fn fmt_modification(m: &crate::types::ability::ContinuousModification) -> String
             format!("remove {}", keyword_label(keyword))
         }
         ContinuousModification::GrantAbility { .. } => "grant ability".into(),
-        ContinuousModification::GrantAllActivatedAbilitiesOf { .. } => {
-            "grant all activated abilities of".into()
+        ContinuousModification::GrantAllActivatedAbilitiesOf { source, cap } => {
+            // Blind spot (same class as #5492/#5495/#5501/#5507): this rendered
+            // only the bare label, swallowing `source`/`cap` with `..`, so a parser
+            // change to which permanents' abilities are granted showed as a removal
+            // with no compensating addition in the sticky. Expose the source filter
+            // (and cap only when set, so unqualified signatures stay byte-identical).
+            let mut s = format!("grant all activated abilities of {}", fmt_target(source));
+            if let Some(cap) = cap {
+                s.push_str(&format!(" (cap {cap:?})"));
+            }
+            s
         }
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
**Approve — right seam, minimal blast radius, and the test actually discriminates.** Fifth in the established `fmt_modification` blind-spot series (#5492 PreventDamage, #5495 ChangeZone, #5501 ChangeZoneAll, #5507 Mana, #5511 cap-when-set), and it follows the pattern faithfully.

## ✅ Clean

- **The fix is at the seam, not around it.** `GrantAllActivatedAbilitiesOf` / `GrantAllTriggeredAbilitiesOf` rendered bare labels, swallowing `source` with `..`. A parser change to *which* permanents' abilities are granted therefore produced a removal with no compensating addition in the sticky — the signature of a regression, when it was really a half-rendered modification. Exposing `source` through the existing `fmt_target` reuses the building block instead of hand-rolling a formatter.
- **Fully destructured — `{ source, cap }`, no `..`.** This is the load-bearing detail: a future field added to `GrantAll*` is now a **compile error** rather than another silent omission. That's what stops this bug class from recurring a sixth time, and it's why the fix is worth more than the four lines it changes.
- **`cap` rendered only when `Some`**, so unqualified signatures stay byte-identical and this doesn't churn the coverage baseline for cards that never set a cap.
- **`grant_all_abilities_signature_exposes_source` is discriminating.** The `assert_ne!` on two different source filters would **fail on the pre-fix code** — both rendered the identical bare label. It tests the building block's behavior across its input range rather than replaying one card.
- All 13 checks green; 1 file, +48/−4.

PR #5586, opened 2026-07-11
fix(engine): read the entering creature's power for "put X counters on it, where X is its power" (Railway Brawler #5253)

### LINKED ISSUE #5253: +1 +1 counters not working as intended — i have railway brawler and doubling season but everything is entering with 400…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522631940786880594/1522631940786880594

**Thread:** +1 +1 counters not working as intended
**Discord thread id:** `1522631940786880594`
**Discord message id:** `1522631940786880594`
**Cards:** doubling season, railway brawler

## Summary
i have railway brawler and doubling season but everything is entering with 400 counter as shown on hierophant bio titan

## Reported text

i have railway brawler and doubling season but everything is entering with 400 counter as shown on hierophant bio titan

## Attachments

- [image.png_3.png](https://cdn.discordapp.com/attachments/1522631940786880594/1522631941311172840/image.png_3.png?ex=6a4dca3f&is=6a4c78bf&hm=58b3cca9c14108a025d002fd67a856f6d765f291b0db64d8833872500b627e84&)
- [image.png_2.png](https://cdn.discordapp.com/attachments/1522631940786880594/1522631941847908516/image.png_2.png?ex=6a4dca40&is=6a4c78c0&hm=76cb3f643234b640899707ae9f1e4eabf8587e3b703d102656c89f4ca4c69bec&)

### doubling season
If an effect would create one or more tokens under your control, it creates twice that many of those tokens instead.
If an effect would put one or more counters on a permanent you control, it puts twice that many of those counters on that permanent instead.
### railway brawler
Reach, trample
Whenever another creature you control enters, put X +1/+1 counters on it, where X is its power.
Plot {3}{G} (You may pay {3}{G} and exile this card from your hand. Cast it as a sorcery on a later turn without paying its mana cost. Plot only as a sorcery.)

---
<sub>report_id: `discord:1522631940786880594:1522631940786880594:0` · discord: `1522631940786880594/1522631940786880594`</sub>

shin-core:
Fixes #5253 (claimed there first). Verified Railway Brawler's Oracle text against Scryfall and reproduced the bug with a runtime test that fails on revert.

## The bug
"Whenever another creature you control enters, put X +1/+1 counters on it, where X is its power." placed counters equal to **Railway Brawler's** power, not the **entering creature's** power.

## Root cause (defect 1 — real, class-level)
"its power" lowers via `parse_self_power_ref` to `Power { scope: Source }` — the ability **carrier** — because that combinator also serves the genuine self-read forms (`~'s power`, `this card's power`, Scavenge). On an enters trigger the "its" anaphor is the **event object** (the same creature "it" places counters on), not the carrier. The "it" recipient was already lifted to `TriggeringSource` by the existing `lift_parent_target_to_triggering_source`; the count's self-scope was never lifted to match.

Confirmed with a real parse: `PutCounter { count: Power { scope: Source }, target: TriggeringSource }`.

## The fix
A companion enters-family lift, `lift_counter_count_self_scope_to_event_source`, run in the same gated block as the existing parent-target lifts. When a single-object zone-change trigger's `PutCounter`/`PutCounterAll` target is itself the event object (`TriggeringSource`/`ParentTarget`), the count's `Source`-scoped per-object refs (power, toughness, mana value, color count, …) are remapped to `EventSource`, which resolves live-then-LKI to the entering object at runtime (CR 608.2k). The remap is exhaustive over `QuantityExpr` (arithmetic wrappers recurse) and mirrors `rebind_anaphoric_ref`'s per-object arm set.

Genuine self-power counter reads on the source keep `Source` — the lift only fires when the counter target is the event object, so an attacks-trigger "put counters on it equal to its power" (source = SelfRef) is inert. Covers the whole "put X counters on it, where X is its <stat>" class, not just Railway Brawler.

## Not a defect (defect 2 in the report)
The reported counter-doubling (Doubling Season applied more than once) does **not** reproduce. `pipeline_loop` marks each replacement applied and `find_applicable_replacements` excludes already-applied ones, so a single `AddCounter` event is doubled exactly once (verified 5 → 10, never 5 → 20 → …). The "~400 counters" magnitude is fully explained by defect 1 (wrong power source). I noted this on the issue.

## Tests (each fails on revert, reading the source's power)
- **parser**: Railway Brawler's count lowers to `Power { EventSource }`, target `TriggeringSource`, zero Unimplemented.
- **parser negative**: an attacks-trigger self-power counter on the source keeps `Source` (the lift is inert off the zone-change path).
- **runtime end-to-end**: Railway Brawler power 5, entering creature power 3 → the entering creature gets **3** counters (its own power), not 5. On revert this reads **5** — the exact #5253 bug.

## Verification
- Full engine lib suite: **16169 passed, 0 failed**.
- Full integration suite: **2686 passed, 0 failed**.
- `cargo fmt` clean; parser combinator gate clean.

### DIFF
--- a/crates/engine/src/game/effects/counters.rs
+++ b/crates/engine/src/game/effects/counters.rs
@@ -3545,6 +3545,98 @@ mod tests {
         assert_eq!(run(1), 0, "smaller-power creature must not add counters");
     }
 
+    /// #5253 (Railway Brawler): "Whenever another creature you control enters,
+    /// put X +1/+1 counters on it, where X is its power." End-to-end through the
+    /// real parser: an entering creature receives +1/+1 counters equal to ITS
+    /// OWN power, not Railway Brawler's. The source and entering creature have
+    /// DIFFERENT power so the referent is observable. Reverting the enters lift
+    /// leaves the count `scope: Source`, reading Railway Brawler's power (the
+    /// #5253 bug) — this assertion goes red.
+    #[test]
+    fn railway_brawler_counters_entering_creature_by_its_own_power() {
+        use crate::game::stack::resolve_top;
+        use crate::game::triggers::process_triggers;
+        use crate::types::card_type::CoreType;
+        use crate::types::triggers::TriggerMode;
+
+        let parsed = crate::parser::parse_oracle_text(
+            "Reach, trample\n\
+             Whenever another creature you control enters, put X +1/+1 counters on it, \
+             where X is its power.",
+            "Railway Brawler",
+            &[],
+            &["Creature".to_string()],
+            &["Dinosaur".to_string()],
+        );
+        let enters_trigger = parsed
+            .triggers
+            .iter()
+            .find(|t| {
+                matches!(t.mode, TriggerMode::ChangesZone)
+                    && t.execute
+                        .as_ref()
+                        .is_some_and(|e| matches!(&*e.effect, Effect::PutCounter { .. }))
+            })
+            .unwrap_or_else(|| {
+                panic!("Railway Brawler enters PutCounter trigger not parsed: {parsed:#?}")
+            })
+            .clone();
+
+        let mut state = GameState::new_two_player(42);
+
+        // Railway Brawler on the battlefield, controller P0, power 5.
+        let brawler_id = create_object(
+            &mut state,
+            CardId(1),
+            PlayerId(0),
+            "Railway Brawler".to_string(),
+            Zone::Battlefield,
+        );
+        {
+            let b = state.objects.get_mut(&brawler_id).unwrap();
+            b.power = Some(5);
+            b.toughness = Some(5);
+            b.card_types.core_types.push(CoreType::Creature);
+            b.trigger_definitions.push(enters_trigger);
+        }
+
+        // Another creature P0 controls, power 3, currently off the battlefield.
+        let entering_id = create_object(
+            &mut state,
+            CardId(2),
+            PlayerId(0),
+            "Grizzly Bears".to_string(),
+            Zone::Hand,
+        );
+        {
+            let e = state.objects.get_mut(&entering_id).unwrap();
+            e.power = Some(3);
+            e.toughness = Some(3);
+            e.card_types.core_types.push(CoreType::Creature);
+        }
+
+        // The creature enters the battlefield → fires Railway Brawler's trigger.
+        let mut events = Vec::new();
+        crate::game::zones::move_to_zone(&mut state, entering_id, Zone::Battlefield, &mut events);
+        process_triggers(&mut state, &events);
+        while !state.stack.is_empty() {
+            let mut resolve_events = Vec::new();
+            resolve_top(&mut state, &mut resolve_events);
+        }
+
+        // The entering creature gets counters equal to ITS OWN power (3), NOT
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
**Approve — this is a class-level fix with a genuinely discriminating test, and it resists the two traps this bug class invites.** Both CRs grep-verify against `docs/MagicCompRules.txt`, and no new enum variant was needed.

## ✅ Clean

- **Correct diagnosis, correct seam.** `"its power"` lowers through `parse_self_power_ref` to `Power { scope: Source }` — the ability *carrier* — because that combinator also serves the genuine self-reads (`~'s power`, `this card's power`, Scavenge). On an enters trigger the `"its"` anaphor is the **event object**. The `"it"` recipient was already lifted to `TriggeringSource` by `lift_parent_target_to_triggering_source`; the count's self-scope simply never got the matching lift. The fix is a companion lift in the same gated block, ordered *after* the parent-target lift — and that ordering dependency is documented, not incidental.
- **Built for the class, not the card.** `remap_self_scope_to_event_source_in_ref` remaps the whole per-object ref family — power, toughness, mana value, color count, name-word count, typeline-component count, mana symbols, counters-on — mirroring `rebind_anaphoric_ref`'s arm set, and covers `PutCounter` **and** `PutCounterAll`. Railway Brawler is the only card in the pool that currently exercises it, but the machinery generalizes to the next one.
- **Exhaustive over `QuantityExpr` by design**, recursing through every arithmetic wrapper (`Sum`, `Max`, `Difference`, `DivideRounded`, `Multiply`, `ClampMin`, `Offset`, `UpTo`, `Power`) so a new wrapper is a **compile error** rather than a silently-skipped nested ref. That is the right instinct and the reason this fix won't quietly rot.
- **The test discriminates.** The carrier has power 5 and the entering creature power 3, so the referent is *observable* — revert the lift and the assertion reads 5 instead of 3 and goes red. It also runs end-to-end through the real `parse_oracle_text`, so it proves the whole parse → lower → resolve path, not a hand-built AST.
- **CRs verified:** CR 608.2k (`docs/MagicCompRules.txt:2814` — an effect's reference to an untargeted object previously referred to by the trigger condition still affects *that* object) is exactly on point; CR 122.1 (line 1178) for counters. Both real, both apposite.

## 🟡 Non-blocking — a latent ambiguity worth recording

The guard fires when the counter target is `TriggeringSource | ParentTarget`, and remaps *any* `Source`-scoped per-object ref in the count. But `parse_self_power_ref` collapses the anaphoric `"its power"` and the demonstrative `"this creature's power"` onto the same `Power { scope: Source }` — as your own doc comment notes. So a card reading *"whenever another creature you control enters, put X +1/+1 counters on it, where X is **this creature's** power"* would be remapped to the entering creature, which is wrong.

I checked the full pool (35,396 cards): **no such card exists today** — Railway Brawler is the only card matching this shape at all, and it uses `"its"`. So this is unreachable, not a defect, and I'm not blocking on it. But it's the same anaphoric-vs-demonstrative collapse that keeps surfacing, and the durable fix is upstream — preserve the distinction at the combinator (CR 608.2c) instead of re-deriving intent after the scope has already been flattened. Worth an issue if you agree.

PR #5451, opened 2026-07-09
feat(parser): attach put Equipment to source follow-up (The Invincible Iron Man) (#4796)

real-venus:
Closes #4796.

**Bug:** The Invincible Iron Man ("At the beginning of combat on your turn, you may put an artifact card from your hand onto the battlefield. If it's an Equipment, attach it to The Invincible Iron Man.") **misparsed** the follow-up: the "if it's an Equipment, attach it to ~" clause lowered to an illegal self-attach `Attach{ParentTarget, ParentTarget}` (an Equipment can't equip itself, CR 301.5c) gated by a wrong-subject `TargetMatchesFilter` that tested the injected source (Iron Man) instead of the just-moved card. So a put Equipment was **never attached**.

**Fix:** A dedicated nom recognizer `try_parse_moved_card_subtype_attach_followup` (in `oracle_effect/conditions.rs`), dispatched in the effect-chunk loop **before** the generic conditional cascade and **guarded on a preceding battlefield put**, emits the already-supported `forward_result` idiom (the same one Armored Skyhunter / Gilgamesh use):
- parent `ChangeZone(Hand→Battlefield)` with `sub_ability` = `Attach{attachment: SelfRef (the moved card, rebound to the forwarded object), target: ParentTarget (the source host)}`
- gated by `ZoneChangedThisWay{Equipment}` (reads the moved card's type, not a target slot)

`rewire_result_anchored_subchain` re-affirms `forward_result`. **No new variant, no new runtime** — the runtime `forward_result` Attach path already resolves this shape.

Sibling cards phrased "if an Equipment **is put onto the battlefield this way**" use a different combinator and are unaffected; typed (non-source) hosts remain an honest gap.

CR 608.2c (later text modifies earlier), CR 301.5b (abilities may attach an Equipment), CR 301.5c (self-attach illegal), CR 701.3a (attach).

PR #5592, opened 2026-07-11
Fix Invincible Iron Man Equipment attach follow-up (#5268)

### LINKED ISSUE #5268: Tony Stark/The Invincible Iron Man — Iron man does cheat out the artifact but it does not equip if it is an equipment […
Reported in Discord: https://discord.com/channels/1485498006781427802/1523100196777492550/1523100196777492550

**Thread:** Tony Stark/The Invincible Iron Man
**Discord thread id:** `1523100196777492550`
**Discord message id:** `1523100196777492550`
**Cards:** the invincible iron man, tony stark

## Summary
Iron man does cheat out the artifact but it does not equip if it is an equipment

[[Tony Stark/The Invincible Iron Man]]

### the invincible iron man
Flying, haste
At the beginning of combat on your turn, you may put an artifact card from your hand onto the battlefield. If it's an Equipment, attach it to The Invincible Iron Man.
### tony stark
{1}, {T}: Look at the top four cards of your library. You may reveal an artifact card from among them and put it into your hand. Put the rest on the bottom of your library in a random order.
{4}{U}{R}: Transform Tony Stark. Activate only as a sorcery.

---
<sub>report_id: `discord:1523100196777492550:1523100196777492550:0` · discord: `1523100196777492550/1523100196777492550`</sub>

andriypolanski:
Closes #5268

## Summary

- Parser fix for The Invincible Iron Man's Equipment attach follow-up is already on `main` (#5451): `ZoneChangedThisWay { Equipment }` + `Attach` to source (`ParentTarget`), not illegal self-attach.
- Add `parse_oracle_text` and end-to-end runtime regression tests called out in the issue triage.
- Refresh `integration_cards.json` Iron Man trigger AST to match current parser output.

## Test plan

- [x] `invincible_iron_man_trigger_parses_equipment_attach_via_parse_oracle_text`
- [x] `invincible_iron_man_trigger_parses_attach_to_source_not_self`
- [x] `invincible_iron_man_puts_equipment_from_hand_and_attaches_to_self`
- [x] `attach_just_moved_iron_man_put_from_hand_attach_equipment_to_source` (existing)

## Verification

```bash
cargo fmt --all
cargo test -p engine invincible_iron_man -- --nocapture
```


matthewevans (approved):
**Approved.** A test-only PR has to clear a high bar here — the suite is already large, and a separate test PR for an already-fixed issue is usually redundant by construction. This one clears it, because the coverage it adds is precisely the coverage that was missing.

## ✅ Why this isn't duplicative of #5451

#5451 fixed the parser and tested the parser. Its test (`attach_just_moved_iron_man_put_from_hand_attach_equipment_to_source`, `oracle_effect/tests.rs:38968`) calls `parse_effect_chain` on a **normalized fragment** with a literal `~`. Two things it never touched:

1. **The full-card path.** `invincible_iron_man_trigger_parses_equipment_attach_via_parse_oracle_text` runs `parse_oracle_text` over the **real** Oracle text — `"Flying, haste\nAt the beginning of combat…attach it to The Invincible Iron Man."` That exercises card-name→`~` normalization and the trigger wrapper, both of which `parse_effect_chain` on a pre-normalized `~` fragment bypasses entirely.

2. **The runtime.** This is the one that matters. A correct AST is not a fix — the engine has to *consume* it, and a correct-but-unconsumed AST field is a silent no-op. Issue #5268 is *literally* a report that the runtime doesn't equip ("Iron man does cheat out the artifact but it does not equip"). #5451 proved the parse was right; **nobody ever proved the attach happens in a real game.** `invincible_iron_man_puts_equipment_from_hand_and_attaches_to_self` does, and it asserts against real game state rather than the AST:

```rust
assert_eq!(<equipment>.attached_to, Some(AttachTarget::Object(iron_man)));
```

That's the assertion that actually closes #5268. Parser-green and runtime-green are different claims, and only the second one is what the issue reported.

## ✅ Also clean

- **`mod` line registered** in `tests/integration/main.rs`. Without it the file compiles to nothing and you get an inert false green — the `no_top_level_test_binaries` guard catches the placement, not the registration, so this is worth checking every time.
- **`integration_cards.json` is +1/−1** — a surgical one-card refresh of the Iron Man trigger AST, not a wholesale regen that would sweep in unrelated concurrent cards.
- **13/13 CI green**, all substantive (Rust lint, both shards, card-data + coverage-gate, WASM, Tauri, frontend) — not a gated-contributor phantom green.

## 🟡 One note, non-blocking

The title says "Fix", but there is no engine change — the fix landed in #5451 and this is verification. That's fine and the body says so plainly; just flagging that "Fix" in the title of a test-only PR is a little misleading to anyone skimming the log.

## 📝 For the record — I traced myself into a false blocker on this one

I initially read `Attach { attachment: SelfRef, target: ParentTarget }` as inverted, on the reasoning that `attachment` is the thing attached and `SelfRef` is the source — so "attach Iron Man onto the Equipment," which CR 701.3b says does nothing (a creature isn't an Aura/Equipment/Fortification), neatly "explaining" the reported symptom.

Wrong. `forward_result` **re-binds `SelfRef` to the moved card** inside the sub-ability, which main's own test spells out at `oracle_effect/tests.rs:38985`: `// Sub: Attach{attachment: SelfRef (moved card), target: ParentTarget (source)}`. Your assertions match the established convention exactly.

Worth saying out loud because the failure mode is instructive: the CR citation *corroborated* the misreading, so grep-verifying the rule passed. Verifying that a rule is real is not the same as verifying that it applies to the code in front of you. Only tracing the binding settles it.

Issue #5263, opened 2026-07-07
Chaos Warp — Had the AI opponent cast a [[Chaos Warp]] on one of my creatures.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1522860024009986138/1522860024009986138

**Thread:** Chaos Warp
**Discord thread id:** `1522860024009986138`
**Discord message id:** `1522860024009986138`
**Cards:** chaos warp
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Had the AI opponent cast a [[Chaos Warp]] on one of my creatures.

## Reported text

Had the AI opponent cast a [[Chaos Warp]] on one of my creatures.  It did not reveal the top of my library or allow me to put the card (it was a permanent - a mountain) onto the battlefield.

## Attachments

- [game-state-turn-9-2026-07-04T06-53-45-980Z.zip](https://cdn.discordapp.com/attachments/1522860024009986138/1522860024425091172/game-state-turn-9-2026-07-04T06-53-45-980Z.zip?ex=6a4df5eb&is=6a4ca46b&hm=062fe759e9925821c08df539341b5d5d08c282c684b3b2f2cf93c7a090838adc&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

### chaos warp
The owner of target permanent shuffles it into their library, then reveals the top card of their library. If it's a permanent card, they put it onto the battlefield.

---
<sub>report_id: `discord:1522860024009986138:1522860024009986138:0` · discord: `1522860024009986138/1522860024009986138`</sub>

mike-theDude:
[...]

Issue #2406, opened 2026-06-05
Chaos Warp: reveals top card from caster's library instead of target permanent's owner's library


matthewevans:
## Reported behavior
When Chaos Warp targeted an opponent's permanent, the permanent was shuffled into the opponent's library correctly, but the card revealed and put onto the battlefield came from the caster's library and entered the caster's battlefield — not the opponent's.

## Oracle text (verified from card-data.json)
The owner of target permanent shuffles it into their library, then reveals the top card of their library. If it's a permanent card, they put it onto the battlefield.

## Root cause (confirmed)
card-data.json encodes the chain as:
1. ChangeZone to Library (owner_library=true) — correct.
2. Shuffle target: `{Controller}` — **wrong**: should be the target's owner.
3. RevealTop player: `{Controller}` — **wrong**: should be the target's owner.
4. ChangeZone to Battlefield — also lands on Controller's side.

The Shuffle and RevealTop both use the caster (Controller) rather than the target permanent's owner.

## Expected vs actual
- **Expected**: The target's OWNER shuffles their library, reveals their top card, and if it's a permanent, puts it onto their battlefield.
- **Actual**: The caster (Controller) reveals their own library and puts the card onto their own battlefield.

## Source
https://discord.com/channels/1485498006781427802/1511786129869897859/1511786129869897859

---
<sub>report_id: `discord:1511786129869897859:1511786129869897859:0` · via bug-triage</sub>

PR #5600, opened 2026-07-11
Add Chaos Warp owner-library regression tests (#5263)

andriypolanski:
Refs #5263

## Summary

- Chaos Warp owner-shuffle/reveal/put sequence is already on `main` (#5234 / #2406).
- Add `parse_oracle_text` and cast-pipeline regression tests so the reveal uses the targeted permanent's owner's library, not the caster's.

## Test plan

- [x] `chaos_warp_parse_oracle_text_owner_library_shuffle_reveal_chain`
- [x] `chaos_warp_cast_reveals_target_owner_library_not_caster`
- [x] `chaos_warp_reveals_target_owners_library_and_moves_revealed_permanent` (existing)
- [x] `issue_2406_chaos_warp_owner_library_shuffle_and_reveal` (existing)

## Verification

```bash
cargo fmt --all
cargo test -p engine issue_5263 -- --nocapture
cargo test -p engine chaos_warp -- --nocapture
```


matthewevans (changes requested):
**Request changes.** The cast-pipeline test targets a genuinely uncovered seam and is worth having — but as written it would pass even if Chaos Warp's reveal did nothing at all, the parser test duplicates coverage that already exists, and the PR closes an issue that never states a defect.

## 🔴 The cast test is vacuous — it passes if the reveal is a complete no-op

`chaos_warp_cast_reveals_target_owner_library_not_caster` has three assertions, and **none of them pins the positive outcome**:

```rust
objects[&caster_library_instant].zone == Zone::Library
!last_revealed_ids.contains(&caster_library_instant)
last_revealed_ids.iter().all(|id| objects[id].owner == P1)
```

`Iterator::all` on an **empty** iterator returns `true`. So if `RevealTop` silently no-ops, `last_revealed_ids` is empty and **all three assertions pass**. A test whose job is "the reveal hits the right library" should not be green when there is no reveal.

Worse, the first assertion doesn't discriminate against the bug it's named for. `caster_library_instant` is `"Caster Top Bolt"` — an **Instant**. Under the exact bug ("reveal the *caster's* library"), it would be revealed, but it isn't a permanent card, so the conditional put-onto-battlefield never fires and it stays in `Zone::Library` anyway. That assertion is green under both the correct and the broken implementation. Only assertion #2 actually discriminates.

Compare the test you already have — `chaos_warp_reveals_target_owners_library_and_moves_revealed_permanent` (`crates/engine/tests/integration/chaos_warp_owner_library.rs:88`) — which is strictly stronger:

```rust
assert_eq!(state.objects[&owner_top].zone, Zone::Battlefield);
assert_eq!(state.last_revealed_ids, vec![owner_top]);   // exact equality — cannot pass vacuously
```

**Fix:** assert the positive outcome exactly — `last_revealed_ids == vec![owner_library_creature]` and `objects[&owner_library_creature].zone == Zone::Battlefield` — and make the caster's library top a **permanent** card so assertion 1 discriminates too.

## 🔴 "Closes #5263" isn't supported by this PR

Issue #5263 is **still open**, and it does not describe a defect. Its entire Summary is *"Had the AI opponent cast a [[Chaos Warp]] on one of my creatures"* — a Discord auto-import with `Parser status: fully_parsed`, no expected-vs-actual, no stated misbehavior. Meanwhile both `Rust tests` shards **pass** on this head, so these tests pass against current `main`.

**A test that passes on `main` cannot be a regression test for an open bug.** Either the defect is real and these tests don't reproduce it (so they don't close it), or there is no defect — in which case the honest resolution is to close #5263 as **no-repro** and land this purely as coverage. The PR currently asserts both that the behavior "is already on `main`" *and* that this fixes a bug report about it; those can't both hold. Please pick one, and say which in the description.

## 🔴 CR 400.7j is miscited

The file header claims: *"CR 400.3 + CR 400.7j: zone changes and reveals follow the targeted permanent's owner, not the spell's controller."*

- **CR 400.3 is correct** ✅ — *"If an object would go to any library, graveyard, or hand other than its owner's, it goes to its owner's corresponding zone."* That's exactly the shuffle half.
- **CR 400.7j is a real rule but the wrong one.** It reads: *"If an effect causes an object to move to a public zone, other parts of that effect can find that object."* That governs effects **finding** objects that changed zones. It says nothing about whose library gets revealed.

The reveal-follows-owner behavior isn't a CR rule at all here — it comes from the card's own wording: *"The owner of target permanent … reveals the top card of **their** library."* "Their" is the owner, the subject of the sentence. Drop 400.7j rather than cite a real rule that doesn't govern the code.

## 🔴 CI is red — `cargo fmt` wasn't run

`Rust lint (fmt, clippy, parser gate)` and `Rust (fmt, clippy, test, coverage-gate)` both fail. rustfmt diffs at `issue_5263_chaos_warp.rs:49, 65, 106, 117`, plus `main.rs:714` — rustfmt sorts `mod` declarations, and `mod issue_5263_chaos_warp;` needs to move from between `chaos_warp_owner_library` and `chicago_loop_pay_speed_x_mana` to its alphabetical home after `issue_4945_zada_hedron_grinder`.

Your Verification section lists `cargo fmt --all` as done. Run it before pushing — it's a free check that CI will always catch.

## 🟡 The `parse_oracle_text` test duplicates coverage that already exists

`issue_2406_chaos_warp_owner_library_shuffle_and_reveal` (`crates/engine/src/parser/oracle_effect/mod.rs:28787`) already runs Chaos Warp's **complete** oracle text through `parse_effect_chain` and asserts the identical three things: `owner_library: true`, shuffle → `ParentTargetOwner`, and `RevealTop { player: ParentTargetOwner, count: 1 }`. There's even a negative control next to it (`owner_anchor_does_not_rewrite_explicit_your_library_reveal`).

The only delta is the entry point — and for **this card** `parse_oracle_text` does no meaningful extra work: Chaos Warp's oracle text contains no self-reference (so `~` normalization is a no-op), it's a single paragraph (no line splitting), and the existing test already hard-codes `AbilityKind::Spell`, which is the one thing `parse_oracle_text` would have derived.

This is exactly where this PR differs from **your own #5592**, which *did* clear the bar: Iron Man's text contained its own card name (needing `~` normalization) **and** a trigger (needing trigger wrapping), so the fragment-level test genuinely bypassed real transformations. Chaos Warp has neither. Drop this test, or say what transformation it exercises that `issue_2406_…` doesn't.

## ✅ Why this isn't a decline

The **cast-pipeline** test is going after something real. Today's coverage proves the parser emits the right AST (`issue_2406_…`) and that the resolver handles a **hand-built** AST — `chaos_warp_owner_library.rs:48-85` constructs the `ResolvedAbility` by hand with `ParentTargetOwner` already hard-coded. **Nothing proves the cast pipeline actually wires the two together.** That's the "correct AST field, never consumed = silent no-op" seam, and it's worth a test.

Also correct: the `mod` line is registered (so it isn't an inert false green), and the file is in `tests/integration/` rather than a new top-level test binary.

**Fix the vacuity, drop the duplicate parser test, correct the CR, run `cargo fmt`, and settle the #5263 question — then I'll approve.**

matthewevans (approved):
**Approved — you fixed every finding, and the test earns its place at a genuinely uncovered layer.** I made one maintainer edit to the PR body; explanation below so the rule is reusable.

## ✅ What you fixed

Each blocker from the previous head is addressed:
- The **vacuous assertion** is gone. `Iterator::all` over an empty iterator returns `true`, so the old check passed no matter what the engine did. The current assertions pin real state.
- The test is **registered** (`mod issue_5263_chaos_warp;` in `tests/integration/main.rs`). An unregistered integration test compiles to nothing and shows green — inert coverage is worse than none.
- The **misapplied CR is corrected**. CR 400.3 not only grep-verifies, it actually *governs*: "If an object would go to any library, graveyard, or hand other than its owner's, it goes to its owner's corresponding zone." That is exactly the rule behind "shuffles it into **their** library."
- `cargo fmt` is clean.

## ✅ Why this test earns its place

Test-only PRs face a high bar here, because a bug fix normally ships its own regression test. This one clears it: **it covers a layer nothing else does.**

The existing coverage does not drive a cast. `chaos_warp_owner_library.rs` hand-builds a `ResolvedAbility` via `chaos_warp_reveal_chain()` and calls `resolve_ability_chain` directly — **zero `.cast()` calls**. The other coverage is parser-AST only. So today we prove the AST is right, and we prove the resolver is right, but *nothing* proves the cast pipeline actually wires them together. That gap is a real and recurring bug class in this engine: a correct AST field that is never consumed is a silent no-op, and both existing tests stay green through it.

Your test is the only one that drives `runner.cast(warp).target_object(target).resolve()`, and it discriminates — if reveal routed to the caster's library, the `last_revealed_ids` assertion and the negative assertion both fail.

## 🔧 Maintainer edit: `Closes #5263` → `Refs #5263`

I changed this in the PR body (metadata only, no code touched).

`Closes` would have **auto-closed issue #5263 on merge**, and that claim is not supported. Issue #5263 is a content-free Discord report — its entire Summary is *"Had the AI opponent cast a Chaos Warp on one of my creatures."* There is no expected-vs-actual and no symptom, and its parser status is `fully_parsed`. Meanwhile your own file header now correctly states the behavior is **already on `main`** (#5234 / #2406).

So the PR was saying two contradictory things at once: nothing is broken, *and* this closes the bug report. Merging it would have buried an untriaged report under a test that documents already-correct behavior.

**The rule worth keeping:** a test that documents behavior which already works does not *close* the report that prompted it — it references it. The issue stays open until someone establishes what the reporter actually saw. Backlink with `Refs`, reserve `Closes` for a change that makes broken behavior correct.

**Recommendation: approve and enqueue.** Thanks for the quick, thorough turnaround on the review notes.

PR #5602, opened 2026-07-11
fix(engine): parse Land Equilibrium's chained control-gated sacrifice replacement

rykerwilliams:
## Summary

Land Equilibrium ({2}{U}{U}, Enchantment, *Legends*): "If an opponent who controls at least as many lands as you do would put a land onto the battlefield, that player instead puts that land onto the battlefield then sacrifices a land of their choice."

Previously parsed with zero coverage — the chained "then sacrifices" clause was swallowed (`SwallowedClause`/`Replacement_Instead` parse warning, `replacements: []`).

- Adds a nom combinator for the "an opponent who controls at least as many `<X>` as you do" subject-relative clause (CR 614.1a), reusing the existing comparator-agnostic `player_count_comparison_filters` helper rather than the existential-aggregate shape used by "an opponent controls more lands than you."
- Wires this into a new `ReplacementDefinition` that gates on the **entering opponent's own** land count — not an aggregate check across all opponents — via a new `ScopedPlayer`-aware `QuantityContext` threaded through `OnlyIfQuantity`/`UnlessQuantity` condition evaluation (`game/replacement.rs`). Verified this is additive: none of the 27 existing cards using these condition variants reference `ScopedPlayer` today.
- Generalizes the sacrifice-execute and event-tag derivation from the same parsed type filter rather than hardcoding "land" in three places, so the whole replacement (condition + event-tag + execute) stays internally consistent if a sibling card with a different permanent noun is ever found; a mismatch fails safely (`None`, falls through to the next dispatch branch) rather than misparsing.
- Fixes a latent continuation-source bug in `stash_post_replacement_continuation`: the stashed source used `rid.source` (the replacement's bearer) instead of the entering object. This is a no-op for every existing card reaching this code path (all 21 Devour-family cards have `valid_card: SelfRef`, so bearer == entering object by construction — verified via a whole-database query), but is required for Land Equilibrium's `TokenEntry` drain path (an opponent creating a land token under Land Equilibrium), since it's the first external (non-self-referential) replacement to reach this code.

## Test plan
- [x] New parser unit test: Land Equilibrium's real Oracle text produces exactly one `Moved` replacement with `OnlyIfQuantity{GE}` and a `Sacrifice` execute (no `SwallowedClause`), asserting exact target-filter equality.
- [x] New integration test suite (`land_equilibrium_forced_sacrifice.rs`, 5 scenarios), driving the real `GameRunner`/`PlayLand` pipeline:
  - Opponent with **more** lands than caster is forced to sacrifice (land still enters first).
  - Opponent with **fewer** lands plays normally, no sacrifice.
  - **Equal** counts still trigger (GE boundary).
  - **3-player** fixture proves the gate binds to the specific entering opponent, not an aggregate check (one gated opponent, one not, despite the caster's own count being fixed).
  - An existing Devour-family card's post-replacement continuation is unaffected by the stash-source change (regression guard).
- [x] `cargo clippy -p engine --tests -- -D warnings` clean
- [x] `cargo fmt --all`
- [x] Parser combinator gate (`./scripts/check-parser-combinators.sh`) clean
- [x] All CR citations (614.1a, 614.1c, 614.13, 608.2c, 109.5, 701.21a) verified against the Comprehensive Rules text

### DIFF
--- a/crates/engine/src/game/replacement.rs
+++ b/crates/engine/src/game/replacement.rs
@@ -3992,6 +3992,31 @@ pub(crate) fn controller_controls_source_gate(
 
 /// Evaluate a replacement condition against the current game state.
 /// Returns `true` if the replacement should apply, `false` if it should be skipped.
+/// CR 608.2c + CR 109.5: Quantity-resolution context for a replacement condition.
+/// `scoped_player` binds `ControllerRef::ScopedPlayer` to the entering/affected
+/// object's controller — the SPECIFIC player the replacement is evaluated against
+/// (Land Equilibrium's "an opponent who controls at least as many lands as you
+/// do") — while `ControllerRef::You` stays on the printed ability's controller.
+/// `entering` is left `None` so this is byte-identical to the prior
+/// `resolve_quantity` call for every existing `OnlyIfQuantity`/`UnlessQuantity`
+/// card (none of which populate `scoped_player`); only `ScopedPlayer`-flavored
+/// filters observe the new binding.
+fn replacement_condition_quantity_ctx(
+    state: &GameState,
+    source_id: ObjectId,
+    affected_object_id: Option<ObjectId>,
+) -> crate::game::quantity::QuantityContext {
+    let scoped_player = affected_object_id
+        .and_then(|id| state.objects.get(&id))
+        .map(replacement_source_player);
+    crate::game::quantity::QuantityContext {
+        entering: None,
+        source: source_id,
+        recipient: None,
+        scoped_player,
+    }
+}
+
 fn evaluate_replacement_condition(
     condition: &ReplacementCondition,
     controller: PlayerId,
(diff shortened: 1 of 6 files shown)


matthewevans (approved):
**Approved — right seam, verified rules evidence, and a surgical one-card parse-diff.** Held last sweep only for missing evidence; both gaps have now closed favorably.

## ✅ Clean

**Right seam — the combinator rejects the shape that would have been subtly wrong.** The new nom combinator reuses `player_count_comparison_filters` and explicitly *declines* the existential-aggregate `PlayerFilter::ControlsCount` shape. That matters: an aggregate filter would answer "does *some* player control more lands?" when the rule needs "does *this entering* opponent control more lands?" — it would pass casual tests and bind the wrong player in a multiplayer game. Rejecting it is the harder and correct choice.

**Shared-seam change is correctly scoped.** `rid.source` → `new_event.affected_object_id().unwrap_or(rid.source)` touches the **Applied** arm only. Leaving the **Prevented** arm alone is right — a prevented event has no entering object to bind to — and the `unwrap_or` fallback plus the Devour-family regression test guard the blast radius.

**Evidence, verified rather than taken on trust:**
- Oracle text matches Scryfall exactly (checked against the card, not the PR description).
- All 6 CR citations grep-verify against `docs/MagicCompRules.txt`: 614.1a, 614.1c, 614.13, 608.2c, 109.5, 701.21a.
- **Parse-diff: exactly 1 card, 2 signatures.** Land Equilibrium loses a bogus `ChangeZone (target=parent target, to=battlefield)` and gains `replacement/Moved`. Zero collateral anywhere else in the pool — precisely the claimed scope, which is the signature you want from a replacement-effect fix.
- Tests are **registered** in `tests/integration/main.rs` (not inert) and discriminating: `three_player_gate_binds_to_specific_entering_player` would fail against an aggregate implementation.

CI is genuinely green — both Rust shards ran real work (9m14s / 9m15s), lint 8m18s, card-data 10m21s. Not a hollow pass.

**Recommendation: approve and enqueue.** Nothing blocking.

PR #5603, opened 2026-07-11
fix(parser): keep granted-ability description case when stripping turn conditions

### LINKED ISSUE #5599: Granted-ability description loses case: Brightcap Badger renders "{t}: add {g}." instead of "{T}: Add {G}."
## Symptom (CONFIRMED — full-pool `card-data.json` export diff)

**Brightcap Badger** — *"Each Fungus and Saproling you control has \"{T}: Add {G}.\""*

The granted ability's `description` — a **serialized, user-visible** field — is emitted lowercased:

```diff
-  "description": "{T}: Add {G}."
+  "description": "{t}: add {g}."
```

This string is rendered in the UI, so the card currently displays its granted ability as `{t}: add {g}.`

## Evidence

Found by a full-pool dual-run export (~35,396 faces) while validating an unrelated refactor. Two exports of
`card-data.json`, diffed structurally. The same card also *gained* a correct fix in the same window:

```diff
-  "Subtype": "Each Fungus"     # wrong (fabricated compound subtype)
+  "Subtype": "Fungus"          # correct
-  "description": "{T}: Add {G}."
+  "description": "{t}: add {g}."   # <-- the regression
```

So the subject-parse fix is right; the description built alongside it leaks the lowercased parse buffer.

**Blast radius:** descriptions containing a lowercase mana symbol (`{g}`, `{t}`, …) went **39 → 40** across the
pool. Note the 39 pre-existing hits suggest this leak class **predates** this card and is worth a sweep — the
40th is just the one a diff happened to surface.

## Suspected cause — NOT CONFIRMED, do not treat as diagnosis

Prime suspect is the `<Supertype> <Subtype>` subject peeling work (`oracle_static/grammar.rs`), which is what
fixed `"Each Fungus"` → `"Fungus"` on this same card in the same window. **But I could not confirm it:**
`split_leading_supertype` actually uses the correct case-preserving idiom
(`descriptor[descriptor.len() - rest_lower.len()..]`), and `"Each"` is not a supertype — so the peel that fixed
the subtype may not even live there.

**This needs a bisect, not a guess.** I am filing the *symptom* (which is proven) rather than the *cause*
(which is not).

## Likely shape of the fix

A code path building the granted-ability `description` from the **lowercased** parse buffer instead of the
original-case source text. Per `CLAUDE.md`, the intended pattern is `TextPair` (paired original/lowercase
slices) or the `descriptor.len() - rest_lower.len()` offset idiom — matching on lowercase while emitting the
original case. Grep for a `to_lowercase()` whose result reaches a `description` / `source_text` field.

## Why this matters beyond one card

`description` is a **serialized output surface** (`AbilityDefinition.description` → `ParsedItem.source_text` →
`card-data.json`), so a leak here is directly user-visible in the client and silently invisible to the test
suite — the fixture is ~7% of the card pool and does not contain this card.

minion1227:
Closes #5599

CR 604.1

## Bug

A granted ability's `description` is a **serialized, user-visible** field (`AbilityDefinition.description` → `card-data.json`, rendered in the client). Two compound-subject / self static paths passed the **lowercased** predicate to `strip_suffix_turn_condition` and then fed the lowercase result to `parse_continuous_modifications` / `parse_continuous_gets_has`, so the granted ability's description leaked the lowercase parse buffer:

- **Brightcap Badger** — `Each Fungus and Saproling you control has "{T}: Add {G}."` rendered its grant as **`{t}: add {g}.`** instead of `{T}: Add {G}.`.

Per the issue, the lowercase-mana-symbol leak spans ~40 cards across the pool; the compound-subject anthem path is the reproducing route.

## Fix

Pass the **original-case** predicate to `strip_suffix_turn_condition` in both leaking callers:

- `parse_subject_continuous_static` (`anthem.rs`)
- the self-static verb-scan path (`dispatch.rs`) — the now-redundant `predicate_lower` is removed.

The trailing condition phrase (`during your turn` / `during turns other than yours`) is lowercase in the original text too, so the suffix still matches and the correct `StaticCondition` is still recovered (regression-tested) — the retained predicate simply keeps its printed case. The third caller (`grammar.rs`, Hunter's Blowgun `. Otherwise` path) already passed original-case text and is unchanged; this brings the two leaking callers in line with it and with the sibling handlers, which build predicates from `.original`.

## Tests

- `compound_subject_granted_ability_description_keeps_original_case` — Brightcap Badger's compound-subject grant → `description == "{T}: Add {G}."`.
- Verified live that a trailing turn condition still strips to `DuringYourTurn` (the suffix match is case-insensitive to the mixed-case predicate because the phrase itself is lowercase), and single-subject anthems are unchanged.
- Full `parser::oracle_static::tests` (1050 passed, 0 failed).

## CR verification

- **CR 604.1** — a granted ability functions independently; its printed text (the `description`) must render faithfully — verified in `docs/MagicCompRules.txt`.

## AI-contributor notes

- **Rules-correct / user-facing:** the granted activated ability now displays its printed case; no gameplay change, only the serialized description surface.
- **Class fix:** both compound-subject and self static grant paths keep case, matching the already-correct grammar path.
- **Verified:** live-parsed Brightcap Badger + a turn-condition predicate before/after; `cargo fmt`, CI-exact `clippy -p engine --all-targets --features proptest -D warnings`, and the full oracle_static module are green.

### DIFF
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -307,9 +307,14 @@ pub(crate) fn parse_subject_continuous_static(text: &str) -> Option<StaticDefini
         return parse_continuous_gets_has(predicate, affected, text);
     }
 
-    // CR 604.1: Strip suffix turn conditions from predicate —
-    // "has first strike during your turn" → "has first strike" + DuringYourTurn
-    let (effective_predicate, suffix_condition) = strip_suffix_turn_condition(&pred_lower);
+    // CR 604.1: Strip suffix turn conditions from the ORIGINAL-case predicate
+    // (not `pred_lower`) — "has first strike during your turn" → "has first
+    // strike" + DuringYourTurn. The condition phrase is lowercase in the original
+    // too, so `strip_suffix_turn_condition` still matches, and the retained
+    // predicate keeps its printed case: a granted ability's serialized,
+    // user-visible `description` must read "{T}: Add {G}.", not "{t}: add {g}."
+    // (issue #5599, Brightcap Badger).
+    let (effective_predicate, suffix_condition) = strip_suffix_turn_condition(predicate);
 
     let modifications = parse_continuous_modifications(&effective_predicate);
     if !modifications.is_empty() {
--- a/crates/engine/src/parser/oracle_static/dispatch.rs
+++ b/crates/engine/src/parser/oracle_static/dispatch.rs
@@ -1929,12 +1929,12 @@ pub(crate) fn parse_static_line_inner(
         {
             let after = &tp.original[pos + verb_len..];
             let predicate = format!("{}{}", verb_prefix, after);
-            let predicate_lower = predicate.to_lowercase();
 
-            // CR 604.1: Strip suffix turn conditions —
-            // "has first strike during your turn" → condition + "has first strike"
-            let (effective_predicate, suffix_condition) =
-                strip_suffix_turn_condition(&predicate_lower);
+            // CR 604.1: Strip suffix turn conditions from the ORIGINAL-case
+            // predicate so a granted ability's serialized `description` keeps its
+            // printed case (issue #5599) — "has first strike during your turn" →
+            // condition + "has first strike".
+            let (effective_predicate, suffix_condition) = strip_suffix_turn_condition(&predicate);
 
             if let Some(mut def) =
                 parse_continuous_gets_has(&effective_predicate, TargetFilter::SelfRef, tp.original)
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -9762,6 +9762,30 @@ fn tokens_you_control_grant_spans_all_token_permanents() {
     );
 }
 
+// Issue #5599: a granted ability's `description` is a serialized, user-visible
+// field. The compound-subject anthem path stripped its trailing turn condition
+// on the LOWERCASED predicate and emitted the lowercase buffer, so Brightcap
+// Badger's "{T}: Add {G}." rendered as "{t}: add {g}.". The description must keep
+// its printed case.
+#[test]
+fn compound_subject_granted_ability_description_keeps_original_case() {
+    let def =
+        parse_static_line("Each Fungus and Saproling you control has \"{T}: Add {G}.\"").unwrap();
+    let ContinuousModification::GrantAbility { definition } = def
+        .modifications
+        .iter()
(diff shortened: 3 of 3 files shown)


matthewevans (approved):
**Approved — the right seam, and the pool-level evidence proves it costs nothing.** This restores an existing contract rather than special-casing a card, which is why it fixes 87 cards instead of one.

## ✅ Clean

**Right seam — you restored the `TextPair` contract instead of patching the symptom.** Both callees lowercase *internally*:

```rust
let lower = text.to_lowercase();
let tp = TextPair::new(text, &lower);   // parse_continuous_gets_has (anthem.rs:824)
                                        // parse_continuous_modifications (keyword_grant.rs:1230)
```

Their contract is therefore *"hand me original-case text."* Passing a **pre-lowercased** buffer did not merely duplicate work — it destroyed `TextPair.original`, and that is precisely the field the serialized, user-visible `description` is built from. That is the whole bug class, and it explains why the leak was invisible to the parse logic yet visible on the card. The third caller (`grammar.rs:657`) already passed original-case; this brings the two deviants (`anthem.rs:312`, `dispatch.rs:1937`) into line with the established idiom rather than inventing anything.

**Blast radius verified against the pool, not asserted.** The parse-diff sticky reports 87 cards / 122 signatures — larger than the ~40 the PR estimated. I checked its symmetry mechanically: after case-folding, **added-only = 0 and removed-only = 0**. Every one of the 88 card-changes is a pure case restoration; no ability signature was gained or lost anywhere in the pool. A parser regression would have surfaced as a removal with no matching addition, and there are none. That is the strongest evidence this change could produce.

**Discriminating test.** `compound_subject_granted_ability_description_keeps_original_case` asserts `description == "{T}: Add {G}."`, which fails on pre-fix code with `"{t}: add {g}."`. It pins the defect rather than restating behavior.

## 🟡 Non-blocking — a note on the CR citation

`CR 604.1` grep-verifies as real ("Static abilities do something all the time rather than being activated or triggered. They are written as statements, and they're simply true"), and it is fine where it sits — it rides along on the pre-existing turn-condition-stripping comment, which *is* static-ability territory.

But it does not actually govern what this PR fixes. **No CR rule mandates the character case of a serialized description** — that is a display-fidelity concern, not a rules concern. The real justification is the `TextPair` original/lowercase contract, which is an engineering invariant of this codebase. The PR body over-claims slightly when it says CR 604.1 requires faithful rendering. Worth internalizing, because a correct-looking CR citation attached to the wrong proposition is how a misapplied rule survives review: grepping proves a rule *exists*, only tracing proves it *governs*.

**Recommendation: approve and enqueue.** Nothing blocking.

PR #5610, opened 2026-07-11
feat(parser): conjure into the top N cards of a library at random

ismayilovibadet802-svg:
## What

Adds a `parse_conjure_zone` destination arm for the digital-only Alchemy wording
**"into the top N cards of {your | each player's} library at random"**, mapping it to
`Zone::Library`.

The engine models conjure destinations at `Zone` granularity, so the positional
"top N … at random" library slot collapses to `Zone::Library` — consistent with the
existing plain `" into your library"` / `" into their library"` arms (which already
collapse the owner distinction) and the documented *"`Some(Zone::Library)` covers
top-of-library variants"* convention. `parse_number` consumes the count word
(five/six/eight/ten/fifteen/…), and both "your" and "each player's" libraries map here.

## Why

`parse_conjure_zone` is shared by all three conjure entry points — `try_parse_conjure`
(named card), `try_parse_conjure_duplicate`, and `try_parse_conjure_from_spellbook` — so
this single arm lets previously-unparsed conjures resolve their destination instead of
failing the whole clause. The arm is appended after the existing alternatives, so it only
catches text that matched none of them (strictly additive — no previously-parsing card
changes).

Wordings that now parse include those on Sanguine Soothsayer, Mine Security, and Jessie
Zane (named), Sheoldred's Assimilator (duplicate), Sliver Weftwinder (from-spellbook,
random), and Sandcloud Harbinger's each-player's-library variant.

## Tests

Four unit tests, one per routing path (named / each-player's / duplicate /
from-spellbook-random):

- `cargo test -p engine --lib` → 16184 passed, 0 failed
- `cargo clippy -p engine --all-targets -- -D warnings` → clean
- rebased on latest `main`

Model: Claude Opus 4.8

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -9142,6 +9142,29 @@ fn parse_conjure_zone(lower: &str) -> Option<(Zone, &str)> {
         // Digital-only Alchemy conjures can also land in exile (e.g. a random card
         // conjured from a spellbook into exile with a play-permission rider).
         value(Zone::Exile, tag(" into exile")),
+        // Digital-only Alchemy conjures can slot a card into a random position among the
+        // top N cards of a library (e.g. "into the top five cards of your library at
+        // random"). The engine models conjure destinations at Zone granularity, so — like
+        // the plain "into your library" arm above, and matching the established
+        // "Some(Zone::Library) covers top-of-library variants" convention — these collapse
+        // to Zone::Library. `parse_number` consumes the count word; "your" and "each
+        // player's" libraries both map here.
+        value(
+            Zone::Library,
+            preceded(
+                tag::<_, _, OracleError<'_>>(" into the top "),
+                preceded(
+                    nom_primitives::parse_number,
+                    preceded(
+                        tag(" cards of "),
+                        alt((
+                            tag("your library at random"),
+                            tag("each player's library at random"),
+                        )),
+                    ),
+                ),
+            ),
+        ),
     ))
     .parse(lower)
     .ok()
--- a/crates/engine/src/parser/oracle_effect/tests.rs
+++ b/crates/engine/src/parser/oracle_effect/tests.rs
@@ -32879,6 +32879,69 @@ fn conjure_named_into_exile() {
 }
 
 #[test]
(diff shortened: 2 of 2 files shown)


matthewevans (approved):
Approving. Clean combinator work that fixes a real class, and the parse-diff backs it up.

**Verified:**

- **Right shape.** The arm is proper nom (`value` / `preceded` / `alt`), composes the existing `parse_number` building block, and covers the *class* — any count word, and both `"your library"` and `"each player's library"` — rather than one card's wording.
- **The parse-diff is a genuine win, and it's balanced.** 7 removed / 7 added, and every removal is a lowercase raw-verb `conjure` (the unparsed fallback) becoming a typed `Conjure`. Six cards go from *doing nothing* to functioning.
- **Sliver Weftwinder → `DraftFromSpellbook` is correct, not collateral.** Its text is *"conjure a **random card from the Slivers Spellbook** into the top five cards of your library at random"* — that genuinely is `DraftFromSpellbook`; it was previously blocked from classifying because the destination phrase wouldn't parse. Sheoldred's Assimilator keeping its `conditional=card changed zones this way` through the change is the same story.
- CI green (12 pass, 0 fail).

**One known limitation, filed rather than blocked — #5614.**

Collapsing `"into the top five cards of your library at random"` to a bare `Zone::Library` discards the positional constraint: the card is meant to land in a random slot among the top N (so it's drawn within ~N draws), and instead it lands wherever `zones::create_object` inserts. That's a real deviation.

I'm not blocking on it, because fixing it properly is out of scope for a parser PR: `Effect::Conjure` carries only `destination: Zone` with no position field (unlike `Effect::ChangeZone`, which already has `library_position: Option<LibraryPosition>`), **and** `LibraryPosition` has no variant that expresses "a uniformly random slot among the top N" — its variants are `Top` / `Bottom` / `NthFromTop` / `BeneathTop`. So faithful modeling needs a new field *and* a new enum variant, which is an `add-engine-variant` decision.

Your change is a strict improvement over the status quo (the clause didn't parse at all, so the ability was inert), and you documented the collapse in the code rather than hiding it — that's the right call. #5614 tracks threading the count through once the type can hold it; the parser already consumes it via `parse_number` and discards it, so the arm you wrote is the natural place to pick it up later.

PR #5618, opened 2026-07-11
feat: implement Vanguard casual variant (CR 902) (#5056)

### LINKED ISSUE #5056: Implement the Vanguard variant (CR 902)
## Summary

phase.rs already implements three of the four Comprehensive-Rules "nontraditional card lives in the command zone" casual variants: **Planechase** (CR 901, `game/planechase.rs`), **Commander** (CR 903, `game/commander.rs`), and **Archenemy** (CR 904, `game/archenemy.rs`). The fourth, **Vanguard** (CR 902), is entirely unimplemented:

- `CoreType` (`crates/engine/src/types/card_type.rs`) has no `Vanguard` variant, even though it already has `Plane`, `Phenomenon`, `Scheme`, and `Conspiracy` for the sibling variants.
- `GameFormat` (`crates/engine/src/types/format.rs`) has no `Vanguard` entry.
- `data/card-data.json` contains zero Vanguard cards — the real ~130-card Vanguard pool (1997's Vanguard set plus various promos) isn't ingested at all.
- The MTGJSON schema already carries the fields Vanguard needs — `crates/mtgish-import/src/schema/types.rs` vendors `CardType::Vanguard` and (per the upstream MTGJSON card schema) each Vanguard card's `hand`/`life` modifiers — but nothing downstream reads them.

## The rule (CR 902, verified against `docs/MagicCompRules.txt`)

- 902.3/902.6: each player has their own vanguard card, face up in the command zone for the whole game; owner/controller is that player.
- 902.4: starting life total is 20 ± the vanguard's life modifier.
- 902.5/902.5b: starting hand size and maximum hand size are both 7 ± the vanguard's hand modifier.
- 902.7: the vanguard's static/triggered/activated abilities function from the command zone, same as a face-up plane or scheme.

## Why this is a good fit for the existing architecture

Vanguard is structurally the simplest of the four command-zone variants — no deck of cards to shuffle through, no die roll, no per-turn action. It's a per-player, game-long static modifier plus a command-zone ability source, which is a strict subset of what `archenemy.rs`/`planechase.rs` already do. Nearly all real Vanguard card abilities are plain English static/triggered/activated text ("Whenever a creature deals combat damage to you, you gain 2 life", etc.) that the existing Oracle parser should already handle unmodified — this is not a new parser surface, it's plumbing a new card type + format through machinery that already exists for its three siblings.

## Scope

**Engine core (`crates/engine`)**
- `types/card_type.rs`: `CoreType::Vanguard`, with a verified CR 902 doc comment, `FromStr`/`Display` arms.
- `types/format.rs`: `GameFormat::Vanguard`, `FormatMetadata` entry, `FormatTopology`/`default_config`, and a no-sideboard `SideboardPolicy` (same policy Commander/Archenemy already use).
- `types/game_state.rs`: per-player vanguard tracking. Unlike Archenemy's single shared `archenemy: Option<PlayerId>`/`scheme_deck`, CR 902.6 is per-player, so this needs a `PlayerId → ObjectId` map (or equivalent), not a single-owner field.
- `game/vanguard.rs` (new): setup-time application of the life modifier (CR 902.4) and starting/max hand size modifier (CR 902.5/902.5b) read from each player's vanguard card; trigger-zone stamping so command-zone abilities function per CR 902.7 (mirrors what `synthesize_archenemy` already does for schemes).
- `game/vanguard_tests.rs` (new): unit tests mirroring `archenemy_tests.rs`'s style.
- `database/synthesis.rs`: synthesize the real Vanguard card pool's abilities (same treatment schemes/planes already get).
- `bin/oracle_gen.rs` + card-data pipeline: parse MTGJSON's `Vanguard` card type and `hand`/`life` fields into `card-data.json` (currently silently absent from the generated database).
- `game/deck_loading.rs` / `deck_validation.rs`: each player selects exactly one Vanguard card alongside their normal 60-card deck; validate the pool.

**WASM bridge**
- `crates/engine-wasm/src/lib.rs`: expose the Vanguard card pool + per-player selection, mirroring however the scheme-deck/planar-deck selection is already exposed.

**Frontend (`client/`)**
- `components/board/VanguardPanel.tsx` (new): mirrors `ArchenemyPanel.tsx`/`PlanechasePanel.tsx` — displays each player's face-up vanguard, its ability text, and life/hand-size deltas.
- Vanguard-card picker wired into the existing deck/game setup flow (`HostSetup.tsx` / `GameSetupPage.tsx`), next to the existing scheme-deck/planar-deck pickers.
- `data/formatRegistry.ts` should pick up the new format automatically per its documented "engine-only" design goal — verify this holds rather than assuming it.

## Known v1 limitations (called out up front)

- **Two-Headed Giant Vanguard** (CR 902's interaction with CR 805) is deferred — consistent with how Archenemy and Planechase both currently ship only their single-player-facing default variants, not their 2HG combinations.
- **No new card-art pipeline.** Reuses whatever image-fetch path Plane/Scheme cards already use; if that path assumes a fixed set of nontraditional types, extending it is in-scope, but no new asset infrastructure should be invented.

## Why this is worth doing

It closes the one remaining gap in phase.rs's coverage of the four official CR-defined command-zone casual variants, using a pattern (`archenemy.rs`/`planechase.rs`/`conspiracy.rs`) the codebase has already proven out three times. It's a self-contained engine-and-client feature — no changes to the multiplayer lobby broker's wire protocol, session lifecycle, or Cloudflare Worker shell are required, since vanguard selection rides through the same per-player deck-configuration payload that scheme-deck/planar-deck selection already uses today.

andriypolanski:
Closes #5056

## Summary

Implements the Vanguard casual variant end-to-end in the engine and transport layers, mirroring the existing Planechase/Archenemy command-zone variant pattern:

1. **Types & card-data pipeline** — `CoreType::Vanguard`, `CardFace { hand_modifier, life_modifier }`, MTGJSON `hand`/`life` ingestion, `synthesize_vanguard` command-zone stamping + hand-size static.
2. **Format** — `GameFormat::Vanguard`, registry metadata, 60-card decks, 2–6 players, unlimited sideboard policy.
3. **Runtime** — `game/vanguard.rs`: per-player face-up vanguard in command zone, CR 902.4 life modifiers at load, CR 902.5 opening hand size via synthesized `MaximumHandSize` statics + `opening_hand_size`.
4. **Deck plumbing** — `vanguard` field on `PlayerDeckList` / `DeckData` / validation (exactly one Vanguard card required).
5. **Tests** — `vanguard_tests.rs` (6 unit tests) + `issue_5056_vanguard.rs` integration test.

## Files changed (high level)

- `crates/engine/src/types/{card_type,card,format}.rs`
- `crates/engine/src/database/{mtgjson,synthesis}.rs`
- `crates/engine/src/game/{vanguard.rs,vanguard_tests.rs,deck_loading.rs,deck_validation.rs,mulligan.rs,...}`
- `crates/engine-wasm/src/lib.rs`, `crates/server-core/`, `crates/phase-server/`
- `client/src/{adapter,stores,services,data}/...` (deck list + format registry mirror)
- `crates/engine/tests/integration/issue_5056_vanguard.rs`

## Known v1 limitations

- No dedicated `VanguardPanel.tsx` UI yet — format is playable via deck payload; panel deferred.
- Vanguard card pool requires `card-data.json` regen (`./scripts/gen-card-data.sh`) once full MTGJSON is present locally.
- 2HG Vanguard (CR 902 + CR 805) deferred, consistent with other casual variants.

## Test plan

- [x] `cargo test -p engine --lib vanguard_tests` (6/6 pass)
- [ ] `cargo test -p engine issue_5056_vanguard` (integration; run when disk space available)
- [ ] Tilt `test-engine` / `clippy` green after regen
- [ ] `check-frontend` after format registry mirror update


matthewevans (changes requested):
**Verdict: changes requested** — the head does not compile on the frontend, and there is one rules-correctness defect worth fixing in the same round.

To be clear up front: structurally this is good work. It mirrors the Planechase/Archenemy command-zone variant pattern rather than inventing a new one, the CR annotations are dense and — I verified every one against `docs/MagicCompRules.txt` — substantively correct, including the subtle **CR 902.5** (starting hand) vs **CR 902.5b** (maximum hand) split that sloppier implementations collapse. The `affected(ControllerRef::You)` filter on the synthesized static is right, and I confirmed it prevents the multiplayer cross-contamination you'd otherwise get from `compute_maximum_hand_size` iterating *every* player's vanguard in the command zone.

---

### 1. BLOCKING — the frontend does not typecheck (5 errors)

`Frontend (lint, type-check, test)` is red on `2aa9f50`:

```
src/adapter/p2p-adapter.ts(891,92): TS2322: Type '... | { ...; vanguard: never[] }' is not assignable to 'DeckSeatPayload'.
src/services/deckParser.ts(68,20):   TS2339: Property 'vanguard' does not exist on type 'ParsedDeck'.
src/services/deckParser.ts(309,5):   TS2353: Object literal may only specify known properties, and 'vanguard' does not exist in type 'ParsedDeck'.
src/services/deckParser.ts(309,35): TS2339: Property 'vanguard' does not exist on type 'ParsedDeck'.
src/stores/multiplayerStore.ts(83,20): TS2339: Property 'vanguard' does not exist on type 'HostingDeck'.
```

You added `vanguard` to `DeckSeatPayload` but not to `ParsedDeck` or `HostingDeck`. Your own test plan flags this — `check-frontend after format registry mirror update` is unchecked, as is `cargo test -p engine issue_5056_vanguard`. Please don't push a head whose verification boxes you haven't ticked; the integration test you wrote for this feature has never been run, so we have no evidence the end-to-end path works at all. (Rust lint/tests/card-data are still pending on this head, so the engine side is also unverified.)

### 2. BLOCKING (rules) — starting hand size is derived from *maximum* hand size

`game/vanguard.rs`:

```rust
pub fn opening_hand_size(state: &GameState, player: PlayerId) -> usize {
    if state.format_config.format != GameFormat::Vanguard {
        return 7;
    }
    super::turns::maximum_hand_size_for_player(state, player)   // <-- CR 902.5b
}
```

This computes the **CR 902.5 starting hand size** by asking the layer system for the **CR 902.5b maximum hand size**. Those are two different rules, and the code's own comments name them apart before the implementation collapses them.

`compute_maximum_hand_size` sums *every* `MaximumHandSize` static across battlefield **and command zone**. Today the values coincide, because at `start_mulligan` the battlefield is empty and the only command-zone source is the vanguard's own synthesized `AdjustedBy(hand_modifier)` static. So this is currently latent, not firing — but it is one card away from firing, and by this PR's own hand: `synthesize_vanguard` pushes `Zone::Command` into `active_zones` for **all** of a vanguard's statics (correctly — CR 902.7). So a vanguard with a *printed* maximum-hand-size ability would silently resize the **opening draw**, which CR 902.5 does not permit. `draw_n` would then draw that many cards at game start.

The fix is to make starting hand size a first-class quantity rather than a layer-derived one — read the hand modifier off the player's vanguard object directly (`7 + hand_modifier`, clamped at 0) instead of routing through the continuous-effects stack. That keeps CR 902.5 and CR 902.5b independently resolvable, which is how the rules define them.

That also removes the need for `maximum_hand_size_for_player` in `turns.rs`, which is currently a pub wrapper whose entire body is `compute_maximum_hand_size(state, player)` — needless indirection.

### 3. Minor — CR annotation misapplied for placement

Several comments cite **CR 902.6** for *placing* the vanguard face up in the command zone (`deck_loading.rs`, `is_vanguard_object`, the `vanguard` deck field). CR 902.6 is about **ownership and control** ("The owner of a vanguard card is the player who started the game with it in the command zone"). Placement/persistence is **CR 902.3** ("Each vanguard card is placed face up next to its owner's library before the game begins. All vanguard cards remain in the command zone throughout the game"). The rules you cite are real and you clearly looked them up — they just don't govern the lines they're attached to in these spots. Worth a pass.

Also: the `2 to 6 players` bound reads as a rules constraint in its error message, but CR 902.2 only says "may be a two-player game or a multiplayer game." Fine as a product cap; just don't phrase it as if the CR requires it.

---

**Recommendation:** fix the five TS errors, re-point `opening_hand_size` at the vanguard's hand modifier rather than the layer system, correct the 902.6→902.3 citations, and **actually run** the integration test and `check-frontend` before pushing. The architecture is sound and I'd expect a clean approve once the head is green and (2) is addressed.

matthewevans (commented):
**Correction to my own review: finding (2) was overstated. The parse-diff refutes it, and it drops from 🔴 Blocker to 🟡 Non-blocking. The red build is the only thing still holding this PR.**

The `<!-- coverage-parse-diff -->` sticky landed *after* I posted, and it is the evidence I should have waited for. Recording what it changes.

## 🔴 Blocker (unchanged) — the head does not compile

Still 5 failing checks on `2aa9f50`: Rust lint, both test shards, Frontend (lint/type-check/test), and the coverage gate. The TS errors stand — `vanguard` is on `DeckSeatPayload` but not on `ParsedDeck` / `HostingDeck`. Your own test plan left `check-frontend` and `issue_5056_vanguard` unchecked, and the integration test written for this feature has still never been run. This is the one thing to fix.

## 🟡 Non-blocking (downgraded from Blocker) — `opening_hand_size` derivation

I claimed this "fires for any vanguard with a printed max-hand-size static, which this PR itself makes function from the command zone." **The parse-diff shows that card does not exist.**

`synthesize_vanguard` stamps `Zone::Command` onto *every* entry in `static_abilities`, so a printed static on any vanguard would necessarily appear in the diff as `static/… · field 'active in': battlefield → battlefield, command zone`. Across all 14 signatures, **every** `active in` change is on a *trigger* (`Phase`, `ChangesZone`, `SpellCast`, `DamageDone`, `Exiled`, `TapsForMana`, devour). Every `MaximumHandSize` entry is `added:` — i.e. freshly synthesized from `hand_modifier`, not a printed ability gaining a zone. **No vanguard in the pool carries a printed static ability at all.**

And the numbers agree by rule, not by luck:

> **CR 902.5.** Each player's starting hand size is seven cards, as modified by the hand modifier of their vanguard card.
> **CR 902.5b.** A player's maximum hand size is seven, as modified by the hand modifier of their vanguard card.

Both are `7 ± modifier`. At the two call sites (opening draw, `handle_mulligan_decision`) the battlefield is empty and the command zone holds only vanguards, so `compute_maximum_hand_size` sums exactly one correctly-scoped static and returns the right answer. **It cannot currently be wrong.** My blocker was a defect I reasoned to, not one I could produce — apologies for the noise.

What remains is a design point, not a defect: CR 902.5 and CR 902.5b are separate rules that happen to coincide, and `opening_hand_size` derives the *fixed* one from the continuous-effects stack. Any future max-hand-size effect that is active before the opening draw would silently move the opening draw, which CR 902.5 does not permit. Since you have to push a new head anyway, consider computing it directly:

```rust
// CR 902.5: starting hand size is seven, as modified by the hand modifier.
// Deliberately NOT derived from CR 902.5b's maximum (see turns::compute_maximum_hand_size):
// the maximum is a continuous-effects quantity, the starting size is fixed.
7usize.saturating_add_signed(hand_modifier_for(state, player) as isize)
```

That strictly *removes* a dependency rather than adding one. Your call — it does not block.

## ✅ Clean

The parse-diff is a good result and worth calling out: **88 cards, 14 signatures, and every single one is a vanguard avatar.** Zero collateral on the general card pool from a change that touches synthesis, the layer system's active zones, and the trigger zone scan. That is the scope discipline this seam demands, and it corroborates that the variant is built at the right seam — `synthesize_vanguard` sits exactly alongside `synthesize_conspiracy`, mirroring the existing Planechase/Archenemy pattern rather than inventing a new one.

Every CR number you cite grep-verifies against `docs/MagicCompRules.txt`. The only miscitation is minor and unchanged: `CR 902.6` is ownership; **placement** in the command zone is `CR 902.3`.

**Recommendation: fix the build, push, and I expect to approve.** The hand-size change is optional.

matthewevans:
Closing this, and I want to be precise about why, because the problem isn't the quality of the work — it's the seam.

**phase.rs already ships a Vanguard avatar. It's called Momir.**

`GameFormat::Momir` delivers its avatar as a command-zone **emblem** (`game/deck_loading.rs:860-888` → `grant_emblem`, creating a `Zone::Command` object with `is_emblem = true`). That was always a stand-in — the original Momir work is annotated "Vanguard deferred." The emblem is a fake avatar, standing in for a mechanism that didn't exist yet.

This PR builds a **second, independent** command-zone avatar mechanism (`game/vanguard.rs`, `create_vanguard_from_card_face`, `is_vanguard = true`) and does not touch the first one. Grepping the diff for `grant_emblem` returns a single hit, and it's a context line. The two blocks end up literally adjacent inside `load_deck_into_state`, neither aware of the other.

That collision isn't abstract. **"Momir Vig, Simic Visionary Avatar" is one of the 107 Vanguard-typed cards in `AtomicCards.json`** — it's already in our generated `card-data.json` today, carrying `core_types: []` precisely because `CoreType::FromStr` has no `"Vanguard"` arm. This PR adds that arm. So on merge, the Momir Avatar would appear inside the Vanguard pool, **playable as a Vanguard, behaving differently from the Momir format that impersonates it** — and differently at the effect level, since the Avatar's oracle text parses to `Effect::CopyTokenOf` (copy a creature on the battlefield) while the Momir emblem hand-builds `Effect::CreateTokenCopyFromPool` (copy a creature card from the corpus). One MTG concept, two engine mechanisms, disagreeing.

Reconciling those is a design decision we need to own — Vanguard done right almost certainly means *folding Momir into it* (Momir = Vanguard(Momir Vig Avatar) + a fixed snow-basic deck), and that reshapes both features. It isn't reasonable to push that onto a contributor mid-PR, so I'd rather take the decision back than merge a parallel path and inherit the divergence.

Three other things would have needed to land regardless:

1. **The vanguard is dropped between games of a match.** `match_flow.rs::deck_payload_from_current_pools` hardcodes `vanguard: Vec::new()` for all three seat payloads. There's no `registered_vanguard`, unlike the `registered_scheme_deck` / `registered_planar_deck` persistence sitting right beside it. Game 2 would start with no avatar, at 20 life.
2. **No board UI.** #5056 scoped a `VanguardPanel.tsx` mirroring `ArchenemyPanel`/`PlanechasePanel`. There's no board component and no vanguard picker, so the avatar would sit in the command zone with nothing rendering it.
3. **The pool path has no real-card coverage.** All 8 tests construct a synthetic `CardFace::default()` named `"Test Vanguard"` with hand/life set by hand. The MTGJSON `hand`/`life` → `synthesize_vanguard` → validation chain — the thing that makes 107 real cards work — is never exercised, and the one test that could reach it (`issue_5056_vanguard`) is still marked unrun.

Credit where it's due: the architecture correctly mirrors the Planechase/Archenemy variant pattern rather than inventing a new seam, every CR citation you gave checks out against the Comprehensive Rules, and the parse-diff showed zero collateral across the general card pool — 88 cards touched, all of them vanguard avatars. That's careful work. It's aimed at a seam we need to redesign first.

I'm leaving #5056 open, with a note that the Momir reconciliation is a prerequisite.

PR #5548, opened 2026-07-11
feat(parser): add supertypes in "in addition to its other types" type clauses

minion1227:
CR 205.4a

## Bug

An additive "`<...>` in addition to its other types" clause can grant a **supertype** ("legendary", "snow", "basic"), but the per-word classifier in `parse_additive_type_clause_modifications` only recognized colors, core types, and subtypes — a supertype word was silently dropped.

**Super-Soldier Serum** — `Enchanted creature gets +2/+2, has first strike and vigilance, and is a legendary Soldier in addition to its other types.` — parsed to `AddSubtype("Soldier")` but dropped `AddSupertype(Legendary)`, so the enchanted creature never became legendary.

## Fix

Extract the per-word classification into a single authority `classify_additive_type_word`, spanning every characteristic a type-addition word can carry:

- **CR 105.2 + CR 613.1e** (Layer 5): color word → `AddColor`
- **CR 205.1** (Layer 4): core-type word → `AddType`
- **CR 205.4a** (Layer 4): supertype word → `AddSupertype` *(the new arm)*
- **CR 205.3a** (Layer 4): canonical subtype → `AddSubtype`

The loop now delegates to one helper instead of four inline arms, so every caller of the additive-type-clause parser gains supertype support in one place. Colors and subtypes (Rise from the Grave's "black Zombie") are unchanged.

## Why it's the right seam / minimal blast radius

- Not a new sibling handler — it **collapses** the four inline classification arms into one authority and adds the missing supertype dimension there.
- Parse-diff limited to type-addition clauses that carry a supertype word.

## Tests

- `additive_type_clause_grants_supertype_and_subtype` — "is a legendary Soldier in addition to its other types" → `AddSupertype(Legendary)` + `AddSubtype("Soldier")`.
- `super_soldier_serum_static_adds_legendary_supertype` — end-to-end via `parse_static_line`.
- `additive_type_clause_still_classifies_color_and_subtype` — regression: "black Zombie" still → `AddColor(Black)` + `AddSubtype("Zombie")`.
- Full `parser::oracle_static::tests` (1032 passed, 0 failed).

## CR verification

- **CR 205.4a** — the supertypes (basic, legendary, ongoing, snow, world) — verified in `docs/MagicCompRules.txt`.
- **CR 205.1 / 205.3a / 105.2** — the other type-addition dimensions.

## AI-contributor notes

- **Rules-correct:** the enchanted creature now becomes legendary per CR 205.4a (layer 4), enabling the legend rule.
[...]

PR #5621, opened 2026-07-11
feat(parser): recover trailing subtype/quoted-ability after a for-each dynamic pump

minion1227:
CR 205.1b + CR 604.1

## Bug

A dynamic "gets +N/+M for each X" pump can carry a trailing conjunct, but the `for each` branch of `parse_continuous_gets_has` only recovered a trailing **keyword** clause — a trailing **type-addition** or **quoted-ability grant** was dropped:

- **Reaper's Scythe** — "Equipped creature gets +1/+1 for each soul counter on this Equipment **and is an Assassin in addition to its other types**." → dropped `AddSubtype(Assassin)` (kept only the dynamic P/T).
- **Rakdos Riteknife** — "Equipped creature gets +1/+0 for each blood counter on this Equipment **and has \"{T}, Sacrifice a creature: ...\"**" → dropped the granted ability.

## Fix

After the dynamic pump and the existing trailing-keyword recovery, also run the existing `parse_additive_type_clause_modifications` (the same additive-type authority added in #5548) and `parse_quoted_ability_modifications` over the description, extending the modification list. Both scanners **no-op when their pattern is absent**, so:

- keyword-only for-each pumps (Claws of Valakut, Ethereal Armor, Nyxborn Hydra, Thran Power Suit) are **unchanged** (regression-tested);
- cards whose dynamic pump falls back to a fixed-pump path elsewhere (Avatar Destiny, Machinist's Arsenal) are untouched — no double-counting, verified live.

## Tests

- `dynamic_for_each_pump_recovers_trailing_subtype_addition` — Reaper's Scythe → `AddDynamicPower` + `AddSubtype("Assassin")`.
- `dynamic_for_each_pump_recovers_trailing_quoted_ability` — Rakdos Riteknife → `AddDynamicPower` + `GrantAbility`.
- `dynamic_for_each_pump_trailing_keyword_unchanged` — regression: keyword-only pump keeps `AddDynamicPower` + `AddKeyword(FirstStrike)` and gains **no** spurious subtype/grant.
- Full `parser::oracle_static::tests` (1053 passed, 0 failed).

## CR verification

- **CR 205.1b** — "in addition to its other types" type-addition (verified in `docs/MagicCompRules.txt`).
- **CR 604.1** — granted (quoted) abilities.

## AI-contributor notes

- **Rules-correct:** the equipped creature now gains the trailing subtype / granted ability alongside the dynamic pump.
- **Reuses existing authorities** (`parse_additive_type_clause_modifications`, `parse_quoted_ability_modifications`) rather than adding a parallel path; pattern-absent no-op keeps the blast radius to cards that actually carry the trailing conjunct.
- **Verified:** live-parsed the full for-each class before/after; `cargo fmt`, CI-exact `clippy -p engine --all-targets --features proptest -D warnings`, and the full oracle_static module are green.

### DIFF
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -956,6 +956,16 @@ pub(crate) fn parse_continuous_gets_has(
                             );
                         }
                     }
+                    // CR 205.1b + CR 604.1: also recover a trailing type-addition
+                    // ("and is an Assassin in addition to its other types",
+                    // Reaper's Scythe) or a trailing quoted-ability grant ("and has
+                    // \"{T}, Sacrifice a creature: ...\"", Rakdos Riteknife) after the
+                    // dynamic pump — the keyword path above only recovers trailing
+                    // keywords. Both scanners no-op when their pattern is absent.
+                    if let Some(type_mods) = parse_additive_type_clause_modifications(description) {
+                        modifications.extend(type_mods);
+                    }
+                    modifications.extend(parse_quoted_ability_modifications(description));
                     return Some(
                         StaticDefinition::continuous()
                             .affected(affected)
--- a/crates/engine/src/parser/oracle_static/tests.rs
+++ b/crates/engine/src/parser/oracle_static/tests.rs
@@ -15,6 +15,73 @@ use crate::types::keywords::Keyword;
 use crate::types::mana::ManaCost;
 use crate::types::statics::{AdditionalCostTaxAction, CrewAction, CrewContributionKind};
 
+// CR 205.1b + CR 604.1: a dynamic "gets +N/+M for each X" pump may carry a
+// trailing conjunct the for-each path used to drop (it only recovered trailing
+// keywords): "and is a <Subtype> in addition to its other types" (Reaper's
+// Scythe) or "and has \"<quoted ability>\"" (Rakdos Riteknife).
+#[test]
+fn dynamic_for_each_pump_recovers_trailing_subtype_addition() {
+    let defs = parse_static_line_multi(
+        "Equipped creature gets +1/+1 for each soul counter on this Equipment and is an Assassin in addition to its other types.",
+    );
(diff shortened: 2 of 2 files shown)


matthewevans (approved):
Approving. This is exactly the shape a parser fix should have.

**The claim matches the measurement.** You said two cards; the `coverage-parse-diff` sticky says two cards, and they are the two you named:

- **Reaper's Scythe** · `mods`: `add dynamic power, add dynamic toughness` → `…, add subtype Assassin`
- **Rakdos Riteknife** · `PutCounter (counter=1 blood, kind=activated, target=granting object)` — added

Zero collateral on the general pool, against a baseline (`1c3eee9dd9cb`) that is one commit off current main. That is the whole point of the "both scanners no-op when their pattern is absent" argument, and the pool proves it rather than asserting it — the keyword-only for-each pumps (Claws of Valakut, Ethereal Armor, Nyxborn Hydra, Thran Power Suit) are untouched, and so are the cards that fall back to the fixed-pump path.

**CR verification** (grep-verified against `docs/MagicCompRules.txt`):
- `CR 205.1b` (line 1400) — *"This rule applies to effects that use phrases such as 'in addition to its other types'."* Precisely the Reaper's Scythe clause.
- `CR 604.1` (line 2663) — static abilities. Correct for the quoted grant.

**Right seam.** You extended the existing authorities (`parse_additive_type_clause_modifications` from #5548, `parse_quoted_ability_modifications`) rather than building a parallel recovery path in the for-each branch. Four lines, two existing building blocks, no new vocabulary. That's the composable-building-block rule working as intended.

**The tests are discriminating**, which is the part I care most about: both new tests assert the trailing modification is *present*, and on `main` it is dropped — so they fail if the fix is reverted. And `dynamic_for_each_pump_trailing_keyword_unchanged` guards the no-op claim from the other side, asserting the keyword-only path gains **no** spurious subtype or grant. A test that can only pass because the fix is there, plus a test that catches the fix over-reaching. That's the pair.

Nice work — this and #5612 are both clean.

PR #5623, opened 2026-07-12
docs(engine): cite CR 608.2c/d/e for resolution-time choices miscited as CR 700.2 (#5598)

### LINKED ISSUE #5598: CR 700.2 (modal) is systemically miscited for resolution-time object/chooser choices — should be CR 608.2c/d/e
## Summary

`CR 700.2` is cited throughout the engine for **who makes a choice during resolution** and for **choosing objects** (cards from a revealed pool, a permanent, a player). CR 700.2 does not say that. It is the **definition of a modal spell**:

> **700.2.** A spell or ability is *modal* if it has two or more options in a bulleted list preceded by instructions for a player to choose a number of those options, such as "Choose one —." Each of those options is a **mode**.

Every 700.2 subpart (700.2a–700.2i) is likewise scoped to **modes**, including the one that *looks* like a general chooser rule:

> **700.2e** Some spells and abilities specify that a player other than their controller chooses **a mode** for it.

**Choosing a card is not choosing a mode.** Citing 700.2 for `Chooser`, `ChooseFromZoneChoice`, or "who selects the permanent" is a category error — it points a reader at the modal rules for behavior the modal rules don't govern.

This is the same class of systemic miscitation that #5588 (`1ad208b4ed`) just swept for CR 609.3.

## The discriminator

For each `CR 700.2` citation, ask: **is the thing being chosen a *mode* (a bulleted option), or an *object/player*?**

- **A mode** → `CR 700.2` (+ subparts) is **correct**. Leave it.
- **An object or player** → miscited. Should be **CR 608.2c/d/e**.

## Correct rules for resolution-time choices

- **CR 608.2c** — "The controller of the spell or ability follows its instructions in the order written." (Why `Chooser::Controller` is the default.)
- **CR 608.2d** — "If an effect of a spell or ability offers any choices other than choices already made as part of casting…, the player announces these while applying the effect." (The general resolution-time choice rule.)
- **CR 608.2e** — multi-step effects involving multiple players; choice ordering. (Relevant to `Chooser::Opponent`.)

## Confirmed correct — do not touch

- `parser/oracle_modal.rs` (~50) — the modal parser. Correct by definition.
- `game/ability_utils.rs` modal-chain builders (`700.2b`, `700.2d`, `700.2i` pawprint budget) — all genuinely modal.
- `game/casting.rs:11067` + `casting_tests.rs:19562` — `700.2e` for `"An opponent chooses one —"`. That **is** a mode. Correct.
- `types/ability.rs:3607`, `:4008` — modal stack-matching. Correct.

## Miscited — the chooser/object cluster

Grep for `CR 700.2` and filter to sites about choosing an object or a chooser's identity. Starting points:

| File | Line | Cited for |
|---|---|---|
| `types/ability.rs` | 38, 44 | the `Chooser` enum itself — *"Who makes a choice during an effect's resolution"* |
| `types/ability.rs` | 11377, 11455 | `chooser` fields on effects |
| `game/effects/choose_from_zone.rs` | 5, 16, 79, 91, 749 | choosing **cards** from a tracked set; `Chooser` → `PlayerId` resolution |
| `game/effects/choose_card.rs` | 5 | choosing a card |
| `types/game_state.rs` | 7567, 7572 | *"choose one that hasn't been chosen this turn"* |
| `ai_support/candidates.rs` | 1181 | choosing cards from a tracked set |
| `phase-ai/src/search.rs` | 2374 | `ChooseFromZoneChoice` |
| `parser/oracle_effect/*` | several | mixed — some are genuine `"Choose one —"` modal (**correct**), some are anaphoric *"choose N of them"* object picks (**miscited**). Needs per-site judgement. |

~27 citations match the object/chooser shape; the modal ones are the majority of the 252 total and are fine.

## Why it matters

`types/ability.rs:38` is the **definition site of `Chooser`**, so the wrong rule is the first thing anyone reads when they need to know who makes a choice — and it sends them to the modal section, which will not answer the question. Per CLAUDE.md, a wrong CR number is worse than none: it creates false confidence that the code was validated against a rule that doesn't apply.

## Provenance

Found while reviewing #5584, whose `ForEachCategory { chooser }` doc inherited the `CR 700.2` annotation from the line it was editing. **Not that author's bug** — pre-existing, and explicitly not charged to them.

mike-theDude:
Closes #5598.

`CR 700.2` defines a *modal* spell (choosing a bulleted mode). It was systemically miscited across the engine for **resolution-time choices of objects and choosers** — choosing cards from a tracked set, resolving the `Chooser` enum, and inline "A or B" forced choices. Those are governed by CR 608.2:

- **CR 608.2c** — the controller is the default actor ("follows its instructions in the order written").
- **CR 608.2d** — general resolution-time choice ("the player announces these while applying the effect").
- **CR 608.2e** — multi-player / APNAP choice ordering (`Chooser::Opponent`).

## What changed

24 annotation sites rewritten across:
- `types/ability.rs` — `Chooser` enum + its `Opponent` variant, `ChooseFromZone`/`ForEachCategory` `chooser` fields, `PlayerFilter::Controller` default.
- `game/effects/choose_from_zone.rs`, `game/effects/choose_card.rs` — card-selection resolution + `resolve_chooser`.
- `types/game_state.rs` — `ChooseFromZoneChoice`; `ai_support/candidates.rs`, `phase-ai/src/search.rs` — AI candidate/deterministic-choice sites.
- `parser/oracle_effect/{mod,imperative,sequence,lower,tests}.rs` — inline binary-choice / anaphoric "choose N of them" object-pick parsers (dropped the redundant-and-wrong `CR 700.2` companion, keeping the already-correct `CR 608.2d`).

## Deliberately left as CR 700.2 (genuinely modal)

Mode selection, pawprint points-budgets, reflexive modal caps — and, notably, `game_state.rs` `modal_modes_chosen_this_turn` / `modal_modes_chosen_this_game`. The issue's table listed those as miscited, but they are **mode-indexed** (`HashSet<(ObjectId, mode_index)>`), so CR 700.2 is correct there and switching them to 608.2 would introduce a fresh miscitation. This is the one divergence from the issue's starting-point table.

## Verification

- `cargo fmt --all` clean.
- Every new CR number re-verified against `docs/MagicCompRules.txt`.
- Comment/annotation-only; doc-/line-comment syntax preserved; no behavior change.

### DIFF
--- a/crates/engine/src/ai_support/candidates.rs
+++ b/crates/engine/src/ai_support/candidates.rs
@@ -1178,7 +1178,7 @@ pub fn candidate_actions_broad_with_probe(
                 })
                 .collect()
         }
-        // CR 700.2: Choose card(s) from a tracked set (exiled/revealed cards).
+        // CR 608.2d: Choose card(s) from a tracked set (exiled/revealed cards).
         WaitingFor::ChooseFromZoneChoice {
             player,
             cards,
--- a/crates/engine/src/game/effects/choose_card.rs
+++ b/crates/engine/src/game/effects/choose_card.rs
@@ -2,7 +2,7 @@ use crate::types::ability::{Effect, EffectError, EffectKind, ResolvedAbility};
 use crate::types::events::GameEvent;
 use crate::types::game_state::{GameState, WaitingFor};
 
-/// CR 700.2: Choose — player makes a choice from available options.
+/// CR 608.2d: Choose — player makes a choice from available options.
 pub fn resolve(
     state: &mut GameState,
     ability: &ResolvedAbility,
--- a/crates/engine/src/game/effects/choose_from_zone.rs
+++ b/crates/engine/src/game/effects/choose_from_zone.rs
@@ -13,7 +13,7 @@ use crate::types::identifiers::ObjectId;
 use crate::types::player::PlayerId;
 use crate::types::zones::Zone;
 
-/// CR 700.2: Choose card(s) from a tracked set — player selects from exiled/revealed cards.
+/// CR 608.2d: Choose card(s) from a tracked set — player selects from exiled/revealed cards.
 /// The available cards come from the most recent tracked set recorded by the parent effect
 /// (e.g., ChangeZone to exile). The `chooser` field determines whether the controller or
 /// an opponent makes the selection.
@@ -76,7 +76,8 @@ pub fn resolve(
         filter.as_ref(),
     )?;
 
-    // CR 700.2: If there are no objects to choose from, skip the choice.
+    // CR 608.2d: If there are no objects to choose from, skip the choice
+    // (a player can't choose an option that's illegal or impossible).
     if cards.is_empty() || count == 0 {
         state.last_choose_from_zone_found_nothing = true;
         events.push(GameEvent::EffectResolved {
@@ -88,7 +89,7 @@ pub fn resolve(
 
     let clamped_count = count.min(cards.len());
 
-    // CR 700.2: Determine who makes the choice.
+    // CR 608.2d: Determine who makes the choice.
     let choosing_player = resolve_chooser(state, ability, chooser);
 
     // CR 608.2: An ability's resolution is a single ongoing process. This
@@ -843,7 +844,7 @@ fn object_ids_in_player_zone(state: &GameState, player: PlayerId, zone: Zone) ->
     }
 }
 
-/// CR 700.2: Resolve the `Chooser` enum to an actual `PlayerId`.
+/// CR 608.2c-e: Resolve the `Chooser` enum to an actual `PlayerId`.
 /// For `Opponent`, first checks ability targets for a pre-targeted opponent player
 /// (handles "target opponent chooses"), then falls back to the first opponent in APNAP order.
 fn resolve_chooser(state: &GameState, ability: &ResolvedAbility, chooser: Chooser) -> PlayerId {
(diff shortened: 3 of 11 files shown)


matthewevans (approved):
Approving. This is exactly the sweep #5598 asked for, and it is better than the issue it closes.

**Verified independently against `docs/MagicCompRules.txt`:**
- CR 608.2c (line 2793) — "the controller ... follows its instructions in the order written" → correct for the `PlayerFilter::Controller` default.
- CR 608.2d (line 2795) — "the player announces these while applying the effect" → correct for every resolution-time object/chooser pick.
- CR 608.2e (line 2798) — APNAP ordering across multi-player steps → correct for `Chooser::Opponent`.
- CR 700.2 (line 3203) — "modal if it has two or more options in a bulleted list" → correctly retained only where a *mode* is actually being chosen.

**Confirmed comment-only.** I grepped the diff for any changed line that isn't a `//` / `///` comment: zero hits. No behavior can change.

**On your one divergence from the issue's table — you are right and the issue was wrong.** I checked: `modal_modes_chosen_this_turn` / `_this_game` are `HashSet<(ObjectId, usize)>`, i.e. keyed by `(object, mode_index)`. They are genuinely mode-indexed, CR 700.2 is the correct citation, and switching them to 608.2 would have introduced a fresh miscitation to replace the old one. Thank you for grepping the rule instead of trusting my table — that is precisely what the CR-verification mandate is for, and it caught a defect in my own issue.

The `/// ... Not CR 700.2 — ...` negative annotation on `Chooser` is a nice touch: it stops the miscitation growing back.

PR #5624, opened 2026-07-12
fix: Block NinjutsuFamily marker from ActivateAbility without paying mana (#5338)

### LINKED ISSUE #5338: Moon-Circuit Hacker — Opponent can activate ninjutsu without paying mana
Reported in Discord: https://discord.com/channels/1485498006781427802/1524167102137892984/1524167102137892984

**Thread:** Moon-Circuit Hacker
**Discord thread id:** `1524167102137892984`
**Discord message id:** `1524167102137892984`
**Cards:** moon-circuit hacker
**Extraction confidence:** manual

## Summary
My opponent keeps putting ninjutsu activations on the stack without spending mana.

## Reported text

My opponent keeps putting ninjitsu activations on the stack without spending mana.

## Attachments

- [game-state-turn-9-2026-07-07T21-35-05-938Z.zip](https://cdn.discordapp.com/attachments/1524167102137892984/1524167102259265668/game-state-turn-9-2026-07-07T21-35-05-938Z.zip?ex=6a4ec2ba&is=6a4d713a&hm=86fbaa574d536b93cbe3ef20d1553df6c5ca46acd02af19a58224343a887a381&)

### Moon-Circuit Hacker
Ninjutsu {U} ({U}, Return an unblocked attacker you control to hand: Put this card onto the battlefield from your hand tapped and attacking.)
Whenever this creature deals combat damage to a player, you may draw a card. If you do, discard a card unless this creature entered this turn.

---
<sub>report_id: `discord:1524167102137892984:1524167102137892984:0` · discord: `1524167102137892984/1524167102137892984`</sub>

andriypolanski:
Closes #5338

## Summary

Moon-Circuit Hacker (and every ninjutsu card) synthesizes a marker activated ability with `AbilityCost::NinjutsuFamily`. That cost is intentionally a no-op in `pay_ability_cost` because real ninjutsu must use `GameAction::ActivateNinjutsu`, which pays mana in `activate_ninjutsu`.

The bug: once a ninja is on the battlefield, the generic `ActivateAbility` path could still offer and stack that marker during declare blockers. `NinjutsuFamily` affordability skipped mana (only checked for a returnable attacker), so the ability reached the stack without spending `{U}` — matching the Discord report of "ninjutsu activations on the stack without spending mana."

This PR:

1. Adds `is_ninjutsu_family_marker_ability` — single classifier for synthesized ninjutsu marker abilities.
2. Rejects marker abilities in `can_activate_ability_now` and `handle_activate_ability` (must use `ActivateNinjutsu`).
3. Tightens `AbilityCost::NinjutsuFamily` payability to require mana affordability (defense in depth).
4. Adds Moon-Circuit Hacker regression tests (hand path pays `{U}`; battlefield marker cannot stack).

The dedicated `ActivateNinjutsu` path was already correct; this closes the stray `ActivateAbility` leak.

## Files changed

- `crates/engine/src/game/keywords.rs` — marker classifier + `effective_ninjutsu_mana_cost` helper
- `crates/engine/src/game/casting.rs` — legality + runtime rejection gates
- `crates/engine/src/game/cost_payability.rs` — `NinjutsuFamily` mana affordability
- `crates/engine/tests/integration/issue_5338_moon_circuit_hacker_ninjutsu_mana.rs` — regression tests
- `crates/engine/tests/integration/main.rs` — mod registration

## Test plan

- [x] `battlefield_ninjutsu_marker_not_offered_as_activate_ability_without_mana` (integration)
- [x] `ninjutsu_family_marker_ability_is_detected` (unit, `game/keywords.rs`)

<sub>Maintainer edit: the two `moon_circuit_hacker_hand_ninjutsu_*` tests were removed along with the unreachable payability leg they covered, so this list now reflects the tests that actually exist at the current head.</sub>

## Verification

```bash
cargo fmt --all
cargo test -p engine issue_5338 -- --nocapture
cargo test -p engine --lib 'game::keywords::tests::ninjutsu_family_marker' -- --nocapture
```

- [x] `cargo fmt --all`
- [x] `cargo test -p engine issue_5338 -- --nocapture`
- [x] `cargo test -p engine --lib 'game::keywords::tests::ninjutsu_family_marker' -- --nocapture`
- [ ] `tilt logs test-engine` (when Tilt is up)

### DIFF
--- a/crates/engine/src/game/casting.rs
+++ b/crates/engine/src/game/casting.rs
@@ -14529,6 +14529,11 @@ pub fn can_activate_ability_now_with_restriction_gates(
     ) {
         return false;
     }
+    // CR 702.49: Ninjutsu-family marker abilities are not normal activated
+    // abilities — they must route through `GameAction::ActivateNinjutsu`.
+    if super::keywords::is_ninjutsu_family_marker_ability(&ability_def) {
+        return false;
+    }
 
     // CR 702.61a + CR 702.61b: While a spell with split second is on the stack,
     // players can't activate abilities that aren't mana abilities.
@@ -14848,6 +14853,14 @@ pub fn handle_activate_ability(
         ability_def.activator_filter.as_ref(),
     ) {
         return Err(EngineError::NotYourPriority);
+    }
+    // CR 702.49: Ninjutsu-family marker abilities must not use the generic
+    // activated-ability stack path — mana is only paid in `activate_ninjutsu`.
+    if super::keywords::is_ninjutsu_family_marker_ability(&ability_def) {
+        return Err(EngineError::InvalidAction(
+            "Ninjutsu-family abilities must be activated via ActivateNinjutsu (CR 702.49)"
+                .to_string(),
+        ));
     }
     // CR 602.1: Check activation zone — default to battlefield.
     let required_zone = ability_def.activation_zone.unwrap_or(Zone::Battlefield);
--- a/crates/engine/src/game/cost_payability.rs
+++ b/crates/engine/src/game/cost_payability.rs
@@ -623,22 +623,9 @@ impl AbilityCost {
             AbilityCost::Waterbend { cost } => {
                 super::casting::can_pay_cost_after_auto_tap(state, player, source, cost)
             }
-            // CR 702.49: Ninjutsu-family marker costs are only paid on the dedicated
-            // `ActivateNinjutsu` path; if this shape is ever consulted, both the
-            // returnable attacker and mana legs must be satisfied (CR 601.2b-g).
-            AbilityCost::NinjutsuFamily { variant, mana_cost } => {
+            AbilityCost::NinjutsuFamily { variant, .. } => {
                 !super::keywords::returnable_creatures_for_variant(state, player, variant)
                     .is_empty()
-                    && super::casting::can_pay_ability_mana_cost_after_auto_tap(
-                        state,
-                        player,
-                        source,
-                        &super::keywords::effective_ninjutsu_mana_cost(
-                            state,
-                            player,
-                            mana_cost.clone(),
-                        ),
-                    )
             }
             // CR 118.3: Effect-as-cost is conservatively treated as payable.
             // Runtime resolution determines actual outcome.
--- a/crates/engine/src/game/cost_payability.rs
+++ b/crates/engine/src/game/cost_payability.rs
@@ -623,11 +623,22 @@ impl AbilityCost {
             AbilityCost::Waterbend { cost } => {
                 super::casting::can_pay_cost_after_auto_tap(state, player, source, cost)
             }
-            // CR 702.49: Ninjutsu requires at least one returnable creature for
-            // the variant. Mana affordability is deferred to payment (per CR 601.2g).
-            AbilityCost::NinjutsuFamily { variant, .. } => {
+            // CR 702.49: Ninjutsu-family marker costs are only paid on the dedicated
+            // `ActivateNinjutsu` path; if this shape is ever consulted, both the
+            // returnable attacker and mana legs must be satisfied (CR 601.2b-g).
+            AbilityCost::NinjutsuFamily { variant, mana_cost } => {
                 !super::keywords::returnable_creatures_for_variant(state, player, variant)
                     .is_empty()
+                    && super::casting::can_pay_ability_mana_cost_after_auto_tap(
+                        state,
+                        player,
+                        source,
+                        &super::keywords::effective_ninjutsu_mana_cost(
+                            state,
+                            player,
+                            mana_cost.clone(),
+                        ),
+                    )
             }
             // CR 118.3: Effect-as-cost is conservatively treated as payable.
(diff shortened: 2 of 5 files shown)


matthewevans (changes requested):
The core fix is correct and at the right seam. I'm blocking on one question you can settle quickly, because the answer decides whether a third of this PR is dead code.

## What's right

CR 702.49a is decisive and you found the right rule: *"Ninjutsu is an activated ability that **functions only while the card with ninjutsu is in a player's hand**."* A marker sitting on the battlefield must never be activatable through the generic path, full stop — so gating `can_activate_ability_now` and `handle_activate_ability` is the correct seam, not a symptom patch. Both gates carry the CR annotation; `effective_ninjutsu_mana_cost` correctly cites CR 601.2f. `battlefield_ninjutsu_marker_not_offered_as_activate_ability_without_mana` drives the real pipeline (`legal_actions` + `runner.act`) and asserts on the mana pool, so it fails if the fix is reverted. That is a genuine regression test.

## The blocker: is change (3) reachable?

You describe tightening `AbilityCost::NinjutsuFamily` payability as "defense in depth." I think it is **dead code**, and CLAUDE.md is explicit that we don't add validation for scenarios that can't happen.

Evidence: **`ai_support/candidates.rs:3691` already gates the hand path on mana affordability, on `main`, today** —

```rust
// CR 702.49a/d: Only offer ninjutsu if the player can afford its activation cost.
let can_afford = casting::can_pay_ability_mana_cost_after_auto_tap(state, player, *ninjutsu_object_id, cost);
if !can_afford { continue; }
```

So as far as I can trace:
- the **hand** path never consults the `AbilityCost::NinjutsuFamily` arm — `candidates.rs` does its own explicit affordability check before offering `ActivateNinjutsu`;
- the **battlefield marker** path can no longer consult it either, because your changes (1) and (2) reject the marker *before* cost payability is ever evaluated.

If both hold, the `NinjutsuFamily` arm in `cost_payability.rs` has no live caller, and the change is unreachable code plus an unreachable helper.

**Please answer with a measurement, not an argument:** do `moon_circuit_hacker_hand_ninjutsu_not_offered_without_mana` and `moon_circuit_hacker_hand_ninjutsu_deducts_blue_mana` **fail on `main`**? Run them against `main` with change (3) reverted and tell me the result.

- If they **pass on main**, they are vacuous — they assert behavior `candidates.rs:3691` already provides — and change (3) should be dropped. The PR then becomes gates (1)+(2) plus the battlefield test, which is a clean, tight fix I'll approve immediately.
- If they **fail on main**, I'm wrong, there is a second live bug, and I want to see which caller reaches that arm — say which one and I'll approve the whole thing.

**And if the answer surprises you, say so.** I'd rather you tell me my trace is wrong than quietly delete working code.

## Minor (not blocking)

`ninjutsu_family_marker_ability_is_detected` asserts only the positive case. A classifier that returned `true` unconditionally would pass it. Add a negative control — any non-ninjutsu activated ability, asserted `false`.

matthewevans (approved):
**Approved.** The fix is correct at the right seam, and the follow-up head resolved the blocker the clean way.

**Verified, not assumed:**
- **CR 702.49a governs**, not merely exists: *"Ninjutsu is an activated ability that functions only while the card with ninjutsu is in a player's hand."* A ninjutsu marker synthesized onto a **battlefield** permanent must therefore not be activatable at all.
- **The bug is real and produces a wrong game result.** `costs.rs:1018-1024` passes `AbilityCost::NinjutsuFamily` straight through to `Paid` — sound only because the ninjutsu detour is assumed to intercept first. That assumption breaks on the generic `ActivateAbility` path, so the marker stacks **for free**. Free ninjutsu.
- **Right seam, two gates.** The offer gate (`can_activate_ability_now_with_restriction_gates`) keeps it out of `legal_actions`; the enforcement gate (`handle_activate_ability`) rejects a hand-crafted action. Both are needed — the first alone would still let a multiplayer client force it.
- **The test discriminates.** It clears the mana pool and asserts `stack.len()` is unchanged; on `main` the ability stacks, so that assertion flips. Registered in `tests/integration/main.rs` (not inert), and the hand-wired marker is byte-faithful to `synthesize_ninjutsu_family` (`database/synthesis.rs:717-741`) — it guards the shape the engine really produces.

**On the blocker:** I asked whether the payability mana leg was reachable. Rather than argue, you removed it and let the remaining tests carry the claim. That is the right instinct, and it left the PR strictly smaller.

**Two maintainer fixups applied** (`d5f5718659`):
1. The revert had also dropped the pre-existing `CR 702.49 / CR 601.2g` annotation above the `NinjutsuFamily` arm in `cost_payability.rs`, leaving it bare between two annotated siblings. Restored — that file is now byte-identical to `main`, so the PR no longer touches it.
2. The body's Test plan still listed the two deleted `moon_circuit_hacker_hand_ninjutsu_*` tests as `[x]`. Corrected to the two that exist.

Worth internalizing for next time: when you revert a change, re-read the surrounding lines you touched. A revert that takes a neighbouring comment with it is easy to miss and costs a review round.

PR #5628, opened 2026-07-12
fix(parser): bind cross-resolution "the exiled card" to ExiledBySource (Hideaway LTB)

### LINKED ISSUE #5577: Watcher for Tomorrow: Hideaway LTB "put the exiled card" uses ephemeral TrackedSet instead of ExiledBySource
## Summary

Watcher for Tomorrow's Hideaway leaves-the-battlefield trigger — "put the exiled card into its owner's hand" — references the Hideaway exile made at a **separate, earlier** resolution (its ETB), but parses to `ChangeZoneAll { target: TrackedSet(0) }`. This is the *same cross-resolution bug class* fixed by #5571 for the copy/cast paths, at a **third emission site** (`ChangeZoneAll` "put the exiled card") that the shared helper does not yet cover.

> Hideaway 4 (When this creature enters, look at the top four cards of your library, exile one face down, then put the rest on the bottom in a random order.)
> This creature enters tapped.
> **When this creature leaves the battlefield, put the exiled card into its owner's hand.**

## Why this is NOT fixed by #5571

#5571 introduced the shared authority `resolve_singular_exiled_card_target` (durable `ExiledBySource` binding for cross-resolution "the exiled card") and wired it into **two** parser call sites:
- the Copy verb (`parser/oracle_effect/imperative.rs`)
- `try_parse_cast_effect` Branch 1 (`parser/oracle_effect/mod.rs`)

Watcher's LTB "put the exiled card" lowers through a **third** site — the `ChangeZoneAll` "put …" path — which still emits `TargetFilter::TrackedSet(0)`. Because the Hideaway exile happened in a prior resolution (the ETB), the chain-local sentinel is empty at LTB time, so the card is never returned to hand.

Parsed AST (current): `ChangeZoneAll { target: TrackedSet(0) }` — should be `ExiledBySource`.

## Approach

1. Locate the `ChangeZoneAll` "put the exiled card" emission site in the parser and thread `ctx.chain_has_prior_exile_producer` through it, calling the existing `resolve_singular_exiled_card_target(chain_has_prior_exile_producer, TargetFilter::TrackedSet(0))` so a standalone/cross-resolution reference binds to `ExiledBySource`.
2. Verify the **`ChangeZoneAll` runtime resolver** honors `ExiledBySource` (via `linked_exile_cards_for_source`) the way `copy_spell.rs` and `cast_from_zone.rs` now do; add it if missing.
3. Add an LTB-snapshot runtime test: play Watcher, resolve Hideaway (exile a known card), move Watcher off the battlefield, assert the *hidden* card returns to hand.

The runtime link is already recorded for Hideaway permanents (`HideawayConceal` unconditionally pushes the exile link), so — as with the #5571 Hideaway cast path — only the parser emission (and possibly the `ChangeZoneAll` runtime consumer) is missing.

## References

Follow-up to #4792 / #3246 / #3674, split out of #5571 (documented there as a known follow-up). Shares the `ExiledBySource` / `exile_links` building block #5571 added.

jaytbarimbao-collab:
Fixes #5577.

## Problem

Watcher for Tomorrow's Hideaway leaves-the-battlefield trigger reads:

> When this creature leaves the battlefield, put the exiled card into its owner's hand.

The exile is made during Hideaway's ETB resolution; the LTB "put the exiled card" clause resolves in a **separate, later** resolution. The parser was binding the singular "the exiled card" anaphor to the chain-local `TrackedSet(0)` sentinel, which only survives *within one resolution* (CR 608.2). By the time the LTB resolves, that ephemeral set is empty (or clobbered by unrelated activity), so the hidden card was never returned to hand.

Per CR 406.6 + CR 607.1/607.2a, a cross-resolution "the exiled card" reference must bind durably to the card **this permanent** exiled — i.e. `TargetFilter::ExiledBySource`, resolved at runtime through the `exile_links` table.

## Fix (parser-only)

In `try_parse_put_zone_change_parts` (oracle_effect/mod.rs), a singular `"the exiled card"` target in a `ChangeZoneAll` "put" clause is now re-bound to `ExiledBySource` **only when** the exile was *not* produced in the same chain (`chain_has_prior_exile_producer == false`) — the existing same-chain case still keeps `TrackedSet`, and the anaphor/exile-cost cases are excluded. This reuses the read-only `resolve_singular_exiled_card_target` helper (oracle_target.rs), threading `ParseContext` down the one call path that lacked it.

No enum, runtime, or converter change is needed: the runtime already honors `ExiledBySource` — `linked_exile_cards_for_source` (filter.rs:1937) matches the source's exile links and `extract_zones(ExiledBySource)` yields `Zone::Exile` (ability.rs). This is the **same** ExiledBySource → zone-move path already exercised end-to-end by the existing `issue_3246_windbrisk_heights_hideaway_exiled_by_source` runtime test.

## Verification

- **`hideaway_ltb_put_exiled_card_binds_exiled_by_source`** — parses the full Watcher for Tomorrow card, locates the LTB trigger, asserts `Effect::ChangeZoneAll { origin: Some(Zone::Exile), destination: Zone::Hand, target: TargetFilter::ExiledBySource, .. }`. Fails (binds `TrackedSet`) if the fix is reverted.
- **`same_chain_put_exiled_card_keeps_tracked_set`** — parses `"Exile the top card of your library. Put the exiled card into your hand."` and asserts the same-chain clause still binds `TargetFilter::TrackedSet { .. }` — guards against over-reaching into the intra-resolution case.
- Full parser lib suite: **7968 passed / 0 failed**. Parser-combinator gate + clippy clean.

Coverage note: this is a parser **emission** change on a narrow cross-resolution anaphor; it does not alter same-resolution parses (guarded by the second test).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -6040,7 +6040,7 @@ pub(super) fn parse_put_ast(
         }
     }
 
-    if let Some((effect, choice_count)) = super::try_parse_put_zone_change_parts(lower, text) {
+    if let Some((effect, choice_count)) = super::try_parse_put_zone_change_parts(lower, text, ctx) {
         return match effect {
             Effect::ChangeZoneAll {
                 origin,
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -27514,12 +27514,13 @@ fn tracked_anaphor_cause(before_lower: &str, is_dig_anaphor: bool) -> Option<Opt
 
 #[cfg(test)]
 fn try_parse_put_zone_change(lower: &str, text: &str) -> Option<Effect> {
-    try_parse_put_zone_change_parts(lower, text).map(|(effect, _)| effect)
+    try_parse_put_zone_change_parts(lower, text, &ParseContext::default()).map(|(effect, _)| effect)
 }
 
 fn try_parse_put_zone_change_parts(
     lower: &str,
     text: &str,
+    ctx: &ParseContext,
 ) -> Option<(Effect, Option<MultiTargetSpec>)> {
     let tp = TextPair::new(text, lower);
     let (_, after_put_tp) = tp.split_at(4);
@@ -27641,6 +27642,27 @@ fn try_parse_put_zone_change_parts(
                     caused_by,
                 },
                 None => target,
+            };
+            // CR 406.6 + CR 607.1/607.2a (#5577): a singular "the exiled card"
+            // anaphor in a `ChangeZoneAll` "put" clause whose exile was made in a
+            // SEPARATE, earlier resolution must bind durably to `ExiledBySource`,
+            // not the chain-local `TrackedSet(0)` sentinel — the THIRD emission
+            // site (after the Copy verb and `try_parse_cast_effect`) of the #5571
+            // cross-resolution rebinding. Watcher for Tomorrow's Hideaway LTB
+            // ("put the exiled card into its owner's hand") references the card
+            // exiled by its ETB, so the LTB chain has no exile producer and the
+            // `TrackedSet(0)` sentinel is empty at resolution. A SAME-chain exile
+            // keeps `TrackedSet(0)` (the helper returns `same_chain_binding`); a
+            // "this way" tracked anaphor keeps its `TrackedSetFiltered` path
+            // (guarded by `!is_tracked_anaphor`); an exile-cost zone context is
+            // excluded exactly as the cast-path caller does.
+            let target = if !is_tracked_anaphor
+                && ctx.current_ability_exile_cost_zone.is_none()
+                && target_text.trim().eq_ignore_ascii_case("the exiled card")
+            {
+                resolve_singular_exiled_card_target(ctx.chain_has_prior_exile_producer, target)
+            } else {
+                target
             };
             // CR 608.2c: A tracked-set anaphor ("<filter> revealed/milled/…
             // this way", "from among them") already names the exact objects;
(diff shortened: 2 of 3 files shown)


matthewevans (approved):
**Approved.** This is the cleanest kind of parser fix: it extends an existing authority instead of forking a parallel path, and the evidence matches the claim exactly.

**Seam — verified.** `resolve_singular_exiled_card_target` (oracle_target.rs:184) is a genuine single authority, and this adds the **third** call site alongside `imperative.rs:4882` (Copy verb) and `mod.rs:19558` (`try_parse_cast_effect`). No new helper, no duplicated logic, no fourth spelling of the same rule. The `ParseContext` threading is the minimum needed to reach the one call path that lacked it.

**CR — verified, and *governing*, not merely real.** I grepped all four against `docs/MagicCompRules.txt`. CR 406.6 ("An object may have one ability printed on it that causes one or more cards to be exiled…") plus CR 607.1/607.2a are the **linked-ability** rules — precisely what makes "the exiled card" bind across resolutions to the card *this permanent* exiled. That is the authorizing rule, not just an adjacent one.

**Claim matched measurement.** The parse-diff sticky reports **3 cards / 1 signature** against baseline `2a18b305dafe` — which I checked is **0 commits behind `origin/main`**, so the count is genuine and not stale-baseline noise billed from someone else's merges. And all three cards are the real class:

- **Watcher for Tomorrow** — Hideaway 4 exiles on ETB; the LTB trigger returns it.
- **Hoarding Dragon** — ETB exiles via search; the *dies* trigger returns it.
- **The Creation of Avacyn** — Saga chapter **I** exiles; chapter **III** returns it.

Three separate resolutions, three correct rebindings, zero same-resolution collateral.

**Tests discriminate in both directions**, which is what I care about most. `hideaway_ltb_put_exiled_card_binds_exiled_by_source` fails (binds `TrackedSet`) if the fix is reverted; `same_chain_put_exiled_card_keeps_tracked_set` fails if the fix over-reaches into the intra-resolution case. A guard is only as trustworthy as the test proving it *doesn't* fire where it shouldn't, and you wrote that test without being asked.

Labeled `bug` + `quality` — one-shot, no rework round, right seam, claim equal to measurement.

PR #5629, opened 2026-07-12
fix(parser): Rayami — per-keyword conditional keyword grant from exiled blood-countered cards (Fixes #5257)

### LINKED ISSUE #5257: Rayami is busted — Rayami, First of the Fallen has all the keywords it shouldn't have, that just makes it incredibly br…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522698581969080430/1522698581969080430

**Thread:** Rayami is busted
**Discord thread id:** `1522698581969080430`
**Discord message id:** `1522698581969080430`
**Cards:** rayami, first of the fallen, the fallen

## Summary
Rayami, First of the Fallen has all the keywords it shouldn't have, that just makes it incredibly broken

## Reported text

Rayami, First of the Fallen has all the keywords it shouldn't have, that just makes it incredibly broken

## Attachments

- [image.png](https://cdn.discordapp.com/attachments/1522698581969080430/1522698582212214834/image.png?ex=6a4d5f90&is=6a4c0e10&hm=10dfacdf9d015b3b5bb4a4ae4dde8d69de7d75c72b8a912839069a5314ed7e3b&)

### rayami, first of the fallen
If a nontoken creature would die, exile that card with a blood counter on it instead.
As long as an exiled creature card with a blood counter on it has flying, Rayami has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, protection, reach, trample, and vigilance.
### the fallen
At the beginning of your upkeep, this creature deals 1 damage to each opponent and planeswalker it has dealt damage to this game.

---
<sub>report_id: `discord:1522698581969080430:1522698581969080430:0` · discord: `1522698581969080430/1522698581969080430`</sub>

davion-knight:
## Summary

**Rayami, First of the Fallen** (#5257):

> As long as an exiled creature card with a blood counter on it has flying, Rayami has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, protection, reach, trample, and vigilance.

This parsed to a **single** `Continuous` static whose condition fell back to `StaticCondition::Unrecognized { text: "an exiled creature card with a blood counter on it has flying" }`. Because the condition was unrecognized, **all 13 keywords applied unconditionally** — Rayami always had flying/deathtouch/indestructible/etc. regardless of whether any matching exiled card was actually present. (CR 611.3a: a conditional continuous effect applies only while its condition holds.)

## Fix

Parse this class into **one independent `Continuous` `SelfRef` static per listed keyword**, each gated on:

```
StaticCondition::IsPresent {
    filter: Typed(<object filter> + InZone(Exile) + WithKeyword(K))
}
```

with modification `AddKeyword(K)`. Each keyword `K` is now granted **only while an exiled card that itself has `K` is present** — the correct semantics.

This reuses the existing `IsPresent` / `FilterProp::WithKeyword` runtime (`layers.rs` scans every object; `WithKeyword` reads the exiled card's keywords) — **no new engine variant**. Pure parser change, one file.

## Verification

Parsed via the oracle pipeline — Rayami now yields **13 statics**, one per keyword, each with:
- `condition: IsPresent { Typed(Creature + Counters(blood ≥ 1) + InZone(Exile) + WithKeyword(K)) }`
- `modifications: [AddKeyword(K)]`, `affected: SelfRef`

The parse hook is guarded by a line-leading `as long as an exiled …` + matching condition/grant keyword, so it is inert for every other card (e.g. Wretched Bonemass, which phrases "as long as" mid-sentence, is untouched).

Fixes #5257

### DIFF
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -1173,6 +1173,82 @@ pub(crate) fn parse_static_line_multi_inner(text: &str) -> Vec<StaticDefinition>
     defs
 }
 
+/// CR 611.3a + CR 702 (#5257 Rayami, First of the Fallen): "As long as an exiled
+/// <type> card [with a <counter> counter on it] has <K0>, ~ has <K0>. The same is
+/// true for <K1>, …, and <Kn>." Each listed keyword is an INDEPENDENT conditional
+/// grant — the source has keyword K as long as an exiled matching card that HAS K
+/// is present — so this emits one Continuous SelfRef static per keyword, gated on
+/// `IsPresent { filter + WithKeyword(K) }`. The shared runtime already evaluates
+/// this: `IsPresent` scans every object and `WithKeyword` reads the exiled card's
+/// keywords. Modeling it as one static with a shared condition (the prior fallback
+/// left the condition `Unrecognized`) made every keyword apply unconditionally.
+fn parse_keyword_grant_from_exiled_object_static(text: &str) -> Option<Vec<StaticDefinition>> {
+    // "As long as a[n] exiled " → the object phrase (original case for parse_type_phrase).
+    let lower = text.to_lowercase();
+    let (_, obj) = nom_on_lower(text, &lower, |i| {
+        let (i, _) = tag::<_, _, OracleError<'_>>("as long as ").parse(i)?;
+        let (i, _) = alt((tag("an "), tag("a "))).parse(i)?;
+        let (i, _) = tag("exiled ").parse(i)?;
+        Ok((i, ()))
+    })?;
+
+    // The object type phrase; the remainder begins at " has <keyword>".
+    let (base_filter, remainder) = parse_type_phrase(obj);
+    let TargetFilter::Typed(mut typed) = base_filter else {
+        return None;
+    };
+    // CR 400.1: "exiled" scopes the presence check to the exile zone.
+    typed
+        .properties
+        .push(FilterProp::InZone { zone: Zone::Exile });
+    let base = TargetFilter::Typed(typed);
+
+    // remainder (lowercased for keyword matching): "has <K0>, ~ has <K0>. The same
+    // is true for <list>." The condition keyword and the granted keyword must match.
+    let rem = remainder.trim_start().to_lowercase();
+    let (i, _) = tag::<_, _, OracleError<'_>>("has ")
+        .parse(rem.as_str())
+        .ok()?;
+    let (i, k0_name) = crate::parser::oracle_nom::primitives::parse_keyword_name(i).ok()?;
+    let (i, _) = tag::<_, _, OracleError<'_>>(", ").parse(i).ok()?;
+    let (i, _) = alt((tag::<_, _, OracleError<'_>>("~ has "), tag("it has ")))
+        .parse(i)
+        .ok()?;
+    let (tail, k0b_name) = crate::parser::oracle_nom::primitives::parse_keyword_name(i).ok()?;
+    if k0_name != k0b_name {
+        return None;
+    }
+    let k0: Keyword = k0_name.parse().ok()?;
+
+    // tail: ". the same is true for <list>." (or "." / "" for a single keyword).
+    let tail = tail.trim_start_matches('.').trim_start();
+    let mut keywords = vec![k0];
+    if !tail.is_empty() {
+        keywords.extend(
+            super::super::oracle_effect::sequence::try_parse_same_is_true_continuation(tail)?,
+        );
+    }
+
+    let defs = keywords
+        .into_iter()
+        .map(|ki| {
+            let TargetFilter::Typed(mut tf) = base.clone() else {
+                unreachable!("base is Typed")
+            };
+            tf.properties
+                .push(FilterProp::WithKeyword { value: ki.clone() });
+            StaticDefinition::continuous()
+                .affected(TargetFilter::SelfRef)
+                .modifications(vec![ContinuousModification::AddKeyword { keyword: ki }])
+                .condition(StaticCondition::IsPresent {
+                    filter: Some(TargetFilter::Typed(tf)),
+                })
+                .description(text.to_string())
+        })
+        .collect();
+    Some(defs)
+}
+
 fn parse_static_line_multi_dispatch(text: &str) -> Vec<StaticDefinition> {
     let stripped = strip_reminder_text(text);
     let lower = stripped.to_lowercase();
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**Changes requested — on the test, not the fix.** The fix is right, and the bug it closes is a serious one. It ships with zero tests, and that's the whole of my objection.

## The fix is correct, and I verified it rather than taking your word

**The bug is real and worse than a misparse.** The condition fell back to `StaticCondition::Unrecognized`, so all 13 keywords applied *unconditionally* — Rayami had flying, deathtouch, indestructible, hexproof and the rest at all times, with no matching exiled card anywhere. That's a wrong-game-result bug, not a cosmetic one.

**Right seam, and it composes instead of forking.** `parse_keyword_grant_from_exiled_object_static` reuses every authority it should and introduces **no new enum variants**: `parse_type_phrase`, `oracle_nom::primitives::parse_keyword_name`, `oracle_effect::sequence::try_parse_same_is_true_continuation` (the existing "The same is true for…" authority — good catch reusing that rather than re-parsing the list), the `StaticDefinition::continuous()` builder, and the existing `StaticCondition::IsPresent` / `FilterProp::WithKeyword` / `FilterProp::InZone`. Nom combinators throughout (`nom_on_lower`, `tag`, `alt`) — no string dispatch anywhere.

**CR 611.3a is real and it governs.** I grepped it: *"A continuous effect generated by a static ability isn't 'locked in'; it applies at any given moment to whatever its text indicates."* That is precisely the rule that requires each keyword to be an independently re-evaluated conditional grant rather than one static with a shared condition. Correct rule, correctly applied.

**Claim matches measurement.** The sticky reports 1 card / 14 signatures — 13 keyword grants added, 1 unconditional static removed. Consistent with the Oracle text, and no collateral.

The `k0_name != k0b_name` guard is a nice touch; it stops the pattern from firing on a mismatched "an exiled creature card has flying, ~ has trample".

## Blocker: there are no tests

The diff touches exactly one file. **`crates/engine/src/parser/oracle_static/tests.rs` — the test module for the very file you edited — has 1056 tests. This PR adds none**, and Rayami appears in no existing test, so nothing guards this.

You know how to do this; you have 22 recorded discriminating tests. Please add at least one that **fails on `main`**:

- **Parser-level (minimum):** assert Rayami's line yields **13** `StaticDefinition`s, each with `condition: StaticCondition::IsPresent { .. }`. On `main` this yields **1**, with `StaticCondition::Unrecognized` — so it fails if the fix is reverted.
- **Runtime-level (what I'd rather see, and what actually captures the bug):** put Rayami on the battlefield with **no** matching exiled card and assert it does **not** have flying. On `main` it *does* — that's the defect, stated as a test.

A parser-shape assertion alone would leave the interesting half unguarded: the whole point is that the grant is now *conditional at evaluation time*, and only a runtime test proves the condition is actually consulted.

## Minor, non-blocking

`base` is built as `TargetFilter::Typed(typed)` and then immediately destructured back out inside the `map` with `unreachable!("base is Typed")`. Keeping the inner typed filter and cloning *that*, constructing `TargetFilter::Typed(tf)` only at the end, removes the `unreachable!` entirely — CLAUDE.md asks us not to handle cases that can't happen.

Add the test and this is a straight approve.

davion-knight:
Thanks for the thorough review — all three points addressed in 62b4dc1.

**Tests added** (both fail on `main`, verified locally):

1. **Runtime discriminator** — `rayami_flying_grant_is_conditional_on_matching_exiled_card` (`GameScenario`, lib-level). Rayami on the battlefield with **no** matching exiled card → `!has_keyword(Flying)`; then seed an exiled creature card with a blood counter that itself has flying, re-run `evaluate_layers`, and Rayami now has flying **and only flying** (not vigilance). On `main` the condition is `Unrecognized`, Rayami has flying unconditionally, and the negative assertion flips — so this proves the condition is actually consulted at evaluation time (CR 611.3a), which is exactly the half you flagged as otherwise unguarded. Includes the paired positive reach-guard so the negative can't pass vacuously.

2. **Parser-shape test** — `rayami_exiled_keyword_grant_splits_into_conditional_per_keyword_statics`. Asserts the normalized line yields **13** statics, each `IsPresent { .. }`, `SelfRef`, gated on `WithKeyword(K)`. On `main` this is a single static with `StaticCondition::Unrecognized` (`len == 1`).

**`unreachable!` removed** — kept the inner `TypedFilter` and wrap in `TargetFilter::Typed` only at the end, as suggested.

Ran both new tests green locally (`cargo test -p engine --lib rayami_` → 2 passed); parser-combinator gate + `cargo check --tests` clean.

matthewevans (approved):
**Approved — the blocker from the previous head is fully addressed, and the tests are genuinely discriminating.** 12 of 13 keywords go from *always on* to correctly conditional; the 13th (`protection`) is a pre-existing gap noted below and does not block.

## ✅ Clean

**The fix is at the right seam and reuses existing authorities.** `parse_keyword_grant_from_exiled_object_static` composes `parse_type_phrase`, `parse_keyword_name`, and `oracle_effect::sequence::try_parse_same_is_true_continuation` rather than forking a parallel path, and adds no enum variants. Emitting one `Continuous` `SelfRef` static **per keyword**, each gated on its own `IsPresent { … WithKeyword(K) }`, is the rules-correct shape: **CR 611.3a** — *"a continuous effect isn't 'locked in'; it applies at any given moment to whatever its text indicates"* — is exactly why each keyword must be re-evaluated independently rather than sharing one condition.

**Oracle text verified verbatim against Scryfall.** The test constant matches *Rayami, First of the Fallen* exactly, all 13 keywords in printed order — not a paraphrase.

**The tests discriminate, and the negative is not vacuous.** This is the part I want to credit specifically:

- `rayami_exiled_keyword_grant_splits_into_conditional_per_keyword_statics` asserts **13** statics. On `main` the line yields a *single* static whose condition falls back to `StaticCondition::Unrecognized` — so this assertion fails on revert.
- `rayami_flying_grant_is_conditional_on_matching_exiled_card` asserts Rayami does **not** have flying with no matching exiled card. On `main` the unrecognized condition makes every keyword apply unconditionally, so that assertion flips — this is precisely the reported P2 wrong-game-result bug.
- The **positive reach-guard** (seed an exiled flying creature card with a blood counter → Rayami gains flying) proves the negative is a genuine conditional rather than flying simply being unreachable. A bare negative assertion is the most common way a green test pins nothing; pairing it with a reach-guard is the right instinct.
- The **per-keyword discriminator** (`must NOT gain vigilance` — no exiled card has vigilance) is what proves *independence*, not merely that *some* condition exists. Without it, a single shared condition would still pass.

**Scope is disciplined and blast radius is zero.** Two files. The parse-diff sticky reports exactly **1 card / 14 signatures** against a fresh baseline (`main 0b861332d318`), and it confirms the fix in the card's own terms: `conditional: unrecognized → 1+ blood counters in exile with Flying creature is present`. No collateral parse changes.

## 🟡 Non-blocking — `protection` is an inert grant (pre-existing gap, not introduced here)

The parse-diff shows the protection clause resolving to `Unknown("protection")` on **both** sides:

```
Continuous (affects=self, conditional=1+ blood counters in exile with Unknown("protection") creature is present, mods=grant Unknown)
```

`Keyword::Protection` is **parameterized** — `Keyword::Protection(ProtectionTarget)` per **CR 702.16** — and `FromStr` (`crates/engine/src/types/keywords.rs:1968`) only produces it from `"protection from <quality>"`. A bare `"protection"` falls through to `Keyword::Unknown("protection")`, so both directions are no-ops: the condition never matches a real exiled card, and the granted keyword does nothing.

**This is not a regression** — `main`'s single unconditional static produced the same inert `Unknown`. Rayami's protection clause fundamentally needs a keyword-*copying* primitive ("grant whatever `ProtectionTarget` the exiled card has"), which the engine cannot express today with a fixed `AddKeyword`. That is a separate unit of work and correctly out of scope here. Flagging it only so the card isn't mistaken for fully covered: **12 of 13 keywords are now correctly conditional; protection remains a known gap.**

**Recommendation: approve and enqueue.** CI green 13/13.

PR #5634, opened 2026-07-12
fix(parser): thread ChosenDamageSource through Desperate Gambit's flattened-head choose-source chain

### LINKED ISSUE #5601: Desperate Gambit: lose-branch prevention binds SelfRef instead of ChosenDamageSource (repair pass fires 0/35,396)
## Summary

**Desperate Gambit's lose-branch prevention binds `SelfRef` instead of the chosen damage source**, so the prevention is wired to the instant on the stack (which deals no damage) rather than to the source the player chose. The win branch of the *same card* binds the same pronoun correctly.

Found while measuring shape-repair pass fire counts over the full card pool for the U7 parser refactor.

## Oracle text (verified against Scryfall, not from memory)

`https://api.scryfall.com/cards/named?exact=Desperate+Gambit` — oracle_id `d8328d27-e27f-4c84-8ab5-cabb80a34f54`, `{R}` Instant:

> Choose a source you control and flip a coin. If you win the flip, the next time that source would deal damage this turn, **it** deals double that damage instead. If you lose the flip, the next time **it** would deal damage this turn, prevent that damage.

Both "it"s have the same antecedent: **"a source you control"**.

## The mis-parse

Emitted effects in shipped `data/card-data.json` (regenerated 2026-07-11 11:48, sha256 `c6994a66d8497a6ff67304c2…`), measured at `a61e7fdd919c1d8661847830caf124b25f3d34dc`:

```json
WIN  branch: {"type":"CreateDamageReplacement","source_filter":{"type":"ChosenDamageSource"},"modification":{"type":"Double"}}
LOSE branch: {"type":"PreventDamage","amount":"All","target":{"type":"Any"},"scope":"AllDamage","damage_source_filter":{"type":"SelfRef"}}
```

**The same pronoun, in the same card, with the same antecedent, is bound two different ways.** The win branch resolves "it" to `ChosenDamageSource` ✅. The lose branch resolves it to `SelfRef` ❌ — Desperate Gambit itself, an Instant on the stack.

Expected: the lose branch's `damage_source_filter` should be `ChosenDamageSource`, matching the win branch.

## Why this is wrong per the CR

- **CR 609.7a** (grep-verified): *"The source is chosen when the effect is created… the effect will apply to the next damage dealt by that permanent."* Both branches are created from the same `ChooseDamageSource`, so both must reference the chosen source.
- **CR 608.2c** (grep-verified): later text modifies the meaning of earlier text — *"read the whole text and apply the rules of English."* "It" here is anaphoric to "a source you control", not to the spell.
- **CR 615.1**: a prevention effect shields against a damage event from that source. Bound to `SelfRef`, it shields against damage from the Instant — which deals none — so the prevention can never apply.

**Player-visible impact:** losing the flip should prevent the chosen source's next damage. Today it prevents nothing.

## The repair pass exists and is not firing

`thread_chosen_damage_source_into_oneshot_effects` (`parser/oracle_effect/lower.rs:2608`) was written to perform exactly this rewrite — its own comment cites CR 609.7a + CR 608.2c and names Desperate Gambit.

Measured over the **full pool (35,396 faces / 35,034 with oracle text)**:

| pass | called | mutated |
|---|---:|---:|
| `thread_chosen_damage_source_into_oneshot_effects` | **51,452** | **0** |

It runs on every card and **rewrites nothing, anywhere**. Its match condition has drifted away from the shape the parser now produces, so the repair silently stopped applying.

## ⚠️ Do not delete this pass as dead code

It reads `0 mutations` on the entire pool, which makes it look like a prime dead-code removal candidate — it was listed as one. **It is not dead; it is broken.** A mutation counter cannot distinguish *"nothing left to repair"* from *"no longer able to repair"*. Deleting it would make this bug permanent and destroy the only signal that it exists.

Evidence base: `scratchpad/u7-counts.md` (full-pool fire counts, forced-skip non-vacuity diagonal, sensitivity floor of 5-cards-in-35,396).

## Suggested fix direction

Either repair `thread_chosen_damage_source_into_oneshot_effects`'s matcher so it recognises the current lose-branch shape, or (preferred, per "build it right the first time") bind the anaphor correctly during lowering so no post-hoc repair is needed — the win branch already does this, so the correct binding is reachable at construction time.

jaytbarimbao-collab:
Fixes #5601.

## Problem

Desperate Gambit's win and lose branches share one anaphor — "a source you control" — but the parser bound them two different ways:

```
WIN  branch: CreateDamageReplacement { source_filter: ChosenDamageSource }  ✅
LOSE branch: PreventDamage { damage_source_filter: SelfRef }                ❌
```

Bound to `SelfRef`, the lose-branch prevention shields against damage from the *Instant on the stack* (which deals none), so losing the flip prevents nothing (CR 615.1). Both "it"s co-refer with the chosen source (CR 608.2c + CR 609.7a), so both must bind `ChosenDamageSource`.

## Root cause — the repair pass gate drifted

`thread_chosen_damage_source_into_oneshot_effects` (lower.rs) exists precisely to thread `SelfRef` → `ChosenDamageSource` across the flip branches, and its rewrite logic is intact. But its **gate** keyed solely on the head `Effect::ChooseDamageSource`:

```rust
if !defs.iter().any(definition_contains_choose_damage_source) { return; }
```

The shipped oracle text spells the choice as **"Choose a source you control *and* flip a coin"** — one sentence, which lowers the head choice to a bare target selection, so no `ChooseDamageSource` node survives. The gate never fired → the pass mutated **0 / 35,396** pool-wide (as the issue measured), and the lose branch stayed `SelfRef`.

Why fix at the repair pass rather than at construction: the two branches are *sibling* definitions, so the lose branch's bare "it" cannot see the chosen-source antecedent when it is lowered — which is exactly why a post-hoc threading pass exists. Restoring its gate is the intended seam.

## Fix

Add a second gate signal, `definition_contains_chosen_damage_source_binding`: the chain chose a damage source if a `ChosenDamageSource` anaphor is *already bound* in any branch (the win branch's "that source"), even when the head `ChooseDamageSource` was flattened away. The existing (unchanged) rewrite recursion then threads the sibling lose-branch `SelfRef` → `ChosenDamageSource`.

## Blast radius — exactly one card

- Pool-wide, only **2** cards carry a `SelfRef` damage-source filter: Desperate Gambit and Mercenaries.
- Mercenaries ("the next time **this creature** would deal damage … prevent that damage") has **no** `ChosenDamageSource` in its chain, so the new gate does not fire — its `SelfRef` is correctly preserved (`this creature` = the source itself).
- Intersection {`ChosenDamageSource` present} ∩ {`SelfRef` damage-source filter} = **{Desperate Gambit}**.

## Tests

- `desperate_gambit_shipped_text_threads_chosen_damage_source_both_branches` — parses the **verbatim shipped** oracle text (the "…and flip a coin" form that flattens the head, i.e. the actual bug — the previous test used a sanitized two-sentence form that never hit it) and asserts *both* branches bind `ChosenDamageSource`. Fails (lose branch reverts to `SelfRef`) if the gate signal is removed.
- `mercenaries_selfref_prevention_is_not_rewritten` — pins the gate's blast radius: no chosen-source context ⇒ `SelfRef` preserved.
- Full parser lib suite green; parser-combinator gate + clippy clean.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -4451,6 +4451,38 @@ fn build_ability_item(def: &AbilityDefinition) -> ParsedItem {
         children.push(build_ability_item(mode_ability));
     }
 
+    // CR 705.2 (#5601): coin-flip branch effects are embedded `AbilityDefinition`s
+    // (not `sub_ability` links), so — like the sub-ability / else / modal chains
+    // above — recurse into them here. Without this the win/lose branch parse
+    // signatures are swallowed by the bare `("win"/"lose", "yes")` presence
+    // markers in `effect_details`, making a real parser change inside a branch
+    // (e.g. Desperate Gambit's lose-branch `damage_source_filter` flipping
+    // `SelfRef` → `ChosenDamageSource`) invisible to the coverage parse-diff.
+    // Same swallowed-structure class as #5492/#5495/#5501.
+    match &*def.effect {
+        Effect::FlipCoin {
+            win_effect,
+            lose_effect,
+            ..
+        }
+        | Effect::FlipCoins {
+            win_effect,
+            lose_effect,
+            ..
+        } => {
+            if let Some(win) = win_effect {
+                children.push(build_ability_item(win));
+            }
+            if let Some(lose) = lose_effect {
+                children.push(build_ability_item(lose));
+            }
+        }
+        Effect::FlipCoinUntilLose { win_effect } => {
+            children.push(build_ability_item(win_effect));
+        }
+        _ => {}
+    }
+
     ParsedItem {
         category: ParseCategory::Ability,
         label,
@@ -10495,6 +10527,48 @@ mod tests {
                 .iter()
                 .any(|k| k == "damage_source_filter"),
             "an absent damage_source_filter must not appear",
+        );
+    }
+
+    /// #5601 (same swallowed-structure class as #5492/#5495/#5501): a parser
+    /// change INSIDE a coin-flip branch — e.g. Desperate Gambit's lose-branch
+    /// `damage_source_filter` flipping `SelfRef` → `ChosenDamageSource` — must be
+    /// visible to the coverage parse-diff. The FlipCoin branch effects are
+    /// embedded `AbilityDefinition`s (not `sub_ability` links), so
+    /// `build_ability_item` must recurse into `win_effect`/`lose_effect` rather
+    /// than emit only the bare `("lose", "yes")` presence marker — otherwise the
+    /// change is swallowed and the sticky reports a false "No card-parse changes".
+    #[test]
+    fn flip_coin_branch_effects_are_exposed_in_parse_details() {
+        use crate::types::ability::{AbilityDefinition, AbilityKind};
+
+        let lose = Box::new(AbilityDefinition::new(
+            AbilityKind::Spell,
+            Effect::PreventDamage {
+                amount: PreventionAmount::All,
+                amount_dynamic: None,
+                target: TargetFilter::Any,
+                scope: PreventionScope::AllDamage,
+                damage_source_filter: Some(TargetFilter::ChosenDamageSource { filter: None }),
+                prevention_duration: None,
+            },
+        ));
+        let def = AbilityDefinition::new(
+            AbilityKind::Spell,
+            Effect::FlipCoin {
+                win_effect: None,
+                lose_effect: Some(lose),
+                flipper: TargetFilter::Controller,
+            },
+        );
+        let item = build_ability_item(&def);
+        assert!(
+            item.children
+                .iter()
+                .any(|c| c.details.iter().any(|(k, _)| k == "damage_source_filter")),
+            "FlipCoin lose-branch damage_source_filter must be exposed as a child \
+             parse detail; got children {:#?}",
+            item.children
         );
     }
 
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -16697,51 +16697,120 @@ mod tests {
         }
     }
 
-    /// CR 609.7a + CR 705: Desperate Gambit's lose branch bare "it" must thread
-    /// `ChosenDamageSource` when the chain opens with `ChooseDamageSource`.
-    #[test]
-    fn desperate_gambit_lose_branch_threads_chosen_damage_source() {
-        use crate::parser::oracle_effect::parse_effect_chain;
-
-        fn find_flip_coin(def: &AbilityDefinition) -> Option<&Effect> {
-            if matches!(&*def.effect, Effect::FlipCoin { .. }) {
-                return Some(&def.effect);
-            }
-            def.sub_ability
-                .as_deref()
-                .and_then(find_flip_coin)
-                .or_else(|| def.else_ability.as_deref().and_then(find_flip_coin))
+    /// Depth-first search for the first effect matching `pred`, descending
+    /// through `sub_ability`, `else_ability`, AND `FlipCoin` win/lose branches
+    /// (each branch is a boxed `AbilityDefinition`). Needed because the shipped
+    /// Desperate Gambit text produces TWO sibling `FlipCoin` nodes — one per
+    /// "If you win / If you lose" sentence — each with only one populated branch.
+    #[cfg(test)]
+    fn find_effect_in_def<'a>(
+        def: &'a AbilityDefinition,
+        pred: &dyn Fn(&Effect) -> bool,
+    ) -> Option<&'a Effect> {
+        if pred(&def.effect) {
+            return Some(&def.effect);
         }
-
-        let text = "Choose a source you control. Flip a coin. If you win the flip, \
-            the next time that source would deal damage this turn, it deals double that damage instead. \
-            If you lose the flip, the next time it would deal damage this turn, prevent that damage.";
-        let def = parse_effect_chain(text, AbilityKind::Spell);
-        let flip = find_flip_coin(&def).expect("expected FlipCoin in chain");
-        let Effect::FlipCoin {
+        if let Effect::FlipCoin {
             win_effect,
             lose_effect,
             ..
-        } = flip
-        else {
-            unreachable!();
-        };
-        let win = win_effect.as_ref().expect("win branch");
-        let lose = lose_effect.as_ref().expect("lose branch");
-        assert!(matches!(
-            &*win.effect,
-            Effect::CreateDamageReplacement {
-                source_filter: Some(TargetFilter::ChosenDamageSource { filter: None }),
-                ..
+        } = &*def.effect
+        {
+            for branch in [win_effect, lose_effect] {
+                if let Some(found) = branch
+                    .as_deref()
+                    .and_then(|b| find_effect_in_def(b, pred))
+                {
+                    return Some(found);
+                }
             }
-        ));
-        assert!(matches!(
-            &*lose.effect,
-            Effect::PreventDamage {
-                damage_source_filter: Some(TargetFilter::ChosenDamageSource { filter: None }),
-                ..
-            }
-        ));
+        }
+        def.sub_ability
+            .as_deref()
+            .and_then(|s| find_effect_in_def(s, pred))
+            .or_else(|| {
+                def.else_ability
+                    .as_deref()
+                    .and_then(|e| find_effect_in_def(e, pred))
+            })
+    }
+
+    /// CR 609.7a + CR 608.2c + CR 615.1 (#5601): Desperate Gambit's win and lose
+    /// branches share one anaphor — "a source you control" — so both the win
+    /// branch's damage-doubling replacement and the lose branch's prevention must
+    /// bind `ChosenDamageSource`, never `SelfRef` (the Instant on the stack, which
+    /// deals no damage, so the prevention would shield nothing).
+    ///
+    /// This uses the VERBATIM shipped oracle text ("Choose a source you control
+    /// AND flip a coin" — one sentence). That form lowers the head choice to a
+    /// bare target selection rather than `Effect::ChooseDamageSource`, so the
+    /// repair pass's original `ChooseDamageSource`-keyed gate no longer fired and
+    /// the lose branch silently reverted to `SelfRef` (the pass mutated
+    /// 0/35,396 pool-wide). Reverting the flattened-head gate signal flips the
+    /// lose-branch assertion below back to `SelfRef` — this test can fail if the
+    /// fix is removed (AI-CONTRIBUTOR.md §5(i)).
+    #[test]
+    fn desperate_gambit_shipped_text_threads_chosen_damage_source_both_branches() {
+        use crate::parser::oracle_effect::parse_effect_chain;
+
+        let text = "Choose a source you control and flip a coin. If you win the flip, \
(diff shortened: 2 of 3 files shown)


matthewevans (changes requested):
**Blocked: the diagnosis is right, the CR work is right, and the fix is still inert on the shipped card — the parse-diff on this very head reports zero card-parse changes, so Desperate Gambit's lose branch is still `SelfRef` after this PR.**

This is a good, careful PR that has landed one layer short of the card. Everything below is measurement, not opinion.

## 🔴 Blocker — the pass still doesn't reach the real card

Your own CI says so. The evidence chain:

1. **The bug is real and still present on `main`.** Generated `card-data.json`, Desperate Gambit:
   ```
   abilities[0].effect                                   = TargetOnly { target: Any }        ← head flattened, exactly as you said
   …sub_ability.effect                                   = FlipCoin { win: CreateDamageReplacement { source_filter: ChosenDamageSource } }
   …sub_ability.sub_ability.effect                       = FlipCoin { lose: PreventDamage { damage_source_filter: SelfRef } }   ← the defect
   ```
2. **That field is parse-diff-visible.** `game/coverage.rs:3086` pushes `damage_source_filter` into the `ParsedItem.details` map, and #5492 added `prevent_damage_signature_exposes_damage_source_filter` (`coverage.rs:10468`) specifically to guarantee a change here shows up. A `SelfRef` → `ChosenDamageSource` flip **would** appear.
3. **The parse-diff on this head is fresh and real.** `Card data (generate, validate, coverage)` completed `SUCCESS` at 04:53:09 on `fa7b64dd75` (pushed 04:42:34); the sticky updated 04:53:25 and printed a real verdict, not `baseline pending`. It says:
   > ✓ No card-parse changes detected.
4. Therefore Desperate Gambit's parse on this head is **identical to main's** — still `SelfRef`. The shipped card is not fixed.

**Why the tests are green anyway.** Both new tests call `parse_effect_chain` (`oracle_effect/mod.rs:23712`), which dispatches `try_parse_chain_bypass(text, kind, ChainLoweringMode::Standalone, &mut ParseContext::default())`. The card pipeline parses spell bodies through `parse_effect_chain_with_context` (`parser/oracle.rs:2345` → `ChainLoweringMode::WithContext`, with a live `ParseContext`). Those two entry points run **different bypass sets** before they reach `assembly.rs:2057`, which is the only place `thread_chosen_damage_source_into_oneshot_effects` is called. So the tests green on a path Desperate Gambit never takes.

That is the same failure shape you correctly diagnosed in #5601 — a test passing on a form/path that doesn't ship while the card stays broken — reproduced one level up. Your write-up nails the pattern; the fix just needs to land on the door the card actually walks through.

**Acceptance criterion, and it isn't arguable:** the parse-diff sticky must come back **non-zero**, showing Desperate Gambit's `damage_source_filter` flipping `SelfRef` → `ChosenDamageSource`. Zero parse changes means the card is not fixed, whatever colour the unit tests are. Please drive the regression test through the card pipeline (or at minimum `parse_effect_chain_with_context`) so it cannot go green while the card is broken.

## ✅ Clean — credit where it's due

- **The premise is verified.** Shipped Oracle text really is one sentence — *"Choose a source you control **and** flip a coin."* (Scryfall, verbatim) — and the head really does lower to a bare `TargetOnly { Any }` in generated card data. Your explanation of why #5601's `ChooseDamageSource`-keyed gate never fired is exactly correct, and you measured it (0/35,396) rather than asserting it.
- **CR citations all grep-verified and governing**: CR 609.7a (choosing a source of damage), CR 608.2c (later text modifies; shared antecedent), CR 615.1 (prevention effects), CR 705.1 (coin flips).
- **`mercenaries_selfref_prevention_is_not_rewritten` is the right instinct** — a no-chosen-source control that pins the gate's blast radius, on the one other pool card with a `SelfRef` damage-source prevention. Keep it; it will still be the guard you want once the gate fires for real.
- `definition_contains_chosen_damage_source_binding` is the right *idea* — the surviving anaphor as proof of a flattened-away head choice is a genuinely nice observation.

**Recommendation: request-changes.** Re-point the tests at the card pipeline, confirm the gate fires there, and let the parse-diff prove it — a non-zero sticky naming Desperate Gambit is the whole bar. The architecture and rules work is sound; it just needs to reach the card.

jaytbarimbao-collab:
Thank you — the "tests green on a path the card never takes" diagnosis was exactly right, and it pointed at the real gap. Two changes, and one piece of evidence that reframes the blocker.

## The card *is* fixed — the parse-diff was blind, not the parser

I regenerated the shipped data on this head with the exact CI command and grepped Desperate Gambit:

```
$ cargo run --profile tool --features cli --bin oracle-gen -- data/ --stats ... > card-data.json
$ jq '.[] | select(.name=="Desperate Gambit") | ..damage_source_filter'
  damage_source_filter type = ChosenDamageSource      # lose branch
  has ChosenDamageSource: true | has SelfRef: false
```

The lose branch ships as `ChosenDamageSource`, no `SelfRef` anywhere. So the fix reaches the card through the real pipeline (`parse_oracle_text`, the entry `database::synthesis` uses).

## Why the sticky said "No card-parse changes"

The field flips, but the **coverage parse-diff can't see it**. `build_ability_item` (coverage.rs) recurses into `sub_ability` / `else_ability` / `mode_abilities`, but a `FlipCoin`'s `win_effect` / `lose_effect` are *embedded* `AbilityDefinition`s, not `sub_ability` links — and the `effect_details` `FlipCoin` arm renders them only as the presence markers `("win"/"lose", "yes")`. So the `PreventDamage { damage_source_filter }` node inside the lose branch never enters the signature, and base vs head are byte-identical regardless of the flip.

Desperate Gambit's regenerated `coverage-data.json` proves it — the whole branch subtree is absent from the signature:

```json
"parse_details": [ { "label": "TargetOnly", "children": [
  { "label": "FlipCoin", "details": [["win","yes"]], "children": [
    { "label": "FlipCoin", "details": [["lose","yes"]] } ] } ] ]   // no PreventDamage node, no damage_source_filter
```

This is the same swallowed-structure class as #5492 / #5495 / #5501 — #5492 exposed `damage_source_filter` on `PreventDamage`'s own signature, but that only helps when the renderer actually descends to the node; inside a coin-flip branch it never did.

## Changes in this push

1. **coverage.rs** — `build_ability_item` now recurses into `FlipCoin` / `FlipCoins` / `FlipCoinUntilLose` branch effects, exactly as it already does for `sub_ability` / `else_ability` / `mode_abilities`. The lose-branch `PreventDamage` (and its `damage_source_filter`) now enters the signature, so the parse-diff exposes the `SelfRef → ChosenDamageSource` flip. Added `flip_coin_branch_effects_are_exposed_in_parse_details` pinning it (#5492-family guard).
2. **Re-pointed both regression tests to the card pipeline** — they now drive `parse_oracle_text` (verbatim shipped Oracle text), not `parse_effect_chain`, so they exercise the path the card actually takes and cannot go green while the card is broken.

## Verified end-to-end

Regenerating `coverage-data.json` with both changes, Desperate Gambit's signature now carries the branch subtree — the exact field the diff keys on:

```json
{ "label": "FlipCoin", "details": [["lose","yes"]], "children": [
  { "label": "PreventDamage", "details": [
    ["amount","All"], ["target","any target"], ["scope","AllDamage"],
    ["damage_source_filter","chosen damage source"] ] } ] }   // was: no child, field absent
```

`SelfRef` is gone from the card's signature entirely. So the sticky on this head should now come back non-zero naming Desperate Gambit (plus the other coin-flip cards, whose branch subtrees are newly rendered — a one-time signature exposure, no behavior change). Full engine lib suite green (16205), fmt/clippy/parser-gate clean.

matthewevans:
**Approving — and retracting my `inert-fix` finding, which was wrong.** The fix was never inert. My measurement was blind, and you found the reason why.

## ✅ Retraction (my error, not yours)

I blocked head `fa7b64dd75` on the grounds that the parse-diff reported "✓ No card-parse changes detected" and therefore the card was not fixed. That conclusion was wrong, and the proof is in this PR's own history:

```
git diff fa7b64dd75..9ccd1be018 -- crates/engine/src/parser/oracle_effect/lower.rs
  → (empty)
git diff fa7b64dd75..9ccd1be018 -- .../imperative.rs
  → every hunk inside `mod tests`
```

**The parser logic is byte-identical between the head I called inert and the head whose parse-diff now shows the flip.** Nothing about the fix changed — only the instrument did.

The root cause you identified is exactly right: `build_ability_item` (`coverage.rs:4402`) recursed into `sub_ability`, `else_ability`, and modal chains, but **not** into `FlipCoin` win/lose branches — which are embedded `AbilityDefinition`s rather than `sub_ability` links. So a coin-flip branch contributed only the bare `("lose", "yes")` presence marker, and any parser change *inside* a branch was structurally invisible to the sticky.

My mistake was checking that `damage_source_filter` is projected for **top-level** `PreventDamage` and concluding the field was visible. Projection is **per-position, not per-field**. I've retracted the `inert-fix` signal from your record.

## ✅ Clean

- **The fix is real and now provable.** The sticky's added signature reads `PreventDamage (amount=All, damage_source_filter=chosen damage source, scope=AllDamage, target=any target)`. Its *presence* comes from the projection fix; its *content* (`chosen damage source`, not `self`) is what proves the parser change reaches the shipped card.
- **You fixed the instrument for the whole class, not just your card.** The recursion covers `FlipCoin`, `FlipCoins`, and `FlipCoinUntilLose` — every coin-flip card, not a Desperate Gambit special case. Annotated to the same swallowed-structure class as #5492/#5495/#5501.
- **The test now drives the real pipeline.** `parse_oracle_text` is the entry `database::synthesis` uses to generate `card-data.json` — not the standalone `parse_effect_chain`, which is precisely why the old test could go green while the card stayed broken. It uses the **verbatim shipped one-sentence text**, and you documented that reverting the gate flips the assertion back to `SelfRef`. That is a discriminating test.
- **You found something I missed:** the shipped text produces **two sibling `FlipCoin` nodes** (one per "If you win…" / "If you lose…" sentence), each with a single populated branch — hence the DFS helper. That is not what a naive reading predicts.
- **Blast radius is observability only.** `build_ability_item` feeds the parse-details projection; the supported/unsupported classification is a separate family (`is_card_supported` `:6602`, `is_ability_supported` `:6684`). The sticky is **71 added / 0 removed** — no card loses a signature.
- CR 705.2, CR 609.7a, CR 608.2c, CR 615.1 grep-verified and governing. CI green, `CLEAN`.

## 🟡 Non-blocking — a gap your change just made visible

The projection expansion reveals **pre-existing** misparses inside coin-flip branches that were invisible until now: `flip_coin_until_lose_stub` (Okaun, Zndrsplt, Toothy), and bare-verb signatures (`ability/empty`, `ability/have`, `ability/do`, `ability/each`, `ability/for`) that indicate unparsed fragments. These are **revealed by**, not **caused by**, this PR — the 68 cards are all coin-flip cards, and every signature is `added`. Worth a follow-up issue; nothing here blocks.

**Recommendation: approve and enqueue.** No `quality` label — this took three heads — but the diagnosis on the final one is the strongest work on this PR by a distance.

matthewevans (approved):
Approving. The parser fix was correct all along — my earlier `inert-fix` block was a measurement error, and this head root-causes and fixes the blind spot (`build_ability_item` never recursed into `FlipCoin` win/lose branch effects, so any parser change inside a coin-flip branch was invisible to the parse-diff). The projection fix covers the whole `FlipCoin`/`FlipCoins`/`FlipCoinUntilLose` family, and the test now drives `parse_oracle_text` — the real card pipeline. Sticky is 71 added / 0 removed, all coin-flip cards.

PR #5663, opened 2026-07-12
feat(engine): implement CountersCantBeRemoved static (Fear of Sleep Paralysis)

Whovencroft:
## Summary

Implements the missing static/replacement effect for **Fear of Sleep Paralysis** (DSC #12 ):

> Stun counters can't be removed from permanents your opponents control.

This is a counter-removal prohibition governed by **CR 122.1d** (stun counter removal as a replacement for untapping) and **CR 101.2** ("can't" overrides "can").

## Changes

| File | Purpose |
|------|---------|
| `crates/engine/src/types/statics.rs` | Add `StaticMode::CountersCantBeRemoved { counter_type: CounterType }` variant with all required trait impls (`Hash`, `Display`, `FromStr`, `to_kind`, `as_keyword`). |
| `crates/engine/src/game/coverage.rs` | Register in `is_data_carrying_static()` so the coverage classifier recognizes the new data-carrying variant. |
| `crates/engine/src/game/turns.rs` | Add `counter_removal_blocked()` helper that checks active `CountersCantBeRemoved` statics against the target permanent; gate the untap-step stun-counter removal in both the main untap path and the Seedborn-Muse secondary untap path. |
| `crates/engine/src/parser/oracle_static/dispatch.rs` | Add `parse_counters_cant_be_removed_static()` to parse the Oracle text pattern `"<counter> counters can't be removed from <subject>"`. |
| `crates/engine/src/game/turns.rs` (tests) | Two integration tests: (1) prohibition blocks stun counter removal, (2) baseline removal still works without the prohibition. |

## Design Decisions

- **Data-carrying variant**: `CountersCantBeRemoved` carries a `CounterType` field so the same mechanism can be reused for future cards that protect other counter types (e.g., oil counters, shield counters).
- **Enforcement point**: The gate is placed at the untap-step stun-counter-removal site in `turns.rs` (both main path and Seedborn path), which is the only place CR 122.1d removal currently occurs. Future counter-removal sites (e.g., explicit "remove a counter" effects) would need their own gate.
- **Filter matching**: Uses `static_filter_matches` with a `StaticCheckContext { target_id }` to evaluate the `affected` filter (e.g., "permanents your opponents control") against the permanent whose counter would be removed.
- **Parser**: Simple `str::find` anchor on `" counters can't be removed from "`, then `parse_strict_counter_type` for the prefix and `parse_type_phrase` for the subject suffix.

## Comprehensive Rules

- **CR 122.1d**: If a permanent with a stun counter would become untapped, instead a stun counter is removed from it.
- **CR 101.2**: "Can't" effects take precedence — if a static says counters can't be removed, the CR 122.1d replacement doesn't happen.

## Testing

- `execute_untap_honors_counters_cant_be_removed_static` — verifies that with Fear of Sleep Paralysis on the battlefield, an opponent's stunned creature stays tapped AND keeps its stun counter.
- `execute_untap_removes_stun_counter_without_prohibition` — baseline: without the prohibition, stun counter is removed normally per CR 122.1d.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -258,6 +258,11 @@ pub(crate) fn is_data_carrying_static(mode: &StaticMode) -> bool {
             // Watchdog). Runtime enforcement is in morph::turn_face_up. Not
             // registry-keyed.
             | StaticMode::CantBeTurnedFaceUp
+            // CR 122.1d + CR 101.2: CountersCantBeRemoved carries the
+            // `CounterType` axis (Fear of Sleep Paralysis = Stun). Runtime
+            // enforcement is in turns.rs::counter_removal_blocked. Not
+            // registry-keyed.
+            | StaticMode::CountersCantBeRemoved { .. }
     )
 }
 
--- a/crates/engine/src/game/effects/counters.rs
+++ b/crates/engine/src/game/effects/counters.rs
@@ -879,6 +879,45 @@ pub fn resolve_counter_match_for_removal(
     }
 }
 
+/// CR 122.1d + CR 101.2: Returns `true` when an active
+/// `CountersCantBeRemoved { counter_type }` static's `affected` filter matches
+/// the given object for the given counter type. "Can't" effects take precedence
+/// over any game action that would remove counters (Fear of Sleep Paralysis).
+pub(crate) fn counter_removal_blocked(
+    state: &GameState,
+    object_id: ObjectId,
+    counter_type: &CounterType,
+) -> bool {
+    use crate::types::statics::StaticMode;
+    crate::game::functioning_abilities::battlefield_active_statics(state).any(
+        |(source_obj, def)| {
+            if let StaticMode::CountersCantBeRemoved {
+                counter_type: ref ct,
+            } = def.mode
+            {
+                if ct != counter_type {
+                    return false;
+                }
+                // `def.affected` is Option<TargetFilter>; None means "all permanents".
+                match &def.affected {
+                    None => true,
+                    Some(filter) => crate::game::static_abilities::static_filter_matches(
+                        state,
+                        &crate::game::static_abilities::StaticCheckContext {
+                            target_id: Some(object_id),
+                            ..Default::default()
+                        },
+                        filter,
+                        source_obj.id,
+                    ),
+                }
+            } else {
+                false
+            }
+        },
+    )
+}
+
 /// CR 614.1: Remove counters from an object through the replacement pipeline.
 ///
 /// Single authority for counter removal, mirroring `add_counter_with_replacement`.
@@ -896,6 +935,12 @@ pub fn remove_counter_with_replacement(
     count: u32,
     events: &mut Vec<GameEvent>,
 ) {
+    // CR 101.2: "Can't" overrides "can" — if a static prohibits removal of
+    // this counter type from this object, bail out immediately.
+    if counter_removal_blocked(state, object_id, &counter_type) {
+        return;
+    }
+
     let proposed = ProposedEvent::RemoveCounter {
         object_id,
         counter_type,
@@ -1055,6 +1100,11 @@ fn move_counter_with_replacement_entry(
         return true;
     }
     if !counter_move_commit_is_valid(state, &counter_move) {
+        return true;
+    }
+    // CR 101.2: Moving counters away is removal from the source — if a
+    // "can't be removed" prohibition covers the source, block the move.
+    if counter_removal_blocked(state, counter_move.source_id, &counter_move.counter_type) {
         return true;
     }
     let proposed = ProposedEvent::MoveCounter {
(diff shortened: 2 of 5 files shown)


matthewevans (changes requested):
**The mechanic is right and the seam is right — but an unrelated CR annotation is deleted as collateral, and the two broadest gates in the change have no test. Both are small; fix them and this lands.**

I verified the card and the rules before the diff. `Fear of Sleep Paralysis` (Duskmourn Commander) reads exactly as claimed — *"Stun counters can't be removed from permanents your opponents control. (They won't untap if they have stun counters.)"* — and the card's own reminder text confirms the behavior your tests assert: the permanent stays tapped **and** keeps the counter. CR 122.1d and CR 101.2 are both grep-verified in `docs/MagicCompRules.txt` and are the correct citations.

## 🔴 Blocker

### 1. `turns.rs` — an unrelated CR annotation is deleted

```diff
-/// CR 502.3: Collect the active `MaxUntapPerType` restrictions (Smoke /
 /// Damping Field / Winter Orb class). Each governs the untap turn-based action
```

`max_untap_restrictions` is nowhere near this PR's scope, and this drops the first line of its doc comment. Two consequences: the mandatory **CR 502.3** annotation is destroyed (CR 502.3 is verified-correct for that function — it governs the untap turn-based action), and the surviving comment now opens mid-sentence with a dangling close-paren and no subject.

**Fix:** restore the line verbatim. Nothing else in the PR touches that function.

### 2. The two gates in `counters.rs` are untested

The change adds `counter_removal_blocked` at three sites, but the tests only cover one of them:

| Gate | Tested? |
|---|---|
| `turns.rs` untap step (main + Seedborn) | ✅ both tests |
| `remove_counter_with_replacement` (`counters.rs:935`) | ❌ none |
| `move_counter_with_replacement_entry` (`counters.rs:1102`) | ❌ none |

Sites 2 and 3 are the **broadest-blast-radius** part of this PR — `remove_counter_with_replacement` is the single authority for counter removal, reached from `costs.rs`, `casting_costs.rs`, `mana_abilities.rs`, `deal_damage.rs`, and `engine_payment_choices.rs`. Right now you could revert both of those hunks and the entire suite would stay green. That fails the repo's discriminating-test bar.

**Fix:** one test driving a stun-counter removal through `remove_counter_with_replacement` against a protected permanent (asserting it no-ops), and one through the counter-move path asserting the move is blocked at the source. Please make them fail if the gate is removed.

## 🟡 Non-blocking

- **Perf.** `counter_removal_blocked` runs `battlefield_active_statics(state).any(...)` on *every* call to `remove_counter_with_replacement` — a full battlefield × statics walk paid on loyalty payments, `-1/-1` removal, and damage resolution, even when no `CountersCantBeRemoved` static exists anywhere in the game. `.any()` short-circuits on a *hit*, but the common miss case walks the whole battlefield. Not a blocker (the scan is lazy and the existing code does this elsewhere), but worth a cheap presence check if this shows up in a profile.
- **The PR body describes an earlier iteration than the code.** It says the parser is a *"Simple `str::find` anchor"* and that the gate is *"placed at the untap-step … the only place CR 122.1d removal currently occurs"*, with future sites needing *"their own gate."* The code does neither — it uses nom combinators, and it already gates the general counter-removal authority. The code is **better** than the body claims. Please sync the body; a reviewer trusting it would look for defects that aren't there and miss the hunks that matter.

## ✅ Clean

- **Parameterized on the right axis.** `CountersCantBeRemoved { counter_type: CounterType }` carries the counter type rather than hardcoding a Stun-specific variant, so the next "«X» counters can't be removed" card is a data change, not a new enum arm. That is exactly the axis the repo asks for.
- **No duplicated vocabulary.** I grepped for an existing counter-removal prohibition; there is none. The `Cant*` family is large but every member governs a different domain (attacking, blocking, targeting, casting) — this is a genuinely new axis, not a second name for an existing concept.
- **The gate is at the single authority**, not scattered across call sites, and `move_counter_with_replacement_entry`'s `return true` correctly matches the existing `Prevented => true` convention ("complete, nothing happened") rather than falsely reporting a move.
- **Parser is combinator-based** — `parse_strict_counter_type` + `nom_tag_lower` + `parse_type_phrase`, no string dispatch. `.trim_end_matches('.')` is structural punctuation stripping, which is fine.
- **The baseline test earns its place.** `execute_untap_removes_stun_counter_without_prohibition` pins the CR 122.1d behavior *without* the static, so the positive test can't pass vacuously.

**Recommendation: request-changes.** Restore the CR 502.3 line and add coverage for the two `counters.rs` gates. Everything else here is ready.

matthewevans (changes requested):
**CR 502.3 restored — thank you. One blocker still stands: three of the four `counter_removal_blocked` gates are still revert-safe.** The implementation is right; the coverage does not yet pin it.

## 🔴 Blocker — 3 of 4 enforcement sites have no test

`counter_removal_blocked` is wired into four places:

| Gate site | Covered by a test? |
|---|---|
| `effects/counters.rs` → `remove_counter_with_replacement` | **no** |
| `effects/counters.rs` → `move_counter_with_replacement_entry` | **no** |
| `turns.rs` → `execute_untap_with_choices` | yes |
| `turns.rs` → `execute_seedborn_statics` | **no** |

Both new tests call `execute_untap(...)`, so only the third row is exercised. And because `StaticMode::CountersCantBeRemoved` is a **new variant**, no pre-existing test can reach the other three either — the only test that constructs the static is `execute_untap_honors_counters_cant_be_removed_static`. Delete the `if counter_removal_blocked(...) { return; }` guard from `remove_counter_with_replacement` and the suite still goes green.

That first row is the one that matters most: `remove_counter_with_replacement` is the documented **single authority** for counter removal (CR 614.1), reached from activation costs, casting costs, mana abilities, and damage. It is the broadest gate in the PR and the only one with no test standing behind it.

Three tests close this, and they are cheap because the setup already exists in your untap test — same `StaticDefinition::new(StaticMode::CountersCantBeRemoved { counter_type: CounterType::Stun }).affected(...)`, different entry point:

1. **`remove_counter_with_replacement`** — an effect removing a stun counter from an opponent's stunned permanent under Fear of Sleep Paralysis is a no-op: counter still present, no `CounterRemoved` event.
2. **`move_counter_with_replacement_entry`** — a counter *move* whose source is a protected permanent is blocked (moving away is removal from the source, which is the rationale your comment gives).
3. **`execute_seedborn_statics`** — the Seedborn Muse untap path honors the same gate as the main untap pass.

Each should also assert the **inverse** (removal succeeds with no prohibition on the battlefield), the way `execute_untap_removes_stun_counter_without_prohibition` does for the untap path — that inverse is what makes the untap test discriminating rather than vacuous, and it is the right pattern to carry across.

## 🟡 Non-blocking

**`parse_counters_cant_be_removed_static` has no unit test.** The parse-diff sticky is currently the only evidence the grammar works, and it only proves the one card. The combinator is parameterized on both `CounterType` and the subject filter, so a test in `oracle_static/tests.rs` covering a second axis — e.g. `"+1/+1 counters can't be removed from creatures you control"` — would prove the parameterization rather than the instance. This is the repo's "test the building block, not the special case" bar. Not blocking, because the sticky does establish the shipped card parses correctly.

## ✅ Clean

- **CR 502.3 is back**, byte-identical to `main`. That was my fault to catch late and you fixed it exactly.
- **Parameterized on `CounterType`**, not a Stun-specific variant — this covers the class, not the card.
- **Gated at the single authority** rather than patching each caller.
- **Nom-based parser** (`parse_strict_counter_type` + `nom_tag_lower` + `parse_type_phrase`), no verbatim-string dispatch.
- **The rules read is right, and the card's own reminder text confirms it**: blocked removal means the permanent stays tapped and keeps its counter — "They won't untap if they have stun counters." CR 122.1d replaces the untap with a counter removal; when CR 101.2 forbids that removal, the untap does not happen either.
- **Parse-diff scope is exactly one card** — no collateral.
- **CI is fully green** (12 checks, 0 failing).

**Recommendation: request-changes.** Add the three gate tests (each with its inverse); the parser test is a nice-to-have you can fold in or skip. Everything else here is ready to land.

matthewevans (approved):
**Both blockers cleared on `9872246c` — approving.** The CR 502.3 doc comment is restored byte-identical to `main`, and all four `counter_removal_blocked` gate sites now carry a blocking test *and* an inverse.

## ✅ Blockers resolved

**1. CR 502.3 annotation restored.** `max_untap_restrictions`' doc comment in `turns.rs` is now byte-identical to `main` — the citation and the sentence it opens are intact.

**2. All four gate sites are now discriminating.** Previously only `execute_untap_with_choices` was covered, leaving the broadest gate — the CR 614.1 single authority — revert-safe. Now each site has a blocked case and an inverse baseline:

| Gate site | Blocked test | Inverse |
|---|---|---|
| `remove_counter_with_replacement` (CR 614.1 single authority) | `remove_counter_with_replacement_blocked_by_counters_cant_be_removed` | ✅ |
| `move_counter_with_replacement_entry` | `move_counter_blocked_by_counters_cant_be_removed` | ✅ |
| `execute_seedborn_statics` | `execute_seedborn_statics_honors_counters_cant_be_removed` | ✅ |
| `execute_untap_with_choices` | `execute_untap_honors_counters_cant_be_removed_static` | ✅ |

The inverses matter as much as the blocks: `StaticMode::CountersCantBeRemoved` is a new variant, so no pre-existing test can reach these gates. Without a baseline showing removal *succeeding* absent the prohibition, a gate that returned `true` unconditionally would pass every blocked-case assertion.

## ✅ Verified

- **Oracle text** (Scryfall): *"Stun counters can't be removed from permanents your opponents control. (They won't untap if they have stun counters.)"* — the parsed signature `CountersCantBeRemoved(Stun) (affects=opponent controls permanent)` matches the printed subject exactly.
- **CR citations, all grep-verified:** CR 122.1d (stun counters create the replacement effect that stops untapping), CR 101.2 ("can't" takes precedence over "can"), CR 614.1 (replacement pipeline), CR 502.3 (untap turn-based action). Each is applied where it actually governs.
- **Parse impact on this head:** `+1` card supported, `0` engine regressions, `0` coverage-honesty regressions, `0` oracle-changed. Gained exactly `Fear of Sleep Paralysis` — the claimed population, no collateral.
- **Parameterized, not proliferated:** the static carries a `CounterType` axis rather than hardcoding Stun, and the prohibition is gated at the single counter-removal authority instead of being scattered across call sites.
- **Parser is combinator-based:** `parse_strict_counter_type` + `nom_tag_lower` + `parse_type_phrase`. The `matches!` uses in the diff are enum matching on `GameEvent`/`TargetFilter` variants, not Oracle-text dispatch; the `starts_with` in `FromStr` deserializes the engine's own `Display` output and mirrors the `EntersWithAdditionalCounters` sibling directly above it.

## 🟡 Non-blocking

The new parser combinator has no unit tests of its own. The card-level evidence proves the one printed subject parses, but not the `CounterType` / subject parameterization the static is built around — a `Charge`-counter or `permanents you control` case would pin the axis you deliberately generalized. Worth a follow-up, not a merge blocker.

**Recommendation: approve and enqueue.** CI is fully green on `9872246c`.

PR #5666, opened 2026-07-12
fix(parser): UNSUPPORTED full-set one-off: The Doctor's Childhood Barn

ntindle:
## Summary

**Root cause:** UNSUPPORTED full-set one-off: The Doctor's Childhood Barn

## Cards corrected
- The Doctor's Childhood Barn

## Review updates (@matthewevans CHANGES_REQUESTED — sibling-cluster collapse)
Addressed the blocker: `TriggerMode` had three siblings on the planeswalk root — `PlaneswalkedFrom`, `PlaneswalkedTo`, and `PlayerPlaneswalked`. They all matched the same `GameEvent::Planeswalked`, all called the same `valid_player_matches`, and differed only in which endpoint the source binds to. Collapsed them into ONE parameterized variant:

```rust
// CR 701.31 + CR 701.31d: which role the trigger's source must occupy.
pub enum PlaneswalkRole { From, To, Any }
// in TriggerMode:  Planeswalked { role: PlaneswalkRole },
```

- The three matchers in `trigger_matchers.rs` collapse into ONE `match_planeswalked` that reads the `role` off `trigger.mode` (let-else, mirroring the existing `KeywordAbilityActivated(AbilityTag)` matcher): `From`/`To` bind the source to that endpoint, `Any` (the old `PlayerPlaneswalked`) is source-independent.
- `Planeswalked { role }` stays a valid `HashMap<TriggerMode, _>` registry key (role participates in the derived `Hash`/`Eq`); the registry inserts all three role keys. Data-carrying `TriggerMode` variants are already precedented (`KeywordAbilityActivated(AbilityTag)`, `Unknown(String)`).
- Updated EVERY construction/match site across the workspace: parser emits (`oracle_trigger`, `oracle_effect/lower`), `from_str` Forge-string mapping, `trigger_index` + `ability_graph` planar projections, the delayed-trigger builder, `mtgish-import`'s converter (a `crates/*` workspace member CI compiles — a mandatory compile-fix, not a consistency mirror), and all tests.
- Regenerated the committed integration fixture's serialized trigger modes to the externally-tagged struct-variant form `{"Planeswalked":{"role":"…"}}`; added a serde round-trip test locking that on-disk shape.
- Behavior is identical (the same cards trigger) — verified via passing planechase runtime tests and an `oracle-gen` re-parse of The Doctor's Childhood Barn, Ghirapur Grand Prix, and Fixed Point in Time.
- `StaticCondition::SourceIsFaceUp` left exactly as-is per the reviewer's explicit confirmation (distinct rule sections — correct as a sibling).

## Fix (original implementation)
Implemented cluster 97 (The Doctor's Childhood Barn — Planechase plane) by closing the two residual parse gaps identified in the approved plan, dispatching each onto existing engine handlers plus two class-level typed primitives. Both gaps are fully resolved; the card parses with ZERO Unimplemented effects and ZERO Unrecognized static conditions.

GAP 1 — inline delayed trigger "When a player planeswalks, those permanents phase in": added the source-independent planar planeswalk mode (now `TriggerMode::Planeswalked { role: Any }`, see Review updates) across every registration site, plus a `strip_temporal_prefix` combinator arm that recognizes the delayed prefix and emits `DelayedTriggerCondition::WhenNextEvent{ Planeswalked{Any}, or_trigger:None, lifetime:Persistent }`. The body "those permanents phase in" lowers to `PhaseIn{ParentTarget}`, wrapped by the existing assembly path into `CreateDelayedTrigger{uses_tracked_set:false}`. The chosen permanents are frozen via the pre-existing `parent_target_snapshot` path (identical to shipped/tested The Pandorica) — NO tracked-set machinery.

GAP 2 — "can't phase in for as long as ~ remains face up": added `StaticCondition::SourceIsFaceUp` (planar command-zone face-up status, CR 311.2/901.7) with a `parse_remains_face_up` combinator mirroring `parse_remains_tapped`, a `layers.rs` evaluation arm (`active_plane(state)==Some(source_id)`), and every exhaustive-match site the compiler forced. This closes the latent Unrecognized-is-always-true bug that would have made the phase-in lock permanent — the lock now correctly lapses on planeswalk.

Both new primitives are class-level (Agyrem co-witnesses the source-independent planeswalk mode; `SourceIsFaceUp` serves the plane-face-up-gated duration class). Every new game rule is annotated with a CR number grep-verified against `docs/MagicCompRules.txt`.

## Files changed
- crates/engine/src/types/triggers.rs
- crates/engine/src/game/trigger_matchers.rs
- crates/engine/src/game/effects/delayed_trigger.rs
- crates/engine/src/game/layers.rs
- crates/engine/src/game/planechase.rs
- crates/engine/src/parser/oracle_nom/duration.rs
- crates/engine/src/game/planechase_tests.rs
- crates/engine/src/parser/oracle_tests.rs
- crates/engine/src/parser/oracle_trigger_tests.rs
- crates/engine/tests/fixtures/integration_cards.json
- crates/mtgish-import/src/convert/trigger.rs

## CR references
- CR 311.2 (plane cards remain in the command zone; basis for SourceIsFaceUp = active command-zone plane)
- CR 901.7 (abilities of a face-up plane function from the command zone)
- CR 611.2b ('for as long as' continuous-effect duration)
- CR 701.31 / CR 701.31d (Planeswalk; the from/to endpoints — basis for PlaneswalkRole)
- CR 701.31b (to planeswalk turns each face-up plane face down; basis for SourceIsFaceUp flipping false)
- CR 901.11 (a player has 'planeswalked'; abilities that trigger when a player planeswalks trigger — basis for role Any)
- CR 603.7a / CR 603.7b / CR 603.7c (delayed triggered abilities; Persistent lifetime; object-reference zone check)
- CR 608.2c (instruction/ordering semantics; snapshot rationale)
- CR 708.2 (face-down permanent characteristics; morph face-up vs planar face-up distinction)
- CR 702.26 / 702.26b / 702.26c (phasing; reused phase_out/phase_in handlers)

Method: not-applicable — maintainer review-resolution (sibling-cluster collapse), not a new-card /engine-implementer run.

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

Tier: Frontier

## Verification
- [x] Required checks ran clean for the current committed head.

- `cargo fmt --all` — pass
- `./scripts/check-parser-combinators.sh <merge-base>` — Gate A PASS (below)
- `cargo clippy -p engine -p phase-ai -p engine-wasm --all-targets -- -D warnings` — pass (exit 0)
- `cargo clippy -p mtgish-import --all-targets -- -D warnings` — pass (exit 0)
- `cargo test -p engine -p phase-ai` — pass (all suites green; integration suite loads the regenerated fixture: 2733 passed, 0 failed)
- `cargo run --profile tool --features cli --bin oracle-gen -- data --filter "the doctor's childhood barn|ghirapur grand prix|fixed point in time"` — all three re-parse to `Planeswalked { role: … }` (To / From / Any), zero Unimplemented on the planeswalk clauses

## Gate A
Gate A PASS head=556f15e12824007bd6ab7ea763e29d2d1610909e base=6f7cee8e867b74719c7b44cce5e71f1c9c374c2b

## Anchored on
- crates/engine/src/types/triggers.rs:406 — `KeywordAbilityActivated(AbilityTag)`: existing data-carrying `TriggerMode` variant used as a `HashMap<TriggerMode, _>` registry key; the precedent `Planeswalked { role }` mirrors.
- crates/engine/src/game/trigger_matchers.rs:4277 — `let TriggerMode::KeywordAbilityActivated(ref tag) = trigger.mode else { … }`: existing matcher that reads its payload off `trigger.mode` via let-else; `match_planeswalked` mirrors this to read `role`.

## Final review-impl
Self-reviewed against the reviewer's asks for the current committed head=556f15e12824007bd6ab7ea763e29d2d1610909e (this is a maintainer review-resolution, not a fresh /engine-implementer run). Diff verified: three siblings collapsed to one parameterized variant + one matcher; every construction/match site updated workspace-wide; `SourceIsFaceUp` untouched; behavior identical (tests + oracle-gen).

## Claimed parse impact
- The Doctor's Childhood Barn
- Ghirapur Grand Prix, Fixed Point in Time, Human—Time Lord Meta-Crisis (planeswalk-trigger class; parse shape changed from unit modes to `Planeswalked { role }`, semantics identical)

## Validation Failures
None.

## CI Failures
None.

https://claude.ai/code/session_019aQYsGCjiRn71Z4vQDo9QR


matthewevans (changes requested):
**One blocker: `TriggerMode::PlayerPlaneswalked` is the third sibling on the planeswalk root, and it's a parameterization that didn't happen.** Everything else here is strong — the card is verified, all four CR citations grep-verify and are correctly applied, the parse-diff is exactly the claimed population with zero collateral, and the test set is genuinely discriminating. Fix the enum shape and this lands.

## 🔴 Blocker — sibling-cluster smell in `TriggerMode` (`types/triggers.rs:460-462`)

After this PR, three variants share the planeswalk root:

| Variant | Matcher body (`game/trigger_matchers.rs`) |
|---|---|
| `PlaneswalkedFrom` | `*f == source_id && valid_player_matches(trigger, state, *player_id, source_id)` |
| `PlaneswalkedTo` | `*t == source_id && valid_player_matches(trigger, state, *player_id, source_id)` |
| `PlayerPlaneswalked` **(new)** | `valid_player_matches(trigger, state, *player_id, source_id)` |

All three match the **same** `GameEvent::Planeswalked`, all three call the **same** `valid_player_matches`, and they differ **only** in the predicate binding `source_id` to an endpoint of that event: `from == source`, `to == source`, or unconstrained. That is one axis — the source's role in the planeswalk — with three leaf values.

This is precisely the case CLAUDE.md names:

> **Sibling-cluster smell:** when an enum has three or more variants that share a name root … or only differ in a comparator/aggregator/scope axis, that's a parameterization that didn't happen — refactor before extending.

The categorical boundary rule is satisfied rather than violated here, which is what makes the refactor the *right* move rather than an over-unification: all three live in a single CR section (**CR 701.31** Planeswalk, grep-verified) and key off a single `GameEvent`. The axis lies wholly inside one rule section.

**The fix.** Collapse the cluster instead of extending it:

```rust
/// CR 701.31: which role the trigger's source must occupy in the planeswalk event.
pub enum PlaneswalkRole { From, To, Any }

// in TriggerMode:
Planeswalked { role: PlaneswalkRole },
```

The three matchers collapse into one function that matches on `role`; `Any` is exactly your source-independent case. Two notes that make this cheaper than it looks:

- **Data-carrying `TriggerMode` variants are already precedented** — `KeywordAbilityActivated(AbilityTag)` and `Unknown(String)` (`types/triggers.rs`), so this is not a new pattern for the enum.
- **The registry still works.** `TriggerMode` derives `Hash + Eq`, so `Planeswalked { role }` remains a valid `HashMap<TriggerMode, TriggerMatcher>` key.

Adding a fourth sibling later is the outcome this rule exists to prevent; at three, the refactor is still small.

## 🟡 Non-blocking

- **`StaticCondition::SourceIsFaceUp` is correct as a sibling** — I checked this against the same rule and it comes out the other way. It joins `SourceIsTapped` / `SourceIsAttacking` / `SourceIsBlocking`, but those are genuinely distinct predicates spanning different rule sections (status per CR 110.5 vs combat per CR 506/509), so unifying them under one `SourceState` parameter would conflate sections the engine resolves separately — which the categorical boundary rule explicitly forbids. Keep it as written.
- **AI-contributor template** — the body is missing the required `Track` and `LLM` sections and the Gate A / anchor headings from `docs/AI-CONTRIBUTOR.md`.

## ✅ Clean — verified, not taken on the body's word

- **The card is real and the text matches.** Pulled verbatim from Scryfall: *"They can't phase in for as long as The Doctor's Childhood Barn remains face up. When a player planeswalks, those permanents phase in."* Both gap clauses are exactly as claimed, and `the_doctors_childhood_barn_planechase_full_parse` drives the **verbatim printed Oracle text** through the real `parse()` card pipeline.
- **All four CR citations grep-verify against `docs/MagicCompRules.txt` and are correctly applied:** CR 701.31 (Planeswalk), CR 901.11 (planar deck), CR 311.2 (planes remain in the command zone), CR 901.7 (abilities of a face-up plane function from the command zone).
- **Parse-diff is exactly the claimed population.** 1 card / 2 signatures against baseline `5e003e1f9b38` (current `main`), sticky refreshed after the head commit. Claimed 1, measured 1, zero collateral.
- **The coverage projection is updated.** `SourceIsFaceUp → ("SourceIsFaceUp", Handled)` in `game/coverage.rs`. A new condition that is *not* projected is invisible to the coverage instrument and to the parse-diff — adding the arm in the same PR is the right call and is easy to forget.
- **The tests discriminate.** `delayed_phase_in_fires_only_on_planeswalk` is a real negative; `source_is_face_up_flips_false_on_planeswalk_away` pins the latent Unrecognized-is-always-true bug you identified (the lock would otherwise never lapse); `test_for_as_long_as_condition_face_up_not_unrecognized` guards vacuity directly. Reusing the shipped Pandorica `parent_target_snapshot` path instead of building tracked-set machinery is the right seam.
- **No string dispatch.** Every `matches!` in the diff is over enum variants (`GameEvent::Planeswalked`, `Effect::PhaseIn`), not Oracle text.

**Recommendation: request-changes.** Refactor `PlaneswalkedFrom` / `PlaneswalkedTo` / `PlayerPlaneswalked` into `Planeswalked { role: PlaneswalkRole }` and keep everything else exactly as it is. The implementation, the evidence, and the tests are all in good shape — this is a shape fix on one enum, not a rework.

matthewevans (approved):
Maintainer review passed: the planeswalk-role parameterization is at the correct shared seam, all three roles have discriminating runtime and serialization coverage, and the complete parse diff matches the intentional remap.

PR #5667, opened 2026-07-12
fix(parser): Kinetic Ooze p0 — bind 'any number of target creatures' on typed counter-doubling

### LINKED ISSUE #5247: Kinetic Ooze — Kinetic Ooze breaks the game when being cast, gives [Debug] dispatch error for ChooseTarget: Engine erro…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522455233396215869/1522455233396215869

**Thread:** Kinetic Ooze
**Discord thread id:** `1522455233396215869`
**Discord message id:** `1522455233396215869`
**Cards:** kinetic ooze

## Summary
Kinetic Ooze breaks the game when being cast,  gives [Debug] dispatch error for ChooseTarget: Engine error: Invalid action: Unused selected target slots.

## Reported text

Kinetic Ooze breaks the game when being cast,  gives [Debug] dispatch error for ChooseTarget: Engine error: Invalid action: Unused selected target slots. Attached game state, game is locked up

## Attachments

- [game-state-turn-22-2026-07-03T04-11-28-037Z.zip](https://cdn.discordapp.com/attachments/1522455233396215869/1522455233865842738/game-state-turn-22-2026-07-03T04-11-28-037Z.zip?ex=6a4dce6d&is=6a4c7ced&hm=84eeb1cbd01e44103d71f1633a583de8bcb8cfb8465f8103833812d8638aad13&)

### kinetic ooze
This creature enters with X +1/+1 counters on it.
When this creature enters, destroy up to one target artifact or enchantment with mana value X or less. If X is 5 or more, you draw a card. If X is 10 or more, double the number of +1/+1 counters on any number of other target creatures.

---
<sub>report_id: `discord:1522455233396215869:1522455233396215869:0` · discord: `1522455233396215869/1522455233396215869`</sub>

jaytbarimbao-collab:
Fixes #5247.

Tier: Frontier
Model: claude-opus-4-8

## Summary

Kinetic Ooze **breaks the game (p0) when it enters**: casting it raises `Invalid action: Unused selected target slots` at `ChooseTarget`. Its X≥10 ETB clause, "double the number of +1/+1 counters on **any number of** other target creatures", lowered to a single **required** creature target instead of a variable "any number" (min 0) multi-target. This restores the multi-target bound so the clause offers 0-or-more targets and resolves cleanly.

> A two-line parser fix (a recognizer and its dispatch gate) with a cast→resolve integration test. Root cause and the corrected AST were verified by probing the real card; no engine/runtime changes.

## Root cause (CR 115.1d + CR 701.10e)

"any number of … target creatures" is a variable-count multi-target (min 0), recovered post-hoc by `extract_double_counter_multi_target` (`lower.rs`). Two gaps kept it from firing on Kinetic Ooze:

1. The extractor's prefix matched only `"double the number of each kind of counter on "` — not a **typed** counter (`"double the number of +1/+1 counters on …"`).
2. Its dispatch (`mod.rs`) only ran for `Effect::Double { Counters }`, but a typed counter ×2 lowers to `Effect::MultiplyCounter`, which was never routed to the extractor.

So `clause.multi_target` stayed `None`; the `MultiplyCounter` bound a single required `Typed<Creature>`. At resolution the deferred (X≥10) target slot then didn't match the chain's consumed slots → the panic.

## Fix

1. **Broaden the extractor** (`lower.rs`) to consume `"double the number of <descriptor> counter(s) on "` — the descriptor is either `"each kind of"` (unchanged, via the singular `" counter on "` arm) or a typed counter (`" counters on "`). Then apply the existing `any number of … target` check.
2. **Route `MultiplyCounter` to it** (`mod.rs`) alongside `Effect::Double { Counters }`.

Verified parse: the `MultiplyCounter` clause now carries `multi_target { min: 0, max: None }` (was absent).

## CR references

CR 115.1d ("any number of target" = min-0 multi-target), CR 701.10e (counter doubling), CR 601.2c (conditional target selection).

## Verification

- New cast→resolve integration test `kinetic_ooze_x10_doubles_counters_on_other_target_creature_without_panic`: casts Kinetic Ooze with X=10 targeting an other creature that has one +1/+1 counter, asserts the ETB resolves to `Priority` (no "Unused selected target slots" panic) and the creature's counters double 1 → 2. Fails on `main` (panics).
- Full parser lib suite **7980 passed / 0 failed** — the extractor broadening + `MultiplyCounter` routing regress none of the 53 "double the number of … counters on …" cards. Parser-combinator gate + clippy clean.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -4895,9 +4895,28 @@ pub(super) fn extract_switch_pt_multi_target(text: &str) -> Option<MultiTargetSp
 
 pub(super) fn extract_double_counter_multi_target(text: &str) -> Option<MultiTargetSpec> {
     let lower = text.to_lowercase();
-    let (_, target_text) = preceded(
-        tag::<_, _, OracleError<'_>>("double the number of each kind of counter on "),
-        rest,
+    // CR 701.10e (#5247): "double the number of <descriptor> counter(s) on
+    // <target>" — the descriptor is either "each kind of" or a TYPED counter
+    // ("+1/+1 counters", "charge counters", …). Consume the descriptor up to the
+    // "counter(s) on " that introduces the target, so a typed counter surfaces
+    // the same `MultiTargetSpec` as the "each kind of" form. Without this, Kinetic
+    // Ooze ("double the number of +1/+1 counters on any number of other target
+    // creatures") drops its "any number of … target creatures" bound and binds a
+    // single required target — the ETB then panics with "Unused selected target
+    // slots" (p0). The singular "counter on" arm preserves the "each kind of
+    // counter on" form unchanged.
+    let (target_text, _) = preceded(
+        tag::<_, _, OracleError<'_>>("double the number of "),
+        alt((
+            (
+                take_until::<_, _, OracleError<'_>>(" counters on "),
+                tag(" counters on "),
+            ),
+            (
+                take_until::<_, _, OracleError<'_>>(" counter on "),
+                tag(" counter on "),
+            ),
+        )),
     )
     .parse(lower.as_str())
     .ok()?;
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -13551,13 +13551,21 @@ fn lower_imperative_clause(text: &str, ctx: &mut ParseContext) -> ParsedEffectCl
     if matches!(clause.effect, Effect::SwitchPT { .. }) && clause.multi_target.is_none() {
         clause.multi_target = extract_switch_pt_multi_target(text);
     }
-    if matches!(
+    // CR 701.10e (#5247): both counter-doubling shapes carry a
+    // "double the number of … counter(s) on <target>" phrase — `Effect::Double`
+    // (the "each kind of counter" form) and `Effect::MultiplyCounter` (a typed
+    // counter ×2, e.g. Kinetic Ooze's "+1/+1 counters"). Both recover the same
+    // `MultiTargetSpec` from the shared extractor; without the MultiplyCounter
+    // arm, "on any number of other target creatures" drops its bound and the
+    // effect binds a single required target (p0 "Unused selected target slots").
+    if (matches!(
         clause.effect,
         Effect::Double {
             target_kind: DoubleTarget::Counters { .. },
             ..
         }
-    ) && clause.multi_target.is_none()
+    ) || matches!(clause.effect, Effect::MultiplyCounter { .. }))
+        && clause.multi_target.is_none()
     {
         clause.multi_target = extract_double_counter_multi_target(text);
     }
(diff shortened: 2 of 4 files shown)


matthewevans (approved):
**Approving — a p0 panic fix that routes a second effect into the *existing* extractor instead of forking a parallel one, with a claim that matches its measurement exactly.** This is the shape these PRs should have.

## ✅ Verified, not taken on the body's word

- **The card is real and the text is verbatim.** `KINETIC_OOZE` in the test matches Scryfall character-for-character, including *"double the number of +1/+1 counters on any number of other target creatures."* No paraphrase, so there is no room for a fabricated clause to survive.
- **CR 701.10e grep-verifies and is the right rule** — *"To double the number of a kind of counters on a player or permanent…"* — cited exactly where the doubling is lowered.
- **The parse-diff is exactly the claim, with zero collateral.** `1 card · 1 signature · ability/MultiplyCounter · field targets: ∅ → 0+` on Kinetic Ooze. This is the measurement that mattered most here: the fix widens a single verbatim `tag("double the number of each kind of counter on ")` into a `take_until` **class** match over any typed counter, and a widening like that is exactly where collateral hides. It bound one card. Claimed 1, measured 1.
- **The test discriminates.** `issue_5247_kinetic_ooze_conditional_multi_target.rs` drives a real `cast → resolve` through `GameScenario` and pins the actual failure (`Invalid action: Unused selected target slots`), not a proxy for it. It is registered in `tests/integration/main.rs`, so it is not an inert false-green.
- **Right seam.** Routing `Effect::MultiplyCounter` into the existing `extract_double_counter_multi_target` (`mod.rs`) rather than writing a second extractor is the correct call — one authority recovers the `MultiTargetSpec` for both counter-doubling shapes. The `alt` ordering is also right: the plural `" counters on "` arm takes typed counters, and the singular `" counter on "` arm preserves the `"each kind of counter on"` form unchanged.
- **Combinator-first.** `preceded` / `alt` / `take_until` / `tag` throughout, and every `matches!` in the diff dispatches on enum variants (`Effect::MultiplyCounter`, `Effect::Double`), never on Oracle text.

## 🟡 Non-blocking

- **The template gate flags you, and it is a formatting artifact — worth a 10-second fix next time.** The audit gate reports `invalid_implementation_method` and wants `anchored on` / `gate a` / `final review-impl`. Those three are `/engine-implementer` pipeline sections, and you honestly declared **Not `/engine-implementer`** with a justification, so they do not apply and I have **not** recorded a template-gap signal against this PR. The `method_valid: false` is almost certainly the bold markers inside the checkbox — `- [x] **Not** \`/engine-implementer\`` — defeating the parser. Dropping the `**` should make the gate read your declaration correctly.

**Recommendation: approve, label `bug` + `quality`, enqueue.** `bug` because a p0 panic is wrong→right regardless of the `fix(parser):` title. `quality` because it is one-shot with no rework round, the body's card count equals the sticky's measured count on the same card, it extended an existing authority instead of forking a path, and it ships a discriminating test that fails on `main`. Closes #5247.

PR #5670, opened 2026-07-12
fix(parser): split "... and flip a coin" so choose-source + flip parse (#5601)

### LINKED ISSUE #5601: Desperate Gambit: lose-branch prevention binds SelfRef instead of ChosenDamageSource (repair pass fires 0/35,396)

mike-theDude:
## Summary

Fixes #5601. Desperate Gambit's real Oracle text joins the choose-source step to the coin flip with a bare `" and "`:

> **Choose a source you control and flip a coin.** If you win the flip, the next time that source would deal damage this turn, it deals double that damage instead. If you lose the flip, the next time it would deal damage this turn, prevent that damage.

The clause splitter's `starts_bare_and_clause_lower` allow-list had no `flip` clause-starter arm, so the whole leading sentence stayed **one** clause. The damage-source candidate parser then rejected the trailing `"and flip a coin"` and the clause collapsed to `TargetOnly { Any }` — with three downstream failures:

1. **`ChooseDamageSource` was dropped**, so `ChosenDamageSource` had nothing to bind at resolution (the *win* branch was broken at runtime too, not just the lose branch).
2. **The coin-flip anchor was swallowed**, so the win/lose conditionals became **two separate `FlipCoin`s** (CR 705.1 — one flip).
3. The lose branch's bare `"it"` stayed `SelfRef`, because `thread_chosen_damage_source_into_oneshot_effects`'s gate never matched (no `ChooseDamageSource` def in the chain — this is why the issue observed the repair pass firing 0/35,396).

## Fix

Add a `flip <count> coin[s]` bare-and clause starter (**CR 705.1**) via `starts_flip_coin_clause_lower`. It **requires the `"coin"`/`"coins"` noun**, mirroring the `FlipCoin`/`FlipCoins` leaf parser (`try_parse_flip_n_coins`), so it never bisects the **CR 710.4** permanent-flip mechanic (`"flip <permanent>"`, the Kamigawa double-faced flip cards). `parse_count_expr` accepts the printed count forms (`a`/`two`/`X`/`that many`), keeping the split general across the whole coin-flip class rather than special-casing one card.

The split restores `"choose a source you control"` → `ChooseDamageSource` and `"flip a coin"` → a single `FlipCoin`, which lets the existing anaphor-repair pass bind the lose branch to `ChosenDamageSource`.

## Tests

- `bare_and_splits_choose_source_and_flip_a_coin` — building-block: the clause splits into two chunks.
- `bare_and_keeps_permanent_flip_conjunct` — negative: `"... and flip ~"` (permanent flip) does **not** split.
- `desperate_gambit_lose_branch_threads_chosen_damage_source` — rewritten to the real single-sentence wording; asserts `ChooseDamageSource` presence, exactly one `FlipCoin`, and both branches bound to `ChosenDamageSource`.

## Verification

`cargo fmt --check` ✅ · `cargo clippy -p engine --lib --tests -D warnings` ✅ · parser-combinator gate ✅ · 354 targeted + coin-flip/sequence/swallow_check regression tests ✅

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -16697,51 +16697,103 @@ mod tests {
         }
     }
 
-    /// CR 609.7a + CR 705: Desperate Gambit's lose branch bare "it" must thread
-    /// `ChosenDamageSource` when the chain opens with `ChooseDamageSource`.
+    /// CR 705.1 + CR 609.7a + CR 608.2c: Desperate Gambit's real Oracle wording
+    /// joins the "choose a source" step to "flip a coin" with a bare " and "
+    /// ("Choose a source you control and flip a coin."). The clause splitter must
+    /// treat "flip a coin" as an independent instruction so the head parses as
+    /// `ChooseDamageSource` (not `TargetOnly { Any }`), the flip becomes a single
+    /// consolidated `FlipCoin`, and both coin-flip branches bind the chosen
+    /// source — the win branch via "that source" and the lose branch via the
+    /// `thread_chosen_damage_source_into_oneshot_effects` anaphor repair (issue
+    /// #5601). The pre-fix parse dropped the choose step, split the flip into two
+    /// `FlipCoin`s, and left the lose branch pointed at `SelfRef`.
     #[test]
     fn desperate_gambit_lose_branch_threads_chosen_damage_source() {
         use crate::parser::oracle_effect::parse_effect_chain;
 
-        fn find_flip_coin(def: &AbilityDefinition) -> Option<&Effect> {
-            if matches!(&*def.effect, Effect::FlipCoin { .. }) {
-                return Some(&def.effect);
+        fn collect_effects<'a>(def: &'a AbilityDefinition, out: &mut Vec<&'a Effect>) {
+            out.push(&def.effect);
+            if let Effect::FlipCoin {
+                win_effect,
+                lose_effect,
+                ..
+            } = &*def.effect
+            {
+                if let Some(win) = win_effect.as_deref() {
+                    collect_effects(win, out);
+                }
+                if let Some(lose) = lose_effect.as_deref() {
+                    collect_effects(lose, out);
+                }
             }
-            def.sub_ability
-                .as_deref()
-                .and_then(find_flip_coin)
-                .or_else(|| def.else_ability.as_deref().and_then(find_flip_coin))
+            if let Some(sub) = def.sub_ability.as_deref() {
+                collect_effects(sub, out);
+            }
+            if let Some(else_def) = def.else_ability.as_deref() {
+                collect_effects(else_def, out);
+            }
         }
 
-        let text = "Choose a source you control. Flip a coin. If you win the flip, \
+        // Real Scryfall Oracle text: single sentence joining choose + flip with
+        // " and ", verified against the printed card.
+        let text = "Choose a source you control and flip a coin. If you win the flip, \
             the next time that source would deal damage this turn, it deals double that damage instead. \
             If you lose the flip, the next time it would deal damage this turn, prevent that damage.";
         let def = parse_effect_chain(text, AbilityKind::Spell);
-        let flip = find_flip_coin(&def).expect("expected FlipCoin in chain");
+
+        let mut effects = Vec::new();
+        collect_effects(&def, &mut effects);
+
+        // The "choose a source you control" step must survive as a real
+        // ChooseDamageSource — without it, ChosenDamageSource has nothing to bind
+        // to at resolution and both branches are inert.
+        assert!(
+            effects
+                .iter()
+                .any(|e| matches!(e, Effect::ChooseDamageSource { .. })),
+            "expected a ChooseDamageSource effect in the chain, got {effects:#?}"
+        );
+
+        // CR 705.1: exactly one coin flip — not two.
+        let flips: Vec<&Effect> = effects
+            .iter()
+            .copied()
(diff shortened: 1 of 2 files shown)


matthewevans:
**Closing — #5601 was already fixed by #5634, which merged this morning. You lost a race, and that is on our timing, not on your work.**

Your commit is dated `2026-07-12T02:47Z`. #5634 merged at `06:18Z` as `6ce54761f6` and closed #5601. Your branch now conflicts with `main` precisely because #5634 rewrote the same seam.

I verified the premise live rather than assuming it, and `main` already carries both of the things this PR fixes:

- **`desperate_gambit_choose_source_you_control_lowers_to_choose_damage_source`** (`oracle_effect/imperative.rs:16673`) — pins that the " and "-joined head lowers to `Effect::ChooseDamageSource` rather than a bare `TargetOnly { Any }`.
- **`desperate_gambit_shipped_text_threads_chosen_damage_source_both_branches`** (`imperative.rs:16761`) — drives the **verbatim shipped Oracle text** ("Choose a source you control **and** flip a coin.") through **`parse_oracle_text`**, the entry point `database::synthesis` actually uses to build `card-data.json`, and asserts both the win and lose branches bind `ChosenDamageSource`.

That second test is the one that makes yours redundant: it exercises the real card pipeline on the real printed text, so it cannot go green while Desperate Gambit is broken. Your version drives `parse_effect_chain` — the standalone entry point, which the shipped card never takes — so `main`'s coverage is strictly stronger here. Worth internalizing for future parser work: **prefer `parse_oracle_text` over `parse_effect_chain` in tests**, because only the former proves the card is fixed.

No defect recorded against you. Diagnosing the " and "-join as the reason the choose step was being dropped was exactly the right read — #5634 reached the same conclusion independently, and got there first.

If you want a follow-up with real value: #5652 and the coverage blind spot I'm filing alongside it are both open and unclaimed.

PR #5675, opened 2026-07-12
fix(parser): guard bounded counter distribution choices

### LINKED ISSUE #5613: parser: binary-choice splitter guards damage distribution but not counters (32 cards), and the underlying re-entry is unbounded

Lowtrixx:
## Summary
- Fixes #5613 for bounded counter distribution parsing.
- Implements the selected card: Ajani, Mentor of Heroes.
- Keeps `one, two, or three target ...` cardinality text from being misclassified as an inline `ChooseOneOf` separator.
- Preserves real external binary choices after bounded target branches by skipping only separators inside bounded-cardinality phrases and continuing the separator scan.

## Card / Oracle
Selected card: Ajani, Mentor of Heroes

Scryfall Oracle verified:
```text
+1: Distribute three +1/+1 counters among one, two, or three target creatures you control.
+1: Look at the top four cards of your library. You may reveal an Aura, creature, or planeswalker card from among them and put it into your hand. Put the rest on the bottom of your library in any order.
-8: You gain 100 life.
```

## Review Notes
- Followed the non-developer track because no Rust toolchain is available on this machine.
- Checked open PRs/issues for an existing Ajani / #5613 / bounded counter-distribution implementation; no duplicate or claimed PR found.
- Plan review completed with a clean final pass.
- Implementation review initially found two issues around external choice scanning and test coverage; both were fixed.
- Final implementation review: Maintainer-Simulation Gate PASS, no blocking findings.

## Validation
- `C:\Program Files\Git\bin\bash.exe scripts/check-parser-combinators.sh`
- `git diff --check`
- Added focused parser regression tests for Ajani and bounded-target choice separator behavior.

## Not Run
- `cargo fmt --all`: `cargo` is not available on PATH.
- Rust unit tests / card-data generation / coverage: Rust toolchain and generated card-data artifacts are unavailable in this environment.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -1,7 +1,9 @@
+use std::ops::Range;
+
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take_till1, take_until};
 use nom::character::complete::{multispace0, multispace1, satisfy};
-use nom::combinator::{all_consuming, eof, map, not, opt, peek, rest, value, verify};
+use nom::combinator::{all_consuming, eof, map, not, opt, peek, recognize, rest, value, verify};
 use nom::sequence::{preceded, terminated};
 use nom::Parser;
 
@@ -9,7 +11,7 @@ use super::super::oracle_nom::bridge::{nom_on_lower, nom_parse_lower, split_once
 use super::super::oracle_nom::duration::{
     parse_duration, parse_for_as_long_as_condition, parse_until_source_exiles_another_card_body,
 };
-use super::super::oracle_nom::error::{OracleError, OracleResult};
+use super::super::oracle_nom::error::{oracle_err, OracleError, OracleResult};
 use super::super::oracle_nom::primitives as nom_primitives;
 use super::super::oracle_nom::quantity as nom_quantity;
 use super::super::oracle_quantity::{
@@ -4709,17 +4711,12 @@ pub(crate) fn extract_bounded_target_multi_target(text: &str) -> Option<MultiTar
         else {
             continue;
         };
-        for (prefix, min, max) in [
-            ("one or two ", 1usize, 2usize),
-            ("one, two, or three ", 1, 3),
-        ] {
-            if let Ok((after_prefix, _)) = tag::<_, _, OracleError<'_>>(prefix).parse(after_verb) {
-                if tag::<_, _, OracleError<'_>>("target ")
-                    .parse(after_prefix)
-                    .is_ok()
-                {
-                    return Some(MultiTargetSpec::fixed(min, max));
-                }
+        for &(prefix, min, max) in BOUNDED_TARGET_PREFIX_PHRASES {
+            if tag::<_, _, OracleError<'_>>(prefix)
+                .parse(after_verb)
+                .is_ok()
+            {
+                return Some(MultiTargetSpec::fixed(min, max));
             }
         }
     }
@@ -4964,6 +4961,11 @@ pub(super) const BOUNDED_TARGET_PHRASES: &[(&str, usize, usize)] = &[
     ("one, two, or three targets", 1, 3),
 ];
 
+pub(super) const BOUNDED_TARGET_PREFIX_PHRASES: &[(&str, usize, usize)] = &[
+    ("one or two target ", 1, 2),
+    ("one, two, or three target ", 1, 3),
+];
+
 /// CR 115.1d + CR 601.2c: Strip exact target-count prefix before a targeted
 /// phrase. "two target creatures" and "X target creatures" both set the exact
 /// number of targets, unlike "up to X target creatures".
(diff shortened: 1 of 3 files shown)


matthewevans:
Thanks for this, and welcome — the diagnosis here is correct, which is worth saying plainly before the outcome.

**Closing as already fixed.** Issue #5613 was closed as completed by #5619 (merged `700d4364`, 2026-07-12T00:39Z), which landed the same guard roughly twelve hours before this branch was authored. That's the whole reason the branch shows `CONFLICTING`: it's based on a pre-#5619 `main`, so the code it edits no longer looks the way it does here.

Concretely, `main` already carries all three pieces:

- **`BOUNDED_TARGET_CARDINALITIES`** (`oracle_effect/lower.rs`) — the single authority for the cardinality vocabulary, measured against the full 34,632-card pool.
- **The splitter guard** in `try_parse_choose_one_of_inline` (`oracle_effect/mod.rs`), keyed on that cardinality axis rather than on a distribution verb — which is exactly the CR 601.2d "damage **or** counters" generalization you identified.
- **The regression test** `distribute_cardinality_clause_parses_within_a_bounded_stack`.

One note for next time, because it would matter even on a rebased branch: this PR introduces `BOUNDED_TARGET_PHRASES` **and** `BOUNDED_TARGET_PREFIX_PHRASES` — two lists that re-spell the same cardinality vocabulary. `main`'s constant is deliberately a single authority, with each consumer composing the trailing noun it needs (`" targets"`, `" target "`) off one stem, so that a future cardinality like *"one, two, three, or four …"* is added in exactly one place. Forking a parallel list is the pattern that authority exists to prevent.

Two things that would have caught this before you wrote the code, and are worth building into your habit here:

1. **Check the issue is still open**, and skim its cross-references — #5619 was linked on #5613 before this branch existed.
2. **Branch from a freshly fetched `main`.** This repo moves fast; a stale base is what turned a correct idea into a conflicting duplicate.

The reasoning was sound and the CR 601.2d insight was the right one. Please do pick up another issue — ideally one with no open PR linked against it, and I'd be glad to look at it.

PR #5677, opened 2026-07-12
feat(parser): L02 condition lane — typed conditions on ~30 Standard cards (BB1–BB8 + BB-FU1–5)

lgray:

## Summary

L02 "condition lane": clears the `SwallowedClause{Condition_*}` parse warnings on ~30 Standard target cards by attaching their typed conditions to the correct ability/trigger/replacement layer, built as 8 reusable building blocks (BB1–BB8) plus 5 in-branch follow-ups (BB-FU1–5) that harden the shared condition/quantity infrastructure. Net +34 newly-supported cards vs the #5668 base with zero engine regressions (coverage-regression gate: engine/oracle-changed 0); swallowed-clause diagnostics drop 1025→986. This revision additionally **enforces the CR 122.1h finality death→exile replacement** — 3 commits (typed `CounterType::Finality` → the `ZoneChange{Battlefield→Graveyard}→Exile` intrinsic replacement → a 4-path discriminating test suite) that make BB3's previously placement-only finality counters actually redirect a dying permanent to exile, resolving the maintainer's remaining blocker.

**Parse-change population (69 cards / 66 signatures):** because the building blocks operate on card *classes*, the head-current parse-diff touches more than the ~30 named targets. The full population is enumerated + classified below — all are intended targets or correct same-signature class-collateral; every `*/REMOVED` pairs with an upgraded typed replacement (no capability loss).

## Parse-change population appendix (categorized, vs base `fa7f157fa`)

| Class | Count | What |
|---|---|---|
| (a) Intended targets | ~29 | The named BB1–BB8 + BB-FU cards. |
| (b) Correct class-collateral | 37 | 28 "entered this turn" live-board→`BattlefieldEntriesThisTurn` ledger (BB-FU1, CR 608.2i — departure-surviving look-back); 9 "make a token, put a counter on *it*" anaphors now binding to the token (BB-FU5); + Wilt in the Heat. Each an Oracle-verified same-typed-signature repair. |
| (a/repr) Representation-only | 3 | `ObjectCount` zone made explicit (`# of X` → `# of in-battlefield X`); no behavior change. |

**Capability-removal clarifications (reviewer-flagged, measured):** Intrepid Paleontologist + Leonardo lost a garbage self-ETB `Moved` misparse (measured inert: counter on the wrong object, die→graveyard not exile) → replaced by correct `Exile`/`GraveyardCastPermission{finality}`. Blasphemous Edict's `ability/PayCost` (a misparsed "pay {B} rather than mana cost") → correct `AlternativeCost` casting option (BB7, tested). Finality die→exile enforcement (CR 122.1h) is now implemented in this branch (the 3 finality commits below), superseding BB3's original placement-only ceiling.

**Verify-flagged, measured correct/deferred:** Lupine Prototype + Naktamun Lorespinner ("a player has ≤N cards") — measured proper existential-over-all-players (the `HandSize` player field is an inert placeholder read per-candidate). Happily Ever After — a pre-existing `there` sentence-fragment misparse (coverage-unsupported, excluded from real play); tracked as a held follow-up, not introduced by this lane.

Every unit went through `/engine-implementer` (trace → plan → `/review-engine-plan` → executor → `/review-impl`), was verified with a discriminating revert-probe, and was ratified at its commit boundary.

### Deferred follow-ups (surfaced by this series, tracked, gated on the user)

These were discovered during L02 and intentionally scoped OUT as standalone follow-ups (novel-vs-inherited disposition; none block the L02 deliverable):
- BB-FU6 — `execute`-condition-path triggers lack a CR 603.4 fire-time intervening-if re-check (codebase-wide).
- BB-FU7 — parser over-lowers non-goad "attacks if able" → `Effect::GoadAll` (spurious `goaded_by`).
- BB-FU8 — "milled-this-way shared-quality + conditional-repeat" feature (Tamiyo + Grindstone).
- BB-FU9 — reflexive "When you do" trigger body doesn't seed `token_created_in_chain` (Iroh class).
- BB-FU10 — `parse_entered_this_turn_ref` live-board→snapshot for the optional-controller any-player class (sibling of BB-FU1).
- BB-FU11 — batched counter-trigger per-kind reproduction aggregation (the 3 KIND cards BB-FU2 kept `batched:false`).
- BB-FU12 — parse granted "first time counters this turn" condition (Danny Pink class).
- mana_spent_to_cast_amount not scrubbed + reset-hygiene + `ReplacementCondition::Or` (Freestrider / Lavinia / Satoru).
- CR-annotation sweep: `110.5b`→`110.5` + `117.1c` cite imprecision.
- Ebondeath Dracolich name-negation parser (cast-permission subject).

(Formerly listed here: finality-counter death→exile **CR 122.1h** enforcement — now **implemented** in this branch, see the finality commits.)

### Known ceilings (intentional, documented)

- **CR-cite imprecision** (`110.5b`, `117.1c`) is deferred to the standalone CR-annotation sweep rather than touched piecemeal here.
- **BB-FU2 per-kind reproduction:** the 3 "same number and kind" reproduction cards (Bold Plagiarist / Aragorn / Captain Marvel) remain `batched:false`, gated on `Effect::Unimplemented`, pending the typed reproduction effect (BB-FU11).

Method: /engine-implementer

## CR references

Primary rules implemented/annotated across the branch:
- **Conditions & triggers:** CR 603.4 (intervening-if), CR 603.2c (batched triggers), CR 603.6c + CR 603.10a (leaves-battlefield look-back), CR 201.5 (self-reference → source), CR 608.2 / 608.2c / 608.2h / 608.2i (information look-back)
- **Casting & permissions:** CR 601.1 / 601.2a / 601.2f / 601.2h / 601.3 (cast process), CR 602.5 (activate-only-if), CR 116.2a + CR 305.1 (land special action), CR 707.10 (copy is not cast)
- **Counters & continuous:** CR 122 / 122.1 / 122.1h (counters), CR 611.2a / 611.2b / 611.3a (continuous effects), CR 400.7 (object identity)
- **Finality death→exile enforcement (this revision):** CR 122.1h (the finality replacement, verbatim) + CR 614.6 (modified event occurs instead) + CR 616.1 / 616.1e (replacement ordering) + CR 700.4 (dies = to graveyard from battlefield) + CR 400.7 (exiled permanent is a new object → counters cease). Contrast annotated against shield CR 122.1c (which removes a counter; finality does not).
- Full per-commit CR annotations are in each commit body (all grep-verified against `docs/MagicCompRules.txt`). Notable correction: BB-FU3 fixes a BB4-introduced miscite on the dying gate (CR 603.6a[ETB] → CR 603.6c + 201.5 + 603.10a + 603.4).

## Verification

- [x] Required checks ran clean (full CI-equiv on the committed head, exact CI invocations).

Full CI-equiv (worktree cargo, exact CI commands) — all GREEN at committed tip **5f7acb903** (the 3 finality commits on top of the maintainer's merge **c49e6a8e9**; base `b2a827081`, upstream/main `5e7b786`):
- `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` — clean (0 warnings)
- `./scripts/check-parser-combinators.sh <base>` — Gate A PASS (finality adds zero production parser lines)
- `./scripts/check-skill-doc.sh` — ✓ · `./scripts/check-engine-authorities.sh <base>` — Gate B PASS (82 zone hits / 62 rows; draw-producers 7 rows) · `./scripts/check-test-card-data-load.sh <base>` — ✓
- `card-data-validate -- data/card-data.json` — OK (35396 parsed); oracle-gen 92.4% implemented — **identical to the pre-finality run** (typed `CounterType::Finality` serializes byte-identically to `"finality"` → parse-neutral, no card-data migration)
- `python3 scripts/draw_replacement_census.py --corpus --check` — PASS (51 rows)
- `coverage-report -- data/ --all` + `coverage-regression-check.sh <R2> --fail-on-engine` — **REGRESSED(engine)=0**, ORACLE CHANGED=0; swallowed-clause 1126→1069. The 3 `REGRESSED(honesty)` cards (Avatar Destiny / Machinist's Arsenal `[Swallow:DynamicQty]`, The Doctor's Childhood Barn `[Effect:when]`) are **stale-base false positives** — the branch is 6 commits behind upstream/main incl. `#5689` (parser "for each" scaling fix, supports the two DynamicQty cards) and `#5688`; the finality diff has zero production parser edits, so it cannot affect them. They clear on merge-forward.
- `cargo nextest run --profile ci --workspace --exclude phase-tauri --exclude mtgish-import --features engine/proptest` (PROPTEST_CASES=32) — 21051 passed / 21 skipped / 0 failed, incl. the 4 new finality death→exile tests (destroy / sacrifice / SBA-lethal / SBA-0-toughness)
- **Finality revert-probe (non-vacuity):** neutralizing the replacement's virtual-candidate injection arm flips all 4 positive assertions Exile→Graveyard (measured FAIL → restore → PASS).

_(The L02 parse-population figures above — 69/66, +34, 1025→986 — were measured vs the #5668 base `fa7f157fa` before the maintainer merged main forward; the finality work is parse-neutral so the population is unchanged, but the absolute coverage counts shifted with main's own swallow-audit evolution `fa7f157fa→b2a827081`. Population re-measure vs the moved base is a reconciliation item.)_

## Gate A

Gate A PASS head=5f7acb903af86b96045dd2792d9cf6c33edee6d5 base=5e7b786704d1d8413d33a6f6ed32f2873824e316

## Anchored on

- crates/engine/src/parser/oracle_nom/condition.rs — `parse_inner_condition`, the single authority for game-state condition recognition that every BB delegates to (no bespoke string matching)
- crates/engine/src/parser/oracle_trigger.rs — the intervening-if extraction chain (`extract_if_condition_with_card_name` → zone-change filter dispatch) that BB4/BB-FU3 extend structurally

## Final review-impl

Final review-impl PASS head=5f7acb903af86b96045dd2792d9cf6c33edee6d5

## Claimed parse impact

L02 target cards (SwallowedClause cleared): the 30 Standard condition-lane cards across BB1–BB8 (activate-only-if, cast-origin, as-long-as, finality/exile cast-permission, conditional enters-with-counter, cast-context cost gates, cast-this-way/goad). Follow-up-touched: Gargoyle Flock (BB-FU1), Stalwart Successor / All Will Be One / Generous Patron / Kros / Putrid Hexhag / Rikku / Danny Pink (BB-FU2, batched-counter flag), Phage the Untouchable (BB-FU4 fence), Otterball / Match-the-Odds class (BB-FU5). BB-FU3 is representation-stable (byte-identical card-data; the fix is structural disambiguation proven by a synthetic production-path test).


matthewevans (changes requested):
## Verdict: changes requested

Substantial, disciplined work — the condition-lane rewrite (`in battlefield you control X entered this turn ≥ 1` → `battlefield entries this turn (X, you) ≥ 1`) is a clean typed re-representation, all 14 CR citations I spot-checked grep-verify against `docs/MagicCompRules.txt`, no new enum variants (no sibling-cluster debt), and the `allow-noncombinator` annotations are in-convention (497 pre-existing uses; the gate honors the documented form). CI is fully green — 12 checks including card-data, the parser gate, and both test shards. Rubblebelt Braggart's `∅` → `not (source is suspected creature)` is exactly right.

Three things block it.

### 1. The population is understated by more than 2x

The body claims "~30 Standard cards" and "+32 supported". The head-current parse-diff sticky (baseline `fa199f8021a0`) measures **69 cards / 66 signatures**.

A wide diff is not automatically scope creep — "build for the class, not the card" produces the same signature, and most of these look like correct collateral repairs. But a reviewer cannot tell the two apart without reading the extras, and the body doesn't list them. Please state the real population card-by-card so the ~39 cards beyond your claim can be checked.

### 2. Two unexplained capability removals

**`replacement/Moved` removed** from **Intrepid Paleontologist** and **Leonardo, Sewer Samurai**. Both cards carry finality-counter reminder text — *"(If a creature with a finality counter on it would die, exile it instead.)"* — and both simultaneously *gain* a correct cast-permission static, which suggests a deliberate re-seam.

But I can find **no generic finality-counter die→exile replacement anywhere in `crates/engine/src/game/`**. If that per-card `replacement/Moved` was the only thing implementing the exile-instead-of-die behavior, this PR silently deletes finality-counter semantics on both cards, and no test would catch it.

Please either (a) point at where that behavior now lives, or (b) confirm it's a genuine loss and restore it. Same question for the unpaired **`ability/PayCost` removed** on one card.

### 3. A condition is being attached twice

> `trigger/YouAttack · field condition: ∅ → attackers declared ≥ 1 and attackers declared ≥ 1`

The same condition conjoined with itself. It's idempotent, so behavior is likely unaffected — but it means the new lane attaches the condition twice, and the next condition where `X ∧ X ≠ X` will not be so forgiving. Worth fixing at the attach site rather than the symptom.

### Non-blocking

- `ability/there · added: there (conditional=# of colors among you control permanent = 5)` — attaching a typed condition to an ability named `there` (a sentence-fragment misparse the lane didn't create, but is now decorating rather than fixing). Worth a follow-up issue.
- **Size.** 8,695 additions / 49 files / 16 commits bundling 13 units (BB1–BB8 + BB-FU1–5) across shared parser infrastructure. Individually the units look sound and the commit boundaries are clean, but this is at the edge of what can be reviewed as one artifact. Future lanes would land faster split at the BB boundary.

Happy to re-review promptly once the population is stated and the removals are explained.

lgray:

Thanks for the detailed review — every point measured and addressed. Rebased onto **#5668** (`fa7f157fa`); all figures below are re-measured against your per-unit audit, and the parse-diff population re-confirmed identical after the rebase.

## B1 — population (69 cards / 66 signatures)

Fully accounted: **~30 named targets + 39 legitimate class-collateral**, dominated by two deliberate class-wide rules migrations — 28 "entered this turn" counts moved to the departure-surviving `BattlefieldEntriesThisTurn` ledger (CR 608.2i) and 9 "make a token, put a counter on *it*" anaphors now binding to the token (BB-FU5) — plus 3 representation-only zone serializations. Every `*/REMOVED` pairs with an upgraded typed replacement (no capability loss). A full categorized card-by-card appendix is now in the PR description.

## B2 — the two capability removals (measured against the true pre-BB3 parent `313374b0b^`)

Both are misparse cleanups, not finality regressions:

- **Intrepid Paleontologist & Leonardo, Sewer Samurai** — at base the cast-permission was entirely *swallowed* (`static_abilities: []`) and the only parse was a garbage self-ETB `Moved` replacement `PutCounter{target: SelfRef, counter_type: "<the whole clause text>"}`. Driven at runtime, that counter lands on the *source card itself*, no cast-permission is granted, and killing the source sends it to the **graveyard, not exile** — inert on both axes. BB3 replaces it with the correct `ExileCastPermission`/`GraveyardCastPermission{enters_with_counter: finality}` and no replacement.
- **The `ability/PayCost` removal is Blasphemous Edict** (not Noctis — Noctis only *gained* `extra_cost: additional`). Its "pay {B} rather than this spell's mana cost if there are thirteen or more creatures" is an alternative cost, misparsed at base as a standalone `PayCost` ability; BB7 re-models it as a proper `AlternativeCost` casting option gated on `ObjectCount(Creature) ≥ 13`, with parse + runtime tests. Same cost, correct representation — nothing lost.

On finality: CR 122.1h die→exile enforcement does not exist anywhere in the engine (only counter-*placement*), so no representation old or new exiled a dying finality-countered permanent. The real fix is a tracked follow-up (class-wide CR 122.1h death→exile, mirroring shield counters CR 122.1c). This lane correctly scopes to counter placement.

## B3 — "condition attached twice" (X ∧ X)

A **display artifact, not a duplicate condition.** Fearless Swashbuckler parses to `And[AttackersDeclaredCount{Pirate}≥1, AttackersDeclaredCount{Vehicle}≥1]` — a faithful `Pirate ∧ Vehicle` parse; the two conjuncts differ only in their `subject.filter`. The parse-diff overlay collapsed them to identical text because the condition formatter (`coverage.rs`) elides the `subject` field via `..`. The attach site (`parse_typed_attacked_this_combat_conjunction`) is single-pass and a whole-DB scan finds zero byte-identical duplicate conjuncts. A dedup here would be *incorrect* — it would make the card fire on a lone Pirate. We've filed the lossy formatter as a tracked coverage-tooling follow-up so the overlay renders the subject filter; the parse itself is correct and unchanged.

## Verify-flagged cards (drove-and-measured)

- **Lupine Prototype & Naktamun Lorespinner** ("a player has ≤N cards") — not a scope bug. The predicate reads each candidate player's own hand (the `player` field on `HandSize` is a documented-inert placeholder ignored by the evaluator). Measured existential: TRUE when any opponent has ≤N cards while the controller has more, FALSE when all players are above. The `{player: Controller}` in the overlay is the diff tool's Rust-`Debug` render, not a serialization defect.
- **Happily Ever After** (the `ability/there` non-blocking note) — good eye; this is a pre-existing `there` sentence-fragment misparse the lane *surfaced* but didn't create. The card is coverage-**unsupported** (its win effect parses to `Unimplemented`), so it's excluded from real games and never resolves a win today. We measured the forced resolution: the `WinTheGame` sub *would* fire on the partial 5-colors gate, so we've filed a tracked follow-up that must rebuild the full three-conjunct condition **and** hoist the trigger-level intervening-if before `there` is implemented — implementing the fragment alone would turn it into a live eager-win. Not fixed in this lane (no real-game behavior change; unsupported before and after).

## Size

Acknowledged — the lane bundled 8 building blocks + 5 in-branch follow-ups. Future condition lanes will split at the building-block boundary to keep review units smaller.

---

No code change was needed for any blocking item (all measured-correct / cleanup / false-positive); the only new commit on the rebase is a test-only adaptation of BB8's three swallow-audit tests to your per-unit `check_swallowed_clauses` signature. Full CI-equiv green at `411e6d7dd` (clippy 0-warnings, nextest 21046/0, coverage-regression engine/honesty/oracle-changed all 0, +34 supported). Happy to expand any of these.

matthewevans (changes requested):
**Verdict: changes requested — the finality-counter lane is rules-incomplete and coverage-unsafe.**

## 🔴 Blocker

- **Finality is serialized as an inert generic counter, not enforced as CR 122.1h's replacement effect.** Evidence: `crates/engine/src/types/counter.rs:64` represents it as `Generic(String)`; `crates/engine/src/game/casting_costs.rs:6901` only adds the pending ETB counter; and `crates/engine/src/game/replacement.rs:5286` synthesizes intrinsic counter replacements only for shield counters, with no battlefield-to-graveyard finality candidate. The new integration suite explicitly limits itself to placement, “never death→exile,” at `crates/engine/tests/integration/l02_bb3_cast_permission_finality.rs:16`. Intrepid Paleontologist's Oracle text says: “If a creature with a finality counter on it would die, exile it instead.” CR 122.1h defines the replacement as: “If this permanent would be put into a graveyard from the battlefield, exile it instead.” A finality-countered creature therefore still reaches the graveyard after destruction, sacrifice, or an SBA in this branch.

  **Recommendation: route this through the full engine cycle before merging. Add the typed finality-counter/replacement building block at the shared zone-change replacement seam (covering destroy, sacrifice, and SBA), with discriminating runtime tests for each; otherwise preserve strict unsupported coverage for these cards instead of clearing their condition warnings.**

## 🟡 Non-blocking

- No additional current-head design finding. The maintainer merge advanced the head to `c49e6a8`; its CI and parse-diff are correctly pending and must be fresh before the next review.

## ✅ Clean

- The earlier apparent duplicate Fearless Swashbuckler condition is a formatter loss, not a duplicated parse: `crates/engine/src/parser/oracle_trigger.rs:4089` retains distinct Pirate and Vehicle filters, while `crates/engine/src/game/coverage.rs:3907` deliberately formats only comparator/count.
- The prior `Moved` and `PayCost` observations were rechecked against the current implementation: the cast-permission/counter plumbing is at the correct selected-permission seam, and Blasphemous Edict's printed “pay {B} rather than pay this spell's mana cost” is an alternative-cost shape. The remaining defect is the missing generic finality rule, not those representation changes.

**Recommendation: keep this PR in requested-changes state; regenerate the parse-diff and rerun current-head CI after the rules-complete finality implementation.**

lgray:

Implemented the full **CR 122.1h** finality death→exile replacement (Option A) — finality counters now redirect a dying permanent to exile instead of the graveyard, not just get placed. Pushed as 3 commits on top of your merge `c49e6a8e9` (preserved as ancestor; no rebase/squash across it).

**Architecture.** It mirrors the shield-counter intrinsic replacement, but on the correct seam: shield replaces `ProposedEvent::Destroy`, whereas CR 122.1h is a *zone-change* redirect, so finality replaces `ProposedEvent::ZoneChange{from:Battlefield, to:Graveyard}`. That single virtual candidate catches battlefield→graveyard death in one place — destroy, sacrifice, SBA lethal damage, SBA zero-toughness, and the other `move_object` SBA paths — because they all converge on that event. It rides the same replacement seam as shield counters and Rest in Peace (both shipped + tested), so it intercepts every battlefield→graveyard death those already do; the SBA zone-change consult (`ZoneChangeCause::is_exempt`) is exhaustive with no death cause exempt, and we measured the four principal paths end-to-end (each with a revert-probe). The applier redirects to exile via `Ok(Modified)` (never `Prevented`) and does **not** consume the counter — CR 122.1h, unlike shield CR 122.1c, has no remove-counter clause. One-shot is guaranteed by the existing `already_applied` guard plus the `to:Graveyard` match (the post-redirect `to:Exile` no longer matches).

**Design calls.** Promoted the counter to a first-class typed `CounterType::Finality` (matching the Shield/Stun/Defense precedent + the enum-over-stringly-typed guideline); serialization is byte-identical to `"finality"`, so there's no card-data migration and the change is parse-neutral. CR 616 materiality is `Unconditional`, consistent with every existing zone-redirect (Rest in Peace, shield-destroy, umbra) — a lone finality counter auto-applies, and co-firing redirects force the CR 616.1e ordering choice; we deliberately did **not** add a new `EventField::Destination` axis (no candidate ever writes a destination field, so it would be dead flexibility). No user toggle — this enforces a printed, previously-inert ability, not a new optional feature.

**Tests.** Four discriminating integration tests, one per distinct death path (destroy / sacrifice / SBA-lethal-damage / SBA-zero-toughness), each driving the real pipeline and asserting exile-not-graveyard, each with a negative twin (no counter → graveyard, the reach-guard proving the driver actually kills) and a dies-trigger differential proving true "instead" (CR 700.4 — the exiled permanent never reaches the graveyard, so `dies` triggers don't fire). Revert-probe: neutralizing the injection arm flips all four positives Exile→Graveyard. One measured refinement over the original plan: we dropped a "counter survives on the exiled permanent" assertion because CR 400.7 makes the exiled object a *new object with no counters* — non-consumption is instead guaranteed **structurally** (the applier takes no `&mut GameState`, so it physically cannot decrement a counter).

CI is green (clippy `--workspace --all-targets` 0 warnings, nextest 21051/0 incl. the 4 new tests, Gate A/B, coverage `REGRESSED(engine)=0`, oracle-gen identical to pre-finality). The change adds **zero** production parser lines, so card-data is unchanged. (Three `REGRESSED(honesty)` cards on the current run — Avatar Destiny / Machinist's Arsenal / The Doctor's Childhood Barn — are stale-base false positives from being a few commits behind `main`'s parser fixes #5688/#5689; the finality diff has no parser edits so it can't affect them, and they clear on the next merge-forward.)

matthewevans (approved):
Approved after resolving the remaining finality runtime, visibility, and parser-audit feedback; current-main CI is green.

PR #5679, opened 2026-07-12
fix(parser): UNSUPPORTED cluster: Secret Lair — The Fourteenth Doctor: enter-as-a-COPY-from-GRAVEYAR

ntindle:
## Summary

Repairs the **reveal-partition** parser class — the *"Put all `<filter>` revealed this way into `<zone>` and the rest on the bottom/top of `<library>` [in a random/any order]"* shape — and lands the Secret Lair headline card **The Fourteenth Doctor** (its cast-trigger mill partition **and** its enter-as-a-copy-from-your-graveyard replacement).

### Real population (correcting the earlier "affecting 1 card(s)")

The earlier body understated this as *1 card*. The head-current parse-diff measures **22 cards** after the reviewer fix below. (@matthewevans measured **23 cards / 38 signatures** at review time; this revision narrows the Muse Vortex guard, which returns Muse Vortex to byte-identical-with-`main` behavior and removes it from the diff → **22 cards, all correct class repairs**.)

**This is "build for the class, not the card" working as intended — not scope contamination.** The fix dispatches onto existing runtime handlers plus one parser-IR field (`rest_library_position`); no new engine enum variant. The headline card **The Fourteenth Doctor** is one of the 22.

#### Why 22, not 1 — the reveal-partition class

On `main` every card of this shape mis-parsed to a single-card `PutAtLibraryPosition(count=1, position=Bottom, target=<subtype>)`: it never put the primary pile into hand/graveyard, and it put *one filtered* card on the bottom instead of the *complement*. Plainly broken. Repaired cards (Fix A, reveal-partition): Goblin Ringleader, Sylvan Messenger, Merfolk Wayfinder, Enshrined Memories, Alrund God of the Cosmos, Brass Herald, Garruk Caller of Beasts, Ajani Unyielding, Lair Delve, Kavu Howler, Enlistment Officer, Grave Defiler, Tidal Courier, and The Fourteenth Doctor's own cast-trigger mill. Each now parses `RevealTop → ChangeZoneAll(primary) → ChangeZoneAll(complement, library, bottom/top)`, with `"in a random order"` → `random_order: true` routed to whichever pile returns to the **library** (CR 401.4) vs `"in any order"` → owner-arranged (`random_order: false`).

The remaining repairs are Fix B clone-source field repairs — e.g. **Activated Sleeper** now *enforces* its `"put there from the battlefield this turn"` restriction (previously dropped entirely) via a `FilterProp::ZoneChangedThisTurn` on the copy-source filter.

---

## Reviewer response — @matthewevans (changes requested)

**Blocker 1 — declared population off by 23×.** Fixed in this body (see *Real population* above): the true population is **22 correct reveal-partition / clone-source class repairs** (was mislabeled "1 card"), and the class is named. None of the 22 is scope contamination.

**Blocker 2 — Muse Vortex mis-binding.** Fixed in `602fe5a54`. The `is_this_way_partition` guard in `parse_put_imperative` was scoped to the bare `"this way"`, which also matched Muse Vortex's **exile-origin** partition — *"put the exiled instant and sorcery cards that weren't cast **this way** into your hand and **the rest** on the bottom of your library in a random order"* — whose primary subject is a negation over the *cast* set, not the `"put all <filter> revealed this way into <zone>"` reveal shape the guard's own comment names. With the broad guard it was hijacked onto the tracked-set path and bound a phantom `ChangeZone { target: TargetFilter::Any }` ("any target") into the hand, on a card with no targeting at all.

- **Fix:** narrowed the guard to the full **`"revealed this way"`** reveal-origin anaphor — exactly the class the comment documents.
- **Result (oracle-gen):** Muse Vortex no longer reaches the tracked-set path. It returns to byte-identical-with-`main` positional handling — `ExileTop(X) → CastFromZone → PutAtLibraryPosition { target: ExiledBySource, position: Bottom }` — with **no phantom `any target`**.
- **Pinned:** `muse_vortex_exile_partition_does_not_bind_phantom_any_target` asserts no hand-bound `ChangeZone`/`ChangeZoneAll` carries `TargetFilter::Any`.
- **Class-wide check:** ran oracle-gen on **all 24** cards that match `"the rest" + "this way"` but *not* `"revealed this way"` (the complete set that could flip under this guard). Zero phantom `Any` bindings, zero partition-clause regressions; the pre-existing `Unimplemented` nodes on Collision of Realms / The Joiner of Cats / Yuna's Whistle are on unrelated clauses (`"who [condition]"`, token-copy `create`, reflexive counter `when`) and are identical to `main`.

**Blocker 3 — unexplained `static_structure` removal.** The card is **The Fourteenth Doctor** — its own clone clause, and the *intended* headline repair (unparsed → parsed), not a regression.

- It is the **only** card in the entire pool whose clone clause reads *"enter as a copy of a `<...>` **in your graveyard**"* (possessive) — the exact pattern Fix B's clone-source-zone table addition (`" in your graveyard"` → `Zone::Graveyard` + `ControllerRef::You`) newly parses. On `main` no clone-source candidate matched `"in your graveyard"` (only `" in a graveyard"` / `" in any graveyard"` / `" on the battlefield"`), so the clause fell through to `Effect::Unimplemented { name: "static_structure" }`.
- It now parses to a `BecomeCopy` replacement gated on `FilterProp::ZoneChangedThisTurn { from: Library, to: Graveyard }` with a reflexive `Haste` sub-ability (0 Unimplemented / 0 Unknown).
- **No other card lost a `static_structure`.** The other clone-from-graveyard cards (Activated Sleeper, Body Double, The Mimeoplasm, Superior Spider-Man, Naga Fleshcrafter, Mirrorhall Mimic, Chameleon) use `"in a/any graveyard"` or battlefield, which `main` already parsed — Fix B only *adds* a `ZoneChangedThisTurn` filter prop to those (a field-level change, not a `static_structure` removal).

---

## Cards corrected

- **The Fourteenth Doctor** (headline): cast-trigger mill partition (Fix A) + enter-as-a-copy-from-your-graveyard replacement with reflexive Haste (Fix B).
- Reveal-partition class (Fix A): Goblin Ringleader, Sylvan Messenger, Merfolk Wayfinder, Enshrined Memories, Alrund, Garruk Caller of Beasts, Ajani Unyielding, Lair Delve, Kavu Howler, Enlistment Officer, Grave Defiler, Brass Herald, Tidal Courier, …
- Clone-source field repair (Fix B): Activated Sleeper (now enforces "put there from the battlefield this turn").

## Fix

**FIX A (cast-trigger reveal partition — `imperative.rs`, `ast.rs`, `oracle_effect/mod.rs`).** The *"put all `<filter>` revealed this way into `<z1>` and the rest on the bottom of `<library>`"* partition previously hijacked to a single-card `PutAtLibraryPosition{Bottom}` filtered to the subtype (dropping the primary move, inverting the partition). Added an `is_this_way_partition` gating pre-filter that diverts the sentence to the tracked-set path, a parser-IR field `rest_library_position: Option<LibraryPosition>` on `PutImperativeAst::ZoneChangeAll`, a `parse_put_rest_library_position` tail-scan helper, and routing so `"in a random order"` (random) vs `"in any order"` (owner-arranged) lands on the correct pile. **Reviewer revision:** the guard's anaphor is narrowed from the bare `"this way"` to `"revealed this way"` so exile-origin negation partitions (Muse Vortex) are not hijacked (Blocker 2).

**FIX B (enter-as-copy-from-graveyard replacement — `oracle_replacement.rs`, `oracle_target.rs`).** Added possessive `" in your graveyard"` to the clone source-zone table (owner-scoped via `ControllerRef::You`, B1), made `parse_zone_changed_this_turn_suffix` `pub(crate)` (B2), consumed the *"that was put there from your library this turn"* clause into `FilterProp::ZoneChangedThisTurn { Library→Graveyard }` on the copy-source filter (B3), and accepted the *"if you do"* reflexive connector (B4). The Fourteenth Doctor now parses to a `BecomeCopy` replacement gated on the this-turn zone change with a reflexive Haste sub-ability; 0 Unimplemented / 0 Unknown.

**DEVIATION (proven by the mandated card-test).** Plan §6 predicted the reflexive condition would be `WhenYouDo`, but `parse_effect_chain` maps `"if you do"` to `EffectOutcome{OptionalEffectPerformed}`. The runtime card-test proved that condition never fires on the `BecomeCopy` copy-completion path, whereas `effects/mod.rs` documents `WhenYouDo` as THE condition for BecomeCopy reflexives. A scoped normalization in `parse_when_you_do_reflexive` converts `EffectOutcome{OptionalEffectPerformed}` → `WhenYouDo` for clone-replacement reflexives only.

## Files changed

- `crates/engine/src/parser/oracle_effect/imperative.rs` (Fix A + Blocker-2 guard narrowing)
- `crates/engine/src/parser/oracle_effect/tests.rs` (Blocker-2 Muse Vortex pin + partition class tests)
- `crates/engine/src/parser/oracle_ir/ast.rs`
- `crates/engine/src/parser/oracle_effect/mod.rs`
- `crates/engine/src/parser/oracle_target.rs`
- `crates/engine/src/game/engine_replacement.rs`
- `crates/engine/tests/integration/the_fourteenth_doctor_graveyard_copy.rs`
- `crates/engine/tests/fixtures/integration_cards.json`

## CR references

- CR 401.4 — arranging 2+ cards placed at a library position (random-order routing).
- CR 608.2c — follow instructions in written order (partition of a tracked/revealed set).
- CR 701.17a — mill (library → graveyard).
- CR 701.20a — reveal.
- CR 400.7 / CR 614.1c — object becomes a new object on zone change / enters-as replacement.
- CR 603.12 — reflexive triggered ability.
- CR 707.9 — copy modifications/exceptions.
- CR 108.3 / CR 400.3 — owner-scoped source zone (possessive "your graveyard").
- CR 702.10 — Haste.

## Verification

- `cargo fmt --all` — clean.
- `check-parser-combinators.sh` (base = `upstream/main` merge-base) — **Gate A PASS**.
- `cargo clippy -p engine -p phase-ai -p engine-wasm --all-targets -- -D warnings` — clean, exit 0.
- `cargo test -p engine -p phase-ai` — PASS (0 failed); includes `muse_vortex_exile_partition_does_not_bind_phantom_any_target`, the reveal-partition class tests, and the `the_fourteenth_doctor_mills_doctor_then_enters_as_hasty_copy` cast-pipeline integration test.
- `oracle-gen` on The Fourteenth Doctor, Goblin Ringleader (class), Muse Vortex, and all 24 `"the rest"+"this way"`-but-not-`"revealed this way"` cards — correct AST, no phantom `Any` bindings, no partition-clause Unimplemented.


matthewevans (changes requested):
## Verdict: changes requested

The partition fix is **architecturally right** and repairs a real, wide misparse class. Three blockers, one of them a genuine correctness regression on a card the PR doesn't mention.

---

### What's good (verified, not assumed)

- **The seam is correct.** No new engine enum variant; you dispatch onto existing runtime handlers plus one parser-IR field (`rest_library_position`). That's the right layer.
- **The CR reasoning is right and the citations govern.** CR 401.4 — *"If an effect puts two or more cards in a specific position in a library at the same time, the owner of those cards may arrange them in any order"* — is exactly on point, and routing randomness to whichever pile returns to the **library** is the correct reading. Your `"in a random order"` (The Fourteenth Doctor → `random_order: true`) vs `"in any order"` (Garruk / Goblin Ringleader → `false`, owner arranges) distinction is right. CR 608.2c and CR 701.17a also verify.
- **The class repair is real.** On `main`, Goblin Ringleader parsed to `PutAtLibraryPosition(count=1, position=Bottom, target=Goblin)` — it never put anything into hand, and it put *a Goblin* on the bottom rather than the rest. Plainly broken. Same for Sylvan Messenger, Merfolk Wayfinder, Enshrined Memories, Alrund, Brass Herald, and the rest of the reveal-partition class. Activated Sleeper was ignoring its `"put there from the battlefield this turn"` restriction **entirely**; your predicate now enforces it. These are 22 correct repairs.
- **The integration test drives the real pipeline** — `GameScenario` + `add_real_card(..., db)`, not a hand-built ability — and the parser tests pin *both* partition shapes including the complement's zone, position, and randomness. That is discriminating.

---

### Blockers

**1. The declared population is off by 23×.**

The body says *"affecting 1 card(s)"*. The head-current parse-diff measures **23 cards / 38 signatures**. Twenty-two of those are correct class repairs — that's "build for the class, not the card" working as intended, and it is *not* scope contamination. But the body has to say so. State the real population and name the class, so the next reviewer doesn't have to reconstruct it from the sticky.

**2. Muse Vortex — the guard is broader than the class it documents, and mis-binds.**

```rust
let is_this_way_partition = nom_primitives::scan_contains(lower, "the rest")
    && nom_primitives::scan_contains(lower, "this way");
```

Muse Vortex reads:

> *"Then put the exiled instant and sorcery cards that weren't cast **this way** into your hand and **the rest** on the bottom of your library in a random order."*

Both substrings hit, so it's pulled into the partition path — but it is an **exile-origin** partition whose primary subject is *itself a negation* (`"...that weren't cast this way"`), not the `"put all <filter> revealed this way into <zone>"` shape your own comment names. The head-current sticky shows the result:

```
ability/ChangeZone · added: ChangeZone (target=any target, to=hand)   [Muse Vortex]
```

`any target` is a **targeting** filter, on a card whose text contains no targeting at all. `main` was also wrong here (`PutAtLibraryPosition(count=Fixed { value: 0 }, ...)` — a no-op), so this isn't correct→broken; it's broken→differently-broken, and now it binds a target where none exists.

Either narrow the guard to the reveal/tracked-set origin it actually means, or handle the exile-origin partition properly — and pin Muse Vortex with a test either way. Right now nothing in the suite would notice this.

**3. Unexplained `static_structure` removal.**

The sticky's collapsed section carries `1 card(s) · ability/static_structure · removed: static_structure` with no example named. Please identify the card and justify the removal, or restore it.

---

### Non-blocking

The `parse_effect_chain` unit tests exercise the standalone entry point rather than the card pipeline (`parse_effect_chain_with_context`). The Fourteenth Doctor is covered end-to-end by the integration test, so this is fine for the headline card — but for the *class* cards, the parse-diff sticky is currently the only pipeline-level evidence.

---

Nice work on the lowering — the random-order routing in particular is a subtle call and you got it right.

matthewevans (approved):
Current 22-card parse diff and full CI are green; the reveal-partition, Muse Vortex, fixture, and suffix-consumption findings are resolved on this head.

PR #5680, opened 2026-07-12
fix(engine): support Ram Through excess rider

### LINKED ISSUE #5632: Ram Through: conditional excess-damage rider drops to Unimplemented (primitive already exists)
## Summary
The trample excess-damage rider on Ram Through is dropped to `Effect::Unimplemented` even though the terminal sentence is byte-identical to the 4 cards that parse it correctly, and the engine primitive (`excess: TargetController` / `ExcessRecipient::TargetController`) already exists. Surfaced by the U20 walk-back binding campaign (task #20 harvest, full-pool sweep).

## Actual parse
```
Effect::Unimplemented { name: "excess", description: "excess damage is dealt to that creature's controller instead" }
```

## Cause
The conditional prefix ("If the creature you control has trample,") defeats the excess-damage recognizer. The 4 cards that fire (Flame Spill, Gandalf's Sanction, Pigment Storm, Ravenous Tyrannosaurus) carry the same terminal sentence without the conditional prefix.

## Fix shape
Cheap, high-confidence: extend the recognizer to accept the conditional prefix (condition parsing already exists via `parse_inner_condition`) and set the typed `excess` field under that condition instead of dropping to Unimplemented.

## Population (from the sweep of all 35,034 oracle texts)
"excess damage": 35 cards → 4 set the typed field correctly, this card is the only recognizer-defeat in class, 5 are Unimplemented in other mechanics, 25 are other clause classes (excess-as-quantity / excess-as-trigger / reminder text — correctly declined). Visible as a gap (Unimplemented), so lower severity than the silent class.

Lowtrixx:
## Summary

- Parse Ram Through's conditional excess-damage rider into the existing DealDamage excess path.
- Add a source-keyword-gated excess recipient.
- Resolve the condition against the actual damage source target.
- Add focused parser and runtime regression coverage.

Fixes #5632

### DIFF
--- a/crates/engine/src/game/effects/deal_damage.rs
+++ b/crates/engine/src/game/effects/deal_damage.rs
@@ -652,7 +652,9 @@ pub(crate) fn apply_damage_after_replacement(
         && matches!(
             (ctx.excess_recipient, &t),
             (
-                Some(ExcessRecipient::TargetController),
+                Some(ExcessRecipient::TargetController {
+                    source_keyword: None
+                }),
                 TargetRef::Object(_)
             )
         ) {
@@ -778,8 +780,12 @@ pub(crate) fn apply_damage_after_replacement(
     // the snapshot / PendingReplacement). `redirected` records that the excess and
     // this leg's lifelink were handed off, so the creature leg gains nothing here.
     let mut redirected = false;
-    if let (Some(ExcessRecipient::TargetController), TargetRef::Object(obj_id)) =
-        (ctx.excess_recipient, t)
+    if let (
+        Some(ExcessRecipient::TargetController {
+            source_keyword: None,
+        }),
+        TargetRef::Object(obj_id),
+    ) = (ctx.excess_recipient, t)
     {
         if redirect_excess > 0 {
             if let Some(controller) = state.objects.get(obj_id).map(|o| o.controller) {
@@ -1070,11 +1076,17 @@ fn active_excess_recipient(
     excess: Option<ExcessRecipient>,
 ) -> Option<ExcessRecipient> {
     match excess {
-        Some(ExcessRecipient::TargetController) => Some(ExcessRecipient::TargetController),
-        Some(ExcessRecipient::TargetControllerIfSourceHasKeyword { keyword })
-            if keywords::object_has_effective_keyword_kind(state, ctx.source_id, keyword) =>
-        {
-            Some(ExcessRecipient::TargetController)
+        Some(ExcessRecipient::TargetController {
+            source_keyword: None,
+        }) => Some(ExcessRecipient::TargetController {
+            source_keyword: None,
+        }),
+        Some(ExcessRecipient::TargetController {
+            source_keyword: Some(keyword),
+        }) if keywords::object_has_effective_keyword_kind(state, ctx.source_id, keyword) => {
+            Some(ExcessRecipient::TargetController {
+                source_keyword: None,
+            })
         }
         _ => None,
     }
@@ -4982,8 +4994,8 @@ mod tests {
                     amount: QuantityExpr::Fixed { value: 5 },
                     target: TargetFilter::Any,
                     damage_source: Some(DamageSource::Target),
-                    excess: Some(ExcessRecipient::TargetControllerIfSourceHasKeyword {
-                        keyword: crate::types::keywords::KeywordKind::Trample,
+                    excess: Some(ExcessRecipient::TargetController {
+                        source_keyword: Some(crate::types::keywords::KeywordKind::Trample),
                     }),
                 },
                 vec![TargetRef::Object(source_id), TargetRef::Object(target_id)],
@@ -5009,7 +5021,10 @@ mod tests {
         );
 
         let (no_trample_state, no_trample_events, no_trample_target) = run(false);
-        assert_eq!(no_trample_state.objects[&no_trample_target].damage_marked, 5);
+        assert_eq!(
+            no_trample_state.objects[&no_trample_target].damage_marked,
+            5
+        );
         assert_eq!(no_trample_state.players[1].life, 20);
         assert_eq!(
             no_trample_events
--- a/crates/engine/src/game/effects/deal_damage.rs
+++ b/crates/engine/src/game/effects/deal_damage.rs
@@ -1061,6 +1061,25 @@ fn stash_remaining_each_source_damage(
     append_to_pending_continuation(state, Some(Box::new(head)));
 }
 
+/// CR 120.4a + CR 608.2c + CR 702: Resolve an excess-redirect rider against
+/// the actual damage source. Unconditional riders apply directly; keyword-gated
+/// riders apply only while the source has the named effective keyword.
+fn active_excess_recipient(
+    state: &GameState,
+    ctx: &DamageContext,
+    excess: Option<ExcessRecipient>,
+) -> Option<ExcessRecipient> {
+    match excess {
+        Some(ExcessRecipient::TargetController) => Some(ExcessRecipient::TargetController),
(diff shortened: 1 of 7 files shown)


matthewevans (changes requested):
**Changes requested.** The architectural direction here is right — replacing the `scan_contains` chain and its two negative guards with a combinator that owns the whole rider shape is exactly the correct move, and `partial_conditional_excess_rider_does_not_patch_damage` is a well-conceived negative test. But three things block it, and the second is the serious one.

## 1. The branch does not compile

`crates/engine/src/parser/oracle_effect/tests.rs:935` and `:958` — `error[E0433]: cannot find type ExcessRecipient in this scope`. The import is missing:

```rust
use crate::types::ability::ExcessRecipient;
```

Both test shards fail to build, which means **all three of your new parser tests have never executed.** Plus 4 `rustfmt` diffs (`deal_damage.rs:5009`, `sequence.rs:3613`, `:5862`, `:5881`) — run `cargo fmt --all`.

## 2. The fix appears to be inert on Ram Through itself

This is the blocker that matters. The head-current parse-diff (posted 13:50Z against head `d6d82e4b`, committed 13:28Z, with the card-data job green) reports **"✓ No card-parse changes detected."**

On current `main`, Ram Through parses to:

```
Spell → sub_ability: DealDamage { damage_source: Target }
          → sub_ability: Unimplemented { name: "excess",
              description: "excess damage is dealt to that creature's controller instead" }
```

A working fix must make that `Unimplemented` node disappear. I checked that the parse-diff *can* see such a change before concluding anything from its silence: `effect_type_name` (`coverage.rs:521`) emits `"Unimplemented"` into the signature, and `build_ability_item` (`:4445`) recurses into `sub_ability` chains — which is exactly where Ram Through's node sits. The instrument is not blind at this position. It reports no change, so **the `Unimplemented` is still there and the card is still broken.**

My strong suspicion as to why — worth confirming rather than taking from me: your three tests all drive `parse_effect_chain`, which runs a fresh `ParseContext`. The card pipeline goes through `parse_effect_chain_with_context`. A rider combinator can pass under the former and never fire under the latter, because the clause text that actually reaches `parse_followup_continuation_ast` in the pipeline is not necessarily the string you hand it in the test. **Please add a test that drives `parse_oracle_text` on Ram Through's real Oracle text and asserts the `Unimplemented` is gone** — that is the test that would have caught this, and it is the one that will prove the fix when it lands.

Note also that `Effect::DealDamage { amount, target, .. }` at `coverage.rs:2186` discards `excess`, so the sticky can never *confirm* your excess field either way. The disappearing `Unimplemented` is the observable, and it is the one to watch.

## 3. Parameterize, don't proliferate

`ExcessRecipient::TargetController` and `TargetControllerIfSourceHasKeyword { keyword }` differ **only in condition shape** — which CLAUDE.md names explicitly as a parameterization axis, at the *first* sibling, not the third. The right form is one variant with an optional gate:

```rust
pub enum ExcessRecipient {
    /// CR 120.4a: excess is dealt to the damaged permanent's controller,
    /// optionally gated on the resolved damage source having a keyword.
    TargetController { source_keyword: Option<KeywordKind> },
}
```

Your own resolver is the evidence: `active_excess_recipient` collapses the gated variant straight back into the ungated one — both arms return `Some(ExcessRecipient::TargetController)`. That collapse is the tell that these were never two things. With the parameterized form the resolver reduces to evaluating one optional gate, and the next keyword-gated rider costs zero new variants.

---

CR citations check out — 120.4a, 608.2c and 702 all grep-verify and genuinely govern here. Get it compiling, get a `parse_oracle_text` test on the real card showing the `Unimplemented` gone, and collapse the enum, and I'll take this.

Lowtrixx:
Thanks for the detailed review — all three points make sense.

I’m updating the branch to:

* fix the missing import and formatting failures;
* add a `parse_oracle_text` regression test using Ram Through’s real Oracle text and verify that the excess-related `Unimplemented` node disappears through the actual card pipeline;
* collapse the two excess-recipient variants into a single parameterized `TargetController { source_keyword: Option<KeywordKind> }` form and update all consumers consistently.

I’m also checking the sibling wording noted by Gemini and will only support it if it resolves to the same damage-source referent in this parser context.

I’ll push the corrected implementation after the full local validation sequence passes and then request another review.

matthewevans (changes requested):
**Request changes: the new source-keyword gate is only exercised through a hand-built `ResolvedAbility`, not through a cast of Ram Through's parsed ability.**

## 🔴 Blocker

- `crates/engine/src/game/effects/deal_damage.rs:4992-5007` constructs `Effect::DealDamage`, its two object targets, and `ExcessRecipient::TargetController { source_keyword: Some(Trample) }` directly, then calls `resolve`. The new integration test at `crates/engine/tests/integration/flame_spill_excess_damage.rs:457-513` is also parser-shape-only for Ram Through.

  Ram Through's real text is “Target creature you control deals damage equal to its power to target creature you don't control. If the creature you control has trample, excess damage is dealt to that creature's controller instead.” (`crates/engine/src/parser/oracle_effect/tests.rs:1004-1012`). Its first selected object must become the damage source (`deal_damage.rs:1130-1141`) before the new gate is evaluated (`:1160-1165`). A direct resolver fixture proves that gate in isolation, but does not prove the parsed `TargetOnly → DealDamage` chain, cast target ordering, and source selection compose in the real pipeline. Add two `GameScenario` + `cast(..).resolve()` cases using Ram Through: trample must redirect excess to the damaged creature's controller; no trample must leave all damage on that creature.

## 🟡 Non-blocking

- The current parse-diff comment predates this head and required Rust/card-data checks are still running, so current-head parse-impact evidence is not yet available. This is not the reason for the request; it must be checked before approval.

## ✅ Clean

- `crates/engine/src/parser/oracle_effect/mod.rs:20285-20306` now unwraps the `TargetOnly` damage-dealer shape consistently for both lookup and mutation, resolving the former inert-path defect.
- `crates/engine/src/parser/oracle_effect/tests.rs:1003-1034` is the requested full `parse_oracle_text` regression and asserts the former `Unimplemented` residue is gone.
- `crates/engine/src/types/ability.rs:8763-8775` correctly parameterizes the old unconditional and keyword-gated recipient cases as one typed variant.
- Gemini's Hunter's Talent sibling example is not applicable: the repository card record at `data/mtgish-cards.json:13273` has its creature deal damage but no excess-damage redirect rider.

**Recommendation: add discriminating cast-pipeline Ram Through regressions for both trample states, wait for the current parse-diff/required checks, then request re-review.**

matthewevans (approved):
+***Approved: the current head clears the Ram Through correctness, feedback, and queue gates.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- Gemini's unresolved Hunter's Talent suggestion is stale and not applicable. [Scryfall's current Hunter's Talent Oracle text](https://api.scryfall.com/cards/e9a31863-9649-4a4f-99e4-c93729938bd7) contains no excess-damage redirect rider; its `that creature` damage instruction is therefore not a sibling of Ram Through's rider.

## ✅ Clean

- `crates/engine/src/parser/oracle_effect/sequence.rs:5867-5888` uses a full nom rider parser; `mod.rs:20294-20310` and `:27119-27168` correctly reach the nested `DealDamage` seam.
- `crates/engine/tests/integration/flame_spill_excess_damage.rs:599-621` has paired real-cast `GameScenario` regressions: trample redirects 3 excess damage, and no trample leaves all 5 on the creature.
- Head-matched CI run `29208679946` is green, including the Card data job; its parse-diff artifact is the intended one-card/one-signature Ram Through change. The local merge rehearsal against current `origin/main` was clean.

matthewevans (approved):

PR #5682, opened 2026-07-12
fix(parser): UNSUPPORTED full-set one-off: Unleash the Flux

ntindle:
## Summary

**Root cause:** UNSUPPORTED full-set one-off: Unleash the Flux

## Cards corrected
- Unleash the Flux

## Fix
Implemented the approved plan for cluster 99 (Unleash the Flux, a Planechase Phenomenon) surgically. The card previously misparsed: "…then you flip a coin. If you lose the flip, repeat this process." routed "repeat this process" into FlipCoin.lose_effect as Unimplemented{"empty"}, silently dropping the loop.

Built one class-level primitive across all layers: a resolution-scoped coin-flip-result condition feeding the existing RepeatContinuation::WhileCondition loop, plus a nom parser reroute.

Changes:
- types/game_state.rs: new Copy struct ResolutionCoinFlip { flipper: PlayerId, won: bool } + Option field on GameState (serde skip-if-None), constructor init, and manual field-equality arm.
- types/ability.rs: new leaf variant AbilityCondition::CoinFlipOutcome { result: CoinFlipResult } (parameterized by the existing CoinFlipResult enum — no sibling split; distinct from EventOutcomeWon which reads the trigger event, not resolution state).
- game/effects/flip_coin.rs: write state.resolution_coin_flip at both single-flip authorities (flip_through_replacement and the Krark resume_after_keep path).
- game/effects/mod.rs: evaluate_condition arm (controller-relative read of resolution_coin_flip); reset the tracker at each WhileCondition iteration start; condition_depends_on_effect_performed exhaustive-union arm.
- game/ability_rw.rs, game/ability_scan.rs, game/coverage.rs: exhaustive-match registration arms mirroring EventOutcomeWon.
- parser/oracle_effect/conditions.rs: new nom stripper strip_coin_flip_conditional (tag/alt/value only) + ability_condition_to_static_condition union arm.
- parser/oracle_effect/mod.rs: reroute in try_parse_repeat_process_directive to try the flip stripper first. It is body-gated: if the trailing body is not "repeat this process", the whole function returns None and the chunk falls through to the inline coin-flip fold, so Krark/Desperate Gambit/Ral Zarek win/lose branches are untouched.

Result: Unleash the Flux parses with 0 Unimplemented and repeat_until = WhileCondition{ CoinFlipOutcome{Lost} }, sub_ability FlipCoin{win:None,lose:None}. Runtime proof: while_condition_lost_flip_loops_until_win drives resolve_ability_chain through the real WhileCondition loop and asserts the loop repeats while lost and stops on a win, with resolution_coin_flip recording the winning flip. Added 9 tests total (3 stripper, 2 directive incl. non-repeat regression, 1 full-card, flip-records, controller-relativity, loop runtime proof).

Verification (cargo run directly — worktree not under Tilt): cargo fmt clean; cargo clippy -p engine --all-targets -D warnings PASS (exit 0); cargo test -p engine 2732 passed/0 failed; full gen-card-data.sh + semantic-audit confirm the card is clean corpus-wide. Krark/Sin/Claim Jumper regression tests pass.

Judgement calls: (1) check-parser-combinators.sh exits 1, but that is a PRE-EXISTING condition — its fixed diff baseline is PR #904 (far behind HEAD fa199f802), so it flags string-dispatch accumulated across untouched files (oracle.rs/sequence.rs/oracle_trigger.rs). Confirmed none of the flagged lines are mine and my added parser lines contain zero string dispatch. (2) Runtime proof strategy: I proved the NEW primitive (flip→condition→WhileCondition loop, controller-relativity, field recording) via focused cast/resolution-pipeline tests rather than building the full multi-player-sacrifice + phenomenon-encounter integration harness, because that composition's remaining surface (player-scope sacrifice fan-out, pending_continuation-before-pending_repeat_until drain ordering) is pre-existing, separately-tested machinery that my change does not modify; the parse test proves the full-card structure and the loop test proves the mechanism. (3) gen-card-data.sh incidentally regenerated two TRACKED data files (known-tokens.toml −9032 lines, oracle-subtypes.json +1) from my local mtgjson/token data differing from the committed baseline — unrelated to the fix. I restored both to HEAD (isolated upstream/main worktree, no other agent's uncommitted work here). card-data.json and semantic-audit outputs are gitignored.

No STOP-AND-RETURN, no scope expansion, no new subsystem — extension only.

## Files changed
- crates/engine/src/game/effects/flip_coin.rs

## CR references
- CR 705.2 — only the player who flips a coin wins or loses the flip; controller-relative gate (verified: grep -n "^705.2" docs/MagicCompRules.txt → "Only the player who flips the coin wins or loses the flip")
- CR 705.1 — flip a coin (verified: grep -n "^705.1" docs/MagicCompRules.txt → "Some cards refer to flipping a coin")
- CR 608.2c — follow instructions in order / read the whole text; home of the 'repeat this process' loop and the resolution-scoped read (verified: grep -n "^608.2c" docs/MagicCompRules.txt)
- CR 104.4b — mandatory-action loop draw; cited in the GameState field-equality comment noting resolution_coin_flip is volatile and never masks a real repeat (verified: grep -n "^104.4b" docs/MagicCompRules.txt)

## Verification
- `cargo fmt --all` — pass
- `check-parser-combinators.sh (base=upstream/main merge-base fa7f157)` — pass
- `oracle-gen data --filter 'unleash the flux'` — pass
Cards confirmed re-parsed correctly: unleash the flux


matthewevans (changes requested):
**Verdict: blocked — the new coin-flip result state can outlive its resolution, and the affected card lacks a production-pipeline regression test; request changes before merge.**

## 🔴 Blocker

`crates/engine/src/game/effects/mod.rs:6062` clears `resolution_coin_flip` only at the start of a `WhileCondition` iteration, while `crates/engine/src/game/effects/mod.rs:8633` reads it whenever `CoinFlipOutcome` is evaluated. The state is documented as resolution-scoped but can therefore survive a completed resolution and affect a later evaluation. CR 608.2c requires that the controller follows the instructions in their written order; the card text says, “each player sacrifices a nonland permanent, then you flip a coin. If you lose the flip, repeat this process.” Move the reset to the authoritative resolution lifetime, including continuation completion, rather than one repeat branch.

`crates/engine/src/game/effects/flip_coin.rs:1617` proves only a bare flip/repeat, while `crates/engine/src/parser/oracle_effect/tests.rs:41813` checks Unleash the Flux only as an AST. Neither test reaches the real each-player sacrifice choices, continuation resume, and repeat decision. Add a registered runtime scenario that drives those choices and demonstrates that an actual loss repeats the whole process and a win stops it.

## 🟡 Non-blocking

`crates/engine/src/types/game_state.rs:1154` introduces `ResolutionCoinFlip { won: bool }`, even though `crates/engine/src/types/ability.rs:17814` already defines `CoinFlipResult`. Reuse that typed domain vocabulary so the stored result and `AbilityCondition::CoinFlipOutcome` cannot drift.

## ✅ Clean

The current parse-diff is scoped to Unleash the Flux: it removes the previous `Unimplemented` lose branch and adds the intended coin-flip continuation. CR 705.2 was verified: only the player who flips the coin wins or loses it.

**Recommendation: request changes — give the flip result a true resolution/continuation lifetime, add the interactive runtime regression, and store the existing `CoinFlipResult` type.**

matthewevans (approved):
**Approve — the current head resolves the requested coin-flip lifetime, typed-state, and runtime-coverage gaps; ready for the merge queue.**

## 🔴 Blocker

- None. The prior requested changes are resolved on `ae42ff638d82c2769f36149486d67d6b930207d9`.

## 🟡 Non-blocking

- None.

## ✅ Clean

- `crates/engine/src/game/effects/mod.rs:5983-6002` establishes the top-level resolution lifetime and `:6078-6085` resets each `WhileCondition` iteration, preventing both cross-resolution and cross-iteration leakage.
- `crates/engine/src/types/game_state.rs:1157-1161` stores the existing `CoinFlipResult`, and `crates/engine/src/game/effects/flip_coin.rs:87-90,464-467` records it at both normal and kept-flip authorities.
- `crates/engine/src/game/effects/flip_coin.rs:1741-1884` drives the parsed Unleash the Flux ability through real sacrifice selections, a losing flip that repeats, and a winning flip that stops. Current CI is green and the parse artifact is exactly one card with two expected signatures.

**Recommendation: approve and enqueue with the existing `bug` label.**

PR #5696, opened 2026-07-12
feat(engine): place conjure into the top N cards of a library at random

andriypolanski:
## Summary

`Effect::Conjure` modeled its destination at **`Zone` granularity** only, so
Alchemy conjures that name a *positional* library slot ("conjure a duplicate of
that creature **into the top five cards of your library at random**") lost that
constraint. The parser arm added in #5610 parsed the count and then discarded
it, collapsing the placement to `Zone::Library`. Because `zones::create_object`
pushes new library objects to the **bottom**, the conjured card landed at the
deterministic end of the library — drawn last / effectively never — instead of a
uniformly-random slot among the top N.

This change threads the count through to the resolver so the card lands where the
rules text says it should. It touches two engine surfaces (a new field plus a new
enum variant), which is why it was deferred out of the parser-only PR #5610.

## Changes

- **`LibraryPosition::RandomWithinTop { n: QuantityExpr }`** — new variant for
  "a uniformly random slot among the top N." Digital-only Alchemy has no CR
  entry, so it is annotated as such rather than citing a fabricated rule. `n` is
  a `QuantityExpr` to mirror `BeneathTop`'s dynamic depth.
- **`Effect::Conjure { …, library_position: Option<LibraryPosition> }`** — new
  field mirroring `Effect::ChangeZone::library_position`. `#[serde(default,
  skip_serializing_if = "Option::is_none")]` keeps old serialized conjures and
  existing card-data output byte-stable.
- **Parser** (`oracle_effect/mod.rs`) — `parse_conjure_zone` now returns the
  optional `LibraryPosition`; the existing `" into the top N cards of … at
  random"` combinator arm populates `RandomWithinTop` instead of throwing the
  count away. Both the named and duplicate conjure paths thread it through.
- **Resolver** (`effects/conjure.rs`) — after `create_object`, a positional
  library conjure repositions the brand-new card at the resolved slot (a direct
  within-library reposition; no cross-zone cleanup / incarnation bump is needed).
- **Shared authority** (`zones::random_top_slot_index`) — one place computes the
  uniformly-random top-N index; the conjure resolver and both `zone_pipeline`
  exhaustiveness arms call it so the behavior is identical everywhere.
- Exhaustive `LibraryPosition` matches in `zone_pipeline.rs` and the
  precise-destructure arm in `ability_rw.rs` updated (no wildcard fallbacks).

## Affected cards (6+)

Goblin Morale Sergeant, Jessie Zane, Fangbringer, Mine Security, Sheoldred's
Assimilator, Sliver Weftwinder.

## Tests

- Parser: `conjure_duplicate_into_top_n_at_random_captures_position`,
  `conjure_named_into_top_n_each_player_at_random_captures_position`,
  `conjure_plain_library_has_no_position`.
- Resolver: `conjure_random_within_top_lands_in_top_n_not_bottom` — asserts the
  invariant (conjured card index `< N`, never the bottom) across 32 seeds.

## Test plan

- [ ] `cargo fmt --all`
- [ ] `clippy` / `test-engine` green (Tilt)
- [ ] `card-data` regenerates cleanly (new `library_position` field + variant)


matthewevans (changes requested):
**Request changes: the new positional-conjure path gets the slot calculation right for one library, but drops player scope and layer invalidation for actual library-top effects.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/mod.rs:9178-9208` deliberately accepts “each player's library”, but `crates/engine/src/game/effects/conjure.rs:224-279` stores no destination-player scope and always uses `ability.controller` as `owner`. Sandcloud Harbinger says, “conjure three cards named Sunscorched Desert into the top ten cards of each player's library at random”; the current resolver instead creates the three cards only in the ability controller's library.

  Model the destination player scope on `Effect::Conjure` and iterate every affected library at resolution, rather than parsing the plural phrase into the controller-only form. Add a multiplayer runtime regression that asserts three new Deserts in every player's top ten.

- `crates/engine/src/game/effects/conjure.rs:143-151` and `:232-279` directly remove and insert a card in `player.library`, including at index `0`, without marking layers dirty. A cached `TopOfLibraryMatches` static therefore continues to evaluate the old top card after a random-top-N conjure picks the top slot.

  Route this through the shared library-reposition/invalidation authority, or explicitly mark layers after the direct reorder, and add a production-path regression with an active top-of-library static. The existing top-N invariant test does not exercise a clean layer cache.

- `crates/engine/src/game/zone_pipeline.rs:627-631` and `:1707-1711` pass `top_n` as both the requested range and available slot count. When a `RandomWithinTop` request reaches either generic pipeline with fewer than `top_n` slots, `move_to_library_at_index` clamps a value sampled from the larger range, biasing toward the bottom. Either compute the real destination-library slot count before calling `random_top_slot_index`, or keep this conjure-only placement out of generic pipeline paths until they can resolve the full context.

## 🟡 Non-blocking

- The current head has failed Rust/card-data checks and no current parse-diff artifact. This confirms it is not approval-ready, but the behavioral gaps above are independently merge-blocking.

## ✅ Clean

- `crates/engine/src/game/effects/conjure.rs:259-267` correctly samples from the actual number of slots for the controller-only conjure path; the single-library top-N invariant test is a useful base case.
- `crates/engine/src/types/ability.rs:8654-8669` uses a typed `LibraryPosition::RandomWithinTop` parameter rather than a one-off boolean or raw count.
- The parser and test correctly preserve a positional top-N constraint for the singular “your library” form.

**Recommendation: request changes—carry destination player scope, invalidate top-of-library layers after direct reorders, and make generic pipeline sampling use the actual library size before re-review.**

matthewevans (changes requested):
**Changes requested — the current head keeps the prior behavioral blockers unresolved.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/mod.rs:9180-9205` accepts both `your library` and `each player's library`, while `crates/engine/src/game/effects/conjure.rs:45-107` creates every card with `ability.controller` and `:222-266` reorders only that controller's library. Sandcloud Harbinger's verified Oracle text is: “When Sandcloud Harbinger enters, conjure three cards named Sunscorched Desert into the top ten cards of each player's library at random.” The current model has no destination-player scope, so it necessarily puts all three cards into one library. Carry a typed player scope through `Effect::Conjure` and resolve once per affected player; add a multiplayer runtime regression asserting three cards in every affected library.

- `crates/engine/src/game/effects/conjure.rs:222-266` directly removes and inserts a library card but does not call the existing `layers::mark_layers_full_if_top_of_library_static_live` authority used after library-top changes. A cached `TopOfLibraryMatches` static can therefore retain the former top card when this reordering selects index zero. Reuse that invalidation seam and add a production-path regression with a live top-of-library static.

## 🟡 Non-blocking

- The `8f001f6` delta only adjusts `CardId` typing/formatting and discards library-position data for spellbook conjure; it does not address either current behavior above.

## ✅ Clean

- The current top-N sampling uses the actual number of available slots, addressing the earlier small-library distribution concern.
- `LibraryPosition::RandomWithinTop` remains a typed position parameter rather than a special-case boolean.

**Recommendation: carry destination player scope and route the reorder through the existing library-top invalidation authority, with multiplayer and live-layer regressions, before re-review.**

matthewevans:
**Maintainer update: held for current-head CI.**

I reviewed the fresh head and pushed `ea100b8507` to keep `RandomWithinTop` at its Conjure-only resolver. The generic zone-pipeline arms were unreachable for this feature, could not resolve the actual destination-library size, and the first also caused the Rust gate to fail. The current resolver continues to carry `PlayerFilter::All` for Sandcloud Harbinger's each-player libraries and invalidates live top-of-library statics after reordering.

The replacement checks are running; I will not clear the prior review state or enqueue until they are complete on this head.

matthewevans (approved):
**Approved for merge queue.** The current head resolves the requested each-player library fan-out and top-of-library layer invalidation, and the maintainer fix keeps random top-N placement at the Conjure resolver rather than an under-contextualized generic zone pipeline. The current parse-diff reports no card-parse delta; required checks may finish in the queue.

Issue #1146, opened 2026-05-26
Gonti, Lord of Luxury — When Gonti is played from the Command Zone, the game automatically puts him into Exile with no…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1508576425995669725/1508576425995669725

**Thread:** Gonti, Lord of Luxury
**Cards:** the command zone
**Parser status:** has_gaps
**Extraction confidence:** 0.90

## Summary
When Gonti is played from the Command Zone, the game automatically puts him into Exile with no trigger or anything.

### the command zone
When you planeswalk here, each player who controls their commander draws a card. Other players may put their commander from the command zone onto the battlefield.
If an ability of your commander would trigger during your turn, that ability triggers an additional time.
Whenever chaos ensues, the number of times each commander has been cast from the command zone becomes 0. (This resets commander tax.)

---
<sub>report_id: `discord:1508576425995669725:1508576425995669725:0` · discord: `1508576425995669725/1508576425995669725`</sub>

matthewevans:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `When ~ enters, look at the top four cards of target opponent's library, exile one of them face down, then put the rest on the bottom of that library in a random order.`  (supported: true)

**Reasoning**: The ChangeZone child inside the Dig chain encodes `target: "triggering source"` — within a ChangesZone trigger that fires on Gonti entering, "triggering source" resolves to Gonti himself, not to the chosen card from the Dig. This causes the engine to exile Gonti immediately upon entry instead of exiling one of the revealed library cards. The correct target should reference the player-selected card from the Dig's tracked set.

---
_Posted by bug-coverage-classifier skill._

PR #5701, opened 2026-07-12
fix(parser): thread the counter rider into the 'exile one of them face down' fusion

### LINKED ISSUE #5631: The Dragon-Kami Reborn: Saga exiles ITSELF instead of a looked-at card (silent #1146-class misparse)
## Summary
**Silent misparse — coverage reports this card as SUPPORTED (zero `Unimplemented` nodes), but chapters I–II resolve wrong.** Surfaced by the U20 walk-back binding campaign (task #20 harvest, full-pool sweep of all 35,034 oracle texts).

## Actual parse
```
Dig { keep_count: 0, reveal: false }            <- pure peek, keeps NOTHING
  └─ ChangeZone { destination: Exile, target: ParentTarget, enter_with_counters: [("hatching", 1)] }
```
This is verbatim the shape documented at `sequence.rs:4391` as the bug the Hideaway fusion exists to prevent (#1146): the Dig short-circuits as a keep_count:0 pure-peek and the sibling `ChangeZone { ParentTarget }` exiles **the trigger source itself** — the Saga exiles itself (with a hatching counter) instead of one of the three looked-at cards. #1146 is live on this card.

## Cause
The counter rider ("with a hatching counter on it") sits between "face down" and the sentence end, so the `ExileOneOfThemFaceDown` recognizer declines — honestly at the recognizer, but the fallback parse is silently wrong rather than a visible gap.

## Fix shape
Recognizer extension, not new engine work: `enter_with_counters` is already carried correctly on the ChangeZone. The recognizer needs to accept an optional counter rider between "face down" and clause end and thread it into the fused Hideaway-style exile.

## Why this matters beyond the card
This is the **supported-but-misparsed** class — invisible to every existing gate (no Unimplemented node, byte-identity gates can't see it). Full-pool sweep population: "exile one of them face down" matches 5 cards; 4 fire the recognizer correctly, this is the only decline.

andriypolanski:
Closes #5631

## Summary

The Dragon-Kami Reborn's chapters I–II read: *"…Look at the top three cards of
your library. **Exile one of them face down with a hatching counter on it**,
then put the rest on the bottom of your library in any order."*

Coverage reported this card as **supported** (zero `Unimplemented` nodes), but it
resolved wrong — a **silent misparse** of the #1146 class. The
`parse_exile_one_of_them_face_down` recognizer required end-of-clause immediately
after `"face down"`, so the CR 122.1 counter rider (`"with a hatching counter on
it"`) made it **decline**. With the recognizer declining, the Hideaway-style
fusion never fired: the `Dig` short-circuited as a `keep_count: 0` pure-peek and a
sibling `ChangeZone { ParentTarget }` exiled the **trigger source itself** (the
Saga, with a hatching counter) instead of one of the three looked-at cards.

Per the issue, this is a **recognizer extension, not new engine work**: the fix
lets the recognizer accept an optional counter rider between `"face down"` and the
clause end, captures those counters, and threads them onto the **chosen** dug card
in the fused exile.

## Changes

- **Recognizer** (`oracle_effect/sequence.rs`, `parse_exile_one_of_them_face_down`)
  — now returns `Option<Vec<(CounterType, QuantityExpr)>>` instead of `bool`.
  `Some(entries)` signals a match (empty = the bare Gonti form; non-empty = the
  rider's counters). The optional rider is parsed by delegating to the shared
  `parse_counter_suffix_body_combinator`, so the **whole class** of
  `"with <count> <type> counter(s) on it"` shapes (a/two/X, `+1/+1`, keyword,
  named) is covered — not just `"a hatching counter"`.
- **Continuation** (`oracle_ir/ast.rs`) — `ContinuationAst::ExileOneOfThemFaceDown`
  gains a `enter_with_counters: Vec<(CounterType, QuantityExpr)>` field
  (`#[serde(default, skip_serializing_if = "Vec::is_empty")]` keeps the bare-form
  serialization byte-stable).
- **Fusion** (`oracle_effect/sequence.rs`, `apply_clause_continuation`) — after the
  existing `append_conceal_sub_ability`, a new `append_counter_riders_sub_abilities`
  appends a `PutCounter { target: ParentTarget }` chain at the deepest sub. Because
  the `DigChoice` resolution binds the chosen (exiled) card onto the continuation
  chain's targets and `ParentTarget` sub-abilities inherit the parent's target
  (mirroring `lower_put_counter_list`), the counters land on the **one card the
  player selected**, not the Saga.
- Exhaustive `ContinuationAst` match/`matches!` sites (`produces_sibling_def`,
  the `dig_continuation_*` helper in `oracle_effect/mod.rs`) updated to the new
  struct-variant shape (no wildcard fallbacks).

## Why this is the right shape

`enter_with_counters` was already parsed correctly onto the (buggy, discarded)
sibling `ChangeZone`; the only defect was that the recognizer's `eof` gate
rejected the rider so the fusion never ran. `Dig` carries no counter field and
`HideawayConceal` carries only a target, so the rider is applied with the existing
`PutCounter` building block chained onto the same `ParentTarget` the conceal
already uses — no new engine variant or field, and `apply_counter_addition`
already places counters on objects regardless of zone (exile included).

## Affected cards

`"exile one of them face down"` matches 5 cards in the full-pool sweep; 4 already
fired the recognizer correctly, and The Dragon-Kami Reborn was the sole decline.
The generalized rider combinator also future-proofs any later choose-one exile
that attaches a counter.

## Tests

- Recognizer: `exile_one_of_them_face_down_with_counter_rider_captures_counters`
  (the hatching rider is captured), `exile_one_of_them_face_down_bare_form_has_no_counters`
  (bare form is a match with an empty rider), plus the updated
  `exile_one_of_them_without_face_down_is_not_gonti_continuation` /
  `delver_pure_peek_is_not_gonti_continuation` scope guards.
- Fusion: `exile_one_of_them_face_down_counter_rider_appends_put_counter_after_conceal`
  — asserts the fused chain is `Dig(keep_count:1, Exile) -> HideawayConceal(ParentTarget)
  -> PutCounter(hatching, ParentTarget)`.

## Test plan

- [ ] `cargo fmt --all`
- [ ] `clippy` / `test-engine` green (Tilt)
- [ ] `card-data` regenerates cleanly (The Dragon-Kami Reborn now parses to the
      choose-one exile + conceal + hatching-counter shape, not a source self-exile)

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -21607,7 +21607,10 @@ fn chain_clause_is_exile_producer(effect: &Effect) -> bool {
 fn continuation_is_exile_producer(continuation: Option<&ContinuationAst>) -> bool {
     matches!(
         continuation,
-        Some(ContinuationAst::ExileOneOfThemFaceDown { .. } | ContinuationAst::ExileLookedAtCard { .. })
+        Some(
+            ContinuationAst::ExileOneOfThemFaceDown { .. }
+                | ContinuationAst::ExileLookedAtCard { .. }
+        )
     )
 }
 
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -21607,7 +21607,7 @@ fn chain_clause_is_exile_producer(effect: &Effect) -> bool {
 fn continuation_is_exile_producer(continuation: Option<&ContinuationAst>) -> bool {
     matches!(
         continuation,
-        Some(ContinuationAst::ExileOneOfThemFaceDown | ContinuationAst::ExileLookedAtCard { .. })
+        Some(ContinuationAst::ExileOneOfThemFaceDown { .. } | ContinuationAst::ExileLookedAtCard { .. })
     )
 }
 
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -834,17 +834,41 @@ fn parse_exile_looked_at_card(lower: &str) -> Option<bool> {
 /// selects exactly one of the N looked-at cards. The "face down" suffix is the
 /// CR 406.3 hidden-information marker required by this class; pure-peek Digs
 /// (Delver of Secrets — no exile clause) never reach this recognizer.
-fn parse_exile_one_of_them_face_down(lower: &str) -> bool {
+///
+/// CR 122.1: An optional counter rider ("... face down with a hatching counter
+/// on it", The Dragon-Kami Reborn) may follow "face down". `Some(entries)`
+/// signals a match (empty `entries` = the bare Gonti form; non-empty = the
+/// counters the fusion threads onto the chosen exiled card). Delegates the
+/// rider to the shared `parse_counter_suffix_body_combinator`, so every
+/// `<count> <type> counter(s) on it` shape (a/two/X, +1/+1, keyword, named)
+/// is covered by the one building block — not just "a hatching counter".
+fn parse_exile_one_of_them_face_down(lower: &str) -> Option<Vec<(CounterType, QuantityExpr)>> {
     let trimmed = lower.trim().trim_end_matches('.').trim_end();
-    (
+    let (rest, _) = (
         tag::<_, _, OracleError<'_>>("exile "),
         alt((tag("one of them"), tag("one of those cards"))),
         multispace1,
         tag("face down"),
-        eof,
     )
         .parse(trimmed)
-        .is_ok()
+        .ok()?;
+    // CR 122.1: optional "with <count> <type> counter(s) on it" rider. Either
+    // the bare form (nothing after "face down") or exactly one counter clause.
+    match preceded(
+        tag::<_, _, OracleError<'_>>(" with "),
+        crate::parser::oracle_effect::parse_counter_suffix_body_combinator,
+    )
+    .parse(rest)
+    {
+        Ok((after, entry)) => {
+            eof::<_, OracleError<'_>>(after).ok()?;
+            Some(vec![entry])
+        }
+        Err(_) => {
+            eof::<_, OracleError<'_>>(rest).ok()?;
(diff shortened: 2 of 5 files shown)


matthewevans (changes requested):
**Request changes — the counter-rider fusion is at the right parser seam, but no runtime regression proves the counter reaches the player-selected exiled card instead of the Saga/source.**

## 🔴 Blocker

- `parser/oracle_effect/sequence.rs:9389-9447` constructs `Dig → HideawayConceal → PutCounter { ParentTarget }`, and the added coverage only inspects that shape. The reported failure is resolution behavior: “Exile one of them face down with a hatching counter on it” must select one looked-at card, exile that card face down, and put the hatching counter on that card—not on The Dragon-Kami Reborn or a discarded sibling. Add a `GameScenario`/resolution-path regression that drives the choice and asserts the selected exile object’s counter plus the Saga’s continued zone. This must fail when the fusion or the counter-target binding is reverted.

## 🟡 Non-blocking

- The optional rider uses the shared counter-suffix combinator and correctly preserves the bare Gonti form; keep that generalization.

## ✅ Clean

- `parse_exile_one_of_them_face_down` now rejects only nonmatching clauses and distinguishes a bare match from a match with counter entries.
- The current parse artifact is contained to The Dragon-Kami Reborn across five expected signatures; no unrelated card changed.

**Recommendation: request changes — add the discriminating real-resolution regression, then re-review this as a `bug` fix.**

matthewevans (approved):
**Approve — the counter-rider fusion now has a discriminating real-resolution regression at the existing Dig continuation seam.**

## 🔴 Blocker

- None. The prior missing runtime test is addressed on `dade59e`: it resolves the full chapter, submits `DigChoice`, and proves the chosen card—not the Saga—receives the hatching counter.

## 🟡 Non-blocking

- Gemini is quota-limited this session; the independent local/adversarial review was completed against the current head instead.

## ✅ Clean

- The shared counter-suffix combinator keeps the bare Gonti form intact while covering the counter-rider class.
- The current parse artifact is contained to The Dragon-Kami Reborn (one card, five expected signature changes).
- Current CI is green, and the runtime test would fail if the fusion or `ParentTarget` binding were reverted.

PR #5702, opened 2026-07-12
fix(parser): keep the mana-value gate on typed cost-modifier spell filters

### LINKED ISSUE #5606: [Card Bug] The Scarlet Witch: Everything becomes cheaper to cast instead of only instants and sorceries
### Card(s)

The Scarlet Witch

### Build/version

0.21.0 a61e7fd

### Game mode

AI

### Actual behavior

When I had Scarlet Witch on the battlefield casting creatures cost the amount of power of Scarlet Witch less to cast

### Expected behavior

Only instant and sorceries should be cheaper to cast

### Steps to reproduce

- Have The Scarlett Witch on the battlefield
- Cast a creature

_No response_

minion1227:
## Summary

Fixes the dropped-spell-filter misparse for **mana-value-gated cost modifiers** (`Closes #5606`). "`[type]` spells you cast with mana value N or greater/less cost {N} less" was lowering to a `ModifyCost` static with `spell_filter: null` and `affected: [Card]`, so the reduction applied to **every** spell. The Scarlet Witch reduced the cost of creatures (and everything else) instead of only instant/sorcery spells with mana value 4 or greater; Krosan Drover, Spectacle Mage, Thryx, and It That Heralds the End share the bug.

Root cause: the trailing "with mana value N or greater" gate sits *after* the "spells you cast" infix, so the type-prefix end-trims (`" you cast"`, `" spells"`) never reached the type words and `parse_type_phrase` was handed the whole tail → `None` → both the type restriction and the mana-value gate were silently dropped.

## Files changed

- `crates/engine/src/parser/oracle_static/static_helpers.rs`

## CR references

- CR 202.3 — mana value; the "with mana value N or greater/less" gate selects spells by mana value.
- CR 601.2f — cost-modification statics apply during the cost-determination step of casting.

Method: not-applicable — authored directly against the `oracle-parser` skill. Pure parser change; no card-data regeneration required (AST-shape fix).

## Track

Developer

## LLM

Model: claude-opus-4-8
Thinking: high

## Verification

- [x] Verification below is for the current committed head (`a1e303414`).

Results on head `a1e303414`:

- `cargo test -p engine cost_mod` — **ok. 9 passed; 0 failed** (4 new: instant/sorcery+MV≥, single-type+MV≤, bare+MV≥, and a no-MV regression; 5 pre-existing cost-mod tests unaffected).
- `cargo clippy -p engine --all-targets --features proptest -- -D warnings` — **Finished, 0 warnings** (exit 0).
- `cargo fmt --all` — clean.

## Anchored on

- `crates/engine/src/parser/oracle_static/static_helpers.rs` — the existing spell-filter derivation this extends: the "with the chosen name" / "of the chosen type" qualifiers are peeled off the subject prefix *before* the type trims and re-attached as `FilterProp`s afterward. The mana-value gate follows the identical peel-then-compose shape.
- `crates/engine/src/parser/oracle_static/static_helpers.rs` (`parse_cost_mod_spell_type_prefix` → `parse_that_clause_suffix`) — the analogous "that has `<property>`" spell-qualifier → `FilterProp` handling; `FilterProp::Cmc` is the same mana-value property that path already emits.

## Final review

Self-review against the review-impl lenses. The peel is a no-op when "with mana value " is absent (`parse_cost_mod_mana_value_qualifier` fails → `strip_*` returns `(prefix, None)`), so cards without the qualifier are unaffected — confirmed by the no-MV regression test and the 5 unchanged pre-existing cost-mod tests. `compose_cost_mod_mana_value` handles all three filter shapes (single `Typed` → absorb prop; multi-type `Or` → `And`-wrap; `None`/bare → card+prop). Both "or greater/or more" (GE) and "or less/or fewer" (LE) comparators covered; the value is a `QuantityExpr::Fixed`.

## Claimed parse impact

- The Scarlet Witch
- Krosan Drover
- Spectacle Mage
- Thryx, the Sudden Storm
- It That Heralds the End

## Validation Failures

None.

## CI Failures

None.


matthewevans (changes requested):
**Request changes — the parser seam and typed filter composition are sound, but the cost reduction has no runtime regression proving the mana-value gate reaches the casting pipeline.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_static/tests.rs:1472-1580` asserts only the parsed `ModifyCost.spell_filter`. The behavior being claimed is that “Instant and sorcery spells you cast with mana value 4 or greater cost {1} less to cast” changes what a player can pay during casting; an AST assertion can remain green even if `FilterProp::Cmc` is ignored in the cost resolver. Add a `GameScenario`/cast-pipeline regression with a qualifying instant or sorcery and an adjacent non-qualifying spell, proving the reduction applies only to the former. This should fail if the new qualifier composition is reverted.

## 🟡 Non-blocking

- None.

## ✅ Clean

- `parser/oracle_static/static_helpers.rs:53-126` uses composable nom parsing and the existing `FilterProp::Cmc` vocabulary; the `Typed`/`Or` composition avoids inventing a sibling spell-filter type.
- The parser tests cover typed, disjunctive, bare, and no-qualifier shapes, including the “or less” sibling.

**Recommendation: request changes — add the discriminating casting regression, then re-review this as a `bug` fix.**

minion1227:
Thanks — addressed the blocker in `5b01b11cc`.

Added a discriminating runtime regression: `crates/engine/src/game/casting_costs.rs::mana_value_gated_cost_reduction_reaches_cost_resolver`. It parses the real static line onto a battlefield source, then drives the cost resolver via `display_spell_cost`:

- qualifying instant (MV 5 ≥ 4) → reduced `{5}` → `{4}`
- sub-threshold instant (MV 3 < 4) → **unchanged** `{3}`
- off-type creature (MV 5) → **unchanged** `{5}`

It flips if the parser fix is reverted (restoring `spell_filter: null` makes the resolver reduce all three), so it proves the `FilterProp::Cmc` gate reaches the casting pipeline end-to-end — not just the AST. `cargo clippy -p engine --all-targets --features proptest` is clean. Ready for re-review as a bug fix.

matthewevans (changes requested):
**Request changes — the parser and resolver-level assertions are sound, but the requested casting-pipeline regression is still missing.**

## 🔴 Blocker

- `crates/engine/src/game/casting_costs.rs:11768-11771` calls `display_spell_cost` directly. That validates the cost-calculation helper, but it does not drive a `GameScenario` cast/payment transition through the production `GameAction`/casting flow that owns the behavior users exercise. Add paired qualifying and non-qualifying cast-pipeline cases (or an equivalent scenario-runner path) that prove the payable/cast result changes only for the MV/type-qualified spell. The regression must fail if the qualifier composition is reverted.

## 🟡 Non-blocking

- Gemini is quota-limited; the current-head local review and green required CI supply the independent evidence for this pass.

## ✅ Clean

- `static_helpers.rs` reuses `FilterProp::Cmc` and composes the typed/disjunctive filter shapes without a new vocabulary.
- The current artifact reports no unrelated parse changes, and the new resolver-level test covers qualifying, sub-threshold, and off-type inputs.

**Recommendation: request changes — add the discriminating cast/scenario regression, then re-review as `bug`.**

minion1227:
Addressed the re-review in `4b4b3313a`.

Added `mana_value_gated_cost_reduction_through_cast_pipeline` — a full `GameScenario` → `GameRunner::cast(..).resolve()` cast/payment regression (not the cost helper). A battlefield permanent grants the reduction; each instant is funded to its full printed cost and cast through the production pipeline, and the assertion is on the **mana actually paid** (leftover pool):

- qualifying instant (MV 5 ≥ 4): funded {5}, pays {4} → **1 mana left**
- sub-threshold instant (MV 3 < 4): funded {3}, pays {3} → **0 left**

The payable/pay result changes only for the MV-qualified spell, and the MV-3 assertion flips to `1` if the qualifier composition is reverted (`spell_filter → null` reduces every spell). The earlier resolver-level test is kept for the type/off-type coverage. `cargo clippy -p engine --all-targets --features proptest` clean.

matthewevans (approved):
**Approve — the mana-value qualifier now reaches the production cast/payment path; current `main` has been ported once to refresh the pending parse baseline.**

## 🔴 Blocker

- None. The prior cast-pipeline gap is resolved at `crates/engine/src/game/casting_costs.rs:11804-11858`: the MV-5 instant leaves one mana after `GameRunner::cast(..).resolve()`, while the MV-3 sibling leaves none; reverting the parser filter composition flips the latter assertion.

## 🟡 Non-blocking

- The parse-diff sticky comment remains baseline-pending immediately after the maintainer port. The loop’s current-head recommendation is `baseline_pending_after_maintainer_merge_ready`; the merge queue will wait for the newly triggered checks rather than adding another noisy port.
- Gemini is quota-limited; the current-head local review and the discriminating scenario test supply the independent evidence.

## ✅ Clean

- `crates/engine/src/parser/oracle_static/static_helpers.rs:60-128` retains the existing `FilterProp::Cmc` vocabulary, composes typed/disjunctive filters correctly, and keeps the no-qualifier path unchanged.
- The source port passed `cargo fmt --all` and the parser combinator gate; the branch contains no workflow, instruction, or generated-file changes.

**Recommendation: approve and enqueue as `bug`; the merge queue should wait for the current-head checks.**

PR #5703, opened 2026-07-12
fix(parser): route 'that player investigates' Clue to the target's controller (#5254)

### LINKED ISSUE #5254: Declaration in Stone — My opponent cast Declaration In Stone and successfully exiled my creature, but it created the cl…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522631993685311589/1522631993685311589

**Thread:** Declaration in Stone
**Discord thread id:** `1522631993685311589`
**Discord message id:** `1522631993685311589`
**Cards:** declaration in stone
**Parser status:** has_gaps

## Summary
My opponent cast Declaration In Stone and successfully exiled my creature, but it created the clue for THEM and not me

### declaration in stone
Exile target creature and all other creatures its controller controls with the same name as that creature. That player investigates for each nontoken creature exiled this way.

---
<sub>report_id: `discord:1522631993685311589:1522631993685311589:0` · discord: `1522631993685311589/1522631993685311589`</sub>

jaytbarimbao-collab:
Tier: Frontier
Model: claude-opus-4-8

Fixes #5254

## Summary

Declaration in Stone's second sentence — *"That player investigates for each nontoken creature exiled this way"* — was mis-executed on two axes: the Clue went to the **caster** instead of *that player* (the controller of the exiled creature, CR 109.4 / CR 608.2h), and the **count** was dropped so exactly one Clue was made regardless of how many creatures were exiled. Both are now fixed: the exiled creatures' controller gets **one Clue per nontoken creature exiled this way**.

Root cause for both: the predicate lowers to the **fieldless** unit variant `Effect::Investigate`, which has neither a player slot for `inject_subject_target` to stamp the subject onto nor a count slot for the `for each … this way` suffix — so both were silently dropped.

The recipient fix is also a class fix — it corrects **Fateful Absence** (*"Its controller investigates"*, single target, no count). A bare *"investigate"* with no player subject and no count (Press for Answers, Hold for Questioning) stays caster-default, single Clue.

## Implementation method

Two symmetric lifts off the fieldless `Effect::Investigate`, both stamped in `lower_subject_predicate_ast` and folded into the chunk's `AbilityDefinition` in the effect-chain loop (no enum churn):

1. **Recipient → `player_scope`.** A transient `ParseContext.pending_player_scope` records the parent-target player anaphor (`ParentTargetController`/`Owner` → `PlayerFilter::ParentObjectTargetController`/`Owner`).
2. **Count → `repeat_for`.** A transient `ParseContext.pending_repeat_for` records the `for each <filter> <cause> this way` suffix, parsed by the shared `parse_for_each_clause` (the parser the `for each …, <effect>` prefix path already uses), accepting **only** `FilteredTrackedSetSize { caused_by: Some(_) }`. Scoped to `Effect::Investigate`, so count-bearing `for each … this way` effects (gain life / draw / counters) are untouched.
3. **Engine driver arm** (`game/effects/mod.rs`) — `ParentObjectTargetController` / `Owner` route through the ability-aware `speed_effects::players_for_filter` (mirroring the existing `AllExcept` arm), then re-impose APNAP. Previously these hit the generic `matches_player_scope` arm and produced an empty fan-out.

Runtime composes with **no engine change to the count**: the `ChangeZone → ChangeZoneAll` chain publishes one **merged** exile tracked set, and `repeat_for` drives **outside** `player_scope` (the Torment of Hailfire shape), so N nontoken creatures exiled ⇒ N Clues to the exiled creatures' controller.

## CR references

- **CR 109.4 / CR 608.2h** — "that player" / "its controller" = the controller of the object targeted earlier in the instruction, resolved via last-known information after it leaves the battlefield.
- **CR 701.16a** — investigate = create a Clue token for the acting player.
- **CR 400.7 / CR 608.2c** — the `… this way` count reads the merged tracked set of objects the chain's exiles moved.

## Verification

- **Recipient regression** (`declaration_in_stone_gives_clue_to_exiled_creatures_controller_not_caster`): opponent's creature exiled → the Clue is controlled by the opponent, **0** to the caster (inverted on `main`).
- **Count regression** (`declaration_in_stone_investigates_once_per_nontoken_creature_exiled`): opponent controls **two** same-name nontoken creatures; casting exiles both and asserts **2** Clues to the opponent, **0** to the caster. Fails on both `main` and a recipient-only fix (count = 1).
- **Parser building-block test**: pins `player_scope = ParentObjectTargetController` **and** `repeat_for = FilteredTrackedSetSize{Exiled}` for Declaration; `player_scope` set but `repeat_for: None` for Fateful Absence; both `None` for a bare `investigate` negative.
- Parser lib suite: **7984 passed, 0 failed**. `cargo fmt --check` clean; parser-combinator gate passes.

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -6556,11 +6556,18 @@ fn resolve_chain_body(
             controller,
         );
         let matching_players: Vec<PlayerId> = match scope {
-            PlayerFilter::AllExcept { .. } => {
-                // CR 608.2c + CR 109.4 + CR 608.2h: the `AllExcept` anchor may be an
-                // ability-target reference (ParentObjectTargetController), which the
-                // generic `matches_player_scope` predicate cannot resolve (it carries
-                // no `ResolvedAbility`). Route through the ability-aware
+            PlayerFilter::AllExcept { .. }
+            | PlayerFilter::ParentObjectTargetController
+            | PlayerFilter::ParentObjectTargetOwner => {
+                // CR 608.2c + CR 109.4 + CR 608.2h: these anchors are all
+                // ability-target references — the `AllExcept` exclude anchor and the
+                // direct `ParentObjectTargetController` / `ParentObjectTargetOwner`
+                // scopes (e.g. Declaration in Stone's "That player investigates",
+                // where "that player" = the controller of the exiled target). The
+                // generic `matches_player_scope` predicate cannot resolve them (it
+                // carries no `ResolvedAbility`, so `parent_target_controller`'s
+                // last-known-information lookup over `ability.targets` is
+                // unavailable). Route through the ability-aware
                 // `speed_effects::players_for_filter`, then re-impose APNAP order by
                 // intersecting against the apnap sequence.
                 let set =
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -17657,6 +17657,21 @@ fn lower_subject_predicate_ast(
             if let Some(wrapped) = wrap_target_subject_damage(clause.clone(), &subject) {
                 return wrapped;
             }
+            // CR 608.2c + CR 109.4 + CR 701.16a: `Effect::Investigate` is a
+            // fieldless unit variant with no player slot for `inject_subject_target`
+            // to stamp the subject onto, so an explicit non-caster subject ("That
+            // player investigates" — Declaration in Stone, where "that player" is
+            // the controller of the exiled target) would be silently dropped and
+            // the Clue handed to the caster. Record the subject as a pending
+            // `player_scope` (consumed by the effect-chain loop) so resolution fans
+            // the Investigate out to the anchored player instead. Only lifts an
+            // explicit parent-target player anaphor; a bare "investigate" leaves
+            // `affected == SelfRef`/`Controller` and is untouched (caster default).
+            if matches!(clause.effect, Effect::Investigate) {
+                if let Some(scope) = player_scope_from_parent_target_subject(&subject.affected) {
+                    ctx.pending_player_scope = Some(scope);
+                }
+            }
             inject_subject_target(&mut clause.effect, &subject);
             sync_subject_into_nested_shuffle_sub(&mut clause, &subject);
             // CR 109.4 + CR 608.2c (issue #534): When the subject phrase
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
**Request changes — recipient binding is repaired, but Declaration in Stone still executes an incomplete “for each … exiled this way” instruction and cannot be approved as rules-correct card support.**

## 🔴 Blocker

- Authoritative Oracle text: “That player investigates **for each nontoken creature exiled this way**.” The integration regression at `crates/engine/tests/integration/issue_5254_declaration_in_stone_clue_owner.rs:40-82` exiles exactly one creature and asserts one Clue; the PR’s scope note explicitly says the multi-exile count still resolves to one. With two or more same-name nontoken creatures, `Effect::Investigate` therefore creates too few Clues. That is part of the spell’s ordered instruction under CR 608.2c; each investigate is “Create a Clue token” under CR 701.16a.

  Complete the count/repeat design at the shared investigate/tracked-exile seam (with a two-or-more-creature runtime regression), or keep Declaration in Stone unsupported until that implementation cycle is complete. Fateful Absence may be split into a separate, complete single-target fix.

## 🟡 Non-blocking

- `ParseContext.pending_player_scope` and the ability-aware player-scope driver are the right recipient-binding seam for the fieldless `Effect::Investigate`; retain that work when completing the count semantics.

## ✅ Clean

- The current regression correctly proves the recipient inversion for the one-exile case, including the negative assertion that the caster receives no Clue.
- `game/effects/mod.rs:6450-6467` correctly routes parent-target controller/owner scopes through the resolved-ability-aware player filter rather than the generic predicate.

**Recommendation: request changes — finish the shared per-exile-count behavior or split off the complete Fateful Absence subset; do not merge partial Declaration in Stone support.**

jaytbarimbao-collab:
Thanks — blocker addressed. Declaration in Stone now executes the complete "for each nontoken creature exiled this way" instruction, so it's no longer partial support.

**Count completed (commit `e680007e3`).** The recipient-binding seam you approved is retained; I added the count at the same fieldless-Investigate seam:

- `Effect::Investigate` has no count slot, and the `for each` **suffix** handler is CopySpell-only, so the count was dropped. I lift it to the Investigate sub-ability's `repeat_for` via a transient `ctx.pending_repeat_for` — mirroring `ctx.pending_player_scope` — parsing the `for each <filter> <cause> this way` tail with the shared `parse_for_each_clause` and accepting **only** `FilteredTrackedSetSize { caused_by: Some(_) }`. Scoped to `Effect::Investigate`, so count-bearing `for each … this way` effects (gain life / draw / counters) are untouched.
- Runtime already composes with no engine change: the `ChangeZone → ChangeZoneAll` chain publishes one **merged** exile tracked set, and `repeat_for` drives **outside** `player_scope` (the Torment of Hailfire shape) — so N nontoken creatures exiled ⇒ N Clues, all to the exiled creatures' controller.

**Two-or-more-creature runtime regression added** (`declaration_in_stone_investigates_once_per_nontoken_creature_exiled`): the opponent controls two same-name nontoken creatures; casting Declaration in Stone exiles both and asserts **2** Clues to the opponent, **0** to the caster. This fails on both `main` and the recipient-only fix (count = 1). The parser building-block test now also pins `repeat_for = FilteredTrackedSetSize{Exiled}` for Declaration and asserts Fateful Absence (no `for each` suffix) stays `repeat_for: None` (single Clue).

Verification: parser lib suite 7984 pass / 0 fail; both integration tests green; `cargo fmt --check` clean; parser-combinator gate passes.

Non-blocking note acknowledged — kept `ParseContext.pending_player_scope` + the ability-aware player-scope driver as the recipient seam, and added a symmetric `pending_repeat_for` for the count so both lifts live in the same place.

matthewevans (approved):
**Approved — the current rebased head resolves Declaration in Stone’s recipient and per-exile-count behavior at the shared fieldless-Investigate seam.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- The current parse artifact reports no card-parse changes; current-head Card data CI is green, so that unchanged artifact is consistent with the runtime/parser plumbing-only diff.

## ✅ Clean

- `crates/engine/src/parser/oracle_effect/mod.rs:17650-17684` carries the parent-target player scope and linked-exile repeat into the existing `AbilityDefinition` axes rather than adding a one-off Investigate variant.
- `crates/engine/tests/integration/issue_5254_declaration_in_stone_clue_owner.rs:86-136` discriminates two linked exiles from the former one-Clue behavior and verifies the caster receives none.
- GitHub’s current-head required checks are terminal green; no unresolved review threads remain.

**Recommendation: approve and enqueue via merge queue.**

PR #5717, opened 2026-07-13
fix(engine): bind "can't be regenerated" through a delayed-trigger-wrapped Destroy

rykerwilliams:
## Summary

Merieke Ri Berit ({W}{U}{B}, Legendary Creature, Ice Age): "Merieke Ri Berit doesn't untap during your untap step.\n{T}: Gain control of target creature for as long as you control Merieke Ri Berit. When Merieke Ri Berit leaves the battlefield or becomes untapped, destroy that creature. It can't be regenerated."

This card was originally suspected to be misparse-backlog root-cause category #6 ("disjunctive or-list collapsed to first branch") — but investigation found that framing is stale for the card's current state: the "leaves the battlefield or becomes untapped" disjunction is **already correctly parsed and resolved at runtime** (both branches of a `WhenNextEvent{trigger, or_trigger}` delayed trigger are checked against a shared matcher — confirmed via an existing runtime test that exercises the same generic mechanism). No fix was needed there, and it's untouched by this PR.

The real bug: the trailing "It can't be regenerated" sentence failed to bind to the `Destroy` effect nested inside the delayed trigger. Instead of setting `cant_regenerate: true` on that Destroy, the parser fell through to a standalone-clause handler and spuriously built an unrelated static "can't be regenerated until end of turn" grant with no basis in the Oracle text — while the actual destroy kept `cant_regenerate: false`, meaning a stolen creature protected by a regeneration shield would incorrectly survive.

Root cause was a 3-part gap: the antecedent-recognizer, its cached registry, and the apply-side mutator all only looked for a **top-level** `Destroy`/`DestroyAll`, none of them descending into a `CreateDelayedTrigger` wrapper. Fixed by extending the existing "unwrap a `CreateDelayedTrigger` wrapper" idiom (already used for `CopySpell` riders elsewhere in the parser) to the `DestroyLike` antecedent class — a new recursive predicate/mutator pair, widened recognition and registry-population call sites, and a corrected apply-arm mutation.

Scope confirmed via a card-data query: exactly 3 cards share the "delayed trigger + can't-be-regenerated" shape today. Gravebind and Whippoorwill's own delayed triggers wrap unrelated effects (an upkeep draw, a death-exile) — their standalone "can't be regenerated" grant is correct as-is and is confirmed unaffected by this fix.

## Test plan
- [x] Parser test: Merieke's real Oracle text produces `cant_regenerate: true` on the nested Destroy, with no spurious static-ability grant anywhere in the chain.
- [x] Parser tests (negative siblings): Gravebind and Whippoorwill still produce their correct standalone grant, unregressed; Terror and Nekrataal (top-level, non-nested Destroy) unaffected.
- [x] Integration test: a stolen creature under a live regeneration shield is destroyed (not saved) when Merieke's delayed trigger fires via the "becomes untapped" branch.
- [x] Integration test: the same holds via the "leaves the battlefield" branch (proving the untouched OR-disjunction mechanism still works correctly end-to-end with the fix applied).
- [x] Integration control test: an independent plain `Destroy{cant_regenerate: false}` still lets the regeneration shield save its victim — proving the shield mechanism itself is genuinely live, isolating the fixed card's result to the `cant_regenerate` binding specifically.
- [x] 3-stage live-revert-and-rerun performed and independently reproduced by a reviewer: reverting the recognition-arm widening alone reproduces the original bug (no panic); additionally reverting the registry widening reproduces the same bug (no panic, `OnMiss::Ignore` no-ops silently); reverting only the apply-arm's mutator (restoring a direct match) instead **panics** — confirming all three edits are independently load-bearing and correctly wired together.
- [x] `cargo clippy -p engine --tests -- -D warnings` clean
- [x] `cargo fmt --all`
- [x] Parser combinator gate clean (no new textual dispatch — only a Rust match-guard's effect-shape pattern was widened)
- [x] All CR citations (608.2c, 701.19a, 701.19b, 701.19c) verified against the Comprehensive Rules text


matthewevans (approved):
**Approve — the delayed-trigger destroy now carries Merieke Ri Berit’s regeneration denial through the production activation path.**

## 🔴 Blocker

- None. `crates/engine/src/parser/oracle_effect/sequence.rs:3176-3202` extends the existing delayed-trigger wrapper traversal to the nearest `Destroy`/`DestroyAll`, and `crates/engine/tests/integration/merieke_ri_berit_cant_regenerate.rs:94-161` proves both disjunctive trigger branches bypass a live regeneration shield.

## 🟡 Non-blocking

- Gemini is quota-limited; the full current-head review, terminal-green required checks, and exact parse artifact supply the independent evidence.

## ✅ Clean

- The current parse artifact is exactly one intended change: remove Merieke’s spurious standalone `CantBeRegenerated` grant.
- Parser tests retain top-level destroy behavior and non-destroy delayed-trigger siblings; the runtime control confirms an ordinary `cant_regenerate: false` destroy still consumes the shield.
- The ported head has no workflow, instruction, generated-file, or scope-contamination changes.

Issue #3260, opened 2026-06-14
Phantasmal Image + Persist — I cast Phantasmal Image and copied Kitchen Finks.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1514652248850825236/1514652248850825236

**Thread:** Phantasmal Image + Persist
**Cards:** kitchen finks, phantasmal image
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
I cast Phantasmal Image and copied Kitchen Finks.

### kitchen finks
When this creature enters, you gain 2 life.
Persist (When this creature dies, if it had no -1/-1 counters on it, return it to the battlefield under its owner's control with a -1/-1 counter on it.)
### phantasmal image
You may have this creature enter as a copy of any creature on the battlefield, except it's an Illusion in addition to its other types and it has "When this creature becomes the target of a spell or ability, sacrifice it."

---
<sub>report_id: `discord:1514652248850825236:1514652248850825236:0` · discord: `1514652248850825236/1514652248850825236`</sub>

Issue #2350, opened 2026-06-05
God-Pharaoh's Gift: second trigger in same turn creates copy of wrong creature (uses first trigger's exiled card)


matthewevans:
## Reported behavior

With two God-Pharaoh's Gift triggers resolving in the same beginning-of-combat step: the first trigger correctly exiled Icetill Explorer and created a copy. The second trigger, targeting Seasoned Pyromancer, created another Icetill Explorer instead of a Seasoned Pyromancer copy.

## Oracle text (verified from card-data.json)

```
At the beginning of combat on your turn, you may exile a creature card from your graveyard. If you do, create a token that's a copy of that card, except it's a 4/4 black Zombie. It gains haste until end of turn.
```

## Expected vs actual

- **Expected**: Each God-Pharaoh's Gift trigger independently exiles a chosen graveyard creature and creates a copy of that specific creature.
- **Actual**: The second trigger re-uses the `TrackedSet` (id:0) from the first trigger's exile step, creating a copy of the first trigger's exiled creature instead of the second selection.

## Notes

card-data.json shows `CopyTokenOf { target: TrackedSet { id: 0 } }` — this id is likely shared across trigger instances rather than scoped per-trigger, causing the second trigger to copy from a stale tracked set.

## Source

https://discord.com/channels/1485498006781427802/1512198303457083563/1512198303457083563

---
<sub>report_id: `discord:1512198303457083563:1512198303457083563:0` · via bug-triage</sub>

Issue #1348, opened 2026-05-29
Imposter Syndrome — [[Imposter Syndrome]] is making token copies of ITSELF when creatures you control inflict combat da…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1509429866859204739/1509429866859204739

**Thread:** Imposter Syndrome
**Cards:** _none detected_
**Parser status:** no_card
**Extraction confidence:** 0.60

## Summary
[[Imposter Syndrome]] is making token copies of ITSELF when creatures you control inflict combat damage to players, not the creatures in question.

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1509429866859204739:1509429866859204739:0` · discord: `1509429866859204739/1509429866859204739`</sub>

matthewevans:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `Whenever a nontoken creature you control deals combat damage to a player, create a token that's a copy of it, except it isn't legendary.`  (supported: true)

**Reasoning**: The CopyTokenOf child lists 'copies: parent target', but in the DamageDone trigger context 'parent target' resolves to the player, not the creature source. 'it' should refer to the creature that dealt damage — wrong reference causes copying the player/itself instead of the attacker.

---
_Posted by bug-coverage-classifier skill._

PR #5724, opened 2026-07-13
fix(engine): fold copy-token exceptions into BecomeCopy instead of layering them (Vizier of Many Faces #5278)

### LINKED ISSUE #5278: Vizier of Many Faces doesn't prompt to copy, dies on entry — This applies to the embalm cast; didn't test yet on normal…
Reported in Discord: https://discord.com/channels/1485498006781427802/1523399920281719015/1523399920281719015

**Thread:** Vizier of Many Faces doesn't prompt to copy, dies on entry
**Discord thread id:** `1523399920281719015`
**Discord message id:** `1523399920281719015`
**Cards:** vizier of many faces

## Summary
This applies to the embalm cast; didn't test yet on normal cast.

## Reported text

This applies to the embalm cast; didn't test yet on normal cast. But with embalm [[Vizier of Many Faces]] didn't offer to enter as a copy of anything, remained a 0/0 and died immediately.

## Attachments

- [game-state-turn-29-2026-07-05T18-45-53-013Z.zip](https://cdn.discordapp.com/attachments/1523399920281719015/1523399920848207972/game-state-turn-29-2026-07-05T18-45-53-013Z.zip?ex=6a4df27c&is=6a4ca0fc&hm=3ab7a65fb540a6197b98e990a76dafbbf825121314bdc76133e9f82c0e5d5cfb&)

### vizier of many faces
You may have this creature enter as a copy of any creature on the battlefield, except if this creature was embalmed, the token has no mana cost, it's white, and it's a Zombie in addition to its other types.
Embalm {3}{U}{U}

---
<sub>report_id: `discord:1523399920281719015:1523399920281719015:0` · discord: `1523399920281719015/1523399920281719015`</sub>

shin-core:
Fixes #5278 (claimed there with the full re-diagnosis). Verified Vizier's Oracle text on Scryfall and re-verified the issue on current main before building.

## The bug moved since it was filed

The reported symptom (embalm token never offered the copy choice, dies at 0/0) was fixed by #5475's liminal-token-entry replacement pass — the prompt IS now offered. But the bug moved one step deeper and became a **panic**: accepting the copy crashes the engine with `unreachable!` at `types/layers.rs:148` — *"RemoveManaCost is consumed at copy resolution; never layered"*. In production WASM that's a wedge.

## Root cause

`handle_copy_target_choice`'s liminal accept branch re-applied the pending `CopyTokenSpec` copy exceptions (Embalm's `SetColor White` / `RemoveManaCost` / `AddSubtype Zombie` + `extra_keywords`) as a `Duration::Permanent` **layered** transient continuous effect after `BecomeCopy` resolved:
- `RemoveManaCost` is stamp-only — it can never be layered → panic on the next `flush_layers`.
- Per **CR 707.9b**, a copy effect's "except" clauses are part of the copy's **copiable values**, not a separate effect layered on top.

## The fix

Fold the exceptions (and `extra_keywords`, as `AddKeyword`) into the `BecomeCopy` ability's `additional_modifications` **before** it resolves, so `become_copy::resolve` — the single composition authority — consumes the stamp-only modifications into the copiable values and layers the remaining exceptions atop the `CopyValues` clone as CR 613.1 continuous copy exceptions. The layered re-application block is deleted (grep-verified it was the only such site).

Notable judgement call: the first attempt stamped the exceptions onto the committed token's base values — empirically **wrong**, because Vizier's enter-as-copy is a *continuous* Clone effect (`CopyValues` layered), so the next layer flush wiped the stamp. Routing through `become_copy::resolve` is the deeper fix and keeps one authority for CR 707.9b composition.

## Class coverage

- Every **Embalm/Eternalize** clone hits this exact path (both keyword override sets carry `RemoveManaCost` → same panic; Cursecloth Wrappings can grant embalm to any Clone-style card).
- The **God-Pharaoh's Gift / Coiling Rebirth** class (token copies with non-panicking exceptions through the same liminal accept) now composes per CR 707.9b instead of via a spurious permanent layered effect.
- Exception-free token copies (populate, Rite of Replication, Saheeli-class) never entered the block — unchanged, regressions green.

## Tests

- `embalm_copy_of_bear_survives_and_carries_stamped_exceptions` — full `apply()` pipeline with verbatim Oracle text: activate Embalm → accept the replacement → choose Bear → **no panic**, token survives as a 3/3 copy of Bear that is white, a Zombie in addition to its other types, with no mana cost. A revert-probe reproduced the exact `layers.rs:148` panic and flips this test.
- `embalm_copy_declined_enters_as_zero_zero_and_dies` — decline → 0/0 dies to SBA (CR 704.5f), with a positive reach-guard that the replacement choice actually surfaced (no vacuous negative).
- Existing copy-class regressions stay green: God-Pharaoh's Gift #2350, The Fourteenth Doctor, Phantasmal Image persist #3260, Impostor Syndrome #1348, `token_copy` (51) + `become_copy` (95) modules.

## Verification

Full engine lib: **16457 passed, 0 failed**. Full integration binary: **2920 passed, 0 failed**. `cargo fmt` clean. All CR annotations (707.9b, 707.2, 614.12a, 613.1, 704.5f, 111.7) grep-verified against `docs/MagicCompRules.txt`.

Separate defect found while tracing (not fixed here, noted on the issue): `parse_clone_suffix` keeps `AddSubtype Zombie` unconditionally and drops the "if this creature was embalmed" condition, so a normal-cast Vizier copy wrongly gains Zombie — happy to take that as a follow-up.

### DIFF
--- a/crates/engine/src/game/engine_replacement.rs
+++ b/crates/engine/src/game/engine_replacement.rs
@@ -1121,7 +1121,7 @@ pub(super) fn handle_copy_target_choice(
                 "Mismatched liminal entry resume".to_string(),
             ));
         }
-        let ability = copy_effect_for_source(state, source_id)
+        let mut ability = copy_effect_for_source(state, source_id)
             .map(|effect_def| {
                 build_resolved_from_def_with_targets(
                     effect_def,
@@ -1161,15 +1161,44 @@ pub(super) fn handle_copy_target_choice(
                 })
             })
         });
+        // CR 707.9b: a copy effect's "except" clauses (Embalm/Eternalize's
+        // "the token has no mana cost, it's white, and it's a Zombie in
+        // addition to its other types", etc.) are part of the copy's COPIABLE
+        // values, so they must be folded into the `BecomeCopy` that copies the
+        // chosen source rather than re-applied as a separate layered effect on
+        // top of it. Fold the pending `CopyTokenSpec` exceptions (from the
+        // synthesized Embalm/Eternalize `CopyTokenOf`, which reach this liminal
+        // accept path only because the copied card carries its own "enter as a
+        // copy" replacement) into the `BecomeCopy`'s `additional_modifications`.
+        // `become_copy::resolve` is the single authority that consumes
+        // `RemoveManaCost` / `SetStartingLoyalty` into the copiable values and
+        // layers the remaining exceptions atop the `CopyValues` clone — the
+        // only rules-correct home for stamp-only modifications that must never
+        // be layered directly.
         let copy_token_exceptions = state.liminal_entries.get(&source_id).and_then(|entry| {
             entry.copy_resume.as_ref().map(|copy| {
                 (
-                    entry.controller,
                     copy.extra_keywords.clone(),
                     copy.additional_modifications.clone(),
                 )
             })
         });
+        if let Some((extra_keywords, additional_modifications)) = copy_token_exceptions {
+            if let Effect::BecomeCopy {
+                additional_modifications: become_copy_mods,
+                ..
+            } = &mut ability.effect
+            {
+                become_copy_mods.extend(additional_modifications);
+                // CR 707.2 + CR 702: "except it has [keyword]" grants ride the
+                // typed `extra_keywords` channel on the token spec; carry them
+                // through as layered `AddKeyword` exceptions so they become
+                // part of the copy's copiable keyword set.
+                become_copy_mods.extend(extra_keywords.into_iter().map(|keyword| {
+                    crate::types::ability::ContinuousModification::AddKeyword { keyword }
+                }));
+            }
+        }
         if !super::effects::token::commit_liminal_token_entry_with_event_emission(
             state,
             resume.event,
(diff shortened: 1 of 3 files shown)


shin-core:
Fixed the Gemini catch in `89aa2609` — the decline test's reach-guard had a vacuous `|| true`; the assertion is now real (`!candidates.is_empty()`) and verified green (the candidate list is genuinely populated when the enter-as-copy choice surfaces). Note the loop structure already guaranteed the `ReplacementChoice` arm was reached — the guard now additionally proves the candidate list is non-empty.

matthewevans (commented):
**Held — external merge gate**

## 🔴 Blocker
- GitHub's current required `Contributor trust` check is `ACTION_REQUIRED` on `ed7b9e3b605a663018572c07c929bf446e7e7cc0`, so the PR remains `UNSTABLE` and cannot stay in the merge queue.

## 🟡 Non-blocking
- The branch is behind current `main`; I did not rebase while the external gate is unresolved.

## ✅ Clean
- The fix uses the existing `BecomeCopy` resolver, which consumes `RemoveManaCost` before the layer pipeline; the new integration coverage drives the real Embalm replacement/choice path and would hit the prior panic if reverted.
- All project CI checks, including card data, are green. The parse-diff reports no card-parse changes. Gemini's sole test assertion finding is resolved.

**Recommendation: Clear the `Contributor trust` requirement, then re-run the current-head queue review (and rebase if `main` has moved further).**

matthewevans (approved):
**Approved — rebased onto current `main`; the current head preserves the correct copy-resolution seam.**

## ✅ Clean

- `crates/engine/src/game/engine_replacement.rs:1178` moves the pending Embalm/Eternalize exceptions into the existing `BecomeCopy` authority. `become_copy::resolve` consumes `RemoveManaCost` into `CopyValues` before the layer system, so the previous `types/layers.rs` panic path is no longer reachable.
- `crates/engine/tests/integration/vizier_of_many_faces_embalm_copy_panic_5278.rs:155` drives the production Embalm → replacement choice → copy target pipeline and asserts the copied Bear survives with the white/Zombie/no-mana-cost exceptions. The declined path retains a non-vacuous replacement-choice reach guard. Gemini's prior `|| true` finding is resolved on this head.
- The parse-diff artifact for the reviewed implementation reports no card-parse changes.

## 🟡 Non-blocking

- `Contributor trust: ACTION_REQUIRED` is a third-party advisory status, not one of `main`'s required GitHub checks. Required Rust and Frontend checks are running again on this rebased head.

**Recommendation: merge-when-ready after the required checks complete.**

PR #5725, opened 2026-07-13
feat(engine): parse "any other player may" and exclude the drawing player from the fan-out

### LINKED ISSUE #5657: Zur's Weirding: mid-chain Unimplemented — 'any other player may pay 2 life' clause silently truncates the replacement chain
## Summary
From Plan 03 P3-1b harvest (pool-verified). A SILENT PARTIAL: the replacement's first stage parses, then an `Unimplemented` node sits MID-CHAIN, truncating the rest.

## Parse
Reveal-instead parses (it's one of the surviving conditional exemplars); 'Then any other player may pay 2 life…' drops to `Unimplemented` mid-chain, losing the pay-life gate, the graveyard route, AND the otherwise-draw completion. The card appears partially supported while its table-politics mechanic does nothing.

## Class note
Mid-chain Unimplemented (vs whole-ability) is its own severity class: the ability LOOKS implemented in coverage summaries that count at ability granularity. Plan 02's per-item audit will catch line-level gaps but a mid-chain node inside one line needs clause-level accounting (Plan 05 substrate).

andriypolanski:
Closes #5657

## Card (pool-verified)

> If a player would draw a card, they reveal it instead. Then any other player
> may pay 2 life. If a player does, put that card into its owner's graveyard.
> Otherwise, that player draws a card.

## The defect

The draw replacement (`ReplacementEvent::Draw`) was recognized, but the second
clause — **"any other player may pay 2 life"** — had no peel in the
opponent-may prefix table (`clause_shell.rs` had `any opponent may` /
`any player may` / `each opponent may`, but **not** `any other player may`). It
therefore fell to the imperative `Effect::Unimplemented { name: "any" }`
fallback *mid-chain*, sitting as a `sub_ability` that severed the following
"if a player does …" and "otherwise …" branches. The whole table-politics
mechanic no-opped while coverage still counted the ability as present.

Mapping it onto an existing scope was not an option: `AnyOpponent`/`AnyPlayer`
are **controller-relative**, but "any *other* player" excludes the **drawing
player**, who need not be the controller (Zur's controller draws too).

### 1. New scope `OpponentMayScope::AnyOtherPlayer` (`types/ability.rs`)

A genuinely new exclusion axis — *affected-player-relative*, not
controller-relative — so it is a new variant, not a parameterization of the two
controller-relative scopes. Unit variant ⇒ serde-stable (old card-data never had
it; new data emits `"AnyOtherPlayer"`).

### 2. Parser peel (`parser/clause_shell.rs`)

Added an `"any other player may "` arm to `try_peel_opponent_may_prefix`,
ordered before the bare `"any player may "` arm. Class-building: it covers every
"any other player may X" clause, mapping to the new scope. The AnyOpponent /
AnyPlayer peels are untouched.

`chunk_actor` (`oracle_effect/mod.rs`) groups `AnyOtherPlayer` with `AnyPlayer`
(no single controller to stamp on a targeted action — each promptee acts on
their own permanents).

### 3. Drawing-player referent stash (`game/replacement.rs`)

`proposed_event_target` now captures `ProposedEvent::Draw { player_id }` as the
drain's `event_target` (previously only `Damage`/`LifeGain` were captured), and
the `Modified`-arm `stash_post_replacement_continuation` call threads it through
(previously `None`). The full reveal-instead substitution zeroes the original
draw and runs its execute via the resident-top post-replacement drain, so the
drawing player is readable via `post_replacement_event_target()` while the
execute chain resolves (CR 615.5 — the resident drain keeps its event context
during dispatch). `None` for every other event kind, so no existing
continuation changes.

### 4. Affected-player-relative fan-out (`game/effects/mod.rs`)

The APNAP "any … may" fan-out now computes a single `excluded` player instead of
a controller-include boolean:

- `AnyOpponent` → exclude the controller (unchanged),
- `AnyPlayer` → exclude no one (unchanged),
- `AnyOtherPlayer` → exclude the drawing player (the drain's stashed event
  target).

This is the only exhaustive `match` on the scope value, so it is the manual
exhaustiveness guard; both other value-matches (`chunk_actor`) were updated too.

## Verification status (IMPORTANT — read before merging)

This environment has **no `cargo` and Tilt is down**, so nothing here was
compiled or run. Changes were made by inspection and pass the editor lint pass
only. Before merge, run through the Tilt loop (`clippy`, `test-engine`,
`card-data`):

- [ ] `cargo fmt --all`
- [ ] `clippy` / `test-engine` green
- [ ] `card-data` regenerates; Zur's Weirding replacement carries
      `optional_for: AnyOtherPlayer` with no `Unimplemented`

## Tests added (parser)

`crates/engine/src/parser/oracle_effect/tests.rs`:

- `any_other_player_may_pay_life_sets_anyotherplayer_scope` — the Zur execute
  body peels to `optional_for: AnyOtherPlayer` (≠ AnyOpponent, ≠ AnyPlayer) and
  the root effect is no longer `Unimplemented` (the reported defect).
- `any_other_player_may_peel_is_distinct_from_any_opponent` — class-level guard
  that the new peel does not alias the opponent peel.

## Known follow-up (not in this PR)

Clause-3/4 **referent resolution** — "that card into its owner's graveyard"
(`ParentTarget` = the revealed card) and "otherwise, **that player** draws a
card" (must resolve to the drawing player, not the controller) — resolves after
the drain is retired, so it needs its own referent threading and **runtime
integration tests** written under the compiler/Tilt loop. This PR fixes the
reported mid-chain-`Unimplemented` collapse and makes the fan-out target the
correct set of players; the branch-consequence referents are a scoped follow-up.


matthewevans (changes requested):
**Request changes — the new scope is the right semantic axis, but this PR leaves Zur’s Weirding’s resolved branches without the required referent binding or runtime proof.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/tests.rs:26872-26925` proves only that the prefix no longer becomes `Unimplemented`; it does not run the replacement pipeline. The PR description explicitly leaves “that card into its owner's graveyard” and “otherwise, that player draws a card” as a follow-up, yet those are the two outcomes that make Zur’s Weirding’s choice meaningful. Complete the affected-player/card referent threading through the resident replacement drain and add a `GameScenario` regression covering both pay-life and decline branches before claiming support.

- `crates/engine/src/game/effects/mod.rs:6863-6873` falls back to excluding nobody when `post_replacement_event_target()` is absent. For `AnyOtherPlayer`, missing affected-event provenance must fail closed or preserve an explicit deferred state; silently prompting the drawing player changes the card’s rules. Add a negative fixture for a non-controller drawing player and make the absence of the required Draw context non-permissive.

## 🟡 Non-blocking

- CI’s Rust lint job is failed and tests/card-data are still pending; no direct local build was run.

## ✅ Clean

- `parser/clause_shell.rs:338-347` adds the missing phrase as an affected-player-relative scope instead of aliasing the controller-relative opponent scope.
- `replacement.rs:6432-6444` captures the replaced Draw’s player at the appropriate event boundary.

**Recommendation: request changes — finish both branch referents, add the discriminating replacement-path tests, and rerun the current-head CI before re-reviewing as `bug`.**

matthewevans (changes requested):
**Request changes — the current head adds the missing enum and Draw-event capture, but it still falls back to a rules-incorrect prompt when provenance is absent and has no runtime proof of either Zur’s Weirding branch.**

## 🔴 Blocker

- `crates/engine/src/game/effects/mod.rs:6873-6877` still maps a missing/non-player `post_replacement_event_target()` to `None`; the subsequent filter excludes nobody. For `AnyOtherPlayer`, that permits the drawing player to accept whenever the Draw provenance is missing. This must fail closed (or preserve an explicit unsupported/deferred state), rather than reinterpret “other” as “any.”
- `crates/engine/src/parser/oracle_effect/tests.rs:26871-26931` remains parser/AST-only coverage. Add registered `GameScenario` coverage through the replacement drain: a non-controller’s draw is revealed, another player accepts and pays 2 life so the revealed card reaches its owner’s graveyard; a decline leaves that same player drawing it. Include an assertion that the drawing player is never offered the choice. These cases must use the actual apply/choice path, not direct helper calls.

## 🟡 Non-blocking

- The current-head CI run is still in progress; the earlier failed lint result belongs to the superseded head and is not used as a blocker here.

## ✅ Clean

- `OpponentMayScope::AnyOtherPlayer` is now committed, is distinct from controller-relative opponent scope, and the replacement boundary records a Draw’s affected player.
- The current parse-diff is limited to the intended Zur’s Weirding support plus the related previously-unimplemented chain signatures.

**Recommendation: request changes — make missing Draw provenance non-permissive and add the discriminating replacement-path tests, then re-review this current head.**

matthewevans (changes requested):
**Request changes — the provenance fallback now fails closed, but this behavioral change still has no replacement-path runtime regression for either result of Zur’s Weirding’s choice.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/tests.rs:26871-26931` only verifies lowering. Add registered `GameScenario` coverage through the real Draw replacement and optional-choice flow: for a non-controller drawing player, another player accepts/pays 2 life and the revealed card goes to its owner’s graveyard; on decline, that player draws the same revealed card. Assert the drawing player is never offered the choice. These tests must reach the actual apply/choice chain, not call parsing or the new fan-out branch directly.

## 🟡 Non-blocking

- The fresh CI workflow for `ca63a021` is in progress; the scan’s failed result is from the superseded `072d971` run.

## ✅ Clean

- `crates/engine/src/game/effects/mod.rs:6881-6890` now fails closed if the resident replaced-Draw target is absent, fixing the prior permissive fallback.
- `AnyOtherPlayer` remains correctly distinct from controller-relative `AnyOpponent`, and the current parse-diff stays limited to the intended support signatures.

**Recommendation: request changes — add the two discriminating runtime branches, then re-review the current head.**

matthewevans (approved):
**Approved — the affected-player binding now follows the replacement continuation through both optional outcomes, and the final generic helper name matches its Draw, life-gain, and damage scope.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- The parser artifact changes eight cards because the shared Draw-replacement recipient rewrite correctly covers the same pronoun class; it is broader than the original single-card description, so no `quality` label is warranted after the maintainer fixups.

## ✅ Clean

- `issue_5657_zurs_weirding.rs` drives the actual replacement and optional-choice pipeline for accept, decline, and excluded-drawer behavior.
- Current-head CI is terminal green, including card-data; the parse-diff artifact is current for `36a81925`.
- The helper stays at the replacement-parser seam and now names the reusable event-recipient behavior accurately.

PR #5726, opened 2026-07-13
fix(parser): recognize "dealt damage to another creature this turn" intervening-if

real-venus:
**Bug:** Wolverine, Best There Is ("At the beginning of each end step, if Wolverine dealt damage to another creature this turn, put a +1/+1 counter on him") parsed with `condition: null` — the intervening-if was **swallowed**, so the counter was placed **unconditionally**.

**Fix:** Extend the dealing-direction source-dealt-damage condition combinator's target axis with `"another creature"` / `"a creature"` (`Typed{Creature, [Another]}`), alongside the existing player/opponent targets. The clause now lifts to `TriggerCondition::QuantityComparison` over `QuantityRef::DamageDealtThisTurn(SelfRef → another creature) ≥ 1`; the runtime damage-record target matcher already handles creature targets. A pure-addition extension of one existing combinator — no new variant, no runtime change. Verified locally: both new parser tests pass, 96 `dealt`-family sibling tests stay green, clippy clean.

CR 120.1 (source of damage), CR 603.4 (intervening-if).

### DIFF
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -324,7 +324,7 @@ fn parse_damage_dealt_this_turn_conditions(input: &str) -> OracleResult<'_, Stat
         parse_subject_was_dealt_excess_damage_this_turn,
         parse_player_was_dealt_damage_threshold_this_turn,
         parse_player_dealt_combat_damage_by_source_this_turn,
-        parse_source_dealt_damage_to_opponent_this_turn,
+        parse_source_dealt_damage_this_turn,
         parse_source_was_dealt_damage_this_turn,
     ))
     .parse(input)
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**Request changes — the parser extension is on the existing damage-history seam, but this behavioral fix needs an end-to-end trigger test and the current head fails Rust lint.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_trigger_tests.rs:1501-1503` fails `clippy::doc_lazy_continuation` in the current CI run: the doc lines following `Wolverine's` are interpreted as an unindented list. Please make the comment a valid paragraph (or use normal `//` comments) and rerun the required lint gate.
- `crates/engine/src/parser/oracle_trigger_tests.rs:1505-1544` proves only the parsed `TriggerDefinition`. This change governs an intervening-if's trigger detection and resolution, so add a `GameScenario`/actual apply-path regression: Wolverine deals damage to another creature → its end-step trigger puts the counter on Wolverine, plus a negative case where it dealt no qualifying damage and no counter is placed. Parser-shape tests alone would still pass if that runtime path stopped evaluating `DamageDealtThisTurn(SelfRef → Another creature)`.

## 🟡 Non-blocking

- `crates/engine/src/parser/oracle_nom/condition.rs:400` still names the now-broader helper `parse_source_dealt_damage_to_opponent_this_turn`; rename it to describe its source-dealt-damage scope.

## ✅ Clean

- The parse-diff artifact for this head shows exactly the intended one-card condition change (Wolverine), and the new `Another` filter is evaluated source-relatively by the existing damage-record matcher.

**Recommendation: request changes — fix CI and add the discriminating runtime regression, then rebase onto current `main` for a fresh review.**

matthewevans (approved):
**Approved — the current head resolves the prior blockers and is ready for the merge queue.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- Gemini's planeswalker suggestion is intentionally not expanded here: the official Scryfall exact-phrase searches return no cards, while the current parse artifact is exactly Wolverine.

## ✅ Clean

- The helper was renamed for its widened source-damage scope.
- The current parse artifact changes exactly Wolverine's intervening-if condition.
- The paired `GameScenario` regressions prove both the qualifying creature-damage path and the non-vacuous player-damage negative path through end-step trigger resolution.
- All required current-head GitHub checks are green.

**Recommendation: approve and enqueue through the merge queue.**

PR #5734, opened 2026-07-13
feat(parser): count "the number of <type> {returned|put into a graveyard} this way"

real-venus:
**Coverage:** Barrel Down Sokenzan ("deals damage equal to **twice the number of Mountains returned this way**") and Volcanic Eruption ("…the number of Mountains **put into a graveyard this way**") parsed the producer sub-effect but dropped the damage sibling as `Unimplemented` — the "equal to X" amount parser (`parse_event_context_quantity`) never reached the tracked-set "this way" helpers for landing-zone-phrased verb tails.

**Fix:** Add `parse_filtered_landing_zone_this_way` — the destination-phrased sibling of `parse_destroyed_or_sacrificed_this_way_filter` — matching "returned this way" / "put into a graveyard this way" on a **controller-agnostic, type-specific** filter → `QuantityRef::FilteredTrackedSetSize{ filter, caused_by: None }`. Wire it into the "the number of" block, and broaden `parse_event_context_quantity` to accept a bare tracked-set `Ref` (the no-multiplier form). The `Bounce`/`Destroy` producers already publish the chain tracked set; "twice" composes via existing `QuantityExpr::Multiply`.

**No new engine variant, no runtime change** (single file). Reveal/tap/per-recipient/aggregate phrasings are intentionally **excluded** (their producers don't publish a countable set) and stay honestly `Unimplemented` — negative tests guard against false coverage. Verified whole-workspace green: `clippy --all-targets` + full engine suite (2934 integration + 16496 lib, 0 failed), 5 parser AST + 3 runtime tests.

CR 608.2c (this-way identity), CR 400.7j (effect finds objects it moved), CR 701.8a (destroy → graveyard).

### DIFF
--- a/crates/engine/src/parser/oracle_quantity.rs
+++ b/crates/engine/src/parser/oracle_quantity.rs
@@ -574,6 +574,13 @@ pub(crate) fn parse_quantity_ref_with_context(
         {
             return Some(qty);
         }
+        // CR 608.2c + CR 400.7j + CR 701.8a: "the number of <type> {returned | put
+        // into a graveyard} this way" — the bounce/destroy landing-zone sibling of
+        // the destroyed/sacrificed path above. Barrel Down Sokenzan (returned) and
+        // Volcanic Eruption (put into a graveyard) count the same tracked set.
+        if let Some(qty) = parse_filtered_landing_zone_this_way(&rest.to_ascii_lowercase()) {
+            return Some(qty);
+        }
         // CR 301.5a + CR 303.4: "the number of <type> attached to <source>" counts
         // objects whose `attached_to` is the source ("him"/"her"/"~" all denote the
         // source — Whiplash's "where X is the number of Equipment attached to him";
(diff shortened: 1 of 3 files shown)


real-venus:
@matthewevans 
Please review this.

matthewevans (changes requested):
**Request changes — the landing-zone quantity behavior and regressions are solid, but the new parser dispatch must be expressed as composed nom combinators.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_quantity.rs:2239-2282` dispatches the suffix grammar by manually iterating an ordered string table. This is parsing dispatch, so it must use one composable nom production (shared optional relative clause plus an `alt()` for the landing-zone phrases), rather than a loop over near-duplicate string alternatives. Keep the existing filter validation and runtime behavior while moving the grammar into nom.

## 🟡 Non-blocking

- The change is correctly limited to controller-agnostic, nontrivial filters; retaining the Builder’s Bane negative is important when refactoring the grammar.

## ✅ Clean

- The current parse artifact is exactly the intended two-card class: Volcanic Eruption and Barrel Down Sokenzan.
- `crates/engine/tests/integration/landing_zone_this_way_quantity.rs:89-197` drives both destroy-to-graveyard and bounce-to-hand tracked-set resolution through the real ability chain, with discriminating count assertions.
- Current-head GitHub checks are terminal green.

**Recommendation: request changes — replace the manual suffix loop with a composed nom parser, preserve the existing class and negative coverage, then re-request review.**

matthewevans (approved):
**Approved — the landing-zone tracked-set parser is now complete, idiomatic, and covered end to end.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The composed nom tail accepts the controller-agnostic specific-filter class while retaining the generic-card and controller-bearing negative guards.
- The current parse artifact exactly matches the two intended cards: Volcanic Eruption and Barrel Down Sokenzan.
- Registered runtime tests drive both destroy-to-graveyard and return-to-hand chains and distinguish the tracked-set count if this parser path is reverted.

PR #5736, opened 2026-07-13
feat(engine): compose optional draw-skip replacements as Optional ×Prevent

### LINKED ISSUE #5655: Obstinate Familiar + Island Sanctuary: 'you may skip that draw' (Optional × Prevent) fails to compose — blocks Plan 03's optional-prevented-draw coverage
## Summary
From Plan 03 P3-1b harvest (pool-verified). Both primitives exist separately — mandatory draw-prevention parses (Living Conundrum, Possessed Portal) and optional replacement mode parses (Abundance) — but the COMPOSITION 'you may skip that draw' drops to `Unimplemented` on both cards.

## Plan 03 consequence (why this outranks a normal parser gap)
The optional-prevented-draw path has ZERO corpus coverage: Plan 03:274 specifies prevented-individual stage semantics whose OPTIONAL variant (player prompted, may decline) cannot be exercised by ANY card today — these 2 cards are blocked solely by this parse gap. The step-2 rewrite will ship that path untested unless this lands first or a synthetic fixture covers it. Island Sanctuary additionally needs the 'If you do, …' rider threaded.

andriypolanski:
Closes #5655 

## Summary

Mandatory draw-prevention (`skip that draw` → `QuantityModification::Prevent`) and
optional replacement mode (`you may instead` → `ReplacementMode::Optional`) both
existed, but **"you may skip that draw"** fell through to
`Effect::Unimplemented` because no peel recognized the optional-skip modal.
This blocked Obstinate Familiar and Island Sanctuary and left Plan 03's
optional-prevented-draw path with zero corpus coverage.

### Parser (`oracle_replacement.rs`)

- New `strip_optional_draw_skip` combinator recognizes
  `"[instead] you may skip that/the draw [instead]"` (class-building — covers
  both card wordings).
- New draw-replacement arm (before mandatory `body_is_draw_skip` and before the
  Abundance `you may instead {effect}` path) lowers to:
  - `ReplacementMode::Optional { decline: None }`
  - `quantity_modification: Prevent`
  - optional `execute` from `parse_when_you_do_reflexive` when an `"if you do, …"`
    rider follows (Island Sanctuary — attached only when the rider parses cleanly)
- `ReplacementCondition::DuringDrawStep` for `"during your draw step"` antecedent
  (Island Sanctuary).

### Types (`types/ability.rs`)

- `ReplacementCondition::DuringDrawStep` — mirrors `DuringUntapStep` for the
  draw step (CR 504.1).

### Runtime (`replacement.rs`)

- **Decline short-circuit:** optional `Prevent` replacements only suppress the
  event on the **accept** branch. Declining leaves the original draw intact
  (`draw_applier` reads `quantity_modification` regardless of branch).

### AI scan (`ability_scan.rs`)

- `DuringDrawStep` → `Axes::NONE`.

## Tests

**Parser** (`oracle_replacement.rs`):

- `optional_draw_skip_lowers_to_optional_prevent_not_unimplemented` — Obstinate
  Familiar
- `island_sanctuary_optional_draw_skip_during_draw_step` — Island Sanctuary
- `mandatory_draw_skip_stays_non_optional` — Living Conundrum regression guard

**Runtime** (`replacement.rs`):

- `optional_draw_skip_decline_leaves_original_draw` — decline restores the
  original `Draw { count: 1 }` event

## Test plan

- [ ] `cargo fmt --all`
- [ ] `tilt logs clippy` / `test-engine` green
- [ ] `card-data` regenerates; Obstinate Familiar + Island Sanctuary replacements
      carry `Optional + Prevent` with no `Unimplemented`

## Known follow-up

Island Sanctuary's `"if you do"` attack-restriction rider (`can't be attacked
except by creatures with flying and/or islandwalk`) is threaded via
`parse_when_you_do_reflexive` when the effect chain parses cleanly; full runtime
verification of that static restriction is a separate follow-up if the rider
effect shape is not yet supported end-to-end.


matthewevans (changes requested):
**Request changes — optional draw-skip needs composable condition parsing and a decline path that never applies `Prevent`.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_replacement.rs:449-456` dispatches `DuringDrawStep` with `scan_contains` and skips `parse_while_antecedent` in the `else` branch. This violates the parser’s nom-only dispatch rule and makes a draw-step gate mutually exclusive with any typed `while` antecedent. Parse both grammar clauses with composable nom combinators and represent their conjunction rather than selecting one condition.
- `crates/engine/src/game/replacement.rs:6454-6463` returns the unmodified Draw only when an optional `Prevent` has no decline ability. With a decline ability, the subsequent draw applier still sees `QuantityModification::Prevent` and suppresses the declined draw. Route every optional-Prevent decline around the applier while preserving the decline continuation, and add a runtime regression for that branch.

## 🟡 Non-blocking

- The new `DuringDrawStep` enum is appropriately registered in `ability_scan.rs`; keep that exhaustive coverage when composing the condition shape.

## ✅ Clean

- `strip_optional_draw_skip` uses nom combinators for the optional-skip phrase itself.
- The parser tests distinguish optional and mandatory draw-skip wording, and the decline test correctly drives the replacement-choice pipeline.

**Recommendation: request changes — compose the timing and antecedent conditions, then make optional-Prevent decline preserve the original draw even when it has a decline effect.**

matthewevans (changes requested):
**Request changes — Island Sanctuary is marked supported while its conditional attack restriction can be silently discarded.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_replacement.rs:443-455` accepts every optional draw-skip replacement after `parse_when_you_do_reflexive(remainder)` returns `None`. That helper returns `None` when its effect chain is unimplemented, so Island Sanctuary's printed rider — “If you do, until your next turn, you can't be attacked except by creatures with flying and/or islandwalk.” — is dropped while the base draw replacement still returns `Some(def)`. The parser test at `:11047-11065` asserts only the draw replacement shape and therefore cannot catch the lost combat restriction. Preserve an unimplemented rider as a strict parse failure, or implement the rider through the existing attack-restriction authority, then add an accept/decline runtime regression.

## 🟡 Non-blocking

- The coverage artifact correctly isolates the intended two-card support delta; the issue is that Island Sanctuary's reported support is incomplete rather than that the corpus baseline is stale.

## ✅ Clean

- The prior blockers are fixed on `848d874`: draw-step and `while` antecedents are composed, and declining optional `Prevent` now bypasses the draw applier. The required CI checks are green for this head.

**Recommendation: request changes — do not report the card as supported until accepting the replacement enforces its conditional combat restriction and declining it leaves both the draw and restriction untouched.**

matthewevans (changes requested):
**Request changes — the new draw-step gate applies during every player's draw step, not just the replacement controller's.**

## 🔴 Blocker

- `crates/engine/src/game/replacement.rs:4538-4540` implements `DuringDrawStep` as only `state.phase == Phase::Draw`. Island Sanctuary says “during **your** draw step”; `apply_draw_player_scope` at `crates/engine/src/parser/oracle_replacement.rs:443-449` scopes *who draws*, but it does not make the source controller the active player. Therefore an otherwise qualifying draw by the controller during an opponent’s draw step is incorrectly offered the replacement. CR 504.1 identifies the active player’s turn-based draw; the condition needs controller-relative active-player scope (preferably by parameterizing the phase-step condition rather than adding another unscoped sibling) and a cross-turn runtime regression.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The `nom` word-boundary scan replaced the previous manual `scan_contains` dispatch, and the new decline short-circuit correctly preserves the original draw rather than applying `Prevent` on the optional decline branch.
- `attach_optional_draw_skip_rider` now fails closed when Island Sanctuary’s unimplemented “if you do” attack restriction cannot be lowered, resolving the prior coverage-honesty blocker.

**Recommendation: parameterize the draw-step gate with controller-relative active-player scope, add a real replacement-path test for a draw during another player’s draw step, then request re-review.**

matthewevans (approved):
**Approved — optional draw-skip now composes timing and quantity gates correctly, with the original draw preserved on decline.**

## 🔴 Blocker
- None.

## 🟡 Non-blocking
- Island Sanctuary's unimplemented attack-restriction rider remains explicitly unsupported rather than being silently claimed as covered.

## ✅ Clean
- The controller-relative draw-step gate rejects the opponent's draw step.
- Optional \`Prevent\` is applied only on acceptance; both decline paths retain the original draw.
- Current-head CI is green and the parse diff is limited to the two intended draw-replacement cards.

PR #5737, opened 2026-07-13
fix(parser): strip outer-sentence comma inside a quoted granted ability's closing quote (#5681)

### LINKED ISSUE #5681: Parser drops "until end of turn" from a quoted granted ability when Oracle puts the comma inside the closing quote (Bronzehide Lion)
## Defect

Bronzehide Lion's dies-trigger grants an activated ability via quote:

> When this creature dies, return it to the battlefield. It's an Aura enchantment with enchant creature you control and "{G}{W}: Enchanted creature gains indestructible until end of turn," and it loses all other abilities.

(Oracle text verified against `data/card-data.json`.)

The trigger itself parses fully (`ReturnAsAura` + `GrantAbility`), but the granted ability's duration is **lost**: both `duration` fields inside the granted definition are `null`, and the phrase survives only as prose — `"description": "gain indestructible until end of turn,"` (note the trailing comma). The card's own first ability parses the identical phrase into a typed `"duration": "UntilEndOfTurn"` correctly.

**Game impact:** activating the granted ability makes the enchanted creature indestructible *permanently* instead of until end of turn.

## Root-cause hypothesis

Oracle's punctuation convention puts the sentence comma **inside** the closing quote (`...until end of turn," and it loses...`). The quoted-ability sub-parse receives `...until end of turn,` and the duration combinator doesn't match with the trailing comma, so the phrase falls through into the description.

## Class measurement (population: all entries in `data/card-data.json`, full pool)

- Cards whose oracle text contains a quoted string with "until end of turn": **120**
  `jq '[to_entries[] | select(.value.oracle_text // "" | test("\"[^\"]*until end of turn[^\"]*\""))] | length'`
- `GrantAbility` definitions whose description contains "until end of turn" with `duration != null` (parsed correctly): **64 cards**
- Same with `duration == null` (the defect): **1 card — Bronzehide Lion**
- Cards whose oracle text contains a duration phrase immediately followed by `,"` (comma inside closing quote, incl. "until your next turn"/"this turn" variants): **1 — Bronzehide Lion** (sole occupant today, but the Oracle convention will produce more)

## Fix shape requested

Fix the class, not the card: normalize trailing sentence punctuation carried inside a closing quote at the quote-boundary layer, before the granted-ability sub-parse runs — so any quoted granted ability followed by a continuing clause parses identically to its unquoted form. Do not special-case the duration combinator to accept a trailing comma.

## Provenance

Surfaced by the per-unit swallow audit (#5668) firing in merge CI: `SwallowedClause / Duration_UntilEndOfTurn` on the trigger line, now pinned in `integration__oracle_parser__snapshot_bronzehide_lion.snap`. When this is fixed, that snapshot's `parseWarnings` entry disappears and the snapshot must be re-accepted in the same commit — that's the designed workflow, not a regression.

jaytbarimbao-collab:
Tier: Frontier
Model: claude-opus-4-8

Fixes #5681

## Summary

When Oracle text puts the sentence comma **inside** the closing quote of a granted ability — `…until end of turn," and it loses all other abilities` — the quoted-body sub-parse received the trailing comma, the duration combinator failed to match `until end of turn,`, and the duration was dropped into prose. Bronzehide Lion's granted `{G}{W}: Enchanted creature gains indestructible until end of turn,` therefore made the enchanted creature indestructible **permanently** instead of until end of turn.

Bronzehide's own printed first ability parses the identical phrase (with a period, `until end of turn.`) correctly, which is the in-card reference for the intended output.

## Implementation method

Per the requested fix shape — normalize at the quote-boundary layer, **not** by teaching the duration combinator to accept a trailing comma:

- Strip a trailing outer-sentence comma at the top of **`classify_quoted_inner`** (`parser/oracle_static/keyword_grant.rs`) — the single seam every quoted-ability isolation site (`return_as_aura::try_parse`, `parse_quoted_ability_modifications`) funnels through, **before** the trigger / keyword / static / `GrantAbility` sub-parse ladder. A quoted ability followed by a continuing clause now parses identically to its unquoted form. This mirrors the sibling keyword-list (`split_keyword_list`) and static-rule legs that already strip trailing punctuation.
- **Only the comma.** An ability's own rules text always terminates in a period, never a bare comma, so a trailing comma inside a closing quote is always the outer sentence's punctuation. A trailing **period** is the ability's own sentence terminator and is preserved (it stays in the grant `description`, and the duration/keyword sub-parsers strip it themselves) — stripping it would corrupt descriptions like `"{T}: Add {G}."`.

No change to the duration combinator (`parse_duration`) or its caller.

## CR references

- **CR 207.2c** — a granted ability's quoted text; the outer sentence's punctuation is not part of it.
- **CR 613.1** — the granted ability's continuous effect (its `until end of turn` duration).

## Verification

- **Discriminating parse assertion** (`bronzehide_lion_quoted_grant_keeps_until_end_of_turn_duration`): after the fix every `until end of turn` in the card is a typed `Duration::UntilEndOfTurn` — two for the printed `{G}{W}` ability and two for the granted copy — so exactly **four** typed-duration fields; on `main` the two grant-side fields are `null` (count = 2).
- **Snapshot** `snapshot_bronzehide_lion` re-accepted in this commit: both grant-side `duration` fields flip `null → "UntilEndOfTurn"`, the trailing comma drops from the grant descriptions, and the `SwallowedClause / Duration_UntilEndOfTurn` `parseWarnings` entry (from the per-unit swallow audit, #5668) is now empty.
- Existing `issue_1332_bronzehide_lion` runtime test still passes.
- Guarded against over-stripping: the period-preservation tests (`compound_subject_granted_ability_description_keeps_original_case`, `conformer_shuriken`) confirm quoted abilities that legitimately end in a period keep it.
- Parser lib suite: **8083 passed, 0 failed**. `cargo fmt --check` clean; parser-combinator gate passes.

## Parse-diff scope (full population)

Bronzehide Lion is the only card whose **duration** phrase sits immediately before `,"` (the reported bug). But the fix is at the shared quote-boundary layer, so it normalizes the trailing outer-sentence comma for **every** quoted granted ability that has one — the coverage-parse-diff bot picks up **5 additional cards**: Sage's Nouliths, White Mage's Staff, Thief's Knife, Black Mage's Rod, Swashbuckler's Whip (FF equipment of the form `has "…," and is a …`).

Those 5 are **semantic-neutral description normalization, not behavior changes**. The parse-diff is symmetric — each signature is both removed and re-added with an identical summary (`Attacks`, `DamageDone`, `SpellCast`, `Tap`), i.e. no signature type is gained or lost; the only delta is the granted ability's `description` string losing its trailing comma (e.g. `"Whenever this creature attacks, untap target attacking creature,"` → `"…untap target attacking creature"`), matching the unquoted form. The full parser lib suite (8083, 0 failed) confirms no semantic regression across the class.


jaytbarimbao-collab:
Superseded by #5842, which landed the identical sentence-comma normalization at the same `classify_quoted_inner` seam (`trim_end_matches(',')`) and already fixes the Bronzehide Lion snapshot on main. Closing this as a duplicate — thanks!

PR #5741, opened 2026-07-13
fix(engine): install "When damage is prevented this way" riders on the shield (CR 615.5)

### LINKED ISSUE #5658: "When damage is prevented this way, …" riders parse as SequentialSibling and are silently dropped — 3-card class (New Way Forward, Phyrexian Vindicator, Outfitted Jouster)
## Summary
**Confirmed by execution** on New Way Forward (Plan 03 P3-1b fixture; probes showed the shield in the prevention tally with prevented=3, found_def=true, has_runtime=FALSE → rider loop `continue`s). The CR 615.5 payoff clause introduced by a separate **"When damage is prevented this way, …"** sentence parses as `sub_link = SequentialSibling` instead of `ContinuationStep`, so `effects/prevent_damage.rs:371-375` never installs it as the shield's `runtime_execute` — **the prevention works, the entire payoff silently does nothing.** Not Unimplemented; claims support.

## Mechanism (code-verified)
`prevent_damage.rs` installs a rider only when `sub_link == SubAbilityLink::ContinuationStep`. The conjunction form ('prevent that damage AND put that many +1/+1 counters on it' — Gatta and Luzzu) produces ContinuationStep and works. The separate-sentence form produces SequentialSibling — whose code comment says it 'resolves on its own by the chain walker', which a prevention rider cannot do (the prevention hasn't happened yet). It is simply lost.

## The class (full-pool sweep, 35,396 faces: 37 cards mention 'prevented this way'; exactly 3 use the separate-sentence rider form)
- **New Way Forward** — 'When damage is prevented this way, New Way Forward deals that much damage to that source's controller and you draw that many cards.' (execution-confirmed dropped; spell-installed shield path)
- **Phyrexian Vindicator** — 'When damage is prevented this way, this creature deals that much damage to any other target.' (same sentence form; STATIC shield — triage should confirm the drop on the Template-arm path)
- **Outfitted Jouster** — 'When damage is prevented this way, sacrifice an Equipment…' (same form; static-while-equipped shield — confirm)
The other 34 cards use the same-sentence form ('…equal to the damage prevented this way') and include working cards (Swans of Bryn Argoll) — NOT in this class.

## Engine relevance
This defect is also why the `Resolved` → `Effect::Draw` continuation edge has ZERO production witnesses (New Way Forward was the pool's only candidate): type-reachable, not production-reachable. A P3-1b bug-pin fixture (`new_way_forward_prevents_damage_but_its_draw_rider_never_fires`) pins current behavior and goes RED when this is fixed — the comments name the two values that flip (0→−3 damage-back, 0→−3 draws).

## Fix shape
Recognize the 'When(ever) damage is prevented this way,' sentence as a CR 615.5 prevention rider (→ ContinuationStep) rather than an independent sibling clause.

nghetienhiep:
A prevention payoff introduced by a separate "When/Whenever/If damage is
prevented this way, ..." sentence parsed as a SequentialSibling and was
silently dropped: prevent_damage.rs only installs a shield's runtime_execute
from a ContinuationStep sub-ability, so the shield prevented damage while its
entire payoff did nothing (New Way Forward, Phyrexian Vindicator, Outfitted
Jouster).

The "this way" back-reference can only bind to the prevention printed
immediately before it, so effect-chain assembly now folds such a clause into
that prevention as a ContinuationStep. Two runtime gaps surfaced by New Way
Forward's execution-confirmed case are fixed alongside:

- "that source's controller" lowered to a dangling ParentTargetController with
  no runtime referent; in a prevention rider it means the prevented event's
  damage source's controller, so it now lowers to PostReplacementSourceController
  (mirrors the static-shield follow-up path).
- fire_combat_prevention_riders never seeded the aggregate rider's drain event
  source, so PostReplacementSourceController could not resolve. It now derives
  the source from the shield's resolved single-object damage_source_filter.

Repurposes the New Way Forward bug-pin into an end-to-end Resolved->Draw
witness and adds building-block parser tests.

Closes #5658


matthewevans (approved):
**Approved — current-head prevention-rider handling is ready for the merge queue.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The current parse artifact contains exactly the expected two-card affected-source controller rebinding class: New Way Forward and Deflecting Palm.
- The previously requested New Way Forward fixture uses `ContinuationStep`, and the current-head parser uses `tag_no_case` without a per-clause lowercase allocation.
- Current GitHub lint, both Rust test shards, card data, frontend, Tauri, trust, and security checks are green.

PR #5742, opened 2026-07-13
fix(engine): parse and resolve Power Leak's dynamic damage prevention

rykerwilliams:
## Summary

Power Leak ({1}{U}, Enchantment — Aura, old-school): "Enchant enchantment\nAt the beginning of the upkeep of enchanted enchantment's controller, that player may pay any amount of mana. This Aura deals 2 damage to that player. Prevent X of that damage, where X is the amount of mana that player paid this way."

This card was flagged under misparse-backlog category #11 ("replacement/instead effect mis-modeled"), but that framing doesn't match the card's Oracle text — there's no "would...instead" clause anywhere in it. Investigation found three distinct, real bugs, plus a separate runtime bug discovered while making the parser fixes actually work end-to-end:

**Parser fixes:**
1. The "amount of mana paid this way" where-X recognizer only matched one exact phrasing ("the total amount..."), not Power Leak's/Errant Minion's possessive-pronoun variant or Liege of the Hollows' "they paid this way" variant — the latter had no parse warning at all, since it silently fell through to an unbindable literal-string placeholder.
2. Power Leak's "deals 2 damage... Prevent X of that damage" was modeled as two separate sequential effects. Since a prevention effect must exist before the damage event it prevents (CR 615.4), and the "prevent the next N" shield this produced is a depleting shield that leaks unconsumed capacity onto later, unrelated damage the same turn, this needed collapsing into one `DealDamage` with a computed `max(2-X, 0)` amount — the same "later text modifies an earlier effect" pattern the engine already uses for `Destroy`'s `cant_regenerate` flag (CR 608.2c).
3. The "at the beginning of the upkeep of enchanted X's controller" trigger-scope recognizer only matched "creature"/"permanent" nouns, so Power Leak's trigger fired on every player's upkeep instead of just the enchanted permanent's controller's. Fixed by delegating to the existing shared Enchant-noun combinator, incidentally fixing 6 sibling cards (Curse Artifact, Cursed Land, Erosion, Feedback, Power Taint, Warp Artifact).

**A separate runtime bug**, discovered because the parser fixes alone didn't make Power Leak resolve correctly: `TargetFilter::TriggeringPlayer` resolves to the wrong player whenever ability resolution pauses mid-chain on a player choice (e.g. Power Leak's own "may pay any amount of mana") and then resumes. Root cause: `stack::resolve_top` unconditionally clears the live trigger-event context once its stack entry leaves the stack, regardless of whether the ability chain actually finished or merely paused into a continuation — so by the time a paused chain resumes, the context needed to resolve "that player" is gone and resolution falls back to the ability's own controller instead. Confirmed via a real A/B test: Warp Artifact (same trigger shape, no payment pause) correctly damages the enchanted controller; Power Leak (same shape, but with a payment pause) incorrectly damaged the Aura's own controller instead.

Fixed by extending an existing type (`ResolvingTriggerContext`, already used by one other pause mechanism) with its one missing field, giving the paused-continuation carrier a single snapshotted field of that type, and wiring the continuation-resume path to push that snapshot as the live context before resuming and restore it afterward — mirroring a save/restore pattern already shipped for an analogous pause type. Two other, narrower pre-existing mechanisms solve the same conceptual problem for two other pause types and were deliberately left untouched (documented in place); consolidating all three is a real, identified follow-up, deferred here to keep this fix's blast radius on the reported bug.

## Test plan
- [x] Parser tests: Power Leak/Errant Minion's `PreventDamage` fold to a computed `DealDamage` amount with no leftover node; Liege of the Hollows' token count binds to the shared X variable instead of a dead placeholder; the 5 existing Join Forces cards' "total amount" binding is unregressed; Warp Artifact/Cursed Land/Feedback's trigger scoping generalizes correctly.
- [x] Runtime tests: a stolen/damaged creature's owner-of-record resolves correctly across the accept-and-pay pause (fails pre-fix with damage landing on the Aura's controller instead); the decline path and the no-pause control path are both covered as siblings.
- [x] A dedicated regression test for a nested-pause carry-over edge case in the continuation-merge path, with a live-revert-and-rerun proof.
- [x] Full `cargo test -p engine --lib`: 16491 passed, 0 failed.
- [x] `cargo clippy -p engine --lib --tests -- -D warnings` clean.
- [x] `cargo fmt --all`.
- [x] Parser combinator gate clean.
- [x] All CR citations verified against the Comprehensive Rules text.


matthewevans (changes requested):
**Blocked — the new behavior has no production-pipeline regression test.**

## 🔴 Blocker

- [`crates/engine/src/game/effects/pay.rs:1861`](https://github.com/phase-rs/phase/blob/f0af1d51720d2630670d74dd8b916c374d5e25ed/crates/engine/src/game/effects/pay.rs#L1861) tests the context restoration by constructing a `ResolvedAbility`, calling `resolve_ability_chain`, and manually clearing the trigger context at lines 159–166. That reaches the new drain code, but it bypasses both the production parser and `stack::resolve_top`—the two boundaries this PR changes and relies on.
- Power Leak’s official Oracle text is: “At the beginning of the upkeep of enchanted enchantment's controller, that player may pay any amount of mana. This Aura deals 2 damage to that player. Prevent X of that damage, where X is the amount of mana that player paid this way.” The current parser tests validate AST shape, and the helper test validates a synthesized chain, but neither proves that the real parsed trigger is scoped to the enchanted controller, survives the stack/pause boundary, and applies `max(2 − X, 0)` to that player.

## ✅ Clean

- The `PendingContinuation` snapshot/restore is at the shared continuation seam, and the parser diff’s Power Leak / Errant Minion / Liege scope is explained by the reviewed Oracle forms.

**Recommendation: add a registered integration test that drives the real parsed Power Leak through an upkeep trigger, accepts a nonzero mana payment, and asserts the enchanted controller—not the Aura controller—takes exactly `2 − X` damage. Include a decline or `X ≥ 2` sibling case if the fixture makes it cheap; then re-request review.**

rykerwilliams:
Added `crates/engine/tests/integration/power_leak_dynamic_prevention.rs`, addressing the specific gap flagged above: it parses Power Leak's real Oracle text, advances a two-player scenario to the enchanted permanent's controller's actual upkeep, and drives the real `WaitingFor::OptionalEffectChoice → PayAmountChoice → SubmitPayAmount` sequence through `stack::resolve_top` — no hand-built ability chain, no manual trigger-context clearing.

Three cases:
- Pay 1 mana → the enchanted controller (not the Aura's controller) loses exactly 1 life (`max(2-1,0)`).
- Decline → the enchanted controller loses the full 2.
- Pay 3 mana → the enchanted controller loses 0 (clamped at the upper edge).

Live-reverted the `drain_pending_continuation` fix locally to confirm this new test actually catches the original bug: with the fix disabled, the first case fails with the enchanted controller taking 0 damage and the Aura's own controller taking it instead — the exact defect this PR closes. Restored and re-verified all three tests + the full `--lib`/`--test integration` suites are green.

The existing `pay.rs` helper tests are left in place as defense-in-depth; this is additive.

Re-requesting review.

matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- Gemini's plural `current_trigger_events` capture concern is addressed at `crates/engine/src/types/game_state.rs`; its CR-format concern is not valid because compound CR citations are the repository convention.

## ✅ Clean

- The registered Power Leak integration test drives the real parsed upkeep trigger through `stack::resolve_top`, the optional-payment pause, and `SubmitPayAmount`, including decline and full-prevention siblings.
- The current parse-diff matches the claimed 9-card scope: seven Aura phase triggers, Power Leak/Errant Minion's computed damage, and Liege of the Hollows' X-token binding.
- The maintainer fix restores `EventContextAmount` precedence—batched match count, die result, then scalar trigger event—and the fresh CI run is green on `c0fba2f`.

PR #5743, opened 2026-07-13
feat(parser): +X/+Y dynamic pump with per-axis quantities (Aspect of Wolf)

nickmopen:
## Summary

A dynamic P/T pump whose two axes bind to **different** quantities — `gets +X/+Y, where X is <A> and Y is <B>` — dropped entirely. **Aspect of Wolf** now grants `+X/+Y` where **X = half the number of Forests you control, rounded down** (power) and **Y = the same, rounded up** (toughness).

## Implementation method

Hand-authored, grammar-only — **no new engine variant**. Both quantities (`QuantityExpr::DivideRounded` over `ObjectCount`) already exist and are runtime-evaluated; the parser just couldn't reach the second axis.

Three seams:
- `grammar.rs` — `parse_variable_pt_pattern` accepts `y` as a dynamic axis (was `x`/digit only).
- `anthem.rs` — `parse_dynamic_pt_in_text` splits the binding clause on ` and y is ` (case-insensitively) and resolves each half via `parse_cda_quantity`, assigning X→power, Y→toughness. Single-quantity clauses (+X/+X, +X/+0, …) are unchanged.
- `lower.rs` — `structurally_bound_where_x_clause` keeps a `<X qty>, and Y is <Y qty>` continuation (a binding continuation, **not** a new instruction) only when both halves independently parse as where-X quantities, so a genuine `…, and <verb>` next-instruction still bounds off.

## CR references

CR 613.4c (Layer 7c additive P/T) + CR 107.1a (rounding). Touched, not invented.

## Files changed

- `parser/oracle_static/grammar.rs`, `parser/oracle_static/anthem.rs` (+ helper `split_x_and_y_where_clause`)
- `parser/oracle_effect/lower.rs` (where-X bounding)
- `parser/oracle_static/mod.rs` (prelude: `RoundingMode`), `tests.rs`

## Verification (local, committed head `f900c7dfdd42690dd03d4bd52b5c0c703b197145`)

- `cargo test -p engine --lib dynamic_pt_pump_binds_x_and_y_axes_to_distinct_quantities` — ok (power = DivideRounded/2/Down, toughness = DivideRounded/2/Up)
- `cargo test -p engine --lib parser::oracle` — ok (**7904 passed, 0 failed**) — full sweep, since the changed functions are core to every dynamic pump and where-X card
- `cargo fmt -p engine -- --check`, `cargo clippy -p engine --lib` (`-D warnings`) — clean
- `scripts/check-parser-combinators.sh <base>`, `scripts/check-engine-authorities.sh <base>` — pass

Transparency: did not run `scripts/ai-gate.sh` (Gate A) — fresh `--release` build, local disk tight; not fabricating a PASS line. CI covers this head.

## Anchored on

- `crates/engine/src/parser/oracle_static/anthem.rs` — the existing single-quantity `where X` resolution in `parse_dynamic_pt_in_text` this generalizes.
- `crates/engine/src/parser/oracle_effect/lower.rs` — the existing comma-bounded `structurally_bound_where_x_clause` (the `and Y is` case is added ahead of it).

## Claimed parse impact

- **Aspect of Wolf** — `+X/+Y` per-axis pump now parses (was dropped). (Phyrexian Ingester shares this `+X/+Y` shape but additionally needs an exiled-card toughness quantity — a separate follow-up.)

### DIFF
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -7302,9 +7302,9 @@ fn structurally_bound_where_x_clause<'a>(clause: TextPair<'a>) -> TextPair<'a> {
     // `parse_dynamic_pt_in_text` then splits it back and assigns each half to its
     // axis. Guarded on both halves parsing so a genuine "…, and <verb>" next
     // instruction still falls through to the comma-bounding below.
-    if let Some((x_part, y_part)) = clause.split_around(", and y is ") {
-        if parse_where_x_quantity_expression(x_part.original).is_some()
-            && parse_where_x_quantity_expression(y_part.original).is_some()
+    if let Ok((_, (x_part, y_part))) = nom_primitives::split_once_on(clause.lower, ", and y is ") {
+        if parse_where_x_quantity_expression(x_part).is_some()
+            && parse_where_x_quantity_expression(y_part).is_some()
         {
             return clause;
         }
--- a/crates/engine/src/parser/oracle_effect/lower.rs
+++ b/crates/engine/src/parser/oracle_effect/lower.rs
@@ -7295,6 +7295,20 @@ pub(crate) fn strip_trailing_where_x<'a>(tp: TextPair<'a>) -> (TextPair<'a>, Opt
 
 fn structurally_bound_where_x_clause<'a>(clause: TextPair<'a>) -> TextPair<'a> {
     let clause = clause.trim_start().trim_end_matches('.').trim_end();
+    // CR 613.4c: a "+X/+Y" pump binds each axis to its own quantity via
+    // "<X quantity>, and Y is <Y quantity>" (Aspect of Wolf). The "and Y is …"
+    // is a continuation of the SAME binding, not a new instruction, so keep the
+    // whole clause when both halves independently parse as where-X quantities —
+    // `parse_dynamic_pt_in_text` then splits it back and assigns each half to its
+    // axis. Guarded on both halves parsing so a genuine "…, and <verb>" next
+    // instruction still falls through to the comma-bounding below.
+    if let Some((x_part, y_part)) = clause.split_around(", and y is ") {
+        if parse_where_x_quantity_expression(x_part.original).is_some()
+            && parse_where_x_quantity_expression(y_part.original).is_some()
+        {
+            return clause;
+        }
+    }
     let mut has_comma = false;
     let mut best_end = None;
 
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -1197,6 +1197,14 @@ pub(crate) fn parse_dynamic_pt_in_text(
     // "gets -X/-Y" (X/Y defined by later conditional sentences, no `{X}` cost)
     // must not emit a bogus `-CostXPaid/-CostXPaid` static.
     let is_distinct_xy = p_mag == PtAxisMag::VarX && t_mag == PtAxisMag::VarY;
+    // `Y` is not a generic cost variable in this grammar. The only supported
+    // Y-bearing form is Aspect of Wolf's ordered `+X/+Y` pair, whose distinct
+    // bindings are carried by the structured where-clause below. Reject every
+    // other placement so `+Y/+X` or `+Y/+Y` cannot silently borrow `CostXPaid`
+    // or an X-only binding.
+    if (matches!(p_mag, PtAxisMag::VarY) || matches!(t_mag, PtAxisMag::VarY)) && !is_distinct_xy {
+        return None;
+    }
 
     // CR 706.2 + CR 706.3b: "where X is the result" binds X to the preceding
     // die roll's result. `parse_cda_quantity` has no "the result" arm; fall
--- a/crates/engine/src/parser/oracle_static/anthem.rs
+++ b/crates/engine/src/parser/oracle_static/anthem.rs
@@ -1153,24 +1153,20 @@ fn strip_trailing_distributive_each(s: &str) -> &str {
     }
 }
 
-/// Split a compound "+X/+Y" pump binding clause `<A> and Y is <B>` into the
+/// Split a compound "+X/+Y" pump binding clause `<A>, and Y is <B>` into the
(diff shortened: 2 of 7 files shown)


matthewevans (changes requested):
**Request changes — the claimed supported-card behavior still lacks a discriminating runtime witness.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_static/tests.rs:28788-28827` verifies only the parser output. It never installs Aspect of Wolf on an enchanted creature, supplies an odd Forest count, or observes the layer result. A regression in static-definition attachment, recipient rebinding, or dynamic quantity resolution would still pass while a supported card gets the wrong P/T. Add an integration test through the game pipeline that shows three Forests yield `+1/+2`, and that fails if the per-axis binding is reverted.

## 🟡 Non-blocking

- The parser test correctly distinguishes the down/up `QuantityExpr`s; keep it as the narrow grammar guard alongside the runtime test.

## ✅ Clean

- The production parser reaches the existing dynamic-P/T authority and uses the existing rounded quantity expression rather than adding new engine surface. The fetched Oracle text matches the comma-separated `X`/`Y` form handled by this diff.

**Recommendation: request changes — add the end-to-end odd-count runtime regression before marking Aspect of Wolf supported.**

nickmopen:
Thanks both — addressed.

**@matthewevans (blocker — runtime witness):** added `tests/integration/aspect_of_wolf_per_axis_xy.rs`. It installs Aspect of Wolf on a 2/2, gives the controller **3 Forests** (odd, so the axes diverge), attaches, runs `evaluate_layers` + `derive_display_state`, and asserts the host resolves to **3/4** — X = half rounded **down** = 1 (power), Y = half rounded **up** = 2 (toughness). Reverting the per-axis binding makes it 3/3 or 4/4 and the test fails. The narrow parser test stays as the grammar guard, per your non-blocking note.

**@gemini (R1):**
- `lower.rs`: replaced `clause.split_around(", and y is ")` with the nom `nom_primitives::split_once_on` combinator (`parse_where_x_quantity_expression` lowercases internally, so the lower halves feed it directly).
- `anthem.rs`: split on `", and y is "` so the joining comma is consumed by the separator, and the sentence period is already stripped upstream by `strip_trailing_where_x` — so both `trim_end_matches(',' / '.')` are gone; only a plain whitespace `trim()` remains.
- Fixed the stale "already lowercased" doc line.

Local: unit + new integration test pass; `parser::oracle` **7905 passed, 0 failed**; fmt / clippy `-D warnings` / parser-combinator + engine-authorities gates clean.

matthewevans (changes requested):
**Request changes — the new per-axis grammar incorrectly treats unbound `-X/-Y` text as a cost-X pump.**

## 🔴 Blocker

- `parse_variable_pt_pattern` now accepts `y` as a dynamic axis unconditionally, while `parse_dynamic_pt_in_text` still defaults every missing binding to `QuantityRef::CostXPaid`. The current parse-diff confirms the resulting regression: **Snowblind** gained a static `Continuous` parse. Its Oracle text is `Enchanted creature gets -X/-Y. If ... X is ... Y is ...`; it has no `{X}` cost and its X/Y values are defined by later conditional sentences. The new static parser therefore emits `-CostXPaid/-CostXPaid`, which resolves to `0` for the Aura rather than its rules-defined reduction. Keep `Y` distinct from `X` and accept the second dynamic axis only when the same `where X is <A>, and Y is <B>` binding was structurally parsed; otherwise preserve the existing unsupported result. Add a regression guard for Snowblind (no synthetic cost-X static), then regenerate/inspect the parse diff.

## 🟡 Non-blocking

- The current parse-diff also adds **Bioplasm**, although the PR claims only Aspect of Wolf. Once the blocker is fixed, either include Bioplasm in the stated scope with an appropriate behavioral witness or narrow the change so the documented parse impact matches the actual card set.

## ✅ Clean

- The new Aspect integration test is a real layer-pipeline witness: three Forests produce `3/4`, so it distinguishes the down/up axes from a shared quantity. The `nom_primitives::split_once_on` replacement addresses Gemini's stale R1 review finding.

**Recommendation: request changes — guard `Y`-axis parsing on the structured paired binding and prove Snowblind remains honest before re-review.**

nickmopen:
Good catch — fixed the regression.

**Blocker (Snowblind synthetic cost-X):** `parse_variable_pt_pattern` now returns a typed `PtAxisMag { Fixed(i32) | VarX | VarY }` per axis instead of a bare `Option<i32>`. `parse_dynamic_pt_in_text` accepts a distinct `+X/+Y` (VarX on power, VarY on toughness) **only when the structured `where X is <A>, and Y is <B>` paired binding is present**; otherwise the pattern is left unsupported — no cost-X synthesis. So:
- **Snowblind** `gets -X/-Y` (X/Y defined by later conditional sentences, no `{X}` cost) now **drops** — added a regression guard `distinct_xy_pump_without_paired_binding_is_unsupported` that fails if it ever re-synthesizes a static.
- Same-letter (`+X/+X`) and mixed (`+X/+1`) axes are unchanged (29 `dynamic_pt` tests green).

**Non-blocking (Bioplasm):** its binding is `"…the exiled creature card's power **and** Y is its toughness"` — **no joining comma**, so it doesn't match the `", and y is "` split and stays unsupported too. The parse impact is now exactly **Aspect of Wolf** as claimed. (Bioplasm additionally needs an exiled-card *toughness* quantity, which doesn't exist yet — a separate `add-engine-variant` follow-up.)

Local: full `parser::oracle` **7906 passed, 0 failed**; the Aspect layer-pipeline integration test + the new Snowblind guard pass; fmt / parser-combinator / engine-authorities gates clean. Please re-check the regenerated parse-diff — it should now be Aspect of Wolf only.

matthewevans (approved):
**Approved — the per-axis parser is now constrained to the supported paired `+X/+Y` form and has a discriminating layer-pipeline regression.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- Aspect of Wolf’s odd three-Forest runtime test distinguishes the rounded-down power axis from the rounded-up toughness axis through attachment and layer evaluation.
- The structured paired binding is the only accepted `Y` form; Snowblind and the unsupported `+Y/+X`/`+Y/+Y` shapes cannot synthesize cost-X semantics.
- The fresh parse-diff contains only Aspect of Wolf, matching the claimed one-card scope.

PR #5744, opened 2026-07-13
fix(parser): stack-scope "spell(s)" in static-ability subjects (Secret Arcade)

### LINKED ISSUE #5740: parser: compound-subject additive type static silently drops the 'spell' stack-scoping (Secret Arcade)
## Summary
`Secret Arcade`'s additive-type static — `Nonland permanents you control and permanent spells you control are enchantments in addition to their other types.` — silently **mis-parses**, not strict-fails. The compound-subject split itself already works (`parse_continuous_subject_filter` → `parse_shared_controller_compound_subject_filter` correctly produces an `Or` of the two `you control` conjuncts), but the second conjunct, `permanent spells you control`, loses its "spell" qualifier entirely and resolves to a bare `Typed{Permanent, controller: You}` filter — identical to `permanents you control`.

Verified Oracle text (Scryfall, `Secret Arcade // Dusty Parlor`, DSC #10):
> Nonland permanents you control and permanent spells you control are enchantments in addition to their other types.

## Root cause
`crates/engine/src/parser/oracle_target.rs` already has a fully-built, tested mechanism for exactly this class of phrase — a bare "spell"/"spells" head noun means the filter targets an object **on the stack** (CR 109.2), not a permanent:
- `target_phrase_mentions_spell_word` / `target_phrase_uses_spell_suffix` detect the word.
- `scope_spell_targets_to_stack` / `stack_spell_filter` rewrite a `Typed` filter into `TargetFilter::StackSpell` (or `And{[StackSpell, Typed]}` when other type/controller constraints remain).
- `scope_target_spell_phrase` is the entry point, wired into `parse_target_with_ctx` at two call sites (`oracle_target.rs:984`, `:1671`) — this is why targeting grammar ("target artifact or enchantment **spell**") already resolves correctly to a stack-scoped filter.

`parse_type_phrase_with_ctx` (`oracle_target.rs:2038`) — the type-phrase grammar used by static-ability **subject** resolution — never calls `scope_target_spell_phrase`. `crates/engine/src/parser/oracle_static/shared.rs`'s `parse_continuous_subject_filter` falls through to `parse_type_phrase(subject)` at its generic fallback (`shared.rs:4305`) for any "you control"-suffixed conjunct not caught by an earlier special case. Since "permanent spells you control" hits that exact fallback, the "spells" head noun is silently swallowed by the bare type grammar instead of being scoped to the stack.

Net effect: the static's `affected` `Or` filter becomes `[NonlandPermanentsYouControl, PermanentsYouControl]`, which collapses to "permanents you control" — i.e. Secret Arcade incorrectly makes **lands** you control enchantments too (never intended), and never actually reaches permanent spells on the stack at all (its real, intended effect).

## Why this isn't new engine plumbing
The runtime side is already fully generic and stack-aware:
- `crates/engine/src/game/layers.rs`'s `apply_continuous_effect_filtered` scans `effect.affected_filter.extract_in_zone().unwrap_or(Zone::Battlefield)` via `zone_cache.ids_for(state, scan_zone)`.
- `crates/engine/src/game/targeting.rs:2172` — `zone_object_ids` already returns `state.stack.iter().map(|e| e.id)` for `Zone::Stack`.
- `crates/engine/src/game/filter.rs:1796` — `filter_inner` (used by `matches_target_filter`, the matcher `apply_continuous_effect_filtered` calls) already dispatches `TargetFilter::StackSpell` / `TargetFilter::StackAbility` to `matches_stack_target_filter`.

So once the parser emits the correct `TargetFilter::StackSpell`-based conjunct, the existing continuous-effect layer application should apply it correctly with no new runtime code — this is a parser-only fix mirroring the compound-subject-dispatcher generalizations already merged in #5219 / #5406 (reuse an existing, narrowly-wired building block across a broader class of callers), except the failure mode here is a **silent semantic loss** rather than a strict-fail/`Unimplemented`.

## Relevant CR
- CR 109.2 — a bare "spell" head noun refers to an object on the stack.
- CR 110.4b — "permanent spell" = an artifact, battle, creature, enchantment, or planeswalker spell.
- CR 611.3a — a continuous effect generated by a static ability applies to whatever its text indicates, not just the battlefield.
- CR 205.1b — additive type grant ("in addition to their other types").

## Proposed fix
Expose `scope_target_spell_phrase` (or an equivalent `pub(crate)` wrapper) from `oracle_target.rs`, and apply it at `parse_continuous_subject_filter`'s generic `parse_type_phrase` fallback in `oracle_static/shared.rs`, mirroring the two existing `parse_target_with_ctx` call sites. No existing GitHub issue covers this card or this gap.

jsdevninja:
## Summary

Fixes #5740. Secret Arcade's additive-type static — `Nonland permanents you control and permanent spells you control are enchantments in addition to their other types.` (Oracle text verified against Scryfall) — silently **mis-parsed**, not strict-failed: the compound-subject split already worked generically, but the `permanent spells you control` conjunct fell through to the bare type-phrase grammar and lost its "spell" qualifier entirely, collapsing to a plain `permanents you control` filter (dropping the Nonland restriction and never reaching the stack).

It's the same structural class as #5219 / #5406 (a generic compound-subject dispatcher whose per-conjunct resolution is too narrow for one conjunct shape) but a different failure mode: **silent semantic loss**, not a strict-fail/`Unimplemented`.

## Root cause

`crates/engine/src/parser/oracle_target.rs` already has a fully-built, tested mechanism for exactly this: a bare "spell"/"spells" head noun means the filter denotes an object **on the stack** (CR 109.2) — `scope_target_spell_phrase` → `scope_spell_targets_to_stack` → `stack_spell_filter`, which rewrites a `Typed` filter into `TargetFilter::StackSpell` (or `And{[StackSpell, Typed]}`). It's wired into `parse_target_with_ctx` (targeting grammar — why `"target artifact or enchantment spell"` already works), but **not** into `parse_type_phrase_with_ctx`, the type-phrase grammar that `parse_continuous_subject_filter`'s generic fallback (`oracle_static/shared.rs`) uses for static-ability subject conjuncts.

## Fix

- Expose `scope_target_spell_phrase` (`pub(crate)`) and apply it at `parse_continuous_subject_filter`'s `parse_type_phrase` fallback, mirroring the two existing `parse_target_with_ctx` call sites exactly.
- Runtime: `layers::apply_continuous_effect_filtered` computed a single scan zone via `extract_in_zone()`, which stops at the first explicit zone marker found *anywhere* in a compound `Or` filter tree — for Secret Arcade's mixed battlefield+stack `Or` (`[nonland permanents you control, permanent spells you control]`), that silently dropped one conjunct's population from the scan entirely regardless of which zone "won." Compute the scan zone **per disjunct** instead — for every non-`Or` affected filter (the overwhelming majority of existing static abilities) this produces exactly the same single zone as before, so existing behavior is unchanged; for a mixed-zone `Or` it now unions both zones' candidate objects.
- `TargetFilter::extract_in_zone()` gained a `StackSpell | StackAbility => Some(Zone::Stack)` arm (mirroring the existing `ExiledBySource => Zone::Exile` precedent in the same function), which both the parser-level and runtime-level fixes rely on.

## Architecture

- Nom/combinator-adjacent: reuses the existing `scope_target_spell_phrase` building block verbatim rather than re-deriving "spell" detection.
- No new `TargetFilter`/`ContinuousModification` variants — `TargetFilter::StackSpell` and `AddType` are already fully wired; this fix only corrects zone inference and reuses existing predicate matching (`filter_inner`'s `StackSpell => matches_stack_target_filter` arm, `zone_object_ids(state, Zone::Stack)` in `targeting.rs`).
- CR-annotated: **CR 109.2** (bare "spell" head noun → stack object), **CR 110.4b** (permanent spell definition), **CR 611.3a** (a continuous effect applies to whatever its text indicates, not just the battlefield).

## Testing

New tests:
- `parse_continuous_subject_filter_scopes_spell_conjunct_to_stack` — the "permanent spells you control" conjunct in isolation resolves to `And{[StackSpell, Typed{Permanent, controller: You}]}`.
- `secret_arcade_compound_additive_type_static_spans_battlefield_and_stack` — the full Secret Arcade line produces `Or` of the battlefield conjunct and the stack-scoped conjunct, both feeding `AddType(Enchantment)`.
- `single_subject_additive_type_static_unaffected_by_spell_scoping` — no-regression guard for a subject with no "spell" word.
- `extract_in_zone_reports_stack_for_stack_spell_and_stack_ability` — unit coverage for the new `extract_in_zone()` arm, bare and composed inside `And`.

**Note on local verification:** this sandbox's local toolchain has no working MSVC linker (Visual Studio install is incomplete, no WSL distro available), so I could not run `cargo check`/`clippy`/`cargo test -p engine` locally. I did verify `cargo fmt --all -- --check` (clean) and `scripts/check-parser-combinators.sh` (exit 0, no new violations) directly, and did an unusually thorough manual trace of every call site involved (confirmed via direct source reads, not memory) given I couldn't compile. Please pay close attention in review/CI — happy to iterate on any CI failures.

## Scope Expansion

The runtime `layers.rs` fix was not part of the original plan (the reference PRs #5219/#5406 were pure parser fixes) — investigation showed the mixed-zone `Or` filter this card produces would otherwise silently drop one conjunct even with the parser fix alone. Flagged explicitly per the repo's "verify the card, not just the rule" principle.


matthewevans (changes requested):
**Request changes — the parser shape is sound, but the shared layer change is not yet proven or generalized enough to merge.**

## 🔴 Blocker

- `crates/engine/src/game/layers.rs:4581-4615` changes the runtime recipient scan, yet the added tests only inspect parser/`TargetFilter` shapes. None creates a Secret Arcade source, puts a controlled nonland permanent and a controlled permanent spell in their respective zones, applies layers, and observes both become enchantments. Add an integration test through the game/layer pipeline that fails when this scan change is reverted.

## 🟡 Non-blocking

- `crates/engine/src/game/layers.rs:4582-4591` handles only a top-level `TargetFilter::Or`. `TargetFilter` can nest an `Or` under an `And` (for example the existing static-subject grammar produces that shape), where `extract_in_zone()` still returns one first zone and drops a sibling zone. Generalize the scan-zone traversal over the filter tree, applying the battlefield default per leaf, and cover a nested shape.

## ✅ Clean

- The change reuses the existing stack-spell filter authority, and the full parse-diff reports only the intended Secret Arcade type change. Gemini's string-parsing concern is not applicable here: this PR reuses the existing spell-phrase scoper rather than adding a new verbatim dispatch.

**Recommendation: request changes — add the discriminating runtime regression and make the multi-zone traversal recursive before reconsidering approval.**

matthewevans (approved):
**Approved — the current-head parser and layer changes are rules-correct and backed by a discriminating two-zone runtime regression.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- `parse_continuous_subject_filter` reuses the established stack-spell scoper, so `permanent spells you control` retains both its controller/type constraints and its stack scope.
- The layer scan now collects zones per `Or` branch, including a nested `Or`, while retaining the prior single-zone result for non-`Or` filters.
- The Secret Arcade layer test drives both a battlefield permanent and a permanent spell on the stack through `evaluate_layers`; reverting the single-zone scan makes one assertion fail.
- Current parse-diff is limited to the expected mixed spell-subject class (8 cards, 4 signatures) and is stamped for this head.

PR #5746, opened 2026-07-13
fix(parser): UNSUPPORTED seam: Grant-and-track suspend to an exiled set + 'does the same' opponent m

ntindle:
## Summary

**Root cause:** UNSUPPORTED seam: Grant-and-track suspend to an exiled set + 'does the same' opponent mirror

## Cards corrected
- The Wedding of River Song

## Fix
Implemented cluster 111 (The Wedding of River Song) — both named seams now parse AND resolve correctly, with a passing runtime card-test. No new engine enum variants (all reuse existing building blocks).

SEAM A — filtered tracked-set suspend grant ("cards exiled this way that don't have suspend gain suspend"):
- Parser (subject.rs `try_parse_exiled_this_way_keyword_grant`): the restrictive clause is no longer a strict-failure. It now captures the excluded keyword with the shared `parse_keyword_from_oracle` combinator (`take_until(" gain")` + `map`, no string dispatch) and rewrites the GenericEffect grant's `target` and each static's `affected` from `ParentTarget` to `TargetFilter::TrackedSetFiltered { WithoutKeywordKind(kind) }`.
- Resolver (effect.rs `register_transient_effect`): added a `Some(TrackedSet | TrackedSetFiltered)` arm that materializes the grant over the SET's members regardless of zone (reaching the exile-zone cards the battlefield scan can't), applying the "that don't have" test once as a one-shot partition (CR 611.2a). Added an `application_is_tracked_set` guard so an inherited player target (from the chained "does the same" clause) can't misbind the grant to the opponent player.

SEAM B — "target opponent does the same":
- Parser (mod.rs): replaced the `Unimplemented` emission with a `TargetOnly { Typed(controller: Opponent) }` clause whose `sub_ability` is a verbatim clone of the antecedent from-hand exile (optional flag preserved, resolution-time timing). sequence.rs: removed the now-dead `does_the_same_unimplemented_name` and updated docs.
- Resolver (mod.rs `optional_prompt_player`): route a does-the-same from-hand optional to the targeted opponent (not the caster), mirroring `controller_for_relative_filter`.

Runtime card-test (casting_tests.rs) proves end to end: controller draws two, exiles their nonland card (2 Time counters = mana value) and gains suspend; the targeted opponent exiles their own nonland card (3 Time counters) and also gains suspend; controller's card is NOT misrouted to the opponent.

## Files changed
- crates/engine/src/game/effects/effect.rs
- crates/engine/src/game/effects/change_zone.rs
- crates/engine/src/game/ability_utils.rs
- crates/engine/src/game/casting_tests.rs

## CR references
- CR 702.62b (Suspend — suspended = exiled + has suspend + a time counter; grant target of Seam A)
- CR 611.2a (continuous effect with no stated duration lasts until end of game; the 'that don't have' test is a one-shot partition at resolution, not self-negating)
- CR 122.1 (counters; enter_with_counters + Seam A)
- CR 608.2c (follow instructions in order / this-way anaphora)
- CR 608.2d (resolution-time choices announced by the acting player — opponent's own hand pick and optional decision)
- CR 601.2c (targets announced as the spell is cast — TargetOnly opponent slot + the deferred-continuation hoist)
- CR 115.1 (targets; TargetOnly opponent slot)
- CR 614.6 (this-way producer cause; tracked-set gate guard)
- CR 701.56a (Time Travel — verified already implemented upstream, unchanged)

## Verification
- `cargo fmt --all` — pass (no changes; nothing to commit)
- `./scripts/check-parser-combinators.sh <upstream/main merge-base 3086067c2>` — pass (Gate A PASS, exit 0)
- `cargo clippy -p engine -p phase-ai -p engine-wasm -- -D warnings` — pass (exit 0, zero warnings; phase-ai + engine-wasm exhaustive matches compile clean)
- `cargo test -p engine -p phase-ai` — pass (engine 16492 passed/0 failed, phase-ai 2950 passed/0 failed, all wedding tests green)
- `cargo run --profile tool --features cli --bin oracle-gen -- data --filter "the wedding of river song"` — pass (fully lowers, 0 Unimplemented, AST matches oracle text)
Cards confirmed re-parsed correctly: the wedding of river song


matthewevans (changes requested):
**Request changes — the tracked-set grant can materialize against an unrelated active set.**

## 🔴 Blocker

- `crates/engine/src/game/effects/effect.rs:515-519` resolves `set_filter`, then ignores the resolved filter’s ID and fetches `resolve_tracked_set_id(state)` instead. That only happens to work for the sentinel/current-chain case used by The Wedding of River Song; any concrete nonzero `TrackedSet` / `TrackedSetFiltered` will iterate the active set and apply the grant to the wrong cards. Extract the ID from `resolved` before reading `state.tracked_object_sets`, and add a regression with a concrete non-active tracked-set ID.

## 🟡 Non-blocking

- `crates/engine/src/game/casting_tests.rs:6897-6951` accepts both optional exiles. Please add the targeted opponent’s decline path as well: this PR changes the deferred continuation and player routing, so the empty opponent set must still reach the final grant without leaking the controller’s choice context.

## ✅ Clean

- The parser reuses the existing `TargetOnly` and tracked-set vocabulary instead of adding parallel engine surface; the exact parse artifact is confined to The Wedding of River Song.
- Gemini’s player-target concern is refuted: `ResolvedAbility::target_player` already finds the first `TargetRef::Player` and safely defaults to the controller, matching the new guard.

**Recommendation: bind materialization to the resolved tracked-set ID, cover both a non-active concrete ID and the opponent-declines continuation, then request re-review.**

matthewevans:
The requested changes on this current head remain unresolved, and there has been no contributor follow-up. Please address them within 7 days; otherwise this PR will be closed automatically as stale. A substantive update will return it to review.

matthewevans:
Closing as stale: the requested changes on this unchanged head were not addressed within the seven-day window stated in the [2026-07-20 warning](https://github.com/phase-rs/phase/pull/5746#issuecomment-5025521646). A substantive update can be proposed in a fresh PR against current main.

PR #5747, opened 2026-07-13
fix(parser): gate Archmage Ascension's optional search-substitute draw replacement (CR 614.1a)

### LINKED ISSUE #5656: Archmage Ascension: optional search-substitute draw replacement drops to Unimplemented
## Summary
From Plan 03 P3-1b harvest (pool-verified; **lower confidence on exact cause** — flagged for triage rather than asserted).

## Parse
The optional search-substitute drops to `Unimplemented`. Likely related to #5654's compound-substitute gap family (optional + multi-action substitute), but the cause was not traced — triage should state what the card parses to and classify.

## Plan 03 relevance
In the 51-card frozen Draw corpus with execute=Unimplemented; implementing it turns the corpus gate red intentionally (assign a reviewed DrawReplacementScope, re-freeze).

nghetienhiep:
"As long as ~ has six or more quest counters on it, if you would draw a
card, you may instead search your library for a card, ..." dropped its
substitute to Unimplemented and lost the counter gate entirely: the
"As long as" clause's own comma steered extract_replacement_effect to
the wrong split, so the effect text still carried the "if you would
draw a card, you may instead" lead-in and never reached the tutor body.

Split the "As long as <state>, if ... would draw ..." gate off ahead of
effect extraction, lift the state into a typed ReplacementCondition
(OnlyIfQuantity), and let the existing optional-modal + effect-chain
machinery parse the substitute. The counter-state lowering reuses the
CountersOn quantity ref resolved against the replacement source, so the
gate covers the whole "as long as <source-counter/quantity> ... would
draw" class, not just this card.

Closes #5656


matthewevans (changes requested):
**Request changes — the gated optional replacement parses correctly, but its runtime eligibility and accept/decline behavior are not yet proven.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_replacement.rs:17546-17594` asserts only the parsed `ReplacementDefinition`. It never installs Archmage Ascension, varies its quest-counter threshold, applies a real draw event, or resolves the optional choice. Add registered `GameScenario` coverage that proves fewer than six counters leaves the draw untouched, six counters offers the replacement, and both accepting and declining preserve the correct draw/search outcome. This must fail if the `OnlyIfQuantity` gate or optional continuation is removed.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The new `AsLongAsDrawGate` fails closed when its condition cannot lower, and reuses `OnlyIfQuantity`/`CountersOn` rather than introducing a card-specific condition.
- The current parse-diff is limited to the claimed Archmage Ascension replacement and search signatures.

**Recommendation: request changes — add the discriminating replacement-pipeline regression, then re-review.**

matthewevans (approved):
**Approved — current-head parser and runtime replacement coverage are complete.**

## 🔴 Blocker
- None.

## 🟡 Non-blocking
- None.

## ✅ Clean
- `crates/engine/src/parser/oracle_replacement.rs:7188-7277` splits the leading gate with existing nom primitives, lowers it into the existing typed `OnlyIfQuantity` condition, and fails closed when the condition cannot be represented.
- `crates/engine/tests/integration/archmage_ascension_gated_draw_replacement.rs:100-280` drives the registered replacement pipeline below the counter threshold and through both accept and decline paths; the buried-card assertion distinguishes a true replacement from a supplementary draw.
- The parse-diff is limited to Archmage Ascension’s claimed draw replacement/search support, and required CI is green on `09b0696`.

PR #5761, opened 2026-07-13
feat(engine): give Conjure a library position for "top N at random"

### LINKED ISSUE #5614: engine: Conjure has no library position, so "into the top N cards at random" collapses to Zone::Library
## Summary

`Effect::Conjure` models its destination at **Zone granularity** (`destination: Zone`), so Alchemy conjures that name a *positional* library slot lose that constraint. Introduced knowingly in #5610, which parses the destination phrase and collapses it to `Zone::Library`.

Affected templating (6+ cards): *"conjure a duplicate of that creature **into the top five cards of your library at random**"* — Goblin Morale Sergeant, Jessie Zane, Fangbringer, Mine Security, Sheoldred's Assimilator, Sliver Weftwinder.

**Gameplay impact:** the card is supposed to land in a random slot among the top N, so it is drawn within roughly N draws. Collapsed to `Zone::Library`, it lands at whatever `zones::create_object` inserts at — a deterministic end of the library — so it is either drawn far sooner or effectively never. That is a real deviation, though a much smaller one than the status quo before #5610, where the whole clause failed to parse and the ability did nothing at all.

## Why it wasn't fixed in #5610

Representing it faithfully needs **two** engine changes, neither of which belongs in a parser PR:

1. `Effect::Conjure` would need to carry a position (it currently has only `destination: Zone`; contrast `Effect::ChangeZone`, which already has `library_position: Option<LibraryPosition>`).
2. `LibraryPosition` has no variant for this shape. Its variants are `Top`, `Bottom`, `NthFromTop { n }`, and `BeneathTop { depth }` — none of which expresses *"a uniformly random slot among the top N."*

So this is a new field **and** a new enum variant, which is an `add-engine-variant` gate decision, not a hunk on a parser PR. #5610 is a strict improvement (6 cards go from non-functional to functional) and its collapse is documented in the code.

## Proposed fix

- Add a `RandomWithinTop { n: QuantityExpr }` variant to `LibraryPosition` (CR has no entry — Alchemy is digital-only — so annotate accordingly rather than fabricating a CR number).
- Give `Effect::Conjure` a `library_position: Option<LibraryPosition>` field, mirroring `Effect::ChangeZone`.
- Route the existing `" into the top N cards of ... at random"` combinator arm in `oracle_effect/mod.rs` to populate it instead of discarding the count.

The parser already consumes the count via `parse_number` and throws it away; the arm is the natural place to thread it through once the type can hold it.

## Evidence

- `Effect::Conjure` definition: `crates/engine/src/types/ability.rs` (`cards`, `destination: Zone`, `tapped` — no position).
- `LibraryPosition` variants: `crates/engine/src/types/ability.rs`.
- `Effect::ChangeZone` already carries `library_position: Option<LibraryPosition>` — the precedent to mirror.
- Card list from `data/mtgjson/AtomicCards.json` and the #5610 parse-diff sticky.

nghetienhiep:
Effect::Conjure modeled its library destination at Zone granularity, so
Alchemy conjures that name a positional library slot ("into the top five
cards of your library at random" — Goblin Morale Sergeant, Jessie Zane,
Fangbringer, Mine Security, Sheoldred's Assimilator, Sliver Weftwinder)
collapsed to Zone::Library and landed at the deterministic bottom, drawn
far sooner or effectively never.

Add a LibraryPosition::RandomWithinTop { n } variant (digital-only, no CR
entry) and an Option<LibraryPosition> field on Effect::Conjure mirroring
Effect::ChangeZone. The parser arm that already consumed the count now
threads it into the position instead of discarding it, and the resolver
reorders the just-conjured card into a uniformly random slot among the
top N (in-library reorder, not a replaceable zone event).

Closes #5614


matthewevans (changes requested):
**Changes requested — this independently reintroduces the positional-conjure gaps already identified for the same card class.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/mod.rs:9180-9205` accepts both `your library` and `each player's library`, while `crates/engine/src/game/effects/conjure.rs:45-107` creates every card with `ability.controller` and `:222-266` reorders only that controller's library. Sandcloud Harbinger's verified Oracle text is: “When Sandcloud Harbinger enters, conjure three cards named Sunscorched Desert into the top ten cards of each player's library at random.” The current model has no destination-player scope, so it necessarily puts all three cards into one library. Carry a typed player scope through `Effect::Conjure` and resolve once per affected player; add a multiplayer runtime regression asserting three cards in every affected library.

- `crates/engine/src/game/effects/conjure.rs:222-266` directly removes and inserts a library card but does not call the existing `layers::mark_layers_full_if_top_of_library_static_live` authority used after library-top changes. A cached `TopOfLibraryMatches` static can therefore retain the former top card when this reordering selects index zero. Reuse that invalidation seam and add a production-path regression with a live top-of-library static.

## 🟡 Non-blocking

- The added many-seed placement test is useful for the singular controller-library path, but it cannot expose either player scope or a cached-layer failure.

## ✅ Clean

- The top-N computation samples from the actual library slots, and the test proves the chosen index varies.
- `LibraryPosition::RandomWithinTop` is a maintainable typed representation of the positional constraint.
- Current GitHub CI is green for `ece8e9a`.

**Recommendation: carry destination player scope and route the reorder through the existing library-top invalidation authority, with multiplayer and live-layer regressions, before re-review.**

matthewevans (changes requested):
**Changes requested — the player scope and layer invalidation are now correct, but multi-card conjures do not preserve the final top-N constraint.**

## 🔴 Blocker

- `crates/engine/src/game/effects/conjure.rs:299-302` samples and inserts each card independently while `crates/engine/src/game/effects/conjure.rs:308-318` mutates the library immediately. For Sandcloud Harbinger, whose verified Oracle text is “When Sandcloud Harbinger enters, conjure three cards named Sunscorched Desert into the top ten cards of each player's library at random,” an earlier card inserted at index 9 can be shifted to index 10 by a later card inserted before it. The final library can therefore contain one of the three outside its top ten. Choose/reserve the full set of final slots (or construct the recipient's final top-N segment atomically) so every conjured copy remains in the named window after the entire instruction resolves. Add a deterministic multi-card regression that forces the old late-insertion displacement and asserts all copies are still within the final top-N range.

## 🟡 Non-blocking

- `crates/engine/src/game/effects/conjure.rs:273-318` honors non-random `LibraryPosition` variants that no Conjure parser currently emits. Keeping that composability is reasonable, but the test suite should stay focused on the parser-reachable random placement path.

## ✅ Clean

- `crates/engine/src/game/effects/conjure.rs:70-85` now correctly fans the effect out to each non-eliminated recipient, and the multiplayer regression checks ownership as well as library membership.
- `crates/engine/src/game/effects/conjure.rs:235-239` uses the existing top-of-library layer invalidation authority, with a production-path regression.
- The current parse-diff reports no unexpected card-parse changes.

**Recommendation: make positional Conjure placement atomic per recipient across the whole card count, prove the adversarial insertion order, then request re-review.**

matthewevans (changes requested):
**Changes requested — the atomic placement is at the right seam, but the regression does not yet prove the displaced-copy failure and the current head fails the zone-authority gate.**

## 🔴 Blocker

- `crates/engine/src/game/effects/conjure.rs:481-556` seeds the production RNG and asserts the final top-three contents, but it does not force or expose the old late-insertion ordering. The replaced per-copy implementation can leave all three copies in the top three for some sampled slot sequences (for example, successive slots 0, 1, 2), so this test's source alone does not establish that reverting the atomic block fails. Add a deterministic production-path witness for the adverse order (or deterministically search fixed seeds and assert the witness) before re-review.

- Current GitHub CI is red: `scripts/check-engine-authorities.sh` reports the stale `reposition_conjured_in_library` census row after this commit renamed the exempt helper to `place_conjured_in_library`. Regenerate the frozen baseline so the required authority gate matches the actual raw-zone sites.

## 🟡 Non-blocking

- The `RandomWithinTop` algorithm assumes the requested window is at least the number of conjured copies; that is true for the parser-reachable Alchemy cards named here, but a future producer with `count > n` will need an explicit semantic policy because all copies cannot fit in that window.

## ✅ Clean

- The new `place_conjured_in_library` seam reserves and constructs the recipient's final window atomically, which addresses the previous multi-copy displacement flaw.
- The current head retains the earlier recipient-scope and top-of-library invalidation fixes; the parse-diff artifact for this head reports no card-parse changes.

**Recommendation: add a deterministic adverse-order regression and refresh the authority baseline, then request re-review.**

matthewevans (changes requested):
**Changes requested — the authority baseline is now correct, but the required adversarial regression is unchanged.**

## 🔴 Blocker

- `crates/engine/src/game/effects/conjure.rs:481-556` is unchanged from the prior reviewed head: it seeds the production RNG and observes one final arrangement, but it does not force or expose the old late-insertion order. The replaced incremental implementation can still keep all copies in the final top three for some sampled sequences, so this test does not demonstrate that reverting the atomic placement fails. Add a deterministic production-path adverse-order witness (or a deterministic fixed-seed search that proves one) before re-review.

## 🟡 Non-blocking

- None.

## ✅ Clean

- `scripts/zone-authority-baseline.txt` now removes the obsolete `reposition_conjured_in_library` census row, and the current GitHub CI run is green.
- The parse-diff artifact remains `No card-parse changes`, matching this resolver-only follow-up.

**Recommendation: retain the atomic implementation and add the missing discriminating regression, then request re-review.**

matthewevans (approved):
**Approved — current head clears the requested multi-copy regression proof.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- `RandomWithinTop` still assumes parser-reachable inputs where the named top-N window can contain the conjured count; any future producer with `count > N` needs an explicit policy.

## ✅ Clean

- `crates/engine/src/game/effects/conjure.rs` constructs each recipient’s final positional-library window atomically at the resolver seam, retaining the recipient scope and shared top-of-library invalidation already reviewed.
- The fixed seed sweep covers an adverse `three into top five` ordering across seeds `0..64`; reverting to the former per-copy placement permits displacement beyond the final window.
- Current required CI is green, and the current parse-diff artifact reports no card-parse changes, matching this resolver-only follow-up.

**Recommendation: approve and enqueue as `enhancement`.**

PR #5765, opened 2026-07-13
Fix targeted loyalty cancel rollback

mike-theDude:
@/tmp/pr-4739-body.md


matthewevans (changes requested):
**Changes requested**

## 🔴 Blocker
- `casting_targets.rs:625-642` now records a targeted loyalty activation only after cost payment returns normally. A positive loyalty cost can pause at `costs.rs:836-846` for a counter-replacement choice; that stores `activation_cost: None`. After the choice, `engine_replacement.rs:883-896` resumes the pending activation and `casting_costs.rs:1508-1529` pushes it without calling `record_loyalty_activation`, so the permanent remains eligible for a second loyalty activation that turn. Preserve a durable “record loyalty activation after payment” marker through the pause/resume path (rather than recording before target selection), and add a regression with competing counter-addition replacements that forces the choice, then verifies the completed activation consumes the loyalty window and a retry is rejected.

## 🟡 Non-blocking
- None.

## ✅ Clean
- The new cancel-path integration test correctly demonstrates that cancelling before target commitment leaves both the loyalty counters and activation history unchanged.

**Recommendation: revise the deferred replacement-payment continuation, add the adversarial regression, then request re-review.**

mike-theDude:
Addressed Matt's blocker in e1b8f4550.\n\nThe follow-up carries a durable pending loyalty-activation marker through replacement-choice cost pauses and consumes it when the activation reaches the stack. Added the requested regression with competing counter-addition replacements to prove the completed targeted +loyalty activation spends the loyalty window and a retry is rejected.\n\nLocal checks: cargo fmt --all; cargo test -p engine --test integration targeted_loyalty_replacement_pause_spends_loyalty_window; cargo test -p engine --lib pending_loyalty_activation_player; cargo test -p engine --test integration the_chain_veil_loyalty_grants. CI is green on e1b8f4550.

matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

None on the current head.

## 🟡 Non-blocking

Gemini review is unavailable due to its daily quota; the final pass used the local diff, current CI, and the targeted integration regression instead.

## ✅ Clean

The maintainer port preserves both behaviors at the shared cost-move seam: the targeted loyalty marker survives a replacement-choice pause, and the pending cast is reattached to the delivered cost move. `targeted_loyalty_replacement_pause_spends_loyalty_window` drives the real choice/resume path and rejects a second activation, so it fails if the marker preservation is removed. Current required CI is green and the current-head parse-diff reports no card-parse changes.

**Recommendation: enqueue as a bug fix.**

PR #5768, opened 2026-07-14
test(parser): pin counter-on-each → untap-those tracked-set anaphora

lourincedaging0-commits:
Adds a building-block parser test for the **"put a +1/+1 counter on each `<filter>`, then untap those `<permanents>`"** composition (Virtue of Loyalty class).

## What it pins

The clause lowers to a `PutCounterAll` primary whose follow-up untap is a **sub-ability targeting the anaphoric tracked set** — `SetTapState(Untap)` on `TargetFilter::TrackedSet`. In other words, "those creatures" resolves to exactly the permanents that just received counters, **not** a fresh re-scan of the "each creature you control" filter.

Asserted shape:
- primary: `Effect::PutCounterAll { counter_type: Plus1Plus1, count: Fixed(1), target: Typed{ Creature, controller: You } }`
- sub-ability: `Effect::SetTapState { target: TrackedSet, scope: Single, state: Untap }`

## Why

This specific `PutCounterAll → SetTapState(Untap)`-on-tracked-set linkage had no dedicated test — the existing `"untap those"` coverage is for unrelated effects (bounce / land untap). Pinning it guards the anaphor against regressing to a filter re-evaluation, which would wrongly untap creatures that entered *after* the counters were placed.

Building-block level (asserts the effect/sub-ability shape, not a single card name), per the repo testing guidance.

## Verification

- `cargo test -p engine --lib put_counter_on_each_then_untap_those_targets_the_tracked_set` → **1 passed**
- Parser-combinator gate → PASS. Test-only; no engine/parser logic changed.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/tests.rs
+++ b/crates/engine/src/parser/oracle_effect/tests.rs
@@ -4841,6 +4841,75 @@ fn radiance_fanout_declines_variable_x_amount() {
     );
 }
 
+/// Building block: "put a +1/+1 counter on each <filter>, then untap those
+/// <permanents>" lowers to a `PutCounterAll` primary with the follow-up untap
+/// as a sub-ability that targets the anaphoric TRACKED SET — i.e. "those
+/// creatures" resolves to exactly the permanents that just received counters,
+/// not a fresh re-scan of the "each creature you control" filter. Pinning the
+/// `PutCounterAll -> SetTapState(Untap)` tracked-set linkage guards the anaphor
+/// against regressing to a filter re-evaluation (which would untap creatures
+/// that entered after the counters were placed). Virtue of Loyalty class.
+#[test]
+fn put_counter_on_each_then_untap_those_targets_the_tracked_set() {
+    let parsed = parse_oracle_text(
+        "At the beginning of your end step, put a +1/+1 counter on each creature you control, then untap those creatures.",
+        "Virtue of Loyalty",
+        &[],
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**REQUEST CHANGES — this standalone parser test duplicates an existing lowerer regression.**

## 🔴 Blocker
- `crates/engine/src/parser/oracle_effect/lower.rs:717-735` already asserts that a `PutCounterAll` head followed by plural untap rewrites the continuation to `TargetFilter::TrackedSet(0)`. The new assertion at `crates/engine/src/parser/oracle_effect/tests.rs:4853-4914` restates that same AST contract without reaching a new production branch or a runtime fixture that can distinguish a stale-filter re-scan.

## 🟡 Non-blocking
- None.

## ✅ Clean
- The test is narrowly scoped, and the current parse-diff correctly reports no card-output changes.

**Recommendation: close this as redundant, or replace it with a registered runtime regression that proves the tracked set excludes a creature entering after counter placement.**

lourincedaging0-commits:
You're right — thanks for the pointer. `put_counter_all_head_plural_untap_rewrites_to_tracked_set` in `lower.rs` already pins the `PutCounterAll`-head → `TrackedSet(0)` contract, so this end-to-end restatement doesn't reach a new production branch. I agree it's redundant, and a runtime fixture that could actually distinguish a stale-filter re-scan isn't meaningfully constructible here (the counter and the untap resolve in one atomic step). Closing.

PR #5770, opened 2026-07-14
test(parser): pin "protection from everything until your next turn" duration

lourincedaging0-commits:
Adds a building-block parser test for the **timed** player-scoped protection grant — "you gain protection from everything **until your next turn**" (The One Ring).

## What it pins

The grant must carry `Duration::UntilNextTurnOf { player: Controller }` so the protection expires at the **start of the controller's next turn** — not at end of turn, and not permanently.

Asserted shape (`parse_effect_chain` → `Effect::GenericEffect`):
- `static_abilities[0].affected == Some(TargetFilter::Controller)`
- `AddKeyword(Protection(Everything))`
- `duration == Some(Duration::UntilNextTurnOf { player: Controller })`

## Why it is additive

The existing `parse_you_gain_protection_from_everything_targets_controller` covers the **untimed** grant — it asserts `affected` + the keyword but **not** any duration. This test adds the timed variant and specifically guards the `until your next turn` duration from collapsing to `UntilEndOfTurn` (which would drop protection a full turn early).

Building-block level (asserts the effect/static/duration shape, not a single card name).

## Verification

- `cargo test -p engine --lib parse_you_gain_protection_from_everything_until_your_next_turn_carries_duration` → **1 passed**
- Parser-combinator gate → PASS. Test-only; no engine/parser logic changed.


matthewevans (changes requested):
**Changes requested — this standalone parser test duplicates an existing end-to-end regression and does not earn additional suite cost.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/tests.rs:26706-26750` asserts the timed player-scoped protection shape already covered by `parse_teferis_protection_end_to_end` at `:26443-26534`: that existing test parses the full official Teferi’s Protection text, asserts `Duration::UntilNextTurnOf { Controller }` at `:26464-26470`, then follows the chained protection clause and asserts both `Protection(Everything)` and `TargetFilter::Controller` at `:26495-26534`. The new assertion therefore adds no distinct failure mode or production path; it remains green whenever the established regression is green. Please withdraw this duplicate, or replace it with a specific untested wrong implementation that the existing Teferi test cannot distinguish.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The refreshed head’s CI is green and the current parse-diff reports no card-parse changes, consistent with a test-only diff.
- The test is correctly colocated in the existing parser suite, so it does not create a new test binary.

**Recommendation: do not enqueue this standalone test unless it is narrowed to a genuinely uncovered regression with a demonstrated distinct failure.**

lourincedaging0-commits:
You're right — `parse_teferis_protection_end_to_end` already parses the full official text and asserts `Duration::UntilNextTurnOf { Controller }` plus `Protection(Everything)` on `TargetFilter::Controller`, so this standalone test adds no distinct failure mode. It stays green whenever that regression is green. Closing as redundant rather than adding suite cost. Thanks for the pointer.

PR #5773, opened 2026-07-14
fix(engine): fire returned creature's ETB after an exile-return (#3673)

### LINKED ISSUE #3673: Fiend Hunter — When fiend hunter leaves the battlefield in this game state, wall of omens' ETB doesn't trigger and I do…
Reported in Discord: https://discord.com/channels/1485498006781427802/1516828792213999798/1516828792213999798

**Thread:** Fiend Hunter
**Cards:** fiend hunter, wall of omens

## Summary
When fiend hunter leaves the battlefield in this game state, wall of omens' ETB doesn't trigger and I don't draw a card.

### fiend hunter
When this creature enters, you may exile another target creature.
When this creature leaves the battlefield, return the exiled card to the battlefield under its owner's control.
### wall of omens
Defender
When this creature enters, draw a card.

---
<sub>report_id: `discord:1516828792213999798:1516828792213999798:0` · discord: `1516828792213999798/1516828792213999798`</sub>

mike-theDude:
## Summary

Fixes #3673. When a permanent returns from exile via an "until this leaves the battlefield" link (Fiend Hunter, Banisher Priest, Oblivion Ring, Journey to Nowhere), the returned permanent's own enters-the-battlefield trigger never fired — Wall of Omens, returned when Fiend Hunter died, did not draw a card.

## Root cause

`check_exile_returns` moves the exiled card back through the zone pipeline, appending the resulting `ZoneChanged` (enter) events to the pipeline's event buffer. But its caller, `run_post_action_pipeline_from`, never scanned those events for triggered abilities — unlike the state-based-action loop directly above it, which does run `process_triggers` on the events an SBA pass produces. So the returned permanent's ETB trigger was never collected and never reached the stack.

## Fix

`crates/engine/src/game/engine_priority.rs`: after `check_exile_returns`, scan the newly-appended events with `process_triggers`, mirroring the SBA loop's inner block. Gated on `WaitingFor::Priority` so a return that paused on an as-enters replacement/aura choice (`BatchMoveResult::NeedsChoice`) is left to its own resume scan. This also correctly lets ETB observers on *other* permanents (Soul Warden class) see the return — a broader win than the single reported card.

CR 603.2 + CR 603.3: a returning permanent's enter is a game event that triggers abilities, which are put on the stack the next time a player would receive priority. CR 704.3: any state-based actions the return implicates are checked on the next pipeline pass once the ETB resolves.

## Anchored on
- crates/engine/src/game/engine_priority.rs:126-127 — SBA loop: slice the just-produced events (`events[events_before..]`) and run `triggers::process_triggers` on them; my exile-return block mirrors this exactly.
- crates/engine/src/game/engine_priority.rs:131 — `begin_pending_trigger_target_selection(state)?` early-return after processing new events to surface a trigger's target prompt; mirrored in my block with the same `consumed_before_priority_trigger_events` cleanup.

## Test plan
- New regression test `exile_return_fires_returned_creatures_etb_trigger` (crates/engine/src/game/engine_exile_return_tests.rs): Fiend Hunter exiles a Wall-of-Omens-class creature via an `UntilSourceLeaves` link; when the source leaves, the returned creature's ETB must be on the stack and resolving it draws a card. Fails before the fix (empty stack), passes after.
- All 16 exile-return tests pass; existing tests returning non-ETB permanents are unaffected, confirming the added scan is a no-op where no ETB matches.
- Full engine lib suite: 16517 passed, 0 failed. `cargo clippy -p engine` clean. `cargo fmt` clean.

## Gate A
```
Gate A PASS head=900b7ba3817dc3c856f2c1ec19955cc5e6d9132c base=7e1ddc9b778c3c3bcfa9e56d1f726561225c099d
```

Model: claude-opus-4-8
Thinking: High
Tier: Frontier


matthewevans (changes requested):
**Blocked — returned ETBs are found, but trigger ordering is split across two batches.**

## 🔴 Blocker

- `crates/engine/src/game/engine_priority.rs:191` calls `process_triggers` for the new return events, then `:208` calls `check_delayed_triggers` for the same post-action batch. Both routes can install `WaitingFor::OrderTriggers`; `check_delayed_triggers` does so through `begin_trigger_ordering` in `crates/engine/src/game/triggers.rs:5846`, which overwrites the normal-return ordering prompt. A return that produces multiple normal ETBs plus a matching delayed trigger can therefore lose the first ordering context or present only the later batch, contrary to the simultaneous-trigger ordering requirement.

  The added Wall of Omens case proves one normal ETB reaches the stack, but it never combines that ETB batch with a delayed trigger or exercises a same-controller ordering prompt. Route the normal return contexts and delayed contexts through the existing combined ordering authority (`process_triggers_with_delayed_phase_events`, or a generalized sibling that preserves the delayed dispatch semantics), then add a regression with both sources that asserts one complete ordering prompt/dispatch.

## ✅ Clean

- The parse-diff artifact reports no card-parse changes, and the test does drive the actual exile-return post-action pipeline.

**Recommendation: request changes — unify the trigger batch before ordering, then cover the normal-plus-delayed interaction.**

mike-theDude:
Addressed the trigger-ordering blocker in 7a48ac45f.

What changed:
- Added a generalized combined normal+delayed trigger path for non-phase event batches while preserving delayed trigger dispatch origin semantics.
- Routed exile-return ETB events through that combined path so normal ETBs and matching delayed triggers share one CR 603.3b ordering prompt.
- Added a post-action pipeline regression asserting two normal ETB observers plus one matching delayed trigger appear in one OrderTriggers prompt.

Local verification:
- cargo fmt --all
- cargo test -p engine exile_return_combines_normal_and_delayed_triggers_in_one_ordering_prompt
- cargo clippy --all-targets -- -D warnings
- cargo test -p engine

matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The revised pipeline batches ordinary return ETBs with matching delayed triggers before the shared ordering decision, preserving the existing target/mode prompt path and preventing duplicate delayed processing.
- The regression test exercises the real post-action pipeline and proves all three triggers appear in one ordering prompt.

**Recommendation: Approved and queued via the maintainer handler.**

PR #5774, opened 2026-07-14
fix(parser): return Auras via "do the same for <type> cards" (Estrid, the Masked)

### LINKED ISSUE #4779: Estrid doesn't return Auras — Playing with [[Estrid, The Masked]] i found an issue with her ultimate.
Reported in Discord: https://discord.com/channels/1485498006781427802/1521831410565120050/1521831410565120050

**Thread:** Estrid doesn't return Auras
**Cards:** estrid, the masked
**Parser status:** has_gaps

## Summary
Playing with [[Estrid, The Masked]] i found an issue with her ultimate.

### estrid, the masked
[+2]: Untap each enchanted permanent you control.
[−1]: Create a white Aura enchantment token named Mask attached to another target permanent. The token has enchant permanent and umbra armor.
[−7]: Mill seven cards. Return all non-Aura enchantment cards from your graveyard to the battlefield, then do the same for Aura cards.
Estrid, the Masked can be your commander.

---
<sub>report_id: `discord:1521831410565120050:1521831410565120050:0` · discord: `1521831410565120050/1521831410565120050`</sub>

lourincedaging0-commits:
Fixes #4779 — Estrid, the Masked's ult never returns Auras.

## The bug

Estrid's `[−7]` ends with **"Return all non-Aura enchantment cards from your graveyard to the battlefield, then do the same for Aura cards."** The `", then do the same for Aura cards"` tail was **silently dropped**, so only the non-Aura enchantments came back — Auras were never returned.

Two causes, both parser-side:
1. **Chunking** — `split_comma_clause_boundary` only splits `", then <verb>"` when the verb is in the imperative table. `"do the same for …"` isn't, so the tail was glued into the first clause (its whole text ended up as the first ability's description) and never dispatched.
2. **No recognizer** — even split off, there was no handler for the `"do the same for <type>"` effect-replication form (the existing `"do the same for <keyword list>"` and `"target opponent does the same"` handlers don't cover it).

## The fix (parser-only — no new engine variant)

- `split_comma_clause_boundary`: treat `", then do the same for <type>"` as a `Then` boundary — directly mirroring the `"faces a villainous choice"` guard right above it (same "verb not in the imperative table" situation).
- New `try_parse_do_the_same_for_type` recognizer + chunk-loop dispatch that **clones the antecedent sibling effect and swaps its type filter**. This reuses the exact antecedent-clone mechanic `try_parse_scoped_does_the_same` uses for the player-scoped fan-out, so it emits an ordinary sibling `Effect` — no new disposition, resolver, or scope.

Result — Estrid now emits **both** returns, zones and controller preserved (CR 608.2c: the antecedent action is replicated modulo the stated type substitution):
- primary: `ChangeZoneAll` Graveyard→Battlefield, `[Enchantment, Non(Aura)]`, controller You
- sub-ability: `ChangeZoneAll` Graveyard→Battlefield, `[Aura]`, controller You

Building-block level: covers the **`do the same for <type>`** clause class, not Estrid alone.

## Verification

- New building-block test `do_the_same_for_type_replicates_mass_zone_change_with_swapped_type` → passes.
- **Full `engine` lib suite: 16526 passed, 0 failed** — the `", then do the same"` chunk-split change causes zero regressions.
- Parser-combinator gate → PASS.

## Summary by CodeRabbit

* **New Features**
  * Added support for Oracle text using “then do the same for \<type\> cards/permanents.”
  * Replicates applicable effects with the specified card type while preserving relevant zones and controller details.

* **Bug Fixes**
  * Prevents unsupported or ambiguous continuations from generating incorrect effects.

* **Tests**
  * Added coverage for type substitutions and rejection of unsupported continuation patterns.


matthewevans (changes requested):
**Changes requested — the general `do the same for <type>` path silently mishandles a parser-visible sibling class.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/mod.rs:25403-25413` clones the antecedent and replaces only `TypedFilter::type_filters`. That is insufficient for **Gruesome Menagerie**: its verified Oracle text is “Choose a creature card with mana value 1 in your graveyard, then do the same for creature cards with mana value 2 and 3. Return those cards to the battlefield.” The continuation represents two further selections, and their mana-value predicates are `FilterProp` values, not `type_filters`; this branch can neither replace `1` with `2`/`3` nor emit both sibling actions. The current parse-diff already names Gruesome Menagerie as changed, while the new test covers only Estrid. Keep this form strict-failing until its full substitution/repetition grammar is modeled, or implement a typed continuation that carries the complete replacement filter(s) and add regressions for Gruesome plus Glimpse of Tomorrow’s tracked-set/“put the rest” chain.

## 🟡 Non-blocking

- `crates/engine/src/parser/oracle_effect/sequence.rs:7551-7554` also accepts `repeat this process for`, which is a broader continuation family than a same-action type substitution. Narrow the recognizer unless every accepted form is modeled by the same typed continuation.

## ✅ Clean

- Estrid’s exact Aura sibling is a useful target case, and the parser uses the shared nom/type-phrase infrastructure.
- Current CI is green; the parse-diff artifact correctly exposes the wider five-card blast radius that needs coverage.

**Recommendation: generalize the continuation around complete filter substitution and cardinality, add the affected-card regressions, then re-request review.**

lourincedaging0-commits:
Thanks — good catch on the blast radius. I took the "keep this form strict-failing until the full substitution/repetition grammar is modeled" path and narrowed the change so it touches **only** the clean pure-type substitution (Estrid's `Aura cards`). Pushed in `b4b705e`:

**Recognizer** (`try_parse_do_the_same_for_type`)
- Dropped the broader `repeat this process for` family (your non-blocking point).
- Accepts a filter **only** when it is a whole card-type substitution: rejects any filter carrying a `FilterProp` predicate (Gruesome Menagerie's `creature cards with mana value 2 and 3`) or a controller scope. The multi-type list (Grim Captain's Call's `Vampire, Dinosaur, and Merfolk`) was already rejected by the whole-consumption remainder guard.

**Chunk split** (`split_comma_clause_boundary`)
- The `, then do the same for …` boundary is now gated on that same recognizer, so an unmodeled continuation is **not split at all** and stays glued exactly as before. Verified against the five-card blast radius:

| Card | Before this PR | After |
|------|----------------|-------|
| Estrid, the Masked | Aura return silently dropped | two `ChangeZoneAll` (non-Aura, then Aura) |
| Gruesome Menagerie | glued | **unchanged** (not split; `ChangeZoneAll=0`, `Unimplemented=0`) |
| Grim Captain's Call | glued | **unchanged** |
| Firemind's Foresight | (`repeat this process for`) | **unchanged** (no longer matched) |

**Dispatch** — only emits the retyped clone when a `Typed` target was actually swapped, so an `Unimplemented`/non-typed antecedent is never cloned verbatim.

So the richer substitution/cardinality forms remain strict-failing until that grammar lands, per your recommendation — I did not try to half-model them here.

**Verification:** added recognizer unit tests (accepts `Aura cards`; rejects the mana-value, multi-type, filter-predicate, and `repeat this process for` forms) next to the Estrid end-to-end test. Full `engine` lib suite: **16527 passed, 0 failed**; parser-combinator gate: PASS. Re-requesting review.

matthewevans (changes requested):
**Blocked — the pure-type continuation still has a real sibling whose type filter is nested, not top-level.**

## 🔴 Blocker

- **Glimpse of Tomorrow** is the other current Oracle instance of the supposedly supported pure form: “Put all non-Aura permanent cards revealed this way onto the battlefield, then do the same for Aura cards.” It passes `try_parse_do_the_same_for_type` because its continuation is the clean `Aura cards` form, but its antecedent carries the revealed-set provenance in `TargetFilter::TrackedSetFiltered`, not a top-level `TargetFilter::Typed`.
- `crates/engine/src/parser/oracle_effect/mod.rs:25415-25420` only swaps when the filter handed to the walker is directly `Typed`; it does not replace the inner filter of `TrackedSetFiltered`. `swapped` therefore stays false for this sibling, so the clone is not emitted. The current parse-diff independently shows only Estrid gained a signature, and the added tests exercise only Estrid plus the recognizer in isolation.

**Recommendation: either extend the substitution at the tracked-set-filter authority (preserving the “revealed this way” provenance) and add a Glimpse regression alongside Estrid, or narrow the accepted form to the actual top-level-target shape. Do not claim the generic pure-type class until both production parser shapes are covered.**

matthewevans:
This PR still has unresolved requested changes on head `b4b705e3191428a2a91178e7c48befa838b21742`, and no contributor follow-up after the blocking review. Please address the requested changes and push an updated head within 7 days. If they remain unaddressed, this PR will be auto-closed under the requested-changes expiry policy.

matthewevans:
@coderabbitai review

matthewevans (changes requested):
**Changes requested — the current head still accepts a pure-type sibling whose type filter is nested in tracked-set provenance, then silently emits no repeated action.**

## 🔴 Blocker

- crates/engine/src/parser/oracle_effect/mod.rs:27112-27129 recognizes the Aura-cards continuation and swaps only a directly Typed filter. crates/engine/src/parser/oracle_effect/mod.rs:30094-30103 wraps revealed-this-way targets in TargetFilter::TrackedSetFiltered. Glimpse of Tomorrow says: “Put all non-Aura permanent cards revealed this way onto the battlefield, then do the same for Aura cards.” Its antecedent therefore has a nested typed filter; swapped remains false and the Aura clone is omitted. Head 1f4bbde only merges main over the previously reviewed implementation, so it does not address this blocker.

  Preserve the tracked-set provenance while substituting the nested type filter at its owning filter traversal and cover the Glimpse resolution/parse path; alternatively narrow this continuation until the nested form is modeled. The Estrid-only assertion does not exercise this production shape.

## 🟡 Non-blocking

- CodeRabbit observations remain valid: the recognizer return type permits the wildcard-to-empty-vector fallback at mod.rs:27113-27115, and the new regression is Estrid-specific. Resolve both with the blocker fix so the typed-only contract and sibling coverage are enforced.

**Recommendation: request changes. Add the nested tracked-set sibling regression and a typed traversal that preserves provenance, then re-request review on a non-merge-only head.**

matthewevans:
@coderabbitai full review

matthewevans:
Closing this PR because the requested changes on current head `1f4bbde70544281fdfdb2ac7f3a87d590c15a637` remain unresolved and there has been no contributor follow-up since the expiry warning of 2026-07-21: https://github.com/phase-rs/phase/pull/5774#issuecomment-5030893042

The outstanding blocker is the accepted pure-type continuation that silently omits the nested tracked-set sibling (Glimpse of Tomorrow), documented in the current-head review: https://github.com/phase-rs/phase/pull/5774#pullrequestreview-4749442667. Please open a fresh PR from current `main` when the full tracked-set substitution/provenance path and its regression are ready.

PR #5776, opened 2026-07-14
fix(engine): exile-and-return resolving-spell replacement for Feather, the Redeemed (#4825)

### LINKED ISSUE #4825: Feather, the Redeemed sends targeted spells to graveyard instead of returning them
Reported in Discord: https://discord.com/channels/1485498006781427802/1522010837462814781/1522010837462814781

**Thread:** Feather, the Redeemed
**Cards:** feather, the redeemed
**Parser status:** has_gaps

## Summary
Instant or sorcery spells that targeted a creature controlled by Feather’s controller go to the graveyard instead of being exiled and returned to hand at the next end step.

### feather, the redeemed
Flying
Whenever you cast an instant or sorcery spell that targets a creature you control, exile that card instead of putting it into your graveyard as it resolves. If you do, return it to your hand at the beginning of the next end step.

---
<sub>report_id: `discord:1522010837462814781:1522010837462814781:0` · discord: `1522010837462814781/1522010837462814781`</sub>

mike-theDude:
Fixes #4825. Instant/sorcery spells that triggered Feather, the Redeemed went to the graveyard instead of being exiled and returned to hand at the next end step.

## Root cause
The "exile that card instead of putting it into your graveyard as it resolves" clause lowered to `ChangeZone{origin:Graveyard, dest:Exile, ParentTarget}`, a runtime no-op (CR 400.7 origin guard) because the spell is still on the stack when Feather's trigger resolves — so it took the CR 608.2n graveyard default.

## Fix (built for the class, not the card)
Lower the clause to the existing `Effect::ExileResolvingSpellInsteadOfGraveyard` (the stack-exile-marker mechanism consulted at spell resolution, already used by Rod of Absorption). Parameterized it with a typed `then_return: Option<ExiledSpellReturn>{ destination, timing }` so the "If you do, return it at the next end step" rider is a parameter value, not a new variant — covering the whole class. A conservative AST fold consumes the rider clause.

The one-shot return delayed trigger is armed **when the replacement is actually applied** (the spell lands in exile) per CR 603.7a, keyed on the concrete exiled object. The marker is cleared on any zone exit (CR 400.7), so a spell **countered or fizzled in response** correctly goes to the graveyard and never returns.

## Testing
- Full engine suite green: 16516 lib + 2972 integration tests.
- New `issue_4825` module (7 tests): parser shape (rider folded, no no-op `ChangeZone`), riderless negative, Rod-of-Absorption regression, serde back-compat, runtime exile→return, counter-negative, fizzle-negative.
- Card-data export regenerated: Feather faithful (`ExileResolvingSpellInsteadOfGraveyard{then_return:{Hand, AtNextPhase{End}}}`, no leftover sub); Rod byte-identical (riderless form).
- Rebased onto latest `main`; targeted suite re-run green post-rebase.

## Non-blocking follow-ups (noted, out of scope)
- CR 616.1 flashback + Feather replacement-ordering choice is collapsed to the strictly-dominant choice (pre-existing static-dest architecture).
- Cross-controller attribution would only matter for a hypothetical opponent-cast carrier of this rider (current class is all "whenever you cast").
- `cargo coverage` / LLM `semantic-audit` deferred to CI / primary-checkout Tilt resources (parse faithfulness already confirmed by tests + regenerated card-data).

### DIFF
--- a/crates/engine/src/analysis/ability_graph.rs
+++ b/crates/engine/src/analysis/ability_graph.rs
@@ -928,7 +928,7 @@ fn effect_projection(effect: &Effect) -> Projection {
         | Effect::AddPendingEntersModifications { .. }
         | Effect::CreateEmblem { .. }
         | Effect::PayCost { .. }
-        | Effect::ExileResolvingSpellInsteadOfGraveyard
+        | Effect::ExileResolvingSpellInsteadOfGraveyard { .. }
         | Effect::PreventDamage { .. }
         | Effect::CreateDamageReplacement { .. }
         | Effect::CreateDrawReplacement { .. }
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -5030,7 +5030,7 @@ fn rw_effect(
             p.writes_external.set(StateKind::StackShape);
             (p, None)
         }
-        Effect::ExileResolvingSpellInsteadOfGraveyard { then_return: _ } => {
+        Effect::ExileResolvingSpellInsteadOfGraveyard { on_exile: _ } => {
             (ext_write(StateKind::StackShape), None)
         }
 
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -3302,7 +3302,7 @@ fn legacy_effect(x: &Effect) -> bool {
         | Effect::SolveCase
         | Effect::SetClassLevel { .. }
         | Effect::AddRestriction { .. }
-        | Effect::ExileResolvingSpellInsteadOfGraveyard
+        | Effect::ExileResolvingSpellInsteadOfGraveyard { .. }
         | Effect::RingTemptsYou
         | Effect::VentureIntoDungeon
         | Effect::VentureInto { .. }
@@ -5030,7 +5030,9 @@ fn rw_effect(
             p.writes_external.set(StateKind::StackShape);
             (p, None)
         }
-        Effect::ExileResolvingSpellInsteadOfGraveyard => (ext_write(StateKind::StackShape), None),
+        Effect::ExileResolvingSpellInsteadOfGraveyard { then_return: _ } => {
+            (ext_write(StateKind::StackShape), None)
+        }
 
         // ---- Pool ----
         Effect::Mana {
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -1071,7 +1071,10 @@ fn scan_effect(x: &Effect) -> Axes {
             acc = acc.or(scan_target_filter(filter));
             acc
         }
-        Effect::ExileResolvingSpellInsteadOfGraveyard => Axes::NONE,
+        // The `then_return` rider is fixed at parse time and only read by the
+        // stack-resolution router when the replacement applies — no game-state
+        // read happens at scan time, so NONE stays correct.
+        Effect::ExileResolvingSpellInsteadOfGraveyard { then_return: _ } => Axes::NONE,
         Effect::PreventDamage {
             amount_dynamic,
             target,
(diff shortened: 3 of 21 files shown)


matthewevans (changes requested):
**Blocked — Lilah’s `If you do, it becomes plotted` continuation still runs before the exile replacement is applied.**

## 🔴 Blocker

- [`crates/engine/src/parser/oracle_effect/lower.rs:1902`](https://github.com/phase-rs/phase/blob/3993957d472059208f0d77783970878766c43485/crates/engine/src/parser/oracle_effect/lower.rs#L1902) folds only the Feather-shaped `CreateDelayedTrigger`/`Bounce` continuation; a structurally different sub-ability is deliberately restored at lines 1921–1923. The parse-diff identifies Lilah as using the new carrier, and its official Oracle text says: “If you do, it becomes plotted.” That leaves Lilah’s `GrantCastingPermission { Plotted, ParentTarget }` as an ordinary chained effect.
- [`crates/engine/src/game/effects/grant_permission.rs:74`](https://github.com/phase-rs/phase/blob/3993957d472059208f0d77783970878766c43485/crates/engine/src/game/effects/grant_permission.rs#L74) resolves `ParentTarget` and then stamps/pushes `Plotted` at lines 201–247 while the spell is still on the stack. [`crates/engine/src/game/zones.rs:315`](https://github.com/phase-rs/phase/blob/3993957d472059208f0d77783970878766c43485/crates/engine/src/game/zones.rs#L315) intentionally preserves `Plotted` on every Stack exit. Thus a Lilah spell that is countered or fizzles can acquire a plotted permission without ever being exiled, then retain it if moved to exile later. The new runtime coverage exercises Feather only, so it does not discriminate this path.

**Recommendation: represent Lilah’s plot grant as an application-time rider of the stack-exile carrier, alongside Feather’s delayed-return rider, and add runtime tests for both successful Lilah exile-and-plot and a countered/fizzled spell receiving neither the plot permission nor `BecomesPlotted` event before re-requesting review.**

mike-theDude:
@matthewevans thanks — fixed. Lilah's plot is now an **application-time rider** of the stack-exile carrier, symmetric to Feather's return rider, exactly as recommended.

**What changed** (latest commit):

- Generalized the carrier rider from the return-only `ExiledSpellReturn` struct to a sum type `ExiledSpellRider { ReturnTo { destination, timing }, BecomePlotted }` on `Effect::ExileResolvingSpellInsteadOfGraveyard { on_exile }`. Both members are CR 603.7a / CR 702.170c consequences of the *same* replacement application.
- The stack router (`stack.rs::resolve_top`) applies the rider **after** the stack→exile move, via `apply_exile_rider`. `BecomePlotted` routes through the single `grant_permission::resolve` authority (stamps `turn_plotted`, emits `BecomesPlotted`, binds the permission to the card's owner). The transient marker is cleared on any *other* zone exit (`zones.rs`), so a countered or fizzled spell takes neither the plot nor the return.
- Parser: `"it becomes plotted"` is now a standalone effect building block. It previously only parsed as a continuation-patch after an Exile `ChangeZone`; as the body of Lilah's reflexive `"if you do, it becomes plotted"` rider it was lowering to `Unimplemented { name: "become" }` on the current head — so the plot never applied at all. The fold (`fold_exile_resolving_rider`) now folds both Feather's return and Lilah's plot onto the typed rider; the swallow-check suppression and coverage display cover both members. Lilah is now fully supported (no `Unimplemented`).

**Tests added:**
- `lilah_exiles_multicolored_spell_and_plots_it` — the exiled spell gains the `Plotted` permission bound to its owner; no delayed trigger is armed.
- `countered_lilah_spell_goes_to_graveyard_and_is_not_plotted` — the countered spell hits the graveyard and is never plotted; the rider is cleared on stack exit.
- Parser SHAPE (`lilah_parses_plotted_rider_onto_carrier`), serde round-trip for both rider members, and a swallow-check regression for Lilah's `"if you do"`.

Full engine suite + clippy green locally. Re-requesting review.

matthewevans (approved):
**Verdict: approved — the current head resolves the requested Lilah timing defect at the same application-time carrier as Feather.**

## 🔴 Blocker

- None. The prior changes request is resolved: `BecomePlotted` is now an `ExiledSpellRider`, applied only after the resolving spell reaches exile.

## 🟡 Non-blocking

- None.

## ✅ Clean

- `crates/engine/src/game/effects/exile_resolving_spell.rs`: plot delegates through the existing casting-permission authority at the correct replacement-application seam.
- `crates/engine/tests/integration/issue_4825_feather_the_redeemed.rs`: successful and counter/fizzle paths discriminate both Feather and Lilah behavior.
- Current parse-diff is exactly Feather and Lilah; fresh GitHub CI is green.

Issue #1095, opened 2026-05-26
Hogaak, Arisen Necropolis — Mana should not be able to be spent to cast Hogaak, Arisen Necropolis.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1508924702884761700/1508924702884761700

**Thread:** Hogaak, Arisen Necropolis
**Cards:** hogaak, arisen necropolis
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Mana should not be able to be spent to cast Hogaak, Arisen Necropolis.

### hogaak, arisen necropolis
You can't spend mana to cast this spell.
Convoke, delve (Each creature you tap while casting this spell pays for {1} or one mana of that creature's color. Each card you exile from your graveyard pays for {1}.)
You may cast this card from your graveyard.
Trample

---
<sub>report_id: `discord:1508924702884761700:1508924702884761700:0` · discord: `1508924702884761700/1508924702884761700`</sub>

matthewevans:
### Coverage classifier (advisory)

**Verdict:** `supported_aspect_defect`
**Matched clause:** *missing entirely from parse_details*

**AST finding:** The Oracle text leads with **"You can't spend mana to cast this spell."** This clause is **not present in parse_details at all**. The parsed nodes are: Convoke, Delve, Trample (keywords), CastFromZone, GraveyardCastPermission. There is no MannaProhibition / CastingRestriction / similar node encoding the "can't spend mana" restriction.

The user's report ("Mana should not be able to be spent to cast Hogaak") aligns exactly: the engine has no AST representation of the restriction, so the casting flow lets mana be spent normally. Secondary issues noted by the reporter ("targets a player on the stack when it shouldn't", "delve doesn't work", "can't select to cast with convoke") suggest the alt-cost-only-via-Convoke/Delve flow needs end-to-end investigation.

Investigation: parser needs to recognize "You can't spend mana to cast this spell" as a casting restriction; casting flow needs to enforce it. May share infrastructure with Force of Will-style "without paying its mana cost" cases, though the polarity is opposite.

mike-theDude:
## Triage
**Area:** parser · **Priority:** p2-wrong-game-result · **Status:** confirmed

**Reported behavior:** The engine lets a player spend mana to cast Hogaak, Arisen Necropolis, but the card can only be cast via Convoke/Delve (never with mana).

**Oracle text (verified from card-data.json):**
> You can't spend mana to cast this spell.
[...]

PR #5777, opened 2026-07-14
fix(engine): enforce Hogaak's "can't spend mana to cast" (#1095)

nickmopen:
## Summary

Hogaak, Arisen Necropolis (issue #1095) — `You can't spend mana to cast this spell.` The line was dropped by the parser as an `Effect::Unimplemented`, so nothing enforced it and mana could be spent to cast Hogaak (a `p2-wrong-game-result`). This fixes it end to end.

- **Parser**: recognize the line in `parse_casting_restriction_line` → new `CastingRestriction::CantSpendMana` (compositional `nom`; straight/curly apostrophe + `~` self-name form).
- **Runtime**: `build_spell_meta` sets `SpellMeta::cant_spend_mana` from the object's casting restrictions; the mana-payment eligibility layer (`mana_payment::ctx_permits_unit`) makes real (non-convoke) pool units ineligible under such a spell context. The same predicate backs both affordability and the spend, so Hogaak can only be paid by convoke/delve and the coverage line is honestly supported.

Addresses the review on the prior revision (model at the spell-payment seam via the payment context; keep coverage honest; add discriminating runtime coverage).

Method: /engine-implementer — parser recognition + a payment-context spend gate.

## CR references

- CR 601.2g / CR 118.3 — "You can't spend mana to cast this spell" (a spell-cost-payment restriction, not a timing window).

## Verification

- `cargo clippy -p engine --tests` — clean
- `scripts/check-parser-combinators.sh` — Gate A PASS (below)
- `cargo test -p engine --lib` — 16535 passed
- `cargo test -p engine --test integration` — 3000 passed (incl. new `hogaak_cant_spend_mana_1095`: convoke pays Hogaak with no mana spent; pool mana alone cannot pay it)

## Gate A

Gate A PASS head=ef312670fbee3eee0c532e44f64147cfc6bbcc71 base=efe74096ba6c2185446b5ef319bbdd0e46c7f5e5

## Anchored on

- crates/engine/src/game/mana_payment.rs:1806 `ctx_permits_unit` — layered onto the existing per-unit `restriction.allows(ctx)` spend-restriction gate used across every eligibility predicate (affordability + spend).
- crates/engine/src/game/casting.rs `build_spell_meta` — sets `SpellMeta::cant_spend_mana` beside `is_face_down`, the existing precedent for a spell-cost spend flag threaded through `PaymentContext::Spell`.

## Claimed parse impact

- Hogaak, Arisen Necropolis


matthewevans (changes requested):
**Blocked — the parser recognizes Hogaak’s clause, but the game still permits spending mana to cast it.**

## 🔴 Blocker

- `crates/engine/src/game/restrictions.rs:1118` makes `CastingRestriction::CantSpendMana` unconditionally true, and a repository-wide use of that new variant stops there: no cast/payment path receives it. In particular, the actual pool-payment authority is `crates/engine/src/game/casting.rs:13016`, so Hogaak can still consume ordinary mana despite `crates/engine/src/types/ability.rs:15529` claiming that the mana-payment path excludes it.

  This is also a coverage-honesty regression: `crates/engine/src/game/coverage.rs:8617` now marks the line supported while the printed restriction is not implemented. A `CastingRestriction` is the timing gate; this is a spell-cost-payment constraint and must travel through the cast/payment context to filter pool mana at every spell-payment entry point. Keep the clause unimplemented/coverage-red until that production seam is wired, or implement the payment restriction there.

- The added tests stop at parser shape and a debug-string assertion. Add a real cast-pipeline regression that proves pool mana cannot pay Hogaak, alongside a legal convoke/delve payment case, so the result would fail if the payment gate is removed.

## ✅ Clean

- The new phrase parser is compositional `nom` and covers straight/curly apostrophes plus the self-reference form; the defect is downstream semantic enforcement, not recognition.

**Recommendation: request changes — model this at the spell-payment seam, retain honest coverage until it is live, and add discriminating runtime payment coverage.**

nickmopen:
Both blockers addressed in `625654f` — the restriction is now enforced at the spell-payment seam, and coverage stays honest.

**🔴 Enforcement.** `CastingRestriction::CantSpendMana` now travels through the payment context: `build_spell_meta` sets `SpellMeta::cant_spend_mana`, and `mana_payment::ctx_permits_unit` makes real (non-convoke) pool units ineligible under that spell context. It's layered onto the existing per-unit `restriction.allows(ctx)` gate, so **every** eligibility predicate — affordability (`can_pay_for_spell`, `reduce_cost_by_pool`) and the spend (`spend_eligible` / generic / pin scans) — filters pool mana identically. Net effect: pool mana can't pay Hogaak, and because affordability now agrees the pool can't cover it, the cast correctly routes onto convoke/delve. The coverage line is honest because the restriction is live.

**🔴 Discriminating runtime coverage.** New registered `hogaak_cant_spend_mana_1095`:
- `convoke_pays_hogaak_without_spending_mana` — real cast pipeline: two creatures convoke the whole cost → Hogaak resolves, `mana_spent_to_cast == false`.
- `pool_mana_alone_cannot_pay_hogaak` — with a full pool and no convoke/delve, the cast is rejected and Hogaak stays in hand with its pool intact. Remove the `ctx_permits_unit` gate and this fails.

Verification: `cargo test -p engine --lib` 16535 pass, `--test integration` 3000 pass, fmt/clippy clean, Gate A PASS. No regressions in the convoke/delve/improvise suites.

matthewevans (changes requested):
**Request changes — the `{S}` payment path can still spend real mana despite `CantSpendMana`.**

## 🔴 Blocker

- `spend_snow_unit` takes no `PaymentContext` and selects any snow unit (`mana_payment.rs:1162-66`, `1489-92`, `2369-77`). Both the live payment and the affordability simulation therefore bypass `ctx_permits_unit`; a spell with `cant_spend_mana` can pay a `{S}` shard from the mana pool. The restriction is modeled as a general casting rule, so every mana-payment route must apply the same gate.

## ✅ Clean

- The updated parser uses composable nom branches, and the ordinary colored, hybrid, generic, and `{Z}` routes now consistently call the new eligibility helper. The new Hogaak scenarios exercise the parser → casting metadata → payment pipeline and are meaningful regressions.

**Recommendation: thread the optional spell context through the snow helper (including its simulated use), apply `spell_permits_unit` to both pinned and fallback selection, and add a focused regression proving snow mana cannot pay when `cant_spend_mana` is set.**

nickmopen:
Good catch — fixed in `ef31267`.

**🔴 {S} snow path.** `spend_snow_unit` (and its `spend_snow` wrapper) now takes the optional `PaymentContext` and applies `spell_permits_unit` to **both** the pinned and fallback selection, exactly like `spend_two_or_more_color_source_eligible`. All four call sites — the live `{S}` spend (`mana_payment.rs:1165`), the affordability sim (`:698`), and the two dry-run/backup probes (`:934`, `:1491`) — thread `spell` through, so no snow-payment route bypasses the gate.

**Regression.** Added `snow_shard_respects_cant_spend_mana`: an ordinary spell pays `{S}` from a real snow unit, but a `cant_spend_mana` spell cannot (both `can_pay_for_spell` and the live `spend_snow_unit` refuse, leaving the snow unit unspent). Remove the gate from the snow helper and it fails.

Verification: `cargo test -p engine --lib mana_payment::` 63 pass (incl. the new snow test + 11 existing snow tests), hogaak integration 2 pass, fmt/clippy clean, Gate A PASS.

matthewevans (approved):
Approved after current-head implementation review and maintainer branch update.

PR #5781, opened 2026-07-14
fix(parser): parse count-form "would draw N or more cards" draw replacements (CR 121.2a)

### LINKED ISSUE #5678: Alms Collector: 'would draw two or more cards' antecedent never parses to a Draw replacement (CR 121.2a instruction-count gap)
## Summary

**Alms Collector never becomes a `ReplacementEvent::Draw` definition.** Its antecedent is a count-form "would draw **two or more** cards", and the parser's draw-replacement branch only matches `"would draw a card"` and `"would draw one or more cards"`.

Oracle text, verified from the pool (`data/card-data.json`, not recalled):

> Flash
> If an opponent would draw two or more cards, instead you and that player each draw a card.

This is a textbook CR 121.2a instruction-count replacement: it modifies the draw **instruction** before any individual card draw happens ("See rule 616.1g"). It should lower to a `Draw` replacement with `draw_scope: InstructionCount`.

## Where

`crates/engine/src/parser/oracle_replacement.rs`, the draw-replacement branch of `parse_replacement_line_inner`. The `alt()` that recognises the antecedent:

```rust
alt((
    value(DrawReplacementScope::IndividualDraw,   tag("would draw a card")),
    value(DrawReplacementScope::InstructionCount, tag("would draw one or more cards")),
))
```

There is no arm for `"would draw two or more cards"`, so `mentions_draw` is false and the line never produces a replacement at all.

## Class size (measured, not estimated)

Scanning every card's Oracle text in the pool for a `would draw <N> or more` antecedent:

| antecedent | cards |
|---|---|
| `would draw a card` | 62 |
| `would draw one or more cards` | 1 (Quantum Riddler) |
| `would draw two or more cards` | **1 (Alms Collector)** |
| `would draw three or more …` | 0 |

So the count-form gap is a class of exactly **one card** today. A fix should still be written as a class (parameterise the threshold) rather than a `tag("two or more")` special case, per the build-for-the-class rule — the phrasing is an open-ended `N or more` axis even if only two values are currently printed.

## Two notes for whoever fixes this

**1. The corpus gate will go red, and that is the designed behaviour.** `scripts/draw-replacement-corpus.tsv` is a frozen 51-row baseline. Fixing this parser gap adds Alms Collector as a **52nd row**, so `draw_replacement_census.py --corpus --check` will fail until the baseline is re-frozen with `--write`. That is the gate working, not a regression — re-freeze it in the same commit and say so in the message.

Alms Collector will also need its scope to satisfy the gate's **cross-check**: the census independently derives the required `DrawReplacementScope` from the definition's shape and fails if the producer disagrees. Expect `InstructionCount`.

**2. This does not contradict the Plan 03 A2 census.** A2 recorded "InstructionCount bucket = Quantum Riddler only", and that remains true **of the parsed corpus** — which is the population A2 measured. Alms Collector is a *pool* occupant the corpus cannot see, precisely because of this gap. Both statements are correct about their own populations.

## Explicitly NOT claimed here

12 cards with a `"would draw a card"` antecedent are also absent from the parsed Draw corpus (Aladdin's Lamp, Mangara's Tome, Ring of Ma'rûf, Scion of Halaster, The Value Knight, Trash Panda, Urabrask Heretic Praetor, and the five Words of War/Waste/Wilding/Wind/Worship). **This is not a list of 12 parser gaps.** At minimum the Words cycle is handled correctly by a different mechanism — those are runtime shields installed at resolution time by `game/effects/create_draw_replacement.rs` ("the next time you would draw a card this turn…"), so they legitimately carry no static Draw replacement in the card data. The remainder are unclassified and would need triage before anyone calls them defects. Recording the observation so it isn't lost, not asserting a count.

nghetienhiep:
Alms Collector ("If an opponent would draw two or more cards, instead you
and that player each draw a card") never became a ReplacementEvent::Draw
definition. The draw-replacement antecedent alt() only matched
"would draw a card" and "would draw one or more cards", so the
two-or-more line set mentions_draw = false and produced no replacement at
all.

CR 121.2a: a count-form antecedent modifies the draw *instruction* before
any individual card draw happens, so its scope is InstructionCount. The
threshold is an open-ended axis, not a fixed pair of cardinals, so parse
it with nom_primitives::parse_number instead of enumerating each word as
its own tag(). This folds the existing "one or more" (Quantum Riddler)
and the new "two or more" (Alms Collector) into one parameterized arm and
covers any future printed threshold. This matches the convention the
Forge converter already uses (mtgish-import convert_replace_would_draw
maps both one-or-more and two-or-more to InstructionCount).

The frozen draw-replacement corpus gains Alms Collector as a 52nd row
(scripts/draw-replacement-corpus.tsv), as the census gate is designed to
require. Its cross-check derives the required scope from the definition
shape, but the execute-shape rule only recognizes InstructionCount via an
EventContextAmount substitute ("draw that many cards plus one") — Alms
Collector's substitute is a fixed "each draw a card", so classify_scope
would misread it as IndividualDraw. Teach classify_scope the count-form
antecedent signal (derived independently from the description, mirroring
the parser) so the cross-check agrees on InstructionCount. The added row
was derived by running the census's own corpus_rows() against the
parser's real output, so the CI --corpus --check (which regenerates
card-data.json) matches.

Closes #5678


matthewevans (changes requested):
**Blocked — the parser recognizes `N` but discards it before the draw-replacement matcher can enforce it.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_replacement.rs:419-436` maps every `"would draw <N> or more cards"` antecedent to the unit variant `DrawReplacementScope::InstructionCount`; the parsed number is not retained. `crates/engine/src/game/replacement.rs:1905-1907` then accepts every positive `ProposedEvent::Draw`, and no runtime path consumes `InstructionCount` to compare the event count. The [official Alms Collector Oracle text](https://scryfall.com/card/c17/2/alms-collector) says, “If an opponent would draw two or more cards, instead you and that player each draw a card.” A one-card draw must therefore not be replaced.
- The new test at `crates/engine/src/parser/oracle_replacement.rs:16883-16928` asserts only the parsed definition. It never drives a one-card and a two-card `ProposedEvent::Draw` through the replacement pipeline, so it cannot catch the lost threshold or prove the claimed runtime support.

Suggested fix: parameterize the existing instruction-count scope with its minimum threshold (rather than adding a parallel replacement type), carry `parse_number` into that value, enforce it at the draw matcher/definition applicability seam, and add a runtime regression proving Alms Collector ignores draw one and replaces draw two.

## ✅ Clean

- The count-form `nom` shape and the independent corpus cross-check are the right parser/census direction; they become valid support once the same threshold reaches the runtime authority.

**Recommendation: request changes for the threshold-threading and runtime coverage.**

matthewevans (changes requested):
**Blocked — the threshold matcher is reached only after production has split a multi-card draw into one-card events.**

## 🔴 Blocker

- `crates/engine/src/game/effects/draw.rs:178-225` starts a draw instruction and immediately calls `draw_through_replacement_with_applied(..., 1, ...)` for each unit. Consequently, the new `event.count >= min` check in `crates/engine/src/game/replacement.rs:5121` sees `count == 1` on every ordinary multi-card draw. Alms Collector's `min: 2` is therefore never applicable in the real pipeline.
- The added runtime test constructs `ProposedEvent::Draw { count: 2 }` and calls `find_applicable_replacements` directly. That proves the matcher in isolation, but bypasses the production split that causes the defect.

## 🟡 Non-blocking

- The parser parameterization and import/census updates correctly preserve the printed threshold; the missing piece is consulting instruction-scoped replacements before the per-card sequence starts.

**Recommendation: move the instruction-scope replacement consult to the pre-split draw-instruction seam and add an end-to-end two-card draw test that proves Alms Collector applies, plus a one-card control that proves it does not.**

matthewevans (changes requested):
**Request changes — a count-form replacement suppresses individual-draw replacements on its own substitute.**

## 🔴 Blocker

- `replace_draw_instruction` keeps `state.draw_consult_scope = Instruction` while it drains the replacement continuation (`replacement.rs:7321-23`, `7348-59`). If that continuation starts a draw, its per-card phase inherits `Instruction`, where `IndividualDraw` definitions are rejected (`replacement.rs:5125-46`). The official Alms Collector text is “If an opponent would draw two or more cards, instead you and that player each draw a card”; each substituted one-card draw must still be eligible for Dredge/Notion-Thief-class individual-draw replacements. CR 121.2a only scopes the *outer* count-form replacement before individual draws; it does not disable replacements on a substituted draw instruction.
- `alms_class_replaces_instruction_at_pre_split_seam` uses a life-gain substitute (`replacement.rs:10579-10612`), so it cannot reach the nested draw path or detect the leaked scope. The returned waiting state from `apply_pending_post_replacement_effect` is also discarded at `7354-57`, so an eligible nested Dredge choice would be lost rather than propagated.

## ✅ Clean

- The threshold is now parameterized and consumed at the pre-split instruction seam, with meaningful one-card and two-card controls; that addresses both prior reviews.

**Recommendation: restore the outer consult scope before draining a substitute, preserve any nested replacement pause through the draw-sequence result, and add an end-to-end fixture where an Alms-Collector-class substitute draw encounters an IndividualDraw replacement (for example Dredge).**

nghetienhiep:
Closing this one — the pre-split instruction-scope interaction with nested individual-draw replacements (Dredge-class on a substituted draw) turned out more involved than fits cleanly here, and I'd rather not leave a partial fix open. Thanks for the thorough reviews; may revisit as a properly scoped follow-up.

PR #5785, opened 2026-07-14
fix(parser): thread multi_target through non-targeted "return up to N cards from graveyard" bounces

### LINKED ISSUE #5784: Ill-Gotten Gains: non-targeted "return up to N cards from graveyard" silently drops its count, bouncing the entire graveyard
## Summary
**Confirmed silent misparse** (verified against `data/card-data.json` and cross-checked with the live parser output on `main`). Not `Unimplemented` — the clause parses as a *valid*-looking `Effect::Bounce`, but the defining quantity is gone.

## Actual parse (current `main`)
The final clause lowers to:
```
Bounce { target: Typed<Card, Owned:ScopedPlayer, InZone:Graveyard>, destination: None, selection: AtResolution }
```
with **no `multi_target` at all**. There is nothing anywhere in the `AbilityDefinition` chain recording "three." At resolution, the `EffectZoneChoice` picker has no bound to enforce, so each player would return their **entire graveyard** to hand instead of up to three cards — the exact opposite of a "restraint" ability card text, and a significant rules violation (CR 601.2c "up to N" semantics never applied).

## Root cause
`TargetedImperativeAst::Return` (the non-targeted, at-resolution bounce variant — used whenever Oracle text omits the word "target", e.g. "return a card from your graveyard to your hand") has no count-carrying field at all, unlike its sibling `Tap`/`Untap`/`Fight`, which already carry `multi_target: Option<MultiTargetSpec>` for exactly this "up to N" shape.

Separately, the "return " clause dispatchers (`parse_targeted_action_ast` in `oracle_effect/imperative.rs` and `try_parse_verb_and_target` in `oracle_effect/mod.rs`) never call the existing `strip_any_number_quantifier` building block for this shape — even though `"return"` is already listed in its `MULTI_TARGET_VERBS` table and the helper is proven correct for the sibling `Tap`/`Untap` "up to N target creatures" pattern. Left unstripped, the leading "up to three " is silently absorbed as unrecognized filler by the downstream destination/target parser.

The *targeted* sibling shape ("return up to two **target** cards from your graveyard to your hand" — Seeds of Renewal, Restock) already works correctly today, via a completely different code path (`strip_optional_target_prefix`, gated on the literal word "target"). Only the non-targeted, "target"-keyword-less shape is affected.

## Runtime is unaffected
`Effect::Bounce`'s at-resolution `EffectZoneChoice` picker already correctly resolves a bounded `ability.multi_target` — this is exercised today by `game::effects::bounce::non_targeted_optional_bounce_still_prompts_at_resolution` and `targeted_up_to_two_graveyard_bounce_moves_chosen_creature` (Wrenn and Six / Macabre Waltz). This is a pure parser-side gap; no game-logic change is required.

## Scope — one confirmed card, a real architectural gap
Ill-Gotten Gains is the only card in the current pool matching this *exact* non-targeted numeric shape (`grep`'d against MTGJSON `AtomicCards.json` for `returns? (up to \w+|that many) cards? from (your|their) \w+ to (your|their) \w+` with no `target` keyword). The sibling card `Bound // Determined` ("Return up to X cards from your graveyard to your hand, where X is the number of colors that creature was") uses the same non-targeted shape but is additionally broken by a separate, pre-existing bug in "where X is `<description>`" CDA-quantity binding (the resolved `Variable` ends up holding the literal description string instead of a resolved quantity) — **out of scope here**, filed as a distinct concern.

### Related, NOT fixed here (follow-up)
The same *class* of defect (a bounded "up to N" quantifier dropped on a non-targeted, at-resolution zone-transfer effect) also reproduces on **Stream of Thought** ("You shuffle up to four cards from your graveyard into your library") via the entirely separate `shuffle ... into` recognizer (`parse_shuffle_ast` in `oracle_effect/imperative.rs`). Different AST family (`ShuffleImperativeAst`, not `TargetedImperativeAst::Return`), so it needs its own fix — flagged here as an untriaged sibling lead, not addressed by the PR accompanying this issue.

## Fix shape
- Add `multi_target: Option<MultiTargetSpec>` to `TargetedImperativeAst::Return`, mirroring the existing `Tap`/`Untap`/`Fight` pattern.
- Wire the existing `strip_any_number_quantifier` building block into both "return " dispatchers, restricted to the genuinely bounded (`max: Some(_)`) result so the unrelated unbounded "any number of" shape (Grave Sifter class, its own pre-existing `return_up_to`/`ReturnToZone.up_to` mechanism) is left untouched.
- Thread the captured spec onto the clause in the shared `lower_imperative_family_ast` interception point, reusing the already-tested `EffectZoneChoice` runtime seam — no game-logic changes.

CR references: CR 601.2c ("up to N" targeting/selection cardinality), CR 115.1d (variable target/object counts), CR 400.6/CR 400.7 (zone-change origin/destination).

jsdevninja:
Fixes #5784.

## Summary

Ill-Gotten Gains's "Exile ~. Each player discards their hand, then returns up to three cards from their graveyard to their hand." silently misparses its defining clause: the final "returns up to three cards from their graveyard to their hand" (no "target" keyword — a non-targeted, at-resolution bounce) lowers to a *valid-looking* `Effect::Bounce` with **no count restriction at all**. At resolution this returns a player's **entire graveyard** to hand instead of up to three cards — the opposite of what the card says.

## Root cause

`TargetedImperativeAst::Return` (the non-targeted bounce variant, used whenever Oracle text omits "target" — e.g. "return a card from your graveyard to your hand") had no count-carrying field, unlike its `Tap`/`Untap`/`Fight` siblings which already carry `multi_target: Option<MultiTargetSpec>` for exactly this "up to N" shape.

Separately, neither "return " clause dispatcher (`parse_targeted_action_ast` in `oracle_effect/imperative.rs`, `try_parse_verb_and_target` in `oracle_effect/mod.rs`) called the existing `strip_any_number_quantifier` building block for this shape, even though `"return"` is already listed in its `MULTI_TARGET_VERBS` table and the helper is already proven correct for the sibling "tap/untap up to N target creatures" pattern. Left in place, the leading "up to three " was silently absorbed as unrecognized filler by the downstream destination/target parser.

The *targeted* sibling shape ("return up to two **target** cards from your graveyard to your hand" — Seeds of Renewal, Restock) already works correctly today via a completely different code path (`strip_optional_target_prefix`, gated on the literal word "target"). Only the non-targeted, "target"-keyword-less shape is affected.

## Fix

1. **`oracle_ir/ast.rs`** — add `multi_target: Option<MultiTargetSpec>` to `TargetedImperativeAst::Return`, mirroring `Tap`/`Untap`.
2. **`oracle_effect/imperative.rs` + `oracle_effect/mod.rs`** — in both "return " dispatchers, call the existing `strip_any_number_quantifier` before any destination/target parsing runs, restricted to the genuinely **bounded** (`max: Some(_)`) result. The unrelated **unbounded** "any number of" shape (Grave Sifter class) already has its own working mechanism (`return_up_to` → `ReturnToZone.up_to`); filtering avoids double-handling it or disturbing that path. Both `Return { .. }` construction sites in each file now thread the captured spec.
3. **`oracle_effect/imperative.rs`** — `lower_targeted_action_ast`'s bare-Effect lowering ignores the new field (mirrors the `Fight` arm's own comment); a new interception arm in `lower_imperative_family_ast` threads `multi_target` onto the clause when present, mirroring the existing `Tap`/`Untap` arm immediately above it.

**No runtime/game-logic changes.** `Effect::Bounce`'s at-resolution `EffectZoneChoice` picker already resolves a bounded `ability.multi_target` correctly — this is already exercised by `game::effects::bounce::non_targeted_optional_bounce_still_prompts_at_resolution` and `targeted_up_to_two_graveyard_bounce_moves_chosen_creature` (Wrenn and Six / Macabre Waltz precedent). This PR only stops the parser from dropping the quantifier before it reaches that machinery.

## Scope

One confirmed card in the current pool (Ill-Gotten Gains) matches this exact non-targeted numeric shape. `Bound // Determined`'s "Return up to X cards from your graveyard to your hand, where X is the number of colors that creature was" uses the same non-targeted shape but is separately broken by a pre-existing "where X is `<description>`" CDA-quantity-binding bug (the resolved `Variable` ends up holding the literal description string) — out of scope here. `Stream of Thought`'s "shuffle up to four cards from your graveyard into your library" reproduces the same class of defect through an entirely separate recognizer (`parse_shuffle_ast`, a different AST family) — also out of scope, flagged in the issue as a follow-up.

## CR references

CR 601.2c ("up to N" targeting/selection cardinality), CR 115.1d (variable target/object counts), CR 400.6/CR 400.7 (zone-change origin/destination).

## Verification

- `return_up_to_n_cards_from_graveyard_binds_bounded_multi_target` — pins the `Fixed` arm: "return up to three cards from your graveyard to your hand" → `Effect::Bounce { selection: AtResolution, .. }` + `multi_target: Some(up_to(Fixed(3)))`.
- `return_that_many_cards_from_graveyard_binds_dynamic_multi_target` — pins the dynamic arm (`QuantityRef::EventContextAmount`), exercising the general combinator capability directly.
- `return_a_card_from_graveyard_has_no_multi_target` — negative: the ordinary singular case stays `multi_target: None`, unaffected.
- `return_any_number_of_cards_from_graveyard_still_uses_legacy_up_to_flag` — negative: the unrelated unbounded "any number of" shape still produces `ChangeZone { up_to: true, .. }` via its pre-existing mechanism, with no `MultiTargetSpec` — proves the new bounded-only filter doesn't regress it.
- `ill_gotten_gains_return_clause_binds_bounded_multi_target` — full-card integration test against the real, Scryfall-verified Oracle text, walking the `Exile → Discard → Bounce` sub-ability chain and asserting the final clause's `multi_target`.

## Implementation method

Not `/engine-implementer`. This sandbox has no working Rust toolchain (no MSVC Build Tools, no WSL, Tilt not running), so I could not compile or run tests locally — the fix was verified by careful static tracing against the existing `Tap`/`Untap` precedent and the already-tested `EffectZoneChoice` runtime seam, cross-checked against real parser output in `data/card-data.json` and the Scryfall API. CI is relied upon for compilation/test verification; flagging this explicitly per repo convention.

### DIFF
--- a/crates/engine/src/game/effects/change_zone.rs
+++ b/crates/engine/src/game/effects/change_zone.rs
@@ -1770,7 +1770,8 @@ mod tests {
     use crate::game::zones::create_object;
     use crate::types::ability::{
         ControllerRef, FilterProp, MultiTargetSpec, PlayerFilter, PtValue, QuantityExpr,
-        QuantityRef, StaticDefinition, TargetFilter, TargetRef, TypeFilter, TypedFilter,
+        QuantityRef, StaticDefinition, TargetChoiceTiming, TargetFilter, TargetRef, TypeFilter,
+        TypedFilter,
     };
     use crate::types::actions::GameAction;
     use crate::types::card_type::CoreType;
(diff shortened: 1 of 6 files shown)


matthewevans (changes requested):
**Blocked — the bounded non-targeted return path still lowers through the legacy `ChangeZone` fallback.**

## 🔴 Blocker

- Current-head CI fails both new parser tests. `return_up_to_n_cards_from_graveyard_binds_bounded_multi_target` expected the bounded at-resolution `Bounce` shape but received a plain graveyard-to-hand `ChangeZone`; the "any number" regression test likewise lost its expected `up_to` behavior. The failing assertions are at `crates/engine/src/parser/oracle_effect/imperative.rs:19865` and `:19956` in run `29308248407`.
- This means the quantifier threading is not intercepting the production parser path for these clauses, so the claimed selection bound is still absent. The red Rust-test shards are therefore a functional failure, not a CI-only hold.

## 🟡 Non-blocking

- The intended reuse of the existing `MultiTargetSpec` machinery remains the right direction once the real return-dispatch/lowering path is identified.

**Recommendation: trace the actual fallback that emits this `ChangeZone`, route the bounded and unbounded return forms through their established selection mechanisms there, and keep both regression tests green before resubmitting.**

matthewevans (changes requested):
**Request changes — the parser shape is correct, but the newly reachable resolution path is not exercised.**

## 🔴 Blocker

- The new tests stop at `AbilityDefinition.multi_target` (`imperative.rs:19883-20018`; `oracle_tests.rs:21260-21326`). None runs the parsed `ChangeZone` clause through `EffectZoneChoice`, chooses a bounded subset, and verifies only those cards move. That leaves the essential contract unproven: `multi_target` must survive parsed-chain construction as `TargetChoiceTiming::Resolution`, `resolution_choice_cardinality` must expose `0..N`, and the chooser continuation must move exactly the chosen cards. The existing `targeted_up_to_two_graveyard_bounce_moves_chosen_creature` test injects a resolved target directly, so it bypasses that at-resolution picker.
- This is a concrete recurring gap for this contributor: the scanner reports `runtime-test-gap` in four PRs in the recent window. The two parser tests and the full-Oracle chain test earn their structural coverage, but at least one should be paired with a real-pipeline fixture for Ill-Gotten Gains (or the two-card sibling Unseal the Necropolis) containing more eligible graveyard cards than the bound.

## ✅ Clean

- The prior wrong-dispatch issue is fixed: the bounded quantifier is stripped before destination parsing, carried on the clause, and the current parse-diff is constrained to Ill-Gotten Gains (`0-3`) and Unseal the Necropolis (`0-2`). The shared `strip_leading_quantifier` helper is at the correct parser seam; all current GitHub checks are green.

**Recommendation: add one integration/scenario test that parses the real card text, reaches the `EffectZoneChoice`, submits fewer than the available eligible cards within the bound, and asserts the selected cards moved while the unselected card remains in its graveyard.**

matthewevans (approved):
**Approved — bounded non-targeted graveyard returns are now exercised through the resolution-time picker.**

## ✅ Clean

- `strip_leading_quantifier` carries the bounded count through the correct parser/lowering seam, and the parse-diff remains limited to Ill-Gotten Gains (`0-3`) and Unseal the Necropolis (`0-2`).
- `ill_gotten_gains_return_clause_resolves_partial_choice_through_effect_zone_choice` parses the actual card text, reaches `EffectZoneChoice`, submits two of four eligible cards under the `up to three` bound, and proves only those two move. Reverting the count threading breaks the asserted timing/cardinality and selection behavior.

**Recommendation: enqueue as the `bug` fix for the dropped bounded-return count.**

PR #5787, opened 2026-07-14
fix(parser): parse Total War's continuity exemption on the destroy filter

jaytbarimbao-collab:
Tier: Frontier
Model: claude-opus-4-8

## Summary

Total War — *"Whenever a player attacks with one or more creatures, destroy all untapped non-Wall creatures that player controls that didn't attack, **except for creatures the player hasn't controlled continuously since the beginning of the turn**."* — silently dropped its trailing continuity exemption. The `DestroyAll` filter parsed only as `[Untapped, Not(AttackedThisTurn)]`, so Total War also destroyed creatures the attacking player **gained this turn**, which CR 302.6 + CR 508.1a exempt.

This was a **documented KNOWN GAP** (the comment at `oracle.rs`, above `parse_continuity_exemption_clause`): the existing continuity recognizer covers only Siren's Call's `"ignore this effect for each creature … didn't control continuously …"` ActivePlayerPunisher shape. Total War phrases the same exemption as a trailing `"except for creatures the player hasn't controlled continuously …"` on the target population, which never reached that recognizer.

The player-scope half of this card (`"that player"` binding to the attacking player) was fixed separately in #5646; this completes Total War's parse.

## Implementation method

Additive, single seam, reuses existing machinery — no new engine variant:

- Add `parse_except_continuity_exemption_suffix` on the target-filter parse path (`oracle_target.rs`), a sibling of the existing `"except for <type-list>"` exclusion. It is placed **after** the controller / `"didn't attack"` relative clauses because the exemption trails them (the early type-list `except for` block runs before those clauses are consumed and never reaches it).
- The exempted set — creatures **not** controlled continuously — is removed from the population, so only creatures the player **has** controlled continuously since the turn began are affected. It attaches the existing `FilterProp::ControlledContinuouslySinceTurnBegan` — the same restriction Siren's Call reaches by its own ActivePlayerPunisher path (runtime: `filter.rs` evaluates it as `!obj.summoning_sick`, CR 508.1a).
- Updated the now-stale KNOWN GAP comment in `oracle.rs` to point at the new recognizer.

## CR references

- **CR 302.6 / CR 508.1a** — a creature the player hasn't controlled continuously since their most recent turn began (i.e. summoning-sick) is exempt.

## Verification

- **Discriminating runtime test** (`total_war_spares_creatures_not_controlled_since_turn_began`): the attacking player controls one idle creature since the turn began (destroyed) and one gained this turn (summoning-sick → **spared**). Reverting the recognizer flips the "Fresh Idler survives" assertion (the filter loses the restriction and destroys it too).
- The existing scope test (`total_war_destroys_attacking_players_idle_creatures_only`, #5646) still passes.
- Parse footprint: **Total War alone** — the sole card with this exemption phrasing (verified against the pool), so the coverage-parse-diff is a clean one-card improvement.
- Parser lib suite: **8101 passed, 0 failed**. `cargo fmt --check` clean; parser-combinator gate passes.

### DIFF
--- a/crates/engine/src/parser/oracle.rs
+++ b/crates/engine/src/parser/oracle.rs
@@ -2920,28 +2920,18 @@ fn sub_ability_is_continuity_exemption(sub: Option<&AbilityDefinition>) -> bool 
     parse_continuity_exemption_clause(text.trim()).is_ok_and(|(rest, ())| rest.trim().is_empty())
 }
 
-// KNOWN GAP (Total War, tracked as a separate follow-up): this recognizer
-// covers only Siren's Call's "ignore this effect for each creature ... didn't
-// control continuously since the beginning of the turn" phrasing. Total War
-// carries the SAME CR 302.6 + CR 508.1a continuity exemption, but through a
-// DIFFERENT shape:
-//   - it is a triggered ability ("Whenever a player attacks with one or more
-//     creatures ..."), NOT an `ActivePlayerPunisher` delayed trigger, so it
-//     never reaches `apply_active_player_punisher`, the only caller of
-//     `sub_ability_is_continuity_exemption` above; and
-//   - it phrases the exemption as "except for creatures the player hasn't
-//     controlled continuously ...", which the "ignore this effect for each
-//     creature ..." tag below does not match.
-// As a result Total War's continuity exemption is currently silently dropped —
-// its DestroyAll filter parses only as [Untapped, Not(AttackedThisTurn)] with
-// no `ControlledContinuouslySinceTurnBegan` restriction and no `Unimplemented`
-// marker. This is a distinct dropped-modifier root cause (NOT the player-scope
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
**Approved — Total War's trailing continuity exemption is now parsed at the target-filter authority.**

## ✅ Clean

- Scryfall's Oracle text says, "except for creatures the player hasn't controlled continuously since the beginning of the turn"; the new suffix attaches the existing `ControlledContinuouslySinceTurnBegan` predicate after the controller and combat clauses at `crates/engine/src/parser/oracle_target.rs:2967`.
- The registered scenario regression drives the actual attack-trigger stack path and distinguishes an established idle creature (destroyed) from a summoning-sick idle creature (spared) at `crates/engine/tests/integration/total_war_attacking_player_scope.rs:175`.
- The refreshed parse-diff reports exactly one changed card and signature: Total War's `DestroyAll` filter.

PR #5790, opened 2026-07-14
test(engine): use Sun Titan's real 'enters or attacks' Oracle text and pin the attack branch

### LINKED ISSUE #5674: test: SUN_TITAN_ORACLE const is not Sun Titan's Oracle text — 'or attacks' half is missing, so the attack branch is unpinned
## Summary

`SUN_TITAN_ORACLE` in `crates/engine/tests/integration/issue_3309_rise_etb_returns.rs:20-21` is **not Sun Titan's Oracle text.** It is an ETB-only fabrication with the "or attacks" half missing, so our only runtime "Sun Titan" test parses a card that does not exist.

## The defect

What the test feeds the parser:

```rust
const SUN_TITAN_ORACLE: &str =
    "Vigilance\nWhen Sun Titan enters, you may return target permanent card with mana value 3 or less from your graveyard to the battlefield.";
```

The actual card (verified against Scryfall, `api.scryfall.com/cards/named?exact=Sun+Titan`):

> Vigilance
> Whenever this creature enters **or attacks**, you may return target permanent card with mana value 3 or less from your graveyard to the battlefield.

Two divergences: the trigger is `Whenever this creature enters **or attacks**` (not `When Sun Titan enters`), and the self-reference is the modern `this creature` form.

## Consequence

`rise_of_dark_realms_optional_etb_return_with_observers_resolves` is a genuinely good runtime test — it drives the parse, fires the trigger, hits `TriggerTargetSelection`, and asserts the graveyard→battlefield return. But because the text it parses has no "or attacks" clause, it produces a plain ETB trigger, **not** the `TriggerMode::EntersOrAttacks` the real card produces.

So the attack half of Sun Titan's trigger — and more generally an `EntersOrAttacks` attack branch feeding a graveyard-scoped target slot — is **unpinned end-to-end**, while a test named for the card gives the appearance of covering it.

The card itself parses correctly on `main` today (`EntersOrAttacks`, `ChangeZone` Graveyard→Battlefield, `Permanent` + `Cmc ≤ 3` + `InZone Graveyard` + `controller: You`), so this is a **coverage** defect, not a behavior defect. Nothing is broken; nothing is guarding it either.

## Fix

Replace the const with the verbatim Scryfall text and drive it through **`parse_oracle_text`** — the entry point `database::synthesis` uses to build `card-data.json` — asserting both halves fire with the graveyard target. Driving `parse_oracle_text` rather than a standalone entry point is what makes the test unable to go green while the shipped card is broken.

## Why this is worth fixing beyond the one const

A test named for a card, parsing text the card does not have, is the exact shape that lets a fix be "verified" against a premise nobody checked. CLAUDE.md's **"verify the card, not just the rule"** exists for this. Worth a sweep for other hand-written oracle consts in `tests/integration/` that have drifted from their printed text.

## Related

- #5264 — Sun Titan "triggers don't trigger" (no engine defect found; see #5671)
- #5671 — the test-only PR that surfaced this

nghetienhiep:
SUN_TITAN_ORACLE was a fabricated ETB-only const missing the 'or attacks'
half, so the only runtime Sun Titan test parsed a card that doesn't exist and
the EntersOrAttacks attack branch feeding a graveyard-scoped return slot was
unpinned end-to-end.

Replace the const with the verbatim Scryfall text driven through
parse_oracle_text, assert it parses as TriggerMode::EntersOrAttacks (not a
plain ETB) with the optional graveyard->battlefield return, and add a runtime
test that attacks with Sun Titan and confirms the attack trigger returns a
graveyard permanent. Reverting the const to the ETB-only text fails both.

Closes #5674

### DIFF
--- a/crates/engine/tests/integration/issue_3309_rise_etb_returns.rs
+++ b/crates/engine/tests/integration/issue_3309_rise_etb_returns.rs
@@ -6,6 +6,7 @@
 //! ETB triggers including an optional graveyard-return (Sun Titan class) and
 //! observer triggers (Soul Warden class).
 
+use engine::game::combat::AttackTarget;
 use engine::game::scenario::{GameScenario, P0, P1};
 use engine::types::ability::TargetRef;
 use engine::types::actions::GameAction;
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
**Approved — the test replaces fabricated Sun Titan text with its authoritative Oracle text and exercises the previously unpinned attack branch.**

## ✅ Clean

- The exact Oracle fixture now parses as `EntersOrAttacks`, rather than silently testing a non-existent ETB-only card.
- The attack scenario drives trigger targeting, the optional resolution choice, and the graveyard-to-battlefield move; restoring the old fixture removes that production trigger path and makes the test fail.
- Existing integration coverage already uses the same scenario conventions, and this test corrects the only Sun Titan fixture rather than duplicating a separate mechanic test.

**Recommendation: approved for the merge queue.**

PR #5800, opened 2026-07-14
fix(coverage): project ReplacementDefinition scoping into the parse-diff signature

### LINKED ISSUE #5673: coverage: parse-diff is blind to ReplacementDefinition scoping (valid_card never projected) — shield-scoping fixes show a false zero diff
## Summary

The parse-diff / coverage projection **cannot see a `ReplacementDefinition`'s scoping**. Any PR that fixes *whom* a replacement or prevention shield applies to shows **"✓ No card-parse changes detected"** — a false negative that reads exactly like "your fix is inert."

Surfaced while reviewing #5669 (self-scoped damage shields, Swans of Bryn Argoll / #5652), whose sticky reported zero parse changes while a runtime bug-pin test on `main` simultaneously went red proving the behavior *did* change.

## The gap

`crates/engine/src/game/coverage.rs:4327-4353` — the replacement projection:

```rust
for repl in &face.replacements {
    // ... children built from repl.execute / decline ...
    items.push(ParsedItem {
        category: ParseCategory::Replacement,
        label: format!("{}", repl.event),   // event only: "DealtDamage"
        source_text: repl.description.clone(),
        supported: execute_supported,
        details: vec![],                     // <-- hardcoded EMPTY
        children,
    });
}
```

`details` is hardcoded `vec![]`, and **`repl.valid_card` is never projected**. So two replacements that differ *only* in scope — `valid_card: None` (fires for damage the object DEALS as well as damage dealt TO it) versus `valid_card: Some(TargetFilter::SelfRef)` (correctly self-scoped) — produce a **byte-identical signature**.

Compare the sibling projections, which do not have this hole:

- **Triggers** project it — `coverage.rs:3597`: `if let Some(vc) = &trig.valid_card { ... }`
- **Statics** project their detail — `details: static_details(stat)`

Only replacements drop it.

## Why it matters

This is a **false negative in a gate we treat as proof**. On #5669 the sticky's "No card-parse changes detected" was the *only* parse-level evidence available, and taken at face value it says the fix does nothing. The fix in fact repairs a whole class (Swans, Phytohydra, Lichenthrope-style shields) — shields that were preventing the source's *own* outgoing damage.

We have been here before: #5634 looked inert for the same structural reason (`build_ability_item` did not recurse into `FlipCoin` win/lose branches), and a correct PR was briefly blocked on it. **Projection is per-position and per-field** — confirming a field is projected somewhere does not mean it is projected everywhere a card can put it. This is the same bug one field over.

## Fix

Project the replacement's scoping into `details`, mirroring what triggers already do:

- `repl.valid_card` (the recipient scope — the field #5669 changes)
- `repl.prevention_shield` / `PreventionAmount`
- the `ReplacementMode` discriminant

A regression test for the projection itself: two `ReplacementDefinition`s differing only in `valid_card` must produce **different** `ParsedItem` signatures. Today they do not.

## Related

- #5669 — the PR that exposed it (blocked on other grounds; its zero sticky is *not* a defect)
- #5652 — the underlying self-scoped-shield bug
- #5634 — the same class of projection blindness, in `FlipCoin` branches

nghetienhiep:
The replacement projection in `build_parse_details` (`coverage.rs`) hardcoded `details: vec![]` and never projected `repl.valid_card`. So two `ReplacementDefinition`s differing only in scope — `valid_card: None` (fires for damage the object *deals* as well as damage dealt *to* it) versus `Some(TargetFilter::SelfRef)` (correctly self-scoped) — produced a **byte-identical** signature.

Any PR that fixes *whom* a replacement or prevention shield applies to therefore showed a false **"✓ No card-parse changes detected"** sticky — the false signature of an inert fix. This is the same class as the trigger/static/`FlipCoin` projection holes (#5492, #5495, #5501, #5507, #5585, #5634): projection is per-position and per-field, and the replacement position dropped it.

## Fix

Add `replacement_details()`, mirroring `trigger_details` / `static_details`, and wire it into the replacement `ParsedItem` in place of `vec![]`. It projects the discriminating scope/modification fields:

- `valid_card` (recipient scope — the field the shield-scoping fixes change) via the existing `fmt_target`
- `valid_player`, the `ReplacementMode` discriminant, and `shield_kind` (including `PreventionAmount`)
- `damage_source_filter` / `damage_target_filter` / `combat_scope`, `damage_modification` / `quantity_modification`, `destination_zone`, `condition`, `redirect_target`, `draw_scope`

Each field is emitted only when set, so an unqualified replacement keeps a byte-identical signature — only a genuine scoping/modification change moves it.

## Test

`replacement_signature_exposes_valid_card_scope` drives the production `build_parse_details_for_face` path: two replacements differing only in `valid_card` (`None` vs `Some(SelfRef)`) must now produce **different** `ParsedItem` signatures, the set scope must appear as a `scope` detail row, and the absent one must not (so unqualified signatures stay stable). Reverting the projection change fails it.

## Verification

Tilt was down locally, so ran cargo directly:

- `cargo test -p engine --lib game::coverage::` — **103 passed, 0 failed** (includes the new test)
- `cargo clippy -p engine --lib` — clean (`-D warnings`)
- `cargo fmt --all` — clean

No support-status or item counts change (details is purely additive), so coverage totals and the draw-replacement census are unaffected. Expected side effect: the parse-diff sticky on this PR will now show a `scope`/scope-field row added to every replacement card — that is the projection landing, not a regression.

Closes #5673

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -4274,6 +4274,15 @@ fn static_details(stat: &StaticDefinition) -> Vec<(String, String)> {
 /// distinct `ParsedItem`s. Without this, a scope-only fix (e.g. a self-scoped
 /// damage shield flipping `valid_card` from `None` to `Some(SelfRef)`) shows a
 /// false "no card-parse changes detected" diff.
+///
+/// Covers every parse-time semantic axis on the struct: recipient/player
+/// scope, mode, shield kind, damage source/target/combat/modification,
+/// quantity modification, destination zone, condition, redirect target, draw
+/// scope, token owner scope/redirect, mana modification/scope, additional and
+/// ensure-all token specs, counter match, and controller overrides
+/// (`enters_under`) and expiry. Deliberately excludes runtime-only state
+/// (`execute`, `runtime_execute`, `consume_on_apply`, `is_consumed`,
+/// `source_controller`) which carries no parse-time signal of its own.
 fn replacement_details(repl: &ReplacementDefinition) -> Vec<(String, String)> {
     let mut d = Vec::new();
     // Recipient scope — the field a shield-scoping fix changes.
@@ -4319,6 +4328,36 @@ fn replacement_details(repl: &ReplacementDefinition) -> Vec<(String, String)> {
     }
     if let Some(scope) = &repl.draw_scope {
         d.push(("draw scope".into(), format!("{scope:?}")));
+    }
+    if let Some(scope) = &repl.token_owner_scope {
+        d.push(("token owner scope".into(), format!("{scope:?}")));
+    }
+    if let Some(redirect) = &repl.token_owner_redirect {
+        d.push(("token owner redirect".into(), format!("{redirect:?}")));
+    }
+    if let Some(mm) = &repl.mana_modification {
+        d.push(("mana mod".into(), format!("{mm:?}")));
+    }
+    if !repl.mana_replacement_scope.is_any() {
+        d.push((
+            "mana scope".into(),
+            format!("{:?}", repl.mana_replacement_scope),
+        ));
+    }
+    if let Some(spec) = &repl.additional_token_spec {
+        d.push(("additional token".into(), format!("{spec:?}")));
+    }
+    if let Some(specs) = &repl.ensure_token_specs {
+        d.push(("ensure tokens".into(), format!("{specs:?}")));
+    }
+    if let Some(cm) = &repl.counter_match {
+        d.push(("counter match".into(), format!("{cm:?}")));
+    }
+    if let Some(cref) = &repl.enters_under {
+        d.push(("enters under".into(), format!("{cref:?}")));
+    }
+    if let Some(expiry) = &repl.expiry {
+        d.push(("expiry".into(), format!("{expiry:?}")));
     }
     d
 }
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**Request changes — `replacement_details` still hides several parsed replacement semantics, so the false-green class remains.**

## 🔴 Blocker

- The new helper promises to project replacement scoping and modification fields, but it omits semantic parse-time axes such as `token_owner_scope`, `token_owner_redirect`, `mana_modification`, `mana_replacement_scope`, `additional_token_spec`, `ensure_token_specs`, `counter_match`, `enters_under`, and `expiry` (`coverage.rs:4277-4323`; definition fields begin at `ability.rs:18927`). A parser fix to any of those fields will still produce an unchanged parse signature. The sole regression covers `valid_card`, so it cannot catch those omissions.

## ✅ Clean

- The `valid_card` regression is a useful production-level check, and the existing projection is correctly wired into every top-level replacement item.

**Recommendation: make the projection exhaustive over parse-time semantic fields (explicitly excluding only runtime state such as resolved continuations), then add at least one regression for an omitted axis such as `token_owner_scope` or `enters_under` so future scope-only changes cannot disappear again.**

matthewevans (approved):
**Approved — replacement parse signatures now include every parse-time semantic axis while excluding runtime-only state.**

## ✅ Clean

- `replacement_details` is wired at the existing `build_parse_details` replacement seam and projects the formerly missing scope, token, mana, counter, controller-override, and expiry fields (`crates/engine/src/game/coverage.rs:4286-4362`).
- The regression tests use `build_parse_details_for_face`, proving a scope-only replacement change alters the production parse signature and that absent optional fields remain absent (`crates/engine/src/game/coverage.rs:10677-10788`).
- The current parse-diff artifact reports the expected replacement detail additions across 2,605 affected cards, and current CI is green.

**Recommendation: merge as the `bug` fix for the replacement parse-diff false-green class.**

PR #3525, opened 2026-06-16
feat(parser): opponent-scoped draw replacements (except-first-draw class)

kiannidev:
## Summary
- Widen the draw-replacement gate from controller-only `"you would draw"` to any `would draw a card` / `would draw one or more cards` phrasing
- Apply `valid_player` scope for opponent and global-player antecedents
- Preserve the shared `ExceptFirstDrawInDrawStep` condition for Notion Thief / Hullbreacher / Chains class text

Closes #3524

## Impact
Unlocks ~9+ cards sharing one antecedent primitive (Notion Thief, Hullbreacher, Chains of Mephistopheles, Magus of the Chains, etc.).

## Test plan
- [x] `cargo test -p engine --lib opponent_draw_replacement`
- [x] `cargo test -p engine --lib parses_any_player_draw`

Made with [Cursor](https://cursor.com)

PR #5804, opened 2026-07-14
fix(parser): parse compound draw-substitute "skips that draw and you draw" (#5654)

### LINKED ISSUE #5654: Notion Thief + Plagiarize: compound 'skips that draw and you draw' substitute drops to Unimplemented (one fix, two cards)
## Summary
From Plan 03 P3-1b harvest (pool-verified). Both cards' replacement substitute is the identical compound: **'instead that player skips that draw and you draw a card'** → `Unimplemented`.

## The class evidence
**Hullbreacher** has the IDENTICAL antecedent but a single-action substitute ('instead you create a Treasure token') → parses correctly to Token. The gap is precisely the COMPOUND substitute: skip + subject-switched draw. One recognizer fix covers both cards.

## Plan 03 relevance
Notion Thief is in the 51-card frozen Draw corpus with execute=Unimplemented — the corpus gate (Gate C / card-data job) will go RED when this is implemented, which is intended: the newly-implemented execute must be assigned a reviewed DrawReplacementScope, then re-freeze with --corpus --write.

jaytbarimbao-collab:
Tier: Frontier
Model: claude-opus-4-8

Fixes #5654

## Summary

Notion Thief and Plagiarize's replacement substitute — *"instead that player skips that draw and you draw a card"* — dropped to `Unimplemented`, so the defining effect of both cards silently did nothing. **Hullbreacher** has the identical antecedent with a single-action substitute (*"instead you create a Treasure token"*) and parses fine; the gap was precisely the **compound** substitute (skip + subject-switched draw). One recognizer covers both cards.

## Implementation method

Two coordinated recognizers on the draw-replacement path (`oracle_replacement.rs`), reusing existing machinery — no new engine variant:

1. **`strip_leading_draw_skip_conjunct`** — for a Draw replacement whose substitute is `"[instead] [subject] skip[s] that/the draw and <effect>"`, drop the `"skips that draw and"` conjunct. The skip **is** the replacement (the original draw is replaced), so only `<effect>` (`"you draw a card"`) is the substitute to run — it now parses to a controller `Draw` execute. A pure skip with no conjunct (`"... skips that draw"`, Living Conundrum) is untouched and still reaches `body_is_draw_skip → Prevent`.

2. **`extract_replacement_effect` now prefers splitting at `", instead "`** over the first `", "`. Plagiarize's *"Until end of turn,"* has an earlier comma, so the first-comma split leaked the antecedent (`"if target player would draw a card, instead …"`) into the effect. Splitting at the replacement marker fixes it; marker-less and trailing-`"instead"` forms fall back to the first `", "` unchanged.

## CR references

- **CR 614.6 / CR 121.6** — the replacement replaces the original draw (the "skip"); **CR 614.1a** — the substitute follows the "instead" marker.

## Verification

- **Discriminating parser test** (`compound_draw_skip_and_draw_substitute_parses_controller_draw`): both cards parse to `ReplacementEvent::Draw` with a `Draw { target: Controller }` execute (not `Unimplemented`), scope `IndividualDraw`. Fails on `main` (Notion Thief → `Unimplemented`; Plagiarize → `Unimplemented` + leaked antecedent).
- Re-froze `scripts/draw-replacement-corpus.tsv` per the census workflow — **2 rows** (notion thief / plagiarize), execute column `Unimplemented → Draw`, scope unchanged (`IndividualDraw`), so `draw_replacement_census.py --corpus --check` passes; `--producers --check` unchanged.
- Parser lib suite: **8138 passed, 0 failed** (the shared `extract_replacement_effect` change caused no regressions across replacement types). `cargo fmt --check` clean; parser-combinator gate passes.


matthewevans (changes requested):
**Request changes — the new supported draw replacements have parser-shape coverage only, not replacement-pipeline coverage.**

## 🔴 Blocker

- The regression asserts only `parse_replacement_line` and the nested `Effect::Draw` AST (`crates/engine/src/parser/oracle_replacement.rs:16938-16968`), while the change upgrades Notion Thief and Plagiarize from `Unimplemented` to supported in the corpus. No test applies an opponent’s draw through the individual-draw replacement drain, proves the original draw is skipped, and proves the controller draws exactly once. That leaves the claimed game behavior unverified and could make coverage green for an inert nested replacement.

## ✅ Clean

- The parser change is narrow, uses the existing draw-replacement shape, and the current artifact is exactly Notion Thief and Plagiarize (2 cards, 4 signatures).

**Recommendation: add registered `GameScenario` coverage for both replacement paths—opponent draw skipped plus controller draw—then re-review the current head.**

jaytbarimbao-collab:
Closing this — on reflection it overclaims. Two issues:

1. **Plagiarize is not cleanly supportable.** Its replacement is scoped to the *cast target* ("if **target player** would draw"), but `ReplacementPlayerScope` only has {You, Opponent, AnyPlayer} — there is no target-bound scope. AnyPlayer would wrongly replace every player's draw (including the caster's), so claiming support for it is incorrect without an engine-level scope addition.

2. **Notion Thief** *is* supportable (the opponent-scoped + except-first-draw runtime from #3525 already exists), but this PR only has parser-shape coverage, not the replacement-pipeline coverage you rightly asked for, and the branch has gone stale.

I'll revisit Notion Thief on its own in a fresh PR with a proper runtime test (add_real_card + rehydrate driving an opponent draw) if #5654 stays open. Thanks for the review.

PR #5805, opened 2026-07-14
fix(parser): map standalone + combat-damage activation-timing gates to enforced variants

lourincedaging0-commits:
A combat-window activation gate with no `during <role>` first half fell through to `Effect::Unimplemented`, dropping the ability's timing restriction entirely. This binds every such phrasing to an **existing** enforced `ActivationRestriction` via the shared `parse_activation_before_window_gate` combinator — no new enum variant, no new helper, pure building-block reuse:

| Oracle phrasing | Restriction | CR | Cards |
|---|---|---|---|
| "before attackers are declared" / "before combat" | `BeforeAttackersDeclared` | 508.1 | Arcum's Whistle, Arcum's Sleigh |
| "before combat damage has been dealt" / "before the combat damage step" / "before combat damage" | `BeforeCombatDamage` | 510 | Angus Mackenzie, Save Point |

### What changed
- The before-window combinator tries the **combat-damage arm before** the `before combat` arm (which is a prefix of `before combat damage` and would otherwise shadow it), longest phrasing first so it consumes fully.
- A new `during combat <window>` branch in `parse_activation_timing_restriction` pairs `DuringCombat` with the before-window gate (CR 509.1 + CR 510) → Save Point yields `[DuringCombat, BeforeCombatDamage]`.
- **Two verbatim-string `strip_suffix` hacks are deleted** ("during your turn, before attackers are declared" and "during combat before combat damage has been dealt") — both are now subsumed by the compound grammar, which the `activate only ` routing arm reaches before those strip points.

Each branch is guarded to fire only when the whole phrase is consumed; the `during your turn` and existing compound forms are untouched.

### Tests
Building-block coverage that the standalone and combat-damage phrasings map to their respective enforced variants. Full engine suite green locally: **16526 passed, 0 failed**.

### DIFF
--- a/crates/engine/src/game/restrictions.rs
+++ b/crates/engine/src/game/restrictions.rs
@@ -4476,4 +4476,67 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn parsed_combat_window_activation_gates_reach_production_enforcement() {
+        let check_window = |oracle: &str,
+                            expected: &[ActivationRestriction],
+                            legal_phase: Phase,
+                            illegal_phase: Phase| {
+            let parsed = crate::parser::oracle::parse_oracle_text(
+                oracle,
+                "Combat Window Test",
+                &[],
+                &["Artifact".to_string()],
+                &[],
+            );
+            let restrictions = &parsed.abilities[0].activation_restrictions;
+            assert_eq!(
+                restrictions, expected,
+                "parser must expose the enforced gate"
+            );
+
+            let mut state = crate::types::game_state::GameState::new_two_player(42);
+            let player = PlayerId(0);
+            state.active_player = player;
+            state.priority_player = player;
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
**Parser fix is complete and verified through the production activation gate.**

## 🔴 Blocker
None.

## 🟡 Non-blocking
None.

## ✅ Clean
- The parser reuses the existing typed `ActivationRestriction` variants and handles the compound timing form without a one-off card branch.
- The added regression parses standalone and compound Oracle text, then drives those restrictions through `check_activation_restrictions`; illegal combat phases reject activation, so a parser regression is observable at the production boundary.
- Head `b3d689bea0305703bc257c85ffd112601b8fa339` is clean with required CI green; the parse diff contains the three expected activation-timing changes and the independent implementation review found no remaining actionable issue.

**Approve and enqueue.**

PR #5807, opened 2026-07-14
fix(parser): compose a color-defining static's color with trailing modifications (Onakke Catacomb)

### LINKED ISSUE #5806: parser: color-defining static drops its color when composed with a keyword/pump (Onakke Catacomb)
## Summary

A continuous color-defining static of the form `<subject> are/is [color] and <keyword|pump>` **silently drops the color**, keeping only the keyword/pump.

- **Onakke Catacomb**: `All creatures are black and have deathtouch.`
  Verified against `client/public/card-data.json` on current upstream: the static parses to a single `AddKeyword { Deathtouch }` modification — the `SetColor { Black }` is gone, so every creature gains deathtouch but is **not** made black.

The color is only lost when it is *composed* with another modification: `All creatures are black.` (solo) parses correctly to `SetColor { Black }`.

## Root cause

`All creatures are black and have deathtouch` hits the `"all creatures "` fast path in `oracle_static/dispatch.rs`, which calls `parse_continuous_gets_has`. That handler matches the trailing `have deathtouch` and returns `AddKeyword { Deathtouch }`, claiming the whole line before the color path (`parse_all_subject_are_color`) is reached. The color functions themselves required the predicate to be a **pure** color expression (`parse_color_predicate(predicate)?`), so `black and have deathtouch` failed there and fell through to the keyword path.

The dispatch comment even asserts the ordering is safe because "`parse_continuous_gets_has`-driven branches … cleanly return None for `are black` predicates" — true for the solo color, but **false** for the `are black and have <keyword>` compound.

## Structural parallel

This is the color sibling of the anthem/type composition already handled elsewhere: `gets +N/+M and has flying` composes a pump with a keyword; `are black and have deathtouch` must compose a **color** (Layer 5, CR 613.1e) with a keyword (Layer 6, CR 613.1f) the same way.

## Proposed fix

Generalize the color-static predicate parse (`parse_all_subject_are_color` / `parse_subject_is_color` in `oracle_static/type_change.rs`):

- A shared `parse_color_and_trailing_modifications(predicate)` peels the leading color expression (itself possibly a multi-color `black and green` list) and parses the `and <keyword|pump>` tail via the existing `parse_continuous_modifications`, returning `(colors, extra_mods)`. A pure color predicate returns `(colors, [])` (unchanged); a non-color predicate returns `None` (routes elsewhere).
- The `"all creatures "` dispatch probes `parse_all_subject_are_color` before `parse_continuous_gets_has`, so a color-leading compound reaches the color path while a bare gets/has line (no leading color) still resolves there.

Result: `[SetColor { Black }, AddKeyword { Deathtouch }]`. Solo color statics and pure keyword/pump anthems are unchanged.

## CR

CR 105.2 (object color) + CR 613.1e (Layer 5, color-changing) + CR 613.1f (Layer 6, ability-adding) — a color-defining static applies the color and the granted ability simultaneously.

## Notes

No existing open issue covers this. #5798 (multi-zone Oxford *compound-subject* color, Painter's Servant/Mycosynth Lattice) is a different axis — that's about the color *subject*; this is about a color *predicate* composed with other modifications.

minion1227:
Closes #5806

## Summary

A continuous color-defining static of the form `<subject> are/is [color] and <keyword|pump>` silently **dropped the color**, keeping only the keyword/pump. Onakke Catacomb's `All creatures are black and have deathtouch` parsed to a lone `AddKeyword { Deathtouch }` — every creature gained deathtouch but was never made black.

Root cause: the `"all creatures "` fast path in `dispatch.rs` calls `parse_continuous_gets_has`, which matches the trailing `have deathtouch` and claims the whole line before the color path (`parse_all_subject_are_color`) is reached. The color functions required the predicate to be a **pure** color expression, so `black and have deathtouch` failed there and fell through to the keyword handler.

## Files changed

## CR references

- CR 105.2 — an object's color(s); a color-defining static sets them.
- CR 613.1e — Layer 5 (color-changing effects) — the `SetColor`.
- CR 613.1f — Layer 6 (ability-adding effects) — the granted keyword; applied simultaneously with the color.

Method: not-applicable — authored directly against the `oracle-parser` skill. Pure parser change; no card-data regeneration required (AST-shape fix).

## Track

Developer

## LLM

Model: claude-opus-4-8
Thinking: high

## Verification

- [x] Verification below is for the current committed head (`c498c2d32`).

Results on head `c498c2d32`:

- `cargo test -p engine --lib parser::` — **ok. 8143 passed; 0 failed** (1 new full-shape test; all snapshot tests unaffected).
- `cargo clippy -p engine --all-targets --features proptest -- -D warnings` — **Finished, 0 warnings** (exit 0).
- `cargo fmt --all` — clean.

### Parse diff (Onakke Catacomb, `All creatures are black and have deathtouch.`)

Before — the color is dropped:

```
modifications: [ AddKeyword { Deathtouch } ]
```

After:

```
modifications: [ SetColor { [Black] }, AddKeyword { Deathtouch } ]
```

## Anchored on

- `crates/engine/src/parser/oracle_static/type_change.rs` (`parse_all_subject_are_color` / `parse_subject_is_color`) — the existing color-static seam this extends. Both already emit `SetColor` for a pure color predicate; the new `parse_color_and_trailing_modifications` peels that same leading color expression and appends the `and <keyword|pump>` tail, reusing the already-in-file `parse_continuous_modifications` (called at type_change.rs for the "and has …" compound-type tail).
- `crates/engine/src/parser/oracle_static/dispatch.rs` (the `"all creatures "` branch) — mirrors the existing "probe the more specific handler first" ordering already used across this dispatch (e.g. `parse_all_permanents_are_type` before the color path); the color path is now probed before `parse_continuous_gets_has` so a color-leading compound is not claimed on its keyword tail.

## Final review

Self-review against the review-impl lenses. `parse_color_and_trailing_modifications` returns `(colors, [])` for a pure color predicate (unchanged behavior), splits at the first `" and <have|has|get|gets|gain|gains|can't>"` boundary (gated on a modification verb so a multi-color `"black and green"` list is never split mid-color), and returns `None` when the tail parses to zero modifications or the head is not a color — so non-color predicates (`"red dragons"`, `"artifacts in addition to their other types"`) route elsewhere untouched. The dispatch probe only claims color-leading `"all creatures are …"` lines; a bare `"all creatures get/have …"` leaves the color path declining and resolves through `parse_continuous_gets_has` exactly as before (regression-tested). Solo color statics and pure keyword/pump anthems are byte-for-byte unchanged (8143 parser tests, incl. snapshots, unaffected).

## Claimed parse impact

- Onakke Catacomb
- the general `<subject> are/is [color] and <keyword|pump>` color-defining static class

## Validation Failures

None.

## CI Failures

None.


matthewevans (approved):
**Approved — the color-plus-modification static now preserves both layers without widening the card surface.**

## ✅ Clean

- Current-head parse evidence reports exactly one changed card, Onakke Catacomb: `grant Deathtouch` → `set color Black, grant Deathtouch`.
- `parse_color_and_trailing_modifications` stays at the existing color-static authority and reuses the continuous-modification parser; the direct production-parser tests cover both the `all creatures` fast path and the generic subject sibling.
- The current head is based on `main` `a27699bd9ee6` with all required CI checks green.

**Recommendation: merge this `bug` fix through the queue.**

PR #5809, opened 2026-07-14
fix(engine): run the base chain when a teamwork-instead swap doesn't fire

### LINKED ISSUE #4772: Too Evil To Stay Dead Nonfunctioning — I played Too Evil To Stay Dead (2B Sorcery, choose target creature in Graveyard…
Reported in Discord: https://discord.com/channels/1485498006781427802/1521707007189717002/1521707007189717002

**Thread:** Too Evil To Stay Dead Nonfunctioning
**Cards:** too evil to stay dead

## Summary
I played Too Evil To Stay Dead (2B Sorcery, choose target creature in Graveyard of CMC<=4, return to battlefield)

It didn't do anything, it just spen

### too evil to stay dead
Teamwork 4 (As an additional cost to cast this spell, you may tap any number of creatures you control with total power 4 or more.)
Choose target creature card in your graveyard with mana value 4 or less. If this spell was cast using teamwork, instead choose target creature card in your graveyard. Return the chosen card to the battlefield.

---
<sub>report_id: `discord:1521707007189717002:1521707007189717002:0` · discord: `1521707007189717002/1521707007189717002`</sub>

tryeverything24:
## Summary

Fixes #4772. Too Evil to Stay Dead ("Choose target creature card in your graveyard with
mana value 4 or less. If this spell was cast using teamwork, instead choose target creature
card in your graveyard. Return the chosen card to the battlefield.") silently did nothing
when cast **without** paying the optional Teamwork additional cost — mana was spent, the
spell resolved, but the targeted creature never returned to the battlefield.

## Root cause

The not-swapped path for three of the four "instead" condition kinds
(`AdditionalCostPaidInstead`, `CastVariantPaidInstead`, `TargetHasKeywordInstead`) only ever
checked `sub.else_ability`, returning early when it was `None` — silently dropping
`sub.sub_ability` (here, the trailing `ChangeZone` that returns the card to the battlefield)
entirely. The sibling `ConditionInstead` arm already had the correct
`sub.sub_ability` tail-runner fallback (added for Throw from the Saddle / Evil's Thrall /
Take the Fall / That's Rough Buddy) — this merges all four condition kinds onto that single,
already-tested path, since the swap-decision check just above already evaluates all four
uniformly.

## Anchored on

- `crates/engine/src/game/effects/mod.rs:7830-7921` (pre-fix) — the `ConditionInstead`
  not-swap tail-runner this change generalizes, with its own existing regression coverage
  (`condition_instead_not_swap_tail_runner_honors_gates`,
  `condition_instead_runs_base_chain_when_not_met`).
- `crates/engine/src/game/effects/mod.rs:6444-6473` — the swap-decision block, which already
  evaluates `AdditionalCostPaidInstead`/`CastVariantPaidInstead`/`TargetHasKeywordInstead`/
  `ConditionInstead` symmetrically, establishing that the not-swap path should be symmetric
  too.

## Testing

Added `crates/engine/tests/integration/issue_4772_too_evil_to_stay_dead.rs`:
- `too_evil_to_stay_dead_without_teamwork_still_reanimates` — the actual regression (fails
  on pre-fix code with the exact reported symptom).
- `too_evil_to_stay_dead_with_teamwork_still_reanimates` — differential control, confirms
  the already-working swap path doesn't regress.

Verified the new test is load-bearing by reverting just the `effects/mod.rs` change and
re-running — it fails with the exact issue #4772 symptom, then passes again with the fix
restored. Both tests pass on a clean, isolated `cargo test -p engine --test integration
issue_4772` run (not relying on any shared/incremental build cache).

```
running 2 tests
test issue_4772_too_evil_to_stay_dead::too_evil_to_stay_dead_without_teamwork_still_reanimates ... ok
test issue_4772_too_evil_to_stay_dead::too_evil_to_stay_dead_with_teamwork_still_reanimates ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 3051 filtered out
```

`cargo fmt --all` run, no diff beyond the change above.

## Validation Failures (honesty disclosure, per docs/AI-CONTRIBUTOR.md)

This was implemented and verified by a Claude Code agent (diagnosis via real code reading
+ empirical test-driven verification, minimal fix reusing an existing tested code path,
regression test proven load-bearing by revert/restore, `cargo fmt` clean), but **not**
through the formal `/engine-implementer` multi-agent plan → review-plan → implement →
review-impl pipeline — that skill wasn't available in the environment this was built in.
No `Gate A` combinator-purity script output or discriminating-test coverage map /
maintainer-simulation matrix is included. Flagging this plainly rather than claiming a
review loop that didn't run.

Model: claude-sonnet-5
Tier: Standard

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -8043,43 +8043,28 @@ fn resolve_chain_body(
         if let Some(ref condition) = sub.condition {
             // CR 608.2c: "Instead" overrides are terminal — the Cow swap above either
             // replaced the parent's effect (condition met) or didn't (condition not met).
-            // For kicker/ninjutsu/keyword-instead, the base has no continuation chain.
-            // For ConditionInstead, the base chain (else_ability) must run when NOT swapped.
+            // When NOT swapped, the base chain (else_ability) runs after the parent's own
+            // effect. This is shared by EVERY "instead" condition kind — additional-cost /
+            // cast-variant / target-keyword gated overrides, and the general
+            // `ConditionInstead` wrapper — because the swap check above (~line 6444)
+            // already treats all four uniformly; the not-swap continuation must too.
+            // Issue #4772 (Too Evil to Stay Dead): "Choose target creature card … with
+            // mana value 4 or less. If this spell was cast using teamwork, instead choose
+            // target creature card in your graveyard. Return the chosen card to the
+            // battlefield." lowers to a base `TargetOnly` clause whose sub is the
+            // teamwork-gated override (`AdditionalCostPaidInstead`), whose own sub is the
+            // trailing `ChangeZone` (`SequentialSibling`, no `else_ability`). Before this
+            // fix, only the `ConditionInstead` arm below had the `sub.sub_ability`
+            // tail-runner fallback, so when teamwork was NOT paid the swap correctly did
+            // not fire but the trailing `ChangeZone` was silently dropped along with the
+            // consumed override node — the spell targeted a card and did nothing.
             if matches!(
                 condition,
                 AbilityCondition::AdditionalCostPaidInstead
                     | AbilityCondition::CastVariantPaidInstead { .. }
                     | AbilityCondition::TargetHasKeywordInstead { .. }
+                    | AbilityCondition::ConditionInstead { .. }
             ) {
-                if let Some(ref base_chain) = sub.else_ability {
-                    let mut resolved = base_chain.as_ref().clone();
-                    if resolved.targets.is_empty() && !ability.targets.is_empty() {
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- `resolve_chain_body` is the shared non-swap authority for all four `instead` condition kinds; the rebased code retains main's continuation-context capture.
- The registered integration test drives both declined and paid Teamwork through casting, target selection, and stack resolution; the decline case fails if the new tail runner is removed.
- The current-head CI is green, Gemini reported no findings, and the parse-diff confirms no parser coverage change.

PR #5815, opened 2026-07-14
feat(parser): compound-subject owned-off-battlefield land-type static (Dune Chanter)

### LINKED ISSUE #5814: parser: compound-subject additive land-type static drops off-battlefield conjunct (Dune Chanter)
## Summary

`Dune Chanter`'s compound-subject land-type static — `Lands you control and land cards you own that aren't on the battlefield are Deserts in addition to their other types.` — silently drops the second conjunct. `parse_continuous_subject_filter` (the generic subject resolver) has no branch for the "`<type>` cards you own that aren't on the battlefield" shape, so `parse_controlled_compound_continuous_subject_filter`'s "X and Y" splitter can't resolve the second half, and the static ends up scoped to battlefield lands only — the off-battlefield land cards the printed text explicitly names are never affected.

Verified against Scryfall (`oracle_id: 7811b50d-af76-43cd-8753-43cf65fbefa3`):

```
Reach
Lands you control and land cards you own that aren't on the battlefield are Deserts in addition to their other types.
Lands you control have "{T}: Add one mana of any color."
{T}: Mill two cards. You gain 1 life for each land card milled this way.
```

## Structural parallel to #5406 / #5219

Same structural gap shape as #5406 (Rukarumel's compound-subject chosen-type static): two subject clauses joined by "and", each individually resolvable via an existing building block, but the combination unreached because the shared resolver had no branch for one of the two shapes.

| Axis | #5406 (chosen type) | This issue (owned off-battlefield land type) |
|---|---|---|
| Predicate-owning dispatcher | `parse_arcane_adaptation_chosen_type_static` (bespoke 3-arm subject match) | `parse_subject_additive_type_static` (already delegates subject resolution to `parse_continuous_subject_filter`) |
| Missing conjunct shape | `X you control and Y you control` (each already resolvable) | `Y cards you own that aren't on the battlefield` (resolvable elsewhere, but only inside a keyword-grant-specific dispatcher, never reachable from the generic subject filter) |
| Fix shape | New compound-subject sibling dispatcher | One new branch inside the already-generic `parse_continuous_subject_filter`, reusing (via extraction) the keyword-grant family's existing off-battlefield-ownership grammar |
| Runtime | Already wired (`AddChosenSubtype`) | Already wired (`AddSubtype`, `FilterProp::InAnyZone`) |

## Root cause

1. `parse_subject_additive_type_static` (`crates/engine/src/parser/oracle_static/anthem.rs`) is the general "`<subject>` are/is `<type>` in addition to its/their other types" dispatcher. It already delegates subject resolution *whole* to `parse_continuous_subject_filter` (`oracle_static/shared.rs`) — no bespoke subject matcher of its own.
2. `parse_continuous_subject_filter("Lands you control and land cards you own that aren't on the battlefield")` reaches `parse_controlled_compound_continuous_subject_filter`, which splits on `" and "` and recursively resolves each conjunct via `parse_continuous_subject_filter` itself:
   - `"Lands you control"` resolves fine (existing `parse_typed_you_control_subject_filter` path).
   - `"land cards you own that aren't on the battlefield"` has **no branch anywhere in `parse_continuous_subject_filter`'s dispatch chain** — it isn't a "you control" phrase, doesn't match any of the other special-shape checks, and the final `parse_type_phrase` fallback only consumes the leading type word, leaving `"cards you own that aren't on the battlefield"` unconsumed (non-empty `rest` → decline).
   - Since the second conjunct fails to resolve, `parse_controlled_compound_continuous_subject_filter` returns `None` entirely (the `?` after the second conjunct's resolution propagates), so `parse_continuous_subject_filter` falls through to its own final fallback and also returns `None` for the whole compound subject.
   - Result: `parse_subject_additive_type_static` declines the whole line, and — because `parse_land_type_change` also declines separately (`Desert` isn't one of the five `BasicLandType` variants per CR 305.6, so it can't route through the basic-land-type machinery either) — the static is lost entirely.
3. The "`<type>` cards you own that aren't on the battlefield" grammar **does already exist**, but only inside `parse_spells_have_keyword`'s Pattern 2 (`oracle_static/keyword_grant.rs`), scoped to the casting-keyword family (Leyline of Anticipation: "Creature cards you own that aren't on the battlefield have flash."). It was never generalized into `parse_continuous_subject_filter`, so no other static family — land type-change, anthem, color, etc. — can reach it.

## Proposed fix (mirrors #5406's shape: extract a shared building block, wire it into the generic resolver)

1. Extract Pattern 2's subject→filter logic (leading type phrase + `ControllerRef::You` + `FilterProp::InAnyZone{Hand,Graveyard,Exile,Command}`) into a new shared `parse_owned_off_battlefield_subject_filter` helper in `oracle_static/shared.rs`, reused — not reimplemented — by `parse_spells_have_keyword`'s Pattern 2. Behavior-preserving extraction: identical output for the existing "Creature cards you own that aren't on the battlefield have flash" test.
2. Add one new branch to `parse_continuous_subject_filter` that tries this helper (after the existing special-shape checks, before the final generic `parse_type_phrase` fallback). Once the second conjunct resolves, `parse_controlled_compound_continuous_subject_filter`'s existing "X and Y" splitter and its `filter_has_source_or_controller_anchor` guard (which already accepts a `Typed` filter with `controller == Some(You)`) handle the rest for free — no new compound-subject dispatcher needed, unlike #5406.
3. No new `ContinuousModification`, `Layer`, `StaticMode`, or runtime — `AddSubtype` and `FilterProp::InAnyZone` are already fully wired.
4. Tests: a parser-shape test on Dune Chanter's compound line (`Or` of a battlefield-scoped conjunct and an off-battlefield-owned conjunct, `AddSubtype{"Desert"}`), a full-oracle integration test (compound static present, sibling "Lands you control have mana ability" static and the mill activated ability unaffected), a no-regression test (a single-subject additive land-type line stays a plain `Typed` filter, not widened into an `Or`), and a filter-shape-preservation test on the extracted helper's original caller (Pattern 2's controller/zone scoping is unchanged after extraction).

## CR references (verified against `docs/MagicCompRules.txt`)

- CR 611.3 — a continuous static effect applies to every object its text indicates (the `Or`-distributed affected set).
- CR 305.6 — the five basic land types (`Desert` is not one of them, so the basic-land-type machinery in `parse_land_type_change` can never apply here).
- CR 305.7 — land-subtype-changing effects.
- CR 205.1b — additive type/subtype retention ("in addition to its/their other types").
- CR 109.4 — objects that are neither on the stack nor the battlefield have no controller.
- CR 109.5 — "you"/"your" on an object falls back to reading its owner when it has no controller.

jsdevninja:
## Summary

Fixes #5814. `Dune Chanter`'s compound-subject land-type static — `Lands you control and land cards you own that aren't on the battlefield are Deserts in addition to their other types.` (Oracle text verified against Scryfall, `oracle_id: 7811b50d-af76-43cd-8753-43cf65fbefa3`) — silently dropped the second conjunct: the static ended up scoped to battlefield lands only, never affecting the off-battlefield land cards the printed text explicitly names.

It's the same *structural* gap shape as **#5406** (Rukarumel's compound-subject chosen-type static): two subject clauses joined by "and", each individually resolvable via an existing building block, but the combination unreached — here because the shared generic subject resolver had no branch for one of the two shapes at all.

## Fix (mirrors #5406, but even more surgical)

`parse_subject_additive_type_static` (the general "`<subject>` are/is `<type>` in addition to its/their other types" dispatcher) already delegates subject resolution *whole* to `parse_continuous_subject_filter` — unlike Rukarumel's predicate-owning dispatcher, it has no bespoke subject matcher of its own. So the fix is one level lower than #5406's new compound-subject sibling function:

- **`parse_owned_off_battlefield_subject_filter`** — a new shared helper in `oracle_static/shared.rs`, extracted (behavior-preserving) from `parse_spells_have_keyword`'s Pattern 2 (`oracle_static/keyword_grant.rs`, the Leyline of Anticipation class: `"Creature cards you own that aren't on the battlefield have flash."`). Reused, not reimplemented, by both callers.
- One new branch inside **`parse_continuous_subject_filter`** tries this helper. Once `"land cards you own that aren't on the battlefield"` resolves, the *already-existing* `parse_controlled_compound_continuous_subject_filter` "X and Y" splitter and its `filter_has_source_or_controller_anchor` guard (which already accepts a `Typed` filter with `controller == Some(You)`) handle the compound union for free — no new compound-subject dispatcher, no `dispatch.rs` wiring needed.
- `Desert` is **not** a basic land type (CR 305.6 lists only Plains/Island/Swamp/Mountain/Forest), so `parse_land_type_change`'s basic-land-type machinery can never apply here regardless of the subject fix — the additive-type path (`parse_subject_additive_type_static` → `parse_additive_type_clause_modifications`) is the only correct route, and it's already fully wired for arbitrary subtypes via the master subtype table.

## Architecture

- Nom combinators throughout; reuses `parse_type_phrase`, `scan_contains`, `scan_split_at_phrase`.
- Builds for the class: any static that names a `<type> cards you own that aren't on the battlefield` conjunct — alone or compounded with a battlefield-scoped conjunct — can now resolve through the generic subject filter, not just the casting-keyword family.
- **No new `ContinuousModification` / `Layer` / `StaticMode` / runtime** — `AddSubtype` and `FilterProp::InAnyZone` are already fully wired end-to-end.
- CR-annotated: **CR 611.3 / CR 305.6 / CR 305.7 / CR 205.1b / CR 109.4 / CR 109.5**.

## Testing

`cargo fmt --all -- --check` ✓ (clean across the whole workspace) · `scripts/check-parser-combinators.sh` ✓ (clean, no forbidden non-combinator patterns in the diff).

New tests (`oracle_static/tests.rs`):
- `parser_shape_dune_chanter_compound_owned_land_type_static` — modeled sentence → `Or` of a battlefield-scoped `Typed{Land, controller: You}` conjunct and an off-battlefield-owned `Typed{Land, controller: You, InAnyZone{Hand,Graveyard,Exile,Command}}` conjunct, plus `AddSubtype{"Desert"}`.
- `parse_dune_chanter_full_oracle_adds_compound_static_and_preserves_siblings` — full printed card: the compound static is present **and** the sibling `"Lands you control have '{T}: Add one mana of any color.'"` static and the mill activated ability are unaffected.
- `single_subject_additive_land_type_static_not_hijacked_by_compound_conjunct_branch` — no-regression: a single-subject additive land-type line (no `" and "`) stays a plain `Typed` filter, not widened into an `Or`.
- `spells_have_keyword_pattern2_owned_off_battlefield_filter_shape_preserved` — no-regression: the extracted helper's original caller (`Creature cards you own that aren't on the battlefield have flash.`) still produces the exact same controller + off-battlefield-zone filter shape after extraction, not just the coarser `type_filters.contains(Creature)` the pre-existing test checks.

**Note on local verification:** this environment doesn't have a working MSVC toolchain (no `link.exe`), so I could not run `cargo check`/`cargo test`/`clippy` locally. I traced every function call in the fix path against the current `main` source by hand (exact signatures, dispatch ordering, and the `filter_has_source_or_controller_anchor` guard that gates the compound splitter), and `cargo fmt --check` plus the parser-combinator gate both ran clean. Please treat CI as the first real compile/test signal on this PR and flag anything it catches.


matthewevans (changes requested):
**Verdict — blocked: the parser now represents Dune Chanter's off-battlefield clause, but the engine cannot apply its type-changing continuous effect outside the battlefield.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_static/shared.rs:3457-3462` emits `AddSubtype { Desert }` for hand, graveyard, exile, and command-zone cards. The Dune Chanter Oracle line under test is explicit: “land cards you own that aren't on the battlefield are Deserts in addition to their other types.”  The only off-zone continuous-characteristics pipeline is `crates/engine/src/game/off_zone_characteristics.rs:82-107`; it collects `Layer::Ability` effects and exposes only effective keywords. Its allowlist at `:190-212` excludes `ContinuousModification::AddSubtype`, so a card in any of those zones never becomes a Desert at runtime. The new tests at `crates/engine/src/parser/oracle_static/tests.rs:22560-22647` assert only parsed AST shape/full-card parsing, not the claimed rules behavior. Please either implement the general off-zone type-characteristics path (including owner-zone matching and a runtime regression covering a hand/graveyard/exile recipient) or keep this line unsupported until that engine capability exists. Parser-only support would make coverage claim behavior the game does not realize.

- `crates/engine/src/parser/oracle_static/shared.rs:3440-3449` introduces a shared grammar using a full-text `scan_contains` guard plus a scan for the first `cards` token. That accepts any subject merely containing the phrase and does not prove that the entire suffix is this ownership grammar. Please make it an all-consuming, composable subject parser (type phrase + `cards you own` + `that aren't on the battlefield`) before reusing it outside the existing keyword-grant path; this is the live Gemini finding and is confirmed by the current head.

## 🟡 Non-blocking

- `crates/engine/src/parser/oracle_static/anthem.rs` correctly declines the early `you control` arm for the compound form, allowing the existing compound-subject resolver to receive both conjuncts.

## ✅ Clean

- The full-Oracle parser test keeps Dune Chanter's mana grant and activated ability in scope, and the existing Leyline-shaped keyword grant receives a shape-regression test. GitHub CI is green for `adbf62e`.

**Recommendation: request changes — land the off-zone runtime characteristic support and a discriminating runtime test, then return for review.**

jsdevninja:
Thanks for the review — both blockers are legitimate and I've pushed a fix (217ba1c8a).

**Blocker 1 (off-zone runtime gap):** Confirmed by reading `game/off_zone_characteristics.rs` directly — `supports_off_zone_keyword_query`'s allowlist (`AddKeyword`, `RemoveKeyword`, `AddDynamicKeyword`, `AddKeywordWithDerivedCost`, `RemoveAllAbilities`, `RemoveChosenKeyword`, `AddChosenKeyword`) has no `AddSubtype`/`AddType` entry, and it's the *only* off-battlefield continuous-effect path. I went with your second option ("keep this line unsupported until that engine capability exists") rather than building the general off-zone type-characteristics path — that's a materially larger engine capability (owner-zone matching for type/subtype changes, Layer-system integration, its own runtime regression suite) than fits this PR's scope, and the codebase already has an established precedent for exactly this situation: Arcane Adaptation/Maskwood Nexus/Rukarumel's off-battlefield "same is true for..." tail is deliberately left as an `Unimplemented` residual for the same reason.

Concretely: reverted the `parse_continuous_subject_filter` wiring so `parse_owned_off_battlefield_subject_filter` is reachable only from `parse_spells_have_keyword`'s Pattern 2 (its original, correct use — `CastWithKeyword` is checked directly against a card's actual zone, so the off-battlefield scope *is* realized there). Dune Chanter's compound off-battlefield type-change line now declines entirely — no static is produced for it — rather than silently claiming a scope the Layer system can't reach. Filed a rules-correctness precedent, not a partial fix.

**Blocker 2 (non-anchored grammar):** Rewrote `parse_owned_off_battlefield_subject_filter` to be `all_consuming` — the whole trimmed subject must be exactly `<type phrase>` + `"cards you own that aren't on the battlefield"`, nothing before the type phrase (now checked via `parse_type_phrase`'s own `rest.trim().is_empty()`) and nothing after the fixed suffix (now checked via `all_consuming(tag(...))` on the scan remainder rather than just probing for `"cards"`). Added `owned_off_battlefield_subject_filter_declines_trailing_garbage` covering both leading-noise and trailing-garbage cases, plus a clean-match-with-trailing-period baseline.

Updated the Dune Chanter tests to assert the honest decline instead of the (incorrect) Or-with-off-battlefield-conjunct shape, while keeping the sibling-static/mill-ability regression checks. CI is green on 02317bfb5 (`Rust lint`, both `Rust tests` shards, `WASM`/`Tauri` compile checks).

The `parse_typed_you_control` dispatch-order fix and its Death Baron regression test are unchanged — that part of the review was flagged non-blocking and isn't affected by this update.

jsdevninja:
@matthewevans please review

matthewevans (approved):

PR #5834, opened 2026-07-14
fix(engine): bind damage_source for anaphoric "each other creature" damage recipients

### LINKED ISSUE #4960: Nova Flame Doesn't do damage — Played Nova Flame onto invisible woman, chose 6 for the cost of X.
Reported in Discord: https://discord.com/channels/1485498006781427802/1522092400816816208/1522092400816816208

**Thread:** Nova Flame Doesn't do damage
**Cards:** invisible woman, nova flame
**Parser status:** has_gaps

## Summary
Played Nova Flame onto invisible woman, chose 6 for the cost of X.

### invisible woman
At the beginning of combat on your turn, if you've cast a noncreature spell this turn, create a 0/3 colorless Wall creature token with defender and reach.
Whenever you attack, you may pay {R}{G}{W}{U}. When you do, target creature gets +1/+0 until end of turn for each creature you control and can't be blocked this turn.
### nova flame
Put X +1/+1 counters on target creature you control. It deals damage equal to its power to each other creature.

---
<sub>report_id: `discord:1522092400816816208:1522092400816816208:0` · discord: `1522092400816816208/1522092400816816208`</sub>

tryeverything24:
## Root cause

Fixes #4960.

Nova Flame ("Deal X damage to each other creature") resolves its damage recipients through the anaphoric "each other creature" expansion path. That expansion path never threaded the spell's actual `damage_source` through to the per-recipient damage effects it builds — recipients ended up with an unbound/default damage source instead of Nova Flame itself.

While fixing this, I found a related bug in the same area: the `Another`-exclusion filter (used to keep "other" from including the source permanent itself) was comparing against the wrong `FilterContext.source_id` in this code path, so in some contexts "other" could incorrectly include the source.

## Fix

- `crates/engine/src/game/effects/deal_damage.rs`: bind `damage_source` correctly when expanding anaphoric "each other creature" recipients.
- `crates/engine/src/parser/oracle_effect/mod.rs`: fix the `Another`-exclusion filter to use the correct `FilterContext` source id.
- New regression test: `crates/engine/tests/integration/issue_4960_nova_flame.rs`.

## Test output (isolated `CARGO_TARGET_DIR`, from-scratch build)

```
running 1 test
test issue_4960_nova_flame::nova_flame_deals_boosted_power_to_each_other_creature ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 3071 filtered out; finished in 0.02s
```

## Validation

- Ran only the new targeted test, not the full suite or clippy, due to local machine constraints (this box OOMs with more than ~2 concurrent full rebuilds). No `mtgish/` files touched.
- Did not go through the `/engine-implementer` skill pipeline.

### DIFF
--- a/crates/engine/src/game/effects/deal_damage.rs
+++ b/crates/engine/src/game/effects/deal_damage.rs
@@ -1560,6 +1560,17 @@ pub fn resolve_all(
     // protection purposes); rebinding it here too keeps recipient-set
     // membership consistent with which object CR 120.1 says the damage is
     // actually FROM.
+    // Only the resolved damage source's OBJECT identity is overridden here —
+    // `source_controller` is deliberately left as `FilterContext::from_ability`
+    // set it (the ability's own controller). Per `FilterContext`'s own
+    // documented invariant (`filter.rs`), `source_controller` is what
+    // `ControllerRef::You` ("creatures you control") resolves against, and
+    // that pronoun refers to the ABILITY's controller (who cast the spell) —
+    // not to whoever happens to control the resolved damage source (e.g. a
+    // stolen creature dealing the damage). An earlier version of this fix
+    // also overrode `source_controller`, which made "you control" resolve
+    // against the wrong player whenever the resolved source's controller
+    // differs from the caster.
     let mut ctx = filter::FilterContext::from_ability(ability);
     if matches!(damage_source, Some(DamageSource::Target)) {
         if let Some(resolved_source_id) = ability.targets.iter().find_map(|t| match t {
@@ -1567,7 +1578,6 @@ pub fn resolve_all(
             _ => None,
         }) {
             ctx.source_id = resolved_source_id;
-            ctx.source_controller = state.objects.get(&resolved_source_id).map(|o| o.controller);
         }
     }
     let matching_objects: Vec<_> = state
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**Blocked — the recipient filter still evaluates “you” against the original ability controller, not the resolved damage source’s controller.**

## 🔴 Blocker

- `crates/engine/src/game/effects/deal_damage.rs:1585-1591` replaces `filter_ctx.source_id` with `ctx.source_id` but leaves `filter_ctx.source_controller` from `FilterContext::from_ability(ability)`. When `damage_source` is a controlled-by-another-object target, `ControllerRef::You` inside the recipient filter is therefore evaluated against the wrong player. The current Gemini thread identifies this exact mismatch, and the new Nova Flame fixture never changes control (`issue_4960_nova_flame.rs:113-117`), so it cannot discriminate the branch. The existing `DamageContext` is already the authoritative source/controller pair; use `ctx.controller` alongside `ctx.source_id`.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The PR correctly moves damage-source resolution before recipient expansion and its existing end-to-end Nova Flame test distinguishes the original zero-damage/anaphoric-power failure. The current parse-diff artifact reports no parser blast radius.

**Recommendation: set `filter_ctx.source_controller = Some(ctx.controller)` and add a production-pipeline control-changing fixture whose recipient filter includes a controller-relative predicate; then request re-review.**

lopesluis73:
Thanks for the review — fixed. `filter_ctx.source_controller` is now set to the resolved damage source's controller (`ctx.controller`) alongside the existing `source_id` override, per CR 109.5. Added a discriminating unit test in `deal_damage.rs` using a synthetic `DamageAll` with a `ControllerMatches` recipient predicate and a source object whose controller differs from the ability's — confirmed to fail without the fix. Ready for re-review.

matthewevans (changes requested):
**Request changes — preserve the ability-controller meaning of `You`.**

## 🔴 Blocker

**[HIGH]** The damage path changes `source_controller`, contradicting the `ControllerRef::You` invariant. Evidence: `crates/engine/src/game/effects/deal_damage.rs:1592-1598`; `crates/engine/src/game/filter.rs:594-596`. Why it matters: filters resolving `You` can target the wrong player. Suggested fix: override only `source_id`; keep the ability controller intact.

lopesluis73:
Fixed — reverted to overriding only `filter_ctx.source_id`, leaving `source_controller` untouched so `ControllerRef::You` still resolves against the ability's controller. Updated the discriminating unit test to pin the corrected direction.

One note for context: while rebasing I found #5873 (already merged, approved) applies the same `resolve_all` restructuring but *does* override `source_controller` the same way this PR's first version did. Rebasing pulled that shared code in, so this PR's diff against current `main` is now just the one-line removal + the corrected test, not a full reimplementation — happy to also send a quick follow-up against #5873's already-merged code if that override should come out there too, just let me know.

Full suite green post-rebase: fmt clean, clippy clean, 16747 lib tests pass, 3205 integration tests pass.

matthewevans (approved):
Approved: reviewed current head b704ccd with discriminating parser/runtime coverage; correct damage-source binding and controller semantics.

PR #5836, opened 2026-07-14
fix(engine): stop object-bound discard from falling back to whole-hand discard when zero targets were forwarded

### LINKED ISSUE #4950: Thoughtseize — Thoughtseize made my opponent discard a land when they had no spells in hand, even though I could and di…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522268124441874624/1522268124441874624

**Thread:** Thoughtseize
**Cards:** thoughtseize

## Summary
Thoughtseize made my opponent discard a land when they had no spells in hand, even though I could and did not select it.

### thoughtseize
Target player reveals their hand. You choose a nonland card from it. That player discards that card. You lose 2 life.

---
<sub>report_id: `discord:1522268124441874624:1522268124441874624:0` · discord: `1522268124441874624/1522268124441874624`</sub>

tryeverything24:
## Root cause

Fixes #4950.

Thoughtseize parses as `RevealHand{card_filter: Non(Land)}` → `DiscardCard{target: ParentTarget}` → `LoseLife`.

When the revealed hand has no nonland card, `reveal_hand.rs`'s `resolve()` correctly detects the empty-eligible case (CR 608.2c: nothing to choose) and returns early without opening a `RevealChoice` and without rebinding any card into the chained discard's targets.

The chained `DiscardCard{target: ParentTarget}` then resolves with `specific_targets` empty (no objects were forwarded). `discard.rs`'s gating condition was:

```rust
if !specific_targets.is_empty() && (declared_target_discard || object_bound_discard) { ... } else { /* whole-hand discard */ }
```

Because `specific_targets` was empty, execution fell into the generic "discard from the whole hand" branch, which force-discards whenever hand size ≤ discard count. With only a land in hand and Thoughtseize's discard count of 1, this force-discarded the land — a card the player never chose and that shouldn't have been eligible at all.

## Fix

`crates/engine/src/game/effects/discard.rs`: drop the `!specific_targets.is_empty()` requirement from the gate. A declared-target or object-bound discard with zero forwarded objects is now a legitimate no-op instead of falling back to unrestricted whole-hand discard.

Two new regression tests cover the reported scenario and its normal-case counterpart:
- `reveal_choose_nonland_discard_is_noop_when_hand_has_no_nonland_card`
- `reveal_choose_nonland_discard_targets_the_chosen_nonland_card`

## Test output (isolated `CARGO_TARGET_DIR`, from-scratch build, foreground, no concurrent builds)

```
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 16588 filtered out; finished in 0.03s
```

All 29 pre-existing `discard.rs` tests plus the 2 new ones pass; no regressions. The sibling `reveal_hand::tests` module (12 tests) was also run as a sanity check — all pass.

## Validation

- Ran the `discard`/`reveal_hand` unit-test modules and confirmed the whole `engine` lib target compiles cleanly; did not run the full integration suite or clippy due to local machine constraints.
- No `mtgish/` files touched.
- Did not go through the `/engine-implementer` skill pipeline.

### DIFF
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -3630,6 +3630,7 @@ fn walk_ability(
         replacement_applied: _,
         dig_found_nothing_for_parent_target: _,
         choose_from_zone_found_nothing_for_parent_target: _,
+        reveal_choice_found_nothing_for_parent_target: _,
     } = a;
 
     // §4.3.2: a definition's own `player_scope` overrides the inherited scope for
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -3628,9 +3628,7 @@ fn walk_ability(
         chosen_players: _,
         sub_link: _,
         replacement_applied: _,
-        dig_found_nothing_for_parent_target: _,
-        choose_from_zone_found_nothing_for_parent_target: _,
-        reveal_choice_found_nothing_for_parent_target: _,
+        parent_target_missing_reason: _,
     } = a;
 
     // §4.3.2: a definition's own `player_scope` overrides the inherited scope for
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -205,6 +205,7 @@ fn resolved_ability_axes(a: &ResolvedAbility) -> Axes {
         sub_link: _,              // SubAbilityLink kind tag
         dig_found_nothing_for_parent_target: _, // bool seam flag
         choose_from_zone_found_nothing_for_parent_target: _, // bool seam flag
+        reveal_choice_found_nothing_for_parent_target: _, // bool seam flag
     } = a;
 
     let mut acc = scan_effect(effect);
@@ -4735,6 +4736,7 @@ pub(crate) fn ability_resolution_choice_freedom(a: &ResolvedAbility) -> Resoluti
         sub_link: _,  // SubAbilityLink kind tag
         dig_found_nothing_for_parent_target: _, // bool seam flag
         choose_from_zone_found_nothing_for_parent_target: _, // bool seam flag
+        reveal_choice_found_nothing_for_parent_target: _, // bool seam flag
     } = a;
 
     // CR 608.2d: an optional effect / optional targeting / opponent-may
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -203,9 +203,7 @@ fn resolved_ability_axes(a: &ResolvedAbility) -> Axes {
         chosen_players: _,        // concrete chosen player ids
         replacement_applied: _,   // replacement provenance set, no dynamic read
         sub_link: _,              // SubAbilityLink kind tag
-        dig_found_nothing_for_parent_target: _, // bool seam flag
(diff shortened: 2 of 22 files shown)


matthewevans (changes requested):
**Changes requested — this fixes Thoughtseize’s empty-choice path by disabling ordinary player-targeted discard.**

## 🔴 Blocker
- `extract_target_filter_from_effect` returns `Some` for `DiscardCard { target: Player }`, so `declared_target_discard` is true for effects such as “target player discards a card.” The new `if declared_target_discard || object_bound_discard` then takes the specific-object loop with an empty `specific_targets` list (a player target is not an object), rather than opening the normal `DiscardChoice` path.
- Current-head CI demonstrates the regression: targeted/player discard no longer happens in Chain of Smog, Sonic Shrieker, Tinybones, Archon, and Skullscorch coverage. Keep the no-op behavior only for an object-reference target such as `ParentTarget`/`LastRevealed` when no object was forwarded; a player target must continue through the generic player-hand discard flow. Please add a focused regression proving both paths.

## 🟡 Non-blocking
- Run `cargo fmt --all`; the current lint job also reports formatting changes in the two added tests.

## ✅ Clean
- The Thoughtseize report and its two scenario tests correctly identify the intended behavior: no eligible nonland card means no discard, while a selected nonland is discarded.

**Recommendation: distinguish object-reference discard targets from player targets at this resolver seam, preserve the normal discard-choice path for players, and request re-review.**

matthewevans (changes requested):
**Changes requested — the repaired behavior is right, but the provenance carrier should be typed rather than adding a third parallel boolean.**

## 🔴 Blocker
- `last_dig_found_nothing`, `last_choose_from_zone_found_nothing`, and the new `last_reveal_choice_found_nothing` are three mutually-exclusive reasons that the immediate `ParentTarget` child has no object. The matching three fields on `ResolvedAbility` repeat the same sum type, forcing every constructor, matcher, scanner, and batch predicate to grow another `false`/ignored field.
- Please replace the parallel booleans with one typed `Option<ParentTargetMissingReason>` relay (for example `Dig`, `ChooseFromZone`, and `RevealHandChoice`), then have each consumer match only its applicable reason. This preserves the narrow Thoughtseize no-op while making future missing-choice sources exhaustive rather than creating a fourth boolean pair.

## 🟡 Non-blocking
- None.

## ✅ Clean
- The new gate correctly restores player-scoped discard behavior while preserving Thoughtseize’s empty-eligible no-op. The two production-pipeline tests cover both the empty and selected-card branches, and the existing CI failures from the prior head are addressed by the branch logic.

**Recommendation: parameterize the missing-parent-target provenance, then request re-review after the refreshed CI and parse-diff artifact complete.**

lopesluis73:
Thanks — refactored. Replaced the three parallel booleans (`last_dig_found_nothing`, `last_choose_from_zone_found_nothing`, `last_reveal_choice_found_nothing` on `GameState`, and their `ResolvedAbility` mirrors) with a single `Option<ParentTargetMissingReason>` pair (`Dig` / `ChooseFromZone` / `RevealHandChoice` variants). `apply_parent_chain_context`'s consume-immediately semantics already treated these as mutually exclusive, so `Option::take()` is a direct, behavior-preserving translation of the prior if-let-then-clear triple — verified via full fmt/clippy/lib/integration suite, no functional change. Ready for re-review.

matthewevans (approved):

PR #5838, opened 2026-07-14
fix(parser): generalize count-form draw-replacement antecedent to any N (CR 121.2a)

### LINKED ISSUE #5678: Alms Collector: 'would draw two or more cards' antecedent never parses to a Draw replacement (CR 121.2a instruction-count gap)

lourincedaging0-commits:
Closes #5678.

## Problem
The draw-replacement branch of `parse_replacement_line_inner` recognized only two antecedents:

```rust
alt((
    value(DrawReplacementScope::IndividualDraw,   tag("would draw a card")),
    value(DrawReplacementScope::InstructionCount, tag("would draw one or more cards")),
))
```

Any other count-form — **"would draw two or more cards"** (Alms Collector), "three or more", etc. — matched neither arm, so `mentions_draw` stayed false and the line produced **no replacement at all**. The opponent drew freely and the card silently did nothing.

## Fix
Per **CR 121.2a**, a count-form antecedent modifies the draw *instruction* before any individual card draw is considered — so the specific count is irrelevant to the scope. Every "N or more cards" is `InstructionCount`. I replaced the verbatim `one or more cards` tag with the existing `parse_number` building block:

```rust
value(
    DrawReplacementScope::InstructionCount,
    (tag("would draw "), nom_primitives::parse_number, tag(" or more cards")),
)
```

- Reuses the shared number combinator (words *and* numerals) instead of enumerating each count as a verbatim tag — this is the general class, not one card.
- The `IndividualDraw` arm stays **first** in the `alt`, so the singular "a card" is never reached by `parse_number`'s `"a" => 1` branch.

## Tests
- `draw_replacement_count_antecedent_maps_any_count_to_instruction_scope` — building-block coverage: "a card" stays `IndividualDraw`; "one / two / three or more cards" all map to `InstructionCount`.
- `alms_collector_two_or_more_draw_replacement_parses` — the issue's card now lowers to a `Draw` replacement scoped to the instruction.

Full engine suite green locally: **16602 passed, 0 failed**.

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -418,9 +418,21 @@ fn parse_replacement_line_inner(text: &str, card_name: &str) -> Option<Replaceme
                 DrawReplacementScope::IndividualDraw,
                 tag::<_, _, OracleError<'_>>("would draw a card"),
             ),
+            // CR 121.2a: a count-form antecedent — "would draw <N> or more cards"
+            // for any N ("one", "two", "three", ...) — modifies the draw
+            // INSTRUCTION before any individual card draw is considered, so every
+            // count collapses to the same InstructionCount scope (Quirion-style
+            // "one or more", Alms Collector "two or more"). Reuses `parse_number`
+            // rather than enumerating each count as a verbatim tag. The
+            // IndividualDraw arm above is tried first, so the singular "a card"
+            // never reaches `parse_number`'s "a" => 1 branch here.
             value(
                 DrawReplacementScope::InstructionCount,
-                tag("would draw one or more cards"),
+                (
+                    tag::<_, _, OracleError<'_>>("would draw "),
+                    nom_primitives::parse_number,
+                    tag(" or more cards"),
+                ),
             ),
         ))
         .parse(i)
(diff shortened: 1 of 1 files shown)


lourincedaging0-commits:
Closing: the one-arm parser fix trips the `draw_replacement_census.py` scope cross-check gate, and doing it correctly requires 3 coordinated changes (typed count-threshold condition in the parser, teaching `classify_scope`, and runtime verification of `EventContextAmount` at draw-match time) plus the still-unimplemented substitute effect. Details posted on #5678. Not shipping a broken/half fix.

PR #5842, opened 2026-07-14
fix(parser): normalize sentence comma inside a quoted granted ability closing quote

### LINKED ISSUE #5681: Parser drops "until end of turn" from a quoted granted ability when Oracle puts the comma inside the closing quote (Bronzehide Lion)

lourincedaging0-commits:
Closes #5681.

## Problem
Oracle's punctuation convention puts the *enclosing* sentence's comma **inside** the closing quote of a granted ability when a clause follows. Bronzehide Lion:

> When this creature dies, return it to the battlefield. It's an Aura enchantment with enchant creature you control and "{G}{W}: Enchanted creature gains indestructible until end of turn," and it loses all other abilities.

The quoted body handed to the inner sub-parse was `...until end of turn,` (trailing comma). `.trim()` strips only whitespace, so the comma reached the duration combinator, which failed to match (`"until end of turn," != "until end of turn"`). The phrase fell through to prose and the `UntilEndOfTurn` duration was **silently dropped** — at runtime the granted ability made the creature indestructible **permanently** instead of until end of turn.

## Fix
Normalize the trailing comma at `classify_quoted_inner` — the single boundary every quoted-grant caller funnels through (aura grants, token grants, keyword-list grants) — so a quoted ability parses identically to its unquoted form.

**Only the comma is stripped.** A trailing *period* is the ability's own terminal punctuation (`"{T}: Add {G}."`, `"...equal to the difference."`) that must be preserved in the serialized `description` (#5599) and does not defeat any inner combinator. (An earlier draft that also stripped periods regressed the `compound_subject_granted_ability_description_keeps_original_case` and `conformer_shuriken` description snapshots — hence comma-only.)

## Tests
- **`classify_quoted_inner_normalizes_trailing_sentence_comma`** — building-block coverage: a comma-tailed body parses identically to its clean form, the `UntilEndOfTurn` duration survives, and a terminal period is preserved.
- **`try_parse_bronzehide_lion_chunk`** — strengthened to assert the granted ability keeps its typed `UntilEndOfTurn` duration (previously only checked the grant was non-empty, which passed even with the bug).
- **`snapshot_bronzehide_lion`** — re-accepted: the granted definition now carries `"duration": "UntilEndOfTurn"` and the `SwallowedClause / Duration_UntilEndOfTurn` parse warning is gone (the designed workflow per the issue).

Full engine lib suite: **16611 passed, 0 failed**. Integration suite: only the Bronzehide snapshot changed (harold_and_bob, old_growth_troll, and the other quoted-grant snapshots unchanged), confirming the blast radius matches the issue's class measurement.

### DIFF
--- a/crates/engine/src/parser/oracle_nom/return_as_aura.rs
+++ b/crates/engine/src/parser/oracle_nom/return_as_aura.rs
@@ -124,7 +124,7 @@ fn parse_loses_clause(input: &str) -> OracleResult<'_, ()> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::types::ability::{ControllerRef, TypeFilter, TypedFilter};
+    use crate::types::ability::{ControllerRef, Duration, TypeFilter, TypedFilter};
 
     fn lower(s: &str) -> String {
         s.to_ascii_lowercase()
@@ -176,17 +176,15 @@ mod tests {
         // turn,"). That trailing comma must be normalized away before the inner
         // sub-parse, so the duration survives as a typed `UntilEndOfTurn` rather
         // than being dropped into prose. Assert on the typed duration, not the
-        // leftover text.
-        let debug = format!("{grants:?}");
-        // The granted duration is nested inside a granted activated ability, so a
-        // Debug-substring check is clearer than walking the full AST.
-        // allow-noncombinator: string assertion in test
-        let has_typed_duration = debug.contains("UntilEndOfTurn");
-        assert!(
-            has_typed_duration,
-            "granted ability lost its UntilEndOfTurn duration to the trailing comma inside \
-             the closing quote: {debug}",
-        );
+        // debug representation.
+        let ContinuousModification::GrantAbility { definition } = grants
+            .iter()
+            .find(|grant| matches!(grant, ContinuousModification::GrantAbility { .. }))
+            .expect("expected the quoted activated ability to be granted")
+        else {
+            unreachable!()
+        };
+        assert_eq!(definition.duration, Some(Duration::UntilEndOfTurn));
     }
 
     #[test]
--- a/crates/engine/src/parser/oracle_nom/return_as_aura.rs
+++ b/crates/engine/src/parser/oracle_nom/return_as_aura.rs
@@ -171,6 +171,22 @@ mod tests {
         }
         assert!(!grants.is_empty());
         assert!(!loses_other);
+        // #5681: the granted ability's "until end of turn" is carried with the
+        // enclosing sentence's comma INSIDE the closing quote ("...until end of
+        // turn,"). That trailing comma must be normalized away before the inner
+        // sub-parse, so the duration survives as a typed `UntilEndOfTurn` rather
+        // than being dropped into prose. Assert on the typed duration, not the
+        // leftover text.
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
**Approved — the shared quoted-grant boundary now removes only the enclosing comma and preserves the typed ability duration.**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- The contributor template is incomplete (audit-only); the current-head CI, parser gate, snapshot, and typed regression assertions provide the review evidence.

## ✅ Clean

- `crates/engine/src/parser/oracle_static/keyword_grant.rs:1839` normalizes at the shared `classify_quoted_inner` boundary.
- The Bronzehide Lion snapshot records `UntilEndOfTurn` and removes the swallowed-clause warning; the tests assert the typed duration both at that boundary and through the return-as-Aura route.

PR #5850, opened 2026-07-15
Fix damage redirection for Palisade Giant / Veteran Bodyguard / Weathered Bodyguards

rykerwilliams:
## Summary

Palisade Giant, Veteran Bodyguard, and Weathered Bodyguards all parse correctly today (`ShieldKind::Prevention { amount: PreventionAmount::All }` + `redirect_target: Some(TargetFilter::SelfRef)` on their `ReplacementDefinition`), but the runtime handler for `ShieldKind::Prevention` shields never read `redirect_target` — it fully prevented the damage instead of redirecting it. The intended recipient never actually took the damage, could never die from it, and never triggered "dealt damage" abilities off it.

A separate, already-correct branch for `ShieldKind::Redirection` does correctly replace the damage event's target per CR 614.9, but is one-shot (consumes the shield after firing) — correct for its own consumers (Soltari Guerrillas, Beacon of Destiny, Jade Monolith, Goblin Psychopath, the en-Kor cycle — all genuinely one-shot "the next time..." spell effects), but wrong for these 3 cards' continuous, ongoing static-ability lifecycle.

## Fix

Extracted the existing CR 614.9 redirect mechanics (resolve recipient → check legality → swap-or-fallback-to-original) out of the one-shot `Redirection` branch into a shared helper parameterized by `consume_after_redirect: bool`, and added a new gate in the `Prevention` branch:

```rust
if redirect_target_for_rid(state, rid) == Some(TargetFilter::SelfRef) && matches!(amount, PreventionAmount::All)
```

The `PreventionAmount::All` conjunct is required, not optional — the shared helper's body contains an `unreachable!()` arm for `PreventionAmount::AllBut` that's a true invariant only for the one-shot `Redirection` shield (which never produces `AllBut`), not for `Prevention` shields in general (which legitimately use `AllBut` for other cards, e.g. Temple Altisaur). Gating on `redirect_target.is_some()` alone would let a hypothetical future card pairing `AllBut`/`Next` with a redirect reach that arm and panic in production — no such card exists today, and this gate keeps it that way honestly.

When the gate passes, the redirect runs non-consuming, preserving the shield's continuous (re-firing every event) lifecycle.

## Scope

This PR fixes exactly 3 cards, verified against real Scryfall Oracle text (not name-substituted test fixtures — an earlier investigation round caught that trap on 2 other cards, see below). Explicitly **out of scope**, left untouched:

- **Pariah** — its real text is "...is dealt to **enchanted creature** instead," which the redirect-detection parser branch doesn't recognize at all today (only literal self-reference). Needs its own parser fix.
- **Pariah's Shield** — architecturally a CR 615.5 prevention-with-additional-effect card (not CR 614.9 redirection); applying this fix's mechanism to it would actually *regress* it, since it's an enchantment and would always fail the redirect-recipient legality check.
- **Palisade Giant's "and other permanents you control"** clause — a separate, pre-existing gap in the same conditional, already documented in an existing test's comments.

## CR references

- CR 614.9 (redirection effects; "the effect does nothing" on an illegal recipient)
- CR 510.2 (simultaneous combat damage)
- CR 614.5 / CR 614.7a (pre-existing, unmodified)
- CR 615 / CR 615.1a (pre-existing, unmodified)
- CR 704.5g (lethal damage → destroy, exercised by the new death test)

## Verification

- `cargo fmt --all` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- 6 new runtime tests (`crates/engine/tests/integration/palisade_giant_redirect.rs`): redirect + death, continuous (non-consumed) lifecycle across multiple events, downstream "dealt damage" trigger firing, illegal-recipient fallback (CR 614.9 "does nothing"), off-battlefield host correctly excluded from candidacy, simultaneous multi-attacker combat damage
- `veteran_bodyguard_tap_redirect.rs`: corrected a stale doc comment that narrated this exact bug as a known, deliberately-unfixed gap, and added the positive `damage_marked` assertion that test had previously and deliberately omitted
- 4 existing one-shot `ShieldKind::Redirection` regression tests pass unmodified (proves the shared-helper refactor is byte-identical for its existing consumers)
- 3 existing parser unit tests pass unmodified (this is a runtime-only fix; parser output is unchanged)

### DIFF
--- a/crates/engine/src/game/replacement.rs
+++ b/crates/engine/src/game/replacement.rs
@@ -4,10 +4,10 @@ use std::sync::LazyLock;
 
 use crate::types::ability::{
     AbilityCost, AbilityDefinition, CombatDamageScope, ControllerRef, DamageModification,
-    DamageTargetFilter, DamageTargetPlayerScope, Effect, EffectScope, PostReplacementContinuation,
-    PreventionAmount, QuantityExpr, QuantityModification, ReplacementCondition,
-    ReplacementDefinition, ReplacementMode, ResolvedAbility, ShieldKind, TapStateChange,
-    TargetFilter, TargetRef,
+    DamageRedirectTarget, DamageTargetFilter, DamageTargetPlayerScope, Effect, EffectScope,
+    PostReplacementContinuation, PreventionAmount, QuantityExpr, QuantityModification,
+    ReplacementCondition, ReplacementDefinition, ReplacementMode, ResolvedAbility, ShieldKind,
+    TapStateChange, TargetFilter, TargetRef,
 };
 use crate::types::card_type::CoreType;
 use crate::types::counter::CounterType;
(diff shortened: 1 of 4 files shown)


matthewevans (approved):

PR #5853, opened 2026-07-15
fix(parser): resolve off-battlefield-owned compound subject (Dune Chanter) [WIP]

lourincedaging0-commits:
WIP toward #5814. Extracts `parse_owned_off_battlefield_subject_filter` (shared by keyword_grant Pattern 2, behavior-preserving) and wires it into the generic continuous-subject resolver so "<type> cards you own that aren't on the battlefield" resolves. The standalone conjunct resolves correctly and the full suite is otherwise green, but the compound "Lands you control and land cards you own..." still collapses to the battlefield conjunct instead of a two-way `Or`. **Draft carries a temporary diagnostic** in the Dune test to surface the compound splitter's direct return from CI logs (local builds blocked by disk). Will remove the diagnostic and finalize once the splitter behavior is pinned.

### DIFF
--- a/crates/engine/src/parser/oracle_static/keyword_grant.rs
+++ b/crates/engine/src/parser/oracle_static/keyword_grant.rs
@@ -522,23 +522,15 @@ pub(crate) fn parse_spells_have_keyword(tp: &TextPair<'_>, text: &str) -> Option
     }
 
     // Pattern 2: "Creature cards you own that aren't on the battlefield have flash"
-    // This grants flash to cards in non-battlefield zones.
-    if nom_primitives::scan_contains(subject, "cards you own that aren't on the battlefield") {
-        let (prefix, _) = nom_primitives::scan_split_at_phrase(subject, |i| tag("cards").parse(i))?;
-        let type_end = prefix.len();
-        let type_part = &tp.original[..type_end];
-        let (base_filter, _) = parse_type_phrase(type_part);
-        let affected = match base_filter {
-            TargetFilter::Typed(mut typed) => {
-                typed = typed.controller(ControllerRef::You);
-                // "aren't on the battlefield" means any zone except battlefield
-                typed.properties.push(FilterProp::InAnyZone {
-                    zones: vec![Zone::Hand, Zone::Graveyard, Zone::Exile, Zone::Command],
-                });
-                TargetFilter::Typed(typed)
-            }
-            _ => base_filter,
-        };
+    // (Leyline of Anticipation class) — grants a casting keyword to cards in non-
+    // battlefield zones. The subject→filter grammar is the shared
+    // `parse_owned_off_battlefield_subject_filter`, reused (not reimplemented) so
+    // the generic continuous-subject resolver can reach the same off-battlefield-
+    // ownership shape (Dune Chanter's additive land-type static).
+    if let Some(affected) = super::shared::parse_owned_off_battlefield_subject_filter(
+        subject,
+        &tp.original[..subject.len()],
+    ) {
         let mut def = StaticDefinition::new(StaticMode::CastWithKeyword { keyword })
             .affected(affected)
             .description(text.to_string())
--- a/crates/engine/src/parser/oracle_static/shared.rs
+++ b/crates/engine/src/parser/oracle_static/shared.rs
@@ -3237,6 +3237,56 @@ fn parse_bare_compound_subtype_subject_filter(subject: &TextPair<'_>) -> Option<
     Some(TargetFilter::Or { filters })
 }
 
+/// CR 109.4 + CR 109.5 + CR 400.3: "`<type>` cards you own that aren't on the
+/// battlefield" — an owner-scoped subject naming cards in every zone EXCEPT the
+/// battlefield. Off the battlefield an object has no controller (CR 109.4), so
+/// "you own" reads the owner (CR 109.5); `FilterProp::InAnyZone` enumerates the
+/// non-battlefield zones (hand, graveyard, exile, command).
+///
+/// Shared by `parse_spells_have_keyword`'s casting-keyword Pattern 2 (Leyline of
+/// Anticipation: "Creature cards you own that aren't on the battlefield have
+/// flash.") and the generic continuous-subject resolver (Dune Chanter's additive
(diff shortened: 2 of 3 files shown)


matthewevans:
**Triaged and held as author-declared WIP on `581baf2a` — not ignored, and not implementation-reviewed.** Recording the state visibly since this draft has been open since 2026-07-15 without maintainer-facing feedback.

## Why it is held

- The PR is a draft and its own description says the compound `Or` case is unresolved, so there is nothing to approve yet.
- `crates/engine/src/parser/oracle_static/tests.rs` carries an intentional `panic!("DIAG5814 split={split:?} ||| whole={whole:?} ...")`. That is a hard merge blocker in any state — it must come out before this can be reviewed, let alone enqueued. It is also the direct cause of the red `Rust tests` shards on this head, so the current CI signal says nothing about the parser change itself.
- The branch is `DIRTY` against current `main`. That staleness is ours (main has churned heavily through the cost/zone campaign since 07-15), so do not spend your own cycles on the rebase — flag it when the diagnostic is out and a maintainer will port it.

## On using CI as the debugger

You noted local builds are blocked by disk. Two things that avoid the round-trip:

- The two authorities in play are readable directly: `parse_controlled_compound_continuous_subject_filter` (`crates/engine/src/parser/oracle_static/anthem.rs:765`) does the `"and "` split, and its `filter_has_source_or_controller_anchor` guard (`crates/engine/src/parser/oracle_static/shared.rs:2828`) returns `None` for the whole compound unless **both** branches carry `controller: Some(ControllerRef::You | ControllerRef::Opponent)`. Whether your right-hand conjunct ("land cards you own that aren't on the battlefield") actually reaches that guard with a `You` controller, or whether the outer resolver matches a higher-priority branch before the compound splitter is consulted at all, are the two candidate explanations — I have not confirmed which, and a `dbg!`-free unit test asserting the splitter's return directly will discriminate between them without a `panic!`.
- A failing `assert_eq!` on the splitter's return prints the same `{:?}` payload the `panic!` does, and leaves the test suite meaningful for everyone else.

**Recommendation: keep this in draft, replace the `panic!` diagnostic with a normal failing assertion on `parse_controlled_compound_continuous_subject_filter`'s return, and mark ready for review once the compound resolves. Flag it here when you do and it will be rebased onto current `main` and reviewed.**

matthewevans:
**Correction to my hold comment above: the rebase I promised is off the table, and this PR should be closed as superseded. The reason is not the conflict or the `panic!` — it is that the work already landed, and the one part of this branch that is not already on `main` was deliberately reverted there on rules-correctness grounds.** I owe you the evidence rather than a one-line close, so all of it is below.

## 1. Issue #5814 was already fixed by an earlier, competing PR

- #5815 (`feat(parser): compound-subject owned-off-battlefield land-type static (Dune Chanter)`) was opened **2026-07-14T16:11Z** and squash-merged **2026-07-16T00:50Z** as `7403be8a5e`.
- It closed #5814, which is now `CLOSED / COMPLETED`.
- This branch's single commit is dated **2026-07-15T06:04Z** — roughly 14 hours *after* #5815 was already open against the same issue.

Two PRs were in flight on one issue and this one lost the race. That is a duplicate-work collision, not a defect in your implementation, and my earlier hold comment missed it because I triaged the branch state without checking the issue for an existing open PR. That check was mine to run.

## 2. The shared-helper extraction is already on `main`, under the same name

`main` carries `parse_owned_off_battlefield_subject_filter` in `crates/engine/src/parser/oracle_static/shared.rs:4504`, extracted from `parse_spells_have_keyword`'s Pattern 2 — the same refactor, the same call site, the same function name as this branch. Rebasing would resolve to "already applied."

## 3. The one remaining delta is the part `main` explicitly rejected

The genuine difference between this branch and `main` is the wiring in `parse_continuous_subject_filter`:

```rust
if let Some(filter) = parse_owned_off_battlefield_subject_filter(&lower, trimmed) {
    return Some(filter);
}
```

#5815 landed that exact branch in its first commit, then **reverted it during its own review**. `main` now carries a standing comment against re-adding it (`shared.rs:4444`, "deliberately NOT wiring ... as a general fallback"). I re-verified the reasoning independently rather than taking the comment's word for it:

- `game/off_zone_characteristics.rs` is the sole off-battlefield continuous-effect path, and its `supports_off_zone_keyword_query` allowlist (line 219) is **keyword-only** — `AddKeyword`, `RemoveKeyword`, `AddDynamicKeyword`, `AddKeywordWithDerivedCost`, `RemoveAllAbilities`, `RemoveChosenKeyword`, `AddChosenKeyword`. There is no `AddSubtype` or `AddType` arm.
- Dune Chanter's line is a **type change**, not a keyword grant: "Lands you control and land cards you own that aren't on the battlefield **are Deserts in addition to their other types**." (verified against Scryfall).

So a `Continuous` static whose `affected` set spans hand/graveyard/exile/command would claim a scope the Layer system never visits for an `AddSubtype` modification. The card reads as fully supported while silently under-delivering, which is worse than the honest `Effect::Unimplemented` gap `main` leaves today — the same posture already established for Arcane Adaptation / Maskwood Nexus / Rukarumel.

Your own CI corroborates this from the other direction: the parse-diff sticky on `581baf2a` reports **"No card-parse changes detected."** The branch moves no card either way in its current state.

## 4. On the `panic!`

`panic!("DIAG5814 split={split:?} ...")` is still present at `tests.rs` in this head and is still what reds the `Rust tests` shards. It is moot now, but for future branches: a failing `assert_eq!` on the authority's return prints the same `{:?}` payload and leaves the suite usable for everyone else.

## 5. Recommendation

Close as superseded by #5815. Nothing here needs to be salvaged onto a rebase — item 2 is already upstream and item 3 must not land as written.

If you want to keep pulling on this thread, the real remaining gap is **engine-side, not parser-side**: `off_zone_characteristics.rs` has no path for non-keyword continuous modifications, which is precisely the gate `main`'s comment names ("Revisit once an off-zone characteristics path exists for non-keyword modifications"). Extending that allowlist to `AddSubtype`/`AddType` — with the Layer-system work to actually realize it — is what would unblock Dune Chanter's second conjunct, plus the Arcane Adaptation / Maskwood Nexus family. That is a design-first change, so it wants an issue and an agreed approach before implementation rather than a direct PR.

PR #5854, opened 2026-07-15
fix(engine): wire "as ~ becomes attached, choose" replacement for Psychic Paper

### LINKED ISSUE #5821: Seam: as-becomes-attached choice + name/type override (Psychic Paper) — requesting `accepted`
## Seam: "as this becomes attached, choose …" replacement + name/type override

**Requesting the `accepted` label** to authorize a protected-scope PR (adds a choice `WaitingFor` + frontend modal) per the AI-contributor protected-architecture policy. Design below; I'll build to whatever shape you approve.

### Card
- **Psychic Paper** — *"As this Equipment becomes attached to a creature, choose a creature card name and a creature type. Equipped creature has ward {1}, it can't be blocked, and its name and creature type are the last chosen name and creature type. Equip {2}."*

### The seam (a class, not one card)
Two composable primitives, both reusable well beyond this card:

1. **As-becomes-attached replacement carrying player choices** (CR 614.1 — "as" is a replacement generating a choice as the object becomes attached). This is the attach-time analogue of the already-common "as ~ enters, choose …" CDA pattern. New `WaitingFor::ChooseAsAttached { … }` + a stored choice bound to the Equipment. Unlocks any "as ~ becomes attached to a permanent, choose …".
2. **Continuous name/type SET from a stored choice** — the equipped creature's name and creature type *become* the chosen values (CR 604 name / CR 205.1 subtypes via the layer system, CR 613). This is the name/type-setting counterpart to the existing P/T- and ability-granting equipment statics. Ward {1} + can't-be-blocked are already supported grants.

### CR grounding (to be grep-verified in the PR)
CR 614.1 ("as" replacement with a choice), CR 613 (layer system; name = CR 201/604, type-changing = CR 205.1/613.1c), CR 702.22 (ward — will grep-verify the exact number), CR 509/blocking restrictions ("can't be blocked").

### Layers
types (as-attached replacement + stored choice + WaitingFor/action/event) · game/replacement.rs or game/equipment (+ `game/mod.rs` if a new module) · game/layers.rs (name/type set) · parser/oracle_static.rs (nom combinator) · AI legal-actions · frontend "choose name + type" modal + i18n.

### Open design question for you
The name-choice input is a *card-name* picker (from the full card pool) + a creature-type picker — a heavier modal than the existing enum-choice modals. Happy to model it on whatever existing name-picker UX you prefer (e.g. how Pithing Needle / Cabal Therapy-style name choices are surfaced), or add a dedicated one.

### Runtime proof
Registered card-test: attach the Equipment, drive the name+type choice, assert the equipped creature's name/type read as the chosen values (layer-correct), ward {1} applies, and it can't be blocked; detach clears the grant.

Part of the Doctor Who set-completion effort. Not yet built — filing for acceptance first because it's protected scope.

nghetienhiep:
Psychic Paper's continuous grant (ward, can't-be-blocked, name/type set
from the last chosen name/type) was already implemented and tested, but
the leading "As this Equipment becomes attached to a creature, choose a
creature card name and a creature type" clause parsed to Unimplemented —
nothing ever actually prompted the choice, so the grant read empty.

Reuses the existing ReplacementEvent::Attached matcher/applier (declared
but never produced by any parser) and routes Effect::Attach's single
resolver through the replacement pipeline, so any "as it becomes
attached, choose ..." definition fires regardless of which ability moves
the attachment. The two-part choice ("a creature card name and a
creature type") is parsed via a small conjunction combinator that
re-dispatches each conjunct through the existing named-choice phrase
table, extended generically rather than special-cased to this card.

Closes #5821

### DIFF
--- a/crates/engine/src/game/effects/attach.rs
+++ b/crates/engine/src/game/effects/attach.rs
@@ -69,6 +69,55 @@ pub fn resolve(
         return Ok(());
     }
 
+    // CR 701.3a + CR 614.1a: Route the attach through the replacement pipeline
+    // so an "as it becomes attached, choose …" definition on the attachment
+    // (`ReplacementEvent::Attached`, Psychic Paper) can bind its choice as the
+    // attachment resolves — the attach-time analogue of "as ~ enters, choose".
+    let proposed = crate::types::proposed_event::ProposedEvent::Attach {
+        attachment_id,
+        target_id,
+        applied: Default::default(),
+    };
+    match crate::game::replacement::replace_event(state, proposed, events) {
+        crate::game::replacement::ReplacementResult::Execute(_) => {
+            if let Some(waiting_for) =
+                deliver_attach(state, attachment_id, target_id, source_id, events)
+            {
+                state.waiting_for = waiting_for;
+            }
+        }
+        crate::game::replacement::ReplacementResult::Prevented => {
+            events.push(GameEvent::EffectResolved {
+                kind: EffectKind::from(&ability.effect),
+                source_id,
+                subject: None,
+            });
+        }
+        crate::game::replacement::ReplacementResult::NeedsChoice(player) => {
+            // CR 616.1: multiple "becomes attached" replacements apply — park
+            // for the ordering choice. `handle_replacement_choice`'s
+            // `ProposedEvent::Attach` arm resumes via `deliver_attach` once
+            // the player orders them.
+            crate::game::replacement::park_waiting_for(state, player);
+        }
+    }
+
+    Ok(())
+}
+
+/// CR 701.3a + CR 614.1a: Perform the attach mutation and, if the attachment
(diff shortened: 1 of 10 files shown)


matthewevans (changes requested):
**Request changes — only actual attachment events may trigger `Attached`.**

## 🔴 Blocker

**[HIGH]** The replacement matches every attachment `ZoneChange` to battlefield, including an unattached Psychic Paper ETB. Evidence: `crates/engine/src/game/replacement.rs:3409-3435`. Why it matters: `Attached` triggers when no attachment occurred. Suggested fix: require an actual attach operation/event and add the unattached-ETB regression.

matthewevans (approved):
**Approved — the attach-time matcher now requires an actual attachment target, and the current runtime regression proves a bare Equipment ETB does not prompt.**

PR #5862, opened 2026-07-15
fix(engine): bind combat-damage sacrifice pools and effect-sacrificed referents (Descendants' Fury #4795)

### LINKED ISSUE #4795: Descendants' Fury does not put the revealed creature onto the battlefield
Reported in Discord: https://discord.com/channels/1485498006781427802/1521155140051140650/1521155140051140650

**Thread:** Descendants Fury
**Cards:** descendants' fury

## Summary
Reporter says Descendants' Fury does not put the revealed card onto the battlefield.

### descendants' fury
Whenever one or more creatures you control deal combat damage to a player, you may sacrifice one of them. If you do, reveal cards from the top of your library until you reveal a creature card that shares a creature type with the sacrificed creature. Put that card onto the battlefield and the rest on the bottom of your library in a random order.

---
<sub>report_id: `discord:1521155140051140650:1521155140051140650:0` · discord: `1521155140051140650/1521155140051140650`</sub>

shin-core:
Fixes #4795 (claimed there with the corrected root-cause). Verified the Oracle text on Scryfall and reproduced end-to-end on current main before building.

## Reproduction (real `apply` pipeline)
Combat → trigger → accept "you may sacrifice one of them" → **two divergences**:
1. The sacrifice was a **silent no-op** (attacker stayed on the battlefield, `EffectResolved{Sacrifice, subject: None}`).
2. After fixing that, the reveal dug through the **entire library** and bottomed everything — the shared-type creature never entered.

The parse is correct; the triage's fast-path hypothesis was stale (the `effect_context_object` capture machinery works). The real breaks are two **bypassed single authorities**:

## Defect 1 — sacrifice pool never bound the combat-damage set
`sacrifice::resolve` matched the raw `TrackedSet(0)` sentinel via `matches_target_filter`, whose id-level ladder has **no combat-damage rung**. "One of them" on a combat-damage trigger lives only in the `CombatDamageDealtToPlayer` event, reachable via `resolve_tracked_set_sentinel`'s `current_combat_damage_source_filter` rung (CR 510.2) — which the sacrifice resolver, **alone among tracked-set consumers** (`change_zone`, `shuffle`, `token_copy`, `cloak`, …), never called. Fix: route the filter through `resolve_tracked_set_sentinel` before deriving the pool (non-sentinel filters pass through unchanged). Same unification discipline as #5512.

## Defect 2 — `CostPaidObject` readers ignored effect-sacrificed referents
Both runtime `TargetFilter::CostPaidObject` arms (filter eval + `resolved_targets`) read only `ability.cost_paid_object`. An **effect** sacrifice (CR 608.2c later-instruction referent) is captured into `effect_context_object` — and the documented `cost_paid_object → effect_context_object` ladder was already implemented for the `ObjectScope::CostPaidObject` **quantity** arms (P/T, mana value) but missing from the **filter** layer. Fix: implement the same fallback in both arms so every `CostPaidObject` reader binds the same referent (CR 608.2k). Class-level: the whole "sacrifice one → later instruction reads the sacrificed object's properties" family.

## Tests (each fails on revert of its seam; revert-probed)
- **Single attacker**: sacrifice happens, reveal digs past a non-matching card, the shared-type creature **enters the battlefield**, rest to the bottom.
- **Multi-attacker**: drives the real `EffectZoneChoice` — the **chosen** creature's type (not a board rescan) drives the reveal; the unchosen type's match stays in the library. This is the identity-binding hostile fixture.
- **Negative**: no shared-type creature in the library → whole library revealed, nothing enters, all to the bottom (fail-closed referent, with a sacrifice reach-guard so the negative isn't vacuous).

## Verification
Full engine lib: **16672 passed, 0 failed**. Integration: **3130 passed, 0 failed** (incl. `issue_735_cost_paid_object_non_regression`). Clippy `-D warnings` clean; fmt clean. CR annotations (510.2, 608.2c, 608.2k, 400.7) grep-verified against `docs/MagicCompRules.txt`.

### DIFF
--- a/crates/engine/src/game/effects/sacrifice.rs
+++ b/crates/engine/src/game/effects/sacrifice.rs
@@ -143,7 +143,7 @@ pub fn resolve(
     // `QuantityExpr` (Fixed/Ref/DivideRounded/...) means a mandatory count;
     // wrapped in `UpTo` means the player may select 0..=count.
     let default_count = QuantityExpr::Fixed { value: 1 };
-    let (filter, count_expr, up_to, min_count) = match &ability.effect {
+    let (raw_filter, count_expr, up_to, min_count) = match &ability.effect {
         Effect::Sacrifice {
             target,
             count,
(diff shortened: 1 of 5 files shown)


matthewevans (approved):

PR #5869, opened 2026-07-15
fix(engine): recognize PayCost as a zone-change-ledger-invisible parent for generic "if you can't" riders

### LINKED ISSUE #4955: Greenbelt Rampager — My opponent resolved greenbelt rampager with two energy in their pool, paid 2, but still bounced i…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522237917857185912/1522237917857185912

**Thread:** Greenbelt Rampager
**Cards:** greenbelt rampager

## Summary
My opponent resolved greenbelt rampager with two energy in their pool, paid 2, but still bounced it to their hand.

### greenbelt rampager
When this creature enters, pay {E}{E} (two energy counters). If you can't, return this creature to its owner's hand and you get {E}.

---
<sub>report_id: `discord:1522237917857185912:1522237917857185912:0` · discord: `1522237917857185912/1522237917857185912`</sub>

tryeverything24:
Fixes #4955.

Greenbelt Rampager: "When this creature enters, pay {E}{E}. If you can't, return this creature to its owner's hand and you get {E}." always bounced the creature, even when the controller had 2+ energy and paid successfully.

## Root cause & fix

The generic "if you can't" rider lowers to `Not { ZoneChangedThisWay { Any } }`, a proxy populated only by effects that move an object between zones. `Effect::PayCost` deducts a resource and moves no object, so the ledger stayed empty regardless of payment success, making the proxy unconditionally true.

Extends `rewrite_cant_rider_for_non_zone_change_parent` (which already carved out this exact bug class for `Effect::TurnFaceUp`, Etrata's "if you can't") to also recognize `Effect::PayCost`, rewriting the rider to `Not { OptionalEffectPerformed }` — fed by `cost_payment_failed_flag`, which correctly distinguishes a paid cost from an unpaid one.

## Testing

Two discriminating tests: paid case stays on battlefield; unpaid case bounces to hand.

Verified locally (rebased onto main): fmt clean, clippy clean, 16672 lib tests pass, 3131 integration tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed payment results carrying over incorrectly between separate cost resolutions.
  * Corrected “if you can’t” effects after payments so they now respond to the actual payment outcome.
  * Fixed Greenbelt Rampager interactions, including energy payments, bouncing, rewards, and controller/owner handling.

* **Tests**
  * Added regression coverage for successful and failed payments, stale payment states, and Greenbelt Rampager scenarios.

### DIFF
--- a/crates/engine/src/game/effects/pay.rs
+++ b/crates/engine/src/game/effects/pay.rs
@@ -36,6 +36,19 @@ pub fn resolve(
         Effect::PayCost { cost, scale, payer } => (cost, scale, payer),
         _ => return Err(EffectError::MissingParam("PayCost".to_string())),
     };
+    // CR 118.12 + CR 608.2c: make the failure signal resolution-local to THIS
+    // `PayCost` instance. `cost_payment_failed_flag` is only ever SET on
+    // failure; without clearing it here, a stale `true` left by an earlier,
+    // unrelated resolution's unpayable cost would survive into this payment,
+    // suppress the `optional_effect_performed` seed for a payment that
+    // SUCCEEDED (`effects/mod.rs`'s mandatory-rider seed guards on
+    // `!cost_payment_failed_flag`), and make a `Not { OptionalEffectPerformed }`
+    // "If you can't" rider fire despite the cost having been paid (Greenbelt
+    // Rampager bouncing after a successful {E}{E} payment, issue #4955 review).
+    // Mirrors the established per-iteration clear in the repeated-payment
+    // driver ("clear the resolution failure flag before THIS payment so a
+    // prior iteration's failure can't be misread as this payment's outcome").
+    state.cost_payment_failed_flag = false;
     let Some(payer) = resolve_effect_player_ref(state, ability, payer_filter) else {
         state.cost_payment_failed_flag = true;
         return Ok(());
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**Request changes — cover controller/owner divergence in the Rampager runtime path.**

## 🟡 Finding

**[MED]** The regression coverage does not exercise a stolen Greenbelt Rampager where controller and owner differ. Why it matters: the affected `if you can't` behavior can still bind player-relative results to the owner rather than the current controller. Suggested fix: add a real cast/ETB runtime regression with a controller≠owner fixture and assert the cost/rider resolves for the controller.

matthewevans (changes requested):
**Request changes — make the `PayCost` rider outcome resolution-local.**

## 🔴 Blocker

**[HIGH]** A successful mandatory `PayCost` can inherit a failure from an earlier, unrelated resolution. Evidence: `crates/engine/src/game/effects/pay.rs:35-42` and `:250-272` set `cost_payment_failed_flag` only on failure; its `PaymentOutcome::Paid` path does not clear it. The new rewrite at `crates/engine/src/parser/oracle_effect/mod.rs:348-364` changes the rider to `Not { OptionalEffectPerformed }`, while `crates/engine/src/game/effects/mod.rs:8505-8521` only seeds that outcome when `!state.cost_payment_failed_flag`. Thus, after any prior unpayable cost, a later paid Rampager cost leaves the stale flag set, no outcome is seeded, and the newly rewritten "If you can't" bounce still runs. The added tests initialize a fresh state and do not exercise that preceding-failure path.

Suggested fix: make the success/failure signal for this `PayCost` instance resolution-local at the payment authority/chain handoff (clearing or replacing the stale global signal before the current mandatory payment), then add a runtime regression that first resolves an unpayable payment and subsequently pays the Rampager ETB cost successfully.

**Recommendation: request changes.**

tryeverything24:
Addressed the stale-flag blocker.

`pay::resolve` now clears `cost_payment_failed_flag` at this `PayCost` instance's chain handoff, making the success/failure signal resolution-local: the flag is only ever SET on failure, so without the clear a stale `true` from any earlier unpayable cost survived into a later successful payment, suppressed the mandatory-rider seed (guarded on `!cost_payment_failed_flag`), and fired the `Not { OptionalEffectPerformed }` bounce despite the cost having been paid. The clear mirrors the repeated-payment driver's established per-iteration semantics ("clear the resolution failure flag before THIS payment so a prior iteration's failure can't be misread as this payment's outcome").

Added the requested regression: a production-path integration test casts the same Rampager twice — the first cast's {E}{E} is unpayable (bounces, grants {E}, and leaves the global flag set), then, funded to exactly {E}{E}, the second cast pays successfully and must STAY on the battlefield with both energy spent and no rider {E}. Fails against the previous head. Also added a `pay.rs` unit test that a pre-set stale flag is replaced by a successful payment's outcome.

Verified locally in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17516 passed, engine integration 3753 passed. Rebased onto current `main` (merge-tree-verified clean against today's head).

matthewevans (commented):
**Current-head review: prior blockers are addressed; holding for the independent pass.**

## ✅ Clean

`pay::resolve` now clears the resolution-local failure flag before this `PayCost`, and the real Rampager recast regression proves a later successful payment does not inherit an earlier failure. The controller/owner regressions also exercise both payment ownership and the owner-hand bounce destination.

**Recommendation: hold until the current CI run and CodeRabbit review complete, then recheck before approval.**

matthewevans:
CodeRabbit's non-Energy coverage note is not a blocking gap. `pay::resolve` clears the outcome flag before every `Effect::PayCost`, and all non-interactive/non-X cost shapes (mana, life, speed, composite, discard) route through `resolve_ability_cost_payment`, which returns `Paid` or sets that same flag on `Failed`. The current fixture corpus contains no additional `pay … If you can't` card to exercise; the Greenbelt runtime tests cover both rider branches, a stale prior failure, and controller/owner divergence. No production change is warranted solely to duplicate the same authority path with a synthetic card.

matthewevans (approved):
Approved: the current head makes the payment outcome resolution-local and covers paid, unpaid, stale-state, and controller/owner runtime paths.

PR #5871, opened 2026-07-15
fix(parser): recognize plural "leave the battlefield" in the generic exile-until duration combinator

### LINKED ISSUE #4235: Cloak and Dagger, Entiwined — The card is asking to put cards onto the battlefield instead of exiling on playing
Reported in Discord: https://discord.com/channels/1485498006781427802/1518700522696085584/1518700522696085584

**Thread:** Cloak and Dagger, Entiwined
**Cards:** _none detected_
**Parser status:** has_gaps

## Summary
The card is asking to put cards onto the battlefield instead of exiling on playing

---
<sub>report_id: `discord:1518700522696085584:1518700522696085584:0` · discord: `1518700522696085584/1518700522696085584`</sub>

tryeverything24:
Fixes #4235.

Cloak and Dagger, Entwined: "...You may exile a nonland card from their hand or the chosen creature until Cloak and Dagger leave the battlefield."

## Root cause & fix

`parse_until_body` (the generic "until [source] leaves the battlefield" duration combinator) only matched the singular verb form "leaves the battlefield". Cloak and Dagger, Entwined's own name is a plural subject, so its printed oracle text uses plural agreement — "until Cloak and Dagger leave the battlefield" — which never matched. This left the exile sub-ability's `duration` field silently `None`, so no `ExileLinkKind::UntilSourceLeaves` link was ever created, and a card exiled by the ETB trigger never returned when Cloak and Dagger left the battlefield.

The sibling *trigger* detector for leaves-the-battlefield already handled both "leaves"/"leave" forms — only the *duration* combinator was missing the alternative. Fixed by mirroring that existing alternation.

## Testing

New regression test (AST-level + full runtime cast→reveal→exile→destroy-source→return-to-hand) — confirmed RED without the fix, GREEN with it.

Verified locally (rebased onto main): fmt clean, clippy clean, 16672 lib tests pass, 3129 integration tests pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved “until source leaves the battlefield” effects when players make interactive card choices.
  - Improved parsing for both singular and plural “leave the battlefield” wording.
  - Prevented unsupported alternative effects from being silently narrowed or misinterpreted.
  - Maintained choice-duration information across state updates and visibility filtering.

- **Tests**
  - Added coverage for interactive exile behavior, duration preservation, and the affected card text.

### DIFF
--- a/crates/engine/src/ai_support/candidates.rs
+++ b/crates/engine/src/ai_support/candidates.rs
@@ -6052,6 +6052,7 @@ mod tests {
             library_position: None,
             is_cost_payment: false,
             enters_modified_if: None,
+            duration: None,
         };
 
         let actions = candidate_actions_broad(&state);
--- a/crates/engine/src/game/costs.rs
+++ b/crates/engine/src/game/costs.rs
@@ -1226,6 +1226,7 @@ fn pay_ability_cost_inner(
                     library_position: None,
                     is_cost_payment: true,
                     enters_modified_if: None,
+                    duration: None,
                 };
                 return Ok(PaymentOutcome::Paused {
                     remaining_cost: None,
--- a/crates/engine/src/game/effects/attach.rs
+++ b/crates/engine/src/game/effects/attach.rs
@@ -318,6 +318,7 @@ fn prompt_resolution_attachment_choice(
                 library_position: None,
                 is_cost_payment: false,
                 enters_modified_if: None,
+                duration: None,
             };
             Ok(true)
         }
--- a/crates/engine/src/game/effects/blight.rs
+++ b/crates/engine/src/game/effects/blight.rs
@@ -89,6 +89,7 @@ pub fn resolve(
         library_position: None,
         is_cost_payment: false,
         enters_modified_if: None,
+        duration: None,
     };
 
     Ok(())
--- a/crates/engine/src/game/effects/bounce.rs
+++ b/crates/engine/src/game/effects/bounce.rs
@@ -239,6 +239,7 @@ pub fn resolve(
                     library_position: None,
                     is_cost_payment: false,
                     enters_modified_if: None,
+                    duration: None,
                 };
                 return Ok(());
             }
@@ -333,6 +334,7 @@ pub fn resolve(
                     library_position: None,
                     is_cost_payment: false,
                     enters_modified_if: None,
+                    duration: None,
                 };
                 return Ok(());
             }
@@ -512,6 +514,7 @@ pub fn resolve_all(
                 library_position: None,
                 is_cost_payment: false,
                 enters_modified_if: None,
+                duration: None,
             };
             return Ok(());
         }
--- a/crates/engine/src/game/effects/cast_from_zone.rs
+++ b/crates/engine/src/game/effects/cast_from_zone.rs
@@ -269,6 +269,7 @@ fn open_private_zone_cast_selection(
         library_position: None,
         is_cost_payment: false,
         enters_modified_if: None,
+        duration: None,
     };
     Ok(())
 }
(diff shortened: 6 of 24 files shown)


matthewevans (changes requested):
**Verdict: Changes requested**

## 🔴 Blocker

- **[HIGH] The newly parsed duration is lost on the ordinary multi-candidate path.** `crates/engine/src/game/engine_resolution_choices.rs` reconstructs `ChangeZoneIterationCtx` from `WaitingFor::EffectZoneChoice` with `duration: None`, as this PR's own module documentation notes. The runtime fixture deliberately removes every competing eligible card so it takes the single-candidate shortcut; it therefore cannot prove the advertised card works when an opponent has two nonland cards. Thread the duration through the choice/resume authority and add a two-eligible-card runtime test that selects one, then verifies the exile link returns it when Cloak and Dagger leaves.

- **[HIGH] This leaves coverage dishonest for the full Oracle sentence.** The parser still accepts Cloak and Dagger while dropping the `or the chosen creature` alternative, the secondary `up to one target creature` declaration it depends on, and the chosen-opponent hand binding. The test's direct zone mutation works around that last error rather than exercising the actual card. Preserve the clause as a strict unsupported marker until its target/anaphor binding is implemented, or complete the general engine/parser work in the same PR.

## 🟡 Non-blocking

- **[MED] The new `CR 603.2c` citation is unrelated.** That rule is about repeated triggering from one event; the verified duration rule is `CR 611.2a`. Correct the annotation while reworking the duration path.

## ✅ Clean

- The plural `leave` parser alternative is composed at the existing duration combinator and the parse-diff is limited to Cloak and Dagger.

**Recommendation: Complete the general choice-resume and Oracle-clause handling, then resubmit with a multi-candidate end-to-end regression.**

tryeverything24:
Addressed both blockers and the citation note.

**Blocker 1 — duration through the choice/resume authority**: `WaitingFor::EffectZoneChoice` now carries a serde-defaulted `duration` field, set from `ability.duration` at the `change_zone` pause site and consumed by the resume authority — `engine_resolution_choices` reconstructs `ChangeZoneIterationCtx` with the carried value instead of hardcoding `None` (the `PayCost` cost-exile reconstruction stays deliberately `None`: no "until …" idiom pays a cost). All construction sites threaded, including the visibility/server-core redaction rebuilds (the duration is a public effect parameter, not private hand info). New runtime regression: a bounded exile driven through the production `resolve_ability_chain` → `change_zone::resolve` → `EffectZoneChoice` → resume pipeline with TWO eligible candidates — asserts the prompt carries the duration, one card is selected, the `UntilSourceLeaves { return_zone: Hand }` link survives the round-trip, and the card returns to its owner's hand when the source is destroyed.

**Blocker 2 — coverage honesty**: the exile imperative now declines any "or the chosen <type>" anaphoric alternative (`parse_exile_ast` guard), so Cloak and Dagger's full sentence lowers to an explicit `Effect::Unimplemented` strict failure instead of a supported-looking exile that silently dropped the printed alternative — pinned by a full-card parser test. The plural-"leave" duration fix itself is pinned on a supported single-referent subset of the same idiom (same plural-name verb agreement). The runtime regression consequently drives the reviewed engine seam with a hand-built ability rather than the printed card, since the card is now honestly unsupported until the chosen-object anaphor is representable.

**Citation**: the duration combinator comment now cites CR 611.2a, replacing the unrelated CR 603.2c.

Verified locally in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17515 passed, engine integration 3751 passed. Rebased onto current `main`.

matthewevans:
**Maintainer status: reviewed current head `51dabfa2874505ba63ad4b8cf36f18d20943743d`; held pending current-head CI and a refreshed coverage parse-diff.**

The prior parse-diff was produced for the July 15 head, so it cannot evidence this update. The earlier duration choice/resume and coverage-honesty blockers were checked against this head; no approval or enqueue will occur until the current evidence is available.

matthewevans (approved):
**Approved:** Current-head review confirms duration survives the multi-candidate `EffectZoneChoice` resume path; the unsupported anaphoric alternative remains an explicit `Effect::Unimplemented`, and the refreshed parse diff is limited to Cloak and Dagger, Entwined.

PR #5872, opened 2026-07-15
fix(engine): bind Elrond's scry-X to the clamped look count

### LINKED ISSUE #1128: Elrond, Master of Healing - Ability is applied incorrectly — Elrond, Master of Healing: His ability is getting misinter…
Reported in Discord: https://discord.com/channels/1485498006781427802/1508715787513495592/1508715787513495592

**Thread:** Elrond, Master of Healing - Ability is applied incorrectly
**Cards:** elrond, master of healing
**Parser status:** has_gaps

## Summary
Elrond, Master of Healing: His ability is getting misinterpreted.

### elrond, master of healing
Whenever you scry, put a +1/+1 counter on each of up to X target creatures, where X is the number of cards looked at while scrying this way.
Whenever a creature you control with a +1/+1 counter on it becomes the target of a spell or ability an opponent controls, you may draw a card.

---
<sub>report_id: `discord:1508715787513495592:1508715787513495592:0` · discord: `1508715787513495592/1508715787513495592`</sub>

tryeverything24:
Fixes #1128.

Elrond, Master of Healing: "Whenever you scry, put a +1/+1 counter on each of up to X target creatures, where X is the number of cards looked at while scrying this way."

## Root cause & fix

No `QuantityRef` existed anywhere for "cards looked at while scrying," so the where-X binder fell through and the whole ability lowered to `Effect::Unimplemented`.

Added `QuantityRef::TriggeringScryLookCount`, backed by `GameState::last_scry_look_count`, set in `apply_scry_after_replacement_without_draw` at the point the requested scry amount is clamped to library size — mirroring the existing `QuantityRef::TriggeringDiscoverValue` precedent (Curator of Sun's Creation's "discover again for the same value") rather than inventing new machinery. Wired through the exhaustive `QuantityRef` match sites plus a new branch in `parse_where_x_quantity_expression`. The existing "up to X target creatures" multi-target machinery already consumed the new quantity with zero further changes.

Elrond's second ability ("becomes the target of a spell or ability an opponent controls" draw trigger) was already correct — verified with new combination test coverage, no fix needed there.

## Testing

New end-to-end test covering both abilities, including a clamped-X edge case (library smaller than the scry amount) and the opponent-targeting draw trigger (positive, negative, and self-cast-negative cases).

Verified locally (rebased onto main): fmt clean, clippy clean, 16673 lib tests pass, 3136 integration tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added support for referencing the number of cards actually looked at during a scry effect.
  * Scry-triggered abilities can now use that count when determining targets and effect values.
  * Scry action records now preserve the effective number of cards viewed.

* **Bug Fixes**
  * Improved handling of triggers created during consecutive or nested scry effects so they are not lost when resolution pauses.


matthewevans (changes requested):
**Verdict: changes requested**

## 🔴 Blocker

`last_scry_look_count` is a global mutable scalar, but the value in Elrond's "this way" clause belongs to each individual triggering scry event. A resolving ability that scries more than once can overwrite it before the queued triggers are constructed; then every Elrond trigger receives the final scry's count. `PendingTriggerContext` already preserves the individual `GameEvent` through target selection and stack resolution, so carry the effective look count on that event (or its typed action payload) and resolve `TriggeringScryLookCount` from the current trigger context instead. Add a production-pipeline regression with at least two scries in one resolution whose look counts differ; it must prove the two queued Elrond triggers expose different target-slot limits and fail with this patch reverted.

The new parser arm is also a verbatim full-Oracle match (`tag("the number of cards looked at while scrying this way")` plus `rest.is_empty()`). Split the grammar into composed nom pieces so the reusable quantity phrase and its scry-context suffix are independently represented, rather than adding a one-card literal dispatch.

## 🟡 Non-blocking

The three tests for Elrond's already-working second ability do not exercise the reported defect or a changed production seam, while the existing parser tests already cover its constituent filters. Please remove them from this bug-fix PR (or move them to a separately justified regression) so each added integration test earns its cost.

Gemini's suggested `CR 701.22b` correction is not applicable: the checked rules text defines 701.22b as *scry 0 produces no event*. Keep the actual citations tied to 701.22a (the scry action) and 701.22d (the post-scry trigger timing), with wording that does not claim a nonexistent clamping subrule.

## ✅ Clean

The parse-diff artifact is scoped to Elrond and Elvish Mariner, matching the new phrase; the current CI is green.

**Recommendation: rework the value as per-trigger event provenance, add the multi-scry regression, then request re-review.**

tryeverything24:
Addressed all four review points.

**Per-trigger event provenance (blocker)**: `GameState::last_scry_look_count` is removed entirely. The effective (post-clamp) look count now rides on the scry's own `PlayerPerformedAction` event (new serde-defaulted `look_count` field), and `QuantityRef::TriggeringScryLookCount` resolves from the CURRENT trigger's preserved event — `current_trigger_event` with the detection-time thread-local fallback, which `push_trigger_event_context` installs around `build_target_slots` — so each queued trigger reads its own scry's count at both announce (slot limits) and resolution.

**Composed grammar (blocker)**: the verbatim full-oracle `tag` is replaced by `parse_scry_look_count_ref` in `oracle_nom::quantity` — composed axes (count-of-cards subject, "looked at" participle, "while scrying" qualifier, reflexive "this way" suffix), registered ahead of the generic `parse_the_number_of` in `parse_quantity_ref`; the where-X binder delegates to it and keeps whole-expression ownership.

**Multi-scry regression**: added ("Scry 3.\nScry 1." in one resolution). With the patch reverted, the queued trigger's slot limit reads the LAST scry's count (1) instead of its own (3) — red. One honest finding pinned in the test rather than papered over: the engine currently queues ONLY the first scry's trigger — the second scry's event, emitted between the two interactive `ScryChoice` pauses of the same resolution, never produces a pending trigger at all (`deferred_triggers` stays at 1; no `OrderTriggers` prompt). That second-trigger loss is a separate, pre-existing pause/resume event-collection defect orthogonal to how the look count is carried; the test hard-asserts the current single-trigger behavior so a future fix surfaces there and the test can be upgraded to the full two-trigger shape. Happy to file it as its own issue.

**Scope + citations**: the three second-ability tests are removed from this PR as requested; citations stay on CR 701.22a/CR 701.22d with no claimed clamping subrule (701.22b not applied — it is the scry-0 no-event rule).

Verified locally in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17515 passed, engine integration 3750 passed. Rebased onto current `main`.

matthewevans (changes requested):
The current head still has two merge blockers.

1. The requested multi-scry proof is not achieved. The new regression intentionally asserts that the second Scry event does **not** trigger, which contradicts `Whenever you scry` and the PR's claimed multi-scry correctness. Repair pause/resume event collection and assert that both triggers independently expose 3 and 1 target slots. Evidence: `crates/engine/tests/integration/issue_1128_elrond_scry_counter.rs:171`.

2. `TriggeringScryLookCount` now reads the preserved trigger event, but `ability_rw` still describes it as reading a global scalar and returns `RwProfile::empty()`, while `ability_scan` returns `Axes::NONE`. Classify it consistently as a per-event/event-live dependency with `EventContextAmount` / current-trigger-event semantics, and add a focused classifier regression. Evidence: `crates/engine/src/game/ability_rw.rs:5666`; `crates/engine/src/game/ability_scan.rs:1795`.

tryeverything24:
Rebased onto current main. Conflicts were mechanical: main's new `resolve_and_apply_library_shuffle` authority in `game/library.rs` vs. this branch's `look_count` field on `PlayerPerformedAction` — kept main's structure and threaded `look_count: None` through the new construction sites.

**Multi-scry trigger loss.** Root cause: `run_post_action_pipeline` scans an action's events only when it settles to `Priority`. In a two-scry resolution, the second scry's `PlayerPerformedAction` is emitted while the first `ScryChoice` submission resumes the continuation, which then pauses on the second `ScryChoice` — a non-Priority settlement — so the event was dropped and its trigger silently lost. Fixed at the `ScryChoice` resume seam in `engine_resolution_choices.rs`: the resumed slice is now parked into `deferred_triggers` via the existing `park_observer_triggers_if_paused` (B2), draining once resolution truly settles. The regression in `issue_1128_elrond_scry_counter.rs` now asserts both triggers fire and independently expose their own scry's look count as target-slot limits (3, then 1), with disjoint targets proving one counter per trigger. Red on the unrepaired collection path — the second `TriggerTargetSelection` never surfaces and the walk runs into `DeclareAttackers` — green with the repair.

**Classifier reclassification.** `TriggeringScryLookCount` resolves from the current trigger's preserved event, so `ability_rw` now returns `reads_event_live()` (mirroring the `EventContextSourceModesChosen` / `TimesCostPaidThisResolution` twins; not one of the frozen D5 legacy tags, so no `legacy_batch_prompt`) instead of `RwProfile::empty()`, and `ability_scan` reports the event axis (`event: true`, mirroring `EventContextAmount`) instead of `Axes::NONE`. Focused classifier regressions added in both files.

Tests: engine lib 17732 passed / 0 failed; integration 4009 passed / 0 failed; fmt, engine-authorities and parser-combinator gates green. Parse-diff against main's published baseline remains scoped to Elrond, Master of Healing + Elvish Mariner.

matthewevans (approved):
Approved at `6a11bd3be800d2b258d3f99163aaae2c17319f53`.

The scry look count is event-scoped through the preserved triggering event; the composed parser is at the existing quantity seam; and the runtime regression drives two paused scries through distinct target-slot limits (3, then 1). Current required checks are green. Earlier change requests and bot findings are resolved on this head; the stale CR 701.22b claim is refuted by the checked local rules text.

PR #5873, opened 2026-07-15
fix(engine): rebind DamageAll recipient-filter source to the resolved damage target

### LINKED ISSUE #4759: Chandra's Ignition — bug 1 When you control multiple creatures, Chandra's Ignition prompts for two targets.
Reported in Discord: https://discord.com/channels/1485498006781427802/1521385320434765934/1521385320434765934

**Thread:** Chandra's Ignition
**Cards:** chandra's ignition, terror of the peaks
**Parser status:** has_gaps

## Summary
bug 1
When you control multiple creatures, Chandra's Ignition prompts for two targets.

### chandra's ignition
Target creature you control deals damage equal to its power to each other creature and each opponent.
### terror of the peaks
Flying
Spells your opponents cast that target this creature cost an additional 3 life to cast.
Whenever another creature you control enters, this creature deals damage equal to that creature's power to any target.

---
<sub>report_id: `discord:1521385320434765934:1521385320434765934:0` · discord: `1521385320434765934/1521385320434765934`</sub>

tryeverything24:
Fixes #4759.

Chandra's Ignition: "Target creature you control deals damage equal to its power to each other creature and each opponent."

## Root cause & fix

The reported "prompts for two targets" symptom is already fixed on `main` via an earlier, unrelated maintainer commit that added a generalized single-target suppression for `Effect::DamageAll` with `damage_source: Some(Target)`.

While building the required end-to-end regression test, found a real, separate bug: `DamageAll::resolve_all` built its recipient-matching `FilterContext` from `ability.source_id` (the spell object) instead of the resolved `DamageSource::Target` creature, so `FilterProp::Another` ("each OTHER creature") never actually excluded the damage source itself — the targeted/boosted creature took its own sweep damage.

Fixed by rebinding `ctx.source_id`/`source_controller` to the resolved target creature before the recipient filter runs, scoped strictly to the `DamageSource::Target` case (mirrors the equivalent fix already applied to the single-recipient `DealDamage` path in #5834/issue #4960).

## Testing

New end-to-end test: cast Chandra's Ignition with multiple creatures on the battlefield, confirm only one target prompt occurs, and confirm the source creature takes zero damage from its own sweep while other creatures and opponents take the full amount.

Verified locally (rebased onto main): fmt clean, clippy clean, 16673 lib tests pass, 3133 integration tests pass.

### DIFF
--- a/crates/engine/src/game/effects/deal_damage.rs
+++ b/crates/engine/src/game/effects/deal_damage.rs
@@ -1545,7 +1545,31 @@ pub fn resolve_all(
 
     // Collect matching object IDs.
     // CR 107.3a + CR 601.2b: ability-context filter evaluation.
-    let ctx = filter::FilterContext::from_ability(ability);
+    //
+    // CR 120.1 + CR 601.2c: When `damage_source` is `Some(Target)`, "other"/
+    // "another"-relative recipient filters (`FilterProp::Another` — "each
+    // OTHER creature") must be evaluated relative to the resolved damage
+    // SOURCE creature, not the ability's nominal source (the spell/permanent
+    // that generated this `DamageAll` effect). Left as `ability.source_id`,
+    // `FilterProp::Another` compares each battlefield candidate against the
+    // spell's own object id — which is never itself a battlefield creature —
+    // so the exclusion never actually fires and the chosen source creature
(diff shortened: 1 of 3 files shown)


matthewevans (approved):

PR #5876, opened 2026-07-15
fix(parser): accept literal thresholds in infix power/toughness comparisons (Wasp, Shrinking Savior)

### LINKED ISSUE #5874: Parser: infix "power/toughness less than N" with a literal threshold is dropped (Wasp, Shrinking Savior)
## Summary

A P/T comparison written in the **infix** form with a **literal** threshold — *"power less than 3"*, *"toughness greater than 4"* — fails to parse, so the filter (and any count built on it) silently loses the constraint. The **postfix** form (*"3 or less"*) and the infix form with a **dynamic** threshold (*"less than the number of cards in your hand"*) both work.

## Affected card

- **Wasp, Shrinking Savior** — *"Whenever Wasp attacks, up to one other target creature gets -3/-0 until your next turn. Then draw a card **for each creature with power less than 0** on the battlefield."*

## Current behavior

The static/effect parses, but *"for each creature with power less than 0"* produces `Draw { count: Fixed(1) }` — a flat **draw 1**, because the `power less than 0` filter is dropped (no `PtComparison`). The parser's swallow-audit flags a `DynamicQty` `SwallowedClause`. Wasp therefore always draws 1 card instead of N.

## Evidence (isolation)

| Phrase (target or for-each) | Parses `PtComparison` |
|---|---|
| `power 3 or less` (postfix, literal) | ✅ |
| `power less than the number of cards in your hand` (infix, dynamic) | ✅ |
| `power less than 3` (infix, **literal**) | ❌ |
| `toughness greater than 4` (infix, **literal**) | ❌ |

## Root cause

`parse_pt_infix_tail` (`oracle_nom/filter.rs`) parses the threshold with `parse_quantity_ref` (dynamic game-quantity references only), whereas the postfix tail uses `parse_quantity_expr_number` (literals + `X`). So an infix comparison against a bare number can't parse the threshold and the whole comparison fails.

## Expected behavior

The infix form accepts a literal number or `X` as the threshold, exactly like the postfix form.

## Rules

CR 208.1 (power/toughness) + CR 107.1 (strict `<`/`>` lowers to `≤ N-1` / `≥ N+1`).

joaovictor712:
Closes #5874.

## The bug

**Wasp, Shrinking Savior** — *"…draw a card for each creature with power less than 0 on the battlefield."* — parsed to a flat `Draw { count: Fixed(1) }`. The `power less than 0` filter was silently dropped, so Wasp always draws **1** card instead of one per creature with power < 0. The parser's swallow-audit flags a `DynamicQty` `SwallowedClause`.

## Root cause

An **infix** P/T comparison with a **literal** threshold doesn't parse. Isolation:

| Phrase | `PtComparison` parsed |
|---|---|
| `power 3 or less` (postfix, literal) | ✅ |
| `power less than the number of cards in your hand` (infix, dynamic) | ✅ |
| `power less than 3` (infix, **literal**) | ❌ |
| `toughness greater than 4` (infix, **literal**) | ❌ |

`parse_pt_infix_tail` parsed the threshold with `parse_quantity_ref` (dynamic game-quantity refs only), while the postfix tail (`3 or less`) uses `parse_quantity_expr_number` (literals + `X`). So an infix comparison against a bare number couldn't parse its threshold and the whole comparison failed.

## Fix

The infix tail now tries the dynamic ref first (unchanged behavior) and **falls back to the literal/X parser** — the same value grammar the postfix tail already uses. General across targets and counts, power and toughness, `less than`/`greater than`.

## Tests

- `test_parse_pt_comparison_infix_less_than_literal` — `"power less than 3"` → `LE` of `Offset(Fixed(3), -1)`.
- `test_parse_pt_comparison_infix_greater_than_literal` — `"toughness greater than 4"` → `GE` of `Offset(Fixed(4), +1)`.

## Verification

`cargo fmt --all` ✓ · engine lib suite **16,612 pass / 0 fail** + 2 new ✓ · `clippy -p engine --all-targets -D warnings` clean ✓

CR 208.1 + CR 107.1.

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -7206,6 +7206,70 @@ mod tests {
         );
     }
 
+    /// CR 208.1 + CR 107.1 — production-path RESOLUTION regression for the infix
+    /// literal-threshold parser fix (Wasp, Shrinking Savior). Parsing the full card
+    /// yields a draw whose count is `ObjectCount` over "creature with power less than
+    /// 0"; this test drives that exact parsed filter through the runtime matcher
+    /// against three battlefield creatures at power −1, 0, and +1 and proves that
+    /// ONLY the negative-power creature satisfies it — i.e. it alone contributes to
+    /// the resolved draw count. If the infix-literal alternative is removed the draw
+    /// count collapses to `Fixed(1)` (no `ObjectCount` filter to extract) and the
+    /// `expect` below fails, so this regression is tied to the parser change.
+    #[test]
+    fn wasp_draw_filter_counts_only_negative_power_creatures() {
+        use crate::types::ability::{AbilityDefinition, Effect, QuantityExpr, QuantityRef};
+        fn find_draw_filter(def: &AbilityDefinition) -> Option<TargetFilter> {
+            if let Effect::Draw {
+                count:
+                    QuantityExpr::Ref {
+                        qty: QuantityRef::ObjectCount { filter },
+                    },
+                ..
+            } = &*def.effect
+            {
+                return Some(filter.clone());
+            }
+            def.sub_ability.as_deref().and_then(find_draw_filter)
+        }
+
+        let parsed = crate::parser::parse_oracle_text(
+            "Whenever Wasp attacks, up to one other target creature gets -3/-0 until your next turn. Then draw a card for each creature with power less than 0 on the battlefield.",
+            "Wasp, Shrinking Savior",
+            &[],
+            &["Creature".to_string()],
+            &[],
+        );
+        let filter = parsed
+            .triggers
+            .iter()
+            .filter_map(|t| t.execute.as_deref())
+            .find_map(find_draw_filter)
+            .expect("Wasp's draw count must be a dynamic ObjectCount over a power<0 filter");
+
+        let mut state = setup();
+        let neg = add_creature(&mut state, PlayerId(0), "Negative");
+        let zero = add_creature(&mut state, PlayerId(0), "Zero");
+        let pos = add_creature(&mut state, PlayerId(0), "Positive");
+        state.objects.get_mut(&neg).unwrap().power = Some(-1);
+        state.objects.get_mut(&zero).unwrap().power = Some(0);
+        state.objects.get_mut(&pos).unwrap().power = Some(1);
+
+        // CR 208.1: "power less than 0" is strict, so power −1 matches while 0 and
+        // +1 do not — exactly one creature contributes to the resolved draw count.
+        assert!(
+            matches_target_filter(&state, neg, &filter, neg),
+            "power −1 must satisfy `power < 0`"
+        );
+        assert!(
+            !matches_target_filter(&state, zero, &filter, zero),
+            "power 0 must NOT satisfy `power < 0`"
+        );
+        assert!(
+            !matches_target_filter(&state, pos, &filter, pos),
+            "power +1 must NOT satisfy `power < 0`"
+        );
+    }
+
     // CR 702.171b: `IsSaddled` matches only objects with the saddled designation.
     #[test]
     fn is_saddled_property_matches_only_saddled() {
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -7208,28 +7208,31 @@ mod tests {
 
     /// CR 208.1 + CR 107.1 — production-path RESOLUTION regression for the infix
     /// literal-threshold parser fix (Wasp, Shrinking Savior). Parsing the full card
-    /// yields a draw whose count is `ObjectCount` over "creature with power less than
-    /// 0"; this test drives that exact parsed filter through the runtime matcher
-    /// against three battlefield creatures at power −1, 0, and +1 and proves that
-    /// ONLY the negative-power creature satisfies it — i.e. it alone contributes to
-    /// the resolved draw count. If the infix-literal alternative is removed the draw
-    /// count collapses to `Fixed(1)` (no `ObjectCount` filter to extract) and the
-    /// `expect` below fails, so this regression is tied to the parser change.
+    /// yields a draw whose count is an `ObjectCount` over "creature with power less
+    /// than 0"; this test drives that exact parsed count through the SHARED quantity
+    /// resolver (`resolve_quantity`, the same authority the Draw effect uses in the
+    /// production resolution pipeline) against three battlefield creatures at power
+    /// −1, 0, and +1, and asserts the concrete count is 1 — only the negative-power
+    /// creature contributes. The per-creature `matches_target_filter` checks pin the
+    /// discriminating boundary. If the infix-literal alternative is removed the draw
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**Verdict: Changes requested**

## 🔴 Blocker

Please add one discriminating integration test that exercises Wasp, Shrinking Savior through the production card/engine path and asserts its draw count uses the `power < 0` battlefield filter. The two new parser-unit cases prove the grammar branch, but do not prove that the card conversion and resolution pipeline retains the parsed filter; this regression must fail when the production parser change is reverted.

## 🟡 Non-blocking

Gemini raised a missing-CR concern, but the current head already documents the relevant comparison handling with CR 208.1 and CR 107.3a/107.1, so no annotation change is needed.

## ✅ Clean

The literal threshold is expressed by reusing `QuantityExpr` and the existing infix comparison path; the parse-diff is current and limited to Wasp, Shrinking Savior, and required CI is green.

**Recommendation: add the end-to-end regression test, then request re-review.**

matthewevans:
**Verdict: Changes requested — the implementation seam is correct, but the proof must cover the full grammar axis.**

## 🔴 Blocker

`crates/engine/src/parser/oracle_nom/filter.rs:351-359` correctly extends the shared infix P/T-tail authority rather than adding a caller-specific arm, but the new tests cover only strict fixed-number examples. Please add:

1. A production-path Wasp, Shrinking Savior regression with creatures at negative, zero, and positive power, proving only the negative-power creature contributes to the resolved draw count.
2. Parser cases for an inclusive literal (`less than or equal to 0`) and the newly admitted `X` threshold, with the full `QuantityExpr` shape asserted.

Those cases prove the actual `0` boundary, the optional comparison clause, and the expanded literal/X axis. They must fail if the new alternative is removed.

## 🟡 Non-blocking

I compared this with #4866. That PR required a separate static-restriction parse because its source-first wording had been routed to the inverse `CantBlock` behavior. This PR does not repeat that mistake: `parse_pt_comparison` is already the shared target/static filter authority, so extending its comparison tail is the right location.

## ✅ Clean

The change reuses `QuantityExpr` and preserves dynamic thresholds by trying `parse_quantity_ref` before the existing literal/X parser. Runtime filter evaluation already resolves `PtComparison` values through the shared quantity evaluator.

**Recommendation: add the three discriminating cases above, then request re-review.**

galuis116:
Thanks for the detailed review - all three points are addressed in cee8fd231.

**1. Production-path resolution regression (negative / zero / positive power).**
`game::filter::tests::wasp_draw_filter_counts_only_negative_power_creatures` parses the full Wasp, Shrinking Savior card through `parse_oracle_text`, extracts the draw count's `QuantityRef::ObjectCount { filter }`, and drives that exact parsed filter through `matches_target_filter` against three battlefield creatures at power -1, 0, and +1. It asserts only the -1 creature satisfies `power < 0`, so exactly one contributes to the resolved draw count. This is tied to the fix: removing the infix-literal alternative collapses the draw count to `Fixed(1)`, the `ObjectCount` extraction returns `None`, and the `.expect(...)` panics -- the test fails.

**2. Inclusive-literal + X threshold parser cases, full `QuantityExpr` asserted.**
- `test_parse_pt_comparison_infix_less_than_or_equal_literal`: `"power less than or equal to 0"` -> `PtComparison { Power, Current, LE, QuantityExpr::Fixed { value: 0 } }` (non-strict, no offset -- pins the `0` boundary and the optional "or equal to" clause on the literal axis).
- `test_parse_pt_comparison_infix_less_than_x`: `"power less than x"` -> `PtComparison { Power, Current, LE, QuantityExpr::Offset { inner: Variable("X"), offset: -1 } }` (the newly admitted X threshold).

Both assert the complete `QuantityExpr` shape and fail if the new `alt((parse_quantity_ref, parse_quantity_expr_number))` branch is reverted, since the literal/X value would no longer parse.

Rebased onto latest `main`; `cargo fmt`, full `engine` lib suite (16679 passed), and `clippy -p engine --tests` are green.

matthewevans (changes requested):
**Verdict: changes requested — the grammar coverage is now clean, but the requested production resolution proof is still missing.**

## 🔴 Blocker

`crates/engine/src/game/filter.rs:7212-7268` parses Wasp and calls `matches_target_filter` for three synthetic creatures, but it never resolves the parsed `QuantityExpr::Ref { qty: ObjectCount { .. } }` through the quantity authority or executes the Draw effect. That proves the parser retained a filter and the leaf matcher understands it; it does not prove Wasp's draw count is one in the production resolution pipeline. A defect in `game/quantity.rs`'s ObjectCount resolution (or in the Draw path's use of the count) would still pass this test.

Please extend the same fixture to invoke the parsed count through the shared quantity resolver and assert the concrete result is `1`, or add an integration scenario that resolves the Wasp trigger and asserts exactly one card drawn. Keep the `-1/0/+1` inputs; that is the discriminating boundary that proves the actual `power < 0` behavior.

## 🟡 Non-blocking

The new parser cases now cover strict literal, inclusive literal, and `X` thresholds with complete `QuantityExpr` assertions, which closes the grammar-axis feedback from the prior review.

## ✅ Clean

`oracle_nom/filter.rs:351-365` remains the shared infix P/T comparison seam, and the current parse-diff is limited to Wasp, Shrinking Savior. CI is green on this head.

**Recommendation: request changes — add the one resolver or end-to-end draw assertion, then re-request review.**

galuis116:
Addressed the resolution-pipeline blocker in 66eb7f7b5.

`wasp_draw_filter_counts_only_negative_power_creatures` now resolves the FULL parsed draw count -- `QuantityExpr::Ref { qty: ObjectCount { .. } }` -- through `game::quantity::resolve_quantity`, the same shared authority the Draw effect consults in production, and asserts the concrete result is `1`:

```rust
let resolved = resolve_quantity(&state, &count_expr, PlayerId(0), neg);
assert_eq!(resolved, 1, "the production quantity resolver must count exactly the one power<0 creature");
```

The `-1 / 0 / +1` inputs are unchanged, so the discriminating boundary is preserved. A defect in `game/quantity.rs`'s `ObjectCount` resolution (or in the count's use) now fails the test -- not merely a broken leaf matcher. The per-creature `matches_target_filter` checks remain as the boundary pin. Removing the infix-literal alternative still collapses the count to `Fixed(1)`, the `ObjectCount` extraction returns `None`, and the `.expect(...)` fails -- so the regression stays tied to the parser change.

Rebased onto latest `main`; `cargo fmt` and the targeted test are green.

matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

None.

## 🟡 Non-blocking

None.

## ✅ Clean

- The shared infix P/T-comparison parser now accepts the complete threshold grammar: dynamic quantities remain first, while literal values and X use the existing typed quantity parser.
- Strict versus inclusive boundaries are covered, and the Wasp regression verifies the parsed `ObjectCount` through the same target-aware quantity resolver used by `Effect::Draw`.
- The real parse diff is limited to Wasp, Shrinking Savior; the current branch is rebased onto `main` and includes the requested CR 208.1 annotation.

**Recommendation: approved and queued for merge.**

PR #5879, opened 2026-07-15
fix(parser): split imperative sibling sentence after a token's quoted ability (Alien Invasion)

### LINKED ISSUE #5844: [Card Bug] Alien Invasion: Creates an alien token for every invasion counter
### Card(s)

Alien Invasion

### Build/version

0.25.0

### Game mode

AI

### Actual behavior

Alien Invasion creates an Alien token for every invasion counter on it

### Expected behavior

It should create 1 Alien token every upkeep phase, with +1/+1 tokens on it for every invasion counter on the enchantment

### Steps to reproduce

- Cast Alien Invasion
- Wait until 2nd upkeep phase

_No response_

joaovictor712:
## Problem

Alien Invasion (#5844) reads:

> At the beginning of combat on your turn, create a 1/1 red Alien creature token with haste and "This token attacks each combat if able." Put a +1/+1 counter on it for each invasion counter on this enchantment, then put an invasion counter on this enchantment.

The engine created **one Alien token per invasion counter** and the "+1/+1 counter â¦ for each invasion counter" pump was **missing entirely**.

## Root cause

The clause splitter (`split_clause_sequence`) closes a sentence after a token/permanent's quoted granted ability (a span ending in `."`) only when the following sentence starts with a whitelisted continuation (`then` / `if` / `otherwise` / `may have` / emblem head). A fresh **imperative** sibling sentence was not recognized, so the whole remainder was swallowed into the token-creation clause:

- its `for each invasion counter` was mis-read as the token **count** â `Token { count: CountersOn(invasion) }` (one token per counter), and
- the `+1/+1 counter on it` pump was dropped.

Isolation probe: with the quoted ability removed, the same card parses correctly (`Token { count: 1 }` + `PutCounter { P1P1, CountersOn(invasion), LastCreated }` + `PutCounter { invasion, 1, SelfRef }`) â confirming the quoted-ability sentence boundary is the sole trigger.

## Fix

Extend `quote_closes_sentence_before_sequence` to also close the sentence when the continuation begins with an imperative game-action verb, via a new `starts_imperative_action_continuation` helper. It reuses the shared clause-starter verb table (`starts_clause_text_lower`) but first excludes the pronoun/determiner subject starters it also recognizes (`it` / `that` / `this` / `those` / `they` / `you` / `all` / `each` / `the`). Those, after a granted quote, are anaphors the token-creation path resolves inline ("It becomes a 2/2 â¦", "The token is goaded"), so they stay attached; only a bare imperative verb marks a genuine sibling effect.

This is a **class-level** fix: every `create a token with "<ability.>" <imperative on the token>` card benefits, not just Alien Invasion.

## Tests

- `sequence.rs`:
  - `quoted_grant_splits_before_imperative_sibling_sentence` â the Alien Invasion remainder splits off.
  - `quoted_grant_keeps_anaphoric_subject_continuation` â an "It gains â¦" anaphor stays attached.
  - Existing guardrails still pass: `quoted_grant_keeps_nonrecognized_capitalized_continuation` ("The token is goaded") and `quoted_grant_splits_before_following_sentence` (Requiem Monolith "may have").
- `oracle_effect/tests.rs`:
  - `alien_invasion_creates_one_token_and_pumps_it_per_counter` parses the full card and asserts (1) the token count is a literal `1`, (2) the "+1/+1 counter â¦ for each invasion counter" pump survives as a `PutCounter` on the last-created token, (3) the enchantment still accrues one invasion counter.

Full `engine` lib suite green (16,682 passed, 0 failed; snapshot tests included); `cargo fmt` and `clippy -p engine --tests` clean.

Closes #5844

### DIFF
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -1444,15 +1444,30 @@ fn quote_closes_sentence_before_sequence(current: &str, remainder: &str) -> bool
 /// `quote_closes_sentence_before_sequence`); only a bare imperative verb marks a
 /// genuine sibling effect here.
 fn starts_imperative_action_continuation(remainder_lower: &str) -> bool {
-    let first_word = remainder_lower.split_whitespace().next().unwrap_or("");
-    // Anaphoric-subject / determiner starters — keep attached (existing behavior).
-    if matches!(
-        first_word,
-        "it" | "its" | "that" | "this" | "those" | "they" | "you" | "all" | "each" | "the"
-    ) {
-        return false;
-    }
-    starts_clause_text_lower(remainder_lower)
+    !starts_anaphoric_subject(remainder_lower) && starts_clause_text_lower(remainder_lower)
+}
+
+/// True iff `remainder_lower` begins with a pronoun/determiner subject word — the
+/// anaphoric continuations ("It becomes …", "The token is goaded", "That creature
+/// gains …") that must stay attached to a granted quote rather than split as a
+/// sibling imperative. Each starter carries a trailing space so only a complete
+/// word matches (e.g. "it " does not fire on "item"), mirroring the whitespace
+/// convention of `starts_clause_text_lower`'s verb tags.
+fn starts_anaphoric_subject(remainder_lower: &str) -> bool {
+    alt((
+        tag::<_, _, OracleError<'_>>("it "),
+        tag("its "),
+        tag("that "),
+        tag("this "),
+        tag("those "),
+        tag("they "),
+        tag("you "),
+        tag("all "),
+        tag("each "),
+        tag("the "),
+    ))
+    .parse(remainder_lower)
+    .is_ok()
 }
 
 fn parse_search_exile_name_suffix(input: &str) -> Result<(&str, ()), nom::Err<OracleError<'_>>> {
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -7926,9 +7926,11 @@ mod tests {
             "create a 1/1 red Alien creature token with haste and \"This token attacks each combat if able.\"",
             "the token + quoted ability must be its own chunk: {chunks:?}"
         );
+        // allow-noncombinator: structural test assertion over the pre-tokenized chunk vector, not parser dispatch
+        let counter_clause_split_off = chunks.len() >= 2
+            && chunks[1].starts_with("Put a +1/+1 counter on it for each invasion counter"); // allow-noncombinator: same, on the `.starts_with` line
         assert!(
-            chunks.len() >= 2
-                && chunks[1].starts_with("Put a +1/+1 counter on it for each invasion counter"),
+            counter_clause_split_off,
             "the imperative counter sentence must split into its own chunk: {chunks:?}"
         );
     }
--- a/crates/engine/src/parser/oracle_effect/sequence.rs
+++ b/crates/engine/src/parser/oracle_effect/sequence.rs
@@ -1403,6 +1403,22 @@ fn quote_closes_sentence_before_sequence(current: &str, remainder: &str) -> bool
     {
         return true;
     }
+    // CR 111.3 + CR 608.2c (issue #5844): a token/permanent granted-ability quote
+    // that ends a sentence can be followed by a fresh IMPERATIVE sibling sentence
+    // that acts on the created object — Alien Invasion: `… and "This token
+    // attacks each combat if able." Put a +1/+1 counter on it for each invasion
+    // counter on this enchantment, then …`. Without splitting here the whole
+    // remainder is swallowed into the token-creation clause: its "for each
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**Request changes — parser dispatch must remain composable nom grammar.**

## 🟡 Finding

**[MED]** `crates/engine/src/parser/oracle_effect/sequence.rs:1447` splits the Oracle text with string operations in parser dispatch. Why it matters: this bypasses the repository's nom parser authority and makes phrase variation brittle. Suggested fix: express the clause split with `alt(tag(...))`, following the existing `starts_clause_text_lower` combinator pattern.

galuis116:
Addressed in a7c0c6eb5.

Replaced the `.split_whitespace()` + `matches!` dispatch with a `starts_anaphoric_subject` combinator built from `alt((tag(...)))`:

```rust
fn starts_anaphoric_subject(remainder_lower: &str) -> bool {
    alt((
        tag::<_, _, OracleError<'_>>("it "),
        tag("its "), tag("that "), tag("this "), tag("those "),
        tag("they "), tag("you "), tag("all "), tag("each "), tag("the "),
    ))
    .parse(remainder_lower)
    .is_ok()
}
```

Each starter carries a trailing space so only a complete word matches (e.g. `it ` does not fire on `item`), mirroring the whitespace convention in `starts_clause_text_lower`'s verb tags. `starts_imperative_action_continuation` is now `!starts_anaphoric_subject(..) && starts_clause_text_lower(..)` -- pure combinator dispatch, no string-method scanning or verbatim string equality.

Behavior is unchanged: the splitter tests (imperative sibling splits, anaphoric subject stays attached) and the guardrails (`quoted_grant_keeps_nonrecognized_capitalized_continuation`, `quoted_grant_splits_before_following_sentence`) all still pass. Rebased onto latest `main`.

matthewevans (changes requested):
**Request changes — do not classify non-anaphoric clause starters as token anaphors.**

**[MED]** `crates/engine/src/parser/oracle_effect/sequence.rs:1456-1467` treats `you`, `all`, and `each` as anaphoric subjects. Evidence: the shared authority `starts_clause_text_lower` explicitly recognizes `all `, `each ` / `each opponent `, and `you ` as clause starters at `sequence.rs:2113-2128` and `2164-2166`, but `starts_imperative_action_continuation` rejects them before consulting that authority. Why it matters: a sentence after a granted-ability quote such as `"…" You gain …` or `"…" Each opponent …` is a fresh sibling instruction, not an anaphor to the created object; it remains glued to the token clause and can be swallowed or mis-lowered, recreating the class of bug this PR is intended to fix.

Please distinguish actual last-created anaphors from independent subject-led clauses, rather than filtering by the first word alone. Add direct splitter coverage that those `You …` and `Each …` continuations split, alongside the existing `It …` non-split guard. The nom conversion itself correctly resolves the earlier parser-dispatch finding.

galuis116:
Addressed in 78b3f0224 -- you're right, `you` / `all` / `each` are not anaphors.

**The guard is now restricted to genuine last-created anaphors:**

```rust
fn starts_anaphoric_subject(remainder_lower: &str) -> bool {
    alt((
        tag::<_, _, OracleError<'_>>("it "),
        tag("its "), tag("that "), tag("this "), tag("those "), tag("they "),
    ))
    .parse(remainder_lower)
    .is_ok()
}
```

`you `, `each `/`each opponent `, and `all ` now fall through to the `starts_clause_text_lower` authority and split, as they should -- they are subject-led sibling instructions, not references to the created object.

I also dropped `the `: it is not a `starts_clause_text_lower` starter, so "The token is goaded" already fails the authority check and stays attached without an explicit guard (that guardrail test still passes) -- the entry was dead weight that only obscured the intent.

**Splitter coverage added** -- `quoted_grant_splits_before_independent_subject_continuation` asserts the exact chunk split for all three independent-subject forms, alongside the existing `It gains ...` non-split guard:

| continuation | expected |
|---|---|
| `"..." You gain 3 life.` | splits |
| `"..." Each opponent loses 2 life.` | splits |
| `"..." All creatures you control get +1/+1 until end of turn.` | splits |
| `"..." It gains haste until end of turn.` | stays attached (guard) |

Full `engine` lib suite green (16,707 passed, 0 failed) -- widening the split surfaced no regressions. `cargo fmt`, clippy, and the parser-combinator gate are clean. Rebased onto latest `main`.

matthewevans (approved):
**Approve — the parser guard is resolved and current-head evidence is complete.**

## ✅ Clean

- `crates/engine/src/parser/oracle_effect/sequence.rs:1446-1481` now excludes only genuine token anaphors before delegating to the shared clause-starter authority; `You`, `Each`, and `All` remain independent sibling instructions.
- `crates/engine/src/parser/oracle_effect/sequence.rs:7930-8005` supplies direct positive and negative splitter coverage, while `oracle_effect/tests.rs:46488-46577` verifies Alien Invasion lowers to one token plus the intended last-created counter effect.
- The current-head parse-diff is fresh (updated after `dff6a7b`) and its 13-card / 20-signature delta is consistent with this generalized quote-sentence boundary. Required CI is green.

PR #5883, opened 2026-07-15
fix(parser): parse subtype-gated token substitution replacements (Fisher's Talent #5636)

### LINKED ISSUE #5636: Fisher's Talent: levels 2 and 3 are silent no-ops — 'create X instead' replacements parse to inert Spell-kind Token abilities
## Summary
**Confirmed silent defect in shipped card-data.json** (surfaced by the U8 Class→DocEmitter conversion, task #8; predicted by ISSUES.md #1 and verified live now that `check_swallowed_clauses` runs in `lower_oracle_ir` on the Class path).

Both are textbook replacement effects (CR 614.1a: "instead" effects are replacement effects).

## Actual parse (shipped)
`replacements: 0`, plus two inert `kind:"Spell"` Token abilities. A Spell-kind ability on a permanent never fires — **levels 2 and 3 of this card do nothing in shipped data**, silently (the card still counts as supported).

## Fix shape
Route the "If you would create [token], create [token] instead" clause class through the replacement-effect parser (the `Replacement_Instead` swallow detector already fires on it — the warning exists, the routing doesn't). Class-card context: the clause arrives via level sections, now emitted at printed source lines after task #8's conversion.

## Class
Supported-but-misparsed (the silent class). Sibling of #5631 in severity: zero Unimplemented nodes, invisible to coverage gates.

Yurii214:
Closes #5636.

## Problem
Fisher's Talent's levels 2/3 — "If you would create a Fish token, create a 3/3 blue Shark creature token instead" (and Shark→Octopus) — parsed to `replacements: 0` plus two inert `Spell`-kind Token abilities. A Spell-kind ability on a permanent never fires, so **levels 2 and 3 silently did nothing** in shipped data (the card still counted as supported).

## Fix
`parse_replacement_line` had no arm for the **subtype-gated one-for-one token substitution** shape "if you would create a `<subtype>` token, create `<token>` instead". The existing token-creation arms cover doubling/additional shapes (Doubling Season "a token" / "one or more tokens"; Xorn "instead create those tokens plus"; Manufactor "instead create one of each") but not substitution.

Add `parse_subtype_token_substitution`, composing two existing, runtime-supported mechanisms:
- the **Xorn subtype gate** (`ReplacementCondition::TokenSubtypeMatches`) so it fires only for the named subtype, and
- the **Divine Visitation substitution payload** — the substitute `Effect::Token` carried in `execute`; the applier (`game/replacement.rs`) swaps only the token's characteristics and keeps the proposed event's count, so N `<subtype>` tokens become N substitute tokens.

The arm requires the first-person **"you would create"** antecedent so the `token_owner_scope(You)` is provably correct (an "if an effect / an opponent would create … instead" line is a different scope and is not captured). `oracle_class.rs` wraps the result with the class-level gate, so the swap is active only at the level that grants it.

No new engine types or variants — a pure parser addition.

## Verification
- New unit test `subtype_gated_token_substitution_fishers_talent` drives `parse_replacement_line` and asserts the `CreateToken` event, `TokenSubtypeMatches([Fish])` condition, `token_owner_scope(You)`, exact `Mandatory` mode, and the 3/3 blue Shark substitute — plus negative cases proving it does **not** steal the Manufactor comma-list shape or a non-substitution "instead" effect.
- `cargo test -p engine --lib` (new test) — green.
- Card-data regen: Fisher's Talent now emits **two** Mandatory `CreateToken` substitution replacements (Fish→Shark, Shark→Octopus); the L2→L3 cascade sequences correctly (create-Fish → Shark → Octopus) through the replacement loop.
- `cargo clippy -p engine --lib -- -D warnings` — clean.
- CR 614.1a / CR 111.1 / CR 109.5 annotated.

Model: claude-opus-4-8[1m]

### DIFF
--- a/crates/engine/src/parser/oracle_replacement.rs
+++ b/crates/engine/src/parser/oracle_replacement.rs
@@ -746,7 +746,10 @@ fn parse_replacement_line_inner(text: &str, card_name: &str) -> Option<Replaceme
     // by the "create <spec> instead" one-for-one swap tail (no "plus" / "one of
     // each"); the Manufactor block above already claimed the "instead create one
     // of each" variant.
-    if nom_primitives::scan_contains(&lower, "you would create a ")
+    // "you would create a" is a prefix of both the "a " and "an " articles (and of
+    // "another …", which the parser's exact frame then rejects), so the guard
+    // covers vowel-starting subtypes ("an Elf token") too.
+    if nom_primitives::scan_contains(&lower, "you would create a")
         && nom_primitives::scan_contains(&lower, "instead")
     {
         if let Some(def) = parse_subtype_token_substitution(&lower, &text) {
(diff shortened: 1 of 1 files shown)


Yurii214:
Addressed the vowel-starting-subtype finding in 4dc26b4: the frame now consumes `you would create ` + `alt(("a ", "an "))` and the fast-path guard covers both articles, so "an Elf/Insect/Octopus token" parses too. No shipped card uses the "an" antecedent today, so this is class completeness (no card-data change); added a regression test for the "an <vowel-subtype>" form.

matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- None.

## ✅ Clean

- The parser lowers to the existing `CreateToken` / `TokenSubtypeMatches` replacement authority rather than adding new replacement machinery.
- The current head resolves the `a`/`an` sibling feedback, and the four Fisher's Talent parse changes exactly match the intended two replacement levels replacing the former bare token effects.

**Recommendation: Merge via the queue.**

PR #5884, opened 2026-07-15
fix(engine): parse and resolve Nether Spirit's dropped self-reanimation intervening-if

rykerwilliams:
## Summary

Nether Spirit (`{1}{B}{B}`, Tempest): "At the beginning of your upkeep, if this card is the only creature card in your graveyard, you may return this card to the battlefield."

The intervening-if was dropped entirely by the parser (`condition: null` in card-data.json). `split_trigger` splits Oracle text at the first unrecognized comma; since "At the beginning of your upkeep" has no comma, the whole "if this card is the only creature card in your graveyard, you may return this card to the battlefield" fell through to the effect side untouched — there was no combinator in `parse_remaining_state_presence_conditions` recognizing "`~` is the only `<type>` [card] in `<zone>`". That also left `trigger_zones` stuck at the structural default `["Battlefield"]` (making the trigger undetectable while the card sits in the graveyard) and `ChangeZone.origin` null.

- Adds `parse_source_is_only_type_in_zone` (`oracle_nom/condition.rs`), composed entirely from existing building blocks (`parse_source_self_token`, `parse_type_phrase`, an `ObjectCount == 1` `QuantityComparison`). When the type-phrase filter carries a non-battlefield zone, wraps the condition in `StaticCondition::And{[SourceInZone{zone}, quantity]}` so the existing, unmodified `trigger_condition_source_zones` walker auto-derives `trigger_zones`/`ChangeZone.origin` — **zero runtime code changed**. Precedent: Jocasta, Automaton Avenger already drives the identical `And{SourceInZone, ...}` derivation end-to-end.
- Removes a stale `game/coverage.rs` exclusion that was silencing the semantic auditor for this exact bug (confirmed via card-data.json scan: the excluded substring matches exactly this one card).

CR 603.4 (intervening-if, checked at trigger time and re-checked at resolution), CR 113.6 / CR 113.6b (ability functions off-battlefield only from the zone it states).

## Test plan

- [x] 3 new parser unit tests (`condition.rs`): graveyard form → `And[SourceInZone(Graveyard), ObjectCount==1]`; battlefield form → bare `QuantityComparison` (no spurious `SourceInZone`); rejection guard (no "only" → parse error)
- [x] Parser-shape test (`oracle_trigger_tests.rs`): asserts `condition.is_some()`, `trigger_zones == [Graveyard]`, `ChangeZone.origin == Some(Graveyard)` for Nether Spirit's verbatim Oracle text
- [x] 6 integration tests (`nether_spirit_only_creature_card_intervening_if.rs`) driving the real `process_triggers`/`stack::resolve_top` pipeline: does-not-fire with 2 creature cards, positive reach-guard, fires-and-returns when sole, decline leaves it in graveyard, CR 603.4 resolution-time recheck fizzles when a 2nd creature card enters in response, two-copies-neither-fires
- [x] Live-revert-and-rerun proof: disabling the new combinator's registration reproduces failure in the parser-shape test and 4/6 integration tests (2 negative-guard tests pass vacuously as expected, guarded by the positive reach-guard test)
- [x] `cargo fmt --all`, parser combinator gate, CR-annotation gate all pass
- [x] `cargo clippy -p engine --lib --tests -- -D warnings` clean
- [x] Full `cargo test -p engine --lib` (16687+ tests) and `--test integration` green after rebase onto current main

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -9967,8 +9967,6 @@ fn line_has_condition_text(lower: &str) -> Option<&'static str> {
             || lower.contains("had no cards in hand")
             // "if no permanents left the battlefield" — turn-event check
             || lower.contains("no permanents left")
-            // "if [this card is] the only creature card in your graveyard" — zone state check
-            || lower.contains("only creature card in your graveyard")
             // "if you discarded a card this turn" — turn-event action check
             || lower.contains("if you discarded")
             // "if 4 or more damage was dealt" — turn-event damage check
--- a/crates/engine/src/parser/oracle_nom/condition.rs
+++ b/crates/engine/src/parser/oracle_nom/condition.rs
@@ -275,6 +275,11 @@ fn parse_remaining_state_presence_conditions(input: &str) -> OracleResult<'_, St
         parse_there_exists_compound_zone_condition,
         parse_there_exists_condition,
         parse_subject_first_zone_count,
+        // CR 603.4 + CR 113.6b: "~ is the only <type> [card] in <zone>" —
+        // self-referential count-equals-one gate (Nether Spirit). Anchored on
+        // the self-token + literal " is the only ", so it never mis-claims the
+        // non-self "a <type> card is in a graveyard" phrase below.
+        parse_source_is_only_type_in_zone,
         // CR 611.3a + CR 702: "a <type> card [with <keyword>] is in a graveyard"
         // — graveyard-presence gate for conditional continuous statics
         // (Tarmogoyf, Cairn Wanderer). Guarded by the "is in a graveyard"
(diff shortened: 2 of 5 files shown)


matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

None.

## 🟡 Non-blocking

None.

## ✅ Clean

- The condition is parsed at the existing state-condition authority and uses the established `ObjectCount == 1` plus `SourceInZone` representation, allowing existing trigger-zone and return-origin derivation to do the runtime work.
- The current parse diff is exactly one card, Nether Spirit: condition, active trigger zone, and `ChangeZone` origin all move from absent/default state to the intended graveyard semantics.
- The registered integration suite reaches trigger collection, the intervening-if fire-time and resolution-time checks, the optional choice, and the graveyard-to-battlefield move. The reviewer-requested direct remainder return is included in the current head.

**Recommendation: approved and queued for merge.**

PR #5885, opened 2026-07-15
fix(parser): accept literal thresholds in infix power/toughness comparisons (Wasp, Shrinking Savior #5874)

### LINKED ISSUE #5874: Parser: infix "power/toughness less than N" with a literal threshold is dropped (Wasp, Shrinking Savior)

Yurii214:
Closes #5874.

## The bug
An **infix** P/T comparison with a **literal** threshold — *"power less than 3"*, *"toughness greater than 4"* — fails to parse, so the filter (and any count built on it) is silently dropped. The **postfix** form (*"3 or less"*) and the infix form with a **dynamic** threshold (*"less than the number of cards in your hand"*) both already work.

**Wasp, Shrinking Savior** — *"…draw a card for each creature with power less than 0 on the battlefield"* — collapses to a flat `Draw { count: Fixed(1) }`, always drawing **1** instead of one per creature with power < 0.

## Root cause
`parse_pt_infix_tail` read the threshold with `parse_quantity_ref` (dynamic game-quantity refs only), while the postfix and exact tails read it with `parse_quantity_expr_number` (literals + X). So an infix comparison against a bare number couldn't parse its threshold and the whole comparison failed.

## Fix
Factor the threshold grammar into a shared combinator `parse_pt_threshold` — *dynamic ref, else literal/X* — and use it in the infix tail. The dynamic ref is tried first, so prior infix behaviour on a dynamic threshold is byte-for-byte preserved; a bare number / X now falls back to the same value grammar the postfix and exact tails already use. Strict `<`/`>` still lower to `LE`/`GE` with a ∓1 offset (CR 107.1: P/T are integers, so "less than N" ≡ "≤ N−1"). The threshold is now an axis independent of the comparison form — general across power/toughness, less-than/greater-than, strict/or-equal-to.

Scoped to the infix tail only; the postfix/exact tails are unchanged (routing them through the union would silently widen them to accept dynamic thresholds — out of scope).

## Tests
**Parser (leaf grammar):**
- `test_pt_infix_literal_threshold_full_grammar_axis` — power/toughness × less/greater-than × strict/or-equal-to, all with literal thresholds, asserting the exact lowered `Comparator` + `Offset`.
- `test_pt_infix_dynamic_threshold_still_parses` — the dynamic-ref path is preserved (no regression).

**Production-path resolution (full grammar axis):**
- `pt_infix_literal_threshold_resolves_through_production_path_full_axis` — parses real *"…draw a card for each creature with `<comparison>`…"* cards through `parse_oracle_text`, extracts the Draw's `ObjectCount`, and resolves it through `resolve_quantity` (the same shared authority the Draw effect uses in production) against one fixed battlefield of creatures A–E at power/toughness (-1,3)(0,5)(1,4)(2,5)(3,2). Case 1 is **Wasp, Shrinking Savior** through its exact printed text (→ 1); the others span the axis — `≤ 1` power → 3, strict `> 4` toughness → 2, `≥ 4` toughness → 3. Remove the fallback and every count collapses to `Fixed(1)` (no `ObjectCount` to extract), so the proof is tied to the fix.

`cargo clippy -p engine --lib -- -D warnings` — clean. CR 107.1 annotated.

Model: claude-opus-4-8[1m]

### DIFF
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -5823,6 +5823,112 @@ mod tests {
         id
     }
 
+    /// CR 208.1 + CR 107.1 + CR 122.1 — production-path RESOLUTION proof for the
+    /// infix literal-threshold parser fix (#5874), across the full grammar axis.
+    /// Each case parses a real `... draw a card for each creature with <comparison>
+    /// on the battlefield` clause through `parse_oracle_text`, extracts the Draw's
+    /// `ObjectCount`, and resolves it through `resolve_quantity` — the same shared
+    /// authority the Draw effect consults in the production resolution pipeline —
+    /// against one fixed battlefield. It spans the axis the leaf grammar tests
+    /// cannot reach on their own: power AND toughness, `less than` AND `greater
+    /// than`, strict AND `or equal to`. Case 1 is the reported card, Wasp,
+    /// Shrinking Savior, through its exact printed text. Remove the infix-literal
+    /// fallback and every `<stat> less/greater than <N>` filter is dropped, so each
+    /// count collapses to `Fixed(1)` (no `ObjectCount` to extract) and the
+    /// `expect` below fails — this regression is tied to the parser change.
+    #[test]
+    fn pt_infix_literal_threshold_resolves_through_production_path_full_axis() {
+        use crate::game::quantity::resolve_quantity;
+        use crate::types::ability::{AbilityDefinition, Effect, QuantityExpr, QuantityRef};
+
+        fn draw_object_count(def: &AbilityDefinition) -> Option<QuantityExpr> {
+            if let Effect::Draw { count, .. } = &*def.effect {
+                if matches!(
+                    count,
+                    QuantityExpr::Ref {
+                        qty: QuantityRef::ObjectCount { .. }
+                    }
+                ) {
+                    return Some(count.clone());
+                }
+            }
+            def.sub_ability
+                .as_deref()
+                .and_then(draw_object_count)
+                .or_else(|| def.else_ability.as_deref().and_then(draw_object_count))
+        }
+
+        // Fixed battlefield: (power, toughness) chosen so every axis case has a
+        // discriminating boundary. Object 0 doubles as the resolution `source`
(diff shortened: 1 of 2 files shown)


Yurii214:
Closing — #5876 landed the same infix literal-threshold fix and resolves #5874 on main. Thanks @matthewevans; no need for a duplicate. (The full-grammar-axis production-resolution test here was the delta, but the merged fix already covers the case.)

PR #5886, opened 2026-07-15
fix(parser): scope spell-cast triggers to a required spell keyword (#4754)

### LINKED ISSUE #4754: Slitherwisp triggers when casting counterspell — Slitherwisp should only trigger when I cast another spell with flash.
Reported in Discord: https://discord.com/channels/1485498006781427802/1521282669407899851/1521282669407899851

**Thread:** Slitherwisp triggers when casting counterspell
**Cards:** _none detected_

## Summary
Slitherwisp should only trigger when I cast another spell with flash.

---
<sub>report_id: `discord:1521282669407899851:1521282669407899851:0` · discord: `1521282669407899851/1521282669407899851`</sub>

jaytbarimbao-collab:
Fixes #4754

## Problem

Slitherwisp reads "Flash\nWhenever you cast **another spell that has flash**, you draw a card and each opponent loses 1 life." The spell-cast trigger dropped the "that has flash" clause: for the payload `spell that has flash`, `parse_spell_qualifier_payload` → `parse_post_spell_modifier("that has flash")` returned `None`, so the parse fell back to `parse_type_phrase`, which consumed "spell" and discarded "that has flash" as leftover. The trigger's `valid_card` ended up with only `FilterProp::Another`, so it fired on **every** other spell — the reported bug: casting a counterspell (no flash) wrongly triggered Slitherwisp.

## Fix

Add a `parse_post_spell_modifier` arm that recognizes a `"that has <keyword>"` / `"with <keyword>"` suffix and emits `FilterProp::WithKeyword { value }`. It is gated on a **fully consumed recognized keyword** (via `parse_keyword_line_core`, checking the remainder is empty), so:

- the numeric/colored qualifiers handled by earlier arms — `"with mana value N"`, `"with {X} in its mana cost"`, `"with one or more <color> mana symbol"` — keep their arms (they run first and return early), and
- a non-keyword tail (`"with power 3 or greater"`, `"with an {X}"`) still declines here and falls through unchanged.

The `Another` prop is still appended by the caller's `add_another_prop`, so Slitherwisp's filter becomes `[WithKeyword(Flash), Another]`. The object matcher already honors `FilterProp::WithKeyword` via `obj.has_keyword` (`game/filter.rs`), so **no runtime change is required** — the existing `valid_card_matches` → `target_filter_matches_object` path now filters cast spells by keyword.

## Tests

- **Parser** (`slitherwisp_cast_another_flash_spell_scopes_to_flash`): the trigger's `valid_card` carries both `WithKeyword(Flash)` and `Another`.
- **Runtime** (`slitherwisp_flash_spell_cast_trigger.rs`, 2 cases): drives the real cast pipeline and asserts Slitherwisp's observable effect (each opponent loses 1 life + controller draws) fires **only** for a flash spell — a plain spell leaves opponent life and hand untouched. This proves the fix is not merely parse-correct-but-inert.

CR references: 702.8a (flash), 603.2 (trigger events).

### DIFF
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -14713,6 +14713,33 @@ pub(crate) fn parse_post_spell_modifier(modifier: &str) -> Option<TargetFilter> 
         }
     }
 
+    // CR 702.8a + CR 603.2: "that has <keyword>" / "with <keyword>" — the spell
+    // must possess the named keyword ability (Slitherwisp: "another spell that
+    // has flash"). Without this the keyword clause is dropped and the trigger
+    // over-fires on every spell (e.g. a non-flash counterspell). Gated on a
+    // recognized keyword whose text is fully consumed by `parse_keyword_line_core`
+    // so the numeric/colored "with mana value …" / "with … mana symbol" qualifiers
+    // handled above keep their arms and a non-keyword tail ("with power 3 …")
+    // still declines here. Emits `FilterProp::WithKeyword`, which the object
+    // matcher honors via `obj.has_keyword` (filter.rs).
+    if let Ok((keyword_text, ())) = alt((
+        value((), tag::<_, _, OracleError<'_>>("that has ")),
+        value((), tag("with ")),
+    ))
+    .parse(modifier)
+    {
+        if let Some((keyword, remainder)) =
+            super::oracle_keyword::parse_keyword_line_core(keyword_text.trim())
+        {
+            if remainder.trim().is_empty() {
+                return Some(TargetFilter::Typed(
+                    TypedFilter::default()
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
**Verdict: Approved**

## 🔴 Blocker

- None.

## 🟡 Non-blocking

- Gemini's `value`-combinator suggestion is not actionable: this parser module already uses `value((), tag(...))` to normalize alternative prefixes, which is the same explicit discard pattern used here.

## ✅ Clean

- The keyword predicate is added at the post-spell-modifier parser seam and reuses the strict all-consuming keyword parser.
- The eight-card parse diff is the intended shared grammar class, with no unexplained cards.
- The Slitherwisp integration tests use the real cast/trigger pipeline and discriminate both the allowed Flash and rejected non-Flash branches.

**Recommendation: Merge via the queue.**

PR #5890, opened 2026-07-16
fix(ai): make cycling decisions deliberate

ntindle:


matthewevans (changes requested):
**Verdict: changes requested — the cycling policy is at a plausible decision seam, but its score contract and PR scope need correction before it can enter the AI evaluation path.**

## 🔴 Blocker

`crates/phase-ai/src/policies/cycling_discipline.rs:20-25` hard-codes `-1.0` and `-2.0`, then returns them from `:48-55`. The project policy contract requires card-equivalent scoring magnitudes to come from `AiConfig::policy_penalties` and be registered for tuning/explicit non-tuning; otherwise the new policy has an untracked production weight. `activation()` at `:35-41` is also intentionally constant but lacks the required `activation-constant` rationale marker. Move the two values into the configured policy-penalty authority, register them with their calibration rationale, and add the marker if the policy should remain globally active.

## 🟡 Non-blocking

`crates/phase-ai/src/tactical_gate.rs:802-889` adds an attacking-only-removal regression that is unrelated to Cycling/Typecycling and changes no production code in this PR. Please split it out so this AI policy diff stays reviewable and its paired-seed result can be attributed to cycling behavior.

## ✅ Clean

`effective_activated_ability` correctly uses the engine’s effective ability index, so runtime-granted Typecycling does not fall back to the printed ability list. The root-scoring tests also exercise finite cycling, pass comparison, and the needed-land case.

**Recommendation: route cycling scores through the configured policy-penalty contract and remove the unrelated tactical-gate test, then request re-review after the paired-seed AI gate reports for that focused head.**

matthewevans (changes requested):
**Changes requested — this adds a cycling policy while bypassing the scoring contract and changing unrelated tactical search behavior.**

## 🔴 Blocker

- `crates/phase-ai/src/policies/cycling_discipline.rs:23-27` introduces raw `-1.0`/`-2.0` policy constants and passes them directly to `PolicyVerdict::preference/strong` at `59-67`. Policy scoring is required to use the repository's configured band/weight contract; these literals make cycling behavior independent of the policy configuration and cannot be calibrated with the rest of the tactical policies.
- The PR also changes shared tactical ranking/deadline behavior in `crates/phase-ai/src/search.rs:1947-2075` and `crates/phase-ai/src/tactical_gate.rs`, outside the cycling policy seam. Current CI is terminal red: Rust tests shard 2 failed in workflow run 29462496610 (the aggregate Rust gate failed with it). Keep that regression separate from a focused cycling policy change.

Please move cycling weights onto the established configuration/band helpers, restore or split the unrelated tactical/search changes, and provide a focused regression that fails on the prior cycling decision while preserving the existing tactical behavior.

## ✅ Clean

- The policy correctly identifies the existing Cycling ability tag rather than matching card text.

**Recommendation: request changes — narrow this to a config-backed cycling policy and resolve the current Rust regression before re-review.**

matthewevans (approved):
**Approved.** The current head moves cycling penalties to the configured policy contract, confines the change to the cycling decision seam, and adds root-pipeline regressions for ordinary Cycling, Typecycling, and the sole-needed-land case. The prior Gemini hand-count finding is incorporated at `cycling_discipline.rs:86-92`.

PR #5893, opened 2026-07-16
Fix AI self-protection life spending

ntindle:
## Summary

- require a concrete, answerable threat before the AI spends life, cards, or permanents on self-protection
- reuse engine targeting/effect binding and runtime activated-ability definitions instead of card-name or printed-slot special cases
- model exact pending combat-damage participation, including first strike/double strike snapshots, infect/wither plus deathtouch, and Aura targeting

## Root cause

The reactive self-protection policy treated broad defensive context as sufficient payoff. For Arco-Flagellant, that made the AI pay 3 life for indestructible once per turn even when nothing threatened it.

The policy now uses an exact tri-state payoff classifier: confirmed payoff scores the activation, confirmed absence rejects it, and genuinely unsupported cases fail open without inventing value. Pure self-protection spell casts are gated separately so Arco-Flagellant itself remains castable.

## Review follow-up

- preserve fail-open behavior when an activation's primary/sub-ability chain contains modal or `else_ability` branches whose selected outcome is not yet known
- cover both a valuable draw mode and a valuable draw `else_ability` at the activation-verdict level, preventing the protection primary from hard-rejecting either shape

## Impact

Arco-Flagellant no longer burns life proactively. The same generic guard applies to other self-protection activations with life, discard, or land-sacrifice costs while preserving legitimate responses to removal and lethal combat damage.

## Validation

All checks pass on corrected head `c9d9a036d`:

- paired-seed AI gate
- decision-cost performance gate
- Rust formatting, clippy, parser gates, and both test shards
- card-data generation, schema validation, coverage, and parse-diff
- frontend lint, type-check, and tests
- WASM, Tauri, lobby-worker, and security checks

The repository pre-commit parser gates also passed locally.


matthewevans (changes requested):
**Blocked: a valuable alternate activation branch can be rejected solely because the primary self-protection branch has no immediate payoff.**

## 🔴 Blocker

`crates/phase-ai/src/policies/reactive_self_protection.rs:64-74` detects self-protection by calling `collect_definition_effects`, which includes modal and `else_ability` branches. `crates/phase-ai/src/policies/self_protection_classify.rs:527-539` then evaluates only `collect_chain_effects`, whose implementation in `crates/phase-ai/src/ability_chain.rs:18-25` walks the primary effect and `sub_ability` chain only. An activated ability with a primary indestructible branch plus a valuable alternate mode/else branch reaches `Some(false)` when no threat exists and is hard-rejected at `reactive_self_protection.rs:79`.

Preserve the existing fail-open rule for mixed/unsupported activation shapes: either evaluate the same complete branch set consistently or return neutral whenever mode/else branches are present. Please add an activation-level regression with a no-threat protection primary branch and an independently valuable alternate branch; it should be selectable after the fix and fail if the current hard-reject path is restored.

## ✅ Clean

The current `ebae47c` CI check set is green, and the required parse-diff artifact reports no card-parse changes. The runtime granted-ability index lookup and combat-step participant snapshot use the intended engine authorities.

**Recommendation: request changes for the modal/else activation classifier gap, then re-run the AI gate on the corrected head.**

matthewevans (approved):
Approved after current-head review.

The alternate modal/else activation branches now fail open before the exact self-protection classifier, with policy-path regressions that reject again if that guard is removed. The runtime activated-ability lookup and combat first/double-strike snapshot use the engine authorities; current required CI is green.

PR #5894, opened 2026-07-16
fix(ai): preserve lone optional sacrifice source

ntindle:
## Summary

- teach `SacrificeValuePolicy` to decline an optional sacrifice-for-one-card effect when the source is the only eligible sacrifice and costs more than the card drawn
- use the typed effect/filter AST and the engine's target-filter matcher rather than card-name or Oracle-text special cases
- stand down when there is other sacrifice fodder, the effect explicitly sacrifices itself, or the ability has a richer branch/repetition/player scope

## Root cause

Very Easy and Easy do not search through optional-effect continuations. The accept and decline candidates therefore received equal tactical scores, and high-temperature selection could accept Wilhelt's end-step trigger. Once accepted, the engine correctly auto-selected Wilhelt when it was the only legal Zombie, so the existing sacrifice-selection policy was reached too late.

## Impact

The AI now preserves Wilhelt (and the general class of valuable filtered sacrifice sources) when the source is the only eligible permanent and the payoff is only one card. It can still accept when another eligible permanent is available.

## Validation

- pre-push `cargo fmt --check`
- workspace Clippy with warnings denied
- parser architecture gates passed
- engine/parser tests: 583 passed
- `phase-ai` tests: 1282 passed, 8 ignored
- card-data schema validation: 35,397 cards parsed
- coverage regression: 0 engine regressions, 0 honesty regressions
- frontend ESLint: 0 errors (18 existing warnings)
- frontend protocol/type check passed
- paired-seed `cargo ai-gate`: 0 failures, 2 passes, 1 non-blocking favorable warning


matthewevans (changes requested):
**Blocked — the new optional-sacrifice policy rejects valid sacrifices at every difficulty, beyond its stated no-continuation-search scope.**

## 🔴 Blocker

**[HIGH]** `crates/phase-ai/src/policies/sacrifice_value.rs:19` returns `PolicyVerdict::reject` without checking difficulty. The policy registry converts every rejection to `NEG_INFINITY` at `crates/phase-ai/src/policies/registry.rs:482`, so Medium, Hard, and CEDH cannot evaluate a favorable sacrifice with a death trigger, recursion, or a copied/granted source ability. Why it matters: the intended limitation is only that VeryEasy/Easy cannot search the continuation; this globally disables valid search-capable decisions. Suggested fix: restrict the hard veto to configurations without continuation search, or use a bounded score that preserves evaluation on search-capable difficulties; add a Hard/CEDH non-veto regression.

**Recommendation: request changes; retain the beginner-difficulty guard without hard-vetoing the full policy surface.**

matthewevans (approved):
**Approved — the current head confines the optional-sacrifice veto to configurations that cannot search the continuation.**

Verified against `8155f94d30b8c1e042abc128e3e9245a77e86a65`:

- `sacrifice_value.rs:145-155` skips the hard reject whenever `search.enabled`, so Medium through CEDH retain continuation evaluation; the direct regression at `:335-356` fails against the prior unconditional veto.
- The full production-path fixture at `:279-293` drives `choose_action` for VeryEasy/Easy; sibling fixtures at `:297-331` preserve an alternate sacrifice and explicit self-sacrifice.
- The policy runs through the existing optional-choice `ActivateAbility` routing, filters via the engine target-filter authority, and only scans the battlefield after the narrow optional one-sacrifice/one-draw shape check.

All required GitHub checks are green at this head. No unresolved actionable review feedback remains.

PR #6021, opened 2026-07-16
fix(parser): compose Class-level trigger conditions instead of overwriting them (#5638)

### LINKED ISSUE #5638: Intermediate Chirography: live Duration_ThisTurn swallow warning, untriaged — 'died under your control this turn' condition
## Summary
Untriaged swallowed-clause warning, live in today's card-data.json (surfaced during task #8's ISSUES.md sweep — the `Duration_ThisTurn` swallow detector fires now that `check_swallowed_clauses` runs on the Class path). Filed per the campaign's harvest directive; needs triage to confirm or clear.

## What's known
`Duration_ThisTurn` swallow warning fires — the "this turn" scoping of the intervening-if condition ("a modified creature died under your control this turn") is the likely subject. Consequence untraced: the condition may be evaluated without the this-turn window (wrong more often than not), or the parse may be correct with a spurious warning.

## Triage ask
Trace the level-3 trigger's parsed condition; classify as (a) silent misparse (condition missing the this-turn window or the modified/died-under-control predicates) → fix, or (b) correct parse → tune the detector. State what the card parses to, not just the warning.

nghetienhiep:
fix(parser): compose Class-level trigger conditions instead of overwriting them (#5638)

parse_class_oracle_text gated level-2+ triggers by unconditionally
assigning trigger.condition = ClassLevelGE, which silently dropped any
printed intervening-if already parsed onto the trigger. Intermediate
Chirography's level-3 ability ("if a modified creature died under your
control this turn") lost its death check entirely and always created
a token at end step. Compose with TriggerCondition::And instead,
mirroring the existing static/replacement class-level wrappers.

Closes #5638

### DIFF
--- a/crates/engine/src/parser/oracle_class.rs
+++ b/crates/engine/src/parser/oracle_class.rs
@@ -352,6 +352,10 @@ fn parse_class_level_trigger(line: &str, card_name: &str, level: u8) -> Option<T
 fn wrap_trigger_with_class_level(trigger: &mut TriggerDefinition, level: u8) {
     let level_cond = TriggerCondition::ClassLevelGE { level };
     trigger.condition = Some(match trigger.condition.take() {
+        Some(TriggerCondition::And { mut conditions }) => {
+            conditions.insert(0, level_cond);
+            TriggerCondition::And { conditions }
+        }
         Some(existing) => TriggerCondition::And {
             conditions: vec![level_cond, existing],
         },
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Approved: the Class-level gate now composes with each printed intervening-if at the parser authority, and the current parse-diff is limited to the four expected Class triggers.

PR #6022, opened 2026-07-16
fix(engine): don't offer an optional counter removal with zero counters (Sun Droplet #4776)

### LINKED ISSUE #4776: Sundroplet — [[Sundroplet]] is allowing its owner to gain life even if it doesnt have any counters on it
Reported in Discord: https://discord.com/channels/1485498006781427802/1521787457799196732/1521787457799196732

**Thread:** Sundroplet
**Cards:** _none detected_

## Summary
[[Sundroplet]] is allowing its owner to gain life even if it doesnt have any counters on it

---
<sub>report_id: `discord:1521787457799196732:1521787457799196732:0` · discord: `1521787457799196732/1521787457799196732`</sub>

shin-core:
Fixes #4776 (claimed there). Verified the card (Sun Droplet) on Scryfall and reproduced end-to-end before building.

## The bug
Sun Droplet: *"At the beginning of each upkeep, you may remove a charge counter from this artifact. If you do, you gain 1 life."* With **zero** charge counters, the player was still prompted to remove one; accepting recorded `OptionalEffectPerformed = true` and fired the `GainLife` rider — **life from nothing**.

## Root cause
The up-front optional-effect feasibility gate `optional_effect_is_infeasible` (`effects/mod.rs`) only special-cased `Effect::PutChosenCounter`; every other optional effect — including `RemoveCounter` — fell through to `_ => false` (always feasible). So the "you may" was offered even when nothing could be removed, and its "if you do" rider then fired.

## The fix
Extend the gate to recognize a `RemoveCounter` whose resolved target object(s) hold **zero** matching counters. The new `counters::remove_counter_optional_is_infeasible` reuses the resolver's own `resolve_defined_or_targets` authority (so feasibility and resolution can't diverge) and honors `RemoveCounter`'s `Option<CounterType>` — `Some(kind)` matches that specific counter, `None` matches any. When infeasible, the up-front prompt is suppressed and its rider does not fire (CR 122.1: removing an absent counter does nothing; CR 603.12 + CR 608.2c: the "if you do" gates on the action occurring). This covers the whole *"you may remove a counter. If you do, X"* class — not just Sun Droplet.

## A note on the reproduction
While building the test I initially wrote Sun Droplet's two abilities on one line and saw only the `DamageReceived` trigger parse (no upkeep trigger) — which would have made this a parser issue. Tracing it, the parser produces **both** triggers correctly when the abilities are newline-separated (how card-data actually stores them); the single-line form was a test artifact. So the upkeep trigger DOES fire in production, and this engine gate is the operative fix. The test uses the real newline-separated text.

## Tests (integration, real upkeep-trigger pipeline, verbatim Oracle text)
- **zero counters**: the optional prompt is NOT surfaced and life is unchanged. **Revert-probed** — restoring the old `_ => false` re-exposes the phantom prompt/lifegain and this test fails.
- **with two counters** (positive reach-guard): the prompt IS offered; accepting removes exactly one counter (2 → 1) and gains 1 life — proving the gate doesn't over-suppress a feasible choice.

## Verification
Full engine lib: **16733 passed, 0 failed**. Integration: **3173 passed, 0 failed**. Clippy `-D warnings` clean; fmt clean. CR annotations (122.1, 603.12, 608.2c) grep-verified.

### DIFF
--- a/crates/engine/src/game/effects/counters.rs
+++ b/crates/engine/src/game/effects/counters.rs
@@ -1488,17 +1488,27 @@ pub fn resolve_multiply(
     Ok(())
 }
 
-/// CR 118.12 + CR 601.2f (by analogy) + CR 122.1: True when a `RemoveCounter`
-/// effect used as an optional "you may" gate cannot be performed because none of
-/// its resolved target object(s) hold a matching counter — "you may remove a
-/// charge counter from this artifact" with zero charge counters (Sun Droplet).
+/// CR 608.2d + CR 118.12 + CR 122.1: True when a `RemoveCounter` effect used as
+/// an optional "you may" gate cannot be performed — none of its resolved target
+/// object(s) hold a matching counter that is *permitted to be removed*. "You may
+/// remove a charge counter from this artifact" with zero charge counters (Sun
+/// Droplet), or with the counter present but frozen by a `CountersCantBeRemoved`
+/// static (Fear of Sleep Paralysis class).
 ///
-/// Removing a counter that isn't there is doing nothing, so the up-front "you
-/// may" must not be offered (and its `EffectOutcome::OptionalEffectPerformed`
-/// rider — "If you do, gain 1 life" — must not fire). Delegates to the same
-/// `resolve_defined_or_targets` authority the resolver uses, so feasibility and
-/// resolution never diverge. Returns `false` (feasible) for any non-`RemoveCounter`
-/// effect so the caller's other arms are unaffected.
+/// CR 608.2d: a player can't choose an impossible option. Removing a counter that
+/// isn't there — or one an effect forbids removing — does nothing, so the
+/// up-front "you may" must not be offered. CR 118.12: the `EffectOutcome::
+/// OptionalEffectPerformed` rider ("If you do, gain 1 life") checks whether the
+/// player chose to perform the action; if the action is impossible the choice is
+/// never offered, so the rider must not fire.
+///
+/// Both the presence check and the removal-prohibition check use the resolver's
+/// own authorities (`resolve_defined_or_targets`, `counter_removal_blocked`), so
+/// feasibility and resolution can never diverge. A `count > 1` request is still
+/// feasible whenever ≥1 permitted counter exists — the resolver removes as many
+/// as available (CR 122.1 "as much as possible"), a nonzero action. Returns
+/// `false` (feasible) for any non-`RemoveCounter` effect so the caller's other
+/// arms are unaffected.
 pub(crate) fn remove_counter_optional_is_infeasible(
     state: &GameState,
     ability: &ResolvedAbility,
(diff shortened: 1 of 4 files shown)


matthewevans (changes requested):
**Blocked — the new feasibility gate still offers an optional counter removal that a static effect makes impossible, allowing its “if you do” rider to fire without a removal.**

## 🔴 Blocker

**[HIGH]** The gate only tests the stored counter count at `crates/engine/src/game/counters.rs:1506`, while actual removal is rejected by `counter_removal_blocked` at `counters.rs:951`. `OptionalEffectPerformed` is set before resolution at `crates/engine/src/game/mod.rs:2074`, so accepting a removal prohibited by `CountersCantBeRemoved` makes a dependent “if you do” rider resolve even though no counter was removed. Why it matters: Fear of Sleep Paralysis-class effects can again grant life from nothing. Suggested fix: require at least one matching counter that is not removal-blocked (and, for an unspecified type, any permitted type), then add a runtime optional-removal-with-prohibition regression.

## 🟡 Non-blocking

**[LOW]** `counters.rs:1491` cites `CR 601.2f` by analogy, but that rule is spell-cost determination rather than a resolution-time optional action. Suggested fix: replace it with directly applicable CR evidence, such as `CR 608.2d` and `CR 118.12` after verification.

**Recommendation: request changes; make feasibility respect removal prohibitions and cover the runtime effect-performed boundary.**

shin-core:
Pushed `a88d8815` addressing both review points.

**[HIGH] removal prohibitions (matthewevans):** `remove_counter_optional_is_infeasible` now requires at least one matching counter that is NOT `counter_removal_blocked` (reusing the resolver's own authority — the same `CountersCantBeRemoved` check `remove_counter_with_replacement` enforces at `counters.rs:953`). A present-but-frozen counter is now correctly an impossible option, so the up-front prompt is suppressed and the "if you do" rider can't fire without a removal. Added a **runtime regression** (`sun_droplet_counter_present_but_removal_blocked_no_prompt_no_lifegain`): a `CountersCantBeRemoved(charge)` static freezes the present counters → no prompt, no lifegain, counters intact. **Revert-probed** — dropping the `counter_removal_blocked` check re-exposes the prompt and flips the test.

**[LOW] CR citation (matthewevans):** replaced the by-analogy `CR 601.2f` (spell-cost determination) with the directly-applicable **CR 608.2d** ("can't choose an option that's illegal or impossible") + **CR 118.12** (the "if you do" clause checks whether the player chose the action, not what events occurred) — both grep-verified against `docs/MagicCompRules.txt`.

**Gemini [MEDIUM] multi-count:** the gate keys on "≥1 permitted counter exists", not "count counters exist" — a `count > 1` removal stays feasible whenever at least one permitted counter is present, and the resolver removes as many as available (CR 122.1 "as much as possible"). So a request to remove 3 with only 2 present is still correctly offered (it's a nonzero action), and one with 0 permitted is suppressed.

Full lib 16733 / integration 3174 green, clippy clean.

matthewevans (approved):
**Approved — the current head closes the optional-counter-removal feasibility gap.**

The gate now excludes counters whose removal is prohibited by the resolver's `counter_removal_blocked` authority, and the integration regression reaches the real optional-decision path to prove no prompt, no rider life gain, and no counter removal. The multi-count review concern is not applicable: the resolver clamps a requested removal to available counters, so one permitted counter makes a nonzero removal feasible.

Issue #92, opened 2026-04-24
Having some problems while playing the game


VinRubio:
Hello, I’m experiencing several issues that are making the game unplayable:

- I am constantly encountering an "Engine connection lost" error. Because of this being so constant, I give up trying to play the game.

<img width="598" height="367" alt="Image" src="https://github.com/user-attachments/assets/71bd3736-28e6-4460-98f2-3cbbade33d8d" />

- Playing Commander with 4 players (myself and 3 AIs) is nearly impossible. Each action takes minutes to process, I timed it, and the AI took almost 9 minutes to complete a single turn.

- I don't know if this is intentional or a bug, but cards are often rendering incorrectly on the battlefield (see the image), though they occasionally appear as intended, so I think it's a bug.

<img width="346" height="210" alt="Image" src="https://github.com/user-attachments/assets/5dc3310a-40bd-456f-a571-d14e46844f67" />

I wasn't sure if you wanted individual bug reports here, but I can create separate issues for these later if preferred. Here are the card-specific issues I’ve found:

- Tataru Taru, when this card enters the battlefield, the caster draws cards, but the target opponent is never given the option to draw.
- Y'shtola, Night's Blessed, this card triggers every end step regardless of whether a player lost 4 life.
- Brokers Ascendancy, Instead of putting a +1/+1 counter on each creature and planeswalker, the game asks me to choose a single creature to receive a counter.
- Atraxa, Praetor's Voice, same thing as Brokers, instead of proliferating, it's asking me to choose a creature to receive a counter.

I'm using Chrome.

Thank you for all the hard work.

VinRubio:
I did some tests regarding the visual problem with the cards, and I found that it's because of my monitor (ultrawide), when I resize the window, the cards start to appear correctly.

matthewevans:
I appreciate the report! I'm doing significant work on improving the AI responsiveness. Thank you for the report and support :)

matthewevans:
@VinRubio can you try now with the latest build and see if that helps at all for you? I have an additional AI optimization in the works as well

VinRubio:
> [@VinRubio](https://github.com/VinRubio) can you try now with the latest build and see if that helps at all for you? I have an additional AI optimization in the works as well

Thank you so much for all your hard work.

I ran some tests and no longer encountered the “engine connection lost” issue. The visual glitch with the cards has also been fixed. 

[...]

PR #6027, opened 2026-07-16
fix(engine): Sidisi, Regent of the Mire — bind sacrifice-cost X and reanimate at X+1 mana value

### LINKED ISSUE #5957: Sidisi, Regent of the Mire — After checking with the Alt key and not seeing anything about being unsupported I thought…
Reported in Discord: https://discord.com/channels/1485498006781427802/1525260931452371115/1525260931452371115

**Thread:** Sidisi, Regent of the Mire
**Discord thread id:** `1525260931452371115`
**Discord message id:** `1525260931452371115`
**Cards:** _none detected_
**Parser status:** has_gaps

## Summary
After checking with the Alt key and not seeing anything about being unsupported I thought I'd post this.

## Reported text

After checking with the Alt key and not seeing anything about being unsupported I thought I'd post this.

Despite having the board state to be able to activate Sidisi's ability (tap: Sac a creature return one from the graveyard that costs 1 mana more. I have a 3 drop on board and a 4 drop in the GY)
 I am unable to activate her ability.

I have attached the exported debug log as the guidelines instruct.

## Attachments

- [game-state-turn-20-2026-07-10T21-56-27-280Z.zip](https://cdn.discordapp.com/attachments/1525260931452371115/1525260931611758735/game-state-turn-20-2026-07-10T21-56-27-280Z.zip?ex=6a5954f0&is=6a580370&hm=e2c3cf78e4f00372d88d993727915d0e9b67a025a99f9b3333a0d7e375facc9b&)

---
<sub>report_id: `discord:1525260931452371115:1525260931452371115:0` · discord: `1525260931452371115/1525260931452371115`</sub>

nghetienhiep:
## Summary

Sidisi, Regent of the Mire (`{T}, Sacrifice a creature you control with mana
value X other than Sidisi: Return target creature card with mana value X
plus 1 from your graveyard to the battlefield. Activate only as a sorcery.`)
was silently misparsed and, separately, could not actually be activated.

- The parser dropped the `plus 1` when it hit `parse_quantity_expr_number`
  (the filter-threshold quantity parser), so the reanimation target's mana
  value gate compiled to `== X` instead of `== X + 1` — it would return a
  creature with the *same* mana value as the one sacrificed rather than one
  greater.
- Independently, the sacrifice cost itself (`with mana value X`) is a
  Shoal-style pitch-bound X: X is defined by whichever creature the player
  chooses to sacrifice, not announced beforehand. That relaxation existed
  only for exile-as-cost (`exile_cost_effective_filter`), so sacrifice-cost
  eligibility evaluated the filter with `X` unresolved (== 0) and found no
  eligible permanents — the ability could never be activated at all.

## Changes

- `parse_quantity_expr_number` (`oracle_nom/quantity.rs`) now composes an
  optional `plus N` / `minus N` tail into `QuantityExpr::Offset`, mirroring
  the existing arithmetic arm in `oracle_util::parse_count_expr`. This is a
  general filter-threshold fix (mana value, power, toughness, counters),
  not a Sidisi-specific string match.
- Renamed `exile_cost_effective_filter` to `pitch_bound_x_effective_filter`
  and applied it in `find_eligible_sacrifice_targets`, so any Shoal-style
  sacrifice cost (not just exile) is payable before X is known.
- `handle_sacrifice_for_cost` binds `chosen_x` from the sacrificed
  permanent's own mana value when the cost filter is pitch-bound, mirroring
  the existing exile-pitch binding used for alternative casting costs.

## Test plan

- [x] New integration test
  `crates/engine/tests/integration/issue_5957_sidisi_regent_of_the_mire.rs`
  drives the real parsed card through `ActivateAbility` → sacrifice →
  target selection → resolution, asserting the mana-value-3 graveyard
  creature is offered as a legal target, the mana-value-2 one is not, and
  the correct creature is reanimated.
- [x] `cargo test -p engine --lib` — 16746 passed, 0 failed
- [x] `cargo test -p engine --test integration` — 3205 passed, 0 failed
  (includes the existing Shoal-pitch-exile coverage, unaffected by the
  rename)
- [x] `cargo clippy -p engine --all-targets --features cli -- -D warnings` — clean
- [x] `cargo fmt --all`
- [x] Verified Oracle text against the Scryfall API before implementing

Closes #5957


matthewevans (changes requested):
**Request changes — the selected-sacrifice X binding is annotated with rules that say the opposite.**

## 🔴 Blocker

- `crates/engine/src/game/casting_costs.rs:2304-2309` says the selected creature defines X, then cites `CR 107.3a + CR 118.9`. The verified text of `CR 107.3a` applies only when X *isn't* defined by the spell or ability text and the controller announces it; `CR 118.9` governs alternative costs for spells, not this activated-ability cost. Sidisi's sacrifice selection defines X, so this needs `CR 107.3c`; if documenting the activation sequence too, `CR 602.2b` is the applicable bridge to casting-cost steps. Please correct the annotation without changing the selected-object binding behavior.
- `crates/engine/src/parser/oracle_nom/quantity.rs:571-576` repeats the unsupported `CR 107.3a` attribution (and `CR 107.1b`, which is about negative numbers) for generic `X plus N` / `X minus N` grammar. Remove that incorrect CR claim, or cite only a verified rule that the code actually implements; the runtime X-definition annotation belongs at the binding seam above.

## ✅ Clean

The current parse-diff artifact now accounts for exactly one intended Sidisi target-filter change, and the required Rust checks are green on this head.

**Recommendation: correct the two annotations on this head, keeping the verified runtime behavior intact, then request re-review.**

matthewevans (changes requested):
**Request changes — this implements Sidisi with an impossible activation order.**

## 🔴 Blocker

- `crates/engine/src/game/cost_payability.rs:187-193` and `crates/engine/src/game/casting.rs:14963-14968` call this a text-defined X and relax the mana-value-X cost filter. Sidisi’s Oracle text has an `{X}{B}` activation cost; it does not define X from the sacrificed creature. The locally verified CR 107.3a therefore requires the controller to choose and announce X. CR 107.3c is inapplicable. Do not make the sacrifice pick bind `chosen_x`.
- `crates/engine/src/game/casting_costs.rs:2304-2312` then binds X when the sacrifice is paid. CR 602.2b applies the 601.2b–i sequence to activation: announce X under 601.2b, choose the return target under 601.2c, then pay costs. A sacrifice cannot retrospectively determine a target’s legality.
- `crates/engine/tests/integration/issue_5957_sidisi_regent_of_the_mire.rs:99-132` encodes the same reversed sequence by selecting the sacrifice before `TargetSelection`. It would accept the rules error above.

## ✅ Clean

The current parse-diff is limited to Sidisi’s intended `X + 1` target-filter change, and the parser arithmetic portion is independently scoped.

**Recommendation: model announced X before target selection and cost payment, with a test that exercises that order; otherwise remove the unsupported Sidisi activation support.**

nghetienhiep:
**Re-verified the two blockers in the latest review against Scryfall — both rest on a premise that doesn't match the printed card.**

Sidisi, Regent of the Mire's actual Oracle text (Scryfall, TDM #92, confirmed just now):

> {T}, Sacrifice a creature you control with mana value X other than Sidisi: Return target creature card with mana value X plus 1 from your graveyard to the battlefield. Activate only as a sorcery.

There is no `{X}{B}` component anywhere in this ability's activation cost — the cost is exactly `{T}` + the sacrifice. X only ever appears as the mana value of the sacrificed creature, never as a mana symbol the controller pays or announces independently.

That puts this squarely under **CR 107.3c** ("If ... the value of X is defined by the text of that spell or ability, then that's the value of X ... The controller ... doesn't get to choose the value"), not CR 107.3a (which only governs X's the controller announces). So `chosen_x` being bound from the sacrificed permanent's own mana value — rather than announced up front — is the rules-correct behavior, not a bug. The CR annotations already reflect this (CR 107.3c + CR 602.2b at `casting_costs.rs:2304` and `cost_payability.rs`), and that correction landed in the prior round.

On sequencing: CR 601.2b (bridged to activated abilities via CR 602.2b) directly addresses this exact shape —

> If the spell has a variable cost ... If the value of that variable is defined in the text of the spell by a choice that player would make later in the announcement or resolution of the spell, that player makes that choice at this time instead of that later time.

That is, the rules *themselves* pull the X-defining choice (which creature to sacrifice) forward to before target selection (601.2c), precisely to avoid the "sacrifice retroactively determines target legality" problem raised in the review. So the engine presenting the sacrifice `PayCost` prompt before `TargetSelection` isn't a reversed/impossible order — it's what CR 601.2b requires for a cost that defines X via a later choice. The integration test doesn't hardcode this order either; it reacts to whichever `WaitingFor` state the engine actually produces.

Given both blockers are built on a cost shape (`{X}{B}`) the card doesn't have, and the actual card's structure is the one CR 601.2b explicitly covers, I don't think there's a correctness issue to fix here — the current implementation matches the printed card and the cited rules. Happy to be pointed to a different reading, but wanted to flag the premise mismatch before making a change that would move the engine away from the real card's behavior.

matthewevans:
**Correction to my prior review:** I am withdrawing only the announced-`X` / target-before-cost finding from the latest requested-changes review.

I incorrectly treated this as a `{X}{B}` activation. The current head remains `8b142db491dfc0639ded512c0da7e9ca10bb3b93`, and the exact Oracle text for **Sidisi, Regent of the Mire** begins: `{T}, Sacrifice a creature you control with mana value X other than Sidisi:`. Here `X` is defined by the selected sacrifice, so that earlier finding does not apply.

An implementation re-review is underway. This correction does not approve or enqueue the PR, and it withdraws no other finding.

matthewevans (changes requested):
**Request changes — the parser change still has no discriminating test.**

## 🔴 Blocker

- **[MED]** The runtime regression at `crates/engine/tests/integration/issue_5957_sidisi_regent_of_the_mire.rs:32-40` loads prebuilt `QuantityExpr::Offset` card data, so it remains green if the new parser arm at `crates/engine/src/parser/oracle_nom/quantity.rs:561-603` is reverted. Evidence: the fixture is deserialized by `CardDatabase::from_export`, rather than parsing the Oracle text; the rest of the test only exercises the already-built ability. Why it matters: the claimed `mana value X plus 1` grammar can regress silently while this test continues to validate only runtime resolution. Suggested fix: add parser-level assertions that fresh Oracle text yields `QuantityExpr::Offset { inner: Variable("X"), offset: 1 }` for `mana value X plus 1`, plus the corresponding `minus` threshold form (and retain the existing runtime Sidisi test).

## ✅ Clean

The current head's text-defined X behavior is not part of this finding. The required checks are green, and the parse-diff is limited to Sidisi's intended target-filter change.

**Recommendation: add the discriminating parser coverage on this head, then request re-review.**

matthewevans:
This PR still has requested changes on its current head. If the requested changes are not addressed within 7 days, it will be automatically closed.

PR #6032, opened 2026-07-16
fix(parser): keep a bare imperative "Draw a card" on the controller after a player-target clause (Stone of Erech)

### LINKED ISSUE #5998: Stone OP — [[Stone of Erech]] wants two targets and DRAWS the card for the opponent instead.
Reported in Discord: https://discord.com/channels/1485498006781427802/1524643242576969889/1524643242576969889

**Thread:** Stone OP
**Discord thread id:** `1524643242576969889`
**Discord message id:** `1524643242576969889`
**Cards:** stone of erech

## Summary
[[Stone of Erech]] wants two targets and DRAWS the card for the opponent instead.

### stone of erech
If a creature an opponent controls would die, exile it instead.
{2}, {T}, Sacrifice Stone of Erech: Exile target player's graveyard. Draw a card.

---
<sub>report_id: `discord:1524643242576969889:1524643242576969889:0` · discord: `1524643242576969889/1524643242576969889`</sub>

joaovictor712:
## Problem

Stone of Erech (#5998):

> `{2}, {T}, Sacrifice Stone of Erech:` **Exile target player's graveyard. Draw a card.**

It asked for **two targets**, and **drew the card for the targeted opponent** instead of the controller.

The parse showed both symptoms in one place:

```
ChangeZoneAll { origin: Graveyard, destination: Exile, target: Player }   <- "Exile target player's graveyard"
sub_ability: Draw { count: 1, target: Player }                            <- BUG: should be Controller
```

`Draw { target: Player }` is a fresh player filter, so `collect_target_slots` surfaces a **second target slot** (the extra prompt), and at resolution the **chosen opponent draws**.

## Root cause

`apply_anchor_subject` inherits an earlier player anchor into a caster-defaulted `Effect::Draw`. That arm exists for a subject-less **CONJUGATED** continuation --

- "That player shuffles, **then draws** a card for each card exiled from their hand this way" (The End, Deadly Cover-Up)
- "... **then draws** that many cards" (Teferi's Puzzle Box, #4241)

-- where the acting player draws. But it also fired on a bare **IMPERATIVE** sentence, which is addressed to the ability's controller.

The leak is general, not Stone-specific -- a preceding *creature* target is fine, a preceding *player* target leaks:

| chain | Draw target (before) |
|---|---|
| `draw a card` | `Controller` (ok) |
| `destroy target creature. draw a card` | `Controller` (ok) |
| `exile target player's graveyard. draw a card` | **`Player`** (bug) |
| `exile target opponent's graveyard. draw a card` | **`Typed{controller: Opponent}`** (bug) |

## Fix

Gate the `Draw` arm on `chunk_continues_anchored_subject`, which reuses the **shared clause-starter authority** rather than introducing new grammar: a conjugated verb ("draws") is not itself a `starts_clause_text` starter but deconjugates to one, while a bare imperative ("draw") is a starter outright.

```rust
fn chunk_continues_anchored_subject(text_lower: &str) -> bool {
    !sequence::starts_clause_text(text_lower)
        && sequence::starts_clause_text_or_conjugated(text_lower)
}
```

This is the same distinction `inherits_carried_targeted_player_subject` already uses for the same question. Only the `Draw` arm is gated; `Shuffle` / `SearchLibrary` / `ChangeZoneAll` keep their current behavior.

## Tests

- `bare_imperative_draw_after_player_target_keeps_controller` -- the Draw keeps `Controller` for **both** the `target player's` and `target opponent's` anchored forms (the class, not one card).
- `conjugated_draw_continuation_still_inherits_player_anchor` -- the inclusion side: "target player shuffles their library, then draws a card" still inherits the anchor, pinning that the gate does not regress the case the arm exists for.

Full `engine` lib suite green (16,749 passed, 0 failed); `cargo fmt`, `clippy -p engine --tests`, and the parser-combinator gate clean.

Closes #5998

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -18435,7 +18435,7 @@ fn repeat_for_each_this_way_suffix(pred_lower: &str) -> Option<QuantityExpr> {
 /// (`Controller`), a bare "that player" placeholder (`Player`), or the generic
 /// subject-parser default (`ParentTargetController`). Does not touch effects
 /// that already carry an explicit owner anaphor (`ParentTargetOwner`).
-fn apply_anchor_subject(effect: &mut Effect, anchor: &TargetFilter) {
+fn apply_anchor_subject(effect: &mut Effect, anchor: &TargetFilter, draw_inherits_anchor: bool) {
     if !target_filter_can_target_player(anchor) {
         return;
     }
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
**Approved — current-head parser review is clean.**

The draw-anchor gate reuses the established clause-starter authority: bare imperative draw sentences retain the controller, while conjugated continuations retain the anchored player. The current parse diff’s seven affected cards are consistent with that grammar boundary. The current head also preserves the existing Alien Invasion and defending-player regression coverage, and the required CI suite is green.

Issue #5994, opened 2026-07-16
Riptide Gearhulk asks for a target among your permanents, instead of opponents — The [[Riptide Gearhulk]] etb should on…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1524666683682328677/1524666683682328677

**Thread:** Riptide Gearhulk asks for a target among your permanents, instead of opponents
**Discord thread id:** `1524666683682328677`
**Discord message id:** `1524666683682328677`
**Cards:** riptide gearhulk
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
The [[Riptide Gearhulk]] etb should only look at opponents for targets, but asks for one from you, and fizzles if skipped.

## Reported text

The [[Riptide Gearhulk]] etb should only look at opponents for targets, but asks for one from you, and fizzles if skipped.

## Attachments

- [game-state-turn-39-2026-07-09T06-38-42-279Z.zip](https://cdn.discordapp.com/attachments/1524666683682328677/1524666684072267979/game-state-turn-39-2026-07-09T06-38-42-279Z.zip?ex=6a59ce80&is=6a587d00&hm=45d0ca53fd69e7f2490412191de984bf7444de8320ac120105aad78cd227ee7a&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

### riptide gearhulk
Double strike
Prowess (Whenever you cast a noncreature spell, this creature gets +1/+1 until end of turn.)
When this creature enters, for each opponent, put up to one target nonland permanent that player controls into its owner's library third from the top.

---
<sub>report_id: `discord:1524666683682328677:1524666683682328677:0` · discord: `1524666683682328677/1524666683682328677`</sub>

mike-theDude:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `When this creature enters, for each opponent, put up to one target nonland permanent that player controls into its owner library third from the top.` (supported: true)

**Reasoning**: The AST models the library-placement effect and ties the permanent filter to `TargetPlayer`, but the target-player selector is just `Player` rather than opponent-constrained. That can explain why the UI asks for the controller own permanents; the parser/targeting contract needs to preserve the per-opponent target relationship.

---
_Posted by bug-coverage-classifier skill._

PR #6034, opened 2026-07-16
test(engine): runtime regression for Riptide Gearhulk optional per-opponent targets (#5994)

joaovictor712:
## What this is now

The parser fix for #5994 landed independently via #6565 (merged), which rewrote `per_opponent_target_fanout_min` with the verb-general `scan_at_word_boundaries` scan. #5994 is closed. This PR is **repurposed to just the runtime target-selection regression** â coverage `main` was still missing â rebased onto current `main`.

## Tests

`crates/engine/tests/integration/riptide_gearhulk_5994.rs` drives Riptide Gearhulk's real ETB through the cast pipeline into `WaitingFor::TriggerTargetSelection` with **two opponents**, each controlling a nonland permanent:

- **`riptide_declining_all_per_opponent_targets_does_not_fizzle`** â declines every optional per-opponent target; asserts the ETB resolves back to `Priority` (both permanents untouched), i.e. it does not stall or fizzle.
- **`riptide_selects_one_opponent_target_and_declines_the_other`** â chooses one opponent's permanent (moved to its owner's library third from the top) while declining the other's optional slot (untouched); asserts the ability resolves.

Both are **non-vacuous**: they pass on current `main` and **fail** if `per_opponent_target_fanout_min` is forced back to a mandatory `min: 1`. On that revert Riptide lowers to a non-targeted `EffectZoneChoice` that stalls short of `Priority`, and the chosen permanent never moves â so the `final_waiting_for()` and zone assertions each flip.

`cargo fmt`, `clippy -p engine --tests`, and the new integration tests are green.

Refs #5994

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue with Riptide Gearhulk’s enter-the-battlefield ability when optional targets are declined.
  * Ensured the ability resolves correctly when no opponents’ permanents are selected.
  * Ensured selected permanents are moved as intended while declined permanents remain on the battlefield.

* **Tests**
  * Added integration coverage for all-declined and mixed target-selection scenarios.

### DIFF
--- a/crates/engine/tests/integration/main.rs
+++ b/crates/engine/tests/integration/main.rs
@@ -767,6 +767,7 @@ mod repro_pilot_crew;
 mod rhys_evermore_remove_counters;
 mod riot_control_regression;
 mod ripples_of_undeath_regression;
+mod riptide_gearhulk_5994;
 mod rite_of_consumption_damage;
 mod roots_of_wisdom_if_you_cant_draw;
 mod roughshod_mentor_green_trample_grant;
--- a/crates/engine/tests/integration/riptide_gearhulk_5994.rs
+++ b/crates/engine/tests/integration/riptide_gearhulk_5994.rs
@@ -1,0 +1,116 @@
+//! Runtime regression for issue #5994: Riptide Gearhulk's per-opponent ETB target
+//! fanout must be OPTIONAL ("up to one target"), so declining a target for an
+//! opponent does not fizzle the ability.
+//!
+//! Riptide's ETB — "When this creature enters, for each opponent, put up to one
+//! target nonland permanent that player controls into its owner's library third
+//! from the top." — lowers to a per-opponent fanout with
+//! `MultiTargetSpec { min: 0, max: PlayerCount { Opponent } }`. Before the parser
+//! fix, the "put ..." verb fell through `per_opponent_target_fanout_min`'s
+//! "gain control of"-only anchor and each per-opponent target lowered as MANDATORY
+//! (`min: 1`); declining one then made the ability fizzle (and the harness's
+//! required-slot check panics), which is what these tests would catch on a revert.
+//!
+//! https://github.com/phase-rs/phase/issues/5994
+
+use engine::game::scenario::{GameScenario, P0, P1};
+use engine::types::game_state::WaitingFor;
+use engine::types::identifiers::ObjectId;
+use engine::types::mana::{ManaType, ManaUnit};
+use engine::types::phase::Phase;
+use engine::types::zones::Zone;
+use engine::types::PlayerId;
+
+/// Riptide Gearhulk's verbatim Oracle text (keywords carried separately).
+const RIPTIDE_ORACLE: &str = "Double strike\nProwess\nWhen this creature enters, for each opponent, put up to one target nonland permanent that player controls into its owner's library third from the top.";
+
+const P2: PlayerId = PlayerId(2);
+
+/// {1}{W}{W}{U}{U} funded into the pool so the cast auto-pays and never surfaces
+/// a `ManaPayment` window.
+fn riptide_mana() -> Vec<ManaUnit> {
+    vec![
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
**Changes requested — two correctness/verification blockers remain on `3ef0bf405d0d0f85c8990d875394b1101dcb1b46`.**

## 🟡 Medium

1. **Correct the CR citation.** Evidence: `crates/engine/src/parser/oracle_effect/lower.rs:3436` and `crates/engine/src/parser/oracle_effect/tests.rs:46610` cite CR 115.10 for optional target count. The local Comprehensive Rules source at `docs/MagicCompRules.txt:886-890` says 115.10 only distinguishes objects affected without targeting; it does not establish that `up to one target` permits zero choices. Remove 115.10 (retaining the verified relevant CR 601.2c is sufficient) or replace it with a verified rule that supports the claim.

2. **Add a production target-selection/resolution regression.** Evidence: `crates/engine/src/parser/oracle_effect/tests.rs:46688` calls only `parse_oracle_text` and asserts `MultiTargetSpec.min`; it does not exercise the reported behavior that declining a per-opponent target must not fizzle Riptide Gearhulk's ETB. Drive the real trigger pipeline through `WaitingFor::TriggerTargetSelection`, including the per-opponent player/object slots: prove one opponent can be selected while another object's optional slot is declined, and prove all optional object slots can be declined while the ability resolves. The test must fail if the `min` lowering reverts to 1.

## ✅ Clean

The verb-general optional-prefix scan is at the shared per-opponent fanout lowering seam and follows the repository's documented word-boundary scan idiom; the direct parser coverage spans `put`, `exile`, and `gain control`.

**Please address both findings, then rerun the current-head parser/CI evidence.** I did not apply a maintainer fixup: the required runtime scenario needs a verified player/object/`None` selection sequence, and the PR explicitly identifies related fanout-resolution behavior as out of scope.

matthewevans:
This PR remains blocked by the requested changes on `3ef0bf405d0d0f85c8990d875394b1101dcb1b46` (the CR citation and the discriminating declined-target runtime-path coverage). Please address them or comment with active progress. If the requested changes are not addressed within 7 days, this PR will be automatically closed.

galuis116:
Update: while rebasing to address your two blockers, I found that #6565 (merged) already landed the parser fix for #5994 â it rewrote `per_opponent_target_fanout_min` with the verb-general `scan_at_word_boundaries` scan, which supersedes this PR's parser change (and independently fixed the "that player controls" scope facet I'd deferred). #5994 is closed.

Rather than push a now-redundant duplicate, I've **repurposed this PR to just the runtime regression you asked for** (`cf8c27a43`), rebased onto current `main`. Main didn't have a Riptide target-selection test, so this is additive coverage:

- `riptide_declining_all_per_opponent_targets_does_not_fizzle` â drives Riptide's real ETB through the cast pipeline into `WaitingFor::TriggerTargetSelection` with two opponents, declines every optional per-opponent target, and asserts the ability resolves back to `Priority`.
- `riptide_selects_one_opponent_target_and_declines_the_other` â selects one opponent's permanent (moved to its owner's library) while declining the other's optional slot, and asserts resolution.

Both are non-vacuous: they pass on current `main` and fail if `per_opponent_target_fanout_min` is forced back to a mandatory `min: 1` (Riptide then lowers to a non-targeted `EffectZoneChoice` that stalls short of `Priority`, and the chosen permanent never moves â so `final_waiting_for()` and the zone assertions each flip).

The CR-citation blocker is moot now that the parser code is `main`'s, but I applied the same correction to the test doc (CR 601.2c â the verified "variable number of targets" rule â not CR 115.10).

Happy to close this instead if you'd rather not carry a test-only PR; the coverage seemed worth keeping given you'd asked for the runtime scenario.

matthewevans (approved):
Approved on cf8c27a430026665b484c2ad5dfa3597413cc5af: the test-only head exercises the real cast/trigger pipeline for all-declined and mixed per-opponent optional targets; the parser fix is already merged in #6565.

PR #6047, opened 2026-07-16
Fix Arcum's Weathervane (one-shot supertype removal)

e11734937-beep:
Model: claude-opus-4-8
Thinking: high
Method: /engine-implementer
Track: Developer

## Summary

Parser-only fix for the one-shot supertype-REMOVAL direction on a targeted permanent — "target &lt;filter&gt; isn't / is not / is no longer &lt;supertype&gt; [until end of turn]". The subject-predicate dispatch (`parser/oracle_effect/subject.rs`) already lowered the supertype-ADDITION direction ("target &lt;filter&gt; becomes &lt;supertype&gt;") to a targeted `Effect::GenericEffect` continuous grant emitting `ContinuousModification::AddSupertype`, but had no arm for the copula-negation removal direction, so those clauses fell through to the imperative fallback and lowered to `Effect::Unimplemented` — at resolution the permanent kept its supertype.

## Parser-only

The new arm (`try_parse_subject_supertype_removal_clause`, added as a sibling right after the "becomes" arm) emits the SAME targeted `Effect::GenericEffect` continuous grant its AddSupertype sibling uses, carrying `ContinuousModification::RemoveSupertype` instead. `game/layers.rs` already applies RemoveSupertype (CR 205.4 layer 4) in both directions, so there is no new effect variant and no resolver-path change — parser-only. Scoped strictly to the five CR 205.4a supertype words via `all_consuming(parse_supertype_word)` (the copula-negation core-type removal path is never claimed) and gated on an explicit target (anaphoric "it isn't legendary" does not match). A trailing duration is peeled and defaults to `Duration::Permanent` per CR 611.2b.

## Anchored on ≥2

Both anchor cards already ship the AddSupertype "becomes &lt;supertype&gt;" sibling on the same card:

- **Arcum's Weathervane** — "{2}, {T}: Target snow land is no longer snow." pairs with the already-supported "{2}, {T}: Target nonsnow basic land becomes snow."
- **Thermal Flux** — the modal "Target snow permanent isn't snow until end of turn." pairs with the already-supported "Target nonsnow permanent becomes snow until end of turn."

## Claimed parse impact

+2 supported (Arcum's Weathervane, Thermal Flux), 0 regressions. Coverage 31339 -> 31341 supported.

## Validation

- `Gate A PASS head=4e7bfc461ee7cc859ac3715bf6e843cb1a360838 base=63e95a4140ece90abe5e66027c752a7b4c9ff82e`
- `Final review-impl PASS head=4e7bfc461ee7cc859ac3715bf6e843cb1a360838`
- `cargo test -p engine -- arcum_weathervane supertype`: 47 unit + 2 integration passed, 0 failed

## CI Failures

None.


matthewevans (approved):
Approved: the targeted supertype-removal parser now preserves its copula boundary, reuses the existing `RemoveSupertype` layer path, and has current-head green verification.

matthewevans (approved):
**Approved — the targeted one-shot supertype-removal grammar is at the existing continuous-effect seam, and the current parse artifact exactly matches Arcum’s Weathervane and Thermal Flux. The runtime integration coverage reaches cast → target → resolve and verifies both permanent and until-end-of-turn duration behavior.**

PR #6049, opened 2026-07-16
fix(engine): surface OrderTriggers when combat-damage assignment fires 2+ simultaneous triggers

### LINKED ISSUE #6014: Screaming Nemesis — I dealt combat damage to my opponent's screaming nemesis and it didn't trigger.
Reported in Discord: https://discord.com/channels/1485498006781427802/1524213910171943032/1524213910171943032

**Thread:** Screaming Nemesis
**Discord thread id:** `1524213910171943032`
**Discord message id:** `1524213910171943032`
**Cards:** screaming nemesis

## Summary
I dealt combat damage to my opponent's screaming nemesis and it didn't trigger.

## Reported text

I dealt combat damage to my opponent's screaming nemesis and it didn't trigger.

## Attachments

- [game-state-turn-11-2026-07-08T00-36-35-829Z.zip](https://cdn.discordapp.com/attachments/1524213910171943032/1524213910398570556/game-state-turn-11-2026-07-08T00-36-35-829Z.zip?ex=6a597a52&is=6a5828d2&hm=4ec4c12635ab1e6167852d16d7f9943f067f813b9717a546339c1b5bef5b5ec0&)

### screaming nemesis
Haste
Whenever this creature is dealt damage, it deals that much damage to any other target. If a player is dealt damage this way, they can't gain life for the rest of the game.

---
<sub>report_id: `discord:1524213910171943032:1524213910171943032:0` · discord: `1524213910171943032/1524213910171943032`</sub>

nghetienhiep:
## Summary

Fixes #6014: Screaming Nemesis's "Whenever this creature is dealt damage" trigger silently failed to fire when Nemesis took combat damage from **two or more blockers** (or, symmetrically, when a banded blocker divides its damage among two or more attackers) in the same combat-damage step.

Root cause: `resolve_combat_damage`'s regular sub-step processes `DamageReceived` triggers internally via `process_combat_damage_triggers`. When two simultaneous `DamageDealt` events both trigger the same player's ability, that internal call sets `state.waiting_for` to `OrderTriggers` (CR 603.3b) — but `resolve_combat_damage` still returns `None`, because the damage step itself is complete. `handle_assign_combat_damage` and `handle_assign_blocker_damage` (the interactive re-entry points used whenever a creature must divide its combat damage among 2+ recipients) treated that `None` as "nothing pending" and unconditionally called `priority::reset_priority` + returned `WaitingFor::Priority`, clobbering the just-set `OrderTriggers` prompt. Both triggers were stranded in `state.pending_trigger_order`, never reaching the stack — the ability appeared not to trigger at all.

The declare-attackers and declare-blockers trigger call sites in the same file already guard against exactly this pattern (`if matches!(state.waiting_for, WaitingFor::OrderTriggers { .. })`); the two damage-assignment re-entry points were simply missing the same guard. This is a general combat-damage-trigger-ordering bug, not specific to Screaming Nemesis — it silently drops any simultaneous-trigger ordering prompt (Enrage-style triggers, damage-based triggers, etc.) raised while dividing combat damage among 2+ recipients. I also mirrored the sibling `GameOver` guard already paired with the `OrderTriggers` guard in `turns.rs`'s analogous call site, so an SBA-driven game end during this same interactive re-entry point isn't clobbered either.

Reproduced with a real cast/combat pipeline test before writing the fix: without it, `waiting_for` lands on plain `Priority` with an empty stack right after combat damage resolves — the ordering prompt (and both triggers) vanish. Confirmed via revert-probe (`git stash` on just the source fix) that the added test fails red on the old code and goes green with the fix.

## Files changed
- `crates/engine/src/game/engine_combat.rs` — added the `OrderTriggers`/`GameOver` guard to the 3 call sites that invoke `combat_damage::resolve_combat_damage` without it (`handle_assign_combat_damage`'s `AsThoughUnblocked` branch, its main body, and `handle_assign_blocker_damage`). The 4th call site (`turns.rs`, the non-interactive default path) already had this guard.
- `crates/engine/tests/integration/screaming_nemesis_combat_damage_multi_trigger.rs` — new discriminating runtime regression: drives the real `GameAction::DeclareBlockers` → `AssignCombatDamage` → `OrderTriggers` → `TriggerTargetSelection` pipeline with Screaming Nemesis blocked by two creatures, and pins that both triggers actually resolve with their own distinct damage amounts (not just that the ordering prompt appears).
- `crates/engine/tests/integration/main.rs` — registers the new test module.

## CR references
- CR 603.3b — two or more abilities have triggered since the last time a player received priority; their controller puts them on the stack in an order of their choosing.
- CR 510.1c — a blocked creature (Nemesis, the attacker in the test) with 2+ blockers divides its combat damage among them as its controller chooses — the interactive re-entry point (`handle_assign_combat_damage`) this regression lives in.
- CR 510.2 — all assigned combat damage is dealt simultaneously, so both blockers' damage to Nemesis is a single turn-based action producing two distinct `DamageDealt` events.
- CR 603.2 — each qualifying event triggers a matching ability independently, so the two simultaneous `DamageDealt` events each trigger Nemesis's ability on their own.

## Implementation method
Manual investigation: reproduced the reported behavior with a throwaway integration test driving the real combat pipeline, bisected the failure by instrumenting `combat_damage.rs`/`triggers.rs` to confirm two `DamageReceived` triggers WERE collected but never reached the stack, traced the loss to the 3 `engine_combat.rs` call sites missing the `OrderTriggers` guard already present at every sibling call site in the same file, applied the identical guard, and replaced the throwaway repro with a permanent discriminating regression test.

## Track
Developer

## Verification
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy -p engine --all-targets --features engine/proptest -- -D warnings` — clean (exit 0)
- [x] `cargo test -p engine --test integration` — 3228 passed; 0 failed; 2 ignored
- [x] `cargo test -p engine --lib` — 16758 passed; 0 failed; 7 ignored
- [x] Revert-probe: reverting just the `engine_combat.rs` hunk makes the new test fail red (`expected an OrderTriggers prompt ... got Priority { player: PlayerId(0) }`); restoring it goes green.
- [x] `./scripts/check-parser-combinators.sh origin/main` — Gate A PASS (no parser files touched)

## Gate A
Gate A PASS head=6f3e0718ee5194e0f7ac7da9f39cbd984c56ef06 base=3b52c67a9025cd20da8b68243667eb8cdad76060

## Anchored on
- `crates/engine/src/game/engine_combat.rs:328` — existing `OrderTriggers` guard in `finish_declare_attackers`, right after its own internal trigger-processing call, before falling through to `priority::reset_priority`.
- `crates/engine/src/game/engine_combat.rs:946` — the same guard in the declare-blockers finish path (`finish_declare_blockers`), confirming this is the file's established idiom for "trigger processing may have set `waiting_for` out of band — check before resetting priority."

## Claimed parse impact
None — no parser files touched, no card's parsed AST changes.

## Scope Expansion
None.

## Validation Failures
None.

## CI Failures
None.

### DIFF
--- a/crates/engine/src/game/combat_damage.rs
+++ b/crates/engine/src/game/combat_damage.rs
@@ -155,8 +155,8 @@ pub fn resolve_combat_damage(
         // combat-damage sub-step is resumed by the priority-pass completeness gate
         // in priority.rs, which re-enters this function once the order is submitted
         // and the resulting triggers resolve.
-        if matches!(state.waiting_for, WaitingFor::OrderTriggers { .. }) {
-            return Some(state.waiting_for.clone());
+        if let Some(waiting_for) = pending_combat_damage_waiting(state) {
+            return Some(waiting_for);
         }
 
         // CR 510.3 + CR 510.3a + CR 510.4: The first-strike combat-damage step is a
@@ -202,7 +202,24 @@ pub fn resolve_combat_damage(
         events,
         !combat.first_strike_done && !has_first_or_double,
     );
+    if let Some(waiting_for) = pending_combat_damage_waiting(state) {
+        return Some(waiting_for);
+    }
     None
+}
+
+/// Returns a terminal or trigger-ordering state produced while combat damage resolves.
+///
+/// CR 603.3b / CR 704.3: trigger ordering and game-over results are completion
+/// states of the combat-damage resolver, so callers must receive them rather than
+/// replacing them with a fresh priority window.
+fn pending_combat_damage_waiting(state: &GameState) -> Option<WaitingFor> {
+    match &state.waiting_for {
+        WaitingFor::OrderTriggers { .. } | WaitingFor::GameOver { .. } => {
+            Some(state.waiting_for.clone())
+        }
+        _ => None,
+    }
 }
 
 /// Which sub-step of combat damage we're collecting assignments for.
--- a/crates/engine/src/game/engine_combat.rs
+++ b/crates/engine/src/game/engine_combat.rs
@@ -601,6 +601,19 @@ pub(super) fn handle_assign_combat_damage(
         if let Some(waiting_for) = super::combat_damage::resolve_combat_damage(state, events) {
             return Ok(waiting_for);
         }
+        // CR 603.3b: resolve_combat_damage's regular sub-step processes
+        // DamageReceived triggers internally and may set `waiting_for` to
+        // OrderTriggers (2+ simultaneous triggers controlled by the same
+        // player) without returning Some(..) — mirror the guard used at every
+        // other trigger-processing call site in this file (finish_declare_attackers,
+        // finish_declare_blockers) so this interactive re-entry point doesn't
+        // strand the ordering prompt under Priority.
+        if matches!(state.waiting_for, WaitingFor::OrderTriggers { .. }) {
+            return Ok(state.waiting_for.clone());
+        }
+        if matches!(state.waiting_for, WaitingFor::GameOver { .. }) {
+            return Ok(state.waiting_for.clone());
+        }
 
         priority::reset_priority(state);
         return Ok(WaitingFor::Priority { player });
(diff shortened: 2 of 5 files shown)


matthewevans:
Resolved in `19d691a05b`: completion-state propagation now lives in `resolve_combat_damage`, which returns `OrderTriggers` or `GameOver` to every caller. The three interactive assignment paths and phase auto-advance no longer duplicate that guard. The existing integration regression still drives the assignment re-entry path and fails when the original `engine_combat.rs` propagation is reverted. Fresh CI and the current-head parse-diff are pending.

matthewevans (approved):
Approved: `resolve_combat_damage` is now the single completion-state authority; current CI and the no-change parse-diff are clean, and the runtime regression discriminates the interactive ordering loss.

PR #6050, opened 2026-07-16
fix(parser): bind "They" to an earlier mass-effect population, not the trigger's source (Ardbert, Warrior of Darkness)

### LINKED ISSUE #5985: Ardbert, Warrior of Darkness — His ability only partially fires for casting white or black spells.
Reported in Discord: https://discord.com/channels/1485498006781427802/1524817886202433566/1524817886202433566

**Thread:** Ardbert, Warrior of Darkness
**Discord thread id:** `1524817886202433566`
**Discord message id:** `1524817886202433566`
**Cards:** _none detected_

## Summary
His ability only partially fires for casting white or black spells.

## Reported text

His ability only partially fires for casting white or black spells. It'll add the +1/+1 counters but the "gain menace" or "gain vigilance" until end of turn for different spell colors doesn't activate

---
<sub>report_id: `discord:1524817886202433566:1524817886202433566:0` · discord: `1524817886202433566/1524817886202433566`</sub>

joaovictor712:
## Problem

Ardbert, Warrior of Darkness (#5985):

> Whenever you cast a white spell, put a +1/+1 counter on each legendary creature you control. **They gain vigilance until end of turn.**
> Whenever you cast a black spell, put a +1/+1 counter on each legendary creature you control. **They gain menace until end of turn.**

The report -- "his ability only partially fires" -- is exact: the counters landed, the keyword grant did nothing. The parse shows why:

```
PutCounterAll { P1P1, target: Typed{Creature, You, [Legendary]} }       <- correct
sub_ability: GenericEffect {
  static_abilities: [{ affected: TriggeringSource,                      <- BUG
                       modifications: [AddKeyword(Vigilance)] }],
  duration: UntilEndOfTurn }
```

`resolve_they_pronoun`'s object-type trigger-subject arm resolved "They" to `TargetFilter::TriggeringSource` -- **the cast spell that triggered the ability**. The grant was applied to an object on the stack, so no creature ever gained vigilance/menace. Both halves of the card are affected.

## Why `ParentTarget` is not the answer

My first instinct was to route this through `parent_target_available` / `ParentTarget`. That is wrong, and the codebase says so: per `chain_has_prior_typed_referent` and `has_typed_target_widened`'s **single-target whitelist**, that axis tracks a **CHOSEN** referent. A mass ("each ...") effect chooses nothing -- `PutCounterAll` is deliberately absent from that whitelist -- so `ParentTarget` has nothing to bind to. The anaphor must inherit the population **filter** itself.

## Fix

Add `ParseContext::chain_prior_mass_population`, seeded by the effect-chain loop from the most-recent earlier mass clause:

```rust
fn clause_ir_mass_population(clause: &ClauseIr) -> Option<TargetFilter> {
    match &clause.parsed.effect {
        Effect::PutCounterAll { target, .. }
        | Effect::PumpAll { target, .. }
        | Effect::DestroyAll { target, .. }
        | Effect::BounceAll { target, .. } => Some(target.clone()),
        _ => None,
    }
}
```

Those are exactly the variants whose `target` is a population filter rather than a chosen target -- the same reason they are absent from `has_typed_target_widened`.

`resolve_they_pronoun`'s object-type arm now prefers that population and falls back to `TriggeringSource` **only when no nearer antecedent exists**, so the existing "Whenever one or more creatures attack, **they** get +1/+0" shape (where the trigger's source genuinely IS the antecedent) is unchanged.

This mirrors the established chain-context pattern (`chain_prior_hand_reveal_target`, `chain_has_prior_exile_producer`) rather than inventing a mechanism, and is **class-level**: any `<mass effect over a population>. They <grant>` trigger benefits.

## Tests

- `they_after_mass_effect_binds_that_population_not_trigger_source` -- production path over the full card: **both** halves bind their grant's `affected` to the population the `PutCounterAll` actually acted on. The expectation is read from the parsed filter itself rather than a hand-written copy, so the test pins the *binding*, not a duplicated literal.

Full `engine` lib suite green (16,754 passed, 0 failed) -- the anaphora change surfaced no regressions. `cargo fmt`, `clippy -p engine --tests`, and the parser-combinator gate clean.

Closes #5985

### DIFF
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -22442,7 +22442,7 @@ fn clause_ir_hand_reveal_target(clause: &ClauseIr) -> Option<TargetFilter> {
     }
 }
 
-/// CR 611.2c: The match arms naming every effect that establishes a NON-targeting
+/// The match arms naming every effect that establishes a NON-targeting
 /// object POPULATION — the `*All` / scope-`All` family whose `target` (or
 /// `filter`) is a population filter rather than a chosen target. This is why they
 /// are absent from `has_typed_target_widened`'s single-target whitelist:
@@ -22473,7 +22473,7 @@ macro_rules! mass_population_filter_arms {
     };
 }
 
-/// CR 611.2c: The NON-targeting mass population filter of an "each <type>"
+/// The NON-targeting mass population filter of an "each <type>"
 /// (`*All` / scope-`All`) effect — the single authority for "which effects
 /// establish an object population". Read-only dual of
 /// [`mass_population_filter_mut`]; both are generated from one variant list
@@ -22482,7 +22482,7 @@ pub(crate) fn mass_population_filter(effect: &Effect) -> Option<&TargetFilter> {
     mass_population_filter_arms!(effect)
 }
 
-/// CR 611.2c: Mutable form of [`mass_population_filter`], for walkers that rewrite
+/// Mutable form of [`mass_population_filter`], for walkers that rewrite
 /// a mass population in place (e.g. `oracle_trigger`'s attached-host "each other"
 /// retarget). Shares the same authoritative variant list.
 pub(crate) fn mass_population_filter_mut(effect: &mut Effect) -> Option<&mut TargetFilter> {
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -22442,6 +22442,25 @@ fn clause_ir_hand_reveal_target(clause: &ClauseIr) -> Option<TargetFilter> {
     }
 }
 
+/// CR 608.2c: If this clause is a mass ("each …") effect, return the object
+/// POPULATION filter it acted on. Feeds `ParseContext::chain_prior_mass_population`
+/// so a later same-chain anaphor ("They gain vigilance until end of turn") binds
+/// to that population.
+///
+/// These are exactly the variants whose `target` is a population FILTER rather
+/// than a chosen target, which is why they are absent from
+/// `has_typed_target_widened`'s single-target whitelist: `ParentTarget` cannot
+/// bind here (nothing was chosen), so the anaphor inherits the filter instead.
+fn clause_ir_mass_population(clause: &ClauseIr) -> Option<TargetFilter> {
+    match &clause.parsed.effect {
+        Effect::PutCounterAll { target, .. }
+        | Effect::PumpAll { target, .. }
+        | Effect::DestroyAll { target, .. }
+        | Effect::BounceAll { target, .. } => Some(target.clone()),
+        _ => None,
+    }
+}
+
 /// CR 400.1/400.2 + CR 601.2a: Map a possessive-hand player reference (as
 /// produced by `parse_hand_possessive_target`) to the `ControllerRef` axis
 /// used to scope a card filter to that same player's hand. Exhaustive over
@@ -27306,6 +27325,16 @@ pub(crate) fn parse_effect_chain_ir(
                 .iter()
                 .rev()
                 .find_map(clause_ir_hand_reveal_target),
+            // CR 608.2c: the most-recent earlier same-chain mass ("each …")
+            // population, so a later "They <grant>" anaphor binds to it rather
+            // than reaching past it to the trigger's own subject (Ardbert,
+            // Warrior of Darkness). Mass effects choose nothing, so
+            // `parent_target_available` / `ParentTarget` cannot carry this.
+            chain_prior_mass_population: builder
+                .clauses()
+                .iter()
+                .rev()
+                .find_map(clause_ir_mass_population),
             // CR 608.2c: bind a bare "it" in this chunk's counter/anaphor to the
             // token created by an earlier clause when that token is the chain's
             // most-recent object referent (Esper Terra's "put up to three lore
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -22442,23 +22442,60 @@ fn clause_ir_hand_reveal_target(clause: &ClauseIr) -> Option<TargetFilter> {
     }
 }
 
+/// CR 611.2c: The match arms naming every effect that establishes a NON-targeting
+/// object POPULATION — the `*All` / scope-`All` family whose `target` (or
+/// `filter`) is a population filter rather than a chosen target. This is why they
+/// are absent from `has_typed_target_widened`'s single-target whitelist:
+/// `ParentTarget` cannot bind to them because nothing was chosen.
+///
+/// The arms live in a macro so the SINGLE authoritative variant list generates
+/// both the shared `&` and `&mut` accessors below — two hand-written matches would
+/// be two lists that drift. Callers must go through those accessors rather than
+/// re-listing variants.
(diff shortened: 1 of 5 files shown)


galuis116:
Thanks for the pass. Both findings are inaccurate -- evidence below so a human reviewer isn't misled.

### [HIGH] "chain_prior_mass_population is never populated" -- it is

It is seeded by the effect-chain loop at `crates/engine/src/parser/oracle_effect/mod.rs:27333`, in the same chunk-context construction as its siblings (`chain_prior_hand_reveal_target`, `chain_has_prior_exile_producer`):

```rust
chain_prior_mass_population: builder
    .clauses()
    .iter()
    .rev()
    .find_map(clause_ir_mass_population),
```

The cited evidence (`context.rs:263`) is the field's *declaration*, not its population site -- the seeding lives in the chain loop, which is exactly where the PR describes it.

This is also self-evident from the test: `they_after_mass_effect_binds_that_population_not_trigger_source` asserts the grant's `affected` equals the population the `PutCounterAll` acted on. If the field were always `None`, the code would fall back to `TriggeringSource` and that assertion would fail. It passes.

### [MED] "CR 201.5c governs pronoun antecedents, not CR 608.2c" -- inverted

Verified against `docs/MagicCompRules.txt`:

- **CR 201.5c** (line 1332): *"Text printed on some cards refers to that card by a **shortened version of its name**. Instances of a card's shortened name used in this manner are treated as though they used the card's full name."* -- this governs **shortened card names** (e.g. "Kroxa" for "Kroxa, Titan of Death's Hunger"). It says nothing about pronouns.
- **CR 608.2c** (line 2793): *"...In some cases, **later text on the card may modify the meaning of earlier text** (for example, "Destroy target creature. **It** can't be regenerated")... read the whole text and **apply the rules of English** to the text."*

CR 608.2c is the rule about a later clause's **pronoun** binding to an earlier clause -- its own worked example is literally the pronoun "It" referring back to a previous sentence, which is precisely this card's shape ("...on each legendary creature you control. **They** gain vigilance..."). CR 201.5c would be the incorrect citation here.

It is also the citation the surrounding authority already uses for this exact question -- `resolve_they_pronoun`, `chain_has_prior_typed_referent`, and `parent_target_available` are all annotated CR 608.2c. Switching this one call site to 201.5c would make it inconsistent *and* wrong.

No changes made. Happy to revisit if a human reviewer disagrees on either point.

matthewevans:
**Follow-up reviewed.** The two Gemini findings do not apply to the current head: `parse_effect_chain_ir` seeds `chain_prior_mass_population` from earlier finalized clauses, and the verified CR 608.2c text is the relevant rule for later text changing the meaning of an earlier clause; CR 201.5c concerns shortened card names.

The PR remains held only for the current-head evidence required for an engine/parser change: the coverage parse-diff comment is absent and the Rust lint/test jobs are still running. I will resume review when those artifacts are available.

matthewevans (changes requested):
**Changes requested — the new anaphora context covers only a subset of the mass-population grammar, so it is not safe to merge as the general fix.**

## 🔴 Blocker

`crates/engine/src/parser/oracle_effect/mod.rs:22435` defines `clause_ir_mass_population` as the prior population for a later `They <grant>` clause, but it recognizes only `PutCounterAll`, `PumpAll`, `DestroyAll`, and `BounceAll`. The same engine surface has other object-population effects with the same `target` semantics: `CounterAll`, `GainControlAll`, `DamageAll`, `DoublePTAll`, `ExploreAll`, and mass `SetTapState`; the existing mass-filter walker at `crates/engine/src/parser/oracle_trigger.rs:3939-3951` already shows that wider family. A trigger whose first clause lowers through one of the omitted object-population effects still reaches `resolve_they_pronoun` at `subject.rs:3045` with no `chain_prior_mass_population`, and silently falls back to `TriggeringSource`.

Please make the prior-population extraction a single, shared authoritative mapping for all object-population effect variants (rather than another partial list), then add sibling regression coverage for at least an omitted `GainControlAll`/`DoublePTAll`-shaped producer plus the existing counter case. The test needs to assert the following grant uses the producer population, not merely that parsing succeeds.

## ✅ Clean

The current parse-diff evidence is fresh for `1131e373` and scoped to Ardbert's two grants. Gemini's two comments do not apply to this head: the chain builder seeds the context before each later chunk, and CR 608.2c is the relevant verified instruction-order rule for the earlier-clause antecedent.

**Recommendation: request changes for the shared mass-population extractor and sibling tests; then re-review the new head.**

galuis116:
Addressed in 4e1b4f84c -- you're right, that was a partial list, not the class.

### Single shared authority

The variant list is now hoisted into one place and generates **both** accessors, so there is no second list to drift:

```rust
macro_rules! mass_population_filter_arms {
    ($effect:expr) => {
        match $effect {
            Effect::CounterAll { target } | Effect::GainControlAll { target } => Some(target),
            Effect::PumpAll { target, .. }
            | Effect::DamageAll { target, .. }
            | Effect::DestroyAll { target, .. }
            | Effect::BounceAll { target, .. }
            | Effect::PutCounterAll { target, .. }
            | Effect::DoublePTAll { target, .. } => Some(target),
            Effect::ExploreAll { filter } => Some(filter),
            Effect::SetTapState { target, scope: EffectScope::All, .. } => Some(target),
            _ => None,
        }
    };
}

pub(crate) fn mass_population_filter(effect: &Effect) -> Option<&TargetFilter>;
pub(crate) fn mass_population_filter_mut(effect: &mut Effect) -> Option<&mut TargetFilter>;
```

I used a macro deliberately: two hand-written matches (one `&`, one `&mut`) would be exactly the drift this review is about.

All the producers you named are now covered -- `CounterAll`, `GainControlAll`, `DamageAll`, `DoublePTAll`, `ExploreAll`, mass `SetTapState`. `clause_ir_mass_population` is now a one-liner over the authority.

### The walker you pointed at now shares it

Thanks for the pointer to `oracle_trigger.rs:3939-3951` -- that was the right precedent. Rather than mirror its list, I made it **delegate** to the same `mass_population_filter_mut`, so the attached-host "each other" retarget and the anaphora seeding provably read one list instead of two copies.

### Sibling coverage

`they_binds_producer_population_across_mass_effect_family` is table-driven over `PutCounterAll` (Ardbert's own shape) plus the previously-omitted `GainControlAll`, `DoublePTAll`, and mass `SetTapState`. Each case:

- reads the population back out of the **parsed producer** via the shared authority (so it pins the binding rather than duplicating a literal filter),
- asserts the following grant's `affected` **equals that population**, and
- asserts it is **not** `TriggeringSource`.

I verified each omitted producer genuinely regressed before the widening -- e.g. `gain control of all creatures target opponent controls. They gain haste` bound the grant to `Typed{Creature, TargetOpponent}` only after the fix; previously it fell through to the trigger's source.

Full `engine` lib suite green (16,770 passed, 0 failed); `cargo fmt`, `clippy -p engine --tests`, and the parser gate clean.

Also noted your confirmation on Gemini's two comments -- no changes were made for either.

matthewevans (commented):
**Current head reviewed; held for fresh CI.** The shared mass-population extractor now covers the previously omitted sibling producers and the parser tests exercise counter, control, P/T, and mass-tap shapes. I also removed the new `CR 611.2c` citations: that rule governs the locked affected set of a resolving continuous effect, whereas the shared accessor also classifies one-shot effects. No additional implementation finding on `89c49b6`.

matthewevans (approved):
Approved after current-head parser review: the shared mass-population authority covers the established *All/scope-All family; the 3-card parse-diff matches the intended antecedent binding and required checks are green.

PR #6056, opened 2026-07-16
feat(engine): implement Jetfire, Ingenious Scientist // Jetfire, Air Guardian

JacobWoodson:
Implements the Transformers DFC **Jetfire, Ingenious Scientist // Jetfire, Air Guardian** (set BOT).

## Summary
End-to-end tracing showed almost every mechanic already existed — More Than Meets the Eye, Living metal, Flying, `convert`→transform, adapt, the "can't be spent to cast nonartifact spells" restriction (`SpellTypeOrAbilityActivation{Artifact, Any}`) and its cross-sentence attachment, and the counter-removal cost → `chosen_x` → "that much" wiring. Four surgical changes complete the two activated abilities:

1. **parser** (`oracle_effect/mana.rs`): `strip_mana_subject_prefix` recognizes **"target player adds"** → `TargetFilter::Player`, a genuine chosen recipient distinct from the "active/that player" anaphors (CR 115.1 + CR 106.4).
2. **engine** (`effects/mana.rs`): `mana_effect_recipient` now deposits mana into the **targeted player** for a `Player` recipient, **provenance-gated** by a new `mana_count_reads_targets` predicate — the identical `TargetFilter::Player` is also emitted as a *count source* ("Add {U} for each card in target player's hand"), where the recipient must stay the controller. No Jeska's Will regression.
3. **parser** (`oracle_effect/mana.rs`): "that much **{C}**" now parses as a count-prefixed colorless production (previously fell to `Unimplemented`).
4. **parser** (`sequence.rs`): `"adapt"` added to the clause-start verb table so **"Convert Jetfire, then adapt 3"** chains the adapt sub-ability.

**No new enum variants.**

## Card text
- **Front** `{4}{U}` Legendary Artifact Creature — Robot 3/4: More Than Meets the Eye {3}{U}; Flying; *Remove one or more +1/+1 counters from among artifacts you control: Target player adds that much {C}. This mana can't be spent to cast nonartifact spells. Convert Jetfire.*
- **Back** Legendary Artifact — Vehicle 3/4: Living metal; Flying; *{U}{U}{U}: Convert Jetfire, then adapt 3.*

## Tests
6 new tests: both faces parse end-to-end (front: chosen-X counter cost + targeted restricted mana + Convert sub-ability; back: convert→adapt chain), the "target player adds that much {C}" clause, targeted-player deposit, the multi-authority recipient-vs-count-source gate (incl. Jeska's Will non-regression), and spend-restriction deposit.

## Verification
- Full engine test suite: **16,725 passed, 0 failed**
- `cargo fmt` clean; `cargo clippy -p engine --lib --tests` clean (`-D warnings`)

## Summary by CodeRabbit

* **New Features**
  * Mana targeting now models recipient vs count-source as distinct roles, enabling correct multi-slot selection and retargeting.
* **Bug Fixes**
  * Fixed mana resolution to scope quantities to the declared count-source role and handle illegal recipient/count-source choices without redirecting.
  * Improved mana scanning/legacy retargeting and strengthened memoization and coverage rendering to correctly reflect role-specific targeting.
* **Tests**
  * Added unit, regression, and integration tests for role parsing/resolution and multi-role mana behavior.
* **Chores**
  * Added a migration script to convert legacy mana-target data into role-based targeting forms.


matthewevans (changes requested):
**Blocked — the new generic target-player mana path still conflates a mana recipient with a count source.**

## 🔴 Blocker

`crates/engine/src/parser/oracle_effect/mana.rs:193-201` recursively parses the bare `add …` clause before it stamps the subject recipient. For `Target player adds {C} for each card in target player's hand`, the bare parser already emits `Effect::Mana.target = Some(TargetFilter::Player)` as the count-source target, so the wrapper leaves that field unchanged. `crates/engine/src/game/effects/mana.rs:35-44` then uses the production count to reinterpret that same field and returns `ability.controller` whenever the count reads the target. That sends mana to the controller even though the sentence's subject is the chosen target player.

The current Jetfire fixture is target-independent (`that much`), so it does not reach this failure mode. The new parser arm accepts the broader subject-led class, however, and the implementation has no typed provenance that distinguishes its recipient target from the count target. The wildcard fallbacks in `quantity_expr_reads_targets` / `quantity_ref_reads_targets` (`effects/mana.rs:87-144`, also raised by Gemini) make that inference incomplete as the quantity vocabulary grows.

Suggested fix: model the two roles explicitly at the `Effect::Mana`/mana-production boundary (recipient vs count-source) and thread their independently surfaced target slots through parsing and resolution. Do not use quantity-shape inference to reinterpret the one overloaded `target` field; add a runtime fixture for the target-recipient + target-derived-count class as well as the existing Jetfire and Jeska's Will cases.

## ✅ Clean

The Scryfall Oracle text confirms Jetfire's front and back clauses, and the change uses the existing `convert`/`adapt` and restricted-mana seams. The Jetfire-specific parser assertions are appropriately detailed.

**Recommendation: request changes for the explicit target-role model, then re-run current-head parser/engine evidence.**

JacobWoodson:
Thanks @matthewevans — the critique was correct, and digging into it showed the problem was worse than "architecturally untidy". Pushed as `70911c11`.

## The bug you predicted is real

For `Target player adds {C} for each card in target player's hand`, the inner `add …` parse already claimed `Effect::Mana.target` as the **count source**, so the subject-led wrapper's `if target.is_none()` guard silently discarded the **recipient** and the mana went to the controller. It could not have been fixed by better inference: `QuantityRef::TargetZoneCardCount` resolves by taking the *first* `TargetRef::Player` in `ability.targets`, so a recipient slot and a count-source slot both land at index 0 and collide. One field genuinely cannot carry two CR 601.2c instances of "target".

## What changed

- **Explicit roles.** New `ManaTargetRole { Recipient, CountSource, Both }` replaces `Effect::Mana.target`'s bare `Option<TargetFilter>`. The role is stamped at **parse time from grammar** — the point that actually knows which grammatical position a filter came from — never from the quantity.
- **Quantity-shape inference deleted outright.** `mana_count_reads_targets`, `mana_production_count`, `quantity_expr_reads_targets`, `quantity_ref_reads_targets` are gone. That also removes the two wildcard `_ => false` arms and the missing-CR annotation Gemini flagged, by construction rather than by patching them.
- **Independently surfaced slots.** One slot per role, recipient first, in `collect_target_slots` *and* its `collect_target_slot_specs` mirror, with matching arms in both assign functions, `chain_has_target_sink`, `minimum_targets_in_chain`, and a position-stable `validate_targets_in_chain` arm. `retarget_slot_violation` enforces per-slot CR 115.7a legality (`each target can be changed only to another legal target`), which the flat retarget pool could not express.
- **Per-role resolution.** Each role resolves against its own slot via `ability_scoped_to_slot`, so shared quantity resolution stays slot-agnostic and untouched. `mana_effect_recipient` now returns `Option<PlayerId>`, so an illegal recipient deposits *nowhere* rather than falling back to the controller (CR 608.2b).

## Blast radius was deliberately held to the new class

`Effect::target_filter()` still returns the sole declared filter (`declared_filters().next()`). Making it return `None` for Mana looked cleaner but cascaded into ~40 call sites — `chain_has_target_sink`, `minimum_targets_in_chain`, the four pre-generic assign branches, the stack/triggers auto-skip gates, `change_targets`, `ai_support/filter` — and silently changed routing for shipping cards. The explicit arms are gated on `mana_multi_role()`, so **all 13 shipping mana-target cards keep byte-identical routing**; only the two-role shape enters new code. Worth noting for reviewers: 10 of the 11 fixture entries carry *context-ref* recipients, so a "first **surfaced** filter" accessor would return `None` for them and reinstate the whole cascade — hence `declared_filters().next()`.

## Fixture migration is name-keyed, not shape-derived

`integration_cards.json` carries 11 legacy `Effect::Mana` targets. A legacy-tolerant `Deserialize` is impossible without re-introducing exactly the inference you rejected: **Carpet of Flowers** (count source) and **Spectral Searchlight** (recipient) are both `Typed` with opposite roles. So the 11 were migrated by card name, with Carpet of Flowers as the canary, and the file stays one minified line (1 insertion / 1 deletion, zero snapshot churn).

## Tests — all three classes you asked for

1. Jetfire: recipient + target-**independent** count → the targeted player's pool.
2. Jeska's Will: count-source only → the **controller's** pool.
3. New end-to-end cast-pipeline fixture for recipient + target-**derived** count: two *different* players with *different* hand sizes (2 vs 5), so a slot mix-up changes both the destination and the amount. Reverting the fix fails it with `left: 0, right: 5`.

18 new tests total. Engine suite 16743 passed / 0 failed; integration 3164 / 0; `cargo clippy --workspace --all-targets -- -D warnings` clean.

## Known gaps, stated plainly

- `retarget_slot_violation` is covered at helper level, not driven through the full `GameAction::RetargetSpell` → `apply_retarget` route. It is gated behind a shape no printed card reaches, so I left it as a declared gap rather than build more synthetic harness — happy to add it if you'd rather.
- `Attach` / `MoveCounters` / `Fight` share the same pre-existing flat-pool retarget gap. Deliberately untouched; `retarget_slot_violation` is the seam they'd extend into if that's fixed on its own merits.
- `cargo coverage`, `cargo semantic-audit`, and parser gate family (D) could not run in my environment (missing `card-data.json` / `python3`). Treating those as *could-not-determine* rather than passing — CI covers them.

One deliberate behavioral note: `validate_targets_in_chain`'s new arm is gated on `mana_multi_role`, so single-role mana keeps the generic companion-aware `split_first` path. An earlier revision broadened it and silently discarded a legal `unless_pay` companion target; that's fixed and pinned by a test.

matthewevans (changes requested):
**[MED]** An illegal count-source target can make the effect count the recipient instead. Evidence: `crates/engine/src/game/effects/mana.rs:110-120` falls back to `ability.clone()` whenever `ability_scoped_to_slot(..., CountSource)` returns `None`; `crates/engine/src/game/quantity.rs:2134-2141` then reads the first player target. Why it matters: for a `Both` mana effect whose recipient remains legal but whose count source becomes illegal, the fallback exposes the recipient as that first target, so the effect produces mana based on the wrong player's hand rather than failing to determine the illegal target's information (CR 608.2b). Suggested fix: distinguish “no count-source role” from “declared count source no longer legal”; for the latter resolve the count through an empty/invalid scoped target view (or skip the dependent mana instruction), and add a cast-pipeline regression with a legal recipient and an illegal count source having different hand sizes.

JacobWoodson:
Confirmed and fixed in `f4789fa9`. You were right, and I reproduced it before touching anything: reverting the fix makes the new regression fail with `left: 2, right: 2` — the count read the **recipient's** hand (2) instead of failing to determine. Exactly the path you described.

The root cause was that `count_scoped_ability` conflated two different `None` returns from `ability_scoped_to_slot`. It now distinguishes **four** cases, not three:

1. **No count-source role declared** — unscoped, which is exactly today's behavior for every single-role mana (Jetfire, and all `target: None`).
2. **Context-ref count source** — resolved through context, mirroring what `mana_effect_recipient` already did. This one is on me: `slot_index` returns `None` for a context-ref filter, not just for an illegal target, so this input was landing in case 4 and silently resolving to 0 **under a CR 608.2b citation that does not apply to it** — nothing there is an illegal target. Latent today (the parser only builds `Player` / opponent-`Typed` count sources) but `with_count_source` is public.
3. **Declared and still legal** — narrowed to that one player.
4. **Declared but illegal** — no player exposed, count resolves to 0 per CR 608.2b.

## Two things beyond the reported instance

**The same defect existed at a second site you didn't cite.** `handle_choose_mana_effect` → `chosen_mana_types_for_prompt` also derives the count — its `SingleColor` arm multiplies the chosen color by it — and was passing the unscoped ability. So a `Both` role with a colour prompt had the identical bug. Both sites now route through `count_scoped_ability`. Fixing only the site you named would have repeated the "fixed the instance, missed the class" mistake from the previous round.

**Scoping is now confined to the player axis.** My first pass cleared the whole `targets` vec, which was too blunt: `ManaProduction::AnyCombinationOfObjectColors { scope: Target }` reads an **object** target out of that same vec and requires nothing about the count source. Clearing it would have made the colour half of the production silently yield no colours — which CR 608.2b doesn't license, since only the part requiring information about the illegal target fails. `retain_only_player_at` now drops only `TargetRef::Player` entries and leaves non-player targets at their original positions. That also tightened case 3, which inherited the same over-broad clear from `70911c11`.

## Regression test

`illegal_count_source_fails_to_determine_instead_of_counting_the_recipient` — real cast pipeline, 3 players, recipient P1 holding 2 cards, count source P2 holding 5, P2 eliminated between `SelectTargets` and resolution. Two reach guards, because `assert_eq!(pool, 0)` on its own is satisfiable by a whole-spell fizzle: the recipient is asserted still legal, and the test collects `ActionResult.events` through the resolve loop and asserts `EffectResolved { kind: Mana }` actually fired. So it pins the per-target "fails to determine" clause rather than CR 608.2b's other clause.

Engine suite 16743 passed / 0 failed; integration 3165 / 0; `cargo clippy --workspace --all-targets -- -D warnings` clean.

## Still open from my earlier note

`retarget_slot_violation` remains helper-level only (not driven through `GameAction::RetargetSpell`), and the `Attach` / `MoveCounters` / `Fight` flat-pool retarget gap is untouched. Both are gated behind shapes no printed card reaches. Say the word if you'd rather I close either now rather than leave them as declared gaps.

matthewevans (changes requested):
**Blocked — the current fix addresses the earlier illegal count-source fallback, but the authorized maintainer port cannot safely complete under the Tilt-only verification policy.**

The attempted merge across current `main` has one conflict in `crates/engine/tests/fixtures/integration_cards.json`. This is a generated single-line fixture, not a textual choice: this PR regenerates 12 mana-related entries, while current `main` has a broad independent regeneration (2,945 keys versus this branch's 2,807). Several entries overlap. Taking either side would silently discard parser-derived fixture behavior; producing the merged artifact requires the project card-data/Tilt path, which is not attached to this isolated worktree and must not be replaced by a direct build.

The targeted code review of `f4789fa` found the new count-scoped view and its real cast-pipeline regression appropriately address the prior recipient-count fallback. The remaining blocker is the verified generated-fixture merge/verification conflict only.

**Recommendation: hold this head until a Tilt-attached maintainer port can regenerate and verify the fixture.**

matthewevans:
**Maintainer status — held on a real merge conflict.**

Current head `4ce98ba3bc3ad9147f0fe0ed6694685cca17e268` is `DIRTY`/`CONFLICTING`, so GitHub's update-branch operation cannot perform the required semantic integration. The earlier maintainer port identified a conflict in the generated integration-card fixture; resolving it requires the generated fixture to be regenerated and verified through the project's Tilt workflow. I have not used a direct build or a pick-a-side merge resolution.

matthewevans (changes requested):
One multi-role retargeting path remains unsafe:

The new `ManaTargetRole::Both` branch can produce a flat target union, but forced retargeting replaces `stack_ability.targets` with a one-element vector and bypasses the per-slot validation used by interactive retargeting. A forced-target effect can therefore drop the other target or place the target in the wrong role. Please preserve the target vector and select/validate the intended role slot (or keep multi-slot mana out of forced retargeting until its semantics are represented), with an end-to-end forced-retarget regression.

matthewevans (changes requested):
**Blocked — the current head still drops multi-role targets during forced retargeting and can corrupt a future migration.**

## 🔴 Blocker

`crates/engine/src/game/effects/change_targets.rs:71-82` takes the forced-retarget path outside `engine.rs::apply_retarget`: after checking a flat union, it overwrites `stack_ability.targets` with `vec![new_target]`. A `ManaTargetRole::Both` spell has two independently declared target slots; this drops the other slot and never applies the positional legality check added for interactive retargeting. CR 115.7a says each target may change only to another legal target and otherwise the original target remains unchanged; CR 115.7b permits changing only one target, not deleting the others (verified in `docs/MagicCompRules.txt:863-869`). The existing forced-retarget test covers a single-target destroy spell, so it cannot exercise this new multi-slot production path.

Please represent which target slot a forced effect changes, preserve all other targets, and validate the replacement against that slot's filter. Add an end-to-end forced-retarget regression for a `Both` mana ability that proves both the untouched target and the role-specific replacement survive; it should fail with this `vec![new_target]` assignment.

## 🟡 Non-blocking

`scripts/migrate-mana-target-roles.mjs:151-155,210,271` documents `Both` as a supported mapping but has field names only for `Recipient` and `CountSource`. If a maintainer follows that documented path, the computed key is `undefined`, producing a malformed `{ role: "Both", undefined: ... }` envelope rather than failing loudly. A legacy single filter cannot reconstruct both roles, so reject `Both` explicitly (and require a dedicated migration) before either writer constructs an envelope.

## ✅ Clean

The role model correctly separates recipient and count-source provenance, and the current parse-diff artifact is present for this head: 58 cards / 21 signatures, updated after the merge commit (`https://github.com/phase-rs/phase/pull/6056#issuecomment-4997648480`).

**Recommendation: request changes. Resolve the forced-retarget behavior with a discriminating runtime test; the migration guard can ride with that repair.**

matthewevans (changes requested):
**Blocked — forced retargeting still destroys one target slot for a multi-role mana ability.**

## 🔴 Blocker

crates/engine/src/game/effects/change_targets.rs:71-82 validates the forced replacement against a flat union and then assigns stack_ability.targets = vec![new_target]. A mana ability with independent recipient and count-source targets therefore loses its other declared target; this path does not call the positional retarget_slot_violation guard added only to interactive apply_retarget in crates/engine/src/game/engine.rs:8456-8483. CR 115.7a requires each target to be changed only to another legal target, not removal or reassignment of a separate target slot.

Use the role-slot authority to identify which slot the forced filter can replace, preserve the full target vector, validate that slot against its own filter, and add an end-to-end forced-retarget regression with distinct legal recipient and count-source players.

**Recommendation: request changes; retain the two target roles through the forced-retarget assignment before re-review.**

matthewevans (changes requested):
**Blocked — forced multi-role retargeting can no-op even when another slot can legally change.**

## 🔴 Blocker

`crates/engine/src/game/effects/change_targets.rs:146-165` chooses the first `ManaTargetRole::Both` filter that accepts `new_target`, but it never requires `current_targets[slot] != new_target`. With overlapping recipient/count-source filters and current targets `[A, B]`, forcing `A` selects slot 0 and writes `A` back to it, even though slot 1 can legally change from `B` to `A`. The current forced-retarget behavior therefore fails to make the available legal change.

CR 115.7a states: “each target can be changed only to another legal target” and “If a target can’t be changed to another legal target, the original target is unchanged.” The forced path must choose the first in-bounds role slot whose filter legally accepts the candidate **and whose current target differs**; if no such slot exists, preserve `current_targets` rather than falling back to `vec![new_target]`.

The current parse-diff artifact is present for this head (58 cards / 21 signatures). The existing distinct-slot regression does not exercise overlapping role filters, so it cannot catch this first-match/no-op branch. Add an end-to-end forced-retarget runtime regression with overlapping `ManaTargetRole::Both` filters and `[A, B]` that proves forcing `A` results in `[A, A]` by changing slot 1 while preserving slot 0.

**Recommendation: request changes — make forced slot selection require a real legal replacement and cover the overlapping-role branch.**

matthewevans (changes requested):
**Changes requested — scope/proof reconciliation is required before this can merge.**

## 🟡 Scope / evidence blocker

**[MED]** The current parser impact is materially broader than the stated Jetfire-only scope, without a reconciliation. Evidence: the PR body describes four surgical Jetfire changes and six tests, while the current-head parse-diff (`<!-- coverage-parse-diff -->`, comment 4997648480, updated 2026-07-29T01:53:01Z) reports 58 cards across 21 signatures: added support for Su-Chi Cave Guard, Rasputin, the Oneiromancer, and Mana Seism, plus target-to-mana-recipient/count-source migrations affecting groups of 20, 12, 7, 5, 4, and 3 cards. Why it matters: these may be valid generalizations, but an unreconciled parser blast radius leaves no auditable proof that behavior-changing non-Jetfire entries are intentional and correctly covered. Suggested fix: add an exact card/signature reconciliation explaining why every non-Jetfire change is intentional, with behavior evidence/tests for each behavior-changing entry; narrow or revert any unintended effect.

This is a scope/proof blocker, not a claim that the additional parser changes are necessarily incorrect. The existing forced-retarget finding remains separately valid.

**Recommendation: request changes with the reconciliation or a narrowed diff.**

JacobWoodson:
@matthewevans — thanks for porting the fixture across `main` yourself in `803e6905`. I saw you're hands-on with the branch, so rather than push onto it, here's the remaining **CR 115.7a forced-retarget refinement** (your other blocker from today) as a patch for you to apply on your terms.

Your branch already has `forced_retarget_targets` (multi-role slot preservation), but not the "must change to a **different** target" guard. As you noted: with overlapping recipient/count-source filters and targets `[A, B]`, forcing `A` selected slot 0 (already `A`, a no-op) instead of changing slot 1 (`B → A`). CR 115.7a requires changing a target only to *another* legal target, and leaving it unchanged if none exists.

### The fix — `crates/engine/src/game/effects/change_targets.rs`

The slot predicate now requires filter-legality **and** `current != new`, and the `vec![new_target]` fallback is dropped (no qualifying slot → all targets unchanged):

```rust
    let slot = match crate::types::ability::mana_multi_role(&stack_ability.effect) {
        Some(role) => role
            .surfaced_filters()
            .enumerate()
            .find_map(|(i, (_slot, filter))| {
                let changes = current_targets.get(i).is_some_and(|cur| *cur != new_target);
                let legal = !crate::game::targeting::validate_targets_for_ability(
                    state,
                    std::slice::from_ref(&new_target),
                    filter,
                    stack_ability,
                )
                .is_empty();
                (changes && legal).then_some(i)
            }),
        None => Some(0), // single-role / non-mana: sole slot, pre-role behavior
    };
    match slot {
        Some(i) if i < current_targets.len() => {
            let mut targets = current_targets.to_vec();
            targets[i] = new_target;
            targets
        }
        // CR 115.7a: no slot can legally change to another target → unchanged.
        _ => current_targets.to_vec(),
    }
```

Behavior across the reachable cases: `[A,B]` force `A` → `[A,A]` (slot 1); `[A,B]` force `C` legal for both → `[C,B]` (first slot that changes); force a target legal for no slot, or already in every slot → unchanged. Single-target / non-mana keeps slot-0 replacement, byte-identical to before.

### Regression test

Adds `forced_retarget_prefers_the_slot_that_actually_changes` to `crates/engine/tests/integration/forced_retarget_multi_role_mana_6056.rs` — overlapping `Both` filters, `[P1, P2]`, force `P1`, driven through real `resolve_top`, asserting `[P1, P1]` with slot 0 unchanged as the reach-guard. **Revert-verified**: without the `current != new` guard it fails `[P1, P2]` (`left: P2, right: P1`).

### How to apply

The commit is on my fork branch [`JacobWoodson:jetfire-rebased`](https://github.com/JacobWoodson/phase) as `46c302de` if you'd rather cherry-pick it. Otherwise the change above plus the test drops straight onto your `803e6905`. I verified the full delta green on top of current `main` (crate now `phase-engine`): lib 18,036 / integration 4,216 / clippy `--workspace --all-targets` clean, including one adaptation main forced — a new `Effect::Mana { target: Some(TargetFilter::Player) }` site in `phase-ai/src/zone_eval.rs` that needs wrapping in `ManaTargetRole::Recipient { .. }`.

### Two LOW residual notes (pre-existing, your call — not introduced here)
1. Non-mana multi-target forced retarget still takes the slot-0 path (no per-slot `changes && legal` search) — unreachable today since only mana roles surface per-slot filters, and it's strictly better than the old collapse.
2. The interactive retarget seam (`retarget_slot_violation`) validates per-slot legality but doesn't enforce "must change," so a `RetargetScope::Single` interactive submission could re-pick the current target where the forced path now forbids it. CR 115.7b vs 115.7d nuance — flagging for your judgment.

matthewevans (changes requested):
**Blocked — the current head does not compile, forced retargeting remains rules-incorrect, and the parser blast radius is unreconciled.**

## 🔴 Blocker

**[HIGH]** The multi-role mana target-slot builder returns the wrong error type. Evidence: `crates/engine/src/game/ability_utils.rs:2643-2645` returns `EngineError::ActionNotAllowed` from a `Result<(), TargetSlotBuildError>` path; the current-head CI run [30509300434](https://github.com/phase-rs/phase/actions/runs/30509300434) reports E0308 there: expected `TargetSlotBuildError`, found `EngineError`. The adjacent no-legal-target paths use `no_legal_target_slots()` at lines 2617 and 2661. Why it matters: the engine cannot compile, so no behavior here is shippable. Suggested fix: return the established `no_legal_target_slots()` value from this branch (or otherwise convert to the function's declared error type) and rerun the affected checks.

**[HIGH]** Forced retarget selection can choose a slot without actually changing it. Evidence: `crates/engine/src/game/effects/change_targets.rs:146-165` selects the first `ManaTargetRole::Both` filter that accepts `new_target`, but does not require `current_targets[slot] != new_target`. With overlapping role filters and current targets `[A, B]`, forcing `A` selects slot 0 and leaves slot 1 unchanged even though slot 1 can legally change to `A`. CR 115.7a states: “each target can be changed only to another legal target”; it also says an original target remains unchanged only when it cannot be changed to another legal target (verified in `docs/MagicCompRules.txt:863`). Why it matters: a forced target-changing effect fails to make the available legal change. Suggested fix: choose the first in-bounds role slot whose filter accepts the candidate and whose existing target differs; if none exists, preserve `current_targets`. Add an end-to-end overlapping-role regression (`[A, B]`, force `A`) that proves slot 1 becomes `A` while slot 0 remains `A`.

## 🟡 Scope / proof

**[MED]** The claimed Jetfire-only scope is not reconciled with the current parser artifact. Evidence: the PR body describes four surgical Jetfire changes and six tests, while the current `<!-- coverage-parse-diff -->` artifact reports 58 cards and 21 signatures, including Su-Chi Cave Guard, Rasputin, the Oneiromancer, Mana Seism, and grouped migrations affecting 20, 12, 7, 5, 4, and 3 cards. Why it matters: these may be valid generalizations, but the current PR supplies no auditable proof that each behavior-changing non-Jetfire entry is intentional and covered. Suggested fix: provide an exact card/signature reconciliation with behavioral evidence for each non-Jetfire change, or narrow/revert unintended impact.

**Recommendation: request changes — restore compilation, make forced slot selection a real legal replacement with a discriminating runtime regression, and reconcile or narrow the parser impact before re-review.**

JacobWoodson:
@matthewevans — here's the exact reconciliation for the parse-diff (58 cards / 21 signatures). Every change falls into one of three buckets; none is unintended, and the non-Jetfire ones are behavior-preserving or correct building-block generalizations.

### Bucket 1 — role-model rename, behavior-preserving (the 14 "Modified fields" signatures, ~50 card-entries)

Every "modified" signature is the **same** paired change: `target: <X> → ∅` **and** `mana recipient|count source: ∅ → <X>`. This is exactly the PR's core: `Effect::Mana.target` changed type from `Option<TargetFilter>` to `Option<ManaTargetRole>`, so the coverage signature now renders the mana target's **role** instead of a generic `target`. The underlying targeting is byte-identical — only the rendered label moved from `target` to the typed role field. No behavior change.

| Group | Old → New | Role | Examples |
|---|---|---|---|
| 20 | `target: parent target's controller` → `mana recipient` | Recipient | Blighted Burgeoning, Dawn's Reflection (land auras) |
| 12 | `target: triggering player` → `mana recipient` | Recipient | Dictate of Karametra, Barbflare Gremlin |
| 7 | `target: scoped player` → `mana recipient` | Recipient | Belbe, Blinkmoth Urn |
| 5 | `target: player` → `mana recipient` | Recipient | Bigger on the Inside, Color Pie |
| 4 | `target: opponent` → **`mana count source`** | CountSource | Carpet of Flowers, Jeska's Will |
| 3 | `target: chosen player` → `mana recipient` | Recipient | Spectral Searchlight, Stadium Vendors |
| 3 | `target: controller` → `mana recipient` | Recipient | Priest of Forgotten Gods, Organ Harvest |

The 4-card `opponent → count source` group is the whole reason for the PR: those cards' targeted player is a **count source**, not a recipient — the exact conflation the role model removes. The rest are recipients. Same targeting, correctly labelled.

### Bucket 2 — correct generalizations of the two mana building blocks (3 cards)

Each went from an unparsed `add` (Unimplemented) to a correct `Mana` effect, via the same building blocks Jetfire introduced. Oracle text confirms each is correct, not merely different:

- **Su-Chi Cave Guard** — "When this creature dies, **add eight {C}**. Until end of turn, you don't lose this mana…" → `Mana({C} x8, expiry=EndOfTurn)`. The **word-number colorless count-prefix** ("eight {C}") — the same arm as Jetfire's "that much {C}", generalized to fixed word counts.
- **Rasputin, the Oneiromancer** — "Remove one or more dream counters: **Add that much {C}**" → `Mana({C} x event amount, activated)`. This is **literally Jetfire's front-face clause** ("Remove one or more +1/+1 counters … : … adds that much {C}") — same `EventContextAmount` path.
- **Mana Seism** — "Sacrifice any number of lands, then **add that much {C}**" → `Mana({C} x event amount)`. Same "that much {C}" clause.

These are "build for the class, not the card" working as intended: the Jetfire arms correctly unlock every card in the colorless-count mana class. The `➕ Added` / `➖ Removed` "add" signatures net to these three plus Jetfire's two faces.

### Bucket 3 — the target card (Jetfire, both faces)

- `Adapt (counters=3)` → Jetfire, Air Guardian (back face).
- `Mana(… mana recipient=player, restrictions=[SpellTypeOrAbilityActivation{Artifact, Any}])` → Jetfire, Ingenious Scientist (front face).

### Summary

54 cards = the role rename (behavior-preserving). 3 cards = correct colorless-count generalizations with Oracle-text evidence above. 2 = Jetfire. Nothing unintended; nothing to narrow or revert.

---

Separately, both HIGH blockers from the latest review are already resolved on my fork branch [`JacobWoodson:jetfire-rebased`](https://github.com/JacobWoodson/phase), which is rebased onto current `main` (linear, no merge commits) and green (`phase-engine` lib 18,036 / integration 4,216 / `clippy --workspace --all-targets` clean):

- **The `ability_utils.rs:2643` compile break** (E0308, `EngineError` vs `TargetSlotBuildError`) — fixed by returning `no_legal_target_slots()` like the sibling arms.
- **The forced-retarget `current != new` refinement** — commit `46c302de` (patch in my previous comment), with the overlapping-role `[A,B]`→force `A`→`[A,A]` regression, revert-verified.

That branch also carries the one adaptation main's `phase-ai` addition needs (`ManaTargetRole::Recipient` wrapper in `zone_eval.rs`). Happy to get it onto the PR head in whatever way works for you.

matthewevans:
**Maintainer update applied; current-head hold.** I merged current `main` into this branch and resolved the sole conflict in the generated `integration_cards.json` fixture by retaining the current-main generated artifact. The updated head is `e14eb2c43400f57d3a0d3531e30500cd0519a88d`.

Current CI has restarted and is not complete yet, so this head is not ready for approval or enqueue. The prior requested-changes verdict remains in force pending a fresh current-head review of the mana target-slot/forced-retarget behavior and current CI/parse evidence.

matthewevans (changes requested):
**[HIGH] Forced multi-role retargeting can still resolve as an illegal no-op.**

Current head: `e14eb2c43400f57d3a0d3531e30500cd0519a88d`.

`change_targets::resolve` accepts the first target in the forced filter that is in the flattened legal pool ([lines 71–90](https://github.com/phase-rs/phase/blob/e14eb2c43400f57d3a0d3531e30500cd0519a88d/crates/engine/src/game/effects/change_targets.rs#L71-L90)). `forced_retarget_targets` then chooses the first compatible role slot without requiring that slot's existing target differ ([lines 146–165](https://github.com/phase-rs/phase/blob/e14eb2c43400f57d3a0d3531e30500cd0519a88d/crates/engine/src/game/effects/change_targets.rs#L146-L165)).

For a `Both` mana ability with `[recipient=P1, count_source=P2]`, forced-to `P1` selects slot 0 and writes `P1` back into it. That leaves both targets unchanged even though slot 1 can legally change from P2 to P1. CR 115.7a requires a changed target to become **another** legal target; CR 115.7b permits exactly one such change.

Select a compatible slot whose current target differs from the forced target, and leave the original targets unchanged only when no slot can be changed. Add the P1/P2 → forced-P1 regression beside the existing forced-retarget test; it must assert `[P1, P1]`, not merely that the two-slot list survives.

The prior target-collapse fix is covered, but its current P3 fixture only exercises a different recipient and cannot detect this no-op branch. Current CI is also still in progress; this functional blocker is independent of that pending evidence.

JacobWoodson:
@matthewevans — to save you the round-trips: [`JacobWoodson:jetfire-rebased`](https://github.com/JacobWoodson/phase/tree/jetfire-rebased) is a single clean solution to all three blockers on this head, and I'm offering to make it the PR head if you're open to it.

It's the Jetfire work **rebased into a linear history on `main`** (no merge commits), which is what makes the parser diff auditable — the net diff is only the Jetfire changes (the reconciliation in my previous comment maps every one of them). On top of that it carries:

- The `ability_utils.rs` **compile fix** — the multi-role mana slot arm returns `no_legal_target_slots()` instead of `EngineError::ActionNotAllowed`, so it builds against current `main`.
- The **forced-retarget `current != new` refinement** (`46c302de`) with the overlapping-role `[A,B]`→force `A`→`[A,A]` regression, revert-verified.
- The `phase-ai/zone_eval.rs` `ManaTargetRole::Recipient` wrapper main's new test needs.

It was green when built (`phase-engine` lib + integration + `clippy --workspace --all-targets`). `main` has moved since, so if you want it as the head I'll **rebase it onto whatever `main` tip is current at that moment** and re-run the full gate before it lands — the fixture step is now a single `node scripts/migrate-mana-target-roles.mjs` (it fail-loud-guards any new mana-target card, which already caught The Warring Triad).

Because making it the head means replacing your `803e6905`, I won't touch the PR branch without your go-ahead. Two ways to do it:

- **You:** `git fetch https://github.com/JacobWoodson/phase jetfire-rebased && git push origin +jetfire-rebased:claude/jetfire-characters-3b456b` (after a freshen).
- **Me:** say the word and I'll rebase it current, verify green, and force-push it to the PR branch.

If you'd rather keep driving `803e6905`, everything's already in your hands — the compile fix, the CR 115.7a patch, and the reconciliation — and I'll stand down on the branch.

matthewevans:
Thanks for the proposed `jetfire-rebased` solution. I re-reviewed the PR's actual current head, `e14eb2c43400f57d3a0d3531e30500cd0519a88d`; it remains unchanged, so the proposal has not resolved this PR's formal blocker.

`forced_retarget_targets` still chooses the first compatible role slot without requiring `current_targets[slot] != new_target`, then writes the target back into that slot. Thus `[P1, P2]` forced to `P1` still selects slot 0 and no-ops although slot 1 can legally become `P1`. The current required Rust/card-data checks are also failing on this head.

I am retaining the existing `CHANGES_REQUESTED` review. Please push the proposed solution (including the `[P1, P2] -> [P1, P1]` runtime regression and its current-head checks) to this PR, or open the rebased branch as a replacement PR for review.

JacobWoodson:
@matthewevans — pushed the solution to this PR head (`e14eb2c4..e93b4732`, fast-forward on top of your merge, nothing force-pushed). Both HIGH blockers are now resolved on the branch:

- **`ability_utils.rs` compile break** (E0308, `EngineError::ActionNotAllowed` vs `TargetSlotBuildError`) — the multi-role mana slot arm now returns `no_legal_target_slots()` like its sibling arms, so it builds against current `main`.
- **Forced-retarget CR 115.7a refinement** — `forced_retarget_targets` now picks the first surfaced slot whose filter legally accepts the candidate **and** whose current target actually differs from it. "Change a target" must change it to *another* legal target, so a slot already holding the candidate is skipped in favor of one that can genuinely change. Added the overlapping-role regression: targets `[P1, P2]`, force `P1` → `[P1, P1]` by changing the count-source slot (revert-verified — dropping the `changes` guard leaves `[P1, P2]` and fails).
- **`phase-ai/zone_eval.rs`** targeted-mana test adapted to the `ManaTargetRole::Recipient` model.

Green locally: `cargo fmt`; `clippy --workspace --all-targets` clean; `phase-engine` integration `forced_retarget` tests pass; `phase-ai` targeted-mana test passes.

matthewevans (changes requested):
Current-head review for `e93b47320d11198cfd70bd4de7734e4f6c3e2d51`:

The CR 115.7a forced-retarget correction is now present and the new overlapping-role regression covers the previously reported no-op. The PR still cannot merge because the schema migration is incomplete.

`Effect::Mana.target` now deserializes as `ManaTargetRole`, but the tracked integration fixture has not been migrated: it still contains 12 legacy target objects and zero `{ role: ... }` envelopes. The current CI merge commit consequently fails both Rust test shards while loading the fixture with `missing field 'role'` (for example, `temporal_anchor...`, run 30728170186). The migration script itself is not sufficient unless its generated fixture output is committed.

Please run the intended fixture migration/regeneration, commit the resulting `crates/engine/tests/fixtures/integration_cards.json`, and rerun the required Rust checks. I will re-review the resulting head.

JacobWoodson:
Follow-up: the required **Rust test shards** were red on the head, and the cause was a stale test fixture, not the retarget logic. Pushed `e93b4732..2e274b1c`.

`crates/engine/tests/fixtures/integration_cards.json` still carried the legacy `Effect::Mana.target: Option<TargetFilter>` shape — the main-merge regenerated it from `main` without re-running the mana-target-role migration — so every fixture-loading test panicked with `missing field 'role'` (both shards).

Re-running the migration surfaced a second, latent defect in the migration script itself: the fixture now carries a u64 sentinel (a `SpecificObject` id of `u64::MAX`, `18446744073709551615`) that arrived on `main` via the Jailbreak fix (#6691). The script's `JSON.parse`/`JSON.stringify` round-trip silently rewrote it to a lossy float (`1.8446744073709552e+19`), which serde rejects as `invalid type: floating point, expected u64`. I made the fixture-mode round-trip **bigint-lossless** — a string-aware scanner quotes 16+ digit integer values behind an ASCII sentinel across parse/stringify, tracking JSON string state so it never touches `[`/`,`+digit sequences inside card text (the fragility a naive regex would have).

Verified locally after the fix:
- migration migrates 12 entries fail-loud clean; u64 sentinel survives as an integer, no float leak;
- previously-failing `phase-ai` subtlety scenario passes; engine `analysis::corpus_tests` (46), `ai_support::candidates` (43), `analysis::ability_graph` (42) all pass;
- `forced_retarget` integration tests (2) still pass; `clippy --workspace --all-targets` clean.

Note: the two hour-long **AI/perf gates** (Decision-cost, Paired-seed AI) were already red before this and this PR changes no AI-behavior or hot-path code — happy to dig in if you consider them in scope for this PR.

matthewevans (changes requested):
Current-head review for `2e274b1cf9180d18a6738f193fd929d3183f8dfc`:

The curated fixture migration fixes the previous blocker, and the earlier CR 115.7a retarget correction remains correct. Required Rust CI is still red because the persisted-game migration is incomplete.

The current head migrates four `*.json.gz` game-state fixtures, but two tracked fixtures still contain legacy `Effect::Mana.target` values without a `ManaTargetRole.role` field:

- `crates/engine/tests/fixtures/witherbloom_sprout_lumaret_simple_4p.json.gz` (four Priest of Forgotten Gods entries)
- `crates/engine/tests/fixtures/dina_conqueror_4p.json.gz` (Priest of Forgotten Gods and The Warring Triad entries)

This is exactly what the current required test run reports: both shards fail deserializing real game states with `missing field 'role'` (run 30731083891; e.g. the loop-shortcut and Dina scenarios). The migration script already has `--state` handling and explicit mappings for these state-only cards; apply it to both remaining fixtures, commit the regenerated gzip files, and let the current-head Rust checks complete successfully.

JacobWoodson:
Update: the fixture fix resolved most of the shard failures but exposed a *second* stale-fixture class — two persisted 4p **GameState dumps** still carried the legacy mana shape, so restoring them panicked with `missing field 'role'` (6 tests across both shards). Pushed `2e274b1c..bee266df`.

- `dina_conqueror_4p.json.gz` (8 mana targets) → `analysis::resource::tests::dina_untargeted_drain_4p_...`
- `witherbloom_sprout_lumaret_simple_4p.json.gz` (4) → the 5 `loop_shortcut::*collapse*` tests

A GameState deserializes through typed serde before any restore-time migration hook runs, so the dump itself has to carry the role encoding. Re-migrated both via the script's `--state` mode (all roles → Recipient: Priest of Forgotten Gods, The Warring Triad), and applied the same bigint-lossless round-trip to `--state` mode since these dumps carry u64 object-id sentinels.

To make sure this was the last of it, I swept **every committed `.json`/`.json.gz` in the repo** (169 files) for a `Mana` effect whose `target` lacks a `role` — **zero** remain. So the card fixture plus these two dumps were the complete set.

Verified locally: the dina resource test and all 71 `loop_shortcut` integration tests pass (including the 6 that were red); both dumps re-scan with zero stale targets and no float/sentinel corruption. CI is re-running on `bee266df`.

matthewevans (approved):
Current-head review complete: persisted fixtures are migrated, required checks are green, and prior retarget/schema findings are resolved.

PR #6064, opened 2026-07-17
fix(engine): keep hidden card information private

nornagon:
## Summary

Fixes information-leak paths involving mulligans, Foretell, stable object identifiers, and library state. Hidden cards and their identities remain private to the appropriate player while preserving legal actions and the current replacement-aware Foretell movement path. Each commit records a concrete Before/Rule/After example.

No directly matching Phase issue was found in the repository issue search.

### CR 400.2 — hidden-zone information

1. **Opponent mulligans:** a card moved from hand to library during a mulligan exposes the movement but not its name or stable identity to an opponent. (cbe92354b7f29d22256b388b0daddbbdb514ae0b)
2. **Public game log:** hidden hand-to-library mulligan moves are omitted from the public log, while genuinely public discards remain visible. (919c0bab96f5d1f7f90d2b22b5db9be88278921e)
3. **Stable spell identities:** opponent-facing events can reference an observable spell on the stack without exposing a persistent hidden-card identifier that would allow the card to be tracked across zones. (21282749dee20afdf0c4c924c92c9bcde99a592a)

### CR 702.143a — face-down Foretell exile

1. **Foretelling a card:** opponents can observe that a card was foretold and moved to exile, but the exile event and public log do not reveal its name or stable identity. (19c1fc517d124c3cc6129e093488e18c26241c9c)

### CR 401.2 and 701.20 — hidden, randomized library order

1. **Library identifiers:** object IDs are allocated using a deterministic replay-safe shuffle instead of submitted deck-list order, preventing observers from inferring hidden card identity or library order from otherwise opaque IDs. (3f0bba53fd23571603c30a4001cca67cd64ec568)

## Validation

- `cargo fmt --all -- --check`
- `git diff --check upstream/main...HEAD`
- `CARGO_PROFILE_DEV_DEBUG=0 cargo check -p engine --lib`

### DIFF
--- a/crates/engine/src/game/casting_tests.rs
+++ b/crates/engine/src/game/casting_tests.rs
@@ -898,15 +898,10 @@ fn foretell_special_action_exiles_and_grants_later_turn_permission() {
     let mut state = setup_game_at_main_phase();
     let object_id = add_foretell_sorcery(&mut state);
     add_mana(&mut state, PlayerId(0), ManaType::Colorless, 2);
+    let mut events = Vec::new();
 
-    let waiting = handle_foretell(
-        &mut state,
-        PlayerId(0),
-        object_id,
-        CardId(143),
-        &mut Vec::new(),
-    )
-    .unwrap();
+    let waiting =
+        handle_foretell(&mut state, PlayerId(0), object_id, CardId(143), &mut events).unwrap();
 
     assert_eq!(
         waiting,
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Approved after current-head review: the hidden-information redaction paths, including the supplementary deck ID allocation and face-down hand-to-exile visibility paths, are covered at their existing authorities. Current required checks are green and the parse-diff reports no card-parse changes.

PR #6072, opened 2026-07-17
fix(engine): scan the Aura's own enchant-zone when choosing a host on non-cast entry

rykerwilliams:
## Summary

CR 303.4f: when an Aura enters the battlefield by any means other than resolving as a spell (e.g. returning from exile) and the effect putting it there doesn't specify a host, the controller chooses a legal object per the Aura's current Enchant ability. `legal_aura_attachment_targets` computed that candidate list from `state.battlefield` only, unconditionally — even though both call sites already had the Aura's `enchant_filter` in hand, and that filter is the single source of truth for which zone(s) it can legally target.

An Enchant ability that targets "a creature card in a graveyard" (Animate Dead, Dance of the Dead), "an instant card in a graveyard" (Spellweaver Volute), or "a card in your hand" (Don't Worry About It) could therefore never find a legal host through this path, even when one genuinely existed — the consult always saw zero candidates and the Aura stayed stranded in its current zone (CR 303.4g).

**Fix**: rewrite `legal_aura_attachment_targets` to resolve the search zone(s) from `enchant_filter.extract_zones()` (defaulting to the battlefield when the filter imposes no zone constraint — the common case, unaffected), flat-map `zone_object_ids` over those zones, and filter with `matches_target_filter` — mirroring the existing `object_count_matching_ids` pattern in `game/quantity.rs`. Uses `matches_target_filter` rather than the spell-targeting enumerator: CR 115.1b says an Aura permanent doesn't target anything, so hexproof/shroud must never remove a candidate here, only an actual Enchant restriction can. Preserves the pre-existing "Enchant player" branch, the CR 303.4d self-enchant exclusion, and the `can_attach_to_object` legality gate unchanged.

Incidentally also now excludes phased-out permanents from the default battlefield case (CR 702.26b) via `zone_object_ids`'s existing filter — strictly more correct, not a behavior regression.

This is the fix that closes the final leg of the Worldgorger Dragon self-loop combo (mass-return mechanism fixed separately in #6055): the returning Animate Dead/Necromancy Aura can now find WGD sitting in the graveyard and re-attach to it, rather than staying stranded in exile.

## Test plan

- [x] New test: a reanimator-class Aura re-entering from exile with a legal graveyard host finds it, attaches, and its ETB reanimation chain fires end-to-end (creature reaches the battlefield under the correct controller)
- [x] New test: an ordinary battlefield-scoped Aura is unaffected — attaches iff a legal battlefield host exists (both a has-host positive and a no-host negative in one fixture)
- [x] New test: a graveyard-scoped Aura with NO legal graveyard creature correctly stays in exile (CR 303.4g) — non-vacuous, a non-matching card is present and enumerated/rejected by the scan
- [x] Zero regression: full `exile_return_tests` (16/16), `animate_dead`-filtered casting tests (9/9), `necromancy`-filtered casting tests (5/5)
- [x] Live-revert-and-rerun proof (done independently by two people): reverting just the scan-logic rewrite reproduces the exact failure (Aura stuck in exile); restoring is green
- [x] `cargo fmt --all`, `cargo clippy -p engine --all-targets -- -D warnings` clean
- [x] CR-annotation gate clean (115.1b, 303.4d, 303.4f, 303.4g, 610.3a, 702.11, 702.18, 702.26b all verified against the Comprehensive Rules)

### DIFF
--- a/crates/engine/src/game/engine_exile_return_tests.rs
+++ b/crates/engine/src/game/engine_exile_return_tests.rs
@@ -1295,3 +1295,385 @@ fn exile_return_combines_normal_and_delayed_triggers_in_one_ordering_prompt() {
         .iter()
         .any(|summary| summary.source_id == delayed_source));
 }
+
+// ---------------------------------------------------------------------------
+// CR 303.4f/303.4g regression: a non-cast Aura re-entering the battlefield must
+// be able to find a legal host in a NON-battlefield zone (graveyard/hand) when
+// its Enchant ability points there (Animate Dead, Dance of the Dead, Spellweaver
+// Volute, Don't Worry About It). Before the `legal_aura_attachment_targets`
+// rewrite the candidate scan only looked at the battlefield, so such an Aura
+// could never find a host through this path and was stuck in exile forever.
+// ---------------------------------------------------------------------------
+
+/// Verbatim Animate Dead Oracle text (matches the repo's canonical corpus form
+/// in `crates/engine/tests/fixtures/integration_cards.json`, mirroring the
+/// `casting_tests.rs` reanimation fixtures).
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
Maintainer review: correct zone-authority fix with discriminating exile-return coverage; required checks are green.

PR #6076, opened 2026-07-17
feat(engine): chosen-name-scoped land-play prohibition (Conjurer's Ban) (#6075)

### LINKED ISSUE #6075: parser+engine: Conjurer's Ban's land-play half strict-fails - no filter-scoped CR 305.1 restriction exists (chosen-name axis)
## Summary

**Conjurer's Ban** (verified Oracle text via Scryfall):

> Choose a card name. Until your next turn, spells with the chosen name can't be cast and lands with the chosen name can't be played.
> Draw a card.

Live-parsed on current `main` (confirmed against `data/card-data.json`): the entire restriction sentence collapses to a single `Effect::Unimplemented{name:"spells", description:"spells with the chosen name can't be cast and lands with the chosen name can't be played"}`. The card currently only lets you choose a name and draw a card — the restriction does nothing at all, on either half.

## Root cause — a missing axis, not a missing card

This isn't a one-off parser miss. `ProhibitedActivity` (`crates/engine/src/types/ability.rs`), the runtime type behind temporary floating restrictions created by resolving spells (Silence: `"Your opponents can't cast spells this turn"` → `Effect::AddRestriction`), already has a filter-scoped variant for the **casting** axis:

```rust
CastSpells { spell_filter: Option<TargetFilter> }   // CR 101.2 / CR 601.2a
```

enforced by `is_blocked_by_cant_cast_spells_for` (`crates/engine/src/game/casting.rs`), which already supports `HasChosenName` (Pithing-Needle-style card-identity filtering) for a temporary spell-sourced ban.

There is **no equivalent for CR 305.1 land-playing**. The only two land-play prohibition mechanisms today are:
- `StaticMode::Other("CantPlayLand")` — a blanket, **player**-scoped permanent static (Aggressive Mining, Worms of the Earth, Limited Resources, Rock Jockey), no card filter at all.
- `ProhibitedActivity::ProhibitPlayFromZone { zone: Zone }` — a temporary restriction, but **zone**-scoped only (Memory Vessel), no card-identity filter.

Neither can express "lands **with the chosen name**, specifically, can't be played" — the land-play sibling of `CastSpells{spell_filter}` was simply never built. This mirrors the exact class of gap fixed by #5219 / #5406 (a proven axis-parameterized sibling exists for one form but a categorically-adjacent sibling never got the same treatment) — except on the "filter-scoped deny axis × CR 305.1" cell instead of the "compound subject" axis.

At the parser layer, `crates/engine/src/parser/oracle_effect/mod.rs`'s `try_parse_cant_cast_spells_effect` (the Silence-class recognizer) only handles the **active-voice**, explicitly player-scoped form ("Your opponents can't cast spells..."). It has no arm for the **passive**, card-scoped, implicitly-all-players form Conjurer's Ban uses ("spells with the chosen name can't be cast") — the same active/passive split the *static* layer already has as two separate functions (`parse_cant_cast_type_spells` vs `parse_passive_cant_be_cast` in `restriction.rs`), just never built at the *effect* layer. There is no land-play recognizer of any card-scoped kind at the effect layer at all.

## Proposed fix

- `ProhibitedActivity::PlayLands { land_filter: Option<TargetFilter> }` (CR 305.1) — the land-play sibling of `CastSpells{spell_filter}`. Categorically a separate variant, not a parameterization onto `CastSpells`: CR 601.2a (casting) and CR 305.1 (land-playing, an explicit special action per CR 116.2a, not a cast) are different rule sections — the same reasoning `ProhibitPlayFromZone`'s own doc comment already gives for staying separate from `CastOnlyFromZones`.
- Runtime: `is_blocked_by_cant_play_lands`, structural sibling of `is_blocked_by_cant_cast_spells_for`, consulted from `handle_play_land` (alongside the two existing checks there) and from the AI legal-action generator (`ai_support/candidates.rs`), which currently only gates the blanket `CantPlayLand` static and would otherwise offer an illegal `PlayLand` candidate for the specifically-named land.
- Parser: passive-voice recognizers for both halves ("spells with `<filter>` can't be cast" / "lands with `<filter>` can't be played"), composed via the same `sub_ability`-chaining shape `try_parse_cant_cast_spells_effect` already uses for trailing continuations, so the class covers any `parse_type_phrase`-resolvable subject — not just "chosen name" literally.

No new `Layer`/CDA/targeting infrastructure — this reuses the entire `GameRestriction`/`ProhibitedActivity` floating-restriction pipeline that Silence already exercises end-to-end; only the land-play axis's filter capability and its passive-voice parser entry are new.

## CR references

- CR 305.1 — playing a land is a special action.
- CR 116.2a — special actions, land-playing specifically.
- CR 601.2a — casting a spell.
- CR 101.2 — "can't" effects take precedence.

## Notes

No existing open issue tracks this (checked `compound`, `strict-fail`, `Conjurer`, `land chosen name`, `lands can't be played` searches). Not a duplicate of #5392/#5246/#5219 (those are the `StaticMode`/chosen-*type* class on permanents; this is the `ProhibitedActivity`/temporary-restriction class on a resolving sorcery, and the axis gap is filter-scoping-on-land-play, not compound subjects).

claytonlin1110:
## Summary

Fixes #6075. **Conjurer's Ban**'s land-play half — "Choose a card name. Until your next turn, spells with the chosen name can't be cast and lands with the chosen name can't be played.\nDraw a card." (Oracle text verified against Scryfall) — previously collapsed the entire restriction sentence into a single `Effect::Unimplemented`; the card only let you choose a name and draw a card.

It's the CR 305.1 (land-play) sibling of the filter-scoped deny axis `ProhibitedActivity::CastSpells { spell_filter }` (CR 601.2a, casting) already uses for temporary spell-sourced restrictions like Silence. That axis had no land-play equivalent — only a blanket player-scoped static (`StaticMode::Other("CantPlayLand")`) or a zone-scoped temporary restriction (`ProhibitPlayFromZone`) existed, neither of which can express "lands **with the chosen name**, specifically, can't be played."

## Fix

- **`ProhibitedActivity::PlayLands { land_filter }`** — the land-play sibling of `CastSpells { spell_filter }`. Kept as its own variant (not a parameter on `CastSpells`) per the categorical-boundary precedent `ProhibitPlayFromZone` already sets against `CastOnlyFromZones`: CR 601.2a (casting) and CR 305.1 (land-playing, an explicit special action per CR 116.2a) are different rule sections. Ran this through the `add-engine-variant` gate (existence check, sibling-cluster check, categorical-boundary check) before adding it.
- **`is_blocked_by_cant_play_lands`** — structural sibling of `is_blocked_by_cant_cast_spells_for`, reusing the existing generic per-object filter evaluator (`filter::matches_target_filter` + `FilterContext`) rather than a spell-record projection (a land isn't cast). Wired into `handle_play_land` alongside the two existing checks there, and into AI legal-action generation (`ai_support/candidates.rs`), which previously only gated the blanket static — without this it would offer an illegal `PlayLand` candidate for the specifically-named land.
- **Parser**: `try_parse_passive_cant_be_cast_effect` is the passive-voice, card-scoped effect-layer sibling of `try_parse_cant_cast_spells_effect`'s active-voice form — mirroring the active/passive split `oracle_static::restriction` already has for the *static* layer (`parse_cant_cast_type_spells` vs `parse_passive_cant_be_cast`), which the *effect* (temporary, spell-sourced) layer never had. Extracted `parse_passive_cant_be_cast_spell_filter` out of the existing static-layer function so both layers share the identical subject grammar ("[type] spells", "spells with mana value N or greater/less", "spells with {X} in their mana cost", "spells with the chosen name") instead of duplicating it. `parse_passive_cant_be_played_land_filter` is the land-play analog (no mana-value/X-in-cost sub-patterns — those are card-casting concepts with no CR 305 analog).
- The compound "spells ... can't be cast **and** lands ... can't be played" composes via the same `sub_ability` tail-chaining shape `try_parse_cant_cast_spells_effect` already uses for trailing continuations — confirmed via `assembly.rs`'s chain-folding step (`assemble_effect_chain`, ~line 2287-2318) that it walks to the *deepest* `sub_ability` before attaching the next top-level clause ("Draw a card."), so the pre-existing chaining machinery composes the full card correctly with zero additional plumbing. The land sub-ability's `duration` is set explicitly from the parent's (not re-derived), since `AddRestriction` resolution (`add_restriction.rs::fill_runtime_fields`) reads `ability.duration` per sub-ability independently — an unset child duration would silently fall back to end-of-turn instead of "until your next turn."

## Architecture

- Nom combinators throughout; reuses `parse_type_phrase`, `TextPair::split_around`, `restriction_scope_matches_player`, `filter::matches_target_filter`.
- Builds for the **class**: any future card templated as "`<X>` spells can't be cast" (bare passive, effect layer) or "`<X>` lands can't be played" (any type-phrase-resolvable filter, not just chosen name) is unlocked, not just Conjurer's Ban.
- No new `Layer`/CDA/targeting infrastructure — reuses the entire `GameRestriction`/`ProhibitedActivity` floating-restriction pipeline Silence already exercises end-to-end.
- CR-annotated throughout: **CR 305.1 / CR 116.2a / CR 601.2a / CR 101.2 / CR 201.2**, each verified against `docs/MagicCompRules.txt` before use.

## Testing

`cargo fmt --package engine` ✓ (clean).

**Verification note:** this environment has no working MSVC linker (Visual Studio Build Tools not installed) and Tilt isn't running, so I could not get a `cargo check`/`clippy`/`test` pass in this session — `cargo fmt` (parse-only, no linking) is the only direct verification available here. I did an extensive manual review instead (cross-checked every type/field/signature against the actual source, traced the exact chain-assembly mechanism in `assembly.rs` to confirm the compound sub-ability composition is correct, and caught/fixed one real bug this way — a use-after-move on `expiry`, which isn't `Copy`). Please run the normal Tilt-based gate (`clippy`, `test-engine`) before merging; I'd appreciate a careful compile-and-test pass given I couldn't do one myself.

New tests (`oracle_effect/tests.rs`):
- `conjurers_ban_compound_cant_be_cast_and_cant_be_played` — the modeled compound sentence → chained `CastSpells{HasChosenName}` + `PlayLands{HasChosenName}`, both sharing the `UntilNextTurnOf` duration.
- `passive_cant_be_cast_single_clause_has_no_land_sub_ability` — no-regression: a bare single-clause passive cast prohibition doesn't fabricate a dangling land sub-ability.
- `passive_cant_be_played_land_type_filter` — pattern-coverage: the land-play axis isn't limited to "chosen name" (a hypothetical type-scoped land ban parses the same way).
- `conjurers_ban_full_oracle_has_no_unimplemented` — full card oracle text (Choose → cast prohibition → land-play prohibition → Draw), asserting zero `Unimplemented` residual anywhere in the chain.

New test (`game/engine_tests.rs`):
- `apply_play_land_rejects_under_cant_play_lands_chosen_name_filter` — runtime: a land matching the restriction (source sitting in the graveyard, as the resolved sorcery would be — proving the chosen-name binding survives its own source's zone change) is rejected; a differently-named land is not.


matthewevans (changes requested):
**Verdict: changes requested — the new land-play gate is reachable, but the PR has no discriminating test that resolves Conjurer's Ban through its actual choose → restriction → stack-resolution path.**

## 🔴 Blocker

- `crates/engine/src/game/engine_tests.rs:3332-3387` constructs a graveyard source with `ChosenAttribute::CardName` and pushes `GameRestriction::ProhibitActivity` directly. The production behavior instead depends on `Effect::Choose` producing that source-bound name, then `Effect::AddRestriction` creating both chained restrictions at `crates/engine/src/game/effects/add_restriction.rs:14-18`, before `handle_play_land` consults the new gate at `crates/engine/src/game/engine.rs:7675-7689`. This test still passes if the full card chain fails to preserve the name/source binding or fails to install the land sub-ability, so it does not prove Conjurer's Ban works. Add an integration test that casts and resolves Conjurer's Ban, submits the named card, then proves the named land is rejected while a differently named land is legal (and, ideally, exercises the matching spell half) after the sorcery has reached the graveyard. That test must fail if either the `PlayLands` sub-ability or its production gate is reverted.

## ✅ Clean

- The parse-diff is current for `5b8aa74` and scoped to Conjurer's Ban; its added `AddRestriction` and removed former `spells` signature match the stated parser replacement.

**Recommendation: add the full cast/resolution regression, then request a fresh review.**

claytonlin1110:
Addressed — added \`conjurers_ban_full_cast_resolve_blocks_named_land_and_spell\` in \`engine_tests.rs\` (kept the original unit-level test alongside it for isolating the runtime gate specifically).

The new test drives the real pipeline end-to-end via \`GameScenario\`/\`GameRunner\`:
- Casts Conjurer's Ban from its verbatim Oracle text and answers the \`NamedChoice\` prompt through the driver (\`.choose_option("Forest")\`), letting the full chain (\`Choose\` → \`CastSpells\` → \`PlayLands\` → \`Draw\`) resolve for real — no hand-assembled \`GameRestriction\`.
- Asserts the sorcery actually reached the graveyard and the draw happened.
- Then issues two further real actions and asserts on their actual \`Result\`: \`PlayLand\` for the named land (rejected) and a differently-named land (legal), plus \`CastSpell\` for a same-named spell (rejected) — proving both halves of the resolved restriction, not just the land gate in isolation.

This fails if \`Effect::Choose\` stops binding the name onto the resolving source, if \`Effect::AddRestriction\` stops installing the \`PlayLands\` sub-ability during resolution, or if \`handle_play_land\`'s gate is reverted.

\`cargo fmt --package engine -- --check\` clean. Same environment caveat as before applies — no working compiler in this session, so I traced the exact scenario/runner API (\`SpellCast::choose_option\`, \`Outcome::assert_zone\`, \`GameRunner::act\`) against source rather than compiling.

matthewevans (approved):
Approved after current-head maintainer review: the Conjurer's Ban resolution-path regression covers the chosen-name spell and land restrictions, and the maintainer import fixup is confined to the existing seams.

matthewevans (approved):
Approved on the refreshed head: cast restrictions now evaluate chosen-name filters through the existing source-aware authority, matching the land-side resolution path and the full-pipeline regression.

matthewevans (approved):
Approved on the refreshed head: the source-aware cast-filter repair no longer leaves its obsolete spell-record projection helper behind.

matthewevans (approved):
Approved on the refreshed head: the test module now owns its `SpellCastRecord` import rather than relying on the production module's private import scope.

PR #6078, opened 2026-07-17
feat(parser): Ward of Bones per-player quantity cast prohibition + same-is-true spell widening

### LINKED ISSUE #6077: Ward of Bones: relative-quantity per-player cast prohibition + 'same is true' spell-type widening not parsed
## Summary

`Ward of Bones`'s per-player prohibitions —

```
Each opponent who controls more creatures than you can't cast creature spells. The same is true for artifacts and enchantments.
Each opponent who controls more lands than you can't play lands.
```

(Oracle text verified against the Scryfall API, `oracle_id: c3ea1497-63ac-46bb-95e9-d98b93a880b3`) — are not modeled by the parser today. No test or dispatch branch references the card, and it isn't in the "claimed parse impact" list of any merged PR.

Root cause: `parse_per_player_conditional_prohibition` (`crates/engine/src/parser/oracle_static/restriction.rs`) is the single handler for the CR 101.2 "Each [scope] who [did X] can't [Y]" class (Angelic Arbiter is the existing member), but its per-player predicate is a fixed 3-arm `alt()` of **turn-activity** predicates only (`attacked with a creature this turn`, `attacked you ... this turn`, `cast a spell this turn`). Ward of Bones's predicate is a **relative permanent-count comparison** ("controls more creatures than you") — a different axis the fixed `alt()` doesn't reach, so the whole line falls through to `Unimplemented`.

This is the same *class* of gap PR #5406 fixed for the chosen-type static family (a hardcoded fixed-arm matcher missing a reachable-but-unmodeled variant) — here the fixed arm is the per-player *condition*, and the same static also needs the "The same is true for X and Y" compound-clause widening PR #5789 generalized for type-changing statics, but applied to a spell-type-restriction static instead.

## Fix approach

- Add a `parse_controls_more_than_you_condition` per-player condition arm, built from **already-wired** primitives: `ParsedCondition::QuantityComparison` comparing `QuantityRef::ObjectCount` scoped via `ControllerRef::ScopedPlayer` (the affected candidate — CR 115.10) against the same count scoped via `ControllerRef::You` (the source's actual controller, per `game::quantity::resolve_quantity_scoped`'s documented split). No new `ParsedCondition` variant, no new evaluator arm — `QuantityComparison` is already generically evaluated by `game::restrictions::evaluate_condition`, and `per_player_condition` is already read at the one call site that matters here (`casting::is_blocked_by_cant_be_cast_for`).
- Add a `parse_cast_typed_spells_predicate` verb arm — the typed sibling of the existing bare "cast spells" arm — extended with an optional "the same is true for `<type>` and `<type>`" continuation (mirrors `same_is_true::parse_sentence`'s two-conjunct "and"-only grammar) that widens the restricted spell types into a single `Or`.

## Scope note

The third clause ("Each opponent who controls more lands than you can't play lands") reuses `StaticMode::Other("CantPlayLand")`, whose runtime check (`check_static_other_by_name`) does not currently read `per_player_condition` at all (only `StaticMode::CantBeCast`/`CantAttack` do, via a different call site). Wiring that up is a separate, slightly wider-blast-radius runtime change (it's a shared function used by several other `Other`-keyed statics) and is left as an explicit follow-up rather than bundled here.

## Verification

- Oracle text confirmed via `https://api.scryfall.com/cards/named?exact=Ward%20of%20Bones` (Eventide, 2008; `mtgo_id: 30278`).
- No existing test, dispatch branch, or "claimed parse impact" list references this card.

jsdevninja:
## Update

While rebasing this branch, `restriction.rs` conflicted with an already-merged fix on `main` that implements this exact card — more completely and more correctly than this PR's original approach:

- It also covers the land-play clause ("Each opponent who controls more lands than you can't play lands"), which this PR had explicitly deferred.
- Critically, it fixes a **rules bug in my original approach**: this PR had modeled the cast restriction as ONE static with a single `Or[Creature, Artifact, Enchantment]` filter gated on a single (creature) count comparison. Per the card's 2008-08-01 ruling, each type is an **independent** prohibition — an opponent with more artifacts but not more creatures must still be barred from casting artifact spells specifically. The upstream fix (`parse_relative_count_typed_cast_prohibitions`) correctly emits one `CantBeCast` static per type, each gated on its own count comparison, and ships with real game-scenario integration tests (`crates/engine/tests/integration/ward_of_bones_relative_count_cast_prohibition.rs`, `ward_of_bones_relative_count_land_prohibition.rs`) proving the per-type independence.

I resolved the conflict by taking the upstream implementation as-is and dropping this PR's now-superseded/incorrect functions and shape-tests (they asserted the collapsed-`Or` model, which upstream's own tests demonstrate is wrong). What's left is a small doc-comment correction to `parse_per_player_conditional_prohibition`'s header, clarifying which clause it owns now that the cast-restriction clause moved to the sibling multi-def parser.

Given the underlying issue (#6077) is already fixed on `main`, this PR no longer adds new capability — it's a documentation-accuracy fix only. Deferring to the maintainers on whether to take it or close it as moot.

---

## Original summary (superseded, kept for history)

Fixes #6077. `Ward of Bones`'s per-player cast prohibition — "Each opponent who controls more creatures than you can't cast creature spells. The same is true for artifacts and enchantments." (Scryfall-verified, `oracle_id: c3ea1497-63ac-46bb-95e9-d98b93a880b3`) — was unhandled at the time this PR was opened. The original approach added a `QuantityComparison`-based per-player condition arm and a typed "cast `<type>` spells[. The same is true for X and Y]" verb producing a single widened `Or` filter. That approach is now superseded per the note above.

---

## Update 2 (review response)

@matthewevans's review requesting runtime proof was attached to `fb19028a4` — the pre-conflict-resolution commit described above, which no longer exists on this branch. #6095 (the concurrently-merged fix, referenced above) already ships exactly that proof in `crates/engine/tests/integration/ward_of_bones_relative_count_cast_prohibition.rs` (per-type block/allow assertions through `can_cast_object_now`, plus a Two-Headed Giant variant). One gap remained — no test pinned the `Comparator::GT` (strict) semantics against an equal-count tie — closed in `406dc391c` (`equal_creature_count_does_not_block_cast`).

### DIFF
--- a/crates/engine/src/parser/oracle_static/restriction.rs
+++ b/crates/engine/src/parser/oracle_static/restriction.rs
@@ -968,10 +968,10 @@ pub(crate) fn parse_per_turn_cast_limit(tp: &str, text: &str) -> Option<StaticDe
 }
 
 /// CR 101.2 + CR 109.5 + CR 508.1 + CR 601.3a: "Each [scope] who [did X] this turn
-/// can't [Y]" — a static prohibition gated on a PER-AFFECTED-PLAYER turn-activity
-/// predicate (Angelic Arbiter).
+/// can't [Y]" — a static prohibition gated on a PER-AFFECTED-PLAYER predicate
+/// (Angelic Arbiter, Ward of Bones).
 ///
-/// The two clauses are:
+/// The clauses are:
 /// - "Each opponent who attacked with a creature this turn can't cast spells."
 ///   → `CantBeCast { who: Opponents }` + `per_player_condition: YouAttackedThisTurn`
 ///   (CR 601.3a cast prohibition).
@@ -979,11 +979,23 @@ pub(crate) fn parse_per_turn_cast_limit(tp: &str, text: &str) -> Option<StaticDe
 ///   → `CantAttack` with `affected = opponents' creatures` +
 ///   `per_player_condition: YouCastSpellThisTurn { filter: None }` (CR 508.1
 ///   declare-attackers prohibition).
+/// - "Each opponent who controls more lands than you can't play lands." (Ward
+///   of Bones, line 2) → `StaticMode::Other("CantPlayLand")`, `affected` =
+///   opponents' player scope, `per_player_condition: QuantityComparison`
+///   (`controls_more_than_you_condition` — a relative permanent-count gate,
+///   CR 109.4, rather than a turn-activity predicate).
 ///
-/// The turn-activity predicate is stored in `per_player_condition` (CR 109.5:
-/// evaluated against the AFFECTED player — the caster, or the attacking creature's
-/// controller), NEVER in `condition` (which is the source-relative functioning
-/// gate). `condition` stays `None` so the prohibition is not globally gated.
+/// Ward of Bones's OTHER clause ("controls more creatures than you can't cast
+/// creature spells. The same is true for artifacts and enchantments.") is NOT
+/// handled here — each named type is an independent per-type prohibition, so
+/// it needs a multi-def result and is owned by the sibling
+/// `parse_relative_count_typed_cast_prohibitions` on the multi-static path.
+///
+/// The per-affected-player predicate is stored in `per_player_condition` (CR
+/// 109.5: evaluated against the AFFECTED player — the caster, or the attacking
+/// creature's controller), NEVER in `condition` (which is the source-relative
+/// functioning gate). `condition` stays `None` so the prohibition is not
+/// globally gated.
 ///
 /// Composed from the shared `strip_casting_prohibition_subject` building block plus
 /// nom `tag`/`alt`/`value` — no string-matching dispatch.
--- a/crates/engine/src/parser/oracle_static/restriction.rs
+++ b/crates/engine/src/parser/oracle_static/restriction.rs
@@ -982,8 +982,8 @@ pub(crate) fn parse_per_turn_cast_limit(tp: &str, text: &str) -> Option<StaticDe
 /// - "Each opponent who controls more lands than you can't play lands." (Ward
 ///   of Bones, line 2) → `StaticMode::Other("CantPlayLand")`, `affected` =
 ///   opponents' player scope, `per_player_condition: QuantityComparison`
-///   (`controls_more_than_you_condition` — a relative permanent-count gate,
-///   CR 109.4, rather than a turn-activity predicate).
+///   (`controls_more_than_you_condition` — a relative permanent-count gate
+///   rather than a turn-activity predicate).
 ///
 /// Ward of Bones's OTHER clause ("controls more creatures than you can't cast
 /// creature spells. The same is true for artifacts and enchantments.") is NOT
--- a/crates/engine/tests/integration/ward_of_bones_relative_count_cast_prohibition.rs
+++ b/crates/engine/tests/integration/ward_of_bones_relative_count_cast_prohibition.rs
@@ -208,6 +208,43 @@ fn more_enchantments_blocks_only_enchantment_spells() {
     );
 }
 
+/// Boundary case: P1 controls the SAME number of creatures as P0 (1 vs 1), not
+/// STRICTLY more. CR 109.4's "more than" is a strict inequality
+/// (`Comparator::GT`); the prohibition must not fire on a tie. This is the one
+/// case the per-type tests above never probe (they only ever use a strict
+/// count difference), so it's the discriminating regression guard against the
+/// comparator silently regressing to `GE` (which would wrongly block casts
+/// whenever the counts are merely equal).
+#[test]
+fn equal_creature_count_does_not_block_cast() {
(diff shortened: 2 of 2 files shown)


matthewevans (changes requested):
**Changes requested — the parser shape is correct, but the newly supported cast prohibition lacks runtime proof.**

## 🔴 Blocker

**[HIGH] Add a discriminating runtime cast test for the parsed Ward of Bones restriction.** Evidence: `crates/engine/src/parser/oracle_static/restriction.rs:948-954` installs the new `CantBeCast` static, while its production consumer evaluates `per_player_condition` only in `crates/engine/src/game/casting.rs:17874-17890`. The added tests at `crates/engine/src/parser/oracle_static/tests.rs:19986-20106` and `:20054-20087` inspect parser output only; none invokes the casting gate. Why it matters: they would remain green if the `ScopedPlayer` quantity were bound to the wrong player, the `QuantityComparison` were ignored in the cast path, or one arm of the three-type `Or` did not prohibit a spell. Suggested fix: stage the parsed Ward of Bones static through the public cast path with unequal creature counts, proving an affected opponent is blocked for creature/artifact/enchantment spells, remains able to cast an instant, and becomes unblocked when its count is no longer greater than the source controller's.

## ✅ Clean

The current-head parse-diff is scoped to Ward of Bones: it adds the expected `CantBeCast(opponents)` creature/artifact/enchantment signature and removes only the former `static_structure` fallback. Current required CI is green.

**Recommendation: add the runtime regression and request re-review on the new head.**

matthewevans (changes requested):
**Changes requested — the original Ward of Bones capability already landed in #6095; the current head has no remaining implementation to review or merge.**

## 🔴 Blocker

**[MED] This branch no longer contains the claimed parser or runtime-test change.** Evidence: `git diff origin/main...9d411a5e` changes only `crates/engine/src/parser/oracle_static/restriction.rs:970-1007`, and the real relative-count parser plus discriminating production-path tests are already on `main` from #6095 (`54cae94fa2`). The previous runtime-test finding is therefore satisfied upstream, not by this PR. Why it matters: the remaining documentation-only delta does not add behavior or retire debt proportional to merge-queue and maintenance cost. Suggested fix: close this PR as superseded by #6095; submit a separate, scoped change only if a demonstrable documentation defect remains.

## 🟡 Non-blocking

**[MED] The parse-diff sticky comment is stale for this head.** Evidence: the `<!-- coverage-parse-diff -->` comment is dated 2026-07-17 and reports one parser signature, while the current head was pushed 2026-07-21 and has no parser-code delta. Why it matters: it cannot serve as current-head review evidence. The current CI run will regenerate it, but that does not make the redundant documentation-only PR merge-worthy.

## ✅ Clean

Scryfall confirms Ward of Bones has the two clauses described in the updated comment, and #6095 contains the corresponding runtime cast coverage, including independent creature/artifact/enchantment count cases.

**Recommendation: close as superseded by #6095; do not enqueue this residual diff.**

jsdevninja:
@matthewevans thanks for the review — a heads-up on timing: it's attached to commit `fb19028a4` (the SHA in the review header), which predates the force-push mentioned in the "Update" section above.

By the time this review landed, `fb19028a4`'s code was already gone: while rebasing onto `main`, this branch conflicted with #6095 (merged concurrently), which implements this same card correctly — one independent `CantBeCast` per type instead of the single collapsed `Or[creature,artifact,enchantment]` gated on one shared count that `fb19028a4` had built (and that your review's "why it matters" scenario — one arm of the `Or` not actually being independently gated — was exactly right to be suspicious of; it just turned out to already be fixed rather than needing a new test written against it).

On the substance of the ask — production-path proof through `can_cast_object_now`/`is_blocked_by_cant_be_cast_for` — #6095 already ships exactly that in `crates/engine/tests/integration/ward_of_bones_relative_count_cast_prohibition.rs`: independent block/allow assertions per type (creature-only, artifact-only, enchantment-only), each with reach-guards proving the untriggered types stay castable, plus a Two-Headed Giant teammate-vs-opponent variant.

One gap remained even in #6095's suite: none of its tests probe an EQUAL count (all use a strict difference), so nothing pinned `Comparator::GT` against a silent regression to `GE`. Added `equal_creature_count_does_not_block_cast` to close that.

Net effect on this PR: it no longer changes the cast-prohibition parsing at all (that's #6095, already on `main`) — it's now a doc-comment clarification to `parse_per_player_conditional_prohibition` plus this one boundary-case test. Re-review on `406dc391c` whenever convenient.

matthewevans (approved):
Approved: the current head adds the previously untested equal-count boundary through the production `can_cast_object_now` gate. It is discriminating against `GT` → `GE`, while the existing strict-count cases in this registered module establish the restriction's active path. The maintainer fixup removed the incorrect CR 109.4 citations; that rule concerns object controllers, not comparison strictness.

PR #6087, opened 2026-07-17
Add opt-in game-log and action dumps to ai-commander

mcbradd:
## Summary
Adds two opt-in env-var dumps to the `ai-commander` sanity runner, for headless audit workflows:
- `PHASE_DUMP_LOG=<path>` — writes the accumulated structured `GameLogEntry` stream (JSON) at game end.
- `PHASE_DUMP_ACTIONS=<path>` — writes each chosen `GameAction` (debug format, one per line).

Zero behavior change when the vars are unset. Motivation: line-by-line rules auditing of headless games (we used this to audit granted-miracle behavior across 9 seeded 4-player games — the existing stdout reports per-turn life totals only, and declined optional offers such as `MiracleReveal` leave no observable trace). Related observation for a possible follow-up: `GameLogEntry` never records that a cast was a miracle cast, so miracle casts are indistinguishable from retail casts in the log; the actions dump is currently the only way to see them.

## Files changed
- crates/phase-ai/src/bin/ai_commander.rs

Method: not-applicable — dev-binary instrumentation, no engine/parser/card logic touched.

## Track
Developer

## LLM
Model: claude-fable-5
Thinking: high
Tier: Frontier

## Verification
- `cargo clippy -p phase-ai --release` — clean, finished in 2m35s
- `cargo build --release -p phase-ai` — builds; binary exercised across 9 seeded 4-player games with both vars set (dumps written, parsed downstream) and without (unchanged behavior).

## Gate A
Not applicable — no parser code touched; pre-commit combinator gate passed on commit.

## Anchored on
- crates/phase-ai/src/bin/ai_commander.rs:41-66 — existing env/arg-driven opt-in configuration in the same binary
- crates/phase-ai/src/duel_suite/attribution.rs:61 — existing precedent for capturing per-action streams for offline analysis

### DIFF
--- a/crates/phase-ai/src/bin/ai_commander.rs
+++ b/crates/phase-ai/src/bin/ai_commander.rs
@@ -171,9 +171,9 @@ fn main() {
     }
 
     let start = Instant::now();
-    let dump_log = std::env::var("PHASE_DUMP_LOG").is_ok();
+    let dump_log_path = read_dump_env("PHASE_DUMP_LOG");
     let mut game_log: Vec<engine::types::log::GameLogEntry> = Vec::new();
-    let dump_actions = std::env::var("PHASE_DUMP_ACTIONS").is_ok();
+    let dump_actions_path = read_dump_env("PHASE_DUMP_ACTIONS");
     let mut actions_log: Vec<String> = Vec::new();
     let mut total_actions: usize = 0;
     let mut last_turn_reported: u32 = 0;
@@ -182,7 +182,7 @@ fn main() {
     let ai_session = phase_ai::session::AiSession::arc_from_game(&state);
 
     loop {
-        let results = run_ai_actions(
+        let mut results = run_ai_actions(
             &mut state,
             &ai_players,
             &ai_configs,
@@ -192,12 +192,12 @@ fn main() {
         if results.is_empty() {
             break;
         }
-        if dump_log {
-            for r in &results {
-                game_log.extend(r.log_entries.iter().cloned());
+        if dump_log_path.is_some() {
+            for r in &mut results {
+                game_log.extend(std::mem::take(&mut r.log_entries));
             }
         }
-        if dump_actions {
+        if dump_actions_path.is_some() {
             for r in &results {
                 actions_log.push(format!("{:?}", r.action));
             }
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**Verdict: changes requested — the opt-in instrumentation is at the right binary seam, but its dump configuration and accumulation path need one local hardening pass before it is safe for large headless audits.**

## 🔴 Blocker

- `crates/phase-ai/src/bin/ai_commander.rs:174-177,195-204,276-283` reads each dump variable twice and treats every `std::env::var` error as if the variable were absent. An invalid-Unicode configured path silently disables capture at startup; a changed/unset value at shutdown can also make a run collect an in-memory dump but write nothing (or a different path). Capture each destination once at startup, distinguish `NotPresent` from invalid configuration, and use the captured paths for the final writes. Evidence: the same values drive both capture and output, but the current branches independently call `std::env::var`.

- `crates/phase-ai/src/bin/ai_commander.rs:195-198` clones every `GameLogEntry` while `results` is dropped after this iteration. Audit runs can reach `MAX_TOTAL_ACTIONS` and accumulate a large trace, so this doubles per-entry allocation/copy work on the very opt-in path intended for large headless captures. Rebind `results` as mutable and extend with `std::mem::take(&mut r.log_entries)`; retain action rendering only when its cached destination is set. Evidence: no later code consumes the individual `results` entries after `total_actions += results.len()`.

The Gemini inline findings at these same lines are confirmed against the current head; there are no other review comments or parser-surface evidence requirements for this binary-only PR.

**Recommendation: make this focused cached-path / move-not-clone fix, then request a fresh review.**

mcbradd:
Pushed a fix for both requested items (new head `c7153934c`):

1. **Cache dump destinations once at startup.** Added `read_dump_env(key: &str) -> Option<String>` that reads `PHASE_DUMP_LOG` / `PHASE_DUMP_ACTIONS` exactly once:
   ```rust
   std::env::var(key)
       .map(Some)
       .or_else(|e| match e {
           std::env::VarError::NotPresent => Ok(None),
           e => Err(e),
       })
       .expect("invalid Unicode in dump-destination env var")
   ```
   `dump_log_path` / `dump_actions_path` are computed once before the loop and drive both the per-iteration capture checks and the shutdown writes — no more double-read, and any non-`NotPresent` error (e.g. invalid Unicode) now panics instead of silently disabling capture.

2. **Stop cloning log entries.** `results` is now `mut`, and the hot path is:
   ```rust
   game_log.extend(std::mem::take(&mut r.log_entries));
   ```
   instead of `r.log_entries.iter().cloned()`. Action rendering is likewise gated on `dump_actions_path.is_some()`.

**Verification** (shared warm target dir, `CARGO_TARGET_DIR=/d/projects/phase/target`):
- `cargo fmt --all` — clean, no diff beyond the intended change.
- `cargo clippy -p phase-ai --release` — clean.
- `cargo build --release -p phase-ai` — builds `ai-commander.exe` successfully.

**Decision-cost perf gate finding:** this failure is unrelated to this diff. The diff only touches a dev test-harness binary (`ai_commander.rs`), which the perf gate never invokes. Looking at the gate's own output for this run:

```
note: card-data hash changed (4bf63d159...→8e688b912...) — likely a card-data-driven
trajectory shift, not a cost-per-node regression; review and refresh if intended
```

8 of 28 counters failed (`mana_aura_trigger_scans`, `restriction_static_mode_gate_scans`, `state_clone_for_legality`, etc.), all engine/mana/legality counters with no connection to this bin. More tellingly, the same "AI gate" workflow's "Decision-cost perf gate" job is currently failing on every other open PR I checked (`feat/life-auction-4131`, `fix/issue-4731`, `feat/text-changing-effects`, `fix/5820-susan-foreman-planar-deck-arrange`, `agent/phase-characteristics-triggers-rules`), starting in the same window (runs before ~07:40 UTC today passed, runs after ~08:27 UTC today fail) — none of those PRs touch `ai_commander.rs` either. This is baseline drift on `main` (most likely from the ongoing card-data/engine changes landing throughout the day), not something introduced here.

Ready for another look.

mcbradd:
Note: attempted to re-request review from @matthewevans via the API but got a 404 (`POST /pulls/6087/requested_reviewers`) — likely a permissions restriction on this fork-based PR. Flagging here instead since the automated re-request didn't go through.

matthewevans (approved):
Approved: the current head caches the opt-in destinations and moves log entries into the dump buffer; the binary-only diff has no remaining review finding.

PR #6059, opened 2026-07-17
Extend SearchFound with bound exile-play permissions

parthmishra:
## Summary
Extends the engine-owned `SearchFound` pipeline with the next two reusable Opposition Agent prerequisites:

- binds searched-library ownership after `CantSearchLibrary` filtering and carries it through standard/scoped choices, serialized batches, and every found-card event;
- binds an exact `ChangeZone`-to-exile plus `GrantCastingPermission` replacement chain, then installs the snapshotted play/`AnyColor` permission only after the card actually reaches exile.

This is intentionally an engine-authority slice, not a full card-support claim. Search-scoped player control and parser/card-data integration remain separate follow-up work.

Context:
- Builds on #5897, which introduced the engine-owned per-card `SearchFound` replacement pipeline.
- Builds on #6037, which established exact `PlayFromExile`/`AnyColor` permission authority and cleanup behavior.
- Continues the narrowed replacement for the earlier broad attempt in #5888.

## Files changed
- `crates/engine/src/types/proposed_event.rs`
- `crates/engine/src/types/game_state.rs`
- `crates/engine/src/game/effects/search_library.rs`
- `crates/engine/src/game/effects/scoped_library_search.rs`
- `crates/engine/src/game/engine_resolution_choices.rs`
- `crates/engine/src/game/replacement.rs`
- `crates/engine/src/game/visibility.rs`
- `crates/engine/src/game/elimination.rs`
- `crates/engine/src/game/effects/mod.rs`
- `crates/engine/src/ai_support/candidates.rs`
- `crates/engine/src/ai_support/mod.rs`
- `crates/phase-ai/src/search.rs`

## CR references
- CR 701.23a — searching and found-card provenance
- CR 614.6 — replacement-selected modified event delivery
- CR 616.1 — frozen replacement ordering and resume authority
- CR 611.2b — lasting play permission created by a resolving effect
- CR 609.4b — exact `AnyColor` payment concession
- CR 601.3 — effect-provided spell-casting permission
- CR 400.7 — exact source incarnation and zone-change identity

## Implementation method (required)
Method: /engine-implementer

## Track
Developer

## LLM
Model: codex-5
Thinking: high
Tier: Standard

## Verification
- [x] Required checks ran clean, or the exact CI-owned alternative is stated below.
- [x] Gate A output below is for the current committed head.
- [x] Final review-impl below is clean for the current committed head.
- [x] Both anchors cite existing analogous code at the same seam.

- `cargo fmt --all` — pass
[...]

PR #6037, opened 2026-07-16
Add exact any-color casting permission authority

parthmishra:
## Summary

Adds the generic, engine-owned `ManaSpendPermission::AnyColor` prerequisite needed for Opposition Agent's "mana as though it were mana of any color" rider.

This change also binds cast-this-way costs and riders to the exact object permission or static source elected for the cast. That identity survives targeting, X selection, distribution, alternate-form casting, during-resolution offers, and payment resumptions, preventing sibling permissions from leaking mana concessions, cost raises, frequency consumption, ETB riders, transformations, or cleanup behavior.

This is the next maintainer-requested narrow slice after merged [#5897](https://github.com/phase-rs/phase/pull/5897), which landed the engine-owned `SearchFound` replacement cycle. It extracts the any-color/provenance prerequisite from the closed broad attempt [#5888](https://github.com/phase-rs/phase/pull/5888). SearchFound permission delivery, search-scoped player control, parser integration, and a full Opposition Agent support claim remain intentionally out of scope.

## Files changed

- `client/src/adapter/types.ts`
- `crates/engine/src/game/casting.rs`
- `crates/engine/src/game/casting_costs.rs`
- `crates/engine/src/game/casting_targets.rs`
- `crates/engine/src/game/casting_tests.rs`
- `crates/engine/src/game/effects/collect_evidence.rs`
- `crates/engine/src/game/engine_casting.rs`
- `crates/engine/src/game/engine_tests.rs`
- `crates/engine/src/game/splice.rs`
- `crates/engine/src/game/visibility.rs`
- `crates/engine/src/types/ability.rs`
- `crates/engine/src/types/game_state.rs`
- `crates/engine/src/types/statics.rs`
- `crates/engine/tests/integration/fireball_x_cost_surcharge_timing.rs`

## CR references

- CR 106.1a-b
- CR 107.4c
- CR 107.4h
- CR 118.9a
- CR 122.1
- CR 205.1b
- CR 601.2a
- CR 601.2f
- CR 603.7
- CR 608.2g
- CR 608.2n
- CR 609.4b
- CR 611.2a
- CR 613.1d
- CR 614.1a
- CR 614.1c
- CR 701.27

## Implementation method (required)

Method: /engine-implementer

## Track

Developer

## LLM

Model: codex-5
Thinking: high

Tier: Standard

## Verification

- [x] Required checks ran clean, or the exact CI-owned alternative is stated below.
- [x] Gate A output below is for the current committed head.
[...]

PR #5897, opened 2026-07-16
Add an engine-owned found-card replacement cycle

parthmishra:
## Summary

Adds a generic, engine-owned replacement cycle for cards found while a player searches their own library. The cycle preserves CR 616 chooser authority, optional and competing replacement ordering, hidden-information boundaries, continuation state, and multiplayer elimination cleanup.

This PR intentionally stops at destination replacement. Opposition Agent player-control and play-permission integration remain deferred to a follow-up after this foundation lands.

## Files changed

- `crates/engine/src/analysis/resource.rs`
- `crates/engine/src/game/effects/scoped_library_search.rs`
- `crates/engine/src/game/elimination.rs`
- `crates/engine/src/game/engine_replacement.rs`
- `crates/engine/src/game/engine_resolution_choices.rs`
- `crates/engine/src/game/replacement.rs`
- `crates/engine/src/game/sba.rs`
- `crates/engine/src/game/visibility.rs`
- `crates/engine/src/types/game_state.rs`
- `crates/engine/src/types/proposed_event.rs`
- `crates/engine/src/types/replacements.rs`

## CR references

- CR 701.23
- CR 614.1
- CR 614.6
- CR 616.1
- CR 800.4a

## Implementation method (required)

Method: /engine-implementer

## Track

Developer

## LLM

Model: codex-5
Thinking: high

Tier: Standard

## Verification

- [x] Required checks ran clean, or the exact CI-owned alternative is stated below.
- [x] Gate A output below is for the current committed head.
- [x] Final review-impl below is clean for the current committed head.
- [x] Both anchors cite existing analogous code at the same seam.

- Full repository pre-push hook — passed: formatting, workspace clippy, release card-data validation, parser gates, engine parser tests (584 passed), phase-ai tests (1281 passed, 8 ignored), Oracle generation, card-data validation, coverage regression check, frontend lint (warnings only), and frontend type-check.
- Targeted found-card replacement, production SearchChoice routing, taxonomy, and elimination tests — passed.
- `cargo test -p engine` — found-card and taxonomy coverage passed after the review fixes; one unrelated local coverage fixture mismatch was caused by a pre-existing uncommitted `crates/engine/data/known-tokens.toml`, which is excluded from this PR. CI runs from a clean checkout.
- Parser/card-data impact — none; this PR claims zero parse impact.

## Gate A

Gate A PASS head=701f1b42b148338208c8800808287bf21acf59f4 base=d62706286601c3e1e0c1360c468e374ffa5e53f1

[...]

PR #6099, opened 2026-07-17
Add search-scoped player control foundation

parthmishra:
## Summary

Adds the engine-owned runtime foundation for controlling a player's decisions while that player searches. The implementation latches search authority and learned-information audiences, preserves exact object incarnations across replacement/delivery pauses, keeps simultaneous multiplayer searches APNAP-correct, and exposes an authoritative filtered submitter projection to server, WASM, and TypeScript consumers.

This intentionally stops before adding the Opposition Agent static ability or parser/card-data wiring; those remain a later, separately reviewable slice.

Method: /engine-implementer

Model: codex-5

Thinking: High

Tier: Standard

## Context

This builds on the previously merged Opposition Agent foundation slices:

- #5897 — engine-owned `SearchFound` replacement cycle
- #6037 — exact any-color casting-permission authority
- #6059 — bound exile-play permissions for found cards

## CR references

- CR 101.4
- CR 400.2 / CR 400.7
- CR 401.2
- CR 616.1
- CR 701.23a / CR 701.23b / CR 701.23e / CR 701.23i
- CR 701.24a
- CR 723.4 / CR 723.5
- CR 800.4a / CR 800.4b

## Verification

- [x] Required checks ran clean, with the exact local-only exception documented below.

- `cargo fmt --all` — clean after reconciliation with current upstream `main`.
- `cargo check -p engine --tests` — passed after reconciliation.
- `cargo clippy -p engine --tests -- -D warnings` — passed after reconciliation.
- Review follow-up — scoped reveal publication now replaces `last_revealed_ids`; focused tests cover the normal APNAP reveal path and two successive scoped reveal batches.
- Focused post-merge suites — 19 SearchFound tests, 7 scoped wire/serde tests, owner-elimination, paused scoped-elimination, latched-audience privacy, standard delivery, and split-delivery regressions passed.
- TypeScript wire tests/type-check/lint — passed before the upstream-only reconciliation; no conflicting client wire changes were introduced by the merge.
- Full engine library before reconciliation — 16,833 passed, 7 ignored; the sole local failure was the token coverage count caused by an unrelated uncommitted `crates/engine/data/known-tokens.toml` change that is excluded from this PR.
- Pre-commit parser/router gates — passed on both commits after the upstream merge.

## Gate A

Gate A PASS head=0cbc3ac475b1fae692f59f4ed0b9e5ec5583c177 base=8406f6532597783e612995a9a6b7c3ce908ab1b5

## Anchored on

- `crates/engine/src/game/zone_pipeline.rs:1031` — existing typed completion deferral across paused zone moves.
- `crates/engine/src/game/visibility.rs:1331` — existing private-access authority used by state and event filtering.

## Final review-impl

Final review-impl PASS head=0cbc3ac475b1fae692f59f4ed0b9e5ec5583c177

## Claimed parse impact

- None — this PR is an engine/runtime authority slice; Opposition Agent parser and card-data wiring remain out of scope.


matthewevans (changes requested):
## Request changes

**[HIGH]** `deliver_collected_cards` appends the completed scoped-search batch to `state.last_revealed_ids` without first replacing the prior batch (`crates/engine/src/game/effects/scoped_library_search.rs:762-775`). That field is the current reveal context: other reveal writers replace it (`dig.rs:149,173`) and the top-level chain prelude clears it (`effects/mod.rs:6345-6348`). Two scoped searches within one resolution chain therefore leave the first search's cards eligible to a later `LastRevealed` consumer after the second search completes.

Reset the reveal context immediately before publishing this scoped batch, then add a regression test that completes two scoped searches in sequence and verifies the second `last_revealed_ids` contains only its own revealed cards.

**Recommendation:** request changes; do not enqueue until the stale-context case is fixed and covered.

matthewevans (approved):
Current-head re-review: the scoped reveal context is now replaced per batch and the two-batch regression covers the prior stale-ID defect. Parse-diff is current with no card changes; existing CR-annotation findings are addressed.

Issue #1301, opened 2026-05-29
Cauldron of Essence debug error:  Cannot pay mana cost — [Debug] dispatch error for ChooseTarget: Engine error: Action…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1508983758580355252/1508983758580355252

**Thread:** Cauldron of Essence debug error:  Cannot pay mana cost
**Cards:** bring back, cauldron of essence
**Parser status:** has_gaps
**Extraction confidence:** 0.90

## Summary
[Debug] dispatch error for ChooseTarget: Engine error: Action not allowed: Cannot pay mana cost
[Debug] dispatch error for ChooseTarget: Engine error:

### bring back
Create two 1/1 white Human creature tokens. (Then exile this card. You may cast the creature later from exile.)
### cauldron of essence
Whenever a creature you control dies, each opponent loses 1 life and you gain 1 life.
{1}{B}{G}, {T}, Sacrifice a creature: Return target creature card from your graveyard to the battlefield. Activate only as a sorcery.

---
<sub>report_id: `discord:1508983758580355252:1508983758580355252:0` · discord: `1508983758580355252/1508983758580355252`</sub>

matthewevans:
## Classifier analysis

**Verdict**: `not_card_data_attributable`

**Matched clause**: `(none)`  (supported: true)

**Reasoning**: The bug is a dispatch error — 'Cannot pay mana cost' — a runtime cost-payment engine error unrelated to how the Oracle text was parsed. The card's AST is intact.

---
_Posted by bug-coverage-classifier skill._

PR #6100, opened 2026-07-17
fix(engine): exclude every non-self cost-paid object from an ability's own target selection

### LINKED ISSUE #4948: Samwise Gamgee — My opponent activated Samwise Gamgee's sacrifice three foods ability, which on the stack said it targe…
Reported in Discord: https://discord.com/channels/1485498006781427802/1522270792572866751/1522270792572866751

**Thread:** Samwise Gamgee
**Cards:** samwise gamgee

## Summary
My opponent activated Samwise Gamgee's sacrifice three foods ability, which on the stack said it targeted "Object 143", and then did nothing when it resolved.

### samwise gamgee
Whenever another nontoken creature you control enters, create a Food token. (It's an artifact with "{2}, {T}, Sacrifice this token: You gain 3 life.")
Sacrifice three Foods: Return target historic card from your graveyard to your hand. (Artifacts, legendaries, and Sagas are historic.)

---
<sub>report_id: `discord:1522270792572866751:1522270792572866751:0` · discord: `1522270792572866751/1522270792572866751`</sub>

tryeverything24:
Fixes #4948.

Samwise Gamgee: "Sacrifice three Foods: Return target historic card from your graveyard to your hand."

## Root cause & fix

This engine pays a non-self sacrifice/discard/exile activation cost before choosing the ability's target (a documented CR 601.2c-vs-602.2b ordering shortcut, see issue #1301's `exclude_cost_paid_object_that_left_battlefield`). A sacrificed Food is itself Historic (it's an artifact) and lands in the graveyard the instant the cost is paid, so it transiently qualified as a legal "target historic card from your graveyard" candidate for the very same activation. CR 704.5d then makes the token cease to exist before the ability resolves, so if it was chosen the ability silently fizzled: "targeted Object N, then did nothing when it resolved" — the exact reported symptom.

`exclude_cost_paid_object_that_left_battlefield` already existed for exactly this problem class, but only tracked a SINGLE cost-paid object (stamped from `chosen.first()`). Samwise sacrifices THREE Foods, so only one of the three was ever excluded. Generalized the tracked field to a `Vec<ObjectId>` (`cost_paid_object_ids`), populated alongside the existing singular stamp at all three non-self cost-payment sites (`handle_sacrifice_for_cost`, `handle_discard_for_cost`, `finish_exile_selection_for_cost`), so the shared exclusion filter now excludes every object the cost consumed, not just the first.

Also closes a sibling defect: a multi-object non-self DISCARD cost had the same single-object leak, with a worse consequence than Samwise's silent fizzle — since a discarded non-token card does NOT cease to exist (CR 704.5d is token-specific), the leaked second discarded card was a legal, resolvable target, letting a "Discard two creature cards: Return target creature card from your graveyard to your hand"-shaped ability return one of its own just-discarded cards for free.

## Testing

Two new end-to-end tests driving the real cast/activate → pay cost → choose target → resolve pipeline. Each scenario includes two legitimate pre-existing graveyard targets so target selection is genuinely ambiguous and the engine pauses on an interactive `TargetSelection` prompt (a lone legal target auto-resolves inline without pausing, which would defeat the point of asserting on `legal_targets`).

Verified locally: fmt clean, clippy clean (`-D warnings`), 3255 integration tests pass (3253 pre-existing + 2 new), 0 regressions.


matthewevans (approved):
Reviewed current head `ecd22a83f7995e5b044dea2ffc361368c8632fed` after the CR-annotation correction. The cost-payment tracking is at the authoritative activation/casting path, preserves the existing tracked-payment semantics, and the regression coverage exercises the production path. Required checks are green.

Issue #2384, opened 2026-06-05
Molten-Core Maestro: mana produced equals pre-counter power (counter applied after mana generation)


matthewevans:
## Reported behavior
Molten-Core Maestro was a 4/4. Casting a 5-mana spell: Maestro should get a +1/+1 counter (becoming 5/5) then produce 5 {R}. Instead it produced only 4 {R}.

## Oracle text (verified from card-data.json)
```
Menace
Opus — Whenever you cast an instant or sorcery spell, put a +1/+1 counter on this creature. If five or more mana was spent to cast that spell, add an amount of {R} equal to this creature's power.
```

## Root cause analysis
AST encodes PutCounter first, then Mana production (Power: Source scope) as SequentialSibling. However the engine resolves the mana amount from the pre-counter power snapshot rather than current power after counter application.

## Expected
4/4 Maestro + 5-mana spell → gets +1/+1 counter (5/5) → produces 5 {R}

## Actual
4/4 Maestro + 5-mana spell → produces 4 {R} (pre-counter power)

## Source
https://discord.com/channels/1485498006781427802/1512135055722545193/1512135055722545193

---
<sub>report_id: `discord:1512135055722545193:1512135055722545193:0` · via bug-triage</sub>

PR #6103, opened 2026-07-17
fix(engine): recognize the symmetric two-way control-swap idiom, and keep it simultaneous

### LINKED ISSUE #4731: Reins of Power targeting incorrectly — [[Reins of power]] is giving the casting player control of all creatures on the…
Reported in Discord: https://discord.com/channels/1485498006781427802/1520487615122636931/1520487615122636931

**Thread:** Reins of Power targeting incorrectly
**Cards:** reins of power

## Summary
[[Reins of power]] is giving the casting player control of all creatures on the board, instead of the intended swap.

### reins of power
Untap all creatures you control and all creatures target opponent controls. You and that opponent each gain control of all creatures the other controls until end of turn. Those creatures gain haste until end of turn.

---
<sub>report_id: `discord:1520487615122636931:1520487615122636931:0` · discord: `1520487615122636931/1520487615122636931`</sub>

tryeverything24:
Fixes #4731.

Reins of Power: "Untap all creatures you control and all creatures target opponent controls. You and that opponent each gain control of all creatures the other controls until end of turn. Those creatures gain haste until end of turn."

## Root cause & fix

"Each gain control of all creatures the other controls" collapsed into a single `Effect::GainControlAll` whose population filter had NO controller anaphor (matching every creature regardless of owner) and whose resolver always assigns the new controller to `ability.controller` — the caster. There was no second, opposite-direction control change for the opponent's side, so the caster ended up controlling every creature on the board instead of a clean two-for-two swap.

A dedicated parser combinator (`try_parse_symmetric_gain_control_all`) now recognizes this idiom and composes two independently correct mechanisms via `sub_ability` chaining:
  - `GainControlAll { target: creatures the target opponent controls }` — the caster takes (same mechanism as the already-correct Hellkite Tyrant "gain control of all artifacts that player controls").
  - `GiveControlAll { target: creatures the caster controls, recipient: the target opponent }` — new mass counterpart of `GiveControl`; the opponent takes.

A second defect surfaced during testing: `resolve_chain_body` deliberately flushes pending layers before a sub-ability resolves (issue #2384, so a P/T-dependent sub-effect sees current characteristics). That flush applies `GainControlAll`'s control-change TCEs before `GiveControlAll` runs, so `GiveControlAll`'s live "creatures you control" filter incorrectly also matched the creatures `GainControlAll` just took — collapsing the swap right back into a one-way grab. Fixed via a new resolution-scoped `GameState::last_gained_control_object_ids` field: `resolve_all` stamps the objects it just took, and `resolve_give_all` excludes them from its own filter, recovering the intended CR 613.9 simultaneity.

Out of scope: "Those creatures gain haste until end of turn" (the oracle's third sentence) resolves to an `AddKeyword` TCE targeting the two PLAYERS, not the four creatures that changed control — a separate, pre-existing "those creatures" scoping bug unrelated to the reported one-way-grab issue. Left for a follow-up rather than bundled here.

## Testing

New end-to-end test driving the real cast → target-selection → resolution pipeline with each player controlling creatures the other doesn't, so a one-way grab and a no-op swap are both distinguishable from the correct two-for-two outcome.

Verified locally: fmt clean, clippy clean on both `engine` and `phase-ai` (`-D warnings`), 3252 integration tests pass, 0 regressions.

## Summary by CodeRabbit

* **New Features**
  * Added support for mass control handoffs and associated event/profiling coverage.
  * Added parsing support for Reins of Power–style two-way “all creatures” control swaps.
* **Bug Fixes**
  * Fixed simultaneous/mass-resolution behavior to prevent one-sided or duplicate control outcomes.
  * Improved oracle/AST duration propagation and printed-signature handling for the new effect.
  * Updated analysis and AI policies to recognize GiveControlAll consistently.
* **Tests**
  * Added regression tests covering two-way control swap parsing and end-to-end resolution.

### DIFF
--- a/crates/engine/src/analysis/ability_graph.rs
+++ b/crates/engine/src/analysis/ability_graph.rs
@@ -858,6 +858,7 @@ fn effect_projection(effect: &Effect) -> Projection {
         | Effect::Dig { .. }
         | Effect::GainControl { .. }
         | Effect::GainControlAll { .. }
+        | Effect::GiveControlAll { .. }
         | Effect::ControlNextTurn { .. }
         | Effect::Attach { .. }
         | Effect::UnattachAll { .. }
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -2796,6 +2796,7 @@ fn legacy_effect(x: &Effect) -> bool {
         | Effect::PumpAll { target, .. }
         | Effect::GainControl { target }
         | Effect::GainControlAll { target }
+        | Effect::GiveControlAll { target, .. }
         | Effect::ControlNextTurn { target, .. }
         | Effect::Bounce { target, .. }
         | Effect::DestroyAll { target, .. }
@@ -4993,6 +4994,10 @@ fn rw_effect(
         Effect::GainControl { target: _ }
         | Effect::GainControlAll { target: _ }
         | Effect::GiveControl {
+            target: _,
+            recipient: _,
+        }
+        | Effect::GiveControlAll {
             target: _,
             recipient: _,
         }
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -724,6 +724,12 @@ fn scan_effect(x: &Effect, mode: ScanMode) -> Axes {
             acc = acc.or(scan_target_filter(target, target_ctx, mode));
             acc
         }
+        Effect::GiveControlAll { target, recipient } => {
+            let mut acc = Axes::NONE;
+            acc = acc.or(scan_target_filter(target));
+            acc = acc.or(scan_target_filter(recipient));
+            acc
+        }
         Effect::ControlNextTurn {
             target,
             grant_extra_turn_after: _,
@@ -5880,6 +5886,7 @@ fn effect_resolution_choice_freedom(e: &Effect) -> ResolutionChoiceFreedom {
         | Effect::Dig { .. }
         | Effect::GainControl { .. }
         | Effect::GainControlAll { .. }
+        | Effect::GiveControlAll { .. }
         | Effect::ControlNextTurn { .. }
         | Effect::Attach { .. }
         | Effect::UnattachAll { .. }
@@ -6151,6 +6158,7 @@ pub(crate) fn effect_is_randomness_bearing(e: &Effect) -> bool {
         | Effect::Dig { .. }
         | Effect::GainControl { .. }
         | Effect::GainControlAll { .. }
+        | Effect::GiveControlAll { .. }
         | Effect::ControlNextTurn { .. }
         | Effect::Attach { .. }
         | Effect::UnattachAll { .. }
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -726,8 +726,8 @@ fn scan_effect(x: &Effect, mode: ScanMode) -> Axes {
         }
         Effect::GiveControlAll { target, recipient } => {
             let mut acc = Axes::NONE;
-            acc = acc.or(scan_target_filter(target));
-            acc = acc.or(scan_target_filter(recipient));
+            acc = acc.or(scan_target_filter(target, target_ctx, mode));
+            acc = acc.or(scan_target_filter(recipient, target_ctx, mode));
             acc
         }
         Effect::ControlNextTurn {
@@ -5136,6 +5136,10 @@ fn effect_target_ctx(e: &Effect, mode: ScanMode) -> FilterReadContext {
         | Effect::DamageEachPlayer { .. }
         | Effect::DestroyAll { .. }
         | Effect::GainControlAll { .. }
+        // CR 732.2a: `GiveControlAll` is `GainControlAll`'s recipient-inverted
+        // sibling (issue #4731's two-way swap) — the same mass-population
+        // battlefield read, so the same fail-closed census classification.
+        | Effect::GiveControlAll { .. }
         | Effect::PumpAll { .. }
         | Effect::BounceAll { .. }
         | Effect::UnattachAll { .. }
@@ -5477,7 +5481,7 @@ enum RelaxReason {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum CensusRole {
     /// Mass battlefield population read - enumerated over every matching object,
-    /// scales with the growing class => fail-CLOSED census. EXACTLY the 28
+    /// scales with the growing class => fail-CLOSED census. EXACTLY the 29
     /// `effect_target_ctx` `LiveBoardCensus` members.
     Census,
     Relax(RelaxReason),
@@ -5491,7 +5495,7 @@ enum CensusRole {
 #[cfg(test)]
 fn effect_census_role(e: &Effect) -> CensusRole {
     match e {
-        // -- CENSUS (28): verbatim mirror of `effect_target_ctx`'s LiveBoardCensus
+        // -- CENSUS (29): verbatim mirror of `effect_target_ctx`'s LiveBoardCensus
         // arm - mass battlefield population reads that scale with growth.
         Effect::EachSourceDealsDamage { .. }
         | Effect::EachDealsDamageEqualToPower { .. }
@@ -5500,6 +5504,9 @@ fn effect_census_role(e: &Effect) -> CensusRole {
         | Effect::DamageEachPlayer { .. }
         | Effect::DestroyAll { .. }
         | Effect::GainControlAll { .. }
+        // CR 732.2a: `GiveControlAll` mirrors its `effect_target_ctx` census
+        // entry (issue #4731's two-way swap sibling of `GainControlAll`).
+        | Effect::GiveControlAll { .. }
         | Effect::PumpAll { .. }
         | Effect::BounceAll { .. }
         | Effect::UnattachAll { .. }
@@ -6928,6 +6935,10 @@ mod tests {
             "EachSourceDealsDamage",
             "ExploreAll",
             "GainControlAll",
+            // CR 732.2a: GiveControlAll — GainControlAll's recipient-inverted
+            // sibling (issue #4731's two-way swap), the same mass-population
+            // battlefield read.
+            "GiveControlAll",
             "GoadAll",
             "PumpAll",
             "PutCounterAll",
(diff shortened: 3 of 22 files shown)


matthewevans (changes requested):
**Request changes — the Reins runtime regression is covered, but the new parser path is a one-card full-text matcher rather than a composable Oracle grammar.**

## 🔴 Blocker

**[HIGH]** The new parser dispatch hard-codes complete Reins of Power clauses. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:16636-16644` defines two whole-sentence strings and accepts them via `SUBJECT_PREFIXES.contains(&lower.as_str())`. Why it matters: this violates the parser’s required nom-combinator and class-based design, so it only supports the two enumerated phrasings instead of the grammatical two-way mass-control-swap class. Suggested fix: express the subject, recipient-anaphor, population, and controller-anaphor axes with the existing nom/target-filter building blocks, then add sibling-form tests that exercise those axes rather than exact text only.

**[MED]** The companion-slot state introduced for this path is carried as raw booleans. Evidence: `crates/engine/src/game/ability_utils.rs:368`, `1528`, `2386`, `2411`, `5745`, `5923`, and `5933`; Gemini’s current-head threads identify the same state. Why it matters: `false`/`true` does not name whether the companion slot is pending or consumed, leaving the chain-state protocol less maintainable and contrary to the repository’s typed-state convention. Suggested fix: replace the field, local, and mutable parameters with a small `Pending`/`Pushed` status enum, as the existing review recommends.

## 🟡 Non-blocking

The required current-head parse-diff artifact is not yet available: `inspect` reports `parse_diff.state=absent` while Card data is pending. It must be checked against the claimed scope before approval; this is CI evidence, not a request to change behavior.

## ✅ Clean

The new integration test drives cast, target selection, resolution, and layer evaluation with two distinct creatures per player, so its four controller assertions distinguish the reported one-way grab from a two-way swap.

**Recommendation: request changes for a composable parser and typed companion-slot state; then regenerate and review the current-head parse-diff before approval.**

matthewevans (changes requested):
**Request changes — current head still has the two earlier architecture blockers, and it now marks Reins of Power as supported while deliberately leaving a rules-bearing rider incorrect.**

## 🔴 Blockers

1. **[HIGH] Coverage honesty: Reins is no longer safely unsupported, but its haste rider is knowingly wrong.**  The new integration test says the third sentence is not asserted because it becomes an `AddKeyword` targeting the two players rather than the changed creatures (`crates/engine/tests/integration/issue_4731_reins_of_power.rs:160-167`). The current parse-diff artifact also reports the Reins haste grant removed. A card must stay explicitly unimplemented for the unsupported remainder, or the chain must carry a typed, frozen affected-object set through the haste grant and test all four creatures receive it. Please do not silently accept the full card with this known incorrect result.

2. **[HIGH] The parser is still an exact full-text allowlist, not a composable grammar.**  `try_parse_symmetric_gain_control_all` accepts only two whole strings through `SUBJECT_PREFIXES` (`crates/engine/src/parser/oracle_effect/mod.rs:17157-17166`). That reintroduces the prohibited one-card phrase matching pattern and cannot generalize independently varying subject, population, and recipient/controller axes. Please parse the idiom with the project nom combinators and typed axes, retaining `Unimplemented` for forms the grammar cannot lower correctly.

3. **[MED] Chain-wide target reuse is represented by raw booleans.**  `SlotAccumulator::target_player_companion_slot_pushed: bool` and `assign_targets_in_chain`'s `player_slot_consumed: bool` (`crates/engine/src/game/ability_utils.rs:372-387, 1585-1593`) encode a cross-link target-selection protocol with two independent flags. Use a small typed state that makes the declaration/reuse transition explicit; it is safer and extensible for distinct player/opponent references.

The main two-way control assertion is useful, but it cannot validate the omitted third sentence. I did not run local builds; GitHub still shows the Rust/AI checks in progress on this head.

matthewevans (changes requested):
**Request changes — this rebased current head has the same implementation patch as the previously blocked head, so all three substantive blockers remain.**

## 🔴 Blockers

1. **[HIGH] Coverage honesty remains broken for Reins of Power.** The test explicitly leaves its haste rider incorrect: the grant targets players rather than the changed creatures (`crates/engine/tests/integration/issue_4731_reins_of_power.rs:160-167`). The parse-diff artifact also reports the Reins haste grant removed. Keep the card explicitly unsupported for that remainder, or carry a typed frozen affected-object set through the grant and test all affected creatures.

2. **[HIGH] The parser remains a two-string allowlist.** `try_parse_symmetric_gain_control_all` matches only complete clauses through `SUBJECT_PREFIXES` (`crates/engine/src/parser/oracle_effect/mod.rs:17157-17166`). Replace this with composed nom grammar axes and retain an explicit unsupported result for forms that cannot be lowered correctly.

3. **[MED] Target reuse remains represented by raw booleans.** `target_player_companion_slot_pushed` and `player_slot_consumed` encode cross-chain target-selection state as booleans (`crates/engine/src/game/ability_utils.rs:372-387, 1585-1593`). Use a typed declaration/reuse state instead.

The new commit is an exact patch-id rebase of the prior reviewed change; required CI is green. No local build was run.

**Recommendation: keep this PR in requested changes until the parser, typed target-state, and complete card semantics are addressed.**

tryeverything24:
Rebased onto current `main` to clear the merge conflict, plus the adaptations the moved upstream required: the `scan_target_filter` signature change threaded through the `GiveControlAll` arm; `GiveControlAll` classified in the new exhaustive census guards (`effect_target_ctx`/`effect_census_role`, the recipient-inverted sibling of `GainControlAll`'s existing census entry — enumerations and counts updated 28→29); and a CR733 authority-matrix row added for this branch's `last_gained_control_object_ids` resolution-scoped ledger (mirroring `last_zone_changed_ids`), re-inserted in the regenerated fixture's format.

To be transparent about scope: this push does NOT yet address the three standing review blockers (composable nom grammar for the two-way swap idiom, typed companion-slot state, and the haste-rider semantics). Those need a substantive rework that I did not want to rush into a conflict-clearing push; it is next on my queue for this PR. If you would rather I close and resubmit once the full rework is ready, happy to do that instead.

Full local gate on the rebased head in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17549 passed, engine integration 3816 passed.

matthewevans (changes requested):
**Request changes — the rebased head still has the original coverage/parser/target-state defects, and its new control-swap plumbing adds two confirmed incomplete consumers.**

## 🔴 Blocker

**[HIGH]** Reins of Power is treated as supported while its haste rider is knowingly mis-scoped. Evidence: `crates/engine/tests/integration/issue_4731_reins_of_power.rs:160-167` says “Those creatures gain haste” targets players rather than the four affected creatures; the current parse-diff sticky artifact is from 2026-07-20, before head `29ab15f`, so it cannot prove current coverage. Why it matters: the card resolves with a rules-bearing sentence wrong while coverage no longer preserves an explicit unsupported remainder. Suggested fix: keep the card honest with `Effect::unimplemented` until the rider is represented with a typed frozen affected-object set, or implement and runtime-test the full rider.

**[HIGH]** The parser is still a one-card, full-text allowlist. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:17249-17257` compares the entire normalized clause with two entries in `SUBJECT_PREFIXES`. Why it matters: the parser only accepts Reins wording and violates the required composable nom grammar/class implementation. Suggested fix: parse the subject, population, recipient, and controller-anaphor axes with the existing nom/filter building blocks; retain an explicit unsupported result for forms the grammar cannot lower.

**[HIGH]** The legacy profile visitor drops `GiveControlAll.recipient`. Evidence: `crates/engine/src/game/ability_rw.rs:2797-2800` matches `Effect::GiveControlAll { target, .. }`, whereas the new recipient is also a `TargetFilter`. Why it matters: a legacy/event-context reference in the recipient side is invisible to the conservative profile walk. Suggested fix: traverse both `target` and `recipient`, as `GiveControl` does.

## 🟡 Non-blocking

**[MED]** Chain-wide companion-target state is still encoded as booleans. Evidence: `crates/engine/src/game/ability_utils.rs:372-387`, `1585-1593`, and `2450-2479` use `target_player_companion_slot_pushed: bool` / `player_slot_consumed: bool`. Why it matters: this cross-link protocol has named transitions (pending vs consumed) that raw booleans obscure. Suggested fix: use a small typed state enum consistently through collection and assignment.

**[MED]** The AI policy scores a paired swap as an unpaired giveaway. Evidence: `crates/phase-ai/src/policies/control_change_awareness.rs:103-126` sets `gives_away_permanent` for `GiveControlAll` even when the same ability contains the complementary `GainControlAll`. Why it matters: Reins-like abilities receive the critical giveaway penalty despite being an exchange-shaped effect. Suggested fix: recognize the complementary pair as one compound swap and retain the penalty for genuinely unpaired giveaways; add a policy regression.

**[LOW]** The `CR 613.9` annotations do not substantiate the claimed simultaneity. Evidence: `crates/engine/src/game/effects/gain_control.rs:135-140,296-307`; verified `docs/MagicCompRules.txt:3044` defines only when one continuous effect overrides another. Why it matters: the cited rule does not explain the snapshot/exclusion mechanism or its timing. Suggested fix: cite the rule that governs the implemented layer/timestamp or dependency behavior, after verifying it against the local CR text.

## ✅ Clean

The four-controller integration assertions remain a discriminating witness for the original one-way-grab regression.

**Recommendation: keep requested changes until the full card stays coverage-honest, the parser is compositional, the missing recipient traversal and AI pairing are fixed, and the parse-diff artifact is current-head evidence.**

matthewevans (changes requested):
**Request changes — the current-head parse artifact confirms an unexplained second-card blast radius and a removed Reins rider.**

## 🔴 Blocker

**[HIGH]** The current `<!-- coverage-parse-diff -->` artifact (updated 2026-07-22 17:48 UTC for head `29ab15f`) changes both Reins of Power and Twist Allegiance: it adds `GiveControlAll` and changes `GainControlAll` for both cards, removes Twist Allegiance’s tracked-set untap, and removes Reins’s haste grant. Why it matters: the PR’s Reins-specific claim does not explain the Twist regressions, while the removed haste signature confirms the known unsupported rider is being silently dropped. Suggested fix: narrow/rework the grammar so unrelated Twist behavior is preserved, and keep Reins explicitly unsupported until its haste rider is represented correctly.

**Recommendation: keep requested changes; the measured card-level diff does not reconcile with the claimed scope.**

matthewevans:
This pull request still has unresolved requested changes, and there has been no contributor follow-up on the current head.

Please push a commit that addresses the existing requested changes and request re-review within 7 days of this warning. If the requested changes remain unaddressed after that window, this pull request will be automatically closed.

matthewevans:
**Closed — requested changes expired without a contributor follow-up.**

The current head 29ab15f188027d05544f5a5d36a36ec411f354b5 still has the unresolved requested-changes reviews. The 2026-07-29 expiry warning (https://github.com/phase-rs/phase/pull/6103#issuecomment-5121677596) gave seven days to push a correction; GitHub shows no author comment, review, or new head after that warning. Closing is solely the requested-changes expiry action, not a judgment about model tier or the author account. A refreshed PR from current main is welcome when the documented blockers are addressed.

PR #6109, opened 2026-07-17
fix(parser): bind "those creatures" after "draw for each creature" to the count filter (#6065)

### LINKED ISSUE #6065: Inspiring Call does not grant indestructible to creatures with +1/+1 counters
## Bug

`Inspiring Call` draws the correct number of cards for creatures you control with +1/+1 counters, but those creatures do not gain indestructible until end of turn.

## Oracle text

> Draw a card for each creature you control with a +1/+1 counter on it. Those creatures gain indestructible until end of turn.

The 2017-11-17 ruling also confirms the affected set is locked in as the creatures that had +1/+1 counters when Inspiring Call resolved: creatures that get counters later do not gain indestructible, and creatures that did gain indestructible keep it even if they later lose their counters.

## Expected

When `Inspiring Call` resolves:

1. Count creatures you control with one or more +1/+1 counters.
2. Draw that many cards.
3. Grant indestructible until end of turn to that same set of creatures.

## Actual

The draw count works, but the matching creatures do not gain indestructible.

## Classifier analysis

**Matched clause**: `Draw a card for each creature you control with a +1/+1 counter on it. Those creatures gain indestructible until end of turn.` (supported: true)

**Reasoning**: Coverage marks the whole spell as supported and includes a supported `grant Indestructible` child, but the child is encoded as affecting `parent target`. The parent `Draw` effect targets the controller and only uses the countered creatures as an `ObjectCount` filter, so the "those creatures" object set is not represented as the target/result set that the keyword grant can consume.

Generated `client/public/card-data.json` currently shows:

```json
{
  "type": "Draw",
  "count": {
    "type": "Ref",
    "qty": {
      "type": "ObjectCount",
      "filter": {
        "type": "Typed",
        "type_filters": ["Creature"],
        "controller": "You",
        "properties": [
          {
            "type": "Counters",
            "counters": { "type": "OfType", "data": "P1P1" },
            "comparator": "GE",
            "count": { "type": "Fixed", "value": 1 }
          }
        ]
      }
    }
  },
  "target": { "type": "Controller" }
}
```

with a sequential child:

```json
{
  "type": "GenericEffect",
  "static_abilities": [
    {
      "mode": "Continuous",
      "affected": { "type": "ParentTarget" },
      "modifications": [
        { "type": "AddKeyword", "keyword": "Indestructible" }
      ],
      "description": "gain indestructible"
    }
  ],
  "duration": "UntilEndOfTurn"
}
```

This likely needs a general building block for "those [objects]" referring to the object set counted by a previous `for each <filter>` clause, rather than treating the counted set as only a numeric quantity.

jaytbarimbao-collab:
Fixes #6065

## Problem

Inspiring Call — *"Draw a card for each creature you control with a +1/+1 counter on it. **Those creatures** gain indestructible until end of turn."* — drew the right number of cards but never granted indestructible.

"Those creatures" lowers to `TargetFilter::ParentTarget`. The chain assembler only rebinds that anaphor when a prior clause is a **tracked-set publisher** (`publishes_tracked_set_from_resolution` — exile/token/`PutCounterAll`/…). `Effect::Draw` is not a publisher (it publishes nothing at resolution; its `target` is the drawing player), so the rebind was skipped and `ParentTarget` survived — resolving at runtime to the Draw's target, the **controller**. The grant landed on a player, so no creatures gained indestructible.

## Fix

`Draw a card for each <filter>` carries the counted creatures in its `count: ObjectCount { filter }`, not in a tracked set. So in the cross-clause pronoun block (`assembly.rs`), when there is **no** tracked-set publisher but the nearest prior clause is such a `Draw`, bind the following "those creatures" grant's `ParentTarget` directly to that count `filter`.

Scoping: the new branch is an `else` to the existing `any_prior_publishes` path, so the `PutCounterAll → TrackedSet(0)` behavior (Nalia de'Arnise) is untouched. It only fires for a `Draw{count: ObjectCount{filter}}` antecedent + an explicit "those creatures" pronoun. Only the grant surface (`GenericEffect`) is rewritten. This is a 1-card class today (Inspiring Call is the only "draw for each creature … those creatures gain" card).

The resulting continuous keyword grant to a Typed creature filter is the standard anthem-style path (Archangel Avacyn etc.) — no runtime change needed.

## Tests

- **Parser** (`inspiring_call_those_creatures_bind_to_draw_count_filter`): the grant binds to the counted Typed creature filter, not `ParentTarget`/`TrackedSet`. The existing Nalia test (`PutCounterAll → TrackedSet(0)`) is the regression guard and stays green.
- **Runtime** (`inspiring_call_indestructible_grant.rs`): casts Inspiring Call (live-parsed) with one creature carrying a +1/+1 counter and one without — the countered creature gains indestructible, the other does not (the grant resolves to the counted set, not to all creatures or the controller).

CR: 603.7.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/assembly.rs
+++ b/crates/engine/src/parser/oracle_effect/assembly.rs
@@ -60,13 +60,14 @@ use super::{
     bind_anaphoric_damage_subject_keep_recipient, collapse_ephemeral_color_choice_mana,
     contains_explicit_tracked_set_pronoun, contains_implicit_tracked_set_pronoun,
     def_is_damage_dealer, def_is_dig_look, def_is_dig_or_mill, def_is_generic_effect_head,
-    def_is_keyword_counter_placement, demote_unbindable_batch_aggregate,
+    def_is_keyword_counter_placement, demote_unbindable_batch_aggregate, draw_object_count_filter,
     fold_cast_copy_of_card_defs, has_explicit_player_target, inject_chosen_color_choice_grant,
     mark_uses_tracked_set, parse_spell_graveyard_replacement_rider,
     publishes_aggregate_set_from_resolution, publishes_tracked_set_from_resolution,
     rebind_tracked_aggregate_to_chain_set, retarget_counter_additional_cost_to_target,
-    rewrite_parent_targets_to_tracked_set, rewrite_rounding_mode, rewrite_that_type_mana_instead,
-    stamp_delayed_returns, try_fold_token_repeat_into_count, wire_optional_cast_decline_fallback,
+    rewrite_grant_parent_to_filter, rewrite_parent_targets_to_tracked_set, rewrite_rounding_mode,
+    rewrite_that_type_mana_instead, stamp_delayed_returns, try_fold_token_repeat_into_count,
+    wire_optional_cast_decline_fallback,
 };
 
 // ===========================================================================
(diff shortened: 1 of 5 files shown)


matthewevans (approved):
Approved — current-head review confirms the intended one-card Inspiring Call parse-diff, the count-filter rebind at the existing assembly seam, and a discriminating cast/resolve regression for countered versus plain creatures.

PR #6125, opened 2026-07-18
Add There Are No Permanents (state trigger)

e11734937-beep:
## Summary

Route the game-state form "when there are no <type> on the battlefield, <effect>" to a real state trigger.

`try_parse_player_trigger` recognizes a bare game-state-condition trigger by lowering the clause through the shared condition authority (`parse_inner_condition`) and bridging the result via `static_condition_to_trigger_condition`. The accept-gate previously admitted only a `TriggerCondition::ControlsType` result, so the global-emptiness form — which `parse_inner_condition` already lowers to `ObjectCount(<filter>) == 0` with no controller restriction (it counts *any* player's matching permanents; CR 110.1 / CR 403.1) and which bridges 1:1 to `TriggerCondition::QuantityComparison` — was rejected at the gate. The line then degraded to Unimplemented and never fired.

This widens the accept-gate in `try_parse_player_trigger` to also accept `TriggerCondition::QuantityComparison` alongside the existing `ControlsType`. It routes to the pre-existing `parse_inner_condition` -> `QuantityComparison` lowering and `check_trigger_condition` evaluation, emitting `TriggerMode::StateCondition` exactly as the sibling arm does. No new trigger mode, mechanism, effect variant, resolver path, or hot path (CR 603.8).

## Anchored on >=2 existing siblings

- The existing `ControlsType` sibling arm in the same accept-gate in `try_parse_player_trigger` (identical shape: lower via `parse_inner_condition`, bridge via `static_condition_to_trigger_condition`, emit `TriggerMode::StateCondition`).
- The existing `static_condition_to_trigger_condition` mapping that already produces `TriggerCondition::QuantityComparison`, already consumed by the `check_trigger_condition` evaluation.

## Model / Method

- Model: claude-opus-4-8
- Thinking: high
- Method: /engine-implementer
- Track: Developer

## Parse impact

- Claimed parse impact: +10 supported, 0 regressions.
- Newly supported: Porphyry Nodes, Drop of Honey, Task Mage Assembly, Mana Vortex, Endrek Sahr Master Breeder, Last Laugh, Lurking Jackals, Opal Avenger, Transcendence, Veiled Crocodile.

## Validation

- Gate A PASS head=e391b0adefa0b3170e3ee8c5a71a603af22283f0 base=af578cb59f8e53243da4d21330b711af59eed6cb
- Final review-impl PASS head=e391b0adefa0b3170e3ee8c5a71a603af22283f0
- `cargo test -p engine --test integration -- there_are_no_permanents`: 4 passed, 0 failed
- Validation/CI Failures: None

### DIFF
--- a/crates/engine/src/parser/oracle_trigger.rs
+++ b/crates/engine/src/parser/oracle_trigger.rs
@@ -13347,15 +13347,21 @@ fn try_parse_player_trigger(lower: &str) -> Option<(TriggerMode, TriggerDefiniti
     // `static_condition_to_trigger_condition`, so every presence filter the
     // condition parser already handles (subtype, type, P/T comparator, keyword,
     // …) is covered without re-implementing filter parsing here. Gated on a
-    // `ControlsType` result so only genuine single-permanent presence conditions
-    // become state triggers; the effect ("sacrifice this creature") is parsed
-    // separately by the caller, exactly as for the `ControlsNone` arm.
+    // `ControlsType` (single-permanent presence) or `QuantityComparison`
+    // (count / emptiness — e.g. "there are no creatures on the battlefield",
+    // which `parse_inner_condition` lowers to `ObjectCount(<filter>) == 0` with
+    // no controller restriction, so it counts *any* player's matching
+    // permanents; CR 110.1 / CR 403.1) result, so only genuine game-state
+    // conditions (CR 603.8) become state triggers; the effect ("sacrifice this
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Maintainer review passed: the shared QuantityComparison state-trigger bridge is correct, the global opponent-scope discriminator is covered, and the existing CR 603.8 annotation resolves the stale bot concern.

PR #6136, opened 2026-07-18
feat(parser+engine): Comeuppance — typed-source prevention riders and compound recipients

mcbradd:
Model: claude-fable-5
Tier: Frontier
Thinking: high

## Summary

Implements **Comeuppance** — *"Prevent all damage that would be dealt to you and planeswalkers you control this turn by sources you don't control. If damage from a creature source is prevented this way, Comeuppance deals that much damage to that creature. If damage from a noncreature source is prevented this way, Comeuppance deals that much damage to the source's controller."*

- **Compound prevention recipient** — `TargetFilter::ControllerAndControlledPermanents { permanent_type: Option<CoreType> }` lowers "you and planeswalkers you control" through `untargeted_damage_filter` into the parameterized `DamageTargetFilter::PlayerOrPermanentsControlledBy` (existing variant, new `permanent_type` axis), with `typed_recipient_valid_card_filter` foreclosing the player-leg drop. Also fixes **Channel Harm's latent dropped source filter** (same seam, pre-existing).
- **Typed source axis on prevention** — "by sources you don't control" parses through the shared `parse_damage_source_subject` grammar (`ControllerRef::Opponent`), the same combinator family the existing "you control" tag uses.
- **Per-event rider provenance** — the prevented event's source is captured once (`proposed_damage_source`) before the applier consumes the event, stashed on the resident post-replacement drain, and consumed by both the source-type gate (`AbilityCondition::PostReplacementDamageSourceMatchesFilter`) and the reflection target (`TargetFilter::PostReplacementDamageSource`) from that single authority — the object-valued siblings of the existing `PostReplacementSourceController` (Swans of Bryn Argoll) mechanism. Batched combat is handled by excluding per-source-reflecting shields from `batched_combat_all_shield`, so each combat damage event carries its own source and amount and is drained inline before the next (CR 615 prevention; CR 616.1 ordering unchanged — the shared pipeline is untouched).
- **Independent-gate chain fall-through** — mutually exclusive riders ("if creature source… / if noncreature source…") each evaluate their own gate at resolution rather than being skipped by parent-effect bookkeeping.

## Anchored on

- `crates/engine/src/types/ability.rs:4861` — `PostReplacementSourceController`: existing event-driven controller ref; the new `PostReplacementDamageSource` is its object-valued sibling (same drain accessor).
- `crates/engine/src/types/ability.rs:4866` / `:4881` — `PostReplacementDamageTarget` / `PostReplacementDamageTargetOwner`: existing prevented-event refs the new source ref is threaded alongside in every match.
- `crates/engine/src/types/ability.rs:17960` — `TriggerCondition::EventDamageSourceMatchesFilter`: the established source-vs-event sibling split the new `AbilityCondition::PostReplacementDamageSourceMatchesFilter` mirrors.
- `crates/engine/src/game/effects/prevent_damage.rs:179` — `TargetFilter::Player => Some(any_player_damage_filter())`: the existing "player recipient yields `Some`, never `valid_card`" pattern the compound arm copies.
- `crates/engine/src/game/effects/prevent_damage.rs:195` — `TargetFilter::Any | ParentTarget => None` in `typed_recipient_valid_card_filter`: the existing exclusion list the new arm extends.
- `crates/engine/src/game/replacement.rs:1290` — `shield_has_per_event_execute_followup`: the pre-existing per-event-vs-batch discriminator; `rider_reflects_per_event_damage_source` is its structural twin.
- `crates/engine/src/game/replacement.rs:7008` — the base `batched_combat_all_shield` guard this PR extends with the `!rider_reflects_per_event…` conjunct.
- `crates/engine/src/parser/oracle_replacement.rs:6682` — `parse_damage_source_subject`: the existing controller-axis tag family ("you control" → `ControllerRef::You`) extended with the opponent axis, same `tag().parse()` idiom.
- `crates/engine/src/parser/oracle_replacement.rs:10192` — `clause_is_prevented_this_way_rider`: the existing rider classifier (`preceded(alt(when/whenever/if), tag("damage is prevented this way,"))`) that the source-gate extractor generalizes.

(All line numbers verified at the PR base by reading `git show upstream/main:<path>`.)

## Gate A

```
Gate G PASS (router/grant architecture: strict router vs permissive grant boundary intact)
Gate A PASS head=9f1befc396438077fd64d6f361585cb2eb009621 base=ac1a581daaa1791e51ad9d5b9e380a4af9450de9
```

## Implementation-review record (final `review-impl`, independent reviewer, against this head)

An isolated reviewer ran the full `/review-impl` lens set against the complete branch diff, including the maintainer-named provenance and per-event-vs-batched cases. Record of material results:

- **Provenance trace — PASS.** Source authority is captured once as `proposed_damage_source` before the applier consumes the event (`replacement.rs`), stashed onto the resident drain's `event_source` at the `Prevented` arm, and read back by *both* consumers (gate and reflection target) via `post_replacement_event_source()` from the same single-authority resident drain, retired per-event by `finish_dispatch`. No cross-event or cross-attacker leak path was found. Batched combat forces the per-event path (exclusion from `batched_combat_all_shield`), so each event carries its own source + amount and is drained inline before the next iterates; `KeepResident` never clobbers a sibling.
- **New-field threading — PASS.** `permanent_type` is threaded at every construction site and its sole runtime consumer; the two new `TargetFilter` variants and the new `AbilityCondition` variant are threaded through every exhaustive match across engine, parser, coverage, and `mtgish-import`. The one wildcard site (`condition_depends_on_effect_performed`) is intentional — the new condition is handled by the dedicated independent-gate branch in `resolve_chain_body`.
- **Vacuous-negative check on all runtime + parser tests — PASS** (every negative assertion paired with a positive reach-guard).
- **Finding (fixture serde-compat) — investigated, refuted with CI evidence.** The reviewer flagged the new `Option<CoreType>` field as a potential fixture-deserialization break absent `#[serde(default)]`. Upstream CI on the prior head (both Rust test shards, which eagerly load `integration_cards.json`) is green — serde's derive treats a missing `Option` field as `None` by default, so old-shape data loads correctly. No change required; noted here for the record.
- **Finding (N≥2 simultaneous attackers untested) — fixed in this head.** The per-event batch routing was production-exercised only at N=1; the sequential two-event test bypassed the combat batch. This head adds a two-attacker single-declare-step combat test asserting each attacker is reflected exactly its own damage amount (revert-fails if routing folds both events into the aggregate or discards the second sibling's continuation).
- **Finding (rider gate dropped when a condition pre-exists) — fixed in this head.** The source-type gate now composes with any existing rider condition via the codebase's existing conjunction form instead of silently skipping; a parser test pins the composed shape. (Unreachable for Comeuppance itself; class-hardening for future riders.)
- **Deferred with note:** a CR 616.1 multi-replacement ordering fixture (prevention + damage doubler on one event) — the shared ordering pipeline is unchanged by this PR and the reflection reads the final prevented amount from the same drain; flagged as a follow-up test rather than a blocker.
- **mtgish-import touch — compulsory.** `target_filter_variant_name` matches `TargetFilter` exhaustively with no wildcard, so the two new variants break the workspace build without the added arms. The new `AbilityCondition` variant is deliberately *not* threaded into mtgish (no exhaustive match exists there) — touch only what the compiler forces.

## Tests

6 original integration tests (combat prevention + reflection, planeswalker leg, noncreature-source → controller leg, own-source negative, creature-recipient negative, per-event sequential pair) + the new two-simultaneous-attackers batch test; parser tests for the compound recipient, both gated riders, the over-trigger guard, Channel Harm/Deflecting Palm regressions, and the composed gate+condition case.

CR 615 (prevention effects) · CR 616.1 (replacement ordering, unchanged pipeline) · CR 514.2 (until-end-of-turn expiry).


matthewevans (changes requested):
**Request changes — this new prevention/replacement mechanism needs a completed engine implementation cycle before it can be merged.**

## 🔴 Blocker

**[HIGH]** The PR introduces coupled parser, serialized ability, and per-event replacement machinery without the required reviewed implementation evidence. Evidence: `crates/engine/src/types/ability.rs` adds the new recipient/source references, `crates/engine/src/game/replacement.rs` adds the per-event rider dispatch, and `crates/engine/src/parser/oracle_replacement.rs` adds the grammar/lowering; the PR body has no `## Anchored on`, Gate A, or final `review-impl` record. Why it matters: a prevention rider that binds an event source must preserve that authority across every replacement/resumption path, and a first-pass 24-file implementation cannot establish that by CI alone. Suggested fix: run the engine-implementer plan → implementation → final review cycle against the current head, including source/recipient provenance and per-event versus batched rider cases, then resubmit the resulting evidence and any required corrections.

## ✅ Clean

The current parse-diff is internally scoped to the compound-recipient and source-filter class (23 cards / 13 signatures); it is not treated as unexplained blast radius.

**Recommendation: complete the implementation-cycle review before another merge-readiness pass.**

mcbradd:
Resubmitting with the requested implementation-cycle evidence — the PR body has been rewritten to carry the full record: `## Anchored on` (nine base-verified analog citations), `## Gate A` (verbatim output on this exact head), and the final independent `review-impl` record, including the two cases your block named:

- **Source/recipient provenance**: traced end-to-end and confirmed sound — one authority (`proposed_damage_source` → resident drain `event_source`), captured before the applier consumes the event, consumed by both the gate and the reflection target, retired per-event. The full trace is in the review record in the body.
- **Per-event vs batched riders**: the review found this correct by trace but under-tested — production combat exercised it only at N=1. This head adds `two_simultaneous_attackers_each_reflected_own_power`: two attackers with distinct powers in one declare-attackers step, each asserted to be reflected exactly its own prevented amount (an aggregate-fold or sibling-discard regression fails it), with a CR 508.1f tapped reach-guard so the prevention negative can't pass vacuously.

The review also surfaced one class-hardening fix shipped here (the rider source gate now conjoins with any co-existing condition via `merge_ability_condition` instead of being dropped — parser test pins the composed shape), one refuted finding recorded with its evidence (fixture serde-compat — green CI on both test shards + serde's missing-`Option`→`None` semantics), and one deferred follow-up noted in the record (CR 616.1 multi-replacement ordering fixture; the shared ordering pipeline is untouched by this PR).

Branch is current with `main` as of this push, so CI regenerates the parse diff against the live baseline for reconciliation.

matthewevans:
Maintainer re-review complete for `9f1befc396438077fd64d6f361585cb2eb009621`.

The requested N≥2 production-combat regression and condition-conjunction fix are present in the reviewed delta. I found no additional concrete blocker in that pass. Rust’s required check suite is still running, so this remains held pending CI; it has not been approved or enqueued.

matthewevans (approved):
Reviewed current head `9f1befc396438077fd64d6f361585cb2eb009621`: the per-event prevention/reflection path and N≥2 production-combat regression address the prior request. Required checks are green. Approved for the merge queue.

PR #6142, opened 2026-07-18
feat(parser+engine): Winnowing — per-player relative keep-one sweeps

mcbradd:
## Summary

Implements **Winnowing** (Lorwyn Eclipsed, ECL): *"Convoke … For each player, you choose a creature that player controls. Then each player sacrifices all other creatures they control that don't share a creature type with the chosen creature they control."*

**Zero new enum variants** — the card rides entirely on existing machinery (`Effect::ChooseAndSacrificeRest` with `CategoryChooserScope::ControllerForAll`, the Tragic Arrogance template):

- **Parser, choose clause** — new single-bare-category pattern in `parse_category_and_sacrifice_rest`: "for each player, you choose a `<type>` that player controls", generic over `CoreType` via the shared `parse_category_item` + `core_type_to_type_filter` (CR 608.2d + CR 701.21a). Placed before the multi-category list pattern, whose missing-"from among" `.ok()?` short-circuit otherwise swallowed the single-category form (root cause of the previous `Unimplemented("for")` parse). Provably disjoint from patterns 2–4 (regressions: Tragic Arrogance, Cataclysmic Gearhulk, Slaughter the Strong, Natural Balance all unchanged).
- **Parser, sweep qualifier** — the keep-one sweep continuation accepts an optional relative qualifier "that don't share a creature type with the chosen creature they control" via the shared `parse_shared_quality_clause` combinator (4 existing consumers), folding `SharesQuality { CreatureType, ParentTarget, DoesNotShare }` into the sacrifice filter (CR 608.2c). This also fixes the dangling-sub-ability parse the trailing qualifier previously caused.
- **Runtime** — the sweep binds each scope player's **own** kept creature as the filter recipient (`FilterContext.recipient_id`), so the shared-type exemption is evaluated per player against that player's chosen creature — not the caster's, not a global keeper. Sacrifices land simultaneously (CR 101.4). Inert for existing keep-one cards (their filters carry no reference props); no `filter.rs` changes (the recipient threading pre-existed).

## Tests

- Parser: pattern + qualifier building-block tests; full-card probe asserting the folded filter **contains** the `SharesQuality` prop and no dangling sub-ability (guards the silent sacrifice-everything failure mode); 4 exemplar regressions.
- 6 integration tests (`winnowing.rs`): mixed-tribe per-player exemption; cross-player negative guard (a creature sharing a type with the *caster's* keeper but not its own controller's is still sacrificed — discriminates per-player binding); empty player skipped; kept Changeling → nothing sacrificed; unkept Changeling survives; non-creatures untouched + one simultaneous batch.
- Full engine suite, clippy, combinator Gates A+G, semantic-audit baseline: all green; Winnowing at 0 audit findings.

CR 701.21a (sacrifice) · CR 608.2c (resolution in the order written) · CR 608.2d (resolution-time choices by the spell's controller) · CR 101.4 (simultaneous actions).


matthewevans (changes requested):
## Request changes

**[HIGH]** The integration-test module is declared but not included in this PR. Evidence: `crates/engine/tests/integration/main.rs:799` adds `mod winnowing;`, while the current head `4b666798df15eb461b6a1ab6cb92e75d005c1819` has no `crates/engine/tests/integration/winnowing.rs`; the current Rust lint run fails with E0583 for that module. Why it matters: this branch cannot compile, and the claimed runtime coverage for the per-player `ParentTarget` binding is absent. Suggested fix: add the intended integration test module (including a two-player, distinct-creature-type runtime case) or remove the declaration until the test is ready.

**[HIGH]** The parser coverage artifact is not yet available for the current merge-main head. Evidence: the sticky parse-diff comment is baseline-pending for `46c5223763`, while the source head is now `4b666798`; Card data is still running. Why it matters: Winnowing support cannot be reconciled with the card-level parser blast radius. Suggested fix: let current CI publish the parse diff, then include/reconcile it before re-review.

matthewevans:
**Blocked — the new parser artifact confirms an unreviewed class-wide change, and the runtime test module is still absent.**

## 🔴 Blocker

**[HIGH]** The current parse artifact changes 25 cards across 14 signatures, not only Winnowing. Evidence: the `<!-- coverage-parse-diff -->` artifact on head `4b666798` adds `ChooseAndSacrificeRest` for 25 cards and removes/replaces 10 prior signatures (including Ascent of the Worthy, Benthic Anomaly, Divine Reckoning, Goliath Hatchery, Canoptek Wraith, and Sakashima); the PR body presents Winnowing as the implementation target and gives no card-by-card reconciliation. Why it matters: the generic bare-category parser and recipient rebinding now change runtime behavior for a broader class, so Winnowing-only parser probes cannot establish rules correctness for the additional cards. Suggested fix: reconcile every affected card against its Oracle text, state the intended class explicitly, and add discriminating runtime coverage for representative non-Winnowing members and the per-player scope branch.

**[HIGH]** The registered integration module is still missing. Evidence: `crates/engine/tests/integration/main.rs:799` declares `mod winnowing;`, but the current tree has no corresponding `winnowing.rs`; the current Rust lint and both Rust-test shards fail with E0583. Why it matters: the claimed six runtime tests do not exist on the reviewed head, leaving the new per-player `ParentTarget` binding unverified. Suggested fix: add the promised registered runtime module, including the cross-player negative discriminator, or remove the declaration and keep coverage honest until it is supplied.

## ✅ Clean

The current parse-diff artifact is now available; it supersedes the earlier baseline-pending evidence.

**Recommendation: request changes; do not enqueue until the class-wide delta and runtime test coverage are reconciled.**

mcbradd:
Both findings addressed on head `dd01924d9`:

1. **Missing module (E0583)** — confirmed and fixed. `crates/engine/tests/integration/winnowing.rs` existed locally with the full test suite but was never staged in the original commit; only the `mod winnowing;` registration landed. The file is now committed on top of the merge-main head `4b666798`. It includes the requested two-player, distinct-creature-type runtime case: `winnowing_scopes_shared_type_reference_per_player` gives each player creatures of different type profiles and asserts each player's kept creature protects only *that player's* type-sharers (per-player `ParentTarget` binding), plus changeling interactions on both the kept and unkept side, a no-creatures player skip, and the noncreature/one-step sweep case.

2. **Parse-diff baseline pending** — the new head is a descendant of the merge-main head, so CI on this push regenerates the parse diff against the current baseline. No parser source changed in the fix commit (test file only), so the parse diff should be unchanged from the previous run's scope.

Local pre-push gates on `dd01924d9`: fmt, clippy `--all-targets -D warnings`, card-data release validation, combinator Gate G + Gate A (`Gate A PASS head=dd01924d9f9ef1beb97e8e6b55c07de4b6027b09 base=4b666798df15eb461b6a1ab6cb92e75d005c1819`), and the 595-test parser suite all pass.

matthewevans:
**Held pending current-head CI.** I applied a maintainer fixup on `e9a4081403` after the prior head's test shards found that the new generic bare-category parser arm also captured ordinary `choose a creature they control` syntax. The arm now runs only through the existing `for each player, you choose` entry point; this preserves Winnowing while restoring the two existing target-selection parser paths.

The earlier missing `winnowing.rs` blocker is resolved on `dd01924d9`, and the current worktree review found the per-player recipient binding and its cross-player runtime discriminator correctly placed. The new push needs the current CI run and parse-diff sticky artifact before approval; no contributor action is requested at this point.

matthewevans (approved):
Approved: the current-head Winnowing implementation uses the existing choose-and-sacrifice authority, scopes each shared-type comparison to that player's chosen creature, and has discriminating parser and runtime coverage.

Issue #175, opened 2026-05-04
Phyrexian mana cards can't be cast by paying 2 life


matthewevans:
## Bug Report

**Source**: Discord / Engine analysis
**Discord thread**: [Phyrexian mana unusable](https://discord.com/channels/1485498006781427802/1485498006781427802)

### Cards
All cards with Phyrexian mana symbols ({U/P}, {B/P}, {R/P}, etc.) — e.g., Gitaxian Probe, Mental Misstep, Dismember

### Actual Behavior
Cards with Phyrexian mana symbols cannot be cast by paying 2 life for the Phyrexian component. The 2-life alternative payment path is either not offered or not functional.

### Expected Behavior
For each Phyrexian mana symbol in a cost, the player should be able to pay either the colored mana or 2 life (CR 107.4f). Both options must be presented and resolved correctly.

### Parser Status
N/A — the Phyrexian mana payment UI/engine flow is broken, not the parser.

### Notes
Affects an entire mechanic class. The `WaitingFor::PhyrexianPayment` state and its resolution path need to be fully wired in the mana payment flow.

matthewevans:
Fixed. Phyrexian mana is fully implemented: interactive `WaitingFor::PhyrexianPayment` choice, `SubmitPhyrexianChoices` action, life deduction (CR 107.4f), and AI auto-resolution. Integration tests confirm casting with life payment works (see `casting.rs::can_cast_phyrexian_cost_spell_when_life_covers_phyrexian_shard` and `escape_phyrexian_cost_deducts_life_after_exile`).

PR #6146, opened 2026-07-18
feat(parser+engine): Mizzix's Mastery — coordinated copy-cast tails and overload self-exile guard

mcbradd:
## Summary

Implements **Mizzix's Mastery** (OTC #175): *"Exile target card that's an instant or sorcery from your graveyard. For each card exiled this way, copy it, and you may cast the copy without paying its mana cost. Exile Mizzix's Mastery. / Overload {5}{R}{R}{R}"*

No new enum variants or effects — the card rides the existing `Effect::CastCopyOfCard` seam (CR 707.12/707.12a) and the existing Overload machinery. Two root-cause fixes:

- **Coordinated copy-cast tails** — the "for each …" prefix suppresses the chunk splitter's comma+and split, so ", and you may cast the copy without paying its mana cost" was swallowed (`SwallowedClause: Optional_YouMay`) and the card produced an inert, uncastable copy. New `try_split_copy_cast_compound` in `lower_imperative_clause` (following the established `try_split_*_compound` pattern) splits the coordinated tail into `CopySpell` + `CastFromZone` via the **shared** `try_parse_cast_effect`, so the untouched `fold_cast_copy_of_card_defs` performs the fusion — the fold stays the single fusion authority, and the count-bearing "cast up to N of the copies" path is preserved. Guarded on without-paying, so God-Eternal Kefnet's `{2}`-reduction copy-cast stays correctly excluded.
- **Overload self-exile guard** — `transform_ability_def` promoted *every* `ChangeZone` to `ChangeZoneAll`, including targetless `ChangeZone { target: SelfRef }` clauses. CR 702.96a replaces only "target" with "each"; the promoted form scanned the battlefield (the resolving spell is on the stack), found nothing, and dropped the self-exiling spell into the graveyard instead of exile. The transform now skips `SelfRef` zone-moves — first overload card with a self-exile clause; any future one is covered.

Also switched the Mizzix parser/swallow fixtures from doctored two-sentence text to the real printed comma+and wording (the doctored text is why this gap survived), with two-sentence (Mnemonic Deluge class) and bare "copy it" (Fork class) regressions retained.

## Tests

- Parser: printed-text probe asserting the folded `CastCopyOfCard{TrackedSet}` shape with the self-exile sub intact; two-sentence regression; bare "copy it" → `CopySpell` regression; swallow fixture on printed text.
- 2 overload unit tests: `SelfRef` not promoted; target-bearing exile in the same tree still promotes.
- 4 integration tests: normal cast → copy resolves (life delta) and Mastery self-exiles; creature card not a legal target; overload → each instant/sorcery exiled+copied, creature spared, accept-one/decline-one (+3 not +6, CR 707.12a per-copy choice), Mastery ends in EXILE; overload on empty graveyard resolves cleanly.
- Full engine suite green (the only failure in the shared dev tree was the test-registration guard tripping on a stray untracked file from a sibling branch — absent in a clean checkout); clippy clean; combinator Gates A+G pass; semantic-audit: Mizzix's Mastery now 0 findings.

## Known pre-existing gap (not introduced or fixed here — separate issue to follow)

Declining **all** copies of a single-card tracked set still casts one: the empty `ChooseFromZoneChoice` selection re-binds to the tracked set in the resume continuation. Affects all 9 `CastCopyOfCard` cards, violates CR 707.12a, lives in unchanged shared machinery — filing separately.

CR 707.12 / 707.12a (casting copies, per-copy choice) · CR 707.10 (copying) · CR 118.9 (cast without paying) · CR 702.96a/c (overload text change) · CR 608.2c (order written).


matthewevans (changes requested):
The overload guard and the composable copy/cast head are on the right seams, but one current-head test helper needs correction before this can land.

`move_to_graveyard` guards `state.objects.get_mut(&id)` and then immediately indexes `state.objects[&id]` to obtain the owner. If the helper ever receives an absent object, it still panics; the preceding guard does not protect that access. Keep the owner read inside the guarded branch (or require the object with a single explicit expectation) and return without mutating the battlefield/graveyard when it is absent.

Evidence: current head `d5715580a20bbe03fc6945ac02070e13026bd0fa`, `crates/engine/tests/integration/mizzixs_mastery.rs:243-254`; the current unresolved Gemini thread identifies the same path. Confidence: high. Rust lint/tests are also still running, so please update the head and let current CI complete.

mcbradd:
Addressed on the new head:

- **`move_to_graveyard` unsafe indexing** (your finding + the Gemini thread): the unconditional `state.objects[&id]` after the guarded `get_mut` is gone. The helper now takes the object through a single `let-else` expectation — the owner is read from the already-retrieved reference inside the guarded path, and an absent object fails immediately with an explicit panic message naming the id rather than an opaque index panic. (Chose the "require with a single explicit expectation" option over silently returning: in a test helper, being handed a nonexistent object is a test bug that should fail loudly, not a state to paper over.)
- **Head currency**: merged upstream `main` (through d04420b15) into the branch, so CI on this push runs against the current baseline and regenerates the parse diff.

matthewevans (changes requested):
**Request changes — the parser and overload seams look sound, but the current head is missing the required current-head parse-diff evidence.**

## 🔴 Blocker

- `crates/engine/src/parser/oracle_effect/mod.rs:16530` changes parser behavior, yet the required `<!-- coverage-parse-diff -->` comment was last updated at 2026-07-18T09:09:00Z for commit `d5715580a20bbe03fc6945ac02070e13026bd0fa`; the current head is `0776c3d646ddc88230dfd505f1cfe9ca8a4361f2`, pushed afterward. This prevents a current-head check that the claimed Mizzix-only signature replacement remains the complete parser blast radius. Regenerate or refresh the parse-diff artifact for this head, then retain the full card-level result on the PR. The existing comment is useful baseline evidence but cannot certify the new commit.

## ✅ Clean

- The prior `move_to_graveyard` finding is resolved at `crates/engine/tests/integration/mizzixs_mastery.rs:243`: the object is now retrieved once through `let Some(...)`, so the unsafe map index is gone.
- The full sticky diff does explain the intended Mizzix shape change on its earlier head: `CopySpell` is replaced by `CastCopyOfCard`; Scryfall's current Oracle text contains the coordinated `copy it, and you may cast the copy` clause; and CR 702.96a/b plus CR 707.12/707.12a support the overload and individual-copy choices. The local rules file and Scryfall API were checked.

**Recommendation: refresh the current-head parse-diff artifact, then request re-review; no implementation redesign is indicated by this pass.**

matthewevans (approved):
Approved after current-head review. The prior `move_to_graveyard` indexing finding is fixed at `mizzixs_mastery.rs:258`; the overload `SelfRef` guard correctly leaves targetless self-exile untouched under CR 702.96a, and the copy/cast splitter reuses the existing `CastFromZone` → `CastCopyOfCard` fold. Maintainer follow-up `d2d3f96` also adds the CR 707.12a single-copy decline regression; it drives the production choice/resume path and would catch a tracked-set rebind.

mcbradd:
Your `normal_cast_can_decline_its_only_copy` test now passes unmodified — and it exposed something better than the #6147 hypothesis.

**Actual root cause** (verified by runtime instrumentation, not the tracked-set re-bind we guessed in #6147): the `ChooseFromZoneChoice` → `SelectCards` partition-forwarding arm in `engine_resolution_choices.rs` forces the **un-selected** cards onto the continuation's `sub_ability` (the chosen-here/rest-there pattern, e.g. Atraxa's "put the rest on the bottom"). For a `CastCopyOfCard` sub that is exactly backwards — it casts the copies the player *declined*. Declining the only copy therefore cast it (your test), and accepting one of two cast the **wrong** one — our overload accept-one test was passing by coincidence because both fixture copies gain the same 3 life. That test now also discriminates chosen-vs-declined correctly.

**Fix** (one hunk, `engine_resolution_choices.rs`): skip the unchosen-partition forwarding when the partition sub is `CastCopyOfCard`, mirroring the existing `PutCounter` "not a rest-consumer" carve-out at the same seam. The chosen copies still reach the caster via the head-targets inheritance the accept path already used. Class-general: fixes the decline path for all nine `CastCopyOfCard` cards from #6147 with no card- or Mizzix-specific branch. CR 707.12a grep-verified ("allows that player to choose individually, for each of those objects, whether or not to cast it").

**Blast radius checked**: genuine zone partitions (Atraxa), per-player accumulation (Breach the Multiverse), per-category (Sanar), and tracked-set-consuming continuations (End-Blaze Epiphany, Windbrisk, anaphor suite) are untouched paths — 43 neighboring integration tests plus all 5 `mizzixs_mastery` tests and the 11 `cast_copy` unit tests pass; clippy `--all-targets` clean.

matthewevans (approved):
Approved after current-head handler fix `87ac14c30b1f3b68ce5497c3fd5ce3fa7a2de215`.

The partition exception remains at the shared choice/resume seam. The overload test now distinguishes the selected +3 copy from the declined +5 copy, so it fails if the continuation receives the unchosen partition. The prior Gemini/helper finding remains resolved; the current parse-diff is Mizzix-only.

PR #6155, opened 2026-07-18
Add Hit the Mother Lode (shared difference-anaphor binder + discover resolution fixes)

mcbradd:
## Summary

Implements **Hit the Mother Lode** (LCI): "Search your library for a card, put it into your hand, then shuffle. Discover X, where X is the difference between the highest mana value among cards in your hand and the mana value of the card you found." — specifically the class of spells whose later clause references "the difference" computed from earlier clauses via a bare anaphor.

### Parser

- `oracle_effect/token.rs`: bare **"the difference"** anaphor now resolves to a bound `QuantityRef::Difference` when a difference binder is in scope — covering the templating class, not the single card.
- `oracle_effect/mod.rs`: hoisted the shared difference-binder recognition out of the trigger parser into the spell clause seam (`parse_effect_chain_ir`), so both triggered abilities and spell effect chains bind "the difference between X and Y" once, at a single authority. The binder is **bind-only** — it records the referent without consuming the clause, preserving downstream parsing.
- `oracle_trigger.rs`: −76 lines — the previously trigger-local binder functions moved to the shared seam.

### Engine

- `game/quantity.rs`: presence helpers so resolution can distinguish "referent exists but is zero" from "referent was never established".
- `game/effects/mod.rs`:
  - Missing-referent guard is **presence-based**, not value-based (CR 608.2h: an undefined value in an instruction means that part of the effect does nothing; a legally-computed 0 is still a real 0).
  - Conditional-sub evaluation deferred to the resolution continuation so the referent is bound before conditions read it (CR 608.2c: instructions performed in the order written).
  - Discovered-referent stamp recorded at resolution time.
- `game/engine_resolution_choices.rs`: free-cast continuation drain — after a discover free-cast completes, resolution correctly resumes the remaining effect chain (CR 608.2g: information from the resolution process is used for the rest of the resolution).

### Tests

- `tests/integration/hit_the_mother_lode.rs` — 5 scenarios: full search→shuffle→discover flow with correct X; X computed as highest-hand-MV minus found-card MV (operand order pinned); discover free-cast path resumes the chain; missing-referent (empty hand / no card found) does nothing per CR 608.2h; decline-free-cast puts the card in hand.
- `oracle_parser.rs` probe pinning the parsed AST shape.

All CR citations grep-verified against `docs/MagicCompRules.txt`: CR 701.57a/b/c (discover), CR 608.2c/g/h/k (resolution semantics), CR 202.3e (mana value of cards in zones), CR 400.7j, CR 111.10a.

> Reviewer footnote: an optional hardening test for the mana-value-0 edge (found card with MV 0) was suggested and is a good follow-up; the presence-based guard already distinguishes that case from a missing referent by construction.

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -6348,6 +6348,64 @@ pub(crate) fn drain_pending_player_scope_sacrifice_after_replacement(
 
 /// Resolve an ability and follow its sub_ability chain using typed nested structs.
 /// No SVar lookup, no parse_ability(). The depth is bounded by the data structure.
+/// CR 608.2c: True when `condition` is a quantity comparison awaiting a
+/// resolution-only object referent that the currently-paused effect will only
+/// bind after its choice resolves (Hit the Mother Lode: "if the discovered
+/// card's mana value is less than 10" — the discovered card is stamped onto the
+/// continuation only after the cast-or-hand offer). Such a condition must NOT be
+/// evaluated synchronously against the pre-choice state, where the referent
+/// reads as absent and the comparison collapses to a false gate; defer the sub
+/// to the continuation so it re-evaluates once the referent is bound.
+fn condition_awaits_resolution_only_referent(
+    condition: &AbilityCondition,
+    state: &GameState,
+    ability: &ResolvedAbility,
+) -> bool {
+    match condition {
+        AbilityCondition::QuantityCheck { lhs, rhs, .. } => {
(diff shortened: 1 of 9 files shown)


matthewevans (approved):
Approved after current-head maintainer review: the current parse-diff artifact covers the intended three-card difference-anaphor class, the runtime coverage exercises decline and free-cast continuations plus X=0/no-hit boundaries, and required checks are green.

Issue #4246, opened 2026-06-23
Braids, Ultimate Nightmare — [[Braids, Arisen Nightmare]] triggers fire correctly, however if the owner of Braids selec…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1518338641997332601/1518338641997332601

**Thread:** Braids, Ultimate Nightmare
**Cards:** braids, arisen nightmare
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
[[Braids, Arisen Nightmare]] triggers fire correctly, however if the owner of Braids selects no, the engine takes life and draws the player cards even though they didnt sacrifice anything.

### braids, arisen nightmare
At the beginning of your end step, you may sacrifice an artifact, creature, enchantment, land, or planeswalker. If you do, each opponent may sacrifice a permanent of their choice that shares a card type with it. For each opponent who doesn't, that player loses 2 life and you draw a card.

---
<sub>report_id: `discord:1518338641997332601:1518338641997332601:0` · discord: `1518338641997332601/1518338641997332601`</sub>

Issue #491, opened 2026-05-18
Braids, Arisen Nightmare — [[Braids, Arisen Nightmare]] dealt the 2 damage to me rather than the opponent, though I did…


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1505776199404752916/1505776199404752916

**Thread:** Braids, Arisen Nightmare
**Cards:** Braids, Arisen Nightmare
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
Braids, Arisen Nightmare dealt the 2 life loss to the reporting player rather than the opponent, while still drawing the reporting player a card.

### Braids, Arisen Nightmare
At the beginning of your end step, you may sacrifice an artifact, creature, enchantment, land, or planeswalker. If you do, each opponent may sacrifice a permanent of their choice that shares a card type with it. For each opponent who doesn't, that player loses 2 life and you draw a card.

---
<sub>report_id: `discord:1505776199404752916:1505776199404752916:0` · discord: `1505776199404752916/1505776199404752916`</sub>

matthewevans:
Fixed in 0f30f06d415e3d0a9d6d8ca3d20b1cbb58b733c3 (with follow-up fa08d7e91fcf4863e423d65df986d9973f3d9c36).

Braids, Arisen Nightmare's end-step ability — "each opponent may sacrifice an artifact, creature, or land; for each opponent who doesn't, that player loses 2 life and you draw a card" — was mis-resolving the decline consequence. The "loses 2 life" was hitting Braids' controller instead of the declining opponent, the gate was inverted (firing on opponents who DID sacrifice rather than who didn't), and the "you draw a card" clause was running once total instead of once per declining opponent.

The decline consequence now resolves correctly per opponent: each opponent who declines loses 2 life, and you draw one card for each such opponent. Verified with a 3-player test (two opponents decline → both lose 2 life, you draw exactly 2) mutation-checked against the reverted fix.

PR #6163, opened 2026-07-18
test(engine): end-step regression for Braids opponent decline (#5988)

### LINKED ISSUE #5988: Braids, Arisen Nightmare bug — There's no consequence to an opponent failing to sacrifice a permanent of the appropriat…
Reported in Discord: https://discord.com/channels/1485498006781427802/1524738069222396034/1524738069222396034

**Thread:** Braids, Arisen Nightmare bug
**Discord thread id:** `1524738069222396034`
**Discord message id:** `1524738069222396034`
**Cards:** _none detected_

## Summary
There's no consequence to an opponent failing to sacrifice a permanent of the appropriate type.

## Reported text

There's no consequence to an opponent failing to sacrifice a permanent of the appropriate type. I should be drawing a card and the opponent should lose 2 life

---
<sub>report_id: `discord:1524738069222396034:1524738069222396034:0` · discord: `1524738069222396034/1524738069222396034`</sub>

andriypolanski:
Fixes #5988

## Summary

Discord report ([#5988](https://github.com/phase-rs/phase/issues/5988)): when an opponent fails to sacrifice a permanent sharing a card type with the controller's sacrificed permanent, there is no consequence — the opponent should lose 2 life and Braids' controller should draw a card.

Classifier verdict: **supported_aspect_defect** — the parser AST is correct; the suspected gap was runtime per-opponent `Not{IfYouDo}` decline routing or `ParentTarget` LKI for the `SharesQuality` filter.

Investigation on current `main` showed the engine already handles this correctly:

- **Chain-level tests** in `braids_arisen_nightmare_decline.rs` (issues #491 / #4246) cover decline-consequence routing, multi-opponent fan-out, controller decline, and accept-path `ParentTarget` LKI — all pass.
- **End-step trigger path** was untested: prior coverage built a hand-resolved `execute` chain rather than firing the real "beginning of your end step" trigger through `apply()`.

This PR adds end-to-end trigger-path regressions. Both scenarios pass on current head without engine changes.

## Oracle text (verified)

> At the beginning of your end step, you may sacrifice an artifact, creature, enchantment, land, or planeswalker. If you do, each opponent may sacrifice a permanent of their choice that shares a card type with it. For each opponent who doesn't, that player loses 2 life and you draw a card.

## Changes

- **`crates/engine/tests/integration/issue_5988_braids_arisen_nightmare.rs`** — two end-step integration tests:
  - **`braids_end_step_opponent_without_match_loses_two_and_controller_draws`** — P1 controls only a land; after P0 sacrifices a creature, P1 loses 2 life and P0 draws.
  - **`braids_end_step_opponent_declines_matching_sacrifice_loses_two_and_controller_draws`** — P1 has a matching creature but declines the optional sacrifice; same consequences.
- **`crates/engine/tests/integration/main.rs`** — register the new module.

No engine, parser, frontend, or AI code changes.

## Test plan

```bash
cargo fmt --all -- --check
cargo test -p engine issue_5988
```

Expected: both integration tests pass.

Related Braids regressions (should remain green):

```bash
cargo test -p engine braids_arisen_nightmare
cargo test -p engine braids_
```

## Notes

If live Discord repros persist after merge, the next investigation target is **UI/AI optional-choice dispatch** (`DecideOptionalEffect { accept: false }` for the opponent), not the decline-consequence engine path — these tests prove the engine applies life loss and draw correctly once the trigger resolves.

### DIFF
--- a/crates/engine/tests/integration/issue_5988_braids_arisen_nightmare.rs
+++ b/crates/engine/tests/integration/issue_5988_braids_arisen_nightmare.rs
@@ -63,7 +63,7 @@ fn braids_end_step_opponent_without_match_loses_two_and_controller_draws() {
             WaitingFor::OrderTriggers { .. } => {
                 runner
                     .act(GameAction::OrderTriggers { order: vec![0] })
-                    .ok();
+                    .expect("order end-step triggers");
             }
             WaitingFor::OptionalEffectChoice { player, .. } => {
                 let accept = player == P0;
@@ -79,7 +79,9 @@ fn braids_end_step_opponent_without_match_loses_two_and_controller_draws() {
                     .expect("controller sacrifice");
             }
             WaitingFor::Priority { .. } if !runner.state().stack.is_empty() => {
-                runner.act(GameAction::PassPriority).ok();
+                runner
+                    .act(GameAction::PassPriority)
+                    .expect("pass priority while stack resolving");
             }
             _ => runner.pass_both_players(),
         }
@@ -130,7 +132,7 @@ fn braids_end_step_opponent_declines_matching_sacrifice_loses_two_and_controller
             WaitingFor::OrderTriggers { .. } => {
                 runner
                     .act(GameAction::OrderTriggers { order: vec![0] })
-                    .ok();
+                    .expect("order end-step triggers");
             }
             WaitingFor::OptionalEffectChoice { player, .. } => {
                 let accept = player == P0;
@@ -146,7 +148,9 @@ fn braids_end_step_opponent_declines_matching_sacrifice_loses_two_and_controller
                     .expect("controller sacrifice");
             }
             WaitingFor::Priority { .. } if !runner.state().stack.is_empty() => {
-                runner.act(GameAction::PassPriority).ok();
+                runner
+                    .act(GameAction::PassPriority)
+                    .expect("pass priority while stack resolving");
             }
             _ => runner.pass_both_players(),
         }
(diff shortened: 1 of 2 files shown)


matthewevans (changes requested):
**Request changes — the new end-step coverage is useful, but it currently masks engine-action failures.**

## 🔴 Blocker

**[MED]** Both new production-pipeline loops discard `OrderTriggers` and `PassPriority` errors via `.ok()`. Evidence: `crates/engine/tests/integration/issue_5988_braids_arisen_nightmare.rs:73-91` and `:136-154`. Why it matters: a changed or failed trigger/priority action can be silently skipped, allowing the loop to reach an incidental terminal state and turn this regression into a false green. Suggested fix: replace these `.ok()` calls with `.expect(...)` (or assert the intended benign error explicitly) so the test proves every production action succeeded.

## ✅ Clean

The test adds a genuinely distinct trigger-dispatch path beyond the existing `braids_arisen_nightmare_decline.rs` execute-chain tests, so it clears the standalone-test value bar once action failures are surfaced.

**Recommendation: make the action outcomes strict, then re-request review.**

matthewevans (approved):
Approved after re-review of `5e6aacbfb511e9fdcedd058089a61af859787325`.

The end-step regressions retain their production trigger-path coverage and now propagate `OrderTriggers` and `PassPriority` failures rather than masking them. The required CI checks are green; the independent security scan passed.

PR #6191, opened 2026-07-19
fix(engine): treat a tracked-set-scaled sacrifice count as a set consumer

### LINKED ISSUE #5991: Malfegor — Its ability that forces the opponent to sacrifice their creatures after I discard my hand is malfunctioning
Reported in Discord: https://discord.com/channels/1485498006781427802/1524669435342753912/1524669435342753912

**Thread:** Malfegor
**Discord thread id:** `1524669435342753912`
**Discord message id:** `1524669435342753912`
**Cards:** _none detected_

## Summary
Its ability that forces the opponent to sacrifice their creatures after I discard my hand is malfunctioning

---
<sub>report_id: `discord:1524669435342753912:1524669435342753912:0` · discord: `1524669435342753912/1524669435342753912`</sub>

tryeverything24:
Fixes #5991.

Malfegor: "Flying. When Malfegor enters the battlefield, discard your hand, then each opponent sacrifices a creature of their choice for each card discarded this way."

## Root cause & fix

The compound trigger parses correctly into a `DiscardCard` producer chained to a `Sacrifice` sub-ability whose count reads `FilteredTrackedSetSize { caused_by: Discarded }` — the standard "for each X discarded/exiled/milled this way" idiom already used elsewhere in the engine (Kylox's `ExileTop`, Seasoned Pyromancer's `repeat_for`).

`effect_references_tracked_set()`, which decides whether a preceding producer effect needs to publish its affected object ids into the chain-scoped tracked set before a downstream consumer reads it, had no arm for `Effect::Sacrifice`. So the discard step's affected ids were never published, and the sacrifice count's `FilteredTrackedSetSize` read an empty or stale set and resolved to 0 — Malfegor's opponent was never asked to sacrifice anything, a fully-parsed card that silently does nothing.

Fixed by adding a `Sacrifice { count, .. } => quantity_hits_tracked(count)` arm, mirroring the existing `ExileTop` arm immediately above it.

## Testing

New end-to-end test driving the real `DiscardCard` → `Sacrifice` resolution chain via `resolve_ability_chain` (the same production entry point every triggered ability resolves through), using the parser's own `parse_effect_chain` output, not a hand-built `ResolvedAbility`. Discriminating scenario: controller discards a 2-card hand, opponent controls 3 creatures — must be asked to sacrifice exactly 2 (not a fixed 1, not 0, the two most plausible pre-fix symptoms).

Verified locally: fmt clean, clippy clean (`-D warnings`), 3545 integration tests pass (3544 pre-existing + 1 new), 0 regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracked-set detection across all effect-embedded quantity expressions, including token “enter with counters” and chained ability behavior.
  * Fixed Malfegor’s “discard your hand, then each opponent sacrifices” so opponents are prompted to sacrifice the correct number.
  * Ensured no sacrifice prompt occurs when the discarded hand is empty.
* **Tests**
  * Added an integration regression suite for Malfegor discard-to-sacrifice, including chained tracked-set counter sizing and coverage for empty-hand behavior.


matthewevans:
Maintainer follow-up: corrected the verified sacrifice citation (`CR 701.21a`) and routed tracked-set count/amount detection through the existing `Effect::count_expr()` authority, retaining the special token P/T and planar-deck fields. The Malfegor end-to-end regression remains in place. Held for fresh CI and parse-diff evidence on 2211732e3.

matthewevans (changes requested):
**Changes requested — the tracked-set consumer walk is incomplete.**

## 🔴 Blocker

`effect_references_tracked_set` now treats `Effect::count_expr()` as though it covers every quantity-bearing field, but `count_expr` is documented as the primary count/amount accessor (`crates/engine/src/types/ability.rs:14399-14403`). `Effect::Token` also evaluates each `enter_with_counters` quantity during resolution (`crates/engine/src/types/ability.rs:9939-9975`, `crates/engine/src/game/effects/token.rs:2039-2043`), while the current special case only examines token power/toughness (`crates/engine/src/game/effects/mod.rs:3740-3745`).

A producer followed by a token effect with fixed count and P/T but `enter_with_counters: [(+1/+1, TrackedSetSize)]` is consequently classified as non-consuming; the producer does not publish its set and the counter quantity resolves incorrectly. Other secondary quantity fields have the same structural risk, so adding only this one field would preserve a misleading generic invariant.

Please replace the primary-count shortcut with an exhaustive, authoritative traversal of every `QuantityExpr` carried by an effect (or an equivalent complete abstraction), and add a runtime regression covering tracked entry counters. The existing Malfegor sacrifice test remains useful but does not exercise this missing path.

## ✅ Clean

The sacrifice case is correctly identified as a tracked-set consumer, and the existing test drives the real discard/sacrifice pipeline.

**Recommendation: rework the quantity traversal, add the entry-counter regression, then request re-review.**

matthewevans (changes requested):
**Changes requested — the current head still has the incomplete tracked-set traversal.**

## 🔴 Blocker

`effect_references_tracked_set` says it walks every quantity position, but its `Effect::Token` arm only inspects `count`, `power`, and `toughness` at [effects/mod.rs:3912-3963](https://github.com/phase-rs/phase/blob/33c122623d707ef173c1d3c107bbc7d9cab58f86/crates/engine/src/game/effects/mod.rs#L3912-L3963). `Effect::Token` also carries `enter_with_counters: Vec<(CounterType, QuantityExpr)>` ([ability.rs:9996-10000](https://github.com/phase-rs/phase/blob/33c122623d707ef173c1d3c107bbc7d9cab58f86/crates/engine/src/types/ability.rs#L9996-L10000)), and the production token resolver evaluates every such quantity ([token.rs:2045-2051](https://github.com/phase-rs/phase/blob/33c122623d707ef173c1d3c107bbc7d9cab58f86/crates/engine/src/game/effects/token.rs#L2045-L2051)).

A producer followed by a fixed-count/fixed-P/T token whose entry-counter quantity is `TrackedSetSize` is therefore classified as non-consuming; the producer does not publish its set and the token receives zero counters. The added Malfegor test only exercises `Effect::Sacrifice` ([issue_5991_malfegor_discard_sacrifice.rs:115-204](https://github.com/phase-rs/phase/blob/33c122623d707ef173c1d3c107bbc7d9cab58f86/crates/engine/tests/integration/issue_5991_malfegor_discard_sacrifice.rs#L115-L204)), so it cannot detect that sibling path.

Please use an exhaustive authority that visits every `QuantityExpr` carried by an effect (including nested token entry counters), then add a production-chain regression with a tracked-set-scaled entry counter. `count_expr()` alone is insufficient because it intentionally exposes only the primary count/amount slot.

## 🟡 Non-blocking

The CR citations at [effects/mod.rs:3942](https://github.com/phase-rs/phase/blob/33c122623d707ef173c1d3c107bbc7d9cab58f86/crates/engine/src/game/effects/mod.rs#L3942) and [issue_5991_malfegor_discard_sacrifice.rs:115](https://github.com/phase-rs/phase/blob/33c122623d707ef173c1d3c107bbc7d9cab58f86/crates/engine/tests/integration/issue_5991_malfegor_discard_sacrifice.rs#L115) cite `CR 701.17a`, which is the mill rule; sacrifice is `CR 701.21a` and discard is `CR 701.9a` (verified in `docs/MagicCompRules.txt:3327-3331,3449-3451`). Please remove the unrelated citation or replace it with the rule the annotated behavior implements.

**Recommendation: complete the quantity traversal and its discriminating runtime test, correct the citations, then request re-review.**

tryeverything24:
Addressed both review points:

**Exhaustive quantity traversal**: replaced the enumerated primary-count match in `effect_references_tracked_set` with a new exhaustive authority, `Effect::for_each_quantity_expr` (`types/ability.rs`), which visits every `QuantityExpr` an effect directly carries — including the secondary slots `count_expr()` doesn't expose (`Token::enter_with_counters`, `Dig::keep_count_expr`, `PutSticker::max_ticket_cost`, dynamic `PtValue::Quantity` P/T, `ChangeZone`'s conditional entry counters, `ForEachCategory` per-member counts, ...) — and recurses through variants embedding another `Effect`/`ResolvedAbility`. The match is exhaustive with no wildcard arm (same convention as `count_expr()`/`target_filter()`), so any newly added variant must be explicitly classified rather than silently reporting no quantities. Deferred definition payloads (`AbilityDefinition`, `ContinuousModification`, ...) are documented as out of scope: their quantities evaluate in their own later resolution context, and current-chain consumption there is signalled by explicit flags (`CreateDelayedTrigger { uses_tracked_set }`).

**Tests**: added a matcher-boundary unit test (a FIXED-count/FIXED-P/T token whose only tracked-set reference is a `TrackedSetSize` entry counter must classify as consuming) and a production-chain regression through `resolve_ability_chain` (whole-hand discard of 2 → token enters with exactly 2 +1/+1 counters; zero counters is the pre-fix failure).

**Citations**: removed the mill rule (CR 701.17a); the comments now cite CR 701.9a (discard) and CR 701.21a (sacrifice).

Verified locally in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17516 passed, engine integration 3751 passed. Rebased onto current `main`.

matthewevans (changes requested):
**Changes requested — `for_each_quantity_expr` is still incomplete on 75391811e4469e880fa6b9d8c4d48a68e7823dce.**

## 🔴 Blocker

The new authority says it visits every `QuantityExpr`, but it classifies `Effect::Mana` as quantity-free at `crates/engine/src/types/ability.rs:14961`. `ManaProduction` carries live `count: QuantityExpr` slots (for example `Colorless` at `:1877-1880`, `AnyOneColor` at `:1888-1890`, and the other dynamic-production variants), and `game/effects/mana.rs:454-480` resolves them during this effect's resolution. A producer followed by one of those consumers still fails to publish its tracked set.

Two more live nested carriers have the same hole: `Effect::PayCost` visits only `scale` at `ability.rs:14772-14776` while its `AbilityCost` contains dynamic values such as `ManaDynamic`, `PayLife`, `Discard`, `PayEnergy`, and `PaySpeed` (`:8156-8175`, `:8251-8257`); and `Effect::RollDie` visits `count` but not `modifier` (`:14792-14793`), though `DieRollModifier::{Add,Subtract}` holds a `QuantityExpr` (`:1099-1103`) resolved in `game/effects/roll_die.rs:73-79`.

Please make the traversal genuinely compositional and exhaustive across these nested domain types (including recursive cost forms such as `Composite`, `OneOf`, `EffectCost`, and `PerCounter` where their quantities resolve in the current chain), then add discriminating production-chain coverage for the newly traversed live seams. This is the same tracked-set-publication defect class the PR is intended to remove, so the current "exhaustive" contract is not safe to merge.

## 🟡 Non-blocking follow-ups

The current CodeRabbit review correctly flags that the empty-hand regression at `issue_5991_malfegor_discard_sacrifice.rs:205-234` has no positive reach guard and carries an unrelated `CR 107.3a` citation. Please either add a direct event/state assertion proving the empty discard step ran, and replace the citation only after verifying it locally, or remove the unsupported citation.

## ✅ Clean

The Malfegor discard-to-sacrifice test and the token entry-counter regression both drive the relevant resolution chain and would catch their respective prior shortcuts. The current parse-diff sticky is `no_changes` for this head, matching the non-parser scope.

**Recommendation: complete the nested quantity traversal and its regressions, address the small test follow-up, then request re-review.**

matthewevans:
Correction to the blocker line reference: `AbilityCost::PayEnergy` / `PaySpeed` are at `crates/engine/src/types/ability.rs:8244-8249` on this head (the previously cited `:8251-8257` range begins `ReturnToHand`). The finding is unchanged.

tryeverything24:
Rebased onto current `main` to clear the merge conflict introduced by the #6193 merge (both added a `mod` line to `tests/integration/main.rs` at the same anchor — kept both). No content changes to the reviewed implementation. Full local gate re-verified post-rebase in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17529 passed, engine integration 3756 passed.

matthewevans (changes requested):
**Request changes — the new exhaustive visitor still omits direct `QuantityExpr` carriers.**

## 🔴 Blocker

**[HIGH]** `Effect::for_each_quantity_expr` is incomplete despite its exhaustive-contract documentation. Evidence: `crates/engine/src/types/ability.rs:14961` groups `Effect::Mana { .. }` with no quantities, while `ManaProduction` directly carries `count: QuantityExpr` (for example `Colorless` at `:1878-1881`, `AnyOneColor` at `:1888-1891`, and `AnyOneColorAmongPermanents` at `:2005-2008`). `Effect::PayCost` at `:14772-14775` walks only `scale`, dropping dynamic `AbilityCost` fields (`ManaDynamic` `:8156-8158`, `PayLife` `:8170-8172`, `Discard` `:8173-8175`, `PayEnergy`/`PaySpeed` `:8260-8266`). `Effect::RollDie` at `:14792-14794` walks only `count`, although `modifier` contains `DieRollModifier::{Add,Subtract}` values at `:1099-1103` and `RollDie` carries that modifier at `:11861-11868`. Why it matters: a tracked-set producer can fail to publish before any of these downstream consumers, yielding a zero or incorrect resolution; the visitor's claimed exhaustive contract would then silently regress as these variants evolve. Suggested fix: compose dedicated exhaustive visitors for `ManaProduction`, `AbilityCost`, and `DieRollModifier`, and add focused regressions proving the tracked-set classifier reaches each missing carrier.

**Recommendation: request changes — complete the visitor and its regression matrix before re-review.**

matthewevans (changes requested):
**Changes requested — the new exhaustive quantity visitor is still incomplete on `610d2abe839c6134bd93c37859b9e28155edd14f`.**

## 🔴 Blocker

**[HIGH]** The new `Effect::for_each_quantity_expr` still drops live nested quantity carriers. Evidence: `crates/engine/src/types/ability.rs:14961` groups `Effect::Mana` as quantity-free although its `ManaProduction` owns dynamic `count` values; `:14772-14776` visits only `PayCost::scale` and not its `AbilityCost`; `:14792-14794` visits only `RollDie::count` and not dynamic `DieRollModifier::{Add,Subtract}` quantities. Why it matters: a preceding producer will still fail to publish its tracked set for those current-chain consumers, so their tracked quantity resolves as zero or wrong despite the visitor’s exhaustive contract. Suggested fix: compose exhaustive visitors for `ManaProduction`, recursive `AbilityCost` (including nested forms), and `DieRollModifier`, invoke them from `Effect`, and add production-chain regressions for each seam.

## 🟡 Non-blocking

**[MED]** The empty-hand regression is a vacuous negative check and retains an unsupported CR citation. Evidence: `crates/engine/tests/integration/issue_5991_malfegor_discard_sacrifice.rs:205-234` only asserts no prompt/no battlefield change and never checks the captured resolution events; `:207` cites `CR 107.3a` for the zero-count no-op. Why it matters: an upstream short-circuit that skips the discard/tracked-set path still passes. Suggested fix: add a discard-resolution reach guard, then verify a replacement citation against the local Comprehensive Rules file or remove it.

## ✅ Clean

The current parse-diff sticky comment reports no card-parse changes, which matches this engine-only change. I also rechecked the current CodeRabbit findings; they identify the same remaining incomplete-traversal and test/citation issues.

**Recommendation: request changes — complete the nested visitor composition and add discriminating production-path tests before re-review.**

tryeverything24:
Completed the nested visitor composition in b94902b:

- **`ManaProduction::for_each_quantity_expr`** — no-wildcard match classifying all 15 variants; every dynamic `count` slot (`Colorless`, `AnyOneColor`, `AnyCombination`, `ChosenColor`, `OpponentLandColors`, `AnyCombinationOfObjectColors`, `AnyTypeProduceableBy`, `AnyInCommandersColorIdentity`, `AnyOneColorAmongPermanents`) is visited from the `Effect::Mana` arm.
- **`AbilityCost::for_each_quantity_expr`** — no-wildcard match over all 31 variants; visits `ManaDynamic`, `PayLife`, `Discard`, `PayEnergy`, `PaySpeed`, recurses through `Composite`/`OneOf`/`PerCounter`, and walks `EffectCost`'s embedded effect. `Effect::PayCost` now visits the cost plus `scale`.
- **`DieRollModifier::for_each_quantity_expr`** — `Add`/`Subtract` values, called from `Effect::RollDie` alongside `count`.

A full sweep of every `QuantityExpr` field in `types/ability.rs` surfaced five more live carriers, now walked with the same no-wildcard discipline: `LibraryPosition` dynamic slots (`ChangeZoneAll`, `PutAtLibraryPosition`, `Conjure`, `Counter`'s countered-spell destination), `ExileFromTopUntil`'s cumulative threshold, `OpponentGuess` proposition sides, `ChooseAndSacrificeRest`'s keeper constraint, and per-card `Conjure` counts. Deferred `AbilityDefinition`/modification payloads remain excluded per the documented convention (doc updated to name `RollDie::results` branch bodies explicitly).

Production-chain regressions (modeled on the entry-counter test — real-parser whole-hand discard producer, hand-built consumer continuation), each verified red with the visitor completion reverted:

- `tracked_mana_production_count_adds_discarded_count_mana` — red: pool gained 0 `{C}` instead of 2.
- `tracked_pay_cost_life_amount_pays_discarded_count_life` — red: payer stayed at 20 life instead of 18.
- `tracked_die_roll_modifier_adds_discarded_count_to_result` — red: d1 result stayed `Some(1)` instead of `Some(3)` (natural 1 + 2 discarded).

Empty-hand regression: added a reach guard asserting the captured events contain the discard resolution's `EffectResolved` for the source (plus a no-`Discarded`-events check), and removed the `CR 107.3a` citation — no Comprehensive Rules text file exists in the repo to verify a replacement against, so none was invented.

Verification: `cargo build -p engine` clean with the no-wildcard matches; full `cargo test -p engine` green (17549 lib + 3821 integration, 0 failures — no existing test depends on non-publication); `check-engine-authorities.sh` and `check-legacy-quantity-callsites.sh` pass; touched files formatted. Engine-only diff, so the expected parse-diff remains `no_changes`.

matthewevans:
Maintainer acknowledgement: rechecked current head bc5384afecc65249d0d98120c3c74b19791d9888 after the latest follow-up. The maintainer compatibility port is present; this remains held solely for the current required CI run, with no further author action requested now.

matthewevans (approved):
Maintainer review complete: tracked-set consumer classification now uses the exhaustive quantity visitor, including nested mana-production and cost carriers; the Malfegor runtime regression and current no-change parse artifact are clean.

PR #6193, opened 2026-07-19
fix(parser): recognize "you exile ... this way" reflexive triggers

### LINKED ISSUE #5989: adryn the usurper and abyssal harvest — the abilty to exile a card from a graveyard and make a copy only works if i do…
Reported in Discord: https://discord.com/channels/1485498006781427802/1524725753227247636/1524725753227247636

**Thread:** adryn the usurper and abyssal harvest
**Discord thread id:** `1524725753227247636`
**Discord message id:** `1524725753227247636`
**Cards:** _none detected_
**Parser status:** has_gaps

## Summary
the abilty to exile a card from a graveyard and make a copy only works if i do it to my graveyard other wise i have to make to copy myself and i have

## Reported text

the abilty to exile a card from a graveyard and make a copy only works if i do it to my graveyard other wise i have to make to copy myself and i have no was to add a creature type to the ones for adryn

## Attachments

- [game-state-turn-41-2026-07-09T10-22-09-313Z.zip](https://cdn.discordapp.com/attachments/1524725753227247636/1524725753436966943/game-state-turn-41-2026-07-09T10-22-09-313Z.zip?ex=6a595cc3&is=6a580b43&hm=aed54b30bf5b72f3e6970f81467b485a73b725c075b38f53500af57d30b67c21&)

---
<sub>report_id: `discord:1524725753227247636:1524725753227247636:0` · discord: `1524725753227247636/1524725753227247636`</sub>

tryeverything24:
Fixes #5989.

Ardyn, the Usurper: "At the beginning of combat on your turn, you may exile up to one target creature card from a graveyard. When you exile a card this way, create a token that's a copy of it, except it's a 5/5 black Demon."

## Root cause & fix

`strip_if_you_do_conditional` (the reflexive "when you <verb> this way" gate recognizer) already had active-voice combinators for "you discard"/"you sacrifice" this way, but no "you exile" arm. The entire "When you exile a card this way, create a token..." clause fell through to `Effect::Unimplemented` — the ability exiled the target and then silently did nothing, matching the reported "I have to make the copy myself" symptom.

Fixed by adding `parse_you_exile_this_way_clause` (`parser/oracle_nom/condition.rs`), mirroring the existing `parse_you_discard_this_way_clause` / `parse_you_sacrifice_this_way_clause` siblings (same active-voice grammar, differing only in the verb and that exile's destination isn't fixed to a single zone the way discard/sacrifice are), and wiring it into `strip_if_you_do_conditional` (`parser/oracle_effect/conditions.rs`) alongside the existing two.

Also investigated the issue's own paraphrase — "only works if I do it to my graveyard" — and disproved it: the parsed exile target filter carries no controller restriction (any player's graveyard is a legal source), so this reads as an imprecise description of the same underlying "nothing happens" bug, not a second, narrower defect.

## Testing

New end-to-end test drives the real activate → target-selection → resolve → optional-effect-decision → reflexive-copy pipeline (not a hand-built `ResolvedAbility`), with two legal graveyard targets — one in EACH player's graveyard — so target selection genuinely pauses and the opponent's-graveyard case is explicitly exercised. Confirms the resulting token is a distinct 5/5 black Demon.

Verified locally: fmt clean, clippy clean (`-D warnings`), full integration suite 3546 tests (3543 pre-existing + 2 pre-existing ignored + 1 new), 0 regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of Oracle text using the “you exile… this way” idiom.
  * Correctly links exile-conditions to reflexive “this way” clauses so follow-up effects fire as intended.
  * Ensures the resulting token copy behavior matches the exiled card’s expected characteristics and related abilities.
* **Tests**
  * Added regression coverage for Ardyn, the Usurper’s exile-and-copy interaction (including correct gating behavior).

### DIFF
--- a/crates/engine/src/parser/oracle_effect/conditions.rs
+++ b/crates/engine/src/parser/oracle_effect/conditions.rs
@@ -963,6 +963,21 @@ pub(super) fn strip_if_you_do_conditional(text: &str) -> (Option<AbilityConditio
                     text[offset..].to_string(),
                 );
             }
+            // CR 603.12 + CR 701.16a: "when you exile a[n] X this way, [body]" —
+            // the reflexive gate created by a preceding "exile [target] X"
+            // instruction (Ardyn, the Usurper, issue #5989). The exile's move
+            // into the (public, CR 400.2) exile zone publishes the card into
+            // `state.last_zone_changed_ids`, which `ZoneChangedThisWay` checks.
+            if let Ok((after_clause, (filter, _negated))) =
+                crate::parser::oracle_nom::condition::parse_you_exile_this_way_clause(rest)
+            {
+                let body_lower = strip_reflexive_conditional_body_separator(after_clause);
+                let offset = text.len() - body_lower.len();
+                return (
+                    Some(AbilityCondition::ZoneChangedThisWay { filter }),
+                    text[offset..].to_string(),
+                );
+            }
         }
     }
     (None, text.to_string())
--- a/crates/engine/src/parser/oracle_effect/conditions.rs
+++ b/crates/engine/src/parser/oracle_effect/conditions.rs
@@ -922,6 +922,24 @@ pub(super) fn strip_if_you_do_conditional(text: &str) -> (Option<AbilityConditio
                 text[offset..].to_string(),
             );
         }
+        // CR 603.12 + CR 701.16a: active-voice "you exile[d] a[n] X this way,
+        // [body]" — the reflexive gate created by a preceding "exile [target]
+        // X" instruction. Runs under BOTH prefixes because the printed idiom
+        // appears in both tenses: past "If you exiled a card this way, …"
+        // (Ardyn, the Usurper's actual Oracle text, issue #5989) and present
+        // "When you exile a card this way, …". The exile's move into the
+        // (public, CR 400.2) exile zone publishes the card into
+        // `state.last_zone_changed_ids`, which `ZoneChangedThisWay` checks.
+        if let Ok((after_clause, (filter, _negated))) =
+            nom_cond::parse_you_exile_this_way_clause(rest)
+        {
+            let body_lower = strip_reflexive_conditional_body_separator(after_clause);
+            let offset = text.len() - body_lower.len();
+            return (
+                Some(AbilityCondition::ZoneChangedThisWay { filter }),
+                text[offset..].to_string(),
+            );
+        }
         if let Ok((after_clause, (filter, _negated))) =
             nom_cond::parse_zone_changed_this_way_clause(rest)
         {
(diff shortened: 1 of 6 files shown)


matthewevans (changes requested):
### Blocking: the added grammar does not match Ardyn's current Oracle text

Scryfall lists Ardyn's reflexive clause as **“If you exiled a card this way, …”** ([card API](https://api.scryfall.com/cards/named?exact=Ardyn%2C%20the%20Usurper)). This PR instead adds a `when you exile … this way` parser and its integration test replaces the printed triggered ability with a synthetic activated ability containing that same `When` wording (and an extra `may`).

`strip_if_you_do_conditional` reaches the new arm only inside `if prefix == "when "`; the existing passive parser starts with an article/type phrase, so it cannot consume the actual active-past `if you exiled a card this way` form either. Consequently the real card still falls through instead of gaining the conditional token-copy effect, despite the green test.

Please parse the actual active-past `if you exiled … this way` construction (at the shared reflexive-condition seam, without regressing the existing passive forms), and replace/add a regression that uses Ardyn's exact current Oracle text and its triggered combat ability. The test should prove the printed path, not a hand-written activated approximation.

tryeverything24:
Addressed the blocker — the PR now parses and proves Ardyn's actual printed text.

**Grammar**: `parse_you_exile_this_way_clause` accepts both tenses ("you exile"/"you exiled") and runs under both the `if ` and `when ` prefixes at the shared reflexive-condition seam (hoisted alongside the other active arms), so the active-past "If you exiled a card this way," reaches it.

**A second, deeper mis-binding found while fixing this**: the context-free condition grammar was already consuming "you exiled a card this way" as `CostPaidObjectMatchesFilter` (the cost-paid class). With no cost snapshot, its runtime fallback requires the moved object's controller to equal the ability controller — so a card exiled from an opponent's graveyard failed the gate, matching the issue's "only works on my own graveyard" report. Added a clause-context rewrite (`rewrite_cost_paid_exiled_reflexive_for_effect_exile_parent`, mirroring the existing cant-rider rewrite): when the preceding non-continuation clause is a `ChangeZone { Exile }` EFFECT, the gate rebinds to `ZoneChangedThisWay`. The cost-paid class (Shilgengar — no preceding exile clause in the chain) and the subject-scoped sacrifice class (Deadly Brew) are untouched.

**Tests**: the integration test now drives Ardyn's full exact Oracle text (Demon anthem + `Starscourge —` ability word) through the real begin-combat trigger: phase advance into BeginCombat, `TriggerTargetSelection` between two graveyard candidates (one in EACH player's graveyard), mandatory exile (no optional prompt — the printed text has no "you may"), automatic reflexive copy; asserts the token is a copy of THAT card by name, a 5/5 Demon, and receives haste from Ardyn's own anthem line. A parser test pins the exact sentence pair lowering to a `ZoneChangedThisWay`-gated `CopyTokenOf` over the tracked set.

Verified locally in an isolated target dir: fmt clean, workspace clippy `-D warnings` clean, engine lib 17516 passed, engine integration 3749 passed. Rebased onto current `main`.

matthewevans (commented):
Re-reviewed the refreshed head: the exact-Ardyn regression and the clause-context rewrite address the prior blocker, and the current CodeRabbit review has no actionable finding. Holding rather than approving while the refreshed CI run and a parse-diff artifact tied to this head complete.

matthewevans (approved):
Approved on current head `cf6c2d2f39aa79e4e2c6c62bfb77ed67d158606b`: the exact-Ardyn regression, shared active-voice reflexive parser, and clause-context rebind resolve the prior finding. Current CI and parse-diff evidence are clean.

Issue #6172, opened 2026-07-18
Stuck decision: CastOffer


LukasMHD:
**What happened**
_briefly describe what you were doing_
cast fungus frolic on brightcap badger
**Diagnostic**
```
Build: v0.29.0 (88f0a7e)
Waiting for: CastOffer
Stuck players: 0
User agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:152.0) Gecko/20100101 Firefox/152.0
```

mike-theDude:
## Classifier analysis

**Verdict**: `not_card_data_attributable`

**Matched clause**: `(none)`  (supported: n/a)

**Reasoning**: The report reaches CastOffer and leaves no player able to act after Fungus Frolic targets Brightcap Badger. That is a cast-offer continuation/action-routing failure, not an AST-faithfulness defect in the named cards.

---
_Posted by bug-coverage-classifier skill._

mike-theDude:
Could you add which player controlled Fungus Frolic and Brightcap Badger, what CastOffer prompt/options appeared, what you selected, and attach the saved game state or log? Those details are needed to distinguish a human action-routing stall from the older fixed AI CastOffer fallback in #4433.

mike-theDude:
Backlog audit (2026-07-25): this was reported on v0.29.0 at the Adventure CastOffer for Brightcap Badger // Fungus Frolic. Current main includes merged PR #6552, which now renders only engine-authorized Adventure faces and dispatches the exact legal ChooseAdventureFace action. Its regressions cover single-face, both-face, and player-routing cases, while engine tests cover the only-Adventure-face-affordable path.

Please close this as fixed/stale. If it still occurs on a current build, reopen with the saved game state and which face or action was offered.

matthewevans:
Closing per mike-theDude's triage above — verified fixed/resolved on current main. If this still reproduces on a current build, please reopen with a fresh repro.

Issue #6080, opened 2026-07-17
[Card Bug] Native AI stalls on OptionalCostChoice (kicker / alternative cost) — ai-commander never reaches GameOver


mcbradd:
### Card(s)
Divine Resilience (kicker), Fierce Guardianship (free-cast alternative cost) — same underlying stall, listed as examples per template guidance.

### Build/version
commit 3da779098f6a87fc5ddc81381c28aa797fff3e68, `cargo build --release -p phase-ai`, Windows 11 x86_64-msvc, nightly-2026-04-19

### Game mode
AI

### Actual behavior
In headless `ai-commander` 4-player games, when a native-AI player casts a spell carrying an optional cost choice, the game stalls: `run_ai_actions` returns no actions while `game_state.waiting_for` is `OptionalCostChoice { player, cost: Choice(Mana { cost: NoCost }, Mana { cost: ... }), ... }`, so the runner's loop breaks and the game ends without reaching `GameOver`.

Observed with two distinct optional-cost shapes:
- Divine Resilience (Kicker {2}{W}): stalls on `Choice(NoCost, {5}{W})` at turn 15 (seed 1009, Easy)
- Fierce Guardianship ("you may cast without paying" while controlling a commander): stalls on `Choice(NoCost, {5}{U})` at turn 23 (seed 1009, VeryHard)

Possibly related to #5250 (kicker prompt on Everflowing Chalice) — if that one reproduces in AI games, this seed may serve as its repro too.

### Expected behavior
The AI decides the optional cost (decline is always a legal default) and the game continues to `GameOver`.

### Steps to reproduce
```
cargo run --release --bin ai-commander -- client/public --seed 1009 --difficulty Easy
```
Deterministic: stalls every run at turn 15, 456 actions, with the `waiting_for = OptionalCostChoice` state printed at exit. `--difficulty VeryHard` with the same seed reproduces the Fierce Guardianship variant at turn 23.

### Logs/screenshots/game-state
Tail of the Easy run (abridged):
```
Turn 15 (active P1)  actions=   456  P0:36 P1:38 P2:37 P3:38
=== RESULT ===
Game did NOT reach GameOver. waiting_for = OptionalCostChoice { player: PlayerId(1),
  cost: Choice(Mana { cost: NoCost }, Mana { cost: Cost { shards: [White], generic: 5 } }),
  times_kicked: 0, pending_cast: PendingCast { ...
  description: Some("Creatures you control gain indestructible until end of turn.") ... } }
```
[...]

Issue #5958, opened 2026-07-16
Treasure cruise froze game — The Ai cast it and since the it has been frozen.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1525245591289073934/1525245591289073934

**Thread:** Treasure cruise froze game
**Discord thread id:** `1525245591289073934`
**Discord message id:** `1525245591289073934`
**Cards:** _none detected_
**Parser status:** no_card
**Extraction confidence:** 0.60

## Summary
The Ai cast it and since the it has been frozen.

## Reported text

The Ai cast it and since the it has been frozen.

## Attachments

- [Screenshot_20260710_155610_Chrome.jpg](https://cdn.discordapp.com/attachments/1525245591289073934/1525245591649652978/Screenshot_20260710_155610_Chrome.jpg?ex=6a5946a6&is=6a57f526&hm=0cc4d8e7cad6a3417c75e2ee587f4d81ed08f99bc24b94e81aaa00bb5c11a75c&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1525245591289073934:1525245591289073934:0` · discord: `1525245591289073934/1525245591289073934`</sub>

mike-theDude:
## Classifier analysis

**Verdict**: `not_card_data_attributable`

**Matched clause**: `Delve` / `Draw three cards.` (supported: true)

**Reasoning**: Treasure Cruise card data has Delve and the draw-three spell effect. The report is that the AI cast it and the game froze, so this is routed to AI/casting-progress handling. It needs the stalled `waiting_for` state or game-state file to determine whether Delve payment, AI choice, or stack resolution is blocked.

---
_Posted by bug-coverage-classifier skill._

Issue #5250, opened 2026-07-07
Everflowing Chalice — When attempting to cast, I am prompted to kick it or stop kicking it.


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1522470587467829260/1522470587467829260

**Thread:** Everflowing Chalice
**Discord thread id:** `1522470587467829260`
**Discord message id:** `1522470587467829260`
**Cards:** _none detected_
**Parser status:** fully_parsed
**Extraction confidence:** 0.90

## Summary
When attempting to cast, I am prompted to kick it or stop kicking it.

## Reported text

When attempting to cast, I am prompted to kick it or stop kicking it. I had 5 available mana, and intended to kick twice for 2 counters. However it continued to prompt me to kick it or stop. I kept clicking to kick it until it was at 6. Then when I clicked stop, it entered with 6 counters - but only tapped 2 mana.

## Attachments

- [game-state-turn-12-2026-07-03T05-12-28-149Z.zip](https://cdn.discordapp.com/attachments/1522470587467829260/1522470587581202522/game-state-turn-12-2026-07-03T05-12-28-149Z.zip?ex=6a4ddcba&is=6a4c8b3a&hm=c13626abc51f5a20fb87583778711e53bd5f6b924cf2cc8142d2ecf1ded6b78e&)

<sub>Restored from the Discord source message on 2026-07-26: the original ingestion truncated this report. See `splitIntoItems` in `scripts/lib/extract.ts`.</sub>

## Oracle text (verified from `client/public/card-data.json`)
_No card names were detected for this report._

---
<sub>report_id: `discord:1522470587467829260:1522470587467829260:0` · discord: `1522470587467829260/1522470587467829260`</sub>

mike-theDude:
**Triage:** Couldn't confirm a defect from the report alone. Multikicker casting is a well-covered area — there's a dedicated `OptionalCostChoice` modal, `times_kicked` tracking, and an existing regression suite guarding a previously-fixed prompt-laundering bug (#454) — and the "pay again / stop kicking" prompt sequence is the **intended** per-kick UX (CR 702.33), so it isn't obviously broken.

Need repro detail to distinguish normal flow from a regression: does the prompt eventually let you finish the cast (declining should end it), or does it truly loop forever / block the cast? Please attach the exact click sequence or a short screen recording.

_Status set to `needs-repro`; classified `cannot-determine`._

matthewevans:
[...]

PR #6194, opened 2026-07-19
fix(ai): surface ai-commander stalls + strict difficulty parse + post-resolution deck counts

mcbradd:
## Summary

Three fixes to `ai-commander`, the AI-only 4-player Commander sanity-check runner pod-lab's gauntlet harness drives at scale:

- **Surface driver stalls instead of exiting as normal game end** (`0ec55a611`). An empty `run_ai_actions` batch while `waiting_for != GameOver` was previously conflated with a clean game end at the driver, hiding the whole p0-softlock issue family (#5250, #4345, #5958, #6172, #3886, #3919, #3233) behind a silent success exit. `run_ai_actions` now returns `AiActionsRun` (Deref'd to the existing `Vec<AiActionResult>` for source compatibility) carrying a typed `AiActionsBreakReason` naming which of its three break doors fired (no AI actor, `choose_action` returned `None`, or `apply()` returned `Err`). `ai_commander` prints a machine-readable `STALL:` line (`waiting_for` variant, turn, active/priority player, pending-cast summary) plus the break reason, and exits with a distinct code (3) instead of falling through to a clean exit — converting the whole silent-softlock family into loud, diagnosable failures. The existing `"Game did NOT reach GameOver"` substring pod-lab's `runner.py` already matches on is preserved verbatim. Addresses #6080 (driver-side surfacing; card-level fallback tracked separately).
- **Hard-fail unknown `--difficulty` labels instead of silently downgrading** (`ebe4f7bb5`). `parse_difficulty` previously matched five exact-case strings and silently fell back to `Easy` for anything else (typos, or `CEDH`, which had no arm at all) — poisoning an entire batch of games with the wrong skill tier with no indication anything went wrong. It now routes through the crate's single-authority `AiDifficulty::from_label`, but layers a hard startup error on top of any label `from_label` would otherwise silently downgrade, and lists every accepted value (`VeryEasy, Easy, Medium, Hard, VeryHard, CEDH`) in the error message.
- **Print per-seat resolved deck counts after name resolution** (`62a84e2b2`). `resolve_deck_list` silently skips any card name the database doesn't recognize, and the only prior print happened before that resolution ran, so nothing could prove a deck actually loaded full-size. Adds a per-seat "Resolved deck sizes (post name-resolution)" block summing each seat's resolved `main_deck`/`commander` counts, so a harness can detect silent-skip drift without reaching into engine internals.

## Test plan
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --workspace --all-targets --features engine/proptest -- -D warnings` clean (pre-push hook)
- [x] `cargo test -p engine --features proptest parser::oracle::tests` clean (pre-push hook)
- [x] `cargo test -p phase-ai --lib` clean (pre-push hook)
- [x] `cargo check --release --bin card-data-validate` clean (pre-push hook)
- [x] parser combinator gate (`scripts/check-parser-combinators.sh`) clean (pre-push hook)
- [x] coverage regression check clean (pre-push hook)
- [x] `pnpm lint` / `pnpm run type-check` clean (pre-push hook)


matthewevans (changes requested):
**Changes requested — the reported stall reason can be lost after real AI actions.**

## 🔴 Blocker

`crates/phase-ai/src/bin/ai_commander.rs:216-219` stores `results.break_reason` only when the returned action list is empty. But `run_ai_actions` can execute one or more actions and then stop on `ApplyFailed` or `ChooseActionNone` (`crates/phase-ai/src/auto_play.rs:156-183`). In that case this loop discards the diagnostic, invokes another batch, and may later print `NoActor` or `unknown` instead of the actual failed action/reason. That defeats the PR's stated purpose: making the original stall diagnosable.

Capture a non-`None` break reason from every batch and stop the driver at that batch boundary (after retaining its completed actions), then add a discriminating regression that exercises a non-empty batch followed by a failing apply/choice and asserts the emitted reason.

## 🟡 Non-blocking

The public `run_ai_actions` return-type change is currently untested across its diagnostic paths. The required regression above should also demonstrate the intended no-regression behavior for ordinary callers that consume action results.

**Recommendation: request changes; please preserve the original break door before retrying the AI loop.**

mcbradd:
Fixed the review blocker: `break_reason` is now captured from **every** `run_ai_actions` batch, not only empty ones.

Previously the driver only checked `results.break_reason` inside the `results.is_empty()` branch, so a batch that completed one or more actions and then stopped on `ApplyFailed`/`ChooseActionNone` had its diagnostic silently discarded — the driver looped again and could later report `NoActor`/unknown on an unrelated empty batch instead of the real cause.

Fix: extracted `driver_step` (`crates/phase-ai/src/auto_play.rs`) as the single place that decides, from a batch's action count and `break_reason`, whether the driver should stop. `ai_commander`'s loop now calls it after retaining/logging that batch's completed actions, and stops at that batch boundary with the original break reason preserved for the stall report.

Regression coverage added:
- `driver_step_preserves_break_reason_from_non_empty_batch`, `driver_step_empty_batch_behavior_is_unchanged`, `driver_step_ordinary_batch_is_unaffected` (unit tests in `auto_play.rs`) — cover the discriminating cases directly.
- `run_ai_actions_non_empty_batch_carries_break_reason` (`crates/phase-ai/tests/ai_quality.rs`) — forces `run_ai_actions` itself to return a real non-empty batch with a `break_reason` (DeclareBlockers applies, then priority reaches an AI seat with no registered config) and asserts `driver_step` handles it correctly, demonstrating ordinary callers are unchanged.

`cargo fmt --all`, `cargo clippy -p phase-ai --all-targets -- -D warnings`, and `cargo test -p phase-ai` (1354 unit + all integration tests) all pass.

The stacked branch (#6195) has been rebased on top of this fix — both the abort-path STALL-block guard (`if !aborted`) and the new break-reason capture are preserved together.

matthewevans (changes requested):
The non-empty-batch regression is now covered, but the diagnostic is still inaccurate for the case the new test exercises.

`run_ai_actions` finds P1 as an AI-controlled actor, then takes the `ai_configs.get(&actor) == None` branch, but records `AiActionsBreakReason::NoActor`. The commander prints that value as the stall cause, so a missing AI configuration is reported as "no actor" even though those require different remediation.

Please add a distinct missing-config reason (including the player) or otherwise preserve that distinction, and update the regression expectation. The completed CI is green; this is a correctness issue in the newly added diagnostic contract.

mcbradd:
You're right — that door reported a caller wiring gap as "no actor", and those need different remediation. Fixed.

## The fix

`run_ai_actions` takes the `ai_configs.get(&actor) == None` branch only *after* it has already found an actor and confirmed the seat is in `ai_players`. That door now records its own reason instead of borrowing `NoActor`:

```rust
    /// `player` is in `ai_players` but has no entry in `ai_configs`. Distinct
    /// from `NoActor`: an actor *was* found and *is* AI-controlled, so the
    /// remedy is caller wiring (register a config for this seat), not "wait
    /// for a human" or "the game ended".
    MissingAiConfig { player: PlayerId },
```

```rust
            None => {
                tracing::warn!(player = ?actor, "AI seat has no config — stopping AI loop");
                break_reason = Some(AiActionsBreakReason::MissingAiConfig { player: actor });
                break;
            }
```

And the regression expectation in `run_ai_actions_non_empty_batch_carries_break_reason` now pins that specific door:

```rust
    assert!(
        matches!(
            results.break_reason,
            Some(AiActionsBreakReason::MissingAiConfig { player: P1 })
        ),
        "expected MissingAiConfig(P1): P1 is an AI seat with no ai_configs entry, \
         which is not the same stall as NoActor"
    );
```

## Why `NoActor` stays fieldless

Deliberate, and I want to be explicit about the reasoning rather than have it read as an oversight. `NoActor` fires when the `find()` over `acting_players()` yields nothing, and that has two shapes — one of which has no player to name at all:

- `crates/engine/src/types/game_state.rs:7341` — `WaitingFor::GameOver { .. } => None` in `acting_player()`, which `acting_players()` funnels through at `:7361` (`_ => self.acting_player().into_iter().collect()`). `GameOver`, and an empty pending set, both produce an **empty** `Vec<PlayerId>`. There is no player to attach.
- `crates/engine/src/types/game_state.rs:7355-7360` — `MulliganDecision` and `OpeningHandBottomCards` both `pending.iter().map(|e| e.player).collect()`, so they can pend **several** players simultaneously. If the batch stops because none of them is an AI seat, picking one of the several to name would be arbitrary.

Attaching a `PlayerId` to `NoActor` would therefore either be a lie (empty case) or an arbitrary pick (simultaneous case) — reproducing the same imprecision you filed, one layer down. The distinction you asked for is carried by the new variant instead.

## Disclosures

- `NoActor` still covers both the empty-`acting_players()` case and the actor-not-in-`ai_players` case; only the missing-configuration cause you named is split out here, and separating that second conflation is filed as follow-up rather than folded into the fix you asked for.
- The regression expectation asserts via `matches!` on the variant's `player` field, so no new derives were needed on the reason enum.
- The STALL printer already formats the reason with `{:?}`, so the new variant surfaces with no commander change; a hand-written `Display` was not added.
- The `player: P1` in the updated expectation is what pins the missing-config door specifically — `NoActor` carries no field and cannot match it — so no second test was added.
- `EngineError` is four variants with a `String` at worst, so the comment was corrected rather than the type boxed; clippy's `large_enum_variant` does not fire on it.

## Bot notes

- **Deck-count indexing** (`ai_commander.rs:165-177`): addressed. `[&payload.player, &payload.opponent].into_iter().chain(payload.ai_decks.iter()).enumerate()` replaces the hardcoded `ai_decks[0]`/`[1]`. `ai_decks` is built with exactly two entries, and both forms visit player → opponent → `ai_decks[0]` → `ai_decks[1]` in that order, so output is byte-identical for the 4-seat case while no longer being able to panic if the seat count changes.
- **`EngineError` boxing**: declined, with the comment corrected instead — see the disclosure above. The comment was the inaccurate part, not the type.

## Proof

Revert the `None =>` arm to `AiActionsBreakReason::NoActor` and `run_ai_actions_non_empty_batch_carries_break_reason` fails on the `matches!` assertion with `expected MissingAiConfig(P1)` — `NoActor` is fieldless, so it cannot satisfy the `player: P1` pattern. (`P1` is `pub const P1: PlayerId = PlayerId(1)` from `engine::game::scenario`, and `PlayerId` derives `PartialEq + Eq`, so that is a real value check and not a catch-all binding.)

## Verification

`cargo fmt --all`, `cargo clippy -p phase-ai --all-targets -- -D warnings` (clean, zero warnings), and `cargo test -p phase-ai` — **1402 passed, 0 failed, 14 ignored**. Pre-push hook (full workspace clippy + card-data pipeline + coverage regression + frontend lint/type-check) run, not bypassed.

matthewevans (commented):
**Re-reviewed after the maintainer branch update.**

The current head contains a clean merge of `main`; the implementation diff remains the same three reviewed `phase-ai` files. The missing-configuration diagnostic is now represented as `MissingAiConfig { player }`, and the non-empty-batch regression exercises the production `run_ai_actions` path. No additional implementation finding from this follow-up.

Required checks restarted on the updated head, so approval/enqueue remains pending their completion.

matthewevans (approved):
Approved after current-head re-review and the completed post-update checks: the AI stall diagnostic preserves non-empty-batch break reasons, distinguishes missing configuration from no actor, and has production-path regression coverage.

PR #6212, opened 2026-07-19
fix(parser): type the condition on bare self-referential combat-requirement statics

JacobWoodson:
## The bug

Frodo Baggins' `As long as Frodo Baggins is your Ring-bearer, it must be blocked if able.` parsed to **two** static abilities, both gated on `StaticCondition::Unrecognized { text: "~ is your Ring-bearer, it" }`.

`layers.rs` evaluates `Unrecognized` as **always-true**, so the must-be-blocked requirement applied unconditionally — Frodo had to be blocked even when he was **not** your Ring-bearer.

## Root cause

`try_split_inverted_as_long_as` already produced a clean split at the `shared.rs:2040` dispatch block (`condition = "~ is your Ring-bearer"`, `effect = "it must be blocked if able"`), and `parse_static_condition` types that condition to the **already-existing** `StaticCondition::IsRingBearer` directly.

But no consumer existed in that block for the bare self-referential combat-requirement shape. All three existing consumers declined, so the line fell through to `try_split_and_must_attack_block`, which cut the requirement out mid-clause and left an orphaned `"it"` — producing the ungrammatical `"As long as ~ is your Ring-bearer, it."` that forced the condition into `Unrecognized`.

## The fix

**Fix A** — add `try_parse_inverted_bare_self_rule_static` as a fourth consumer of the existing `InvertedSplit`, composing `parse_static_condition` (→ `parse_inner_condition`, the single authority for game-state conditions) with `parse_rule_static_predicate_nom` under `all_consuming`. No new enum variant; `IsRingBearer` already existed and was already wired.

Generic across two axes — any typeable condition × any `RuleStaticPredicate`. Fixes **Frodo Baggins** and **Enkira, Hostile Scavenger**, which carry two *different* `StaticCondition` variants (`IsRingBearer`, `SourceIsEquipped`) through identical grammar.

**The attached-subject gate.** For `As long as enchanted creature is red, it loses all abilities.` (**Ray of Frost**, paper) the pronoun binds the **host** permanent, not the source (CR 608.2c). That line is correctly claimed by a consumer *downstream* of this insertion point, so an ungated branch would win the race and strip the **Aura's own** abilities. `condition_binds_attached_subject` scans at word boundaries rather than peeking a prefix, because **Dog Umbra**'s attached noun is mid-string. Fails closed.

**Fix B** — restore single-authority condition delegation on the terminal `"as long as"` fallback in `dispatch.rs`, aligning it with its four in-file siblings. Narrowing-only (`parse_static_condition` enforces full consumption); **zero** corpus change, so this is defensive hygiene rather than a behavior change.

## Verification

Runtime gating is verified through the real production entry `combat::validate_blockers_for_player` — not only at the AST layer, since the bug was a runtime always-true evaluation.

- `cargo test -p engine` — 17245 + 8 + 9 + 3562 passed, 0 failed
- `cargo clippy --all-targets -- -D warnings` — exit 0
- `check-parser-combinators.sh` — Gate G / Gate A PASS
- `cargo semantic-audit` — flagged-card set identical before and after
- Coverage 31494 → **31496** (88.768% → 88.774%)
- Full-corpus export diff: **42 lines, Frodo and Enkira only**

Revert-failure was proven empirically (each fix temporarily neutered behind a probe) rather than argued. Tests that do **not** discriminate are labelled as reach-guards rather than presented as proofs. The runtime blocker fixture is deliberately 1/4: with a 2/2, CR 701.54c (the Ring emblem's "can't be blocked by creatures with greater power") makes blocking a Ring-bearer Frodo illegal, which would have made every positive assertion pass vacuously.

## Rules

CR 509.1c, CR 604.1, CR 608.2c, CR 611.3a, CR 701.54a, CR 701.54c, CR 701.54e — each verified against `docs/MagicCompRules.txt`.

## Known gap (not addressed here)

Lines with a **non-self** effect subject (`As long as <cond>, creatures you control attack each combat if able.`) still ride the pre-existing fallback and still emit `Unrecognized` — the same class of bug, different subject shape. This branch deliberately claims only self-referential subjects; fixing the rest requires deriving `affected` from the subject and untangling the suspected latent `Source*` collapse documented at `condition.rs:1705-1712`. A test asserts the current behavior so it fails loudly when that fallback is repaired.


matthewevans (approved):
Approved after current-head review: the parse diff is limited to the intended typed conditions for Frodo and the Enkira sibling class, and the combat runtime tests exercise the live Ring-bearer gate.

PR #6214, opened 2026-07-19
fix(client): render a name/text tile for tokens with no official art (#6156)

### LINKED ISSUE #6156: Kibo Banana tokens render as black squares because no official token art exists
## Summary

Banana tokens created by Kibo, Uktabi Prince currently render as black squares instead of a usable token/card visual.

This appears to be an art/data fallback problem, not a rules issue. Kibo's exported card data correctly creates a colorless artifact token named `Banana` with the expected activated mana/life ability, but `Banana` is not a normal `card-data.json` key and likely cannot be resolved through the generated Scryfall token-image data because Wizards does not appear to have produced an official paper Banana token/card image.

## Evidence

- Local exported card data for `kibo, uktabi prince` has the correct Oracle text and a `Token` effect named `Banana`.
- `jq -r 'keys[] | select(test("banana"; "i"))' client/public/card-data.json` only finds ordinary cards containing banana in the name, not a standalone `Banana` token entry.
- The frontend token image path expects either `obj.token_image_ref` via `client/public/scryfall-token-images.json` or a token/name lookup fallback.
- Public token references list Banana as a token without paper printing and note "No official art."

## Expected behavior

When Kibo creates Banana tokens, the UI should show a deliberate fallback representation instead of a black square. Possible resolutions to discuss:

- A generic artifact-token fallback frame/text tile for true tokens with no official art.
- A curated local fallback asset for non-paper tokens whose art cannot be sourced from Scryfall/MTGJSON.
- A generated or symbolic Banana token image, if that is acceptable for the project's asset policy.
- A specific engine/card-data token-image reference strategy for no-official-printing tokens.

## Repro

1. Start a game with Kibo, Uktabi Prince.
2. Activate Kibo's `{T}` ability.
3. Observe the created `Banana` tokens on the battlefield.

## Notes

This should be discussed before implementation because the right fix may affect the general policy for unofficial/no-paper tokens, not just Kibo.

minion1227:
Closes #6156

## Summary

Banana tokens created by Kibo, Uktabi Prince — and any other true token with no official paper printing — render as a featureless dark square with no way to tell what the permanent is. This implements the issue's first-listed resolution ("a generic artifact-token fallback frame/text tile"): a deliberate name (+ Oracle text) tile, so the permanent stays identifiable.

## Root cause

`useCardImage` returns `src: null` for a token whose art can't be resolved (no `token_image_ref` in `scryfall-token-images.json` and no Scryfall name hit — true for Banana, which Wizards never printed on paper). `CardImage` folded that `!src` case into its **loading** branch:

```tsx
if (!faceDown && (isLoading || !src)) {  // gray-700 animate-pulse
```

So a token that will *never* resolve art animated the loading pulse forever — a blank/black square. The component already had a rich name + Oracle-text tile, but only for the separate `imageError` (broken ``) case.

## Fix

- Split "still resolving art" (`isLoading`) from "resolution finished with no image" (`!src`). Only the former pulses.
- Route `!src` **and** `imageError` through one shared `CardArtFallback` tile (extracted; previously the error tile was inline). It renders the card/token name and, when the engine knows it, the Oracle text.

Built for the class: any artless card or token is covered, not a hard-coded name. Face-down cards (`CARD_BACK_URL`) and the normal loaded-art path are untouched.

- `client/src/components/card/CardImage.tsx` — the pre-existing `imageError` fallback tile (name + `RichLabel` Oracle text) is the exact visual `CardArtFallback` now generalizes; this change widens its trigger from "broken ``" to "broken `` OR no resolvable art".
- `client/src/components/card/__tests__/CardPreview.test.tsx` — the sibling component test's `vi.mock("../../../hooks/useCardImage.ts", …)` + React Testing Library `render`/`screen` pattern is the harness reused verbatim in the new `CardImage.test.tsx`.

## Scope / policy note

The issue flags that the broader **asset-sourcing** policy for no-paper tokens (curated local art, generated art, an engine token-image-ref strategy) should be discussed. This PR deliberately stays policy-neutral: it only replaces the blank square with an identifiable text tile using data the client already has, and leaves any art-sourcing decision as separate follow-up.

## Test plan

New `client/src/components/card/__tests__/CardImage.test.tsx` (RTL + Vitest, mocking `useCardImage`):

- `pnpm exec vitest run src/components/card/__tests__/CardImage.test.tsx`

Cases: (1) `isLoading` → pulse, no text tile; (2) resolved with `src: null` → name tile, and **no `` element is emitted** (nothing can render as a black square); (3) Oracle text appears in the tile when known; (4) a resolved `` that fails to load swaps in the same tile.

## AI-contributor disclosure

Authored with an LLM (Claude). Pure frontend/display change — no engine, parser, or MTG-rules logic touched, so no CR annotations apply. Combinator-purity Gate A passes trivially (no parser diff). Local `pnpm`/Vitest could not be run in the authoring sandbox (network-restricted dependency install); CI runs the frontend suite. Please flag anything you'd like reworked.


matthewevans:

This current head (`232f5ffcd7dc505c6bb8dde24110add983b01bba`) was triaged as a frontend-only change (`client/src/components/card/CardImage.tsx`, `client/src/components/card/__tests__/CardImage.test.tsx`) by `minion1227`. The local frontend-review allowlist does not include this author, so this route does not perform an implementation-diff review or approve the PR.

minion1227:
Understood on the `defer-fe` routing — flagging for whenever a maintainer has capacity, no rush.

State of this one:

- **CI green** across all checks, including `Frontend (lint, type-check, test)`.
- **Automated review came back clean** ("no additional feedback"), so there are no outstanding comments to address.
- Scope is deliberately narrow: it only splits "art still resolving" from "resolution finished with no art" and routes the latter through the tile the component already used for broken images. Face-down cards and the normal loaded-art path are untouched.
- It stays **policy-neutral** on the open question in #6156 — no curated/generated assets, no token-image-ref strategy. Just an identifiable tile instead of a blank square, leaving the asset-sourcing decision entirely open.

Happy to split it smaller, fold it into a larger frontend batch, or drop it if the artless-token fallback isn't a direction you want. If a frontend-review exception isn't something you want to hand out, that's completely fine — I'd just appreciate knowing so I can aim future work at engine-side issues instead.

matthewevans:

This refresh applies to the current head (`111ac0726681f55eb0e4e30912b224d353d90c71`). It remains a frontend-only change, and `minion1227` is not in the local frontend-review allowlist. It therefore remains routed for explicit maintainer ownership rather than receiving an implementation-diff review or approval here. The `defer-fe` label is a routing marker, not a verdict on the change.

matthewevans (changes requested):
**[MED]** The text-box preview never enters its new 404 fallback. Evidence: `client/src/components/modal/CardTextboxPreview.tsx:36-75` creates `artError` and branches on it, but the rendered `` has no `onError` handler to call `setArtError(true)`. Why it matters: a failed image request still leaves a blank/broken preview in the choice modals this PR explicitly intends to keep identifiable. Suggested fix: attach `onError={() => setArtError(true)}` and add a component test that fires that error and asserts the named fallback.

minion1227:
Thanks — you were exactly right, and it was my bug. I added the `artError` state and the fallback branch in the previous commit but never attached `onError`, so nothing could set it and the branch was unreachable dead code.

Fixed in `8358d773a`:

- `onError={() => setArtError(true)}` on the `` in `CardTextboxPreview`.
- New `client/src/components/modal/__tests__/CardTextboxPreview.test.tsx` with the test you asked for: fires the error, asserts the named fallback replaces the ``. Verified discriminating — removing the handler fails that case while the other two still pass. The file also pins the two previously-uncovered states (absent while in flight, named strip once settled with no art).

Verified locally on that head: 2078 tests pass, `tsc -b --noEmit` clean, `eslint --quiet` clean. Gate A PASS head=`8358d773adbc192ec3ca9c4511897ecbc2ad60ca`.

Two notes on the rest of the PR while you have it open:

**Known remaining, deliberately out of scope.** A review sweep enumerated ~31 card-art render sites in `client/src`; this PR now handles both `src === null` and load failure at 8 of them (the in-game paths that actually reproduce #6156: `CardImage`, `ArtCropCard`, desktop hover, both mobile preview arms, `StackEntry`, `CardTextboxPreview`). The remaining ~26 — zone piles, reveal overlays, deck builder, draft, menus — still lack `onError`, and several render an anonymous blank box rather than a name. `CardArtFallback` now exists to make each a one-line swap, but sweeping them is a separate change rather than a #6156 fix. Happy to file a tracking issue, or fold them in here if you would rather have it in one pass.

**Unrelated CI red.** `Card data (generate, validate, coverage)` fails on the `Draw replacement corpus freeze (Plan 03 / CR 121.2)` step. That is inherited, not from this PR: `main` at `e259c4c6` ("Add Birgi, God of Storytelling // Harnfel, Horn of Bounty") fails the same job and step, and `e259c4c6` entered this branch as the second parent of a merge commit. That job's cache key hashes only MTGJSON plus `crates/engine/**`, so a frontend-only diff cannot affect it. It should clear here once main is green.

matthewevans (approved):
Approved after the current head fixed the missing image-error transition in `CardTextboxPreview` and added a component-level regression test that fires the image error and verifies the named fallback.

PR #6217, opened 2026-07-20
fix(engine): stamp real occurrence refs on self-owned trigger installs

JacobWoodson:
## The crash

A token created from a catalog preset carried its rules-text trigger as a bare `TriggerDefinition`. `Definitions::push<U: Into<T>>` routes that through `From<TriggerDefinition> for TriggerEntry`, which stamps `TriggerDefinitionOccurrenceRef::Unmaterialized` — a variant marked `#[serde(skip_serializing)]` and explicitly rejected from an observable state by `validate_trigger_definitions`.

The engine computed the right result and then died handing it to the UI. `engine-wasm`'s `to_js` panics on serialization failure, so the first state send after such a token was created killed the worker:

```
panicked at crates/engine-wasm/src/lib.rs:75:29:
serde_json serialization failed: the enum variant
TriggerDefinitionOccurrenceRef::Unmaterialized cannot be serialized
```

## User-visible symptom

Reported as "Mitotic Slime's tokens don't work." Sacrificing the Slime put its dies-trigger on the stack correctly, but resolving it created Ooze tokens whose live trigger entries could not serialize — so the worker died before the tokens reached the client. The trigger stayed on the stack forever, and reloading restored the same pre-resolution checkpoint and panicked again ("Engine connection lost", immediately recurring).

Debug-spawning the same Ooze tokens worked, which made this look like a token-ability parser gap. It isn't: debug spawn marks layers dirty, so a layer pass rebuilds live entries from base with proper `Printed` refs before anything serializes. Tokens created mid-resolution get serialized inside that window.

**No engine test caught this because in-process tests never serialize.**

## The fix

Adds `GameObject::push_printed_trigger` / `relive_printed_trigger` as the single authority for *"this object gains a trigger that is part of what it is"* (CR 111.3 token abilities, CR 707.9a copiable values, synthesized printed riders). It keeps `base_trigger_definitions` and the live list in lockstep and stamps a real `Printed` occurrence ref, so callers never do index math or push bare definitions.

Grants from a *separate* source are unchanged — those still go through `install_trigger_candidate` and its `Granted`/`ExpandedGrant` producer key.

Converted, all self-owned installs:

| Site | |
|---|---|
| `effects/token.rs` | catalog token materialization ← **the reported crash** |
| `effects/token_copy.rs` | ×2 |
| `effects/copy_spell.rs` | |
| `game/blitz.rs` | dies-draw trigger |
| `database/synthesis.rs` | evoke + offspring |

## Compiler gate

`From<TriggerDefinition> for TriggerEntry` is now `#[cfg(any(test, feature = "test-support"))]`, matching the sibling `Definitions` conversions in `definitions.rs`. A future raw push into a live list is a **compile error** instead of a runtime engine-kill. The deserializer's legitimate legacy-payload path calls an explicit `unmaterialized_legacy` constructor.

The gate is also how the site list above was proven complete — the compiler enumerated all 9 rather than me asserting I had found them.

## Stickers: deliberately not converted

`game/stickers.rs` is the 9th site and is left on the explicit legacy constructor with a `FIXME`. A sticker-granted trigger is **not** a printed slot: `rebuild_public_zone_stickers` reverts to base and re-applies, so writing to base would make the sticker permanent. It wants a Layer-6 grant producer key, but neither `TriggerProducerOrigin::Static` (no owning static definition) nor `Transient` (no continuous-effect id) describes a sticker without a design decision.

**A stickered object with an ability sticker still fails to serialize.** That is pre-existing, and this change makes it explicit and greppable rather than silent. Worth filing separately rather than guessing at CR 123 semantics.

## Evidence

- The **token** fix is backed by a reproducer from a real crashed session (state dump + console trace) and a red/green regression test.
- The other **7** are pattern-matched to the same defect class. They compile and the suite is green, but there is no gameplay reproducer for blitz/copy/evoke the way there was for the token path. Worth a reviewer's eye on the occurrence semantics at each.

## Verification

- `cargo test -p engine --lib` — 17,260 passed, 0 failed
- `cargo test -p engine --test integration` — 3,574 passed, 0 failed
- `cargo fmt --all`
- New test verified **red** without the fix (`got Unmaterialized`)

The regression test asserts the occurrence ref, runs `validate_trigger_definitions`, **and serializes both the object and the full state** — the check whose absence let this ship.

### DIFF
--- a/crates/engine/src/database/synthesis.rs
+++ b/crates/engine/src/database/synthesis.rs
@@ -3127,13 +3127,11 @@ pub(crate) fn ensure_evoke_etb_sac_trigger(obj: &mut crate::game::game_object::G
             .iter_all()
             .any(|entry| is_evoke_sac(entry.definition()))
         {
-            obj.trigger_definitions.push(build_evoke_etb_sac_trigger());
+            obj.relive_printed_trigger(is_evoke_sac);
         }
         return;
     }
-    let trigger = build_evoke_etb_sac_trigger();
-    std::sync::Arc::make_mut(&mut obj.base_trigger_definitions).push(trigger.clone());
-    obj.trigger_definitions.push(trigger);
+    obj.push_printed_trigger(build_evoke_etb_sac_trigger());
 }
 
 fn offspring_etb_copy_trigger_for_ordinal(origin_ordinal: u32) -> TriggerDefinition {
@@ -3224,15 +3222,14 @@ pub(crate) fn ensure_paid_offspring_etb_copy_triggers(
             if !obj.trigger_definitions.iter_all().any(|entry| {
                 is_offspring_etb_copy_trigger_for_ordinal(entry.definition(), origin_ordinal)
             }) {
-                obj.trigger_definitions
-                    .push(offspring_etb_copy_trigger_for_ordinal(origin_ordinal));
+                obj.relive_printed_trigger(|trigger| {
+                    is_offspring_etb_copy_trigger_for_ordinal(trigger, origin_ordinal)
+                });
             }
             continue;
         }
 
-        let trigger = offspring_etb_copy_trigger_for_ordinal(origin_ordinal);
-        std::sync::Arc::make_mut(&mut obj.base_trigger_definitions).push(trigger.clone());
-        obj.trigger_definitions.push(trigger);
+        obj.push_printed_trigger(offspring_etb_copy_trigger_for_ordinal(origin_ordinal));
     }
 }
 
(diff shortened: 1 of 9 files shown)


matthewevans:
CI has a concrete, reproducible blocker on this head: the generated Draw-replacement corpus now includes `bard, king of dale\tDraw\tMandatory\tnone\tDraw\tnested-draw\tIndividualDraw`, while `scripts/draw-replacement-corpus.tsv` lacks that row. The current Oracle text is an individual-card replacement (draw two instead), so `IndividualDraw` is the correct reviewed scope. Please add the alphabetized row after `asmodeus the archfiend` and rerun CI; that will allow the card-data job to reach and publish the required parse-diff. Maintainer repair commit `142e782215` is available locally but could not be pushed because the configured token was rejected by the fork remote.

matthewevans (approved):
Maintainer review complete: printed-trigger provenance fix and regression coverage are clean on the current head.

Issue #6221, opened 2026-07-20
GATED_SETS is unset — 20 unreleased Hobbit preview cards are in the pool unbanned


mcbradd:
## Summary

`GATED_SETS` is unset everywhere in the pipeline, so the release gate is currently a no-op — and 20 unreleased The Hobbit (HOB) preview cards are in the exported card pool with their real (unbanned) legalities.

## The designed mechanism (implemented, but unconfigured)

`crates/engine/src/database/set_gating.rs` module docs describe a hybrid release gate:

- A single env var, `GATED_SETS` (comma-separated set codes, case-insensitive), read at card-data / set-list / draft-pool **generation** time. "To gate a set before release, generate with `GATED_SETS=MSH,MSC,TMSH`."
- Gated cards **stay in card-data** (browsable) but have their legalities overridden to `Banned` in every format (`oracle_gen.rs` applies `all_formats_banned()` to every card whose printings are all gated), which excludes them from every format-scoped deck-builder pool; the sets themselves are hidden from draft/picker/set-list output.
- Sets whose MTGJSON `releaseDate` is on or before `GATED_SETS_AS_OF` (default: UTC today) are **automatically ungated** even while still listed — so the config self-expires on release day and a forgotten env cleanup cannot leave cards banned (#4365).
- When unset or empty, gating is a no-op.

## Current state

- `GATED_SETS` appears nowhere in `.github/workflows/`, `scripts/` (only a comment at `gen-card-data.sh:26`), or the `Tiltfile`. Only `GATED_SETS_AS_OF` is exported (`gen-card-data.sh:29`). Gating is therefore a no-op in every generation path, CI included.
- MTGJSON `SetList.json`: HOB "The Hobbit" and HOC "The Hobbit Commander" both have `releaseDate: 2026-08-14` and `isPartialPreview: true` (spoiler-season data, subject to change).
- The current export contains **20 HOB-only preview cards** (38 atomic cards have an HOB printing, 23 are HOB-only, 20 of those parse into the pool) — e.g. Bard, King of Dale; Bard the Bowman; Bifur, Melodic Rider; Bilbo, Thief in the Night; Bolg of the North; Desolation of Smaug; Elrond, Moon-Reader; Gollum, Riddle Master.
[...]

PR #6222, opened 2026-07-20
chore(card-data): re-freeze draw-replacement corpus for Bard, King of Dale (HOB set unlock)

mcbradd:
## What

Re-freezes the Plan 03 draw-replacement corpus baseline (`scripts/draw-replacement-corpus.tsv`) with exactly one new row:

```
bard, king of dale	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
```

One line, no parser or engine change. Regenerated with `scripts/draw_replacement_census.py --corpus --write`; `--corpus --check` passes afterward (51 rows).

## Why every PR is red

The CI step "Draw replacement corpus freeze (Plan 03 / CR 121.2)" has failed on main and every PR since 2026-07-20T00:00Z. The trigger is **wall-clock, not any merged commit**: the card-data job caches `AtomicCards.json` under the ISO-week key `mtgjson-atomic-$(date +%Y-W%V)` (exact-match, no restore-keys — see `ci.yml`), so at Monday 00:00 UTC the key rolled to the new week and forced a fresh MTGJSON download that now carries The Hobbit (HOB) spoiler data, including Bard, King of Dale, whose draw-replacement row the frozen corpus had never seen.

Tonight's merges are exonerated — including #6174, which touched only the swallow-detector. Note the #4365 release-date gate did **not** engage: `GATED_SETS` is unset in CI (gating is a documented no-op when unset), and HOB's MTGJSON releaseDate is 2026-08-14 (`isPartialPreview: true`). The single-row failure reproduces locally with no gating configured. The gating gap is filed separately as #6221.

Failing runs:
- PR #6195: [Card data job 88256861966](https://github.com/phase-rs/phase/actions/runs/29711854779/job/88256861966)
- PR #6214: [Card data job 88254786623](https://github.com/phase-rs/phase/actions/runs/29710999164/job/88254786623)
- main's own post-midnight push CI is red the same way: [Card data job 88249911454](https://github.com/phase-rs/phase/actions/runs/29708804507/job/88249911454)

## Why the row is correct (CR 121.2, not CR 121.2a)

Bard, King of Dale (The Hobbit, HOB #144): "If you would draw a card except the first one you draw in each of your draw steps, draw two cards instead." This is an individual-draw replacement per CR 121.2 (cards are drawn one at a time): fixed substitute count, no `EventContextAmount` read — so it is *not* a CR 121.2a instruction-count modifier. Its census shape and classification are identical to the already-blessed precedent rows:

```
alhammarret's archive	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
teferi's ageless insight	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
```

Engine-emitted `draw_scope` and the census's independently derived scope agree (`IndividualDraw`) — a disagreement would have raised `ClassificationError`, not a diff row.

## This blesses PREVIEW text

HOB is `isPartialPreview: true` and releases 2026-08-14, so this row freezes spoiler-season text. If Bard's Oracle text changes before release, the exact-match freeze goes red again and forces a re-review — by design, that is the gate doing its job, not a regression. Current Scryfall preview text matches the parsed row exactly.

## Effect of merging

- Unblocks #6195, #6214, main, and every other open PR.
- Observational, **not addressed here**: the same set-unlock event moved coverage ratios (card pool 34653 → 34702; 20 HOB-only preview cards entered the export), which is the likely cause of main's separate post-midnight "Rust (fmt, clippy, test, coverage-gate)" failure ([job 88250786912](https://github.com/phase-rs/phase/actions/runs/29708804507/job/88250786912)). Left to a separate change.

### DIFF
--- a/scripts/draw-replacement-corpus.tsv
+++ b/scripts/draw-replacement-corpus.tsv
@@ -27,6 +27,7 @@ abundance	Draw	Optional	none	Choose	-	IndividualDraw
 alhammarret's archive	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
 archmage ascension	Draw	Optional	none	SearchLibrary	-	IndividualDraw
 asmodeus the archfiend	Draw	Mandatory	none	ExileTop	-	IndividualDraw
+bard, king of dale	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
 blood scrivener	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
 breathstealer's crypt	Draw	Mandatory	none	Draw	nested-draw	IndividualDraw
 chains of mephistopheles	Draw	Mandatory	none	Discard	nested-draw	IndividualDraw


matthewevans:
**Not enqueued — this branch is now an obsolete duplicate.**

The row itself is correct: Bard's current Oracle text is an individual-draw replacement under CR 121.2, and the CI `Card data` check passed on this head. However, current `main` already contains the identical `bard, king of dale\tDraw\tMandatory\tnone\tDraw\tnested-draw\tIndividualDraw` row in commit `3a5d8671`.

Merging this behind branch produces the current `main` tree with no corpus delta, so it adds no remaining value and should not enter the merge queue. Thanks for identifying the weekly-MTGJSON corpus drift promptly.

Issue #5683, opened 2026-07-12
Conditional "instead" clauses lower to unconditional chains — engine executes BOTH halves (94-face class)


matthewevans:
## Defect class

Oracle text of the form "do A. [If C,] do B **instead**" (CR 614.1a: effects using "instead" are replacement effects) should lower to a branch — one half or the other. On a measured subset of the pool it instead lowers to an **unconditional chain or sibling**, and `SubAbilityLink::SequentialSibling` is documented in `types/ability.rs` as "independent of the parent … Always resolves" — so the engine executes **both** effects and drops the condition entirely.

This is a correctness bug on cards the coverage report marks **supported**, not a coverage gap.

## Verified witnesses (Oracle text and parse both read from the full-pool export)

- **A-Paragon of Modernity** — "{3}: This creature gets +1/+1 until end of turn. If exactly three colors of mana were spent to activate this ability, put a +1/+1 counter on it instead."
  Parse: `Pump → sub_ability PutCounter(SequentialSibling)`, `condition: null` on both → always pumps AND always adds the counter; the three-color gate is gone.
- **Bring the Ending** — "Counter target spell unless its controller pays {2}. Corrupted — Counter that spell instead if its controller has three or more poison counters."
  Parse: two sibling `Counter` abilities, neither conditioned → counters unconditionally; both the Corrupted gate and the "instead" exclusivity are gone. (Independently re-verified from the export by the reviewer of this report.)
- **Back for Seconds** — the bargained alternative carries `AdditionalCostPaid` but as a SIBLING, not an alternative → when bargained, both halves run.
- **Anthem of Rakdos** (adjacent shape, same family) — the Hellbent damage replacement parses to a `Continuous` static with `modifications: []` — an empty shell holding only the condition; the replacement itself is gone.

## The parser can already do it right

33 faces pool-wide use the correct `condition` + `else_ability` branch (e.g. Accumulate Wisdom, Celestial Reunion, Flow State), and the kicker/bargain family is correctly captured via `AbilityCondition::AdditionalCostPaidInstead` (e.g. Bloodchief's Thirst). The bug is the remaining condition shapes falling through to chain lowering.

## Population (full pool, 35,396 faces)

- **94 faces** carry an "instead" clause with no typed representation in the parse (measured by the de-vacuumized `Replacement_Instead` swallow detector after its description-channel evidence was replaced with typed carriers — before that fix the detector reported 11 of these 94).
[...]

PR #6223, opened 2026-07-20
test(parser): pin Bring the Ending Corrupted "instead" as a branch (#5683 Class B anchor)

xfodev:
Part of #5683 (conditional "instead" clauses lowering to unconditional chains).

## Context

#5683 names four regression anchors. I dumped all four through the production parser (`parse_oracle_text`) against current `main`, and **Bring the Ending is already lowered correctly** — the cross-line self-replacement binder (the same path that fixed Anoint with Affliction) produces a single `Counter` ability whose Corrupted override attaches as a `ConditionInstead`-gated `sub_ability`, not two unconditional `Counter` siblings. This PR **locks that anchor** so it can't silently regress back to the double-counter chain while the other three classes (Paragon / Back for Seconds / Anthem of Rakdos) are worked separately.

Verified AST (current `main`):
```
Counter{StackSpell} unless_pay:{2}
  └─ sub: condition = ConditionInstead{ QuantityCheck{ TargetControllerCounter{Poison}, GE, 3 } }
          effect    = Counter
```

## Test

`crates/engine/tests/integration/bring_the_ending_corrupted_instead_branch_5683.rs` (registered in `tests/integration/main.rs`) asserts through `parse_oracle_text`: exactly one ability (not two siblings), base `Counter` retaining its `unless … pays {2}` cost, and the override sub gated by `AbilityCondition::ConditionInstead`. RED if the override ever regresses to an unconditional sibling.

## Verification

Locally verified: I compiled the exact assertion body against the current engine rlib (via a bootstrapped `zig cc` linker) and it passes — this pins **current, real** behavior, not an aspiration. (The full `cargo test` binary OOMs on this host, so CI runs the registered test.)

`Gate A PASS head=01ef81645… base=e259c4c65…` (mirrors the existing `anax_instead_branch_not_chain` regression, CR 614.1a / 614.6 / 614.15).

Model: claude-opus-4-8
Thinking: High
Tier: Frontier

### DIFF
--- a/crates/engine/tests/integration/bring_the_ending_corrupted_instead_branch_5683.rs
+++ b/crates/engine/tests/integration/bring_the_ending_corrupted_instead_branch_5683.rs
@@ -1,0 +1,81 @@
+//! Issue #5683 (Class B regression anchor) — CR 614.1a + CR 614.6: a
+//! "Counter that spell **instead** if <cond>" ability-word override is a BRANCH,
+//! not a second sibling ability.
+//!
+//! CR 614.1a: "Effects that use the word 'instead' are replacement effects."
+//! CR 614.6:  "If an event is replaced, it never happens…"
+//!
+//! Witness (Oracle read from the pool export):
+//!   Bring the Ending —
+//!     "Counter target spell unless its controller pays {2}.
+//!      Corrupted — Counter that spell instead if its controller has three or
+//!      more poison counters."
+//!
+//! The #5683 defect class is "instead" clauses lowering to an unconditional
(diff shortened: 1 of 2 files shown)


matthewevans (approved):
**Approved.** The corrected fixture now uses Bring the Ending's authoritative `Instant` type. The registered parser regression test discriminates the intended shape: one `Counter` ability retains `unless … pays {2}`, while the Corrupted override is a `ConditionInstead` sub-ability rather than an unconditional sibling.

PR #6229, opened 2026-07-20
fix(engine): resume interrupted APNAP queues when a chooser leaves the game

### LINKED ISSUE #4823: Parked replacement choice: pre-existing strand-freezes when the chooser leaves the game (PPT phase-advance, team-draw skip, deferred-trigger drain)
## Summary

When a player who owns a **parked replacement choice** (`WaitingFor::ReplacementChoice`) leaves the game (any elimination path: life ≤ 0, poison, draw-from-empty, commander damage, concede, 2HG cascade), `do_eliminate` marks them out and the post-loop reconcile rewrites `waiting_for` to `Priority{next}` (`elimination.rs:116-121`) — but several **coupled continuation slots** are left stranded because the engine only drains them via the `(WaitingFor::ReplacementChoice, GameAction::ChooseReplacement)` resume, which is now unreachable.

PR #4680's follow-up fix (commit `1020f5334`) cleared the specific slot that caused an **SBA freeze** (`pending_replacement`) plus its tightly-coupled continuation. This issue tracks the **remaining pre-existing strand-freezes** on the *same* trigger, which were deliberately deferred because they need proper **interrupted-queue resumption** (skip the departed player and continue the APNAP drain), not field-nulling.

These predate PR #4680 and are 3+ player only (in 2-player games eliminating the chooser ends the game via `GameOver` first).

## The stranded slots and their consequences (verified against code)

1. **`pending_phase_transition_progress` → phase-advance freeze.**
   `auto_advance` early-returns while this is `Some` (`turns.rs:1919`). It is set at `turns.rs:174`, cleared only at `turns.rs:201` (queue empty) or drained via `drain_pending_phase_transition_progress` (`turns.rs:193`, reachable only from the replacement resume). A step-end empty-mana drain (CR 703.4q) that pauses on a CR 616.1 `EmptyManaPool` ordering choice retains the rest of the APNAP queue in PPT while parking `ReplacementChoice{player}`. If that player leaves, PPT is stranded → the phase transition never completes → soft-lock. (`turns.rs:1908-1915` documents this exact pause.)

2. **`pending_team_draw_step` → next living player's draw silently skipped.**
   Global slot seeded in **all** formats, not just 2HG (`execute_draw`, `turns.rs:1319-1330`, seeds `[active_player]` only when empty). A draw-step draw can park on a CR 616.1 competing-draw replacement; if the drawer leaves, the stale `[dead drawer]` entry survives, and the next living player's `execute_draw` sees a non-empty queue → does not seed → `drain_pending_team_draw_step` (`turns.rs:1345-1352`) drains the dead entry → that player's turn-based CR 504.1 draw is skipped.

3. **`pending_continuation` → deferred-trigger drain freeze.**
   `can_drain_deferred_triggers` returns `false` while this is `Some` (`triggers.rs:4468`). If it belongs to the departed chooser, deferred triggers never drain.

The remaining resume-drained slots (`pending_counter_moves`, `pending_counter_additions`, `pending_batch_deliveries`, `pending_copy_token_resolution`, `deferred_step_trigger_resume`, `pending_step_end_mana_handlers`) are not known to gate a progression loop (the only external `is_some()` guard among them, `pending_batch_deliveries` at `engine.rs:3528`/`3566`, is inside the `ReturnAsAuraTarget` handler), but they hold incomplete in-flight work and should be considered in the same fix.

## Why this needs its own change (not part of the SBA-freeze fix)

Nulling these is **wrong** — they hold interrupted **APNAP queues shared across players** (the phase-transition drain, the team-draw queue) and may belong to the spell **controller's** resolution, not the affected **chooser's** choice. The correct fix is to **resume** the interrupted operation for the remaining players (continue the drain skipping the departed player), which is materially larger new logic with its own correctness surface and tests.

## Suggested direction

A single authoritative "a parked-continuation owner left the game" routine (CR 800.4a), invoked from the elimination path, that for each interrupted-queue slot either resumes the drain for the remaining players or cleanly finalizes the transition — rather than the current per-slot teardown. Regression tests per slot: chooser of a step-end empty-mana choice is eliminated (assert phase advances); chooser of a draw-step competing-draw is eliminated (assert next player still draws); chooser holding `pending_continuation` is eliminated (assert deferred triggers drain).

## References
- Fix for the SBA-freeze facet: commit `1020f5334` (`do_eliminate` in `crates/engine/src/game/elimination.rs`).
- CR 800.4a (player leaving), CR 616.1 (replacement-order choice belongs to the affected player), CR 703.4q / CR 500.4 (empty mana pool), CR 504.1 (draw step).

dale053:
## Summary

Fixes #4823. In a 3+ player game, when a player with a parked replacement choice leaves the game, elimination only redirected the surface-level `WaitingFor` prompt to `Priority` — it never re-invoked the drain function for whichever field that player was actually blocking, stranding:

- `pending_phase_transition_progress` (phase-advance freeze, after an empty-mana-pool CR 616.1 ordering choice)
- `pending_team_draw_step` (next living player's draw silently skipped)
- `pending_continuation` (deferred-trigger drain freeze)

- Hardens `resolve_chooser` (`choose_from_zone.rs`) to re-check liveness on every call instead of trusting a stale snapshot, with CR 800.4g/h fallback redirection.
- Adds `redirect_departed_chooser`, which re-points an in-flight per-player/per-category `ChooseFromZoneChoice` at a live chooser in place — preserving the already-offered `cards`/`count`/`up_to`/`constraint` payload rather than letting the generic collapse silently reduce it to an empty pick.
- Adds `prune_player_ordered_queues` + `resume_interrupted_apnap_queues` in `elimination.rs`: a `Priority`-gated cascade (team-draw → phase-transition → continuation, in that order — phase-transition has the fewest other retry paths in the engine, so it must run before continuation can strand it) that resumes all three stranded queues for the remaining players per CR 800.4a.
- Terminal `GameOver` branch now clears six pending fields it previously left dangling.

Two narrower, structurally-similar gaps were discovered during review and are documented in-code as explicit, tracked follow-ups rather than fixed here (consistent with this repo's scope discipline): `PendingChooseOneOf`/`PendingVoteBallotIteration`'s current-actor departure isn't redirected the same way, and a departed non-chooser zone owner's already-offered cards aren't revalidated against `exile_owned_objects_on_player_left_game`. Neither reproduces a softlock; both are narrower than this issue's reported symptoms.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo clippy -p engine --all-targets -- -D warnings` clean
- [x] `cargo test -p engine --lib elimination::` — 30 passed
- [x] `cargo test -p engine --lib 'game::effects::choose_from_zone::tests'` — 29 passed
- [x] New end-to-end test drives the real `eliminate_player` production entry point and verified to fail when the fix is reverted
- [x] All CR annotations (800.4a, 800.4g, 800.4h, 101.4, 608.2c, 616.1) verified against `docs/MagicCompRules.txt`

Model: Claude Sonnet 5


matthewevans:
This PR is being closed without implementation review.

It changes rules-critical APNAP/elimination/replacement-resolution behavior, but the PR provides no evidence that it followed the required `engine-implementer` pipeline (anchored plan, Gate A, and final `review-impl` for the committed head). Its stated local validation also conflicts with this repository’s Tilt-first workflow, and the required Rust check is currently failing.

This is not an objection to the model named in the PR. The closure is based on the required engineering process and a recurring contributor pattern already recorded in maintainer review history. Future PRs from this contributor that touch rules-critical behavior without the required `engine-implementer` pipeline and independently reproducible validation will be closed immediately rather than reviewed or repaired by maintainers.

PR #6230, opened 2026-07-20
fix(engine): auto-tap self-sacrifice mana abilities like Gold tokens

### LINKED ISSUE #6157: Gold tokens are not used by automatic mana payment
## Summary
Gold tokens can be activated manually/clicked to produce mana, but the automatic payment path does not consider them as available mana sources. Treasure tokens are considered and auto-paid correctly in the same kind of payment flow.

## Observed behavior
- A player controls one or more Gold tokens.
- Clicking a Gold token to use its mana ability works.
- Casting/activating something with automatic payment enabled does not automatically use those Gold tokens.
- Treasure tokens with the analogous sacrifice-for-mana role are selected by auto payment.

## Expected behavior
Automatic mana payment should treat supported Gold-token mana abilities as available mana sources when they can pay the pending cost, just as it does for Treasure tokens.

## Notes
Likely engine-side, not frontend-side: manual activation works, while auto-payment source discovery/selection appears to skip Gold. Relevant surfaces include `crates/engine/src/game/mana_sources.rs`, `crates/engine/src/game/casting_costs.rs`, and the auto-payment preview path in `crates/engine/src/game/preview.rs`.

## Repro needed
Add a minimal engine integration test that creates a Gold token, attempts to cast a generic or colored-cost spell with `CastPaymentMode::Auto`, and asserts the Gold token is sacrificed/consumed and the spell advances. Include a Treasure control case if needed to lock parity.

bitfathers94:
fix(engine): auto-tap self-sacrifice mana abilities like Gold tokens

Auto-tap source discovery required every scanned mana ability to have a
{T} cost component, so Gold's "Sacrifice this token: Add one mana of any
color." (no {T}) was invisible to CastPaymentMode::Auto even though the
sacrifice targets only the token itself and needs no player choice —
exactly as deterministic as a {T} cost. Treasure worked only because its
cost happens to also include {T}.

Extend the single-authority eligibility check to also accept an
unambiguous self-sacrifice cost (Sacrifice targeting SelfRef, count 1),
and exclude that same shape from the "needs manual payment" fallback path
so it isn't double-counted. Ambiguous sacrifice costs (Krark-Clan
Ironworks' "Sacrifice an artifact") still require a manual choice and are
unaffected.

Closes #6157

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic mana payment for Gold and Treasure when their mana abilities use choice-free self-sacrifice costs, including tapless variants.
  * Automatic selection now preserves eligible mana sources even when tapped or summoning-sick, while continuing to enforce normal tap gating for other abilities.
  * Refined eligibility to avoid auto-using composite mana abilities that include player-choice elements (LED-style), preventing incorrect auto-tap/auto-pay behavior.

* **Tests**
  * Added integration coverage for auto-payment success with Gold and Treasure, including tapped/summoning-sick eligibility.
  * Added regression coverage confirming choice-containing composite abilities are not auto-selected.


matthewevans (changes requested):
The new self-sacrifice predicate is never reached for a tapped source. `auto_tap_mana_options` still returns early on `obj.tapped` before `scan_mana_abilities` can call `is_active_tap_mana_ability`; that early gate is only valid for a `{T}` cost. A tapped Gold token can legally pay “Sacrifice this token: Add one mana of any color”, but the proposed auto-payment path offers no source and the spell remains unpayable. Please make the untapped/summoning-sickness prefilter conditional on the selected ability requiring `{T}` (or let the existing activation/payability predicate make that decision), and add a tapped-Gold regression test.

bitfathers94:
Thanks for the precise catch, @matthewevans — you were exactly right that the new predicate was never reached for a tapped source. Fixed and rebased on latest main.

**What changed**
The untapped/summoning-sickness prefilter is now conditional on the selected ability requiring `{T}`. There were two object-level early returns dropping a tapped/summoning-sick permanent *before* `scan_mana_abilities` could evaluate it:
- `mana_sources.rs::auto_tap_mana_options` (the Gold-token path)
- `casting_costs.rs::collect_sorted_auto_tap_source_options` (the outer board sweep)

Both now gate the tapped/summoning-sick/can't-tap exclusion behind a new helper `object_has_tapless_self_sacrifice_mana_ability` (a self-sacrifice mana ability with **no** `{T}`/`{Q}` component, per CR 106.12 / CR 302.6). A tapless self-sac source (Gold) now falls through to the payability predicate; a `{T}`-gated source (e.g. Treasure's `{T}, Sacrifice`) is still correctly excluded when tapped. I also added a cheap per-ability guard inside `scan_mana_abilities` on the auto-tap path so no `{T}`/`{Q}` ability of a tapped source leaks through.

**Regression test**
`tapped_gold_token_still_auto_taps_for_self_sacrifice_mana` (integration/issue_6157): a Gold token with `tapped=true` **and** `summoning_sick=true` must still auto-pay via self-sacrifice. Revert-verified — it panics (spell unpayable) on the pre-fix code and passes after.

Green locally: `cargo build`/`clippy` clean (0 warnings), the 3 issue_6157 integration tests pass, and the full `mana` lib suite (1241 tests) is green. Ready for another look.

matthewevans:
Current-head review completed. The tapped/summoning-sick Gold regression now reaches the per-ability gate, while `{T}`/`{Q}` abilities remain gated there; the integration test exercises the reported path. I also verified the Sacrifice citations against Wizards' current Comprehensive Rules, where Sacrifice is CR 701.21, so the older bot citation finding is not applicable.

No approval or enqueue yet: the current Rust and frontend required checks are still running, and CodeRabbit's current-head review is in progress. I will leave this held for those external gates rather than override them.

matthewevans (changes requested):
**Blocking: the auto-tap eligibility predicate currently accepts a cost that still requires a player choice.**

## 🔴 Blocker

`crates/engine/src/game/mana_sources.rs:282-291` uses `cost_has_component`, whose `Composite` implementation is `costs.iter().any(...)`. Consequently, a mana ability with `Composite[Discard, Sacrifice(SelfRef, 1)]` is classified as an “unambiguous” self-sacrifice merely because one component matches. That classification is consumed both by the manual-payment exclusion at `mana_sources.rs:1276-1283` and by the auto-tap eligibility gate at `mana_sources.rs:1820-1824`, so the ability can be selected automatically even though it still needs a discard choice.

This is a real existing cost shape, not a hypothetical: the Lion's Eye Diamond activation test builds exactly `Composite[Discard { selection: Chosen }, Sacrifice(SelfRef, 1)]` at `crates/engine/src/game/mana_abilities.rs:6209-6223`; the activation pipeline then exposes `WaitingFor::PayCost { kind: Discard }` at `mana_abilities.rs:1450-1468` and asserts that prompt at `6240-6256`. Auto-tap must not bypass that prompt.

The established authority already models the needed invariant: `mana_abilities.rs:3027-3041` is a deny-by-default full-tree whitelist, allowing only `Tap`, `Sacrifice(SelfRef, 1)`, and composites whose *every* component is choice-free. Please make the auto-tap predicate share/extract that full-cost authority (or apply the same full-tree invariant), then add a regression proving the LED-shaped composite remains on the manual-payment path while bare Gold and `{T} + self-sacrifice` Treasure remain eligible.

## ✅ Clean

The current head fixes the prior tapped/summoning-sick Gold reachability finding; that issue is independent of this composite-cost blocker.

**Recommendation: request changes. Tighten eligibility to the full, choice-free cost tree and cover the existing LED shape before approval.**

bitfathers94:
Good catch, @matthewevans — you're right that `cost_has_component`'s `any`-over-`Composite` let a partial match through, so an LED-shaped `Composite[Discard{Chosen}, Sacrifice(SelfRef,1)]` was misclassified as deterministic. Fixed by reusing the authority you pointed to rather than adding a special case.

**What changed**
- `mana_abilities.rs`: exposed `cost_component_choice_free` as `pub(crate)` (the existing deny-by-default full-tree whitelist: only `Tap`, `Sacrifice(SelfRef,1)`, and `Composite`s whose *every* component is choice-free).
- `mana_sources.rs`: `has_unambiguous_self_sacrifice_component` now requires **both** (1) the whole cost tree passes that shared `cost_component_choice_free` authority, **and** (2) a `Sacrifice(SelfRef,1)` component is actually present. Both consumers — the manual-payment exclusion (`:1276-1283`) and the auto-tap eligibility gate (`:1820-1824`) — go through this one predicate, so they're fixed together.

Result: LED (`Composite[Discard, Sacrifice]`) is rejected by the full-tree whitelist → stays on the manual-payment path and keeps its `WaitingFor::PayCost{ Discard }` prompt. Bare Gold (`Sacrifice(SelfRef,1)`) and Treasure (`Composite{Tap, Sacrifice(SelfRef,1)}`) still pass → remain auto-tap eligible.

**Regression (end-to-end)**
`led_shaped_discard_sacrifice_not_auto_tapped_stays_manual`: board holds an LED-shaped source **and** a Gold token; auto-resolving a {1} spell pays with Gold, drives to `WaitingFor::Priority` (no bypassed prompt), and **LED remains on the battlefield**. Revert-verified — reverting the predicate to the `any`-match makes LED contaminate the auto-tap plan and the payment fails, so the test genuinely pins the invariant. (Kept the predicate-level unit test too.)

Green locally: `cargo build`/`clippy` clean (0 warnings under `-D warnings`), all 4 issue_6157 integration tests pass, full `mana` lib suite 1242 green. Ready for another look.

matthewevans:
Maintainer fixup `3ef7451` replaces the LED-only test's vacuous `try_resolve().is_err()` assertion with a positive `WaitingFor::ManaPayment` reach guard and assertions that the pending cast and LED source are preserved.

The current head is held pending its required CI run and CodeRabbit review. No approval or merge-queue enrollment has been performed.

matthewevans (approved):
Approved on current head 4656be66: the LED-only regression now drives the real CastSpell action and asserts the manual ManaPayment boundary; Gold/Treasure/tapped-Gold and LED cases remain covered. Current CodeRabbit completed successfully; auto-merge will wait for the remaining required checks.

PR #6231, opened 2026-07-20
fix(parser): route "different kinds of counters" quantity to DistinctCounterKindsAmong

### LINKED ISSUE #6219: [Card Bug] Perrie, the Pulverizer's attack ability does not work
## Card(s)

Perrie, the Pulverizer

## Build/version

Not provided.

## Game mode

Unknown.

## Actual behavior

Perrie's enters-the-battlefield ability works and places a shield counter, but its second ability does not work when Perrie attacks.

## Expected behavior

Whenever Perrie attacks, the triggered ability should target a creature its controller controls. That creature should gain trample and get +X/+X until end of turn, with X equal to the number of distinct kinds of counters among permanents that player controls.

## Steps to reproduce

1. Put Perrie onto the battlefield and confirm its enters-the-battlefield ability resolves.
2. Control one or more permanents with counters, then attack with Perrie.
3. Observe that Perrie's attack-triggered ability does not produce the expected trample and +X/+X effect.

Exact board state and whether the trigger appears on the stack were not provided; targeted runtime reproduction should distinguish trigger dispatch from quantity resolution.

> When Perrie enters, put a shield counter on target creature. (If it would be dealt damage or destroyed, remove a shield counter from it instead.)
>
> Whenever Perrie attacks, target creature you control gains trample and gets +X/+X until end of turn, where X is the number of different kinds of counters among permanents you control.

## Classifier analysis

**Matched clause**: “Whenever Perrie attacks, target creature you control gains trample and gets +X/+X until end of turn, where X is the number of different kinds of counters among permanents you control.” (supported: true)

**Reasoning**: The attack trigger, controlled-creature target, duration, trample grant, and dynamic power/toughness modifications are represented. However, X currently lowers to `CountersOnObjects` with a counter type serialized as `"different kinds of"`. That models a literal counter type instead of the cardinality of distinct counter kinds among matching permanents. The engine already has the general `DistinctCounterKindsAmong` quantity primitive, so this phrase should lower to that typed reference and be verified through the attack-trigger runtime pipeline.

---
_Posted using the repository's bug-triage and bug-coverage-classifier workflows._

bitfathers94:
## Summary

Fixes #6219: Perrie, the Pulverizer's attack trigger ("Whenever Perrie
attacks, target creature you control gains trample and gets +X/+X until end
of turn, where X is the number of different kinds of counters among
permanents you control") never applied the +X/+X boost. Trample landed fine,
but X always resolved to 0.

Root cause: the "the number of [counter type] counters among <filter>"
generic quantity arm treated "different kinds of" as a literal counter type
name. Since no real counter is ever named that, `CountersOnObjects` always
counted zero. The engine already has the correct primitive for this —
`QuantityRef::DistinctCounterKindsAmong` — used by the "for each kind of
counter on/among <filter>" `repeat_for` reading (Bribe Taker), but nothing
routed the "the number of different kinds of counters among <filter>"
dynamic-quantity surface form to it.

Added `parse_distinct_counter_kinds_among_tail` to
`oracle_nom/quantity.rs`, a nom combinator that recognizes "different
kind[s] of counters {on|among} <filter>" and emits
`QuantityRef::DistinctCounterKindsAmong { filter }`, nested alongside
`parse_distinct_subtypes_among` in `parse_number_of_inner` so it's tried
before the generic controlled-type/counter-type fallback.

## Files changed

- `crates/engine/src/parser/oracle_nom/quantity.rs` — new combinator +
  registration + 2 parser unit tests (plural/singular surface forms).
- `crates/engine/tests/integration/perrie_the_pulverizer_attack_trigger_6219.rs`
  — new runtime regression driving the real attack trigger through
  declare-attackers → stack → resolution, asserting the +2/+2 pump and
  trample grant land on Perrie.
- `crates/engine/tests/integration/main.rs` — registers the new test module.

## CR references

- CR 122.1 (counters, distinct kinds by name/description)

## Implementation method

Method: manual (parser combinator + runtime regression), traced through
the existing `DistinctCounterKindsAmong` primitive and its `for each`
sibling before writing the fix.

## Track

Developer

## LLM

Model: Claude (Sonnet 5) / Thinking: standard / Tier: n/a

## Verification

- [x] `cargo clippy -p engine --all-targets --features engine/proptest -- -D warnings` — 0 warnings
- [x] `cargo test -p engine --lib` — 17301 passed, 0 failed, 6 ignored
- [x] `cargo test -p engine --test integration` — 3585 passed, 0 failed, 2 ignored (includes the new Perrie regression)
- [x] `./scripts/check-parser-combinators.sh origin/main` — `Gate A PASS head=df13846e3c03fa29902d4bd93e8f4eefd84972f7 base=2f4608051c4d694d3ec3fd59c3e68c26d3746559`
- [x] Revert check: `git apply -R` on the parser-fix hunk flips the new integration test red (X resolves to 0, P/T stays 3/3); re-applying turns it green (5/5 + trample) — confirms the test actually pins the fix.

## Gate A

Gate A PASS head=df13846e3c03fa29902d4bd93e8f4eefd84972f7 base=2f4608051c4d694d3ec3fd59c3e68c26d3746559

## Anchored on

- `crates/engine/src/parser/oracle_nom/quantity.rs:1626` — `parse_for_each_distinct_counter_kinds_among` (the `repeat_for` sibling reading of the same `DistinctCounterKindsAmong` primitive)
- `crates/engine/src/parser/oracle_nom/quantity.rs:1595` — `parse_number_of_distinct_colors_among_permanents_tail` (the same "the number of `<distinct-X>` among `<filter>`" CDA-quantity shape, colors side)

## Claimed parse impact

1 card (Perrie, the Pulverizer) moves from `supported_aspect_defect` (X
always 0) to fully correct. No other card's Oracle text begins with
"different kind of counter" / "different kinds of counters" in a way the
prior generic arm would have handled differently — this arm is nested
strictly before the generic controlled-type fallback and only fires on
that literal adjective prefix, so no existing `supported:true` card
regresses.

Scope Expansion: None.
Validation Failures: None.
CI Failures: None.

### DIFF
--- a/crates/engine/src/parser/oracle_nom/quantity.rs
+++ b/crates/engine/src/parser/oracle_nom/quantity.rs
@@ -1468,7 +1468,16 @@ fn parse_number_of_inner(input: &str) -> OracleResult<'_, QuantityRef> {
         // generic controlled-type/type-filter arms whose leading token would
         // otherwise commit. Nested together to stay within nom's top-level `alt`
         // arity (nom 8.0 max: 21 items).
-        alt((parse_distinct_subtypes_among, parse_turns_taken_this_game)),
+        //
+        // CR 122.1: "different kind[s] of counters {on|among} <filter>" — the
+        // dynamic-quantity reading of the counter-kind cardinality (Perrie, the
+        // Pulverizer: "X is the number of different kinds of counters among
+        // permanents you control") — shares this nest for the same reason.
+        alt((
+            parse_distinct_subtypes_among,
+            parse_turns_taken_this_game,
+            parse_distinct_counter_kinds_among_tail,
+        )),
         // CR 201.2 + CR 603.4: "differently named <type-phrase>" (distinct-by-name)
         // and "different <power|mana value> among <type>" (distinct-by-quality —
         // Celebrate the Harvest's "the number of different powers among ..."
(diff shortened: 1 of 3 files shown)


matthewevans (approved):
Approved on current head `df13846e3c03fa29902d4bd93e8f4eefd84972f7`: this reuses the existing distinct-counter-kind resolver, preserves the typed quantity seam, and includes a discriminating runtime attack regression.

PR #6233, opened 2026-07-20
fix(parser): keep the "up to N target" count on gain-control effects

### LINKED ISSUE #6205: [Card Bug] The Super Hero Civil War: Can only select one creature
### Card(s)

The Super Hero Civil War

### Build/version

0.30.0

### Game mode

AI

### Actual behavior

When casting The Super Hero Civil War it asks to select target creatures, but only one

### Expected behavior

Should allow to select two creatures with a total mana value of up to 6

### Steps to reproduce

- Opponent has multiple creatures with mana value less than 6
- Cast The Super Hero Civil War

[game-state-turn-11-2026-07-19T17-39-49-295Z.json](https://github.com/user-attachments/files/30168823/game-state-turn-11-2026-07-19T17-39-49-295Z.json)

minion1227:
## Summary

The Super Hero Civil War's chapter I reads "Gain control of up to two target creatures with total mana value 6 or less," but only one creature could ever be selected. The count was parsed and then thrown away.

This is a class defect, not a single-card one. The CI parse-diff shows **7 cards** affected — **3 were genuinely under-targeted**:

| Card | Oracle | Before | After |
|---|---|---|---|
| The Super Hero Civil War | "up to **two** target creatures with total mana value 6 or less" | 1 | 0-2 |
| Jace, Ingenious Mind-Mage | "up to **three** target creatures" | *no count* | 0-3 |
| Domineering Will | "target player gains control of up to **three** target nonattacking creatures" | 1 | 0-3 ¹ |

¹ Domineering Will's parsed **count** is now correct, but the card stays non-functional at runtime for two pre-existing reasons unrelated to this fix, both measured and documented in the test file: `GiveControl` never receives a recipient target slot (`effect_references_target_player` inspects only the effect's target filter, never `GiveControl`'s `recipient` field), and `"nonattacking"` parses to `Subtype("Attacking")`, which no creature has. Neither is introduced here and neither is fixed here.

The other 4 ("up to **one** target": Pyreswipe Hawk, Rangers of Ithilien, Scroll of Isildur, Jon Irenicus) already selected correctly and only change *representation* — from an implicit `optional (up to)` marker to an explicit `0..=1` range. A dedicated test pins that their optionality (`min = 0`) survives that move, since that is the one place this change could regress behavior rather than fix it.

Closes #6205

## Root cause

Both `gain control of ` entry points called `strip_optional_target_prefix`, which **already parses** the `up to N` quantifier and returns a `MultiTargetSpec`. Both bound it to `_`:

```rust
let (target_text, _) = super::strip_optional_target_prefix(rest);
```

So the spec never reached `ParsedEffectClause.multi_target`, and target selection fell back to a single optional slot.

The defect is specific to this verb. Every other "up to N target" shape kept its count — Call of the Death-Dweller (2), Patch Up (3), The War in Heaven (3) — including the same `TotalManaValue` constraint. The constraint and the trailing saga duration always parsed correctly; only the count was dropped.

`GiveControl` ("target player gains control of …") lowers through these same paths, which is why Domineering Will was capped at one target by the same line.

## Fix

Carry the spec on `TargetedImperativeAst::GainControl` and thread it onto the clause at lowering, mirroring the existing `Tap`/`Untap`/`Return` arms rather than inventing a new mechanism.

Both entry points are fixed:
- `oracle_effect/imperative.rs` — the plain imperative path (Jace's loyalty ability).
- `oracle_effect/mod.rs` — the compound/rider path that saga chapters with a trailing duration take (The Super Hero Civil War). Patching only the first left the issue half-fixed, so both are covered here.

Only the **targeted** form can carry a count — the mass "gain control of all/each" (`all: true`) has no target slots, so the new lowering arm is guarded on `all: false, multi_target: Some(_)` and untargeted mass gain-control is untouched.

The `Tap`/`Untap`/`Return`/`Support` arms in `lower_imperative_family_ast`, which solve exactly this problem: the bare-`Effect` lowering cannot carry a variable target count, so an AST that captured one is lowered in the family layer and the count is threaded onto the clause. The new arm sits directly below the Tap/Untap arm and follows its shape. No new parsing was written — `strip_optional_target_prefix` already returned the value; this stops discarding it.

CR annotations use the codebase's established pair for variable target counts, `CR 115.1d + CR 601.2c` (601.2c: "If the spell has a variable number of targets, the player announces how many targets they will choose"), matching the existing `multi_target` annotations in `oracle_ir/ast.rs`.

## Test plan

Registered integration test `crates/engine/tests/integration/gain_control_multi_target_6205.rs` — 5 tests, all passing:

- `saga_chapter_gain_control_keeps_up_to_two` — The Super Hero Civil War → 2 (the reported card, compound/rider path).
- `loyalty_gain_control_keeps_up_to_three` — Jace, Ingenious Mind-Mage → 3 (plain imperative path).
- `give_control_keeps_up_to_three` — Domineering Will → 3 (the `GiveControl` sibling, parse level).
- `up_to_one_stays_optional_with_an_explicit_upper_bound` — Pyreswipe Hawk → `min 0, max 1`, guarding the representation-only subgroup against losing optionality.
- `single_target_gain_control_declares_no_count` — plain "gain control of target creature" declares **no** count.
- `jace_ultimate_steals_three_creatures_through_target_selection` — **runtime regression** (added on review): activates the loyalty ability, answers one `WaitingFor::TargetSelection` slot per `GameAction::ChooseTarget` (CR 601.2c), and asserts all three creatures actually changed controller (CR 613.1b). Reverting the source fix leaves creatures two and three with their original controller.

There is deliberately no `GiveControl` runtime witness — see footnote 1; a passing one is not constructible today and a failing one would fail for an unrelated defect. Measuring the same ability's slots with and without `multi_target` gives 3 vs 1 creature slots and no player slot in either case, which both confirms the restored count reaches real slot generation and isolates the pre-existing gap.

The last two are the discriminating guards: the fix cannot degenerate into "always attach a `multi_target`", nor silently drop optionality on the `up to one` group.

Verified locally: `cargo fmt --all` clean; `cargo clippy -p engine --all-targets --features proptest -- -D warnings` exit 0; 5/5 tests pass; the count assertions fail on revert of the source change.

## AI-contributor disclosure

Authored with an LLM (Claude). Parser-only change; no runtime/game-state logic is touched, so behavior flows through the existing `multi_target` target-selection path. Combinator-purity Gate A: no new string matching — the change binds an already-parsed combinator result instead of discarding it, plus one lowering arm.

Every CR number cited was verified against `docs/MagicCompRules.txt` before use (115.1d at line 847, 601.2c at line 2461). The first draft cited 115.1d alone and was corrected to the established pair, because 115.1d covers *whether* an ability is targeted, not the variable *count*, which is 601.2c.

Two review-driven corrections since opening: the lowering arm was destructured by value to drop an unnecessary `clone()` and a wildcard match arm (per @gemini-code-assist), while still routing through `lower_targeted_action_ast` so `Effect::GainControl` is not duplicated across two sites; and the CI parse-diff revealed the class is 7 cards spanning `GiveControl` too, not the 2 originally claimed — the summary and tests above were expanded to match rather than left understated. Happy to rework anything.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/imperative.rs
+++ b/crates/engine/src/parser/oracle_effect/imperative.rs
@@ -2153,11 +2153,20 @@ pub(super) fn parse_targeted_action_ast(
             .parse(input)
         })
         .is_some();
-        let (target_text, _) = super::strip_optional_target_prefix(rest);
+        // CR 115.1d + CR 601.2c: Preserve the "up to N target" count instead of discarding
+        // it. `strip_optional_target_prefix` already parses the quantifier; the
+        // spec was previously dropped here, so "gain control of up to two target
+        // creatures" collapsed to a single optional slot and only one creature
+        // could be chosen (issue #6205).
+        let (target_text, multi_target) = super::strip_optional_target_prefix(rest);
         let (target, _rem) = parse_target_with_ctx(target_text, ctx);
         #[cfg(debug_assertions)]
         assert_no_compound_remainder(_rem, text);
-        return Some(TargetedImperativeAst::GainControl { target, all });
+        return Some(TargetedImperativeAst::GainControl {
+            target,
+            all,
+            multi_target,
+        });
     }
     // Earthbend: "[you ]earthbend [N] [target <type>]". The optional "you "
     // subject (Fatal Fissure's delayed trigger "you earthbend 4") names the
@@ -2380,7 +2389,7 @@ pub(super) fn lower_targeted_action_ast(ast: TargetedImperativeAst) -> Effect {
             target,
             subject: TargetFilter::SelfRef,
         },
-        TargetedImperativeAst::GainControl { target, all } => {
+        TargetedImperativeAst::GainControl { target, all, .. } => {
             if all {
                 Effect::GainControlAll { target }
             } else {
@@ -11399,6 +11408,34 @@ pub(super) fn lower_imperative_family_ast(ast: ImperativeFamilyAst) -> ParsedEff
             };
             let mut clause = parsed_clause(lower_targeted_action_ast(ast));
             clause.multi_target = multi_target;
+            clause
+        }
+        // CR 115.1d + CR 601.2c: "gain control of up to N target …" (The Super
+        // Hero Civil War, Jace, Ingenious Mind-Mage). Same shape as the
+        // Tap/Untap arm above: the bare-Effect lowering cannot carry a target
+        // count, so a GainControl AST that captured one is lowered here and the
+        // count is threaded onto the clause. Only the TARGETED form can carry
+        // it — the mass "gain control of all/each" (`all: true`) has no target
+        // slots. Destructured by value (rather than the `ast @ …` + re-match
+        // used above) so the count moves out without a clone and the pattern
+        // stays exhaustive; the variant is rebuilt with `multi_target: None` so
+        // lowering still flows through the single authority
+        // (`lower_targeted_action_ast`) instead of duplicating `Effect::GainControl`.
+        ImperativeFamilyAst::Structured(ImperativeAst::Targeted(
+            TargetedImperativeAst::GainControl {
+                target,
+                all: false,
+                multi_target: Some(multi_target),
+            },
+        )) => {
+            let mut clause = parsed_clause(lower_targeted_action_ast(
+                TargetedImperativeAst::GainControl {
+                    target,
+                    all: false,
+                    multi_target: None,
+                },
+            ));
+            clause.multi_target = Some(multi_target);
             clause
         }
         // CR 115.1d + CR 601.2c: "return up to N cards from your graveyard to
--- a/crates/engine/src/parser/oracle_effect/mod.rs
+++ b/crates/engine/src/parser/oracle_effect/mod.rs
@@ -14644,7 +14644,12 @@ fn try_parse_verb_and_target<'a>(
         let all = alt((tag::<_, _, OracleError<'_>>("all "), tag("each ")))
             .parse(rest_lower.as_str())
             .is_ok();
-        let (target_text, _) = strip_optional_target_prefix(rest);
+        // CR 115.1d + CR 601.2c: preserve "up to N target …" here too — this compound/rider
+        // path (saga chapters carrying a trailing duration, e.g. The Super Hero
+        // Civil War) is the second `gain control of` entry point, and dropping
+        // the count here left issue #6205 half-fixed. Mirrors the `fight ` arm
+        // directly above.
+        let (target_text, multi_target) = strip_optional_target_prefix(rest);
         let (target, rem) = parse_target_with_ctx(target_text, ctx);
         let rem_lower = rem.to_ascii_lowercase();
         if tag::<_, _, OracleError<'_>>(" during that player's next turn")
@@ -14676,7 +14681,14 @@ fn try_parse_verb_and_target<'a>(
                 rem,
             ));
         }
-        return Some((TargetedImperativeAst::GainControl { target, all }, rem));
+        return Some((
+            TargetedImperativeAst::GainControl {
+                target,
+                all,
+                multi_target,
+            },
+            rem,
+        ));
     }
     // Earthbend: "earthbend [N] [target <type>]"
     if let Ok((rest, _)) = tag::<_, _, OracleError<'_>>("earthbend ").parse(lower) {
--- a/crates/engine/src/parser/oracle_ir/ast.rs
+++ b/crates/engine/src/parser/oracle_ir/ast.rs
@@ -1037,6 +1037,15 @@ pub(crate) enum TargetedImperativeAst {
         /// True for the untargeted mass form ("gain control of all/each …"),
         /// lowered to `Effect::GainControlAll`; false for targeted GainControl.
         all: bool,
+        /// CR 115.1d + CR 601.2c: Variable target count for "gain control of up to N target
+        /// …" (The Super Hero Civil War's "up to two target creatures with total
+        /// mana value 6 or less"; Jace, Ingenious Mind-Mage's "up to three target
+        /// creatures"). `None` for the common single-target "gain control of
+        /// target creature". Carried onto `ParsedEffectClause.multi_target` at
+        /// lowering so the targeting system surfaces N optional slots — without
+        /// it the count collapsed to one and only a single creature could be
+        /// chosen (issue #6205). Mirrors the same field on `Tap`/`Untap`.
+        multi_target: Option<MultiTargetSpec>,
     },
     ControlNextTurn {
         target: TargetFilter,
--- a/crates/engine/tests/integration/gain_control_multi_target_6205.rs
+++ b/crates/engine/tests/integration/gain_control_multi_target_6205.rs
@@ -15,7 +15,9 @@
 //!     creatures" parsed no count at all.
 //!   * Domineering Will — "target player gains control of up to three target
 //!     nonattacking creatures"; `GiveControl` shares these parse paths, so it
-//!     was capped at one as well.
+//!     was capped at one as well. Parse-level only: the give-control path does
+//!     not surface creature target slots at runtime for an unrelated reason,
+//!     documented with the runtime section at the bottom of this file.
 //!
 //! The remaining four ("up to one target": Pyreswipe Hawk, Rangers of Ithilien,
 //! Scroll of Isildur, Jon Irenicus) already selected correctly and only change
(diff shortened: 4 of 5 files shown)


minion1227:
Thanks @gemini-code-assist — the wildcard arm and the `clone()` were both real warts and are fixed in `0d14845`.

I took the destructuring but not the inlined `Effect::GainControl { target }`: building the effect directly in this arm would duplicate lowering across two sites, so if `lower_targeted_action_ast`'s `GainControl` arm ever changes, this arm would silently diverge. Instead the variant is destructured by value and rebuilt with `multi_target: None`, so the count moves out without a clone, the match stays exhaustive, and lowering still flows through the single authority:

```rust
ImperativeFamilyAst::Structured(ImperativeAst::Targeted(
    TargetedImperativeAst::GainControl { target, all: false, multi_target: Some(multi_target) },
)) => {
    let mut clause = parsed_clause(lower_targeted_action_ast(
        TargetedImperativeAst::GainControl { target, all: false, multi_target: None },
    ));
    clause.multi_target = Some(multi_target);
    clause
}
```

Reviewing that region also surfaced a defect on my side that I have fixed: my arm had been inserted between the existing Tap/Untap doc comment and its arm, orphaning that comment. The new arm now sits below the Tap/Untap arm and comment adjacency is restored.

Separately, the CI parse-diff showed the class is larger than I claimed: **7 cards**, including `GiveControl`. Domineering Will ("up to three target nonattacking creatures") was a third genuinely-capped card, and four "up to one" cards change representation only. I have added `give_control_keeps_up_to_three` and `up_to_one_stays_optional_with_an_explicit_upper_bound` (asserting `min = 0` survives) and corrected the PR body, which had understated the blast radius. 5/5 tests pass; clippy `-D warnings` is clean.

matthewevans (changes requested):
**Request changes — add one production target-selection regression before this parser fix lands.**

## 🔴 Blocker

**[MED]** The new tests only call `parse_oracle_text` and inspect `ParsedEffectClause.multi_target` (`crates/engine/tests/integration/gain_control_multi_target_6205.rs:32-75`); they never enter the `WaitingFor`/`GameAction::ChooseTarget` pipeline that consumes the clause populated at `crates/engine/src/parser/oracle_effect/imperative.rs:11424-11439`. Why it matters: the claimed defect is that a player can select only one target, so AST-shape assertions cannot prove the production target slots are created and accepted. Suggested fix: add a `GameScenario` regression that casts or resolves a representative “up to two/three target” gain-control ability, selects distinct legal targets through the normal action flow, and asserts all selected creatures change controller; retain the existing parser-class tests.

## ✅ Clean

The updated head resolves Gemini’s stale wildcard/clone finding, keeps lowering through `lower_targeted_action_ast`, and the full parse-diff’s seven-card set matches the corrected scope claim.

**Recommendation: request-changes — add the runtime discriminator, then rerun the current-head review.**

minion1227:
Runtime discriminator added in `a26d033` (kept as a separate commit so your review anchors stay valid). The parser-class tests are retained as you asked.

## The runtime regression

`jace_ultimate_steals_three_creatures_through_target_selection` activates the loyalty ability with three distinct opponent creatures, answers one `WaitingFor::TargetSelection` slot per `GameAction::ChooseTarget` (CR 601.2c), and asserts all three actually changed controller (CR 613.1b). The cast driver consumes one declared object per slot, so all three only land if all three slots exist — revert the source fix and slots two and three are never created, leaving creatures two and three with P1.

6/6 pass; `clippy -p engine --all-targets --features proptest -D warnings` exit 0.

## Your gate caught two real things — thank you

I first wrote the `GiveControl` witness against Domineering Will and it failed with all three creatures unmoved. Rather than assume, I measured the same ability's slots with and without `multi_target`:

```
WITH multi_target (this PR):  3 slots — all creature slots, NO player slot
WITHOUT (pre-fix):            1 slot  — creature slot,      NO player slot
```

That isolates two pre-existing gaps, both independent of this PR:

1. **`GiveControl` never gets a recipient slot.** `ability_needs_companion_target_player_slot` defers to `effect_references_target_player`, which inspects only the effect's target *filter* (`effect_bound_filter_matches`) and never `GiveControl`'s separate `recipient` field. So the recipient never resolves and `resolve_give` installs no `ChangeController` at all — `transient_continuous_effects` stays empty. Any "target player gains control of …" card is inert at runtime today.

2. **`"nonattacking"` parses to `Subtype("Attacking")`.** No such creature subtype exists, so Domineering Will additionally has zero legal creature targets.

The comparison above also confirms the count this PR restores does reach real slot generation (1 → 3), and that neither gap is introduced here.

So there is deliberately **no** `GiveControl` runtime witness: a passing one is not currently constructible, and a failing one would fail for an unrelated defect. Its count stays pinned at parse level, with the measurement recorded in the test file. I have not smuggled a fix for either gap into a parser count fix — happy to open follow-up issues for both, or to fix the recipient slot here if you would rather it land together.

One correction to my own PR body while I am here: I originally listed Domineering Will as "fixed". Accurate statement is that its parsed count is now correct, but the card remains non-functional at runtime for gap 1 above. The body has been updated.

minion1227:
Blocker addressed in `a26d033` (+ a comment correction in `32cf1bf`).

## The runtime discriminator

`jace_ultimate_steals_three_creatures_through_target_selection` drives Jace, Ingenious Mind-Mage's verbatim `−9: Gain control of up to three target creatures.` through the real pipeline: a loyalty activation (CR 606.1 + CR 602.2b), three distinct opponent creatures declared, one `WaitingFor::TargetSelection` slot answered per `GameAction::ChooseTarget` (CR 601.2c), then asserts all three changed controller (CR 613.1b). The driver consumes one declared object per slot, so all three land only if all three slots exist — no AST field is asserted.

Fail-on-revert is measured, not claimed. Reverting the source arm to the discarded binding and re-running gives:

```
got [PlayerId(0), PlayerId(1), PlayerId(1)]
```

Exactly one creature changes controller and two stay put — the reported defect reproduced through production target selection. The parser-class tests are retained, and `single_target_gain_control_declares_no_count` still passes under the revert, so the witness is discriminating rather than blanket-failing.

## No `GiveControl` runtime witness — and why

I tried to add the symmetric give-control witness and could not make it honest, so it is documented instead of shipped. Building the target slots for `"target player gains control of up to three target creatures"` with and without `multi_target` gives **3 creature slots vs 1** — so the count this PR restores does reach real slot generation — but **no player slot in either case**. `ability_needs_companion_target_player_slot` defers to `effect_references_target_player`, which inspects the effect's bound target filter and never `GiveControl`'s separate `recipient` field, so the recipient never resolves and `resolve_give` installs no `ChangeController`.

That gap is pre-existing and independent of this fix. It is also not a one-line arm: `GiveControl`'s recipient is often *not* a chosen target ("an opponent gains control of that creature", `ScopedPlayer` / `SpecificPlayer`, all resolved at resolution time via `unique_recipient_from_filter`), so firing the companion-slot check on any `GiveControl` would manufacture a spurious player slot for that whole class — the same hazard the `Sacrifice` / `Bounce { AtResolution }` / `UnattachAll` carve-outs exist to prevent. Doing it right means discriminating a targeted recipient from a resolution-time one, which is a design change I did not want to smuggle into a parser count fix. Happy to file it as a follow-up issue, or to fold it in here if you'd rather this PR carry it.

So `GiveControl`'s count stays pinned at parse level (`give_control_keeps_up_to_three`), with the reasoning recorded in the test file rather than papered over by a test that would pass or fail for an unrelated reason.

Two corrections to my own earlier claims, for the record:
- The PR body's Jace Oracle text was wrong — I had `"until end of turn. Untap those creatures. They gain haste until end of turn."` appended, which is not on the card. The real line is `−9: Gain control of up to three target creatures.`, and both the parser test and the runtime witness now use it verbatim.
- Domineering Will is fixed **at parse level only**, not end-to-end, for the recipient reason above. The original PR body implied otherwise; the file header now states the limit.

Verification: `cargo fmt --all` clean; `cargo clippy -p engine --all-targets --features proptest -D warnings` clean; full `cargo test -p engine` green (17308 + 3591 passing, 0 failures).

matthewevans (approved):
Approved: the GainControl target-count is threaded through both parser entry paths to the existing clause-level multi-target authority, and the Jace regression proves three selected targets resolve through the production pipeline.

matthewevans:
Maintainer follow-up: I reviewed the current head `32cf1bfa6406bd1622afe34c568568750617d4b1`. The new Jace scenario is a real production target-selection regression: it selects three distinct creatures through `AbilityActivation` and asserts all three controller changes. That addresses the prior runtime-test request.

This PR is now held only for the current-head Rust required check and a refreshed `<!-- coverage-parse-diff -->` artifact. The existing parse-diff comment predates this head, so I will not approve or enqueue until both current-head evidence items are available. No contributor action is requested at this point.

PR #6235, opened 2026-07-20
fix(client): avoid duplicate card data on iOS

parthmishra:
## Summary

- keep saved-game restoration on the shared engine worker instead of loading a second main-thread WASM/card corpus
- delay `CardImage` fallback data lookup until artwork definitively fails
- route card face, parse-detail, and ruling queries through the shared worker
- add memory-ownership guardrails and focused regression tests

## Root cause

`GameProvider` loaded the full card database through the main-thread runtime before `WasmAdapter` restored via the worker. At the same time, every initial `CardImage` mount queried that main-thread runtime while artwork was still resolving. On memory-constrained iOS WebKit, this duplicated the WASM module and full card corpus and could terminate the WebContent process around game startup or mulligan.

These paths were present on upstream `main` before the hand-preview work and are independent of it.

## User impact

Mobile Safari and the installed PWA can enter games and pass mulligan without falling into the “A problem repeatedly occurred” reload loop. Desktop behavior is unchanged.

## Validation

- physical iPhone manual verification: startup crash no longer reproduced
- TypeScript build (`tsc -b --noEmit`)
- ESLint (0 errors; 18 pre-existing warnings)
- 44 focused Vitest tests
- production PWA build
- full pre-push Rust, parser, AI, card-data, coverage, lint, and type-check gates

### DIFF
--- a/client/src/adapter/__tests__/boundary-guardrails.test.ts
+++ b/client/src/adapter/__tests__/boundary-guardrails.test.ts
@@ -71,6 +71,31 @@ describe("adapter boundary guardrails", () => {
     }
   });
 
+  it("keeps gameplay card-data ownership in the shared engine worker", () => {
+    const root = repoRoot();
+    const gameProvider = readFileSync(
+      resolve(root, "client/src/providers/GameProvider.tsx"),
+      "utf8",
+    );
+
+    // The shared WasmAdapter loads the corpus before game creation/restoration.
+    // Importing the main-thread card-data service here would allocate a second
+    // WASM module and full corpus whenever gameplay mounts or resumes.
+    expect(gameProvider).not.toMatch(/from ["']\.\.\/services\/cardData["']/);
+
+    const cardDataHook = readFileSync(
+      resolve(root, "client/src/hooks/useEngineCardData.ts"),
+      "utf8",
+    );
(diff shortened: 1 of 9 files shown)


matthewevans:

This current head (`c739f7e4e75e4555441ad9e11cfcc8246ab482d7`) was triaged as a frontend-only change (`client/src/adapter/__tests__/boundary-guardrails.test.ts`, `client/src/adapter/__tests__/wasm-adapter.test.ts`, `client/src/adapter/engine-worker-client.ts`, `client/src/adapter/engine-worker.ts`, `client/src/adapter/wasm-adapter.ts`, `client/src/components/card/CardImage.tsx`, `client/src/components/card/__tests__/CardImage.test.tsx`, `client/src/hooks/useEngineCardData.ts`, `client/src/providers/GameProvider.tsx`) by `parthmishra`. The local frontend-review allowlist does not include this author, so this route does not perform an implementation-diff review or approve the PR.

matthewevans (approved):
Reviewed against current `origin/main`. The shared adapter now owns initialization and card-DB loading for all three display queries; the resume path still loads the DB before rehydration, and `CardImage` defers the fallback lookup until art resolution has actually failed. No blocking findings.

PR #6237, opened 2026-07-20
fix(card-data): weekly MTGJSON refresh + vintage gate; version-keyed CI caches

matthewevans:
Local card-data generation ran forever against a frozen MTGJSON cache
(downloads were only-if-absent), regenerating stale catalogs that fought
git for crates/engine/data/known-tokens.toml and oracle-subtypes.json —
the reported Tilt infinite rebuild loop.

- gen-card-data.sh: refresh the MTGJSON cache when the ISO-week stamp
  (data/mtgjson/.refresh-week) is stale or PHASE_REFRESH_MTGJSON=1;
  MTGJSON_SKIP_REFRESH=1 is the offline hatch. All five download gates
  honor the refresh flag; decks re-download on refresh.
- Vintage gate: crates/engine/data/mtgjson-vintage records the newest
  .meta.date ever promoted. Older inputs may still generate untracked
  outputs but are refused write access to tracked catalogs, so a stale
  cache can never overwrite newer committed data again. Stamp advances
  monotonically (tmp+cmp+promote); malformed .meta.date hard-fails.
- CI: deploy/release mtgjson-full caches now key on the published
  .meta.version (week-based fallback) via a shared composite action,
  closing the Monday race where an ISO-week key seeded before MTGJSON's
  ~06:00 UTC publish pinned the whole week to stale data (observed
  2026-07-20: prod served the 07-12 release all morning).


matthewevans:
The flagged `permissions: contents: write` top-level block pre-exists this PR on `origin/main` (this PR only touches the MTGJSON cache-key steps in this file — the permissions block is unchanged). Scoping permissions per-job is a valid hardening follow-up but belongs in its own change against the release workflow.

PR #4712, opened 2026-07-01
feat(client): show unimplemented-mechanics badge on stack spells

jsdevninja:
## Summary

- Extract shared `UnimplementedMechanicsBadge` from `CardImage` for reuse across card surfaces
- Show the amber `!` warning on stack spells when `unimplemented_mechanics` is non-empty
- Add unit test verifying badge visibility on `StackEntry`

Closes phase-rs/phase#4711

## Test plan

- [x] `pnpm --dir client exec vitest run src/components/stack/__tests__/StackEntry.test.tsx` — 2/2 passed
- [ ] Manual: cast a card with partial support and confirm the badge appears on the stack before resolution

PR #6243, opened 2026-07-20
feat(client): show the unimplemented-mechanics badge on stack spells

### LINKED ISSUE #4711: Show unimplemented-mechanics warning badge on stack spells
## Problem

Hand cards and battlefield permanents already show an amber ! badge when unimplemented_mechanics is non-empty. Stack entries skip this badge.

## Proposed solution

Reuse the same badge on StackEntry; extract shared UnimplementedMechanicsBadge from CardImage.

## Scope

Display-layer only — no engine changes.

minion1227:
## Summary

Hand cards and battlefield permanents already show the amber `!` when the engine reports `unimplemented_mechanics`. Stack entries skipped it — and that is the surface where it matters most, since a spell on the stack is about to resolve and is the last moment a player can respond to it.

Extracts the badge out of `CardImage` into a shared `UnimplementedMechanicsBadge` and renders it on both surfaces, so the two cannot drift apart.

Closes #4711

## Screenshot

Both entries are the real `StackEntry` component. Left: a spell whose printed abilities include unimplemented mechanics (`cascade`, `storm`) — amber `!` at bottom-right. Right: a fully-supported spell — no badge.

unimplemented-mechanics badge on stack spells

Rendered in headless Chrome against a real `gameStore` state. Disclosure: the harness stubs `@wasm/engine`, because building the WASM bundle needs `wasm-pack`, which I could not install in a network-restricted sandbox. That is why the **card art is blank** — every other element (badge, yield pill, "Next", borders, layout) is the genuine component output. Happy to redo this as a live in-game capture if you want one; I just can't produce it from here.

## The screenshot caught a real placement bug

Worth recording, because it justifies the screenshot requirement. My first pass put the badge **bottom-left**, having checked that top-left (coalesce `×N`, chosen `X=`) and top-right (`Casting…` / `Next`) were occupied. Rendering it showed the badge **overlapping the auto-pass yield pill**, which also lives bottom-left.

Bottom-right is the stack entry's only genuinely free corner. No DOM assertion would have caught the overlap — both elements render, both are in the document, and both pass every query. Only pixels show it.

## Fix

- **New** `client/src/components/card/UnimplementedMechanicsBadge.tsx`.
  - Owns the emptiness guard: renders `null` for an absent field *or* an empty array, so no call site repeats `mechanics.length > 0`. That duplicated guard across surfaces is what this issue exists to remove.
  - Takes a typed `variant` (`"overlay"` inside card art, `"corner"` hung off a stack entry) rather than a raw `className`, so placement is a closed set rather than stringly-typed styling.
- `CardImage.tsx` — inline span replaced by the component; visual output unchanged (`overlay` is the default variant and reproduces the previous classes).
- `StackEntry.tsx` — renders the badge from `sourceObj?.unimplemented_mechanics`.

Display-only, as scoped in the issue. `unimplemented_mechanics` is an engine-computed projection forwarded verbatim; nothing here derives, filters, or re-interprets it, and there are no engine changes.

## Anchored on

`CardArtFallback` in the same directory — the component I extracted from `CardImage` in #6214 for the same reason (one shared surface instead of per-caller copies). This follows its conventions deliberately: exported props interface, doc comments on the props, and a named `variant` union for render scale rather than passing classes down.

## Test plan

`client/src/components/card/__tests__/UnimplementedMechanicsBadge.test.tsx` — building-block coverage across the component's whole input range, not a single call site:

- renders nothing when the field is absent, and nothing for an empty list (the engine omits the projection for supported cards, but serde can also round-trip `[]`)
- renders when at least one mechanic is unimplemented
- names every mechanic in `title` and the accessible name, comma-joined — the glyph is a bare `!`, so that label is the only place the detail is legible
- keeps the amber identity across both variants, changing only placement
- defaults to `overlay`, pinning that existing card surfaces are unchanged

`client/src/components/stack/__tests__/StackEntry.test.tsx` — two paired cases: a partially-supported spell shows the warning; a fully-supported one does not. The second is the discriminating guard — without it the first would still pass if the badge rendered unconditionally.

Verified locally: **2116 tests pass** (245 files, full client suite), `tsc -b --noEmit` exit 0, `eslint` exit 0 on all five changed files.

## Credit

#4712 by @jsdevninja proposed this same extraction and was closed for inactivity after the screenshot request went unanswered. I arrived at the same shape independently from the issue text; the placement fix, typed `variant`, building-block tests, and screenshot are new here.

## AI-contributor disclosure

Authored with an LLM (Claude). Frontend-only; no engine, parser, or rules logic is touched, so no CR annotations apply. The badge consumes an existing engine projection and adds no derived state to the client, per the display-layer rule in CLAUDE.md.

Unlike my previous frontend PRs (#6214, #6215), I was able to run the frontend suite, type-check, and lint locally this time, so those results are measured rather than deferred to CI. Happy to rework anything.

### DIFF
--- a/client/src/components/card/CardImage.tsx
+++ b/client/src/components/card/CardImage.tsx
@@ -7,6 +7,7 @@ import type { TokenImageRef } from "../../adapter/types.ts";
 import { CARD_BACK_URL } from "../../services/scryfall.ts";
 import { getBevelBorderStyle } from "./cardFrame.ts";
 import { CardArtFallback } from "./CardArtFallback.tsx";
+import { UnimplementedMechanicsBadge } from "./UnimplementedMechanicsBadge.tsx";
 import { ManaSymbol } from "../mana/ManaSymbol.tsx";
 
 interface CardImageProps {
@@ -131,14 +132,7 @@ export function CardImage({
           style={borderStyle ?? { border: "1px solid #4b5563" }}
         />
       )}
-      {unimplementedMechanics && unimplementedMechanics.length > 0 && (
-        <span
-          className="absolute top-0.5 left-0.5 bg-amber-500 text-black text-[8px] font-bold rounded-sm px-0.5 leading-tight"
-          title={t("card.unimplemented", { mechanics: unimplementedMechanics.join(", ") })}
-        >
-          !
-        </span>
-      )}
+      <UnimplementedMechanicsBadge mechanics={unimplementedMechanics} variant="overlay" />
       {tapIndicator && (
         <span
           className="absolute top-1 right-1 flex items-center justify-center rounded-full bg-black/70 p-1 shadow-md ring-1 ring-white/20"
--- a/client/src/components/card/UnimplementedMechanicsBadge.tsx
+++ b/client/src/components/card/UnimplementedMechanicsBadge.tsx
@@ -1,0 +1,61 @@
+import { useTranslation } from "react-i18next";
+
+/**
+ * Render scale for the badge. Mirrors the `artCrop` / `fullCard` variant
+ * vocabulary `CardArtFallback` already uses, so every card-surface component in
+ * this directory names its render modes the same way.
+ *
+ * - `overlay` — pinned inside the art of a hand/battlefield card, where the
+ *   badge shares the corner with the tap indicator and must stay unobtrusive.
+ * - `corner` — hung outside the border of a stack entry, matching the ×N and
+ *   status pills that already sit on that surface.
+ */
+export type UnimplementedMechanicsBadgeVariant = "overlay" | "corner";
+
+export interface UnimplementedMechanicsBadgeProps {
+  /**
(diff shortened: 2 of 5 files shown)


matthewevans:

This current head (`eb1a0fc5e93a3009caf3929fc7f3abaf026f437c`) was triaged as a frontend-only change (`client/src/components/card/CardImage.tsx`, `client/src/components/card/UnimplementedMechanicsBadge.tsx`, `client/src/components/card/__tests__/UnimplementedMechanicsBadge.test.tsx`, `client/src/components/stack/StackEntry.tsx`, `client/src/components/stack/__tests__/StackEntry.test.tsx`) by `minion1227`. The local frontend-review allowlist does not include this author, so this route does not perform an implementation-diff review or approve the PR.

matthewevans (approved):
Maintainer review complete: the shared engine-projection badge uses the display-layer seam and has discriminating stack coverage.

PR #6249, opened 2026-07-21
test(engine): guard Edgar Markov Eminence command-zone trigger (#817)

### LINKED ISSUE #817: [Card Bug] Edgar Markov Eminence ability not triggering
### Card(s)

Edgar Markov

### Build/version

_No response_

### Game mode

AI

### Actual behavior

Casting a vampire spell does not trigger the Eminence ability on Edgar

### Expected behavior

Casting a vampire spell should trigger eminence and create a 1/1 vampire token

### Steps to reproduce

Play commander with Edgar Markov as your commander and cast any vampire spell

_No response_

mike-theDude:
## Summary

Adds a discriminating runtime regression for #817. Edgar Markov's Eminence trigger was reported as not firing from the command zone. Investigation shows the fix already landed via `20d4abc52 Fix Eminence command-zone triggers` (the parser now collects `Command` + `Battlefield` from the trigger's `Or` intervening-if into `trigger_zones`), but the only existing Eminence coverage (`the_ur_dragon_eminence.rs`) hand-builds the cost-reduction **static** and never exercises the parser → trigger → resolution path. So the reported trigger bug had **zero runtime guard** and could silently regress.

This test drives the real parser (`from_oracle_text`) + full cast pipeline:

- **Positive:** casting another Vampire spell with Edgar Markov in the command zone creates exactly one 1/1 Vampire token.
- **Negative control:** a non-Vampire spell creates no token (guards against an over-broad command-zone trigger scan).

Both pass on current `main`, confirming #817 is fixed-unreleased.

## CR annotations (grep-verified against `docs/MagicCompRules.txt`)

- CR 207.2c — "Eminence" is an ability word with no independent rules meaning.
- CR 113.6b — an ability that mentions a zone functions only from that zone; the parsed trigger's `trigger_zones` must include `Command`.
- CR 603.4 — intervening-"if" is checked at trigger time and at resolution.
- CR 408 — the command zone.

## Verification

```
test edgar_eminence_triggers_another_vampire_from_command_zone ... ok
test edgar_eminence_ignores_non_vampire_spell_from_command_zone ... ok
```

Closes #817.

https://claude.ai/code/session_01QL7TL2ECn25ccYLahLssxV


matthewevans (approved):
Approved after a maintainer fixup restored rustfmt and made the regression use Edgar Markov's exact shortened Oracle self-reference.

PR #6252, opened 2026-07-21
feat(ai-commander): add --games-file batch mode with panic isolation and stack-size fix

mcbradd:
## Summary

pod-lab's simulation-acceleration plan (Tier 1) spawns `ai-commander` as a fresh OS process per game — measured at ~1000 games/run, each re-parsing the ~97MB card database (~1.8s/game measured setup cost, real but secondary to per-game gameplay-phase contention which is out of scope for this PR). `--games-file <path>` lets one process play a list of `seed,difficulty` games in a loop against the same already-loaded card database and feed, amortizing that load cost across the whole batch.

- `--games-file` points at a file of `seed,difficulty` lines (one game per line). When given, `--seed`/`--difficulty` are ignored; every line plays against the SAME loaded database/feed. Malformed lines hard-fail the whole batch at startup (same discipline as the existing `parse_difficulty`/`parse_action_cap` validators), before the DB load.
- Single-game invocation (no `--games-file`) is **unchanged**: the pre-existing per-game logic was extracted into `play_one_game()` verbatim (it now returns `RunOutcome` instead of calling `std::process::exit` directly) so both the single-game path and the batch loop drive the exact same code. Verified byte-identical stdout for 3 real seeds (see Test plan).

## Why the panic-isolation / flush / stack-size changes are here, not incidental

Running many games in one process collapses the crash isolation that one-process-per-game gets for free from the OS — a panic partway through a batch would otherwise lose every game still queued behind it. These three changes are deliberate correctness work this PR ships together with batching, not scope creep:

- **Panic isolation** (`run_batch_isolated`): each game's play-through is wrapped in `catch_unwind` (with `AssertUnwindSafe`, since `GameState` et al. aren't `UnwindSafe` by default) so one bad game — this binary has real `unwrap()`/`expect()` calls — can't take down the rest of the batch. Prints a machine-readable `GAME <seed> PANICKED: <message>` line and continues. The panic-reporting path itself is nested inside the isolation boundary too: `println!`/`flush` can panic on a genuine I/O failure (e.g. a broken stdout pipe under an external process harness), and that must not defeat the isolation either. A quiet panic hook suppresses the default noisy stderr dump for the duration of the batch only (restored afterward).
- **Immediate per-game flush**: each game's result (or panic line) is flushed to stdout the moment it finishes, not buffered until the batch completes — so a process killed mid-batch (pod-lab enforces an external wall-clock timeout on the whole process) never loses an already-completed game's result.
- **Stack-size fix**: `main()` now only parses CLI args; the actual game-driving work (both the pre-existing single-game path and the new batch path) runs on a spawned thread with a 32MiB stack — mirroring `duel_suite::run`'s identical pre-existing fix for the same reason (deep AI search recursion overflows the platform default) — joined from `main()` with the exit code propagated correctly (a single-game-mode panic now correctly still exits non-zero, via the `join` `Err` path mapping to exit 101). **This was not theoretical**: every seed tried this session overflowed the plain main thread's default stack immediately after game start, even at Easy difficulty, under this crate's normal release profile (`opt-level = 'z'`). Batching increases cumulative per-process execution, making this more likely to matter, not less, even though no single game's peak recursion depth changes.

## Test plan

- [x] `cargo test --release --bin ai-commander -p phase-ai` — 17 tests pass, including new coverage for `parse_games_file`/`parse_games_file_line` (valid/malformed lines, blank-line skipping, whole-batch-fails-on-any-malformed-line) and `run_batch_isolated` (visits every game in order; a panic on a middle game does not prevent later games from running or being recorded — a `#[cfg(test)]`-only fake `play` closure proves the loop mechanics without needing to drive a real game).
- [x] `cargo build --release --bin ai-commander -p phase-ai` and `cargo clippy --release --bin ai-commander -p phase-ai -- -D warnings` — clean.
- [x] Result-equivalence check: ran 3 real seeds (90001/90002/90003, Easy) individually via `--seed N --difficulty Easy` and together via `--games-file`. Diffed stdout (filtering `elapsed`-containing lines, which are wall-clock and expected to differ) — byte-identical, including winner, total actions, turns played, and every player's final life/hand/library/graveyard/battlefield line.
- [x] Confirmed a malformed `--games-file` line hard-fails at startup (before the DB load) with a clear error and exit code 1.
- [x] Diff scoped to `crates/phase-ai/src/bin/ai_commander.rs` only.

### DIFF
--- a/crates/phase-ai/src/bin/ai_commander.rs
+++ b/crates/phase-ai/src/bin/ai_commander.rs
@@ -10,7 +10,20 @@
 //!   cargo run --release --bin ai-commander -- client/public --seed 7 --difficulty Easy
 //!   cargo run --release --bin ai-commander -- client/public --difficulty Easy \
 //!       --difficulty-p2 VeryHard --action-cap 50000
+//!
+//! Batch mode (pod-lab simulation-acceleration plan, Tier 1 item 1): play many
+//! games in one process instead of one process per game, amortizing the card
+//! database load (~1.8s/game measured) across the whole batch. `--games-file`
+//! points at a file of `seed,difficulty` lines (one game per line); when it is
+//! given, `--seed`/`--difficulty` are ignored and every line is played in turn
+//! against the SAME loaded database and feed. Each game's play-through is
+//! panic-isolated (`run_batch_isolated`) and its result is flushed to stdout
+//! immediately, so a batch survives an individual game panicking or the whole
+//! process being killed mid-batch (pod-lab enforces an external wall-clock
+//! timeout on the process).
+//!   cargo run --release --bin ai-commander -- client/public --games-file games.txt
 
+use std::any::Any;
 use std::collections::{HashMap, HashSet};
 use std::io::Write as _;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


matthewevans (approved):
Approved: the batch-mode review and maintainer hardening are complete on the current head; the merge queue will wait for the fresh required CI run.

matthewevans (approved):
Approved after maintainer conflict resolution: current head preserves batch-mode game reuse and the all-card-names setup authority; existing regression coverage for both paths remains in place.

matthewevans (approved):
Approved on the current maintainer-fixup head: GameRunContext resolves the CI argument-limit regression without changing game setup, batch isolation, or the existing regression coverage.

PR #6256, opened 2026-07-21
fix(engine): resolve Kellan, the Kid hand-pick cast during resolution…

### LINKED ISSUE #5945: Kellan the kid errors saying ability doesn't exist — [[Kellan the kid]]
Reported in Discord: https://discord.com/channels/1485498006781427802/1525476267594485840/1525476267594485840

**Thread:** Kellan the kid errors saying ability doesn't exist
**Discord thread id:** `1525476267594485840`
**Discord message id:** `1525476267594485840`
**Cards:** kellan, the kid
**Extraction confidence:** 0.40

## Summary
[[Kellan the kid]]

## Reported text

[[Kellan the kid]]

## Attachments

- [Screenshot_20260711_071705_Chrome.jpg](https://cdn.discordapp.com/attachments/1525476267594485840/1525476267715989604/Screenshot_20260711_071705_Chrome.jpg?ex=6a5974bc&is=6a58233c&hm=21440ec9ef9a0ce4c60cf8467abf0b07985affd367dbd4b0046bc6c445539b9f&)

### kellan, the kid
Flying, lifelink
Whenever you cast a spell from anywhere other than your hand, you may cast a permanent spell with equal or lesser mana value from your hand without paying its mana cost. If you don't, you may put a land card from your hand onto the battlefield.

---
<sub>report_id: `discord:1525476267594485840:1525476267594485840:0` · discord: `1525476267594485840/1525476267594485840`</sub>

andriypolanski:
Closes #5945

## Summary

Kellan, the Kid errored with "ability doesn't exist" when its trigger was accepted. The optional hand-pick `CastFromZone` self-installs a resolution-time `EffectZoneChoice`, but the generic sub-stash in `resolve_chain_body` prepended the "If you don't, put a land" sub ahead of that continuation, so the resume fed the wrong ability into `complete_hand_pick_cast_from_zone`.

Fixes, all built for the class (any "cast a permanent spell from your hand" grant with a `Not(OptionalEffectPerformed)` land-drop fallback), not the one card:

- Skip the generic sub-stash for a hand-pick `CastFromZone` that installed its own continuation (CR 608.2c/608.2g).
- Freeze the "equal or lesser mana value" gate to a `Fixed` when the pick is opened, while the trigger event is live; the finalize-time re-evaluation on the selection resume no longer needs the (by-then cleared) trigger context (CR 202.3 + CR 608.2h).
- Exclude lands from a Cast-mode hand pick — a land is played, not cast (CR 601.2 / CR 305.1).
- An empty selection ("If you don't") or an empty eligible pool routes the land-drop fallback through the decline authority with `optional_effect_performed` reset, and the consumed `Not(OptionalEffectPerformed)` gate is stripped so the fallback's own optional accept doesn't re-trip it (CR 608.2c/608.2d).

## Verification

- [ ] cargo nextest run --profile ci --partition count:1/2 --workspace --exclude phase-tauri --exclude mtgish-import --features engine/proptest --status-level fail --final-status-level fail
- [ ] cargo nextest run --profile ci --partition count:2/2 --workspace --exclude phase-tauri --exclude mtgish-import --features engine/proptest --status-level fail --final-status-level fail
- [ ] cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings
- [ ] cargo fmt --all --check


matthewevans (changes requested):
**Blocked — the new decline fallback is not correct for every condition shape it explicitly accepts.**

## 🔴 Blocker

- `crates/engine/src/game/effects/mod.rs:2637` strips the consumed decline gate only when the entire condition is exactly `Not(OptionalEffectPerformed)`. But `should_resolve_subability_on_optional_decline` deliberately accepts `And`/`Or` conditions containing that gate at `crates/engine/src/game/effects/mod.rs:2654-2667`. An optional decline branch such as `And([Not(OptionalEffectPerformed), <independent condition>])` therefore reaches the fallback, then re-enters after accepting its own “may” with the original `Not` still present; the accept flips `optional_effect_performed` to true and suppresses the effect. This is the same failure mode the new helper documents for the simple form, now left in the composite form.

Evidence: the current-head implementation and its existing composite-branch classifier. The missing case is rules-relevant under CR 608.2c: the parent’s decline gate has already selected the branch, while the independent conjunct must still be evaluated for the branch’s own resolution.

Suggested fix: make the consumed-gate transformation structural—remove only the selected `Not(OptionalEffectPerformed)` leaf from an `And` while preserving and re-evaluating sibling predicates; treat an `Or` only with semantics proven by the parser/lowering shape. Add a production-path regression test for an optional compound decline branch plus a sibling-condition negative case.

## ✅ Clean

- The missing-player bot comment is not a confirmed finding: this path is internal game state and a missing ability controller is not a recoverable user-input case. The exact-context and fallback-routing concerns above are the blocker.

**Recommendation: request changes for the compound decline-condition path, then re-review the new head.**

matthewevans (changes requested):
**Blocked — the revised decline-gate transform still accepts an `Or` branch it cannot safely re-enter.**

## 🔴 Blocker

- `crates/engine/src/game/effects/mod.rs:2678-2682` intentionally preserves every `Or`, but `should_resolve_subability_on_optional_decline` at `crates/engine/src/game/effects/mod.rs:2722-2735` selects any `Or` containing a performed gate. An optional fallback with `Or([Not(OptionalEffectPerformed), IsYourTurn])` is selected after the parent declines because the first term is true. If the fallback controller is not the active player, accepting its own may sets `optional_effect_performed = true`; re-entry then evaluates `Or([false, false])` and silently skips the selected effect. The new tests prove only the `And` branch and the positive Armored Kincaller `Or` branch, not this accepted negative-`Or` shape.

Evidence: the current head condition selector and transform above. CR 608.2c requires resolving the selected instruction in order; CR 608.2d makes the fallback may a new choice. Both citations were verified in `docs/MagicCompRules.txt`.

Suggested fix: make selection and transformation use the same proven condition grammar. Either structurally consume the parent-decline condition for all accepted `Not(OptionalEffectPerformed)` forms, preserving only predicates that remain independent after selection, or narrow the selector to the exact lowering shapes with a parser proof. Add a production-path regression for the negative-`Or` case and its sibling truth-value case.

**Recommendation: request changes; re-review the next head.**

matthewevans (approved):
Approved on current head `40500653cd387d0898b65b3e2b637a9a41b05041`.

The decline-gate transform now consumes the selected negative `Or` branch correctly, and the production-path regression covers accepting that branch when its sibling is false. The current parse-diff artifact reports no card-parse changes.

matthewevans (approved):
Approved after the maintainer-caused merge update at `b24dcec652644a8a3c0f9589bef6ffce1942be3e`. The only resolution preserved both adjacent integration-test registrations in sorted order; no Kellan behavior changed.

matthewevans (approved):
Approved after the maintainer-caused update to `main` at `0a2e74f2b2e910ef3153319c5b164466c2c5b3f0`. The merge is conflict-free: the final PR diff remains limited to the reviewed Kellan cast/decline continuation fix and its registered runtime regressions. Existing no-change parse-diff evidence remains scope-consistent; the new CI run will regenerate required verification for this merge head.

PR #6238, opened 2026-07-20
feat(engine): drive & collapse accepted CR 732.2a infinite-loop shortcuts at the phase boundary

lgray:
I've fully playtested this branch end to end and confirmed it correctly **offers CR 732.2a loop shortcuts to the user** in a live 4-player game: the Witherbloom / Sprout Swarm object-growth loop and the Kilo, Apogee Mind + Freed from the Real + Relic of Legends + Pentad Prism counter loop both surface the shortcut prompt and resolve to the correct board state (naming ~500 counters resolves quickly). The next tasks are to **widen the scope of combos that we trigger on and accept**.

:robot: _AI text below_ :robot:

## Summary

Delivers the **acceptance + materialization** half of the CR 732.2a loop-shortcut feature. When a player accepts an offered infinite loop at the phase boundary (the offer / opponent response window landed in PR-7 #5672), the engine now drives the loop to a player-named finite count and collapses the unbounded growth axis — copy-token piles, `+1/+1` (and other) counters, life, and mana — into the correct terminal board state, rendering `∞` on the affected resource. Covers object-growth loops (tapped token piles) and persistent-axis loops (counters / life / mana) with APNAP accept + interruptibility, a zone-gated object-growth census firewall (CR 603.4 / 603.6a / 113.6), and multi-action loop-shortcut driving for infinite-mana combos.

Multi-commit epic (24 commits) developed on the fork; each component was built through `/engine-implementer` (plan → review-engine-plan → implement → review-impl).

## Implementation method (required)

Method: /engine-implementer

## CR references

Core: `CR 732.2a` / `CR 732.2b` / `CR 732.6` (loop shortcuts + response window), `CR 500.4` / `CR 500.5` (phase-boundary auto-advance / SBA), `CR 603.4` / `CR 603.6a` / `CR 113.6` (zone-gated loop-firewall observer scans), `CR 704.5a` / `CR 704.5f` / `CR 704.5j` (mandatory-loop winner / state-based actions), `CR 608.2b`–`CR 608.2d` (per-iteration re-check + predictability). Plus supporting annotations for counters (`CR 122.1`), life (`CR 119`), and keyword/interaction rules — ~89 unique `CR` annotations in the diff, each grepped against `docs/MagicCompRules.txt`.

## Verification

- [x] Required checks ran clean locally, or the exact CI-owned alternative is stated below.
- [x] Gate A output below is for the current committed head.
- [x] Final review-impl below is clean for the current committed head (recent components; note below).
[...]

PR #6259, opened 2026-07-21
fix(engine,client): address review findings on CR 732.2a loop-collapse

lgray:
Play-tested at `247ffb7` (rebased to `b211832`) and combo detection works in live games. Will check again after addressing blockers.
Play-tested at `1ed7cfd` (rebased to `e6ebd63`) and combo detection works in live games with blockers addressed.

## Summary

Follow-up to **#6238** (merged) addressing the 1 MED + 5 LOW findings from the maintainer's independent `review-impl` on the CR 732.2a loop-collapse feature: a defense-in-depth boundary-mint pause guard, the declined-axis `∞` lifecycle documentation + regression test, the single-fodder invariant assertion + unit test, an accept-time production-routing E2E test, and two frontend display fixes (the `∞` counter tooltip and an i18next plural). No behavior change to the certified/offer path; the MED guard is unreachable today (firewall-gated) and the rest are coverage/docs/display.

Method: /engine-implementer

(plan → review-engine-plan [2 rounds to clean] → implement → review-impl [clean, every finding's revert-flip observed]. Two findings changed fix after code-probing: LOW-1 both reviewer options proven wrong/impossible → document + test; LOW-2 k>1 proven structurally impossible → assert + unit test, not capture-and-multiply.)

## CR references

`CR 732.2a` (offer firewall / no-conditional-actions), `CR 732.2b` (loop shortcut never forced — declined-axis capability marker), `CR 707.2` (token-copy surface identity), `CR 616.1` (2+ materially-ordered replacements → controller orders; the MED test's pausing construction), `CR 111.10` (predefined-token creation), `CR 605.3b` (off-stack mana ability — the LOW-3 test's loop-seeding activation class). All grepped against `docs/MagicCompRules.txt`.

## Verification

- [x] Required checks ran clean locally (Tilt does not watch this worktree; run directly), post-rebase on the current head.
- [x] Final review-impl below is clean (the reviewed diff is byte-identical to this head except the #6254 import union, re-verified green post-rebase).

Rebased onto `upstream/main` (`6817abeb4`, which now contains #6238's squash `549582524`) — one content conflict in `combo_infinite_pile.rs` resolved as a union import (`mark_layers_full` from #6254 + `GameScenario` from this PR). Measured post-rebase on head `247ffb761`:

- `cargo test -p engine --test integration combo_infinite_pile` — **21 passed / 0 failed**
- `cargo test -p engine --lib derived_fodder_class` — **1 passed** (the LOW-2 single-new-object gate)
- `cargo check --workspace` — **clean** (Finished, 2m26s)
- `cargo clippy -p engine --all-targets` — **clean, 0 warnings / 0 errors** (16m06s)
- `pnpm type-check` — **exit 0** · `pnpm test` — **2138 passed / 246 files** (18 todo, 4 skipped; i18n locale parity green)
- `cargo fmt --all --check` — **clean**

Each new test carries an observed revert-flip: removing the MED guard fails the "preserve the Tokens `∞` axis" assertion; forcing the LOW-3 routing branch changes the registered stash shape; deleting the LOW-2 single-new-object gate flips the `is_none()` assertion; ∞-gating the LOW-1 re-offer fails; neutralizing `isUnbounded` leaks the finite count; reverting the LOW-5 plural key renders "Create 1 tokens" at value=1.

## Gate A

Gate A PASS head=247ffb76188651139168f03f3f629c8188b327bf base=836ff312ae2073c99af28d286b0c4915faa8a458

## Anchored on

- `crates/engine/src/game/engine_resolution_choices.rs` — the existing `pending_mana_ability` early-return in the same `SubmitPayAmount` handler; the MED guard mirrors that pause-aware `WaitingFor` idiom for the copy-token mint.
- `client/src/viewmodel/cardProps.ts` + the render sites' existing `isUnbounded` membership read from engine `derived.unbounded_counters` (`PermanentCard.tsx` / `ArtCropCard.tsx`); the LOW-4 tooltip fix threads that same engine-provided flag (display-only, no new frontend logic).

## Final review-impl

Final review-impl PASS head=247ffb76188651139168f03f3f629c8188b327bf

## Claimed parse impact

None — engine game-logic (a defensive boundary guard, docs, tests) + frontend display/i18n; no parser or card-data change. (Coverage parse-diff on #6238 already read "No card-parse changes".)

## Combo-detector series

| Pos | PR | Delivers |
|-----|----|----------|
| PR-7 | #5672 | Loop shortcut **offer** + APNAP accept-or-shorten window (CR 732.2a–c). |
| PR-8 | #6238 (merged) | Loop-shortcut **acceptance + phase-boundary collapse/materialization** (tokens/counters/life/mana → `∞`). |
| **this PR** | (this PR) | **Review follow-up** on #6238 — boundary-mint pause guard, declined-axis `∞` lifecycle doc + test, single-fodder invariant, accept-time routing E2E, `∞` tooltip + plural i18n. |

**Predecessor: #6238 — https://github.com/phase-rs/phase/pull/6238 (MERGED)** — this PR addresses its post-merge review findings.

## Deferred (disclosed)

- **Declined-axis `∞` lifecycle (answering the reviewer's LOW-1 question):** a declined axis's `∞` mark is kept as an intentional **capability marker** (CR 732.2b never forces a shortcut; the axis was not applied, so game totals stay correct) and is retired by a later genuine re-detection re-collapse or the debug toggle. Collapse-to-finite was rejected as rules-wrong (it would hide a still-usable capability). If you'd prefer different UX here, say so and I'll adjust.
- **Object-growth enabler-registration gap (F2):** `materialize_object_growth_shortcut` never calls `register_unbounded_loop_enablers`, so `zones.rs`'s enabler-departure auto-revoke is inert for object-growth `∞` marks. This is **pre-existing** (predates this PR) and class-wide — deferred to a separate follow-up rather than widened here.

### DIFF
--- a/client/src/components/board/PermanentCard.tsx
+++ b/client/src/components/board/PermanentCard.tsx
@@ -919,7 +919,12 @@ export const PermanentCard = memo(function PermanentCard({
               // counter unboundedly — render ∞ instead of the (still-finite) real count.
               const isUnbounded = unboundedCounterTypes.includes(type);
               return (
-                <CounterTooltip key={type} type={type} count={count}>
+                <CounterTooltip
+                  key={type}
+                  type={type}
+                  count={count}
+                  isUnbounded={isUnbounded}
+                >
                   <span
                     className={`flex items-center gap-0.5 rounded px-1 text-[10px] font-bold text-white ${COUNTER_COLORS[type] ?? "bg-purple-600"}`}
                   >
--- a/client/src/components/card/ArtCropCard.tsx
+++ b/client/src/components/card/ArtCropCard.tsx
@@ -214,7 +214,12 @@ export const ArtCropCard = memo(function ArtCropCard({ objectId }: ArtCropCardPr
                   // counter unboundedly — render ∞ instead of the (still-finite) real count.
                   const isUnbounded = unboundedCounterTypes.includes(type);
                   return (
-                    <CounterTooltip key={type} type={type} count={count}>
+                    <CounterTooltip
+                      key={type}
+                      type={type}
+                      count={count}
+                      isUnbounded={isUnbounded}
+                    >
                       <span
                         className={`rounded-full flex items-center justify-center font-bold text-white shadow-md border border-black/50 ${COUNTER_COLORS[type] ?? "bg-purple-600"}`}
                         style={counterStyle}
--- a/client/src/components/card/__tests__/ArtCropCard.test.tsx
+++ b/client/src/components/card/__tests__/ArtCropCard.test.tsx
@@ -346,4 +346,40 @@ describe("ArtCropCard", () => {
     expect(screen.getByText("2")).toBeInTheDocument();
     expect(screen.queryByText("∞")).not.toBeInTheDocument();
   });
+
+  // LOW-4 (CR 732.2a / CR 701.34a): the ∞ pill and its TOOLTIP must agree. Before the fix the
+  // badge showed ∞ while the tooltip interpolated the raw finite count ("… : 2"), contradicting
+  // the pill. The tooltip summary must say ∞, not leak the count.
+  it("tooltip renders ∞ and hides the finite count when the counter is unbounded", () => {
+    mockUseCardImage.mockReturnValue({ src: "card.png", isLoading: false, isRotated: false, isFlip: false });
+    const permanent = pentadWithCharge();
+    useGameStore.setState({
+      gameState: {
+        objects: { [permanent.id]: permanent },
+        derived: { unbounded_counters: { [permanent.id]: ["charge"] } },
+      } as never,
+    });
+
+    render(<ArtCropCard objectId={101} />);
+
+    // The tooltip summary line reads "∞ <label> counters", never the finite "2 <label> counters".
+    expect(screen.getByText(/∞ \S+ counters/i)).toBeInTheDocument();
+    expect(screen.queryByText(/2 \S+ counters/i)).not.toBeInTheDocument();
+  });
+
+  it("tooltip shows the finite count when the counter is NOT unbounded (matched pair)", () => {
+    mockUseCardImage.mockReturnValue({ src: "card.png", isLoading: false, isRotated: false, isFlip: false });
+    const permanent = pentadWithCharge();
+    useGameStore.setState({
+      gameState: {
+        objects: { [permanent.id]: permanent },
+        derived: { unbounded_counters: {} },
+      } as never,
+    });
+
+    render(<ArtCropCard objectId={101} />);
+
+    expect(screen.getByText(/2 \S+ counters/i)).toBeInTheDocument();
+    expect(screen.queryByText(/∞ \S+ counters/i)).not.toBeInTheDocument();
+  });
 });
--- a/client/src/components/mana/PayAmountChoiceUI.tsx
+++ b/client/src/components/mana/PayAmountChoiceUI.tsx
@@ -107,7 +107,13 @@ export function PayAmountChoiceUI() {
               className={gameButtonClass({ tone: "emerald", size: "md" })}
             >
               {loopCollapseAxis
-                ? t(`mana.loopCollapseAmount${loopCollapseAxis}`, { value })
+                ? // `count` drives i18next pluralization for the Tokens axis
+                  // (loopCollapseAmountTokens_one/_other); the ×N-framed
+                  // Counters/Life/Mixed keys keep reading `value`.
+                  t(`mana.loopCollapseAmount${loopCollapseAxis}`, {
+                    value,
+                    count: value,
+                  })
                 : t("mana.payAmount", { value, resource: resourceLabel })}
             </button>
           </div>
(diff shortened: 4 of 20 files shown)


matthewevans (changes requested):
**Blocked — the new paused-token early return breaks the boundary collapse’s atomic bookkeeping for mixed materialization stashes.**

## 🔴 Blocker

- `crates/engine/src/game/engine_resolution_choices.rs:2168` removes the complete materialization stash before the loop. If a preceding `Counters` or `Life` item is applied and appended to `collapsed`, then the new `Tokens` branch pauses and returns at `:2251-2254`, `clear_collapsed_materializations` at `:2339` is skipped. The applied finite counter/life change is left paired with its stale unbounded-axis/display state, while the full stash has already been consumed. That violates the stated “only actually-applied items are cleared” invariant and leaves subsequent loop handling with inconsistent state.

Evidence: current-head control flow; the added regression covers a tokens-only stash, so it cannot exercise this preceding-materialization case. Gemini’s transactional-warning is confirmed in this narrower form: the immediate defect is stale collapse bookkeeping (not necessarily an immediate duplicate application).

Suggested fix: make the submit transaction pause-safe—either prove/encode a single-item stash before applying, or persist/resume the remaining collapse state and clear the already-applied entries before returning. Add a mixed `Counters + Tokens` (and/or `Life + Tokens`) pause regression that asserts the finite prior effect occurs once and only the still-paused Tokens axis remains unbounded.

## ✅ Clean

- The CR-comment formatting suggestion is not a finding; the relevant comments already carry verified CR context and compound CR citations are permitted by project convention.

**Recommendation: request changes for atomic mixed-axis pause handling, then re-review the new head.**

lgray:

Fixed in `1ed7cfdb4`. Took the "clear the already-applied entries before returning" option you offered, and additionally made the axis ordering deterministic so the acceptance assertion holds regardless of stash registration order.

**What changed** (`engine_resolution_choices.rs`, `PayableResource::LoopCollapse` arm):

1. **Tokens processed last.** `items.sort_by_key(|i| matches!(i, PersistentAxisMaterialization::Tokens(_)))`. Tokens is the *only* pause-prone axis — its per-cycle fodder mint can raise an ETB replacement `NeedsChoice`, whereas `apply_counter_addition` is a direct `+=`, `apply_life_gain` is declined-if-observed (so it only applies when no replacement exists), and `DriveSequence` drives synchronously. A stable sort commits the deterministic Counters/Life axes first, so a Tokens pause can never strand a not-yet-applied finite axis — independent of the Tokens→Counters→Life registration order in `materialize_object_growth_shortcut`.

2. **Pause-safe clear.** `clear_collapsed_materializations(player, &collapsed)` now runs *before* the `pending_copy_token_resolution` early-return, cashing out the axes already applied this pass. The paused Tokens axis isn't in `collapsed`, so its ∞ axis + pile are preserved for manual play.

Net: a mixed loop's finite non-token effect commits exactly once and only the still-paused Tokens axis remains unbounded — the acceptance you asked for, verbatim.

**Regression added:** `med_mixed_counter_tokens_pause_commits_finite_counter_and_keeps_only_tokens_unbounded` — a mixed **Counters + Tokens** pause with Counters registered *after* Tokens, so the sort is load-bearing. Both discriminators measured (revert → observe FAIL → restore):
- remove the sort → Tokens pauses at index 0, the counter is never applied (stuck at base) → FAIL.
- remove the clear → the committed Counter axis keeps a stale ∞ mark → FAIL.

I built the Counters+Tokens variant (your "and/or") rather than Life+Tokens: Counters is the naturally-unobserved axis on the real 4p `OFFER_STATE` *and* exercises the harder `clear_collapsed_materializations` surviving-target arm; Life and Counters are treated identically by both the sort and the clear (both non-`Tokens` variants). Happy to add an explicit Life+Tokens variant too if you'd prefer it spelled out.

One design-choice note: this is the "clear the already-applied entries" branch of your suggestion, not literal persist-and-resume-the-remaining-collapse. Game totals are exact either way; the paused Tokens axis is left ∞ for manual play rather than resumed automatically (CR 732.2b never forces the shortcut). Say the word if you specifically want resume semantics instead.

CR annotations: `CR 732.2a` (boundary collapse) / `CR 732.2b` (a shortcut is never forced), both grep-verified against `docs/MagicCompRules.txt`.

Verification (worktree, run directly — not Tilt-watched): `combo_infinite_pile` 22/22 · `cargo check --workspace` clean · `clippy -p engine --all-targets` 0 warnings (20m40s) · `fmt --check` clean. Independent review-impl came back clean with both revert-flips re-observed by the reviewer.

matthewevans:
**Reviewed current head `e6ebd635`.** The mixed-axis pause/cleanup blocker is resolved, including the production replacement-choice regression.

This remains held because the required parse-diff artifact is still baseline-pending, so the current-head card-level evidence is unavailable. Please let that artifact populate (or update the branch if GitHub requires it); no further implementation change is requested at this point.

matthewevans (approved):
Approved: the mixed-axis pause transaction now clears only committed finite axes, preserves the paused Tokens axis, and is covered through the replacement-choice pipeline.

PR #6276, opened 2026-07-21
feat(engine,client): badge permanents that a copy effect turned into copies

### LINKED ISSUE #5932: Indistinguishable Copy — I can't tell which card is a Reveillark and which is a Phantasmal Image copy of a Reveillark i…
Reported in Discord: https://discord.com/channels/1485498006781427802/1525691902937010326/1525691902937010326

**Thread:** Indistinguishable Copy
**Discord thread id:** `1525691902937010326`
**Discord message id:** `1525691902937010326`
**Cards:** phantasmal image

## Summary
I can't tell which card is a Reveillark and which is a Phantasmal Image copy of a Reveillark in the attached game state.

## Reported text

I can't tell which card is a Reveillark and which is a Phantasmal Image copy of a Reveillark in the attached game state.

## Attachments

- [game-state-turn-16-2026-07-12T02-34-16-774Z.zip](https://cdn.discordapp.com/attachments/1525691902937010326/1525691903121821746/game-state-turn-16-2026-07-12T02-34-16-774Z.zip?ex=6a5994cf&is=6a58434f&hm=eafaae46f03da9bd7830e94622bba2fe544b0dd683edaa812329a51dd00d0a43&)

### phantasmal image
You may have this creature enter as a copy of any creature on the battlefield, except it's an Illusion in addition to its other types and it has "When this creature becomes the target of a spell or ability, sacrifice it."

---
<sub>report_id: `discord:1525691902937010326:1525691902937010326:0` · discord: `1525691902937010326/1525691902937010326`</sub>

minion1227:
## Summary

A Phantasmal Image that entered as a copy of Reveillark rendered identically to the real Reveillark — same art, same name, same P/T — so the board gave no way to tell them apart.

Closes #5932

## Root cause

The board already has a "Copy" badge, but it was gated on a **token-copy** heuristic:

```ts
const isCopy = obj.is_token === true && obj.display_source !== "Token" && !obj.face_down;
```

A real card under a copy effect is not a token, so it never qualified.

There was no serialized signal the frontend could use instead — I checked the two obvious candidates and both are dead ends:

- **`GameObject.is_copy`** means "not represented by a card" (CR 707.10) and is deliberately cleared when a copy resolves onto the battlefield (`game/stack.rs:835-836`, `game/sba.rs:4869`). For Phantasmal Image it is correctly `false`.
- **`copy_modifications`** lives on `CopyChosenSelection` (a `WaitingFor` payload), not on `GameObject`.

The copy itself is a Layer 1a `CopyValues` continuous effect (CR 613.2a + CR 707.2), not a property of the object — and it overrides `printed_ref`, which is why even image lookup follows the copied card.

## Fix

The engine classifies it, since the client cannot and should not derive it (display-layer rule).

**Engine** — `DerivedViews::copied_permanents`: battlefield permanents whose copiable values a live copy effect currently supplies. Membership is decided with the same `matches_target_filter` the layer engine uses to pick a continuous effect's recipients, so the projection and the effect that actually rewrote the object's characteristics can never disagree. Using the shared matcher (rather than special-casing the `SpecificObject` filter copy effects usually carry) also keeps it correct for any filter shape a future copy effect might use.

**Client** — `PermanentCard` unions the projection with the existing token-copy case, so the badge now covers the whole class — Clone, Vesuvan Doppelganger, and every "enters as a copy" permanent — instead of only the token half.

**CR 708.2** — face-down permanents are excluded on **both** sides. Their characteristics are only those the face-down rules grant, so surfacing "Copy" would leak what one really is. Keeping the guard on each side means neither alone can leak it.

## Anchored on

`DerivedViews::battlefield_keyword_badges` — the existing per-object display projection where the engine classifies something the client would otherwise have to reinterpret, keyed by `ObjectId` and consumed through a `useGameStore` selector. This follows that shape exactly: same struct, same serde `skip_serializing_if` treatment, same selector pattern in `PermanentCard`.

## Test plan

**Engine** (`derived_views.rs`) — three cases:
- flags the permanent a live `CopyValues` effect applies to, and *not* the original it copied
- reports nothing on an ordinary board (discriminating guard against "every permanent is a copy")
- never reports a face-down permanent (CR 708.2)

**Client** (`PermanentCard.test.tsx`) — badge appears for a permanent in `copied_permanents`, is absent on an ordinary permanent, and is absent when that permanent is face-down.

**Fail-on-revert verified**: restoring the old token-only condition makes `badges a real card that a copy effect turned into a copy` fail (1 failed / 37 passed) while both guards stay green — so the positive test is discriminating rather than vacuous.

Verified locally: `cargo fmt --all` clean; `cargo clippy -p engine --all-targets --features proptest -- -D warnings` exit 0; engine `derive_views` tests pass; **2163 client tests pass** (251 files); `tsc -b --noEmit` exit 0; `eslint` exit 0.

## AI-contributor disclosure

Authored with an LLM (Claude). Every CR number was verified against `docs/MagicCompRules.txt` before use: 613.2a (line 2976, Layer 1a copiable effects), 707.2 (line 5610, copiable values), 708.2 (line 5701, face-down characteristics). No parser changes, so the combinator-purity gate is not engaged.

I deliberately did not have the client infer copy-ness from what it can see (e.g. a name/`printed_ref` mismatch): that would be deriving game state the engine owns, and the copy effect overrides `printed_ref` anyway, so there is nothing left to compare. Happy to rework anything.

### DIFF
--- a/client/src/adapter/types.ts
+++ b/client/src/adapter/types.ts
@@ -2574,6 +2574,14 @@ export interface DerivedViews {
    * matters on the battlefield. Keyed by ObjectId-as-string.
    */
   battlefield_keyword_badges?: Record<string, Keyword[]>;
+  /**
+   * CR 613.2a + CR 707.2: battlefield permanents whose copiable values are
+   * currently supplied by a copy effect (Clone, Phantasmal Image, Vesuvan
+   * Doppelganger). Such a permanent renders identically to what it copied, so
+   * the engine classifies it here rather than leaving the client to guess.
+   * Face-down permanents are excluded per CR 708.2. Absent when empty.
+   */
+  copied_permanents?: ObjectId[];
   /** Keyed by attacking commander's current controller (PlayerId as string). */
   commander_damage_by_attacker?: Record<string, CommanderDamageView[]>;
   /**
--- a/client/src/components/board/PermanentCard.tsx
+++ b/client/src/components/board/PermanentCard.tsx
@@ -261,6 +261,13 @@ export const PermanentCard = memo(function PermanentCard({
       s.gameState?.derived?.battlefield_keyword_badges?.[String(objectId)]
       ?? EMPTY_KEYWORD_BADGES,
   );
+  // CR 613.2a + CR 707.2: whether a live copy effect supplies this permanent's
+  // copiable values. Engine-classified because a copy of a permanent lives in a
+  // Layer 1a continuous effect, not on the object — and the copy overrides
+  // `printed_ref` too, so a copy is pixel-identical to its source here.
+  const isCopiedPermanent = useGameStore((s) =>
+    (s.gameState?.derived?.copied_permanents ?? []).includes(objectId),
+  );
   // CR 732.2a / CR 701.34a: the counter-type keys the engine marks as ∞ (unbounded
   // counter-growth loop) for this object. Values match the object's `counters` map keys
   // (e.g. "charge"); the pill renders ∞ instead of ×N for any type in this set.
@@ -552,7 +559,16 @@ export const PermanentCard = memo(function PermanentCard({
   // CR 708.2: a face-down permanent has no characteristics other than those
   // its face-down rule grants, so never surface "Copy" on it — that would leak
   // that it's a token-copy (matches the `!face_down` guard on the keyword strip).
-  const isCopy = obj.is_token === true && obj.display_source !== "Token" && !obj.face_down;
+  // Two independent ways a permanent is a copy, unioned so the badge covers the
+  // whole class rather than only the token half (issue #5932):
+  //   - a token minted as a copy (`is_token` + card art), and
+  //   - a real card under a live copy effect (`copied_permanents`) — Clone,
+  //     Phantasmal Image, Vesuvan Doppelganger. These were previously missed
+  //     entirely, so two Reveillarks were indistinguishable on the board.
+  // Both stay subject to the same face-down guard below.
+  const isCopy =
+    (obj.is_token === true && obj.display_source !== "Token" && !obj.face_down)
+    || (isCopiedPermanent && !obj.face_down);
 
   // Filter out loyalty counters — shown separately as the loyalty badge
   const counters = Object.entries(obj.counters).filter((entry): entry is [string, number] => entry[1] != null && entry[0] !== "loyalty");
--- a/client/src/components/board/PermanentCard.tsx
+++ b/client/src/components/board/PermanentCard.tsx
@@ -565,10 +565,12 @@ export const PermanentCard = memo(function PermanentCard({
   //   - a real card under a live copy effect (`copied_permanents`) — Clone,
   //     Phantasmal Image, Vesuvan Doppelganger. These were previously missed
   //     entirely, so two Reveillarks were indistinguishable on the board.
-  // Both stay subject to the same face-down guard below.
+  // CR 708.2: the face-down guard leads, so it covers BOTH sources — a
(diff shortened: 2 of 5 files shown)


matthewevans (changes requested):
**Request changes — the copy badge remains after a temporary copy effect has lapsed.**

## 🔴 Blocker

**[MED]** The derived badge treats inactive copy effects as live. Evidence: `crates/engine/src/game/derived_views.rs:479-491` checks stored `CopyValues` and target matching only, while the layer application authority in `crates/engine/src/game/layers.rs:4590-4611` also rejects failed `ForAsLongAs` durations and source conditions. `crates/engine/tests/integration/become_copy.rs:2010-2061` proves Zygon Infiltrator's copy lapses when its target untaps while the TCE remains stored.

Why it matters: after that untap, the permanent is no longer a copy but still displays the UI “Copy” badge.

Suggested fix: share the layer-active continuous-effect eligibility predicate with the derived projection, and add a derived-view regression asserting no badge after Zygon's target untaps.

## ✅ Clean

Battlefield ordering is deterministic (`im::Vector` insertion order), so the existing projection ordering is not a correctness concern.

**Recommendation: gate the badge on the same active-effect predicate as layers, add the lapsed-copy regression, then regenerate current-head artifacts.**

minion1227:
Blocker fixed in `e111c2b` — you were right, and it was a real defect rather than a doc nit.

## The liveness gate

My predicate asked "is a `CopyValues` effect stored and does its filter match", which is not the same question as "is it applying". `gather_transient_continuous_effects` already knew the difference, so I extracted its three gates — departed `UntilHostLeavesPlay` host, `ForAsLongAs` duration that no longer holds, false source condition — into `transient_effect_is_live`, and both the layer engine and `derive_views` now call it.

Extracted rather than duplicated deliberately: a second copy of that predicate would be free to drift, which is exactly the class of bug you just caught. `gather_transient_continuous_effects` keeps its existing behaviour; the condition gate simply runs inside the shared predicate and the retained recipient-context condition is unchanged.

## The regression

`derive_views_drops_the_copy_flag_once_a_temporary_copy_effect_lapses` builds Zygon's target-relative `ForAsLongAs { IsTapped { Target } }` copy with `set_transient_duration_subject` bound to the target, asserts the badge shows while the target is tapped, then untaps it and asserts the badge is gone. It explicitly asserts the TCE is **still stored** at that point, so it exercises liveness rather than sweep/removal.

Verified discriminating: replacing the gate with `true` makes it fail, so it cannot pass vacuously.

## Also addressed

- `copied_permanents` is now sorted, matching what its doc comment already claimed. Noted your point that `im::Vector` insertion order is already deterministic — the sort is belt-and-braces so the payload depends only on which permanents are copies, not on battlefield churn.
- The client's duplicated `!obj.face_down` check is factored out so the CR 708.2 guard leads and covers both copy sources (@gemini-code-assist's suggestion).

Current head verified: **17442 lib tests**, **3676 integration tests**, `clippy -p engine --all-targets --features proptest -- -D warnings` exit 0, 2163 client tests, `tsc`/`eslint` clean. The `gather_transient_continuous_effects` refactor touches every continuous effect, so I ran the full lib + integration suites rather than the targeted ones.

matthewevans (approved):
Approved after maintainer fixup 73bad4c: merge-owned Layer-1 CopyValues effects are excluded, with a discriminating merge projection regression.

matthewevans (approved):
Approved current maintainer test fix 35e75d0: the merge exclusion is effect-scoped, with coverage that preserves a later independent copy effect on the merged host.

PR #6279, opened 2026-07-21
feat(engine): CR 801 Limited Range of Influence + CR 809 Emperor Variant

### LINKED ISSUE #6277: feat(engine): CR 801 Limited Range of Influence + CR 809 Emperor Variant
## Summary

`FormatConfig::range_of_influence: Option<u8>` (CR 801, Limited Range of Influence) has existed on `FormatConfig` since it was first introduced, but it is entirely dead: every one of the ~13 `FormatConfig` constructors sets it to `None`, and there is no consumer anywhere in the engine (`grep -rn "RangeOfInfluence|within_range|seat_distance" crates/engine/src` returns zero hits beyond one dead comment in `game/engine.rs:3463` noting the gap). Likewise, `GameFormat` has no `Emperor` variant — CR 809's Emperor Variant, the format CR 801.1 says ROI is "always used" in, is entirely unimplemented.

This proposes implementing both together: CR 801 as general-purpose, opt-in infrastructure (usable by any future multiplayer format, not just Emperor — CR 801.1: "can be applied to most multiplayer games... often used for games involving five or more players"), and CR 809 (Emperor) as the first concrete consumer, following the exact same "self-contained variant runtime module" pattern already used by `game/archenemy.rs` and `game/planechase.rs`.

## Scope

Backend/engine only (no frontend work) — mirrors how Archenemy/Planechase/Momir all shipped their engine PR without bespoke client work first; the client already generically renders any `GameFormat`/`FormatConfig` the engine returns via `GameFormat::registry()`.

**In scope** (each independently CR-verified against `docs/MagicCompRules.txt`):
- CR 801.2/801.2a-d: seat-distance + range-of-influence predicates, computed live per call (a documented simplification of CR 801.2c's turn-start snapshot).
- CR 801.3: attack legality. CR 801.4: target legality. CR 801.6: activated-ability activation legality. CR 801.8/801.9: Aura/Equipment/Fortification attach legality. CR 801.14: "wins the game" opponent-elimination scoped to range.
- CR 809: `GameFormat::Emperor`, `FormatConfig::emperor()`, `FormatTopology::FixedTeams { team_size: 3, team_count: 2, turn_structure: IndividualTurns }` (the first real consumer of the existing-but-dead `TurnStructure::IndividualTurns` variant), `emperor_players: Vec<PlayerId>`, an asymmetric per-role ROI accessor (CR 809.3a: emperors get range 2, generals get range 1), and the CR 809.5a-c asymmetric team win/loss (a general's individual loss does **not** end their team's game; an emperor's loss/win/draw does, unlike Two-Headed Giant's symmetric "any teammate's loss ends the team" model).

**Explicitly deferred** (documented as follow-ups, not silently dropped):
- CR 804 (Deploy Creatures Option, a separable CR-809.3b default).
- CR 801.5a-c (choice-making range constraints), CR 801.7/801.7a (triggered-ability range gating — no safe minimal seam identified in `triggers.rs`'s ~34,700 lines without a deeper follow-up investigation), CR 801.10/801.11 (mass-effect/information-gathering range scoping — touches many individual effect resolvers, too invasive for a first pass), CR 801.12 (world rule), CR 801.13a/801.13b (replacement/prevention effect scoping), CR 801.15/801.16 (draw/loop-draw propagation — no "the game is a draw" effect exists yet in this engine to extend, and the loop-detection subsystem is intentionally left untouched), CR 801.17/801.18 (moot while the above are deferred).
- **CR 809.3c (attack-only-the-adjacent-seat restriction) is deferred, not implemented**, for a specific reason: CR 809.2 places the emperor "in the middle" of a team that "sits together on one side of the table," implying real Emperor games use rectangular/oval seating where the two teams face each other across a gap — a geometry this engine's sequential/circular `seat_order` model doesn't represent. A literal circular-seat-adjacency implementation (`seat[i-1]`/`seat[i+1]`) leaves the middle seat of each 3-player team — where CR 809.2 places the emperor — with **zero adjacent opponents**, which cannot be the intended behavior. Rather than guess at an unverifiable table-geometry model, this is left for a follow-up once a concrete seating model is agreed on. Emperor players are still subject to their (asymmetric) CR 801.3 range-of-influence restriction on attacks in this pass, just not the additional 809.3c narrowing.
- CR 809.6/809.6a (team sizes other than 3, team counts other than 2 — v1 ships the simplest legal case CR 809.1's own example describes: two teams of three).

## Prior art check

No existing open issue or PR references "Emperor variant," "range of influence," or CR 801/809. The one related closed PR, **#2686** ("multiplayer ROI enforcement, replay viewer, sealed limited, parser coverage"), was closed for bundling ROI enforcement together with an unrelated frontend replay viewer, a frontend sealed-limited page, and unrelated parser fixes in one PR, plus a process ask to open an issue first — **not** a rejection of ROI itself. This issue is filed first, and the implementation is scoped to backend-only, addressing both points raised there.

## Architecture

Every new CR 801 check is a single additive early-return/filter guard inside an existing legality/resolution function, gated on `FormatConfig::range_of_influence_for_player(viewer).is_some()`. When `None` (every one of the 17 pre-existing `GameFormat` variants, unconditionally), every new guard is a proven no-op — existing formats are unaffected by construction, not by convention. Concrete seams (all read and confirmed against current source, not assumed): `game/targeting.rs::can_target`, `game/combat.rs::attacker_can_attack_target`, `game/casting.rs::can_activate_ability_now_with_restriction_gates`, `game/effects/attach.rs::attachment_illegality`, `game/effects/win_lose.rs::resolve_win`.

New files: `game/range_of_influence.rs` (seat-distance + range predicates), `game/emperor.rs` (asymmetric elimination-cascade + win-condition helpers). Touched: `types/format.rs` (new variant, field, constructor, per-role ROI accessor, five existing exhaustive-match sites), `game/elimination.rs` (the existing `eliminate_players_simultaneously` 2HG-only cascade guard generalizes to a new `team_elimination_cascades_from` predicate that is unconditionally unchanged for 2HG and only newly-true for an Emperor player who is specifically an emperor; `check_game_over` gets a new Emperor branch mirroring its existing Archenemy branch's asymmetric-winner shape), `game/deck_validation.rs` (two exhaustive matches found during self-review need a trivial new Emperor arm, treated like FreeForAll/2HG/Limited — a plain per-player deck, no special legality).

## Verification note

I do not have a working Rust build toolchain in this environment this session (no MSVC linker/Windows SDK, no WSL, no Tilt) — every claim above was verified by reading the actual current source at the cited functions/line numbers (not assumed from memory), including a full manual re-derivation confirming `game/topology.rs` needs zero changes for `team_size: 3` and confirming (by grepping every `.is_eliminated = true` write site in the crate) that `game/elimination.rs::do_eliminate` is the sole production path that can eliminate a player, which is the invariant the new asymmetric Emperor win-condition branch relies on. The forthcoming PR will be explicit about this same limitation and will rely on upstream CI plus maintainer review for compiled/test-suite confirmation.

claytonlin1110:
Closes #6277.

## Summary

Implements CR 801 (Limited Range of Influence Option) as general-purpose, opt-in infrastructure, and CR 809 (Emperor Variant) as its first concrete consumer — following the same "self-contained variant runtime module" pattern already used by `game/archenemy.rs` and `game/planechase.rs`.

`FormatConfig::range_of_influence: Option<u8>` has existed since `FormatConfig` was introduced but was entirely dead (every constructor set it to `None`, zero consumers anywhere). This PR gives it real consumers while keeping it a proven no-op for every one of the 17 pre-existing `GameFormat` variants.

## Files changed

- `crates/engine/src/game/range_of_influence.rs` (new) — CR 801.2 seat-distance predicates.
- `crates/engine/src/game/emperor.rs` (new) — CR 809.5b asymmetric elimination-cascade helper.
- `crates/engine/src/types/format.rs` — `GameFormat::Emperor`, `FormatConfig::emperor()`, `emperor_players: Vec<PlayerId>`, `range_of_influence_for_player()`, the five existing exhaustive `GameFormat` matches (`legality_format`/`sideboard_policy`/`label`/`registry`/`for_format`), plus a `starting_life_for_seat`/`starting_life_for_player` fix (see below).
- `crates/engine/src/game/elimination.rs` — the 2HG-only cascade guard generalizes to `team_elimination_cascades_from`; a new `check_game_over` branch mirrors the existing Archenemy branch's asymmetric-winner shape.
- `crates/engine/src/game/targeting.rs`, `combat.rs`, `casting.rs`, `effects/attach.rs`, `effects/win_lose.rs` — one additive CR 801 guard each in an existing legality/resolution function.
- `crates/engine/src/game/deck_validation.rs` — two more exhaustive `GameFormat` matches found during self-review (not in my original 5-site list), given the same "plain deck, no special check" treatment as FreeForAll/2HG/Limited.
- `crates/engine/src/game/mod.rs`, `casting_tests.rs` — module registration and one test respectively.

## CR references

- CR 801.2/801.2a-d (seat-distance + range-of-influence definition), CR 801.3 (attack legality), CR 801.4 (target legality), CR 801.6 (activated-ability activation), CR 801.8/801.9 (Aura/Equipment/Fortification attach legality), CR 801.14 ("wins the game" opponent-elimination scoping).
- CR 809.1-809.5c (Emperor Variant: two teams of three, one emperor per team, CR 809.3a asymmetric range of influence, CR 809.4 individual turns, CR 809.5a-c asymmetric team win/loss), CR 809.7 (team resources, including life, are NOT shared — distinct from CR 810.8's Two-Headed Giant model).

## Root-cause-shaped design note (why every CR 801 guard is provably inert for existing formats)

Every new check is a single additive early-return/filter guard, gated on `FormatConfig::range_of_influence_for_player(viewer).is_none_or(...)`. `range_of_influence_for_player` returns `None` for every format except Emperor (which is asymmetric by role, `Some(2)`/`Some(1)`) or a future format that explicitly sets the flat `range_of_influence` field. `None` makes every guard's boolean check unconditionally `true` — the existing behavior of every pre-existing format is unchanged by construction, not by convention.

## A bug I found and fixed during self-review

Emperor's topology reuses `FormatTopology::FixedTeams { team_size: 3, .. }` (to get individual-turn/team-grouping machinery in `game/topology.rs` for free — no changes needed there, verified by hand-tracing every `has_shared_team_turns()`-gated function). But `FormatConfig::starting_life_for_player`/`starting_life_for_seat` already had a `FixedTeams` branch that **divides** `starting_life` by `team_size` — correct for Two-Headed Giant (CR 810: life genuinely shared, one pool split across the team), **wrong** for Emperor (CR 809.7: resources, including life, are explicitly NOT shared). Reusing `FixedTeams` for topology without accounting for this would have silently given every Emperor player `20 / 3 = 6` starting life instead of a full 20 each. Both methods now special-case `GameFormat::Emperor` to return the flat, undivided `starting_life`, with a regression test proving Two-Headed Giant's genuine division (`30 / 2 = 15`) is unchanged.

## Scope: in, deferred, and why

**In scope**, each independently CR-verified: CR 801.2/801.3/801.4/801.6/801.8/801.9/801.14 (listed above); CR 809's core (`GameFormat::Emperor`, asymmetric ROI, individual turns, asymmetric win/loss).

**Deferred, documented (not silently dropped)** — see `game/range_of_influence.rs` and `game/emperor.rs`'s module docs for the full per-item reasoning:
- CR 804 (Deploy Creatures Option) — a separable CR 809.3b default.
- **CR 809.3c (attack-only-the-adjacent-seat restriction)** — CR 809.2 places the emperor "in the middle" of a team that "sits together on one side of the table," implying real Emperor games use a table geometry (two teams facing each other across a gap) this engine's sequential/circular `seat_order` model doesn't represent. Worked a concrete 6-seat example: a literal circular-adjacency reading (`seat[i-1]`/`seat[i+1]`) leaves the *middle* seat of each 3-player team — exactly where CR 809.2 places the emperor — with **zero adjacent opponents**, which cannot be the intended behavior. Rather than guess at an unverifiable geometry, Emperor players are subject only to their (asymmetric) CR 801.3 range-of-influence restriction on attacks in this pass, not the additional 809.3c narrowing.
- CR 801.5a-c (choice-making range constraints), CR 801.7/801.7a (triggered-ability range gating — no safe minimal seam identified in `triggers.rs`'s ~34,700 lines without a deeper follow-up), CR 801.10/801.11 (mass-effect/information-gathering range scoping — touches many individual effect resolvers), CR 801.12 (world rule), CR 801.13a/801.13b (replacement/prevention effect scoping), CR 801.15/801.16 (draw/loop-draw propagation — no "the game is a draw" effect exists yet in this engine to extend, and the recently-built combo/loop-detection subsystem is intentionally left untouched), CR 801.17/801.18 (moot while the above are deferred), CR 809.6/809.6a (team sizes/counts other than the CR 809.1 baseline of two teams of three).
- Frontend: none in this pass — the client already generically renders any `GameFormat`/`FormatConfig` via `GameFormat::registry()`, matching how Archenemy/Planechase/Momir shipped their engine PRs without bespoke client work.

## Anchored on

- `game/archenemy.rs` / `game/planechase.rs` — the two existing "variant runtime module" precedents this mirrors structurally.
- `game/elimination.rs`'s existing Archenemy branch in `check_game_over` — the direct template for the new Emperor branch's asymmetric-winner shape (naming the emperor specifically, not an arbitrary teammate — see the hostile-fixture test `emperor_winner_is_the_emperor_not_an_arbitrary_teammate`, which places the emperor at the *highest*-id seat on their team specifically so a naive "pick the first living player" implementation would name the wrong winner).
- `game/combat.rs`'s existing `attack_passes_neighbor_restriction` (`StaticMode::AttackOnlyNeighbor`, CR 508.1c) — read as the closest sibling mechanism to CR 809.3c, informing the decision to defer 809.3c rather than force an unverified adjacency model onto it.
- `FormatConfig::starting_life_for_player`'s existing `OneVsMany` branch (Archenemy's asymmetric 40/20 split) — the precedent for a format needing its own non-generic branch rather than reusing a sibling topology's division logic, which is exactly the bug described above.

## Verification

I do not have a working Rust build toolchain in this environment this session (no MSVC linker/Windows SDK, no WSL, no Tilt) — I could not run `cargo test`/`cargo clippy`/`cargo build`. `cargo fmt --all` **does** work (no linking involved) and passed clean over this diff three times across the course of implementation (it also caught my own dead-code slip — I initially wrote `game::emperor::has_team_based_win_loss`/`emperor_of_team` helpers I never ended up calling from production code, and removed both once I noticed on self-review, to avoid an unused-function warning).

In place of running the suite, I verified by hand, tracing the actual current source at every cited function:
- Grepped the whole crate for every `.is_eliminated = true` write site to confirm `game/elimination.rs::do_eliminate` (via `eliminate_players_simultaneously`) is the *sole* production path that can eliminate a player — the invariant the new `check_game_over` Emperor branch's winner-naming logic depends on (a living team always has a living emperor).
- Manually re-derived every `game/topology.rs` function (`team_id`, `team_members`, `teammates`, `apnap_choice_groups_from`, `next_turn_representative`, `priority_pass_representative`) against `FixedTeams { team_size: 3, team_count: 2, turn_structure: IndividualTurns }` to confirm zero changes were needed there — `has_shared_team_turns()` short-circuits every turn/priority-representative function to the plain individual-player path before any team-size-specific logic runs.
- Grepped beyond my original 5-site list of exhaustive `GameFormat` matches and found 2 more in `deck_validation.rs` (now fixed) — and confirmed every other `GameFormat::` reference elsewhere in the workspace (`engine-wasm`, `server-core`, `lobby-broker`, `phase-server`, `card_subset.rs`, `derived_views.rs`, `deck_loading.rs`) is either a non-exhaustive `if`/wildcard-armed `match` (no change needed, confirmed the wildcard's default is semantically correct for Emperor) or an unrelated test fixture.
- Every type/variant/field name used in new tests was copied from currently-existing, already-passing sibling code in the same file (e.g. `combat.rs`'s `attack_only_neighbor_left_restricts_to_left_neighbor_and_their_planeswalkers` for the `declare_attackers`-based attack test; `attach.rs`'s `attachment_illegality_cant_be_enchanted_blocks_aura` for the attach test; `win_lose.rs`'s `win_the_game_eliminates_all_opponents` for the win-propagation test), not invented — including catching and fixing a seat-distance arithmetic mistake in my own first draft of one test (a 3-player table can never produce a distance-2 pair; fixed by using 5-player tables throughout).

I'd appreciate CI (and/or a maintainer with a working local toolchain) confirming the actual `cargo test -p engine` / `cargo clippy` run, since I was not able to produce that signal myself this session.

## Scope note

This PR's ~937 lines is a deliberately conservative scope given the no-compiler constraint above — CR 801 alone has 18 lettered subclauses reaching into nearly every engine subsystem (targeting, triggers, replacement/prevention, mass effects, information-gathering, the world rule, win/draw/loop-draw propagation). I traced concrete, verified seams for the subset I judged safely implementable and hand-verifiable without a compiler in one pass, and explicitly documented everything else as a follow-up rather than rushing wider coverage onto unverified seams (particularly `triggers.rs` at ~34,700 lines, where I did not find a seam I was confident enough in to touch blind).

## Track

Developer

## LLM

Model: claude-sonnet-5


matthewevans (changes requested):
**Request changes — the PR enables a rules-incorrect Emperor format.**

## 🔴 Blocker

**[HIGH]** Range of influence is recomputed live rather than snapshotted at the turn boundary. Evidence:  expressly changes membership immediately when a player leaves or rejoins, while CR 801.2c says the affected players are determined *"as each turn begins"* and gives the example that a departure changes range at the next turn start.  also retains eliminated intermediate seats (), so it cannot produce the required next-turn topology. Why it matters: targeting, casting, combat, attachment, and win/loss checks receive a different range mid-turn than the rules permit. Suggested fix: introduce a turn-boundary range snapshot over living seats, make all CR 801 consumers read it, and add a three-player/mid-turn-elimination test that remains unchanged until the next turn then updates.

**[HIGH]**  is user-selectable while the implementation intentionally omits mandatory default rules. Evidence:  defers both CR 809.3b and CR 809.3c; CR 809.3 says Emperor uses the following default options, and CR 809.3b requires Deploy Creatures while CR 809.3c restricts attacks to immediately adjacent opponents (including planeswalkers and battles). Why it matters: an enabled Emperor game produces rules-illegal attack and casting behavior, including an emperor attacking a non-adjacent opponent within range. Suggested fix: route this through the full engine-implementer cycle and implement the topology/Deploy-Creatures/adjacency authorities before exposing Emperor as a playable format.

## ✅ Clean

The no-op guard for existing formats and the Emperor-specific asymmetric range values are correctly located concepts, but they cannot compensate for the missing CR 801 snapshot and CR 809 defaults.

**Recommendation: rework this as a complete rules-correct variant implementation; do not enqueue the current head.**

matthewevans:
Correction to my request-changes review: shell quoting stripped the path anchors. The decision is unchanged.

- HIGH — range_of_influence.rs lines 12-19 explicitly recompute range live when a player leaves or rejoins, violating CR 801.2c: the players in range are determined as each turn begins. seat_distance at lines 44-62 also retains eliminated intermediate seats, so it cannot calculate the required next-turn topology. The fix belongs at a turn-boundary snapshot authority, read by every CR 801 consumer, with a mid-turn elimination test that changes only at the next turn.

- HIGH — emperor.rs lines 20-34 explicitly defer CR 809.3b and CR 809.3c even though CR 809.3 makes both default Emperor options. A selectable Emperor format must implement Deploy Creatures and immediate-adjacent-only attacks for creatures, planeswalkers, and battles. This needs the full engine-implementer cycle before the format is exposed.

The previous review remains a request for changes; this comment supplies the missing literal file anchors.

matthewevans:
**Closing this PR.** The current head is not eligible to land: it exposes an incomplete, rules-incorrect CR 801/809 variant, as detailed in the request-changes review. The PR also provides no valid implementation-method, Gate A, final review-impl, or local-toolchain verification evidence for a change of this scope.

Please select an existing open issue and submit a narrowly scoped PR against it only after running the local toolchain and the required implementation/review workflow. Engine-mechanic planning must use a higher-level frontier model; the Sonnet 5 planning route used here is not sufficient. Do not reopen this feature as a partial Emperor or range-of-influence implementation; it requires a complete, rules-correct engine design.

PR #6281, opened 2026-07-21
fix(parser): treat single-quoted spans as opaque in token-list splitters (Reef Worm)

JacobWoodson:
## The bug

Reef Worm creates a **cascading token**: a Fish that on death makes a Whale that on death makes a Kraken. The Oracle text nests the grants by quote depth — the Fish's ability in `"…"`, the Whale's in `'…'`:

```
create a 3/3 blue Fish creature token with "When this token dies,
create a 6/6 blue Whale creature token with 'When this token dies,
create a 9/9 blue Kraken creature token.'"
```

#4230 made the **outer** double-quoted span opaque to the token-list splitters, so the outer token stopped being misread as a `ChooseOneOf`. That fix was only half the story: the splitters' opaque unit covered **double quotes only**. One level down — parsing the Fish's trigger effect — the delimiter is **single** quotes, so `split_choice_list_items` severed the Whale clause at the `, create` inside `'…, create a 9/9 … Kraken.'`.

The result: Whale and Kraken came out as **siblings** (a bogus `ChooseOneOf`) with mangled descriptions (`"create a 6/6 … 'When ~ dies"`, `"create create a 9/9 … token.'"`), instead of a Whale that carries its own Kraken dies-trigger. So killing the Fish made *both* a Whale and a Kraken at once, and the Whale had no ability — the entire cascade past stage 1 was wrong.

## The fix

Adds a single-quoted span as an opaque unit in both `split_choice_list_items` (the splitter Reef Worm hits) and `split_create_token_sequence` (same latent defect class).

The single-quote unit is anchored on `" '"` — **space then quote**. An ability opener is always space-preceded (`with '…'`); a possessive apostrophe (`owner's`) never is. So the anchor matches grant delimiters and leaves possessives as plain characters — no new ambiguity.

## Test

The existing `reef_worm_nested_token_is_not_modal_choice` asserted only the outer token's *name* and `..`-ignored everything else, which is exactly why the depth-2 break survived #4230. Replaced it with `reef_worm_nested_token_cascade`, which walks the full chain:

```
Fish → GrantTrigger(dies) → Whale → GrantTrigger(dies) → Kraken
```

Verified **red** without the fix — it reproduces the sibling `ChooseOneOf` with the mangled descriptions.

## Verification

- `cargo test -p engine --lib` — 17,386 passed, 0 failed
- `cargo test -p engine --test integration` — 3,649 passed, 0 failed
- `./scripts/check-parser-combinators.sh` — Gate G / Gate A pass
- `cargo fmt --all`

Follow-up to #4230.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of nested ability text inside choice lists by correctly treating specific single-quoted spans as opaque units.
  * Prevented contractions/possessives and embedded apostrophes from being misread as span terminators.
  * Fixed end-to-end nesting for dies-triggered token cascades (Fish → Whale → Kraken), including the leaf token.
  * Ensured sibling quoted items remain distinct and aren’t incorrectly merged or split.

* **Tests**
  * Replaced Reef Worm regression with full cascade-structure assertions.
  * Added quote-span regression tests covering embedded apostrophes, item bounding, sibling separation, and the card ingestion path.


matthewevans (changes requested):
**[HIGH]** The new single-quoted opaque unit still splits valid nested grants at an apostrophe. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:4707` and `crates/engine/src/parser/oracle_effect/mod.rs:4863` use `take_until("'")`, so an inner `can't` or `owner's` is consumed as the closing quote; the subsequent `, create a …` is visible to the list splitter again. Why it matters: the PR claims this preserves nested token grants, but any such grant with an ordinary contraction or possessive can still become a bogus choice/sequence and resolve the cascade incorrectly. The existing token parser already documents and tests the required pairing rule: `crates/engine/src/parser/oracle_effect/token.rs:1114-1121` and `1247-1285` select the structural closing delimiter so apostrophes remain content. Suggested fix: factor/reuse a structurally anchored quote-span parser for both splitters, and add a full Fish → Whale → Kraken regression where the single-quoted ability contains `can't` or a possessive before an internal list separator.

matthewevans:
The current-head parse-diff artifact is now available and reports **no card-parse changes detected**: https://github.com/phase-rs/phase/pull/6281#issuecomment-5036307595. That conflicts with the PR's stated Reef Worm behavior change. The synthetic `parse_effect_chain` fixture is not enough to establish that the real card-data path changed; please add current-head evidence through the actual Reef Worm card parse/runtime path and explain the parse-diff result before this can be reconsidered.

matthewevans:
@coderabbitai review

matthewevans:
@coderabbitai full review

matthewevans:
@coderabbitai full review

matthewevans:
@coderabbitai full review

matthewevans (changes requested):
**Request changes — the current head still has a quote-boundary parser defect and does not prove a Reef Worm card-path change.**

## 🔴 Blocker

`crates/engine/src/parser/oracle_effect/mod.rs:4706` and `:4862` still use `take_until("'")` after recognizing the space-anchored opener. An apostrophe inside the span (for example, `can't` or `owner's`) becomes the closing delimiter, exposing a later `, create …` to the surrounding token-list splitter. This is the same current-head defect identified in the existing maintainer and CodeRabbit reviews; the current test at `tests.rs:44792` has neither form. Use one structurally paired single-quote span parser for both splitters, with a closing boundary that preserves embedded apostrophes, and cover a Fish → Whale → Kraken cascade containing a contraction or possessive before an internal separator.

`crates/engine/src/parser/oracle_effect/tests.rs:44838` only calls `parse_effect_chain`; it does not exercise Reef Worm's production card ingestion path. The current-head parse-diff artifact reports “No card-parse changes detected,” so the claimed Reef Worm behavior remains unproven. Add a `parse_oracle_text` regression using Reef Worm's Oracle text and retain the cascade assertions; it must distinguish this change from the base behavior.

**Recommendation: request changes.** Resolve the delimiter boundary and add the production-path regression, then rerun the parse-diff artifact on the resulting head.

matthewevans:
@coderabbitai review

matthewevans:
@coderabbitai full review

JacobWoodson:
Thanks — both blockers addressed.

**Structural close (was: `take_until("'")`).** Replaced the inline unit in both splitters with a shared `opaque_single_quoted_span` combinator that applies the same pairing rule as `token::find_anchored_single_quoted_span`: opener anchored on `" '"` (a possessive never STARTS a span) and closer = the LAST `'` in the input (an embedded `can't`/`owner's` is content, not the close). Single quotes never nest in Oracle templating, so the final `'` is unambiguously the ability's close. The combinator is generic over the nom error type so both splitters share one implementation (the choice splitter uses the default `Error`, the sequence splitter uses `OracleError`).

- `opaque_single_quoted_span_closes_at_last_apostrophe` — unit test: two embedded apostrophes stay content; a bare possessive does not open a span.
- `nested_token_cascade_survives_embedded_apostrophe` — Fish → Whale → Kraken where the nested single-quoted grant contains `opponent's` before the internal `, create`. Verified **red** with the old first-apostrophe close (Whale/Kraken split into siblings).

**Production card-path regression.** Added `reef_worm_card_path_builds_cascade`, which runs Reef Worm's actual printed Oracle text through `parse_oracle_text` (not `parse_effect_chain`) and asserts all three cascade tokens are present, nested outer→inner (Fish, then Whale, then Kraken), and not a `ChooseOneOf`.

Verification: `cargo test -p engine --lib` (17,481), integration (token/choice), clippy, fmt, parser gates G/A all green on the merged-main base.

matthewevans (changes requested):
**Request changes — the current head has two parser-regression gaps.**

## 🔴 Blocker

`crates/engine/src/parser/oracle_effect/mod.rs:4858-4863` opens at `" '"` then uses `after_open.rfind('\'')` across the entire remaining list input. With two sibling list items that each carry a single-quoted grant, the first span consumes the separator and the second item through its closing quote. That prevents the surrounding `separated_list1` from seeing the sibling boundary, so the new opaque unit is not compositional for valid list items. Make the close belong to the current quoted item rather than the tail-wide final apostrophe, and add a regression with two sibling quoted items (including an embedded possessive/contraction in the first span).

`crates/engine/src/parser/oracle_effect/tests.rs:44953-44970` calls the card path, but then proves the cascade only through `format!("{dies:?}")` substring presence and debug-print order. A flat Fish/Whale/Kraken sibling shape can produce the same depth-first text and pass those assertions. Reuse the existing `granted_dies_token` structural traversal to follow `Effect::Token` → `GrantTrigger` → `Effect::Token` for Fish, Whale, and Kraken on the `parse_oracle_text` result.

## ✅ Clean

The current parse-diff sticky artifact reports no card-parse changes, which is consistent with this being a parser-shape correction rather than a coverage count change. The production-path test is therefore the right evidence to strengthen; it must assert the nested AST shape directly.

**Recommendation: request changes.** Fix the item-bounded single-quote close and replace the debug-order check with structural traversal, then rerun the existing parser/card-data CI evidence on the new head.

JacobWoodson:
Thanks — both blockers addressed.

**Item-bounded close.** The close is no longer a tail-wide `rfind`; it is the first `'` **not followed by an ASCII alphanumeric**. A contraction/possessive apostrophe (`can't`, `owner's`) is always glued to a following letter, so it stays content; a real closing quote is followed by punctuation, a space, `"` or end of input — so the span closes at its own quote and `separated_list1` keeps seeing sibling boundaries. Two new regressions, both verified **red** with the tail-wide close:
- `opaque_single_quoted_span_is_item_bounded_and_keeps_contractions` — unit: sibling spans stay separate; the first span keeps an embedded possessive.
- `sibling_quoted_token_items_stay_separate` — splitter-level: a disjunctive list with two quoted items (contraction embedded in the first) still splits into exactly two items, each keeping its own span.

The doc comment records why this deliberately differs from `token::find_anchored_single_quoted_span`'s last-`'` rule (that helper sees a single token's suffix; this combinator runs inside a list), and the one shape the rule cannot hold — a bare *plural* possessive (`opponents'`, apostrophe then space) inside the span. No printed card nests one in a single-quoted grant; flagging rather than hiding it.

**Structural card-path traversal.** `granted_dies_token` (the `Effect::Token` → `GrantTrigger`(ChangesZone) → execute hop) is hoisted to module level and shared. `reef_worm_card_path_builds_cascade` now walks Fish → Whale → Kraken structurally on the `parse_oracle_text` result and asserts the Kraken leaf — no substring or debug-print-order checks remain.

Verification: `cargo test -p engine --lib` (17,482), integration (token/choice), clippy, fmt, parser gates G/A all green.

matthewevans (changes requested):
**Verdict: Request changes — the shared quote parser still misclassifies a normal possessive as a closing delimiter.**

## 🔴 Blocker

**[MED]** Do not treat every apostrophe followed by whitespace as the end of a single-quoted span. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:4858-4882` explicitly documents that `opponents'` is unsupported, and the closer stops on its apostrophe because the next character is a space. Why it matters: this helper is used by both token and choice-list splitters; a quoted nested grant containing a plural possessive before an internal `, create` exposes that separator and recreates the malformed split for a valid class of Oracle text. Suggested fix: make the quote-boundary rule distinguish a delimiter from plural possessives without narrowing the supported grammar, and add a Fish → Whale → Kraken regression with `opponents'` before an internal separator.

JacobWoodson:
Addressed — the plural possessive is now span content, not a close.

**The rule.** A plural possessive hangs its apostrophe off a trailing `s` with prose continuing after whitespace (`opponents' creatures`); a genuine close never has that shape, because a quoted granted ability ends with terminal punctuation (`.`, or the Bronzehide comma convention) or a non-`s` keyword. So the close scan now treats `'` as content when it is alphanumeric-glued (`can't`, `owner's`) **or** preceded by `s`/`S` and followed by whitespace — everything else closes the span. This widens the grammar rather than narrowing it: sibling-item boundedness is unchanged (a close after `.'`/`,'` is still a close, and both sibling regressions still pass).

The one residual shape — quoted text ending in a bare `s`-word with *no* terminal punctuation immediately before the close — is unprintable under Oracle's quoted-grant punctuation convention; the doc comment records it explicitly rather than leaving it implied.

**Regression.** `nested_token_cascade_survives_embedded_apostrophe` is now parameterized over both apostrophe shapes — singular `opponent's` and plural `opponents'` — each placed in the nested grant **before** the internal `, create` separator, asserting the full Fish → Whale → Kraken nesting structurally. Verified **red** for the plural variant with the rule disabled.

Verification: `cargo test -p engine --lib` (17,482), integration (token/choice), clippy, fmt, parser gates G/A all green.

matthewevans (changes requested):
**Verdict: Request changes — the plural-possessive heuristic now rejects a valid closing delimiter.**

## 🔴 Blocker

**[MED]** Do not classify every `s'` followed by whitespace as a plural possessive. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:4883-4892` advances past any such apostrophe; therefore a sibling-list item ending `with 'This token has prowess' or ...` treats the closing quote after `prowess` as content. Why it matters: this shared item-bounded splitter then consumes into the next item or fails to find a close, so the generalized quoted-grant class remains malformed even though the new `opponents'` case passes. Suggested fix: use a delimiter rule that can distinguish a completed quoted span from a plural possessive without assuming every quoted ability ends in punctuation, and add the `prowess' or` sibling-list regression alongside the plural-possessive test.

**Recommendation: request changes; keep the `opponents'` coverage while making the span boundary general.**

JacobWoodson:
Addressed — `prowess` was a fair kill of the `s'` heuristic, so it's gone entirely.

**The rule is now lookahead against the item grammar, not a character heuristic.** An apostrophe (not glued to a following alphanumeric) closes the span when it is **preceded by terminal punctuation** (`.'`/`,'`) **or followed by a close boundary** — end of input, the outer `"`, sentence punctuation, or a list-separator phrase (` or `/` and `), via a small `single_quote_close_boundary` helper. A plural possessive fails both sides: it follows a letter and is followed by ordinary continuing prose (` creatures`, ` turns`). No assumption remains that a quoted ability ends in punctuation — `…has prowess' or …` closes because ` or ` is a boundary.

**Regressions.** `sibling_quoted_token_items_stay_separate` is parameterized over both first-span shapes: the punctuation-terminated grant with an embedded contraction, and the bare `s`-keyword close (`…has prowess'`) whose close is only recognizable from the ` or ` boundary. The `opponents'`/`opponent's` cascade regressions stay green under the same rule — both apostrophe families are covered by one mechanism now.

**Residual, documented on the combinator:** a possessive *directly* followed by a separator word (`…all your opponents' or …`) — a dangling possessive Oracle prose never prints.

Verification: `cargo test -p engine --lib` (17,482), clippy, fmt, parser gates G/A all green.

matthewevans (commented):
**Reviewed the refreshed quote-boundary implementation.**

The item-boundary close now covers both previously reported cases: a plural possessive remains content, while a bare `prowess' or …` close remains a delimiter. The sibling-list and nested-cascade regressions exercise those paths. I found no additional concrete parser blocker on this head.

This is held from approval until the in-progress CI and current-head card-data/parse-diff evidence finish. No local build was run.

matthewevans (approved):
Approved: current head uses the shared item-bounded single-quoted span parser, preserves contraction and possessive content, and has structural sibling/cascade regressions. The current parse-diff reports no unrelated card changes.

PR #6283, opened 2026-07-21
ship/native engine shell

matthewevans:
- **feat(desktop): native engine lifecycle — download, verify, spawn, adopt, GC (Workstream E)**
- **fix(desktop): non-fatal native-engine GC, single-instance first, fast-fail health poll**
- **test(desktop): de-flake migration temp dirs with a per-process sequence**


matthewevans:
All three Gemini findings addressed in 86fb649896: KILL is now re-verified against the spawn-record binary after the TERM grace (PID-reuse race); the BSD `ps -o comm=` match accepts a basename (PID still comes from our own spawn record); the Windows pre-delete before rename is removed — `std::fs::rename` already replaces the destination atomically there.

PR #6289, opened 2026-07-21
fix(parser): don't hoist a quoted grant's unless-pay onto the outer clause (Mage's Attendant)

JacobWoodson:
## The bug

Mage's Attendant creates a Wizard token whose granted activated ability is a **soft counter**:

```
create a 1/1 blue Wizard creature token with "{1}, Sacrifice this token:
Counter target noncreature spell unless its controller pays {1}."
```

The token entered with **no abilities at all** — silently, no `Unimplemented` marker. (Debug-spawning the token worked, because that path builds abilities from the catalog, not from this parse — which is why it looked fine in play.)

## Root cause

The clause-level unless-pay extractor (`extract_resolution_unless_pay_modifier`) scanned the **whole chunk** for `unless … pays`, reaching **inside the quoted granted ability**. It matched the Wizard's own `unless its controller pays {1}`, hoisted it onto the outer *create-token* spell, and in stripping that clause from the text it broke the closing `"` of the grant. The token's suffix then had an unbalanced quote, so ability extraction found nothing.

Confirmed by parse dump: the create-token ability came out with `unless_pay = Some(...)` and the token with **0** static abilities.

## Not a missing feature

The soft-counter machinery is complete. The granted activated ability parses correctly on its own — `classify_quoted_inner` produces a `GrantAbility` whose `Counter` effect already carries the `unless_pay` — exactly as **Mana Leak** and **Daze** do at spell level (both have populated `unless_pay` in `card-data.json`). The only defect was the outer scanner reaching into the quote.

## The fix

`extract_resolution_unless_pay_modifier` now masks double-quoted spans (`strip_double_quoted_spans`) before deciding whether an unless-pay belongs to this clause. When every `unless` in the chunk is inside a quoted grant, there's nothing to extract at this level — the pay belongs to the granted ability, where `classify_quoted_inner` already captures it — so the text is returned untouched and the quoted span stays balanced.

A chunk whose unless-pay is genuinely **outside** quotes is unaffected (the mask still shows it), so no soft-counter spell regresses.

Same class as the Reef Worm splitter bug (#6281): a scanner that wasn't quote-aware reaching inside a quoted granted ability.

## Test

`token_grant_soft_counter_keeps_unless_pay_on_the_token` asserts the outer create-token spell carries **no** `unless_pay` and the Wizard token carries the granted `Counter` ability with its `unless_pay` intact. Verified **red** without the fix (the pay is hoisted onto the create-token spell).

## Verification

- `cargo test -p engine --lib` — 17,387 passed, 0 failed
- `cargo test -p engine --test integration` (counter/token/unless) — 509 passed, 0 failed
- `cargo fmt --all`
- Parser combinator gates G / A pass (Family D detector could not run — `python3` unavailable in this environment)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed oracle-effect parsing so “unless” clauses inside double-quoted granted-ability text are no longer misidentified as clause boundaries or cost triggers.
  - Kept “unless-pay” conditions correctly attached to the granted token’s ability rather than moving to the surrounding create-token effect.
  - Improved effect text cleanup to maintain correct slice boundaries, including with non-ASCII characters.
- **New Features**
  - Added quote-aware masking for double-quoted spans that preserves original byte length to improve parsing reliability.
- **Tests**
  - Added regression coverage for quoted and nested “unless”/“unless-pay” scenarios, unterminated quotes, and Unicode byte-offset alignment.


matthewevans (changes requested):
**[MED]** Make `unless` extraction quote-aware rather than handling only the all-inner case.

The new early return applies only when the quote-stripped text contains no `unless`, but the generic `scan_preceded(&lower, tag("unless "))` still scans the original text. A token grant whose quoted ability has an inner `unless` before a real outer resolution `unless` will therefore still hoist the inner clause into the outer effect. Select an unquoted `unless` (for example with byte-preserving quote masking) and add a test containing both inner and outer clauses.

matthewevans:
**Blocked — the Mage's Attendant regression test is useful, but the current-head parse diff exposes two unreviewed card changes.**

## 🔴 Blocker

**[MED]** Make the `unless` selector quote-aware at the scan site and cover the affected quoted-ability class. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:30615-30621` bypasses extraction for every all-quoted `unless`, while the current-head `<!-- coverage-parse-diff -->` reports Musician changing from `if` to `Unsupported unless clause` and Whipgrass Entangler changing from direct `CantAttack, CantBlock` to a granted static ability. Musician's Oracle text contains `If it doesn't have "… destroy this creature unless you pay …"`; Whipgrass Entangler likewise puts its combat tax inside its quoted granted ability. Why it matters: the PR claims only Mage's Attendant and supplies no regression assertions for either additional card, so this broad parser-surface change is neither shown intentional nor pinned against future regressions. Suggested fix: select an unquoted `unless` rather than using the global early return, then add explicit expected-shape tests for Mage's Attendant, Musician, and Whipgrass Entangler (or explain and pin any intentional signature change).

## ✅ Clean

The Mage's Attendant fixture correctly demonstrates that its soft-counter payment belongs to the granted token ability.

**Recommendation: request changes; preserve quoted spans without changing unrelated quoted-ability parses.**

matthewevans:
@coderabbitai review

matthewevans:
@coderabbitai full review

matthewevans (changes requested):
**Request changes — the quoted-`unless` fix still needs its parser-class regression coverage.**

## 🔴 Blocker

**[HIGH]** Preserve the existing unterminated-quote fallback. Evidence: the proposed `scan_unquoted_preceded` returns `None` when its `delimited('"', …, '"')` parse fails, while `strip_double_quoted_spans` deliberately leaves an unterminated span unmasked. Why it matters: a legitimate earlier resolution-level `unless` then goes unselected in text such as `Destroy target creature unless its controller pays {1}. "unterminated`; the new scanner changes malformed-input behavior without a reach test. Suggested fix: retain the legacy unmasked scan fallback and add that case.

**[HIGH]** Add exact current-card shape tests for the parse-diff siblings and mixed placement. Evidence: the current parse-diff artifact identifies Musician and Whipgrass Entangler, but the test diff only exercises Mage's Attendant and primitive scanning. Why it matters: Musician's quoted upkeep-destruction payment and Whipgrass Entangler's quoted combat-tax static can silently change coverage/shape again, and a scanner-only test does not prove extraction selects an outer `unless` after an inner quoted one. Suggested fix: assert each card's expected parsed grant shape and an inner-then-outer `unless` extraction path.

## ✅ Clean

Mage's Attendant correctly identifies the original hoisting defect, and replacing the wildcard test match is directionally appropriate.

**Recommendation: retain the quoted-span regression, then add the two card-shape and mixed-placement tests before re-review.**

JacobWoodson:
Thanks — addressed both blockers plus the MED finding.

**General quote-awareness (was: all-inner guard).** Replaced the early-return guard with a byte-length-preserving mask (`mask_double_quoted_spans_preserving_len`) applied to all three `unless` scan sites in `extract_resolution_unless_pay_modifier`. An `unless` inside a quoted grant is masked out; an `unless` outside quotes keeps its text, and because the mask preserves byte length the matched offset still indexes the original `text` for the cleaned-effect slice. New test `inner_quoted_unless_does_not_shadow_a_real_outer_unless` covers the inner-then-outer case — verified red before the change (it extracted the inner grant's `you` payer instead of the outer `its controller`).

**Unterminated-quote fallback.** The mask mirrors `strip_double_quoted_spans` — an unterminated span passes its remainder through unmasked, so a legitimate resolution-level `unless` after a malformed span is still found. Covered by `unterminated_quote_still_exposes_outer_unless` and a `mask_preserving_len` unit test.

**Sibling card-shape tests.** Added `granted_ability_inner_unless_stays_on_the_grant`, a `parse_oracle_text` regression for both parse-diff siblings: Whipgrass Entangler (granted `can't attack or block unless its controller pays {1}` static — asserts the `UnlessPay` condition stays on the grant) and Musician (quoted upkeep `destroy … unless you pay {1}` — asserts no hoist). I also confirmed Whipgrass was previously mis-hoisting (`outer_unless_pay=true` on the base), so this change fixes it rather than regressing it; Musician's parse is unchanged by the masking.

Verification: `cargo test -p engine --lib` (17,486), targeted integration (counter/token/unless), clippy, fmt all green on the merged-main base.

matthewevans (changes requested):
**Request changes — the quote-aware scanner is the right seam, but the current head still has one offset-safety defect and two non-discriminating card-shape checks.**

## 🔴 Blocker

**[MED]** Keep mask offsets aligned with the original Oracle text. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:30679` builds `lower` with `to_lowercase()`, then `:30702` masks that buffer; the `before_unless` byte offset from the mask is used to slice the original `text` at `:30711`. Why it matters: Unicode lowercasing can change byte length (for example, a capital dotted I expands), so a valid quoted card-name/text prefix can make the later original-text slice select the wrong boundary or panic. Suggested fix: use a byte-preserving ASCII lowercase buffer for this ASCII grammar scan (or otherwise derive the slice position from the original text), and add a Unicode-prefix offset regression.

**[MED]** Make the Musician and Whipgrass regressions assert their actual parsed grants. Evidence: `crates/engine/src/parser/oracle_effect/tests.rs:47855-47864` selects any Whipgrass ability without `unless_pay` then searches its debug rendering, while `:47877-47880` uses `all(...)`, which passes on an empty ability list and proves neither Musician activation nor quoted upkeep grant was parsed. Why it matters: the two parse-diff cards can regress to an empty or unrelated shape while these tests remain green. Suggested fix: add reach guards for the expected activation and nested granted ability, then assert that Whipgrass’s specific granted static and Musician’s specific granted upkeep-destroy ability carry the inner `UnlessPay`, while their outer activations do not.

## 🟡 Non-blocking

**[LOW]** Make the `ContinuousModification` selection exhaustive. Evidence: `crates/engine/src/parser/oracle_effect/tests.rs:47873-47878` has a wildcard fallback. Why it matters: a new enum variant can silently bypass this test-selection logic. Suggested fix: enumerate the current non-`GrantAbility` variants as `None` so the compiler revisits this test when the enum grows.

## ✅ Clean

The prior quoted-`unless` scan-site finding is addressed on this head. The CodeRabbit dig/peek note targets `oracle_effect/mod.rs:21269-21284`, which is outside this PR’s changed hunks, so it is not a current-head finding for this change.

**Recommendation: request changes — preserve original-text offsets and replace the sibling tests with exact, non-vacuous grant-shape assertions; then re-review the new head.**

JacobWoodson:
Thanks — addressed both blockers; a note on the LOW.

**Offset safety (MED).** Switched the scan buffer to `to_ascii_lowercase()`. The grammar matched here is entirely ASCII, so the fold is identical for matching purposes but now byte-length-preserving, so the mask offset indexes the original `text` exactly. Your test case also surfaced a *second* offset defect I fixed: the cleaned effect was `text[..before_unless.trim_end().len()]`, but `before_unless` is a slice of the byte-preserving mask (quoted spans → spaces), so trimming it first dropped a quoted prefix in the effect (`… named "X" unless …`). Now all six return sites slice the original at the untrimmed offset and trim the original. New `unless_extraction_offsets_survive_unicode_case_fold` asserts the exact cleaned string for a quoted `İ` prefix before a real outer unless — verified **red** with `to_lowercase` (both the misalignment and the dropped quote).

**Non-vacuous sibling tests (MED).** Rewrote both with reach guards:
- Whipgrass: assert exactly one Activated ability, its `unless_pay` is None, and its sole `GrantStaticAbility` definition carries `StaticCondition::UnlessPay` (the per-Cleric condition) — no more debug-string search.
- Musician: assert exactly one ability, it is the `{T}` `PutCounter` activation, and its `unless_pay` is None. One correction to the premise: Musician's `If it doesn't have "…", it gains that ability` is a conditional self-referential grant that does **not** currently lower to a typed ability (the quoted upkeep-destroy grant is unimplemented), so there is no parsed `UnlessPay` shape to assert for it. The invariant this fix owns — and the test now guards non-vacuously — is that the inner `unless you pay {1}` is not hoisted onto the `{T}` activation. Happy to add the granted-shape assertion in a follow-up once that conditional grant is implemented.

**Exhaustive selector (LOW).** I kept a documented wildcard on the `ContinuousModification` match: the enum has ~50 variants, so a full exhaustive arm reads as noise in a test selector, and the `UnlessPay` assertion already prevents a silent shape regression. Glad to enumerate it if you'd prefer the compile-revisit guarantee regardless.

Verification: `cargo test -p engine --lib` (17,487), integration (unless/counter/token), clippy, fmt, parser gates G/A all green.

matthewevans (commented):
Reviewed current head `e838e1293e`: the quote-aware selector now masks complete quoted spans, uses byte-preserving ASCII folding for original-text offsets, and the remaining test selector asserts exactly one granted ability. The current parse-diff/required CI evidence is still pending, so this is not an approval.

matthewevans (approved):
Approved: current-head review confirms quoted `unless` extraction masks complete grants, preserves original-text byte offsets, and pins Mage’s Attendant plus the Musician/Whipgrass card paths. Required CI and the fresh parse-diff are green.

PR #6293, opened 2026-07-21
fix(tauri): Set WEBKIT_DMABUF_RENDERER_FORCE_SHM=1 to avoid nvidia graphics bug at no performance cost

lgray:
`WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` on linux machines to avoid blank-screen, but keep acceleration, with nvidia graphics drivers. Without this tauri renders a blank screen, and the game is near unplayable without GPU acceleration, so `WEBKIT_DISABLE_DMABUF_RENDERER=1` is not the correct solution.

This PR adds a line to the Tiltfile to ensure this environment variable is set.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -97,9 +97,6 @@ local_resource('tauri',
     cmd = 'cargo build -p phase-server && mkdir -p client/src-tauri/binaries && cp target/debug/phase-server ' + SIDECAR_DEST,
     serve_cmd = 'pnpm tauri:dev',
     serve_dir = 'client',
-    # Keep the dmabuf renderer (disabling it tanks in-game perf) but force
-    # shared-memory buffers, dodging the GPU-path glitches.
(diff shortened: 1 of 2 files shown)


lgray:

Addressed the review guidance: the env var is now set programmatically in `phase_tauri::run()` (7c6f8a714) — before the first webview is created, compile-gated `#[cfg(target_os = "linux")]`, and only when the variable isn't already present so an operator's value wins. The Tiltfile setting is removed in the same commit: with the fix in-app, dev builds (`pnpm tauri:dev`) exercise the exact code path production ships, instead of masking it behind a dev-only env.

Verification: `cargo check` and `cargo clippy -- -D warnings` clean on the `phase-tauri` crate; `cargo fmt` applied; pre-commit parser gates PASS (no engine/parser surface touched).

PR #6307, opened 2026-07-21
fix(engine): support "prevent N of that damage" static prevention shields

### LINKED ISSUE #5902: Heart-shaped herb does not reduce damage — [[heart-shaped herb]] should reduce incoming player damage by 1, but isn't a…
Reported in Discord: https://discord.com/channels/1485498006781427802/1526749271230255246/1526749271230255246

**Thread:** Heart-shaped herb does not reduce damage
**Discord thread id:** `1526749271230255246`
**Discord message id:** `1526749271230255246`
**Cards:** heart-shaped herb

## Summary
[[heart-shaped herb]] should reduce incoming player damage by 1, but isn't affecting it at all.

## Reported text

[[heart-shaped herb]] should reduce incoming player damage by 1, but isn't affecting it at all.

## Attachments

- [game-state-turn-25-2026-07-15T00-35-33-973Z.zip](https://cdn.discordapp.com/attachments/1526749271230255246/1526749271615995974/game-state-turn-25-2026-07-15T00-35-33-973Z.zip?ex=6a59790f&is=6a58278f&hm=7cfea6fdbf470f706e3110b872759a994c5e02b27ebc13a38a2546aef679e4d9&)

### heart-shaped herb
If a source an opponent controls would deal damage to you, prevent 1 of that damage.
{2}, {T}, Sacrifice this artifact: You may sacrifice a creature. If you do, return that card to the battlefield under its owner's control with three +1/+1 counters on it and you become the monarch.

---
<sub>report_id: `discord:1526749271230255246:1526749271230255246:0` · discord: `1526749271230255246/1526749271230255246`</sub>

rsnetworkinginc:
## Summary

Heart-Shaped Herb's own text ("If a source an opponent controls would deal
damage to you, prevent 1 of that damage.") never installed a replacement
effect at all: `parse_damage_prevention_replacement`'s amount extraction
only recognized `prevent all`, `prevent all but N`, `prevent the next N
damage`, and the literal `prevent that damage` — a bare `prevent N of that
damage` (no "all"/"all but"/"the next" qualifier) matched none of those
branches, so the whole function returned `None` and the ability silently
fell through. This matches the reported symptom exactly ("isn't affecting
it at all").

This idiom is shared by several other real cards beyond Heart-Shaped Herb
(Sphere of Purity, Orbs of Warding, Urza's Armor, Guardian Seraph, Djeru —
With Eyes Open, Protection of the Hekma, Daunting Defender, Shield of the
Realm, ...), so the fix is generic rather than card-specific.

## Fix

Reworked per @matthewevans's review (add-engine-variant Stage 1 =
`REFUSE_WITH_EXISTING_SLOT`): **no new `PreventionAmount` variant**. The
class reuses the existing shared per-event damage-subtraction authority —
`DamageModification::Minus { value: N }` on a `DamageDone` replacement, the
same continuous, non-consumed representation CR 702.64 Absorb already
synthesizes (`database::synthesis::build_absorb_replacement`) and applies
in Branch 1 of `damage_done_applier`.

- **Parser** (`parser/oracle_replacement.rs`): "prevent N of that damage"
  now builds the replacement with `damage_modification: Minus { value: N }`
  (leaving `shield_kind` as the default `None`), carrying the same
  recipient/source/combat filters as before. Shield-style prevention
  (all / all-but / depleting next / redirection-context "prevent that
  damage") still routes through `ShieldKind::Prevention`. Only a literal
  digit after "prevent " qualifies — not the pronoun "X" — so the unrelated
  dynamic "prevent X … where X is …" idiom (Cover of Winter class) is
  untouched, and the match is anchored on the trailing "of that damage".

- **Applier bookkeeping** (`game/replacement.rs`): the shared
  `DamageModification::Minus` path did not previously emit `DamagePrevented`.
  Fixed **at that shared authority** (per the review) rather than in a
  parallel arm: when a `Minus` modification actually reduces an event it now
  emits `DamagePrevented` — per-event outside a combat batch, and aggregated
  into the per-shield `combat_prevention_tally` inside one (a single
  post-batch `DamagePrevented` via `fire_combat_prevention_riders`, matching
  the `ShieldKind::Prevention` shields). CR 702.64 Absorb and the
  `Minus { value: u32::MAX }` prevent-all sentinel now emit it too.

CR 615.1a (prevention shield) + CR 702.64b (continuous, non-depleting) +
CR 609.7b (controller-scoped source restriction).

Closes #5902

## Testing

- 2 parser unit tests
  (`heart_shaped_herb_prevent_n_of_that_damage_is_minus_modification`,
  `bare_prevent_n_of_that_damage_generalizes_without_source_controller_clause`)
  pin the parsed `damage_modification`/`shield_kind`/`damage_source_filter`
  shape, including a sibling card with no controller qualifier to confirm
  the fix generalizes.
- 1 discriminating integration test (`issue_5902_heart_shaped_herb.rs`)
  drives the real `deal_damage::resolve` pipeline: damage from an
  opponent-controlled source is reduced by 1, the shield re-fires on a
  SECOND opponent-source event (non-consumed — proving it is not a depleting
  shield), damage from the controller's own source is unaffected, and
  `DamagePrevented` fires for the prevented point.

Local parity (isolated target dir): `cargo fmt --check` clean;
`cargo clippy --workspace --exclude phase-tauri --all-targets --features
engine/proptest -- -D warnings` clean; full engine lib suite 17567 passed /
0 failed; full integration suite 3836 passed / 0 failed.

## Summary by CodeRabbit

* **New Features**
  * Added support for the anchored grammar **“prevent N of that damage”**, enabling continuous, non-depleting prevention per qualifying damage event.

* **Bug Fixes**
  * Ensured plain arithmetic **minus** no longer triggers prevention events or prevented-amount bookkeeping.
  * Updated prevention provenance so **DamagePrevented** and prevented-amount handoff work correctly, including combat-damage batching behavior.
  * Corrected **Absorb**-style prevention and aligned typed prevention handling during import conversions.

* **Tests**
  * Added/expanded coverage for **Heart-Shaped Herb**, generalized **prevent** wording, and **Benevolent Unicorn** regression.

### DIFF
--- a/crates/engine/src/database/synthesis.rs
+++ b/crates/engine/src/database/synthesis.rs
@@ -3703,16 +3703,19 @@ pub(crate) fn entry_replacement_for_grant_static(
 
 /// CR 702.64a: Absorb N — "If a source would deal damage to this creature,
 /// prevent N of that damage." A continuous, self-recipient damage replacement:
-/// `DamageModification::Minus { value: N }` saturating-subtracts N from each
-/// damage event whose recipient is this creature (`valid_card: SelfRef`). It is
-/// NOT a consumed shield, so it re-applies to every source and every event
-/// independently (CR 702.64b). No new variant — mirrors the continuous
-/// damage-prevention statics (Benevolent Unicorn class) and the self-scoped
-/// `valid_card(SelfRef)` damage replacements (persistent prevention shields).
+/// `DamageModification::PreventionMinus { value: N }` saturating-subtracts N
+/// from each damage event whose recipient is this creature (`valid_card:
+/// SelfRef`). `PreventionMinus` is the CR 615 prevention provenance of the
+/// shared `Minus` subtraction authority — Absorb genuinely PREVENTS damage, so
+/// it emits `DamagePrevented` bookkeeping, unlike the plain-arithmetic
+/// `Minus` statics (Benevolent Unicorn class). It is NOT a consumed shield, so
+/// it re-applies to every source and every event independently (CR 702.64b),
+/// like the self-scoped `valid_card(SelfRef)` damage replacements (persistent
+/// prevention shields).
 fn build_absorb_replacement(n: u32) -> ReplacementDefinition {
     ReplacementDefinition::new(ReplacementEvent::DamageDone)
         .valid_card(TargetFilter::SelfRef)
-        .damage_modification(DamageModification::Minus { value: n })
+        .damage_modification(DamageModification::PreventionMinus { value: n })
         .description(format!(
             "CR 702.64a: Absorb {n} — if a source would deal damage to this creature, \
              prevent {n} of that damage."
@@ -3728,7 +3731,7 @@ fn is_absorb_replacement(r: &ReplacementDefinition, n: u32) -> bool {
         && matches!(r.valid_card, Some(TargetFilter::SelfRef))
         && matches!(
             r.damage_modification,
-            Some(DamageModification::Minus { value }) if value == n
+            Some(DamageModification::PreventionMinus { value }) if value == n
         )
 }
 
@@ -25200,9 +25203,11 @@ mod absorb_synthesis_tests {
     //! CR 702.64a shape tests: Absorb was parsed/typed but had no runtime.
     //! `synthesize_absorb` installs a continuous self-recipient `DamageDone`
     //! replacement that subtracts N from each incoming damage event
-    //! (`DamageModification::Minus { value: N }`, `valid_card: SelfRef`). The
-    //! continuous, non-consumed, per-source/per-event semantics (CR 702.64b) come
-    //! for free from `Minus`; CR 702.64c (each instance separate) is one
+    //! (`DamageModification::PreventionMinus { value: N }` — the CR 615
+    //! prevention provenance of the shared `Minus` subtraction —
+    //! `valid_card: SelfRef`). The continuous, non-consumed,
+    //! per-source/per-event semantics (CR 702.64b) come for free from the
+    //! shared subtraction arm; CR 702.64c (each instance separate) is one
     //! replacement per instance.
     use super::*;
     use crate::game::effects::deal_damage;
@@ -25287,9 +25292,9 @@ mod absorb_synthesis_tests {
         assert!(
             matches!(
                 r.damage_modification,
-                Some(DamageModification::Minus { value: 2 })
+                Some(DamageModification::PreventionMinus { value: 2 })
             ),
-            "CR 702.64a: prevent N (=2) of the damage"
+            "CR 702.64a: prevent N (=2) of the damage (prevention provenance)"
         );
     }
 
--- a/crates/engine/src/game/replacement.rs
+++ b/crates/engine/src/game/replacement.rs
@@ -1559,9 +1559,15 @@ fn damage_done_applier(
             applied,
         } = event
         {
-            // CR 615.1a: captured before the match consumes `modification` (the
-            // `Plus`/`SetTo` arms move their non-`Copy` payload out).
-            let is_minus_prevention = matches!(modification, DamageModification::Minus { .. });
+            // CR 615.1a: typed prevention provenance, captured before the match
+            // consumes `modification` (the `Plus`/`SetTo` arms move their
+            // non-`Copy` payload out). ONLY `PreventionMinus` — the CR 615
+            // prevention provenance of the shared subtraction — does prevention
+            // bookkeeping below; plain arithmetic `Minus` (Benevolent Unicorn's
+            // "that much damage minus 1") reduces the amount without preventing
+            // anything.
+            let is_minus_prevention =
+                matches!(modification, DamageModification::PreventionMinus { .. });
             let new_amount = match modification {
                 DamageModification::Double => amount.saturating_mul(2),
                 DamageModification::Triple => amount.saturating_mul(3),
@@ -1599,10 +1605,16 @@ fn damage_done_applier(
                     .max(0) as u32;
                     amount.saturating_add(added)
                 }
-                // CR 615.1 + CR 614.1a: Saturating subtract. `Minus { value: u32::MAX }`
-                // is the continuous prevent-all sentinel — yields 0 for any amount and
-                // is not consumed (continuous, not shield-style).
-                DamageModification::Minus { value } => amount.saturating_sub(value),
+                // CR 615.1 + CR 614.1a: Saturating subtract — the ONE shared
+                // subtraction authority for both provenances. `Minus` is plain
+                // arithmetic (CR 614.1a); `PreventionMinus` is CR 615 prevention
+                // provenance over the identical formula
+                // (`PreventionMinus { value: u32::MAX }` is the continuous
+                // prevent-all sentinel — yields 0 for any amount and is not
+                // consumed; continuous, not shield-style). Only the prevention
+                // provenance does the `DamagePrevented` bookkeeping below.
+                DamageModification::Minus { value }
+                | DamageModification::PreventionMinus { value } => amount.saturating_sub(value),
                 // CR 614.1a: Conditional — if amount < source's power, set to power.
                 // References the replacement source's (rid.source) post-layer power.
                 DamageModification::SetToSourcePower => {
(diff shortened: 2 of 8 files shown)


matthewevans (approved):
Maintainer review complete: rules, parser architecture, runtime discrimination, parse impact, and current-head CI verified.

matthewevans (changes requested):
## Maintainer review: changes requested

**[HIGH] The new `PreventionAmount::EachEvent` duplicates the existing parameterized per-event damage-subtraction authority.** Evidence: `crates/engine/src/database/synthesis.rs:3704-3715` already models CR 615 / CR 702.64 “prevent N of that damage” as `DamageModification::Minus { value: n }` and explicitly documents “No new variant”; `crates/engine/src/types/ability.rs:19711` defines that slot, and `crates/engine/src/game/replacement.rs:1609` applies its saturating subtraction to each event. This PR instead adds the same concept to the separate shield-amount axis and a parallel resolver arm.

Why it matters: the mandatory add-engine-variant Stage 1 result is `REFUSE_WITH_EXISTING_SLOT`. Keeping two semantic representations for the same continuous per-event subtraction splits parser and resolver authority and makes future prevention cards depend on which path happened to create them.

Suggested fix: reuse or generalize the existing `DamageModification::Minus` replacement pathway for this Oracle class and remove `PreventionAmount::EachEvent`. If that shared pathway does not currently emit the required `DamagePrevented` bookkeeping, fix the bookkeeping at that shared authority instead of introducing a parallel semantic representation.

Current-head CI is green, but that does not contradict the architecture finding: it verifies the duplicated path works, not that the duplicate should exist.

rsnetworkinginc:
Reworked per the 23:08Z review. Removed `PreventionAmount::EachEvent` entirely and re-emitted the bare "prevent N of that damage" class on the existing shared authority you pointed to: the parser now builds these shields as `DamageModification::Minus { value: N }` `DamageDone` replacements — the same continuous, non-consumed representation `database::synthesis::build_absorb_replacement` uses for CR 702.64 Absorb, applied in Branch 1 of `damage_done_applier` — carrying the same recipient/source/combat filters. No new prevention-amount variant, and no parallel resolver arm.

The one gap you flagged — the shared `Minus` pathway didn't emit `DamagePrevented` — is fixed at that shared authority rather than in a parallel path: `damage_done_applier`'s `Minus` branch now emits `DamagePrevented` when it reduces an event (per-event outside a combat batch, aggregated into `combat_prevention_tally` for the single post-batch event inside one), so Absorb and the `Minus { value: u32::MAX }` prevent-all sentinel get it too.

Local parity on an isolated target dir, all green: `fmt --check`; `clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings`; full engine lib 17567 passed / 0 failed; full integration 3836 passed / 0 failed.

matthewevans (changes requested):
**Changes requested — the revised parser now uses the correct shared `DamageModification::Minus` seam, but its bookkeeping treats ordinary arithmetic reductions as prevention.**

## 🔴 Blocker

**[HIGH]** `DamageModification::Minus` is not exclusively prevention. Evidence: [`crates/engine/src/game/replacement.rs:1564`](https://github.com/phase-rs/phase/blob/00cf1a716989f292e776a84f48747c3dc1ab1ce4/crates/engine/src/game/replacement.rs#L1564) classifies every `Minus` as prevention, while [`crates/engine/src/parser/oracle_replacement.rs:17403`](https://github.com/phase-rs/phase/blob/00cf1a716989f292e776a84f48747c3dc1ab1ce4/crates/engine/src/parser/oracle_replacement.rs#L17403) deliberately parses Benevolent Unicorn's “damage minus 1” arithmetic replacement into the same variant. Why it matters: that card now emits `DamagePrevented` and can fire prevention triggers even though it prevents no damage. Suggested fix: carry typed prevention provenance from the parser/synthesis producer through the shared `Minus` path, and add a Benevolent-Unicorn-style negative runtime regression.

**[HIGH]** The new `Minus` prevention path omits the replacement-follow-up amount binding. Evidence: [`crates/engine/src/game/replacement.rs:1666`](https://github.com/phase-rs/phase/blob/00cf1a716989f292e776a84f48747c3dc1ab1ce4/crates/engine/src/game/replacement.rs#L1666) emits/tallies prevention without the `last_effect_count` handoff that the shield path performs at [`replacement.rs:1911`](https://github.com/phase-rs/phase/blob/00cf1a716989f292e776a84f48747c3dc1ab1ce4/crates/engine/src/game/replacement.rs#L1911). Why it matters: a valid “damage prevented this way” continuation can read stale or missing `EventContextAmount`; the combat aggregate needs the same binding before its rider fires. Suggested fix: mirror the shield path's per-event and post-batch binding semantics, with a discriminating continuation regression.

**[MED]** The new Oracle grammar is not an anchored nom parser. Evidence: [`crates/engine/src/parser/oracle_replacement.rs:9834`](https://github.com/phase-rs/phase/blob/00cf1a716989f292e776a84f48747c3dc1ab1ce4/crates/engine/src/parser/oracle_replacement.rs#L9834) adds `starts_with` plus `scan_contains`, allowing a later, non-adjacent “of that damage” phrase to satisfy the branch. Why it matters: it violates the parser's combinator gate and can bind an unrelated suffix. Suggested fix: compose the existing numeric parser with an immediately following `tag("of that damage")` through the lower-case nom bridge.

## 🟡 Non-blocking

The required parse-diff evidence is stale for this head: its sticky comment is from 2026-07-21 23:01 UTC and still reports the removed `EachEvent` representation, while this head was committed at 2026-07-23 02:15 UTC. Please let the current card-data job publish a fresh artifact before re-review.

## ✅ Clean

The prior duplicate `PreventionAmount::EachEvent` was removed, and the Heart-Shaped Herb scenario now exercises repeated opponent-source damage plus the own-source boundary.

**Recommendation: address the typed provenance and continuation-binding gaps at the shared authority, replace the ad hoc parser scan with composable nom parsing, then push a fresh head for review.**

rsnetworkinginc:
Pushed e3960d9 addressing all three findings.

**Typed provenance.** The shared subtraction now carries typed provenance: `Minus` stays plain CR 614.1a arithmetic; the new `DamageModification::PreventionMinus` carries the CR 615 / CR 702.64 prevention provenance through the *same* applier arm (`Minus { value } | PreventionMinus { value } => saturating_sub`), so there is still exactly one subtraction authority. Provenance is a sibling variant tag rather than a field on `Minus` because the dormant `crates/mtgish-import` constructs `Minus { value }` literals that a new mandatory field would break, and CONTRIBUTING freezes that crate for contributor PRs.

Producer audit — every non-test `Minus` construction site:
- `parser/oracle_replacement.rs` "that much damage minus N" (Benevolent Unicorn) → arithmetic `Minus` (the only arithmetic producer)
- `parser/oracle_replacement.rs` bare "prevent N of that damage" → `PreventionMinus`
- `database/synthesis.rs` `build_absorb_replacement` (CR 702.64 Absorb) → `PreventionMinus`
- `crates/mtgish-import/src/convert/replacement.rs:418/:425` (PreventThatDamage → `Minus { u32::MAX }`; PreventSomeOfThatDamage → `Minus { N }`) — semantically prevention, but dormant and contributor-frozen, so left as-is; a one-word maintainer-side change if that crate is revived. No other construction sites exist.

**Amount binding.** The `PreventionMinus` site now mirrors the Branch 2 shield path: outside a batch it emits `DamagePrevented` and stamps the per-event prevented amount into `last_effect_count` (CR 615.5); inside a batch both the emission and the stamp defer to the post-batch aggregate via the per-replacement tally consumed by `fire_combat_prevention_riders` (CR 510.2 + CR 615.13), with the execute-template per-event stamp and the per-source-reflecting rider exclusion mirrored.

**Anchored grammar.** The amount branch is now `terminated(parse_number, tag(" of that damage"))` through `nom_parse_lower` — the anchor must be immediately adjacent to the number.

Each regression was verified red-first: with the provenance split reverted, arithmetic `Minus` emitted `DamagePrevented` (applier unit test + Benevolent Unicorn runtime test both fail); without the stamp, the continuation test observes a seeded stale count (`Some(999)` instead of `Some(2)`); under the previous scan, the anchor test binds a non-adjacent "of that damage". At this head: engine lib 17571 passed / 0 failed, integration 3837 passed / 0 failed; `cargo fmt --check`, `check-engine-authorities.sh`, and `check-parser-combinators.sh` all pass. The parser change does not alter the claimed coverage class (literal-digit "prevent N of that damage"); the card-data job will publish a fresh parse-diff artifact for this head.

matthewevans:
**Held — current-head implementation review is clean; enqueue remains gated on current CI evidence.**

## ✅ Clean

At `e3960d981a47464fa2b308133df1a6b4a9a919f9`, the typed `PreventionMinus` provenance keeps the shared subtraction authority while separating prevention bookkeeping from ordinary arithmetic `Minus`; the parser is now anchored with the existing nom combinators, and the added runtime cases exercise both provenances and the continuation binding.

## 🟡 External gate

The required Rust/card-data checks are still running, so the parse-diff sticky artifact is not yet current for this head. No contributor action is requested.

**Recommendation: wait for the current check run and fresh parse-diff, then route the clean head to the authorized handler.**

matthewevans (approved):
Approved at 4e46120b0af58ebfd5cdb53002be80faac7db622: current CI, parse-diff, and resolved review threads verified; the fixed conversion regression remains covered by the production conversion entry test.

PR #6318, opened 2026-07-22
feat(engine): CR 801 limited range of influence — snapshot foundation + targeting

minion1227:
## Summary

Makes `FormatConfig::range_of_influence` (CR 801) live. It had existed since introduction but was entirely dead — every mention `None`, no consumer anywhere in the engine. This ships it as **opt-in multiplayer infrastructure** plus its first real consumer (CR 801.4 targeting). No shipped format sets the field, so nothing changes for existing games until a format opts in.

Scoped deliberately narrowly (snapshot foundation + targeting), leaving the other CR 801 consumers and CR 809 Emperor as follow-ups.

Refs #6277

## Why this is snapshotted, not computed live — the crux

A prior attempt (#6279) was closed with a `CHANGES_REQUESTED` blocker for recomputing range **live** on every query. That is rules-incorrect, and the failure is invisible in two-player but changes legality mid-turn in exactly the multiplayer games the option exists for.

**CR 801.2c**: "The particular players within each player's range of influence are determined **as each turn begins**." The rule's own example: if a player leaves, the seats they separated come into range *at the start of the next turn*, not the instant they leave.

So range is snapshotted once per turn and every consumer reads the snapshot:

- `RangeOfInfluenceSnapshot` + `GameState::range_of_influence: Option<_>`. `None` is the unlimited case, so games not using the option pay a single `Option` check.
- `refresh_for_turn` is called from the turn-start authority (`turns::start_next_turn`) **and** from both game-start paths, so turn 1 is covered too — see the review-response note below.
- Distance is measured over **living seats** (`seat_order` minus `eliminated_players`), so an eliminated seat stops padding the gap between the players it separated — the other half of the 801.2c example, and the second thing #6279 got wrong.

Both of #6279's rejection reasons are fixed structurally; the turn-boundary snapshot *is* the design rather than a patch over a live computation.

## Review round — what changed since the first commit

**@matthewevans (blocker): the snapshot was never initialized for turn 1.** Correct, and the consequence was exactly as stated: `retain_targets_in_range` treats a `None` snapshot as "unlimited", so a range-limited format ignored CR 801.4 for the entire first turn. `refresh_for_turn` now runs in `start_game_with_starting_player` and `start_game_skip_mulligan` right after seat order is finalized, alongside the existing `start_next_turn` call. Pinned by a **legality** regression, not a field assertion: `on_turn_one_target_enumeration_already_rejects_out_of_range_players` starts a real game and asks `find_legal_targets` what is targetable on turn 1, with no hand-rolled `refresh_for_turn` anywhere. Reverting the seeding makes it fail with all five seats returned.

**CodeRabbit: the range filter was bypassable by early returns.** Correct — `retain_targets_in_range` ran only at the final return of `find_legal_targets_with_context`, so `SpecificPlayer`, player-only typed filters, stack-ability filters, `AttachedTo` and the `Or` recursion skipped CR 801.4. Restructured as an unfiltered inner function plus a wrapper that filters once, so no interior return can bypass it. `has_legal_target_with_context` now delegates to the filtered enumeration when the option is on, so "has a legal target" agrees with what can actually be chosen; the unlimited fast path is untouched.

**CodeRabbit: nonbattlefield zones must be scoped by owner.** Correct. Range follows the controller on the battlefield and stack (CR 801.2d, CR 108.4) and the owner everywhere else (CR 108.3) — a graveyard or hand card has no controller, so reading the residual field judged range by stale state. Pinned by two tests that give the card a deliberately *misleading* residual controller (in-range owner / out-of-range controller and the reverse), so reading `controller` outside the battlefield inverts both assertions.

**CodeRabbit: model the configured range per player.** Correct, and implemented rather than deferred. `FormatConfig::range_of_influence` is now `Option<RangeOfInfluence>` — a `default` plus per-player overrides — instead of `Option<u8>`. CR 801.2a states verbatim that "different players may have different ranges of influence", and the old model could not represent it. `RangeOfInfluence::uniform(n)` builds the common case (CR 809 Emperor, the CR 801.2a example tables); `for_player` resolves an override or the default. `snapshot` resolves the range **per observer** inside the seat loop, so reach is correctly directional — a range-2 seat can reach a range-0 seat that cannot reach back. Retyping the existing field rather than adding a parallel one keeps all 11 `None` constructors valid and leaves one authority for the option.

**Test coverage through the production path (both reviewers).** Every targeting assertion now goes through `find_legal_targets` — the funnel every spell/ability targeting query routes through — rather than the range helpers, covering player targets, battlefield object targets, and nonbattlefield (graveyard) card targets.

## Files changed

- `crates/engine/src/game/range_of_influence.rs`
- `crates/engine/src/game/targeting.rs`
- `crates/engine/src/game/engine.rs`
- `crates/engine/src/game/turns.rs`
- `crates/engine/src/game/mod.rs`
- `crates/engine/src/types/format.rs`
- `crates/engine/src/types/game_state.rs`
- `crates/engine/tests/integration/range_of_influence.rs`
- `client/src/adapter/types.ts`

## CR references

CR 801, CR 801.2, CR 801.2a, CR 801.2b, CR 801.2c, CR 801.2d, CR 801.4, CR 809, CR 108.3, CR 108.4

Each grep-verified against `docs/MagicCompRules.txt` before use. CR 801.3 / 801.5 / 801.6 / 801.7 appear only in the module's scope note naming the consumers this PR does **not** implement.

Method: not-applicable — this is engine infrastructure for an existing dead config field plus its first consumer, not a card add or a parser misparse. No parser changes.

## Track

Developer

## LLM

Model: claude-opus-4-8
Thinking: high

## Verification

- `cargo fmt --all` — clean.
- `cargo clippy -p engine --all-targets --features proptest -- -D warnings` — exit 0, no warnings.
- `cargo test -p engine --test integration range_of_influence` — **16 passed; 0 failed**.
- `pnpm exec tsc -b --noEmit` (client) — exit 0.
- **Fail-on-revert check** — with the game-start seeding removed *and* the zone-scoped owner lookup reverted to `object.controller`, the run is **12 passed; 4 failed**: `on_turn_one_target_enumeration_already_rejects_out_of_range_players` (returned all five seats), `the_first_turn_snapshot_is_seeded_at_game_start`, `target_enumeration_scopes_nonbattlefield_cards_by_owner`, and `a_graveyard_card_is_scoped_by_its_owner_not_a_stale_controller`. Both fixes restored afterwards. The tests fail without the fix, which is the point of them.
- Full `cargo test -p engine` (both shards) and the card-data/coverage gate are CI-owned; the prior head ran all of them green.

## Gate A

Gate A PASS head=40e09694911462eb60b0f7753cd5f0be8773a5a7 base=bdb1f3de5c37fc216d79c683c854be6448aad44c

## Anchored on

- `crates/engine/src/game/archenemy.rs` — the self-contained multiplayer-variant runtime module pattern (`cfg(test)`-free, tests in `tests/integration/`); `range_of_influence.rs` mirrors it exactly.
- `crates/engine/src/game/planechase.rs` — same seam, and its `reveal_starting_plane` call in `start_game_skip_mulligan` is the precedent for seeding per-game variant state at the game-start path rather than only at `start_next_turn`.

## Final review-impl

Final review-impl PASS head=40e09694911462eb60b0f7753cd5f0be8773a5a7

## Claimed parse impact

None. No parser changes — the combinator-purity gate is not engaged and parse-diff has no changed cards.

## Validation Failures

None.

## CI Failures

None.

## AI-contributor disclosure

Authored with an LLM (Claude Opus 4.8). The new `GameState::range_of_influence` field is covered by the `_gamestate_partition_is_total` guard with a rationale for why `PartialEq` needn't compare it (a per-turn snapshot derived from fields the partition already covers, never a per-iteration accumulator). Happy to rework anything, including widening scope to the remaining CR 801 consumers or CR 809 Emperor if you would prefer them together — kept narrow deliberately given #6279's history.

### DIFF
--- a/crates/engine/src/game/engine.rs
+++ b/crates/engine/src/game/engine.rs
@@ -10372,6 +10372,14 @@ pub fn start_game_with_starting_player(
     }
     state.phase = Phase::Untap;
 
+    // CR 801.2c: the players in each player's range of influence are determined
+    // "as each turn begins" — including turn 1. `start_next_turn` handles every
+    // later turn, but the first turn is set up here, so seed the snapshot now
+    // that seat order is final. Without this a limited-range game would treat
+    // every target as in range for the whole first turn. No-op unless the
+    // format opts in.
+    super::range_of_influence::refresh_for_turn(state);
+
     events.push(GameEvent::TurnStarted {
         player_id: starting_player,
         turn_number: 1,
@@ -10418,6 +10426,11 @@ pub fn start_game_skip_mulligan(state: &mut GameState) -> ActionResult {
     state.priority_player = starting_player;
     state.current_starting_player = starting_player;
     state.phase = Phase::Untap;
+
+    // CR 801.2c: seed the turn-1 range-of-influence snapshot (see the mirrored
+    // comment in `start_game_with_starting_player`). No-op unless the format
+    // uses a limited range.
+    super::range_of_influence::refresh_for_turn(state);
 
     events.push(GameEvent::TurnStarted {
         player_id: starting_player,
--- a/crates/engine/src/game/mod.rs
+++ b/crates/engine/src/game/mod.rs
@@ -133,6 +133,7 @@ pub mod printed_cards;
 pub mod priority;
 pub mod public_state;
 pub mod quantity;
+pub mod range_of_influence;
 pub mod replacement;
 pub mod replay;
 pub mod restrictions;
--- a/crates/engine/src/game/range_of_influence.rs
+++ b/crates/engine/src/game/range_of_influence.rs
@@ -1,0 +1,151 @@
+//! CR 801: Limited range of influence.
+//!
+//! An optional multiplayer rule capping how far, measured in player seats, a
+//! player can affect the game. It is opt-in per format through
+//! [`FormatConfig::range_of_influence`]; `None` means unlimited, which is every
+//! format shipped today, so games that do not use the option pay nothing beyond
+//! one `Option` check.
+//!
+//! # Why the range is snapshotted rather than computed live
+//!
+//! CR 801.2c: "The particular players within each player's range of influence
+//! are determined **as each turn begins**." The rule's own example makes the
+//! consequence explicit — if a player leaves the game, the seats that were
+//! separated by them come into range *at the start of the next turn*, not the
+//! instant they left.
+//!
+//! Computing range live would therefore be wrong in a way that is invisible in
+//! a two-player game and silently changes legality mid-turn in the multiplayer
+//! games the option exists for: a spell that was legal when it was cast could
+//! become illegal on resolution because an intervening player was eliminated in
+//! between. So the snapshot is the authority, refreshed exactly once per turn
+//! by [`refresh_for_turn`], and every CR 801 consumer reads it rather than
+//! recomputing.
+//!
+//! # Scope of this module
+//!
+//! This is the CR 801.2 foundation plus the CR 801.4 targeting consumer. The
+//! remaining consumers (CR 801.3 attacking, CR 801.5 choices, CR 801.6
+//! activation, CR 801.7 triggering) and the CR 809 Emperor variant are
+//! deliberately not implemented here — see the module's issue for the split. No
+//! shipped format sets `range_of_influence`, so nothing changes for existing
+//! games until a format opts in.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use serde::{Deserialize, Serialize};
+
+use crate::types::game_state::GameState;
+use crate::types::identifiers::ObjectId;
+use crate::types::player::PlayerId;
+
+/// CR 801.2c: the players within each player's range of influence, fixed at the
+/// start of the current turn.
+///
+/// Absent entirely when the game does not use the option, so `None` is the
+/// "unlimited range" case rather than a degenerate map naming every player.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RangeOfInfluenceSnapshot {
+    /// Keyed by the observing player; the value is every player inside that
+    /// player's range, including themselves (CR 801.2b).
+    within: BTreeMap<PlayerId, BTreeSet<PlayerId>>,
+}
+
+impl RangeOfInfluenceSnapshot {
+    /// CR 801.2 + CR 801.2b: whether `other` is inside `observer`'s range.
+    ///
+    /// A player absent from the snapshot (joined after it was taken, or already
+    /// eliminated when it was) is treated as out of range for everyone but
+    /// themselves — CR 801.2b holds unconditionally.
+    pub fn contains(&self, observer: PlayerId, other: PlayerId) -> bool {
+        if observer == other {
+            return true;
+        }
+        self.within
+            .get(&observer)
+            .is_some_and(|players| players.contains(&other))
+    }
+}
+
+/// CR 801.2 + CR 801.2c: compute the range-of-influence snapshot for the CURRENT
+/// seating, or `None` when the game does not use the option.
+///
+/// Distance is measured over the seats still in the game: CR 801.2c's example
+/// has a departed player's neighbours become adjacent, so an eliminated seat
+/// must not keep padding the distance between the players it used to separate.
+/// Measuring over living seats is what makes that example come out right.
+pub fn snapshot(state: &GameState) -> Option<RangeOfInfluenceSnapshot> {
+    let range = state.format_config.range_of_influence?;
+
+    // CR 801.2c: seats still in the game, in seat order. `seat_order` is the
+    // authority on seating; `eliminated_players` removes the departed.
+    let living: Vec<PlayerId> = state
+        .seat_order
+        .iter()
+        .copied()
+        .filter(|player| !state.eliminated_players.contains(player))
+        .collect();
+
+    let seats = living.len();
+    let mut within: BTreeMap<PlayerId, BTreeSet<PlayerId>> = BTreeMap::new();
+    for (index, &observer) in living.iter().enumerate() {
+        let mut in_range = BTreeSet::new();
+        for (other_index, &other) in living.iter().enumerate() {
+            if seat_distance(index, other_index, seats) <= u32::from(range) {
+                in_range.insert(other);
+            }
+        }
+        // CR 801.2b: a player is always within their own range, even if the
+        // configured range is 0.
+        in_range.insert(observer);
+        within.insert(observer, in_range);
+    }
+
+    Some(RangeOfInfluenceSnapshot { within })
+}
+
+/// CR 801.2: distance between two seats at a table, measured the short way
+/// around. Players sit in a circle, so the two seats at the ends of the seat
+/// order are adjacent.
+fn seat_distance(a: usize, b: usize, seats: usize) -> u32 {
+    if seats == 0 {
+        return 0;
+    }
+    let forward = (a + seats - b) % seats;
+    let backward = (b + seats - a) % seats;
+    forward.min(backward) as u32
+}
+
+/// CR 801.2c: re-determine every player's range because a turn is beginning.
+///
+/// Called from the single turn-start authority so the snapshot advances in
+/// lockstep with turns — the rule's example (a departure taking effect at the
+/// next turn start) falls out of calling this here and nowhere else.
+pub fn refresh_for_turn(state: &mut GameState) {
+    state.range_of_influence = snapshot(state);
+}
+
+/// CR 801.2 + CR 801.2b: whether `other` is within `observer`'s range.
+///
+/// `true` for every pair when the game does not use the option, so callers can
+/// gate unconditionally without first asking whether the option is on.
+pub fn player_in_range(state: &GameState, observer: PlayerId, other: PlayerId) -> bool {
+    match &state.range_of_influence {
+        Some(snapshot) => snapshot.contains(observer, other),
+        None => true,
+    }
+}
+
+/// CR 801.2d: an object is within a player's range if its controller is.
+///
+/// An object that no longer exists is reported out of range rather than in it,
+/// so a stale id can never widen what a player may reach.
+pub fn object_in_range(state: &GameState, observer: PlayerId, object: ObjectId) -> bool {
+    if state.range_of_influence.is_none() {
+        return true;
+    }
+    state
+        .objects
+        .get(&object)
+        .is_some_and(|object| player_in_range(state, observer, object.controller))
+}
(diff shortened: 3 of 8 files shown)


matthewevans (changes requested):
**Verdict: Request changes — the range snapshot is never initialized for turn 1.**

## 🔴 Blocker

- **[HIGH]** Initialize the limited-range snapshot at game start. Evidence: `GameState::new` initializes `range_of_influence` to `None` (`crates/engine/src/types/game_state.rs:15389`); both `start_game_with_starting_player` and `start_game_skip_mulligan` set up and emit turn 1 without calling `refresh_for_turn` (`crates/engine/src/game/engine.rs:10343-10445`); and `retain_targets_in_range` returns without filtering when the snapshot is `None` (`crates/engine/src/game/targeting.rs:399-415`). Why it matters: for a format that opts into range 1, every target is legal for the entire first turn. That conflicts with CR 801.2c, which determines the particular players in range “as each turn begins.” Suggested fix: refresh after the final game-start seat-order setup in both start paths (or move both through one shared start-of-turn authority), and add a real first-turn target-enumeration/legality regression—not only a helper test that calls `refresh_for_turn` manually.

## ✅ Clean

- The snapshot approach, eliminated-seat handling, and the target-enumeration seam are otherwise sensible for the stated CR 801.2/801.4 slice.

**Recommendation: fix the first-turn initialization and demonstrate target rejection through the production targeting path, then request re-review.**

minion1227:
Pushed `46020f2` addressing the full round — the maintainer blocker, all four CodeRabbit findings, and a real test bug CI caught. Every point was valid.

## Maintainer blocker — turn-1 init (HIGH)

Correct and important: `refresh_for_turn` ran only in `start_next_turn`, so turn 1 — set up by the game-start path — had no snapshot, and a limited-range format treated every target as in range for the whole first turn. Now seeded in both `start_game_with_starting_player` and `start_game_skip_mulligan`, right after seat order is finalized. Added `the_first_turn_snapshot_is_seeded_at_game_start`, which starts a real game and asserts the snapshot is real (reaches seat 1, not seat 2), and `target_enumeration_drops_out_of_range_players_through_the_real_pipeline`, which drives `find_legal_targets` and asserts CR 801.4 rejection — not helper-only tests.

## CodeRabbit

1. **Targeting filter bypassed by early returns** — you were right, my "single funnel" claim was false. `retain_targets_in_range` ran only at the final return, so `SpecificPlayer`, player-only typed, stack-ability, and the `Or` recursion skipped CR 801.4. Split into an unfiltered inner + a wrapper that filters once, covering every path. `has_legal_target_with_context` delegates to the filtered enumeration when the option is on, keeping the unlimited-format fast path untouched.
2. **Owner vs controller by zone** — fixed: controller on battlefield/stack (CR 801.2d), owner elsewhere (CR 108.3), since a graveyard/hand card has no controller.
3. **Per-player range (CR 801.2a)** — documented as the implemented subset. The snapshot is already observer-keyed, so per-player ranges are a config-layer extension (a per-seat range map feeding the same loop), not a redesign. This slice covers Emperor (uniform range 1).
4. **Production-path tests** — added, see above.

## Test bug CI caught

The departure regression used a 3-seat table, where the circle wrap makes every seat distance-1 from every other, so the "Rob between them" precondition was false. Fixed to 5 seats, matching the CR example.

## Verification honesty (unchanged from the PR body)

`cargo check -p engine --all-targets` is clean (rc=0). I still cannot *run* the engine test suite in my sandbox — the ~130MB test binary link exceeds a hard 10-minute command cap — so the run is CI's. I diagnosed the shard-2 failure by reasoning about the seat topology rather than a local run; CI will confirm the fix.

matthewevans (changes requested):
**Verdict: Request changes — the new head fixes turn-one initialization, but the limited-range model still omits a required dimension.**

## 🔴 Blocker

- **[HIGH]** Represent a range for each observing player before presenting this as the CR 801 foundation. Evidence: `crates/engine/src/game/range_of_influence.rs:74-85` explicitly documents the implementation as a uniform subset, and `snapshot` at `:91-112` reads one `FormatConfig::range_of_influence` value for every observer. CR 801.2a says, verbatim, “Different players may have different ranges of influence.” Why it matters: any legal asymmetric configuration cannot be represented, so each targeter can be judged against the wrong range. Suggested fix: make the format/game configuration supply a per-player range (with a uniform convenience default), consume that value inside the observer loop, and add an asymmetric-seat snapshot/targeting regression.

## 🟡 Non-blocking

- The new production test exercises only player targets. `object_in_range` now has battlefield/stack versus other-zone ownership behavior at `crates/engine/src/game/range_of_influence.rs:158-170`, but `crates/engine/tests/integration/range_of_influence.rs:104-173` never drives either object case through `find_legal_targets`. Add those sibling assertions with the per-player fix so the centralized funnel cannot regress silently.

## ✅ Clean

- The turn-one refresh and outer `find_legal_targets_with_context` wrapper correctly address the prior first-turn and early-return defects; the current parse-diff is a clean no-change artifact.

**Recommendation: complete the observer-specific configuration and the object-target production coverage, then request a current-head re-review.**

minion1227:
Pushed `96a554c` — per-player range implemented, plus the object-target coverage.

## CR 801.2a per-player range (blocker)

You were right to reject the scope-documentation. `FormatConfig::range_of_influence` is now `Option<RangeOfInfluence>` — a `default` plus per-player overrides — rather than `Option<u8>`. Retyping instead of adding a parallel field keeps all 11 existing `None` constructors valid and leaves a single authority for the option:

```rust
pub struct RangeOfInfluence {
    pub default: u8,
    pub per_player: BTreeMap<PlayerId, u8>,   // CR 801.2a overrides
}
RangeOfInfluence::uniform(n)   // CR 809 Emperor, and CR 801.2a's range-1/range-2 tables
config.for_player(observer)    // override, else default
```

`snapshot` resolves the range **inside** the observer loop, so reach is correctly directional. The new regression pins exactly that: a range-2 seat and a range-0 seat at one table, asserting seat 0 reaches seat 2 while seat 1 reaches nobody but itself (CR 801.2b still holds at range 0), and that seat 2 does *not* reach back into seat 0. A match-wide value cannot represent any of that.

## Object-target production coverage (non-blocking)

Added, driven through `find_legal_targets` rather than the helpers:
- `object_targets_are_range_filtered_through_the_real_pipeline` — a battlefield object controlled by an in-range seat stays targetable; one controlled by an out-of-range seat is filtered (CR 801.4 + CR 801.2d).
- `a_graveyard_card_is_scoped_by_its_owner_not_a_stale_controller` — a graveyard card whose residual `controller` names an **in-range** seat while its owner is out of range must stay out of range, so the zone-scoped ownership rule can't silently regress.

## Note

While preparing this, the zone-scoped ownership fix from `46020f2` was clobbered in my shared working tree (another agent is editing `client/src/adapter/*` in the same checkout). I restored it from the commit and re-verified; that restore is included here. I also had to force-push after my local branch was rebased onto a newer base out from under me — I verified first that every commit on the remote was mine and that the new head is a strict content superset, so nothing was dropped.

`cargo check -p engine --all-targets` is clean (rc=0). Test execution remains CI's — the ~130MB engine test-binary link exceeds my sandbox's 10-minute command cap. Shard 2/2 went green on the previous head after I fixed the seat-topology bug in my own departure test, so the loop is working.

minion1227:
Noting the timing rather than contesting the decision: this was closed at 03:41 and the per-player range fix landed at 03:45, so the head you closed on is not the current one. `96a554c` implements the CR 801.2a blocker as specified — per-player ranges resolved inside the observer loop, with an asymmetric-seat regression — plus the object-target production coverage you flagged as non-blocking. Current head is 14/14 green, both Rust test shards included.

Entirely your call whether that is worth reopening or whether you would rather this came back as a fresh, smaller PR (for instance the CR 801.2 snapshot alone, with the CR 801.4 targeting consumer split out). I am happy either way, and happy to leave CR 801 alone if you would prefer to own the design given two attempts have now closed.

One thing worth flagging for whoever picks this up next: the CR 801.2c turn-boundary snapshot and the turn-1 seeding are the two non-obvious parts. Recomputing range live passes casual review and is wrong, and `start_next_turn` alone misses turn 1 entirely — both were real defects found here.

minion1227:
All four review points are addressed on `40e0969`. Short version:

**@matthewevans — turn-1 initialization (blocker).** Confirmed and fixed. `refresh_for_turn` now runs in both `start_game_with_starting_player` and `start_game_skip_mulligan` right after seat order is finalized, alongside the existing `start_next_turn` call. Per the ask, the regression is a **legality** test rather than a snapshot-field assertion: `on_turn_one_target_enumeration_already_rejects_out_of_range_players` starts a real game and asks `find_legal_targets` what is targetable on turn 1, with no hand-rolled `refresh_for_turn` anywhere. Removing the seeding makes it fail with all five seats returned.

**CodeRabbit — filter bypassable by early returns.** Confirmed and fixed. `retain_targets_in_range` ran only at the final return, so `SpecificPlayer`, player-only typed filters, stack-ability filters, `AttachedTo` and the `Or` recursion skipped CR 801.4 entirely. Split into an unfiltered inner function plus a wrapper that filters once, so no interior return can bypass it. `has_legal_target_with_context` delegates to the filtered enumeration when the option is on; the unlimited fast path is untouched.

**CodeRabbit — owner vs controller outside the battlefield.** Confirmed and fixed. Controller on battlefield/stack (CR 801.2d, CR 108.4), owner elsewhere (CR 108.3). Both regressions give the card a deliberately *misleading* residual controller, so reading `controller` outside the battlefield inverts the assertions rather than merely failing to exercise them.

**CodeRabbit — per-player range (CR 801.2a).** Confirmed and implemented rather than deferred. `range_of_influence` is now `Option<RangeOfInfluence>` (a `default` plus per-player overrides), and `snapshot` resolves the range **per observer** inside the seat loop, so reach is directional — a range-2 seat can reach a range-0 seat that cannot reach back. Retyping the existing field rather than adding a parallel one keeps all 11 `None` constructors valid and leaves a single authority for the option.

Every targeting assertion now runs through `find_legal_targets` — player targets, battlefield object targets, and nonbattlefield (graveyard) card targets. The client's `FormatConfig` type was updated to mirror the new serialized shape, since the engine now emits an object where TS still declared `number | null`.

Verification on this head: `clippy -p engine --all-targets --features proptest -D warnings` exit 0, `cargo test -p engine --test integration range_of_influence` 16 passed / 0 failed, client `tsc -b --noEmit` exit 0. Fail-on-revert checked explicitly — backing out the game-start seeding and the owner scoping turns that run into 12 passed / 4 failed, naming exactly the four tests written for those two fixes.

Ready for re-review.

minion1227:
Following up on my own last message — I hadn't seen that #6277 was closed as `not_planned` a few seconds after this PR. That answers the question I asked above, so treat "ready for re-review" as withdrawn: I'm standing down on CR 801 rather than resubmitting.

Leaving the notes in case they're useful to whoever owns this later: the turn-boundary snapshot (CR 801.2c) has to be seeded at both game-start paths and not only `start_next_turn`, the CR 801.4 filter belongs on a wrapper around the target enumeration rather than at its final return, and range must follow `owner` outside the battlefield/stack (CR 108.3). Sorry for the churn on this one.

PR #6296, opened 2026-07-21
fix(engine): don't force a MustAttack carrier to attack itself (#5952)

andriypolanski:
Closes #5952

## Summary

Discord report: **Fumiko the Lowblood** appears to "goad" herself — her static "Creatures your opponents control attack each combat if able." was forcing Fumiko (a creature *you* control) to attack.

The parse is correct: the static is a `MustAttack` with `affected: Typed { controller: Opponent }`. The bug is in **runtime enforcement**, not the AST.

Root cause: the must-attack logic had a `has_local_must_attack` fast-path that reported a requirement for any creature merely **carrying** a `MustAttack` static — checking only `sd.mode == StaticMode::MustAttack` and **ignoring the static's `affected` filter**. Fumiko carries an opponent-scoped `MustAttack`, so she satisfied "local must-attack" against herself (CR 508.1d + CR 109.5).

Fix: `check_static_ability` / `check_static_ability_sources` are already the single authority — they run the full applicability predicate (`affected` filter, condition, per-player gates) against each carrier and return the carrier id, correctly handling both an intrinsic self static (Juggernaut's `SelfRef`) and a remote scope (Fumiko's `Opponent`). Remove the redundant, buggy `has_local_must_attack` duplicate and route both call sites through that authority, dropping the now-unused `obj` parameter from `must_attack_sources_gated`.

## Changes

- **`crates/engine/src/game/combat.rs`** — delete `has_local_must_attack`; the enforcement bool (`creature_must_attack_with_attackable_players_gated`) and the source collector (`must_attack_sources_gated`) now rely solely on `check_static_ability` / `check_static_ability_sources`, which honor the static's `affected` filter. Drop the unused `obj` parameter from `must_attack_sources_gated`.
- **`crates/engine/tests/integration/fumiko_the_lowblood_5952.rs`** (+ `main.rs` mod) — runtime regression through the real parse → layer → combat-enforcement pipeline: Fumiko and her controller's other creatures are NOT forced during their own combat, while a creature an opponent of Fumiko's controller controls IS forced (positive control).

## Test Plan

- [ ] `cargo fmt --all -- --check`
- [ ] `cargo clippy -p engine --all-targets --features engine/proptest -- -D warnings`
- [ ] `cargo test -p engine` (full suite: 21,174 pass)
- [ ] `cargo test -p engine --test integration fumiko`
- [ ] `cargo test -p engine --test integration curse_of_the_nightly_hunt_must_attack` (sibling cross-permanent MustAttack, still green)

[...]

PR #6319, opened 2026-07-22
fix(engine): spelled-out goad requirements must not goad (CR 701.15a)

lgray:

## Summary

The parser lowered two cards that spell out the goad *effect* without the goad keyword — Kardur, Doomscourge and Maximum Carnage chapter I — to `Effect::GoadAll`, which stamped the goad **designation** and marked only the creatures present at resolution. Both are wrong: per CR 701.15a only a spell or ability that *goads* makes a creature goaded, and per CR 611.2c the affected set of a rules-modifying continuous effect stays dynamic. This lowers the compound to `Effect::GenericEffect` carrying one `StaticDefinition` with `StaticMode::MustAttack` + a new nullary `StaticMode::MustAttackAwayFromSource`, so the requirement is created without the designation and the affected set stays live.

Both cards' official rulings say exactly this:

- **Maximum Carnage** (2025-09-19): *"Although the effects of the first chapter ability are the same as the goad keyword action, that ability doesn't cause any creatures to become goaded. Effects that refer to 'goaded creatures' won't apply."*
- **Maximum Carnage** (2025-09-19): *"affects all creatures until your next turn, regardless of whether or not they were on the battlefield at the time the ability resolved."*
- **Kardur, Doomscourge** (2021-02-05): *"affects all creatures your opponents control, including any that enter the battlefield after the ability resolves."*

Cards that read the designation (Serene Sleuth, Hot Pursuit, Baeloth Barrityl, The Rani, The Kami Knight's "can't be goaded") therefore stop matching these two — which is the rules-correct outcome. Real goad cards (Agitator Ant, Disrupt Decorum, Marisi, Day of the Moon, …) use the keyword and keep lowering to `Goad`/`GoadAll`; a regression test pins that.

Method: /engine-implementer

## CR references

- **CR 701.15a** — only a spell/ability that *goads* makes a creature goaded (the load-bearing rule: these two cards create the requirement without the designation).
- **CR 701.15b** — "Goaded is a designation a permanent can have"; the two clauses of the goad requirement.
- **CR 701.15c** — each distinct avoided player is an additional requirement.
- **CR 611.2c** — a resolution-created continuous effect that modifies no characteristics modifies the rules of the game ⇒ dynamic affected set.
- **CR 508.1d** — the attack-requirement maximisation solver.
- **CR 109.5** — "you" is the controller of the object; for a triggered ability, the controller when it triggered (the player reference is snapshotted, the object set is not).

Every CR number in the diff was grep-verified against `docs/MagicCompRules.txt` before being written, and re-verified by two independent review agents.

## Verification

| Command | Result |
|---|---|
| `cargo fmt --all` | clean, zero diff |
| `scripts/check-parser-combinators.sh 23824afe7` | Gate A PASS + Gate G PASS |
| `cargo clippy --workspace --all-targets --features engine/proptest -- -D warnings` | exit 0, zero warnings |
| `cargo test -p engine --test integration bbfu7` | 13 passed / 0 failed |
| `cargo test -p engine --lib` | 17489 passed / 0 failed / 6 ignored |
| `cargo test-all` | 23463 run, 23455 passed, 8 failed — all 8 pre-existing, see below |
| `cargo coverage` | `supported_cards` 31502 → 31504 (+2) |
| card-data blast radius (regen at base vs head) | exactly **2** keys changed, 0 added, 0 removed |
| `FORGE_TEST_FULL_DB=1 cargo test -p engine --lib ordering_parity_sweep` | sweep line **byte-identical** to base (`unexplained=18` both sides) |

**The 8 `cargo test-all` failures are pre-existing on `main`, proven from git objects without a build.** They are all in `crates/mtgish-import`, a crate this PR does not touch (`git diff --stat 23824afe7...HEAD -- crates/mtgish-import` is empty). For `every_list_field_is_in_ordering_manifest` specifically: at `23824afe7` itself, `ORDERING_MANIFEST` in `crates/mtgish-import/src/diff/ordering.rs` contains **zero** entries for `mode_pawprints`, `spellbook` or `extra_core_types`, while those `Vec` fields **do** exist in `crates/engine/src/types/ability.rs` at the same commit — so the test fails on `23824afe7` by construction. This PR adds no `Vec<...>` struct field at all (the single `Vec<` line it adds is a function-local `let`).

`--all-features` additionally surfaces 16 compile errors under the `forge` feature; all six files involved are byte-identical to `23824afe7` (`git diff --quiet 23824afe7 HEAD -- <f>` for each), and `forge` is not in the repo's gate alias (`clippy-strict = "clippy --all-targets -- -D warnings"`).

### Non-vacuity evidence (measured revert-probes, not asserted)

The acceptance suite began life as a **probe gate**: before any design work, it was run at the pre-fix tip and measured **1 passed / 4 failed**, so every behavioural leg was proven to flip. One leg (`requirement_binds_a_creature_that_enters_after_resolution`) initially passed *vacuously* because a pre-existing goaded creature's own requirement already made the empty declaration illegal; removing that creature flipped it to FAIL. Measured, not assumed.

A late independent review found that **no runtime test discriminated the `MustAttack` half** of the grant — every "declining to attack is illegal" leg was also satisfied by the away-from term alone, so a silent regression in the `MustAttack` path (which #6296 had just rewritten) would have gone undetected. `grafted_must_attack_half_reaches_the_combat_authority` closes that hole, with a measured revert-probe: deleting **only** the `AddStaticMode { mode: StaticMode::MustAttack }` element from `must_attack_away_static_definition()` makes it fail (`left: []`, `right: [ObjectId(2)]`) along with the parse-shape assertion in `neither_card_lowers_to_goad`, which pins the same two-element `modifications` vec; the other 11 tests stay green. Measured: `11 passed; 2 failed`. Restoring the element returns 13/13; `imperative.rs` md5 verified identical before and after. Two independent agents ran this probe separately and got the same counts.

## Gate A

Gate A PASS head=2a91e56ee582f6bceed49ddbe35f5e0ded7601f6 base=23824afe75218442ce199f411169ed5462421585

## Anchored on

- `crates/engine/src/game/combat.rs:2631` — `must_attack_player_directives_for_creature`, the existing per-object `active_static_definitions` scan that the new `players_to_attack_away_from_gated` authority mirrors exactly (same seam, same unfiltered-scan shape, deliberately left unchanged by #6296).
- `crates/engine/src/game/layers.rs` — the existing `ContinuousModification::AddStaticMode` arm that grafts `StaticDefinition::new(mode).affected(TargetFilter::SelfRef)` onto recipients; the new mode rides this path rather than introducing one.

## Final review-impl

Final review-impl PASS head=2a91e56ee582f6bceed49ddbe35f5e0ded7601f6

## Claimed parse impact

- Kardur, Doomscourge
- Maximum Carnage

Both were **unsupported before this change** (one swallowed clause each) and become supported: `supported_cards` 31502 → 31504. Card-data blast radius is exactly these two keys.

## Disclosed follow-ups (not fixed here)

1. `crates/phase-ai/src/policies/effect_classify.rs` classifies the new mode via `_ => Contextual` where `MustAttack` is `Harmful`. Aura-scoped and unreachable today; fixing it would require `cargo ai-gate` plus a paired-seed baseline refresh, which is disproportionate for an unreachable path.
2. `players_to_attack_away_from_gated` and its sibling `must_attack_player_directives_for_creature` both scan `active_static_definitions` without an `sd.affected` check. Safe today for both (grafted-only producers, always `affected: SelfRef`), and documented in-code with the condition under which it must change — but it is a latent class-wide trap that belongs in one follow-up covering *both* siblings, not a one-sided fix here.

## Summary by CodeRabbit

* **New Features**
  * Added support for “must attack a player other than you” via a dedicated combat restriction mode.
  * Updated oracle parsing so BB-FU7-style “away-from” text becomes combat requirements rather than goading behavior.
* **Bug Fixes**
  * Corrected attacker-legality and requirement derivation for away-from restrictions, including proper controller anchoring and dynamic affected-object evaluation.
* **Tests**
  * Added a new integration test covering parsing, eligibility, timing/expiration, edge cases, and verifying goad cards still set goad designations.

### DIFF
--- a/crates/engine/src/ai_support/filter.rs
+++ b/crates/engine/src/ai_support/filter.rs
@@ -1029,6 +1029,7 @@ impl LegalityPoisonGates {
                     | StaticMode::MustAttack
                     | StaticMode::MustAttackPlayer { .. }
                     | StaticMode::Goaded
+                    | StaticMode::MustAttackAwayFromSource
                     | StaticMode::CanAttackWithDefender
                     | StaticMode::MaxAttackersEachCombat { .. }
                     | StaticMode::CombatAlone { .. }
--- a/crates/engine/src/analysis/resource.rs
+++ b/crates/engine/src/analysis/resource.rs
@@ -2075,6 +2075,7 @@ fn static_mode_references_growing_class(mode: &crate::types::statics::StaticMode
         | StaticMode::MustBeBlocked { .. }
         | StaticMode::MustBeBlockedByAll { .. }
         | StaticMode::Goaded
+        | StaticMode::MustAttackAwayFromSource
         | StaticMode::CombatAlone { .. }
         | StaticMode::CantCrew
         | StaticMode::CantPhaseIn
--- a/crates/engine/src/game/combat.rs
+++ b/crates/engine/src/game/combat.rs
@@ -4236,7 +4236,7 @@ pub(crate) fn players_to_attack_away_from_gated(
 ) -> HashSet<PlayerId> {
     let mut players = goading_players_for_creature_gated(state, creature_id, has_goad_static);
     if let Some(obj) = state.objects.get(&creature_id) {
-        // CR 611.2c: the avoided player is the anchor snapshotted at graft time.
+        // CR 109.5: the avoided player is the anchor snapshotted at graft time.
         // `unwrap_or(obj.controller)` is a defensive default with NO producer
         // today — the parser is the sole producer of this mode and always routes
         // through the transient graft, which stamps the anchor (layers.rs). It is
--- a/crates/engine/src/game/combat.rs
+++ b/crates/engine/src/game/combat.rs
@@ -2730,12 +2730,16 @@ fn creature_must_attack_with_attackable_players_gated(
             StaticMode::MustAttack,
             &static_target_ctx(obj_id),
         );
-    // CR 701.15b: Goaded creatures must attack each combat if able.
-    let is_goaded = !goading_players_for_creature_gated(state, obj_id, gates.has_goad).is_empty();
+    // CR 508.1d + CR 701.15b (first clause): a creature under an "attacks a
+    // player other than X if able" requirement — whether from the goad
+    // designation or from the spelled-out `MustAttackAwayFromSource` grant —
+    // also attacks each combat if able.
+    let must_attack_away =
+        !players_to_attack_away_from_gated(state, obj_id, gates.has_goad).is_empty();
     let has_attackable_must_attack_player = must_attack_players_for_creature(state, obj)
         .iter()
         .any(|player| attackable_players.contains(player));
-    if !has_must_attack && !is_goaded && !has_attackable_must_attack_player {
+    if !has_must_attack && !must_attack_away && !has_attackable_must_attack_player {
         return false;
     }
     // Exemptions: tapped, defender (no override), summoning sick.
@@ -3043,10 +3047,16 @@ enum AttackRequirement {
         player: PlayerId,
     },
     /// CR 701.15b (second clause) + CR 701.15c: `creature` attacks a player other
-    /// than `goader` if able. Obeyed iff `creature` attacks a player ≠ `goader`.
-    Goad {
+    /// than `avoided` if able. Obeyed iff `creature` attacks a player ≠
+    /// `avoided`. Two source classes produce it — the goad DESIGNATION
+    /// (`goaded_by` / `StaticMode::Goaded`, CR 701.15a/b) and the spelled-out
+    /// requirement without any designation (`StaticMode::MustAttackAwayFromSource`
+    /// — Kardur, Doomscourge; Maximum Carnage chapter I) — so the variant is
+    /// named for the requirement, not for goad. Each distinct avoided player is
+    /// an additional requirement (CR 701.15c).
+    AttackAwayFrom {
         creature: ObjectId,
-        goader: PlayerId,
+        avoided: PlayerId,
     },
 }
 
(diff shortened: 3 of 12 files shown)


lgray:

Verified your evidence-gap commit `41ddeca8d` rather than re-doing it — both CodeRabbit items were already yours, so this is confirmation plus one leftover you may want to sweep in.

### Your `statics.rs` CR correction is right

Grep-verified against `docs/MagicCompRules.txt`:

- **CR 611.2c** governs the *affected-object set* (fixed vs. dynamic) — "A continuous effect … that doesn't modify the characteristics or change the controller of any objects modifies the rules of the game, so it can affect objects that weren't affected when that continuous effect began."
- **CR 109.5** governs the *"you" anchor* — "The words 'you' and 'your' on an object refer to the object's controller … For a static ability, this is the current controller of the object it's on."

Citing 611.2c for the snapshot was self-contradictory, since this PR *relies* on 611.2c keeping the affected set **dynamic** (that's the `requirement_binds_a_creature_that_enters_after_resolution` leg). Good catch to split them.

### Independent verification at `41ddeca8d`

`cargo test -p engine --test integration bbfu7` → **13 passed / 0 failed**, including `cant_attack_restriction_beats_the_spelled_out_requirement`, which carries your new reach assertion. That assertion is a real non-vacuity improvement: without it the test's legality delta could have come from a missing grant rather than from the Pacifism restriction it claims to be measuring.

The `coverage-parse-diff` bot also confirms the intended blast radius exactly — 2 cards / 4 signatures, `GoadAll` removed → `MustAttack, MustAttackAwayFromSource` added, Kardur + Maximum Carnage only.

### Leftover: the same misattribution survives at a verbatim-parallel twin

CodeRabbit only flagged the `statics.rs` instance, but the identical claim is still in the tree. Comment-only, no behavior change — take or leave.

**1. `crates/engine/src/game/combat.rs:4239`** — this is the direct twin of the line you corrected:

```rust
// CR 611.2c: the avoided player is the anchor snapshotted at graft time.
```

suggested:

```rust
// CR 109.5: the avoided player is the anchor snapshotted at graft time
// ("you" in a resolving ability is that ability's controller).
```

Worth noting that this comment's own closing sentence already paraphrases 109.5 verbatim — it justifies the `unwrap_or(obj.controller)` default with *"'…other than you' on a permanent means that permanent's controller"*, which is exactly CR 109.5's "For a static ability, this is the current controller of the object it's on." So the body argues 109.5 while the citation says 611.2c.

**2. `crates/engine/src/game/layers.rs:5293`** — looser, same axis:

```rust
// CR 611.2c + CR 701.15b: the avoided player is the INSTALLING player,
```

suggested:

```rust
// CR 109.5 + CR 701.15b: the avoided player is the INSTALLING player,
```

The claim on that line ("the avoided player is the INSTALLING player, not the carrier's controller") is an anchor claim, so 109.5 is the operative rule. Adjacent blocks in the same file already pair both rules correctly, so this one is imprecise rather than wrong.

I spot-checked the other ~12 `CR 611.2c` citations in the diff (`effect.rs`, `layers.rs`, the tests) — those all describe affected-set-stays-dynamic and are **correct as written**.

No push from me: this is your branch, the PR is `BEHIND` and you'll likely rebase, and a comment-only push would just re-trigger CI into that.

matthewevans:
Confirmed and addressed on the current head: the two remaining controller-anchor comments now cite CR 109.5 in `f05ac72fb3`. The CR 611.2c citations that describe the dynamic affected set remain unchanged. Fresh CI and parse-diff evidence are running on this head before approval/enqueue.

matthewevans (commented):
Reviewed the current head `f05ac72fb3`: the only new code delta is the two verified CR 109.5 controller-anchor citation corrections. The parse-diff remains limited to Kardur, Doomscourge and Maximum Carnage, and the earlier CodeRabbit items are resolved. This is not an approval; fresh required CI is still running on the current head.

matthewevans (approved):
Approved after current-head review: the spelled-out requirements remain separate from goad, with the dynamic affected set and controller anchor covered.

PR #6320, opened 2026-07-22
Card/plargg and nassari

Whovencroft:
## Summary
Adds engine support for **Plargg and Nassari**.

Closes the previously-unsupported `choose` gap on the upkeep trigger's middle sentence — "An opponent chooses a nonland card exiled this way." — across four seams:

1. **Parser (anaphor ):** `try_parse_choose_exiled_anaphor`'s "exiled this way" referent only supported the untyped chain-tracked-set form ("choose a card exiled this way") and fell through honestly on a typed qualifier. The typed form ("a **nonland** card exiled this way") now lowers to the same linked-exile shape as the Koh, the Face Stealer "exiled with ~" arm: `ChooseFromZone { zone: Exile, zone_owner: AllOwners, filter: And { Typed, ExiledBySource } }` (CR 607.2a linkage; CR 400.1 shared-zone scan). The card-type qualifier now also accepts negated `non<type>` forms (CR 205.2b), and a chooser-prefix alternation ("an opponent chooses" / "target opponent chooses" → `Chooser::Opponent`, CR 608.2d) mirrors `parse_choose_anaphoric`'s existing convention.
2. **Parser (subject layer):** `strip_subject_clause` peels "an opponent" and deconjugates the predicate to "choose …", losing the chooser. `lower_subject_predicate_ast` now rebinds the bare single-opponent subject onto `ChooseFromZone`'s `chooser` before the honesty gate; typed or property-carrying subjects still fall through to `Effect::unimplemented`.
3. **Runtime (choice continuation):** the `ChooseFromZoneChoice` handler bound the **chosen** card as the parked continuation's target list — correct for chosen-consumer continuations (End-Blaze Epiphany's "you may play that card") but inverted for Plargg's "cast up to two spells from among the **other** cards exiled this way". When the continuation head is an exile-linked `CastFromZone` whose target carries the `FilterProp::Another` exclusion, the handler now binds the **unchosen** complement (CR 608.2c); all other continuations are untouched.
4. **Runtime (candidate pool):** `resolve_candidate_cards` returned tracked-set-derived candidate pools **raw**, ignoring the effect's own card filter — the player-scope completion publishes *every* card the exile clause moved (lands included) as the chain tracked set, so the opponent was offered the lands too. A typed filter (`Some(...)`) now narrows tracked-set and explicit-target pools through `matches_target_filter` (CR 608.2d); `filter: None` keeps the raw set, so the untyped "choose one of them" tracked-set anaphors are unchanged. Surfaced by the new runtime regression test on this PR's first CI run.

## Files changed
- crates/engine/src/game/effects/choose_from_zone.rs
- crates/engine/src/game/effects/exile_from_top_until.rs

## CR references
- CR 101.4 (APNAP order for the multiplayer opponent choice)
- CR 205.2b (`non<type>` card-type qualifier)
- CR 400.1 (exile is a shared zone — ownership must not gate the pool)
- CR 607.2a ("exiled this way" is a linked-ability reference to the exiling instruction)
- CR 608.2c (effect resolution — the complement binding for "the other cards")
- CR 608.2d ("an opponent chooses" delegates the resolution-time selection)

Method: /engine-implementer

## Track
Non-developer

## LLM
Model: manus-1.5
Thinking: high

## Verification
- `./scripts/check-parser-combinators.sh origin/main` — `Gate G PASS` + `Gate A PASS head=7aaadbda3080edcb0b7cb6cae9e05565460879fc base=bdb1f3de5c37fc216d79c683c854be6448aad44c`
- `cargo check -p engine --lib` — `Finished` with zero errors and zero warnings (on the initial implementation commit; later commits validated by CI)
- Non-developer track: `cargo test -p engine` (including the five new tests below), `cargo fmt --check`, `cargo clippy-strict`, `cargo coverage`, and `cargo semantic-audit` are delegated to the CI-owned checks on this PR — the ephemeral sandbox was reset mid-run three times, so local test execution could not complete.
- CI status at head: Rust lint (fmt, clippy, parser gate), both test shards, Card data (generate, validate, coverage), WASM/Tauri compile checks, frontend, and lobby worker all green.
- New tests on this PR: `parse_choose_nonland_card_exiled_this_way`, `parse_opponent_chooses_nonland_card_exiled_this_way`, `parse_choose_card_exiled_this_way_still_tracked_set` (imperative.rs), `plargg_and_nassari_full_trigger_chain_choose_then_cast_others` (tests.rs), `plargg_opponent_chooses_nonland_hit_then_cast_pool_is_the_other_hits` (exile_from_top_until.rs, runtime `apply()`-driven).

## Gate A
Gate A PASS head=7aaadbda3080edcb0b7cb6cae9e05565460879fc base=bdb1f3de5c37fc216d79c683c854be6448aad44c

## Anchored on
- crates/engine/src/parser/oracle_effect/imperative.rs:4040 — Koh, the Face Stealer's "exiled with ~" arm: the existing `FromZone { Exile, AllOwners, And { Typed, ExiledBySource } }` lowering the typed "exiled this way" referent now reuses (CR 607.2a + CR 400.1).
- crates/engine/src/parser/oracle_effect/imperative.rs:3822 — `parse_choose_anaphoric`'s chooser alternation ("an opponent chooses N of them" → `Chooser::Opponent`), the CR 608.2d delegation convention the new chooser prefix mirrors.
- crates/engine/src/game/engine_resolution_choices.rs:3798 — the existing chosen/unchosen partition seam in the `ChooseFromZoneChoice` handler, which already forwards the unchosen remainder to the zone-partition sub-ability; the complement binding extends the same seam.

## Final review-impl
Final review-impl PASS head=7aaadbda3080edcb0b7cb6cae9e05565460879fc
- Full-diff pass against CLAUDE.md and the review-impl checklist: no string-method dispatch, no match-on-string-literal arms, no chained `if let Ok(tag)` blocks, no hand-built `Effect::Unimplemented` literals in the diff.
- Seam check: typed-anaphor lowering sits in the existing anaphor combinator (not a new dispatch path); the chooser rebind sits at the subject/predicate lowering seam where every other subject rebind lives; the complement binding sits at the single continuation-binding authority.
- Test discrimination: the runtime test drives the real pipeline (`resolve_ability_chain` → `WaitingFor::ChooseFromZoneChoice` → `apply(GameAction::SelectCards)`) and fails if the complement binding, the opponent chooser rebind, or the typed-anaphor filter is reverted. The untyped-referent regression guard pins End-Blaze-style impulse reductions to the chain tracked set.

## Claimed parse impact
- Plargg and Nassari
- Author of Shadows (middle sentence "Choose a nonland card exiled this way." — same typed exiled-this-way anaphor; the card's remaining clauses are unchanged)

## Validation Failures
None.

## CI Failures
Three red runs during iteration, each root-caused and fixed on this branch; the final head is green.
1. `Rust lint` — two rustfmt diffs in test/handler code introduced after a sandbox reset dropped the local `cargo fmt` pass. Fixed in 5578e01 by applying the exact rustfmt output.
2. `Rust tests (both shards)` — E0433: `WaitingFor` missing from the runtime test's local import list. Fixed in de0e127.
3. `Rust tests (shard 2/2)` — the new runtime regression test failed: the opponent's candidate pool contained all six exiled cards (lands included) because `resolve_candidate_cards` ignored the effect filter for chain-tracked-set pools. Fixed in 7aaadbd (Summary seam 4); shard 1, Card data, and all other checks were already green on the same run, and no pre-existing test changed behavior.

## Summary by CodeRabbit

- **New Features**
  - Added a new multiplayer “Choose Opponent” prompt flow when an opponent must be selected to make the next zone decision, including UI modal and localized text.
- **Bug Fixes**
  - Improved “an opponent chooses…” behavior by adding an intermediate opponent-selection step when multiple eligible opponents exist.
  - Tightened candidate selection/filtering for linked zone choices and free-cast windows, including correct handling for typed nonland and exile-by-source scenarios.
  - Ensured free-cast candidate eligibility and viewer-facing prompts consistently preserve originating source context.
- **Tests**
  - Added regression coverage for the Plargg/Nassari trigger chain, including opponent-chooser pauses and linked exile-to-free-cast behavior.

### DIFF
--- a/client/src/adapter/types.ts
+++ b/client/src/adapter/types.ts
@@ -1686,6 +1686,9 @@ export type WaitingFor =
   | { type: "ManifestDreadChoice"; data: { player: PlayerId; cards: ObjectId[]; source_id: ObjectId } }
   | { type: "LearnChoice"; data: { player: PlayerId; hand_cards: ObjectId[] } }
   | { type: "ClashChooseOpponent"; data: { player: PlayerId; candidates: PlayerId[]; ability: unknown } }
+  // CR 608.2d: "an opponent chooses" from a zone (multiplayer) — the controller
+  // picks WHICH opponent makes the choice before the zone choice is presented.
+  | { type: "ChooseFromZoneOpponentChooser"; data: { player: PlayerId; candidates: PlayerId[]; ability: unknown } }
   | { type: "ChooseAnnouncingOpponent"; data: { player: PlayerId; candidates: PlayerId[]; choice_index: number; choice_count: number; target_type?: CoreType; pending_cast: unknown } }
   | { type: "ClashCardPlacement"; data: { player: PlayerId; card: ObjectId; remaining: [PlayerId, ObjectId][] } }
   | { type: "VoteChoice"; data: {
@@ -2119,6 +2122,8 @@ export type GameAction =
   // CR 702.99a: answer to CipherEncodeChoice — a creature to encode on, or null to decline.
   | { type: "CipherEncode"; data: { creature: ObjectId | null } }
   | { type: "ChooseClashOpponent"; data: { opponent: PlayerId } }
+  // CR 608.2d: answer to ChooseFromZoneOpponentChooser — which opponent will choose.
+  | { type: "ChooseZoneOpponentChooser"; data: { opponent: PlayerId } }
   | { type: "ChoosePileOpponent"; data: { opponent: PlayerId } }
   | { type: "ChooseAnnouncingOpponent"; data: { opponent: PlayerId } }
   | { type: "ChooseAssistPlayer"; data: { player: PlayerId | null } }
--- a/client/src/adapter/types.ts
+++ b/client/src/adapter/types.ts
@@ -1481,6 +1481,13 @@ export type CastOfferKind =
       filter: TargetFilter;
       zones: Zone[];
       exile_instead_of_graveyard?: boolean;
+      // CR 406.6: source of the granting ability (engine serde-default;
+      // absent in payloads predating the field).
+      source?: ObjectId;
+      // CR 607.2a: THIS resolution's "exiled this way" batch (Plargg and
+      // Nassari); omitted when empty (no batch restriction). Display-only
+      // pass-through — the modal renders `candidates`.
+      member_pool?: ObjectId[];
     };
 
 // CR 103.5b: Which declare-point action a pending BottomCards obligation
--- a/client/src/components/modal/ZoneOpponentChooserModal.tsx
+++ b/client/src/components/modal/ZoneOpponentChooserModal.tsx
@@ -1,0 +1,76 @@
+import { useTranslation } from "react-i18next";
+
+import type { GameAction, PlayerId, WaitingFor } from "../../adapter/types.ts";
+import { useGameDispatch } from "../../hooks/useGameDispatch.ts";
+import { useCanActForWaitingState } from "../../hooks/usePlayerId.ts";
+import { useGameStore } from "../../stores/gameStore.ts";
+import { getOpponentDisplayName } from "../../stores/multiplayerStore.ts";
+import { ChoiceModal } from "./ChoiceModal.tsx";
+
+type ZoneOpponentChooserWaitingFor = Extract<
+  WaitingFor,
+  { type: "ChooseFromZoneOpponentChooser" }
+>;
+
+interface ZoneOpponentChooserModalContentProps {
+  waitingFor: ZoneOpponentChooserWaitingFor;
+  seatOrder?: PlayerId[];
+  dispatch: (action: GameAction) => void | Promise<void>;
+}
+
+/**
+ * CR 608.2d: "An opponent chooses …" from a zone in a multiplayer game — the
+ * controller decides which opponent makes the choice before the zone choice
+ * itself is presented to that opponent (Plargg and Nassari's release notes:
+ * "you choose which opponent gets to choose one of the exiled nonland cards").
+ */
+export function ZoneOpponentChooserModalContent({
+  waitingFor,
+  seatOrder,
+  dispatch,
+}: ZoneOpponentChooserModalContentProps) {
+  const { t } = useTranslation("game");
+  const candidates = [...waitingFor.data.candidates].sort((a, b) => {
+    const aIdx = seatOrder?.indexOf(a) ?? a;
+    const bIdx = seatOrder?.indexOf(b) ?? b;
+    return aIdx - bIdx;
+  });
+
+  return (
+    <ChoiceModal
+      title={t("zoneOpponentChooser.title", "Choose Opponent")}
+      subtitle={t(
+        "zoneOpponentChooser.subtitle",
+        "Choose which opponent makes the choice.",
+      )}
+      options={candidates.map((opponent) => ({
+        id: String(opponent),
+        label: getOpponentDisplayName(opponent),
+      }))}
+      onChoose={(id) => {
+        dispatch({
+          type: "ChooseZoneOpponentChooser",
+          data: { opponent: Number(id) },
+        });
+      }}
+    />
+  );
+}
+
+export function ZoneOpponentChooserModal() {
+  const canActForWaitingState = useCanActForWaitingState();
+  const dispatch = useGameDispatch();
+  const waitingFor = useGameStore((s) => s.waitingFor);
+  const seatOrder = useGameStore((s) => s.gameState?.seat_order);
+
+  if (waitingFor?.type !== "ChooseFromZoneOpponentChooser") return null;
+  if (!canActForWaitingState) return null;
+
+  return (
+    <ZoneOpponentChooserModalContent
+      waitingFor={waitingFor}
+      seatOrder={seatOrder}
+      dispatch={dispatch}
+    />
+  );
+}
--- a/client/src/components/modal/ZoneOpponentChooserModal.tsx
+++ b/client/src/components/modal/ZoneOpponentChooserModal.tsx
@@ -1,6 +1,6 @@
 import { useTranslation } from "react-i18next";
 
-import type { GameAction, PlayerId, WaitingFor } from "../../adapter/types.ts";
+import type { GameAction, WaitingFor } from "../../adapter/types.ts";
 import { useGameDispatch } from "../../hooks/useGameDispatch.ts";
 import { useCanActForWaitingState } from "../../hooks/usePlayerId.ts";
 import { useGameStore } from "../../stores/gameStore.ts";
@@ -14,7 +14,6 @@ type ZoneOpponentChooserWaitingFor = Extract<
 
 interface ZoneOpponentChooserModalContentProps {
   waitingFor: ZoneOpponentChooserWaitingFor;
-  seatOrder?: PlayerId[];
   dispatch: (action: GameAction) => void | Promise<void>;
 }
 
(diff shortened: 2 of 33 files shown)


matthewevans (changes requested):
**Request changes — this support path is not rules-correct in multiplayer and does not preserve Plargg and Nassari’s two-spell limit.**

## 🔴 Blocker

**[HIGH]** Let the controller choose which opponent makes the middle choice. Evidence: `crates/engine/src/game/effects/choose_from_zone.rs:930-945` resolves `Chooser::Opponent` to the first APNAP opponent; `exile_from_top_until.rs:1054-1072` only asserts that the chooser is not the controller. The official Plargg and Nassari release notes state: “In a multiplayer game, you choose an opponent to choose a nonland card exiled with Plargg and Nassari.” Why it matters: in a 3+ player game the engine silently selects a different chooser instead of presenting the controller’s required choice. Suggested fix: add a typed resolution-time opponent-selection step that binds the selected opponent before the `ChooseFromZoneChoice` prompt, then cover the 3+ player selection path.

**[HIGH]** Preserve and enforce the printed “up to two spells” bound. Evidence: `crates/engine/src/game/engine_resolution_choices.rs:3847-3849` binds every unchosen card, while `crates/engine/src/game/effects/cast_from_zone.rs:325-336,620` treats all bound object targets as permission recipients; the added runtime test has only three players and creates `cast_sub` without a cardinality bound at `exile_from_top_until.rs:981-1008`. Why it matters: with four players and one nonland hit each, one choice leaves three cards and grants all three free-cast permissions, exceeding the card’s explicit limit of two. Suggested fix: carry a typed, resolution-time cast-choice bound through `CastFromZone` and test a four-player fixture that permits exactly two of three unchosen hits.

**[MED]** Do not parse `target opponent chooses` without retaining the target. Evidence: `crates/engine/src/parser/oracle_effect/imperative.rs:3968-3975` newly lowers both “an opponent” and “target opponent” to the same `Chooser::Opponent`; `oracle_effect/mod.rs:18608-18617` returns without wrapping `subject.target`, although `choose_from_zone.rs:934-939` can only honor a targeted chooser when that player target is present. Why it matters: the newly accepted targeted form can prompt an arbitrary opponent rather than the declared target. Suggested fix: model the targeted-opponent form separately and preserve its target slot through lowering, with an end-to-end regression.

## 🟡 Non-blocking

The current-head engine/parser parse-diff sticky comment is absent even though the card-data check ran; regenerate it after the correction so the claimed Plargg and Nassari/Author of Shadows impact can be compared with the measured card set.

**Recommendation: request changes and re-run the current-head parse-diff after the multiplayer chooser and bounded cast flow are implemented.**

matthewevans (changes requested):
**Request changes — this head correctly removes the unbound `target opponent chooses` grammar, but both blocking Plargg and Nassari behaviors remain.**

## 🔴 Blocker

**[HIGH]** Let the controller choose which opponent makes the middle choice. Evidence: `crates/engine/src/game/effects/choose_from_zone.rs:930-945` still resolves `Chooser::Opponent` to the first APNAP opponent; `exile_from_top_until.rs:1054-1072` only proves the chooser is not the controller. The official Plargg and Nassari release notes say that, in multiplayer, *you choose an opponent* to make this choice. Why it matters: with three or more players the engine silently selects a chooser instead of performing the required choice. Suggested fix: add a typed resolution-time opponent-selection step and a 3+ player regression.

**[HIGH]** Preserve and enforce the printed “up to two spells” bound. Evidence: `crates/engine/src/game/engine_resolution_choices.rs:3847-3849` still binds every unchosen card and `crates/engine/src/game/effects/cast_from_zone.rs:325-336,620` grants each bound card a permission; `exile_from_top_until.rs:981-1008` has no cardinality-bound cast choice. Why it matters: with four players and one nonland hit each, choosing one leaves three cards and grants all three free-cast permissions, exceeding the card’s limit. Suggested fix: carry a typed resolution-time maximum through `CastFromZone`, with a four-player regression that permits exactly two of three unchosen hits.

**Recommendation: keep this PR blocked until the controller-selection and max-two cast flow are implemented and verified.**

matthewevans (changes requested):
**Request changes — the refreshed head fixes the original chooser and two-cast defects, but it still leaks prior-resolution cards into the current free-cast window and leaves the new choice state unplayable in the browser.**

## 🔴 Blocker

**[HIGH]** Scope “cards exiled this way” to this trigger resolution, not every card ever linked to Plargg and Nassari. Evidence: `crates/engine/src/game/effects/free_cast_from_zones.rs:118-145` scans the whole shared exile zone and filters it with `ExiledBySource`; `crates/engine/src/game/filter.rs:2292-2306` implements that filter as the source’s complete live linked-exile ledger. After a previous upkeep left a linked nonland card in exile, the next trigger’s window can therefore offer that old card whenever it was not the newly chosen card. The verified Oracle text says “from among the other cards exiled this way,” which confines the offer to this resolution’s exile batch. Suggested fix: carry a typed, concrete current-resolution member set through the choose → free-cast continuation, and intersect the offer with it before applying the chosen-card exclusion; add a two-upkeep regression where an older linked card is unavailable.

**[HIGH]** Complete the frontend and serialized-surface wiring for the new interactive state. Evidence: the engine creates `WaitingFor::ChooseFromZoneOpponentChooser` at `crates/engine/src/types/game_state.rs:8553-8567`, but `client/src/adapter/types.ts:1688-1690` has no corresponding `WaitingFor` union member, the action union has no `ChooseZoneOpponentChooser`, and `client/src/game/waitingForRegistry.ts:125-127` / `client/src/pages/GamePage.tsx:1731-1736` register and render only the existing opponent-choice dialogs. Why it matters: the new four-player path pauses on an unhandled browser state and reaches the fail-loud exit modal instead of allowing the controller to choose an opponent. Suggested fix: thread the new waiting/action variants across the adapter, add a localized opponent-picker modal (reusing the existing opponent-choice presentation), register/mount it, and add the browser-facing dispatch test.

## 🟡 Non-blocking

**[MED]** Reconcile the claimed parse scope with the current parse-diff artifact. Evidence: the current `<!-- coverage-parse-diff -->` reports five changed cards — Author of Shadows, Plargg and Nassari, Forgotten Lore, Search for Survivors, and Shrouded Lore — while the PR’s claimed-card list names only the first two. Why it matters: the three unclaimed parser changes cannot be evaluated against an explicit intended class. Suggested fix: either name and justify those cards as the intended grammar class, with representative assertions, or narrow the parser change so the artifact matches the stated scope.

## ✅ Clean

The current-head four-player engine test now proves that the controller selects the opponent and that the immediate window starts with two remaining casts; those are material improvements over the previously reviewed head.

**Recommendation: request changes; keep the PR blocked until the resolution-scoped set and browser choice flow are implemented with discriminating regressions.**

matthewevans (changes requested):
**Request changes — the new frontend and opponent-picker work resolves the prior browser and multiplayer-choice blockers, but the current head still treats an all-time source exile ledger as this trigger resolution's `this way` set.**

## 🔴 Blocker

**[HIGH] Scope the free-cast pool to the cards exiled during the current resolution.**

Evidence: `crates/engine/src/game/effects/free_cast_from_zones.rs:118-145` now scans every object in `state.exile` and evaluates `TargetFilter::ExiledBySource`; `crates/engine/src/game/filter.rs:2292-2306` resolves that filter with `linked_exile_cards_for_source(state, source_id)`, which is the live, source-wide link ledger. Nothing in this path consumes the current resolution's tracked set. Consequently, on a later upkeep Plargg and Nassari can offer unchosen nonlands exiled by an earlier upkeep, even though the Oracle text says “the other cards exiled this way.”

Thread a resolution-scoped linked set through the choose/free-cast continuation (and keep the chosen-card exclusion against that same set), then add a two-upkeep regression proving the second offer cannot contain the first upkeep's cards. The current one-upkeep test cannot expose this lifetime error.

matthewevans (changes requested):
The prior stale-exile fix is present, but several current-head merge blockers remain:

1. The empty `member_pool` continuation discards `player` and resumes without `set_priority`, leaving the opponent-chooser state unable to drain. Bind the player and restore priority before resuming.
2. The counted free-cast parser arm lowers even when it parsed a cast constraint, but `FreeCastFromZones` cannot carry that constraint. Do not silently drop it: gate this arm on `constraint.is_none()` or retain strict unsupported handling.
3. `ZoneOpponentChooserModal` sorts engine candidates by `seatOrder`; that game ordering belongs in the engine. Render the supplied order.
4. Register `local.zone-opponent-chooser-unsupported` in the public capability registry and update its test.
5. The current parse diff has five changed cards, including Forgotten Lore, Search for Survivors, and Shrouded Lore, while the PR claims only Plargg and Author of Shadows. Reconcile and test the additional parser impact before re-review.

matthewevans (approved):
Approved after current-head review: the chooser, continuation, and resolution-scoped exile pool are covered by the refreshed runtime regressions.

PR #6323, opened 2026-07-22
fix(engine): read entered-this-turn quantities from the entry ledger (BB-FU10)

lgray:

## Summary

`parse_entered_this_turn_ref` emitted the live-board `QuantityRef::EnteredThisTurn` for Oracle text whose "entered the battlefield under your control this turn" clause is an **entry event**, so a permanent that entered this turn and has since left stopped counting — contradicting CR 608.2i (look-back). This migrates only the entry-event surface to the ledger-backed `QuantityRef::BattlefieldEntriesThisTurn { player, filter }`, leaves the subject-noun surface (Tromell, CR 608.2h live set) alone, and repairs three defects the migration exposed in the entry-record matcher.

Method: /engine-implementer

## CR references

`CR 608.2i` (look-back — the migration's basis), `CR 608.2h` (once-determined live set — why Tromell must NOT migrate), `CR 702.73a` (Changeling CDA), `CR 205.3m` (creature-type namespace), `CR 403.3` (permanents exist only on the battlefield), `CR 109.1`, `CR 109.2`, `CR 109.5`, `CR 111`, `CR 400`, `CR 102.2`, `CR 102.3`, `CR 205.4b`, `CR 308.1`, `CR 603.3b`, `CR 603.6a`, `CR 611.3a`, `CR 701.21`, `CR 732.2a`.

Every number was grep-verified against `docs/MagicCompRules.txt` before being written; the whole-diff gate reports zero unverified. The pre-existing `// CR 608.2b: Disjunction` miscitation at `filter.rs:2833` was deliberately **not** propagated (neither 608.2b nor 608.2i is a disjunction rule).

## Verification

- `cargo fmt --all --check` — exit 0
- `cargo clippy --all-targets -- -D warnings` — exit 0, zero diagnostics
- `cargo nextest run -p engine` — **21258 passed / 0 failed / 8 skipped**
- `cargo nextest run -p engine -E 'test(combo_infinite_pile) | test(kilo_live_offer_from_real_dump) | test(sprout_inalla_realistic_offer) | test(loop_shortcut)'` — **96/96 passed**
- `./scripts/check-parser-combinators.sh` — exit 0 (Gate G + Gate A PASS)
- `cargo combo-verify` — **13 confirmed / 4 gated / 37 deferred / 0 failed (of 54)**; all **54 rows row-for-row identical** to base
- BASE-vs-POST `data/card-data.json` — 35477 cards both sides, 0 added, 0 removed, **5 changed** (see Claimed parse impact)
- BASE-vs-POST `data/coverage-data.json` — `supported_cards` **31504 both sides**, supported-flip multiset diff **0 / 0**
- `FORGE_TEST_FULL_DB=1 … ordering_parity_sweep` — `unexplained=18` on **both base and head**, all metrics and all **18 rows byte-identical** (delta zero). Pre-existing Urza-block failure, opt-in gate, not default CI.

Base and head card-data were generated from **md5-identical MTGJSON inputs** (`AtomicCards.json` `0dd2d256…`, verified after both runs), so every delta above is code-caused, not input drift.

## Gate A

Gate A PASS head=96d5a3b0b434052651e4a95a77941839bcd7fb18 base=836ff312ae2073c99af28d286b0c4915faa8a458

## Anchored on

- `crates/engine/src/game/filter.rs:2840` — `zone_change_record_matches_type_filter`: the existing snapshot-record type-filter matcher that already threads `all_creature_types` live against a snapshotted record. Step 8 mirrors it exactly.
- `crates/engine/src/parser/oracle_nom/condition.rs:7725` — `parse_or_more_entered_count`: BB-FU1's already-shipped condition-side emitter of the same ledger variant, i.e. the sibling surface this PR's quantity-side migration matches.

## Final review-impl

Final review-impl PASS head=96d5a3b0b434052651e4a95a77941839bcd7fb18

## Claimed parse impact

- Hobgoblin Bandit Lord
- Geralf, the Fleshwright
- Cloudspire Coordinator
- Kinbinding
- Bioengineered Future

Three further cards change behavior with **no AST change**, so they cannot appear in a card-data diff — measured and confirmed absent from it:

- Cleaving Reaper
- Lilypad Village
- Nimble Trapfinder

They benefit from the `TargetFilter::Or` repair in the entry-record matcher; two of them gain a previously-impossible activation. The unlock is driven, not asserted: `bbfu10_lilypad_village_activation_unlocked_by_composite_fix` drives `check_activation_restrictions` and measures ILLEGAL → LEGAL.

## Disclosures

**1. A CR 732.2a firewall reclassification suppresses three shipping combo offers.**
`ability_scan.rs` classified `BattlefieldEntriesThisTurn` as `sibling: false`, while `analysis/resource.rs` already clears `battlefield_entries_this_turn` as *"append-only event journals a loop pumps"*. Those two statements contradict each other, and the combination lets the engine hand out a **false ∞ certificate** today. This PR flips the axis to `sibling: true`, which is what the module's own ⛔ INVARIANT requires.

Measured cost: three battlefield permanents — **Park Heights Pegasus**, **Smuggler's Share**, and **The Prydwen, Steel Flagship** (one card; the comma is part of its name) — stop receiving a CR 732.2a loop-shortcut offer they receive today. This is a deliberate, disclosed trade, not an oversight; restoring them under a narrower predicate is tracked as follow-up **BB-FU10-N**. `loop_shortcut.rs` carries a driven regression test that asserts the suppression and names BB-FU10-N as the item that will flip it back.

DB-wide census (measured at the pre-rebase base `e48c7d6e1`; the suppression itself is pinned at the current head by a committed driven test): 35454 faces, `sib_true` 16939 → 16945; the six added faces are exactly the predicted set and the removed set is **empty**. Step 1 in isolation measures 16937 — it *relaxes* Geralf and Hobgoblin Bandit Lord, which is precisely the false-certificate relaxation Step 0c closes.

**2. Not fixed, and deliberately so.** The entry-record snapshot is taken pre-layer, so only Changeling is repairable from it (the keyword travels; granted subtypes do not). 192 MTGJSON cards grant types via continuous effects and remain invisible to any snapshot-only matcher. `ParsedCondition::YouHadAngelOrBerserkerEnterThisTurn` still compares subtypes with a bare `eq_ignore_ascii_case`, so **Cleaving Reaper**'s Changeling axis stays wrong — that arm never routes through the matcher, so this PR's chokepoint cannot reach it. Our Cleaving Reaper claim above is about the ledger-count axis only.

**3. Coverage classification is filter-aware now, but only where the classifier looks.** `extract_card_features` walks static-ability conditions and effect positions (21 corpus cards reach the ledger arm, all `Handled`) but treats a trigger intervening-if as opaque, so it cannot see 20 measured trigger-side cards. **Tunnel Tipster** — `face-down creature entered … this turn` — is therefore still silently `supported=true` with a constant-false condition. That is a **pre-existing** BB-FU1-era defect, disclosed here, not introduced and not fixed.

**4. Instrument blindness, stated plainly.** The combo corpus contains none of the three suppressed cards, so `combo-verify`'s row-for-row identity is **expected** and is *not* evidence that no offer was suppressed. The suppression was measured directly by DB census and by a driven fixture instead.

## Re-verified after rebase

Rebased onto `upstream/main` = `8fe91f145` (29 commits). **Every gate above was re-measured at the rebased head against a freshly generated base**, and every result is identical to the pre-rebase measurement — including the blast radius holding at exactly the same 5 cards, which was the drift most worth checking given the upstream tip (`fix(parser): stamp Permanent on duration-less additive type grants`) touches this PR's seam. A dedicated review pass confirmed no semantic interaction: this PR's `condition.rs` diff is comment-only, and `entry_type_filter_matches` reads the pre-layer entry snapshot, which never observed the grant under either duration.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of “entered the battlefield this turn” effects, including controller-specific and combined type conditions.
  * Departed permanents that entered this turn are now counted correctly where applicable.
  * Improved support for subtype checks, including Changeling interactions.

* **Bug Fixes**
  * Fixed ledger-based conditions that could previously return incorrect counts or remain inactive.
  * Prevented unsupported conditions from producing misleading results.
  * Improved loop detection when battlefield-entry effects depend on changing board state.

* **Documentation**
  * Clarified rules references and behavior for battlefield-entry look-back effects.

### DIFF
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -2183,9 +2183,42 @@ fn scan_quantity_ref(x: &QuantityRef, mode: ScanMode) -> Axes {
             acc
         }
         QuantityRef::BattlefieldEntriesThisTurn { player, filter } => {
+            // CR 732.2a: axis-2 self-assertion. This tally is a board-derived
+            // AGGREGATE — `record_battlefield_entry` (game/restrictions.rs) APPENDS
+            // to `battlefield_entries_this_turn` on every battlefield entry, so a
+            // sibling resolution, or a loop cycle that grows the board, changes its
+            // value.
+            //
+            // The engine ALREADY classifies this journal as loop-pumped:
+            // `project_out_resources` clears it at analysis/resource.rs:2977 under
+            // "CR 400 (zones) / CR 603.6a (ETB) / CR 701.21 (sacrifice) / CR 111
+            // (tokens): append-only event journals a loop pumps". A `sibling: false`
+            // here contradicted that, and in the unsafe direction: it let the
+            // CR 732.2a firewall certify a loop as bounded while a live observer read
+            // the growing class. Per the module header above (ADD-1) over-veto is the
+            // ONLY permitted error direction (#4603-preserving).
+            //
+            // Per the ⛔ INVARIANT on the `TargetFilter::Typed` arm of
+            // `scan_target_filter`, a board-AGGREGATE caller MUST self-assert
+            // `sibling: true` and must NOT delegate its board-read signal to the
+            // `Typed` arm, whose `LoopFirewall` relaxation otherwise erases the
+            // signal at the two `ability_definition_reads_sibling_mutable_for_loop`
+            // scans inside `fire_time_conditions_read_growing_class` —
+            // analysis/resource.rs:1699 (trigger `execute` bodies) and :1722
+            // (battlefield ability bodies) — neither of which has a `projected` twin
+            // in `fire_time_conditions_read_projected_resource`.
+            //
+            // The CR 603.3b APNAP ordering gate (game/triggers.rs, the
+            // `c2_order_independent` term) is the OTHER consumer of this arm and is
+            // provably UNAFFECTED: it scans in `ScanMode::Conservative`, where the
+            // `Typed`/`Or[Typed]` filter every producer emits already forces
+            // `sibling: true` — measured.
+            //
+            // The `filter` census intent stays `SnapshotOrEvent`: it is matched
+            // against a `BattlefieldEntryRecord`, never a live board.
             let mut acc = Axes {
                 event: false,
-                sibling: false,
+                sibling: true,
                 projected: true,
             };
             acc = acc.or(scan_player_scope(player));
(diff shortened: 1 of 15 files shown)


matthewevans (commented):
**Reviewed the current head.**

I found no additional concrete correctness blocker in the migrated entry-event path. The ledger migration, the preserved live subject-noun path, and the entry-record filter guard follow the existing condition-side model.

This remains unapproved pending the in-progress CI and current-head card-data/coverage evidence. No local build was run.

matthewevans (approved):
Approved: the entry-event grammar now selects the ledger only when control modifies the entry event, retains the live subject-noun path, and exercises current-head parser, quantity, activation, and layer-invalidation paths. The parse-diff’s four affected cards match the migrated semantic shape; no quality label is requested because the claimed and measured card sets differ.

PR #6335, opened 2026-07-22
Fix Toxrill, the Corrosive — bind a counter anaphor to the recipient in per-recipient statics

### LINKED ISSUE #5929: Toxril, the Corrosive — [[Toxrill, the Corrosive]] 's ability puts slime counters on enemies, however the slime counter…
Reported in Discord: https://discord.com/channels/1485498006781427802/1525714983088951306/1525714983088951306

**Thread:** Toxril, the Corrosive
**Discord thread id:** `1525714983088951306`
**Discord message id:** `1525714983088951306`
**Cards:** toxrill, the corrosive
**Parser status:** has_gaps

## Summary
[[Toxrill, the Corrosive]] 's ability puts slime counters on enemies, however the slime counters do NOT give -1/-1 to enemies they are on.

### toxrill, the corrosive
At the beginning of each end step, put a slime counter on each creature you don't control.
Creatures you don't control get -1/-1 for each slime counter on them.
Whenever a creature you don't control with a slime counter on it dies, create a 1/1 black Slug creature token.
{U}{B}, Sacrifice a Slug: Draw a card.

---
<sub>report_id: `discord:1525714983088951306:1525714983088951306:0` · discord: `1525714983088951306/1525714983088951306`</sub>

minion1227:
## Summary
Fixes the counter-anaphor scope misparse for **Toxrill, the Corrosive** and its class.

CR 611.3a: a continuous effect from a static ability "isn't 'locked in'; it applies at any given moment to whatever its text indicates." So a bare anaphoric pronoun (`it` / `them` / `him` / `her`) in a per-recipient continuous static names the object **receiving** the effect, not the ability's source.

`parse_counter_object_scope` (`oracle_nom/quantity.rs:1203`) has no view of the enclosing static, so it bound every counter anaphor to `ObjectScope::Source`. Toxrill therefore scaled "-1/-1 for each slime counter on them" off Toxrill's **own** (always empty) slime pile, and the penalty never applied — the exact symptom reported.

The rebind is gated on the **pronoun form**, not merely on the affected set. The much larger Door of Destinies / Joraga Warcaller / Lion Sash class writes "…for each charge counter on ~", naming the source outright, where `Source` is already correct — a blanket rebind would have broken 32 cards to fix 5. A second test pins that class.

No new enum variant: `ObjectScope::Recipient` already existed and the layer evaluator already resolves it per affected object (`layers.rs:6078`). Only the recipient-varying classifier in `quantity_expr_uses_recipient` was missing its `CountersOn` arm. This is the quantity-axis twin of `StaticCondition::RecipientHasCounters`, documented in `types/ability.rs:7329` as "the recipient analog of `HasCounters`" — both axes now agree on what the pronoun in "…counters on it" refers to.

Closes #5929

## Files changed
- `crates/engine/src/parser/oracle_static/mod.rs`
- `crates/engine/src/game/quantity.rs`
- `crates/engine/tests/integration/counter_anaphor_binds_to_recipient.rs`

## CR references
- `CR 611.3a` — a static's continuous effect applies to whatever its text indicates, so the pronoun rebinds to the current recipient
- `CR 613.4c` — layer 7c, the effects/counters that modify power and/or toughness
- `CR 122.1` — a counter is a marker placed on an object

Method: not-applicable — single-seam parser lowering fix; planned and implemented in-session against a regenerated 35k-card corpus rather than via the multi-agent pipeline.

## Track
Developer

## LLM
Model: claude-opus-4-8
Thinking: high

## Verification

- `cargo fmt --all` — clean
- `cargo clippy -p engine --all-targets --features proptest -- -D warnings` — exit 0, no warnings
- `cargo test -p engine --lib` — `ok. 17490 passed; 0 failed; 6 ignored`
- `cargo test -p engine` (integration) — `ok. 3727 passed; 0 failed; 2 ignored`
- `./scripts/gen-card-data.sh` — exit 0, 35452 cards, 31507/35482 supported (88.80%)
- `cargo semantic-audit` — exit 0; **zero** findings for any of the 5 changed cards
- `cargo coverage` — all 5 changed cards `supported=true, gap_count=0`; control class (Joraga Warcaller, Door of Destinies, Lion Sash, Hancock, Earthen Goo, Intrepid Adversary) also `supported=true, gap_count=0`
- **Revert-proof:** commenting out the single `bind_counter_anaphor_to_recipient` call and re-running the new integration test fails at `counter_anaphor_binds_to_recipient.rs:56` with `left: (1, 1) right: (2, 2)` — the source-scoped read applied Toxrill's own 3 slime counters uniformly to every affected creature. The test drives `GameScenario` → `evaluate_layers` → `derive_display_state`, i.e. the production layer pipeline, not a parser-shape assertion.

## Gate A
Gate A PASS head=d39c97130e68af2902e45cd6d67df34c3266717f base=cf3381e5a8ca97351a573eac5e510de892d78dc6

## Anchored on
- `crates/engine/src/parser/oracle_static/mod.rs:223` — `populate_active_zones_from_condition`, the existing post-parse lowering transform in `lower_static_ir` that this one is registered beside, so every static builder inherits it from one authority
- `crates/engine/src/parser/oracle_static/shared.rs:461` — existing `// CR 611.3a: bind the combat state to the recipient, not the source` recipient-binding in this same module, same CR, same rebind direction
- `crates/engine/src/game/quantity.rs:711` — `continuous_modification_dynamic_quantity`, whose exhaustive no-wildcard arm list the new `_mut` mirror follows verbatim

## Final review-impl
Final review-impl PASS head=d39c97130e68af2902e45cd6d67df34c3266717f

Self-review found and fixed one defect before this head: `affected_names_the_source` originally cited `CR 604.2` for "an absent filter is self-scoped". CR 604.2 governs how long a static's continuous effect is active, not the default affected set — the citation did not describe the code, so it was replaced with a plain statement of the engine's `Option<TargetFilter>` representation. Two anaphora helpers were also retargeted from `CR 608.2k` (which scopes to cost/trigger-condition referents) to `CR 611.3a`, matching the convention already used at `shared.rs:461`.

## Claimed parse impact
Regenerated the full corpus and audited every `CountersOn` scope in every static. Before this change **zero** cards parsed to a recipient-scoped counter read; after, exactly these 5 do, and no other card's scope changes:

- Toxrill, the Corrosive — "Creatures you don't control get -1/-1 for each slime counter on them."
- Clamavus — "Each creature you control gets +1/+1 for each +1/+1 counter on it."
- Thelon of Havenwood — "Each Fungus creature gets +1/+1 for each spore counter on it."
- Luxior, Giada's Gift — "Equipped creature gets +1/+1 for each counter on it."
- Spark Rupture — "…is a creature with power and toughness each equal to the number of loyalty counters on it."

Corpus-wide scope distribution after the change: `Source` 90, `Recipient` 10 (5 cards × power+toughness). The 32-card explicit-`~` class was asserted unchanged.

## Validation Failures
None.

## CI Failures
None.

Tier: Frontier

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed continuous static effects so counter references (including recipient-scoped “bare counter” anaphors) are evaluated per recipient.
  * Corrected “for each … counter(s) on …” parsing and resolution, ensuring explicit “counter on ~” stays source-referential while pronoun-based forms bind appropriately.
  * Improved separation of source vs recipient counter magnitudes when both are referenced in the same effect.

* **Tests**
  * Added integration coverage for recipient-scoped “bare counter” anaphors in continuous effects.
  * Added cases verifying explicit “counter on ~” behavior and correct handling when both source and recipient counters are used.


matthewevans (changes requested):
### Blocking: counter-source provenance is lost before static lowering

`parse_counter_object_scope` lowers `it`/`them` and `~` to the same `ObjectScope::Source`. The new static lowering then scans the whole description for any recipient pronoun and rewrites every `CountersOn { object: Source, .. }` in every dynamic modification. A static that reads counters on both `~` and on its recipient will therefore rebind the explicit source read to the recipient.

This needs per-quantity anaphoric provenance carried through parsing/lowering, rather than a description-wide post-pass. Please add a mixed source/recipient runtime regression; the Toxrill and Joraga cases do not cover that combination.

minion1227:
Thanks — the blocking finding was correct, and the description-wide post-pass is gone.

**Where the provenance was actually lost.** Not `parse_counter_object_scope` (that path feeds the "the number of … counters on it" form). The anthem class comes through `parse_for_each_counters_on_source` (`oracle_nom/quantity.rs`), which parsed the object with `parse_source_self_ref` — a combinator that accepts `~` **and** "it" and returned `ObjectScope::Source` for both. That collapse is the defect. The object is now dispatched:

- explicit self-reference (`~`, "this Equipment") → `ObjectScope::Source`, bound at parse time
- bare objective pronoun ("it" / "them" / "him" / "her") → `ObjectScope::Anaphoric`, deferred

`ObjectScope::Anaphoric` is the variant already documented for this, and `rebind_anaphoric_object_scope` already lists `CountersOn` under it — the counter combinator simply never emitted it.

**Binding happens at lowering, per quantity.** `lower_static_ir` walks the continuous modifications *and* the `UnlessPay` condition tree (CR 118.12 — a combat tax carries its per-counter magnitude there, not in a modification). A self-referential subject binds back to `Source`, which is why Myr Prototype's AST is unchanged and `combat_tax_self_ref_subject_you_pay_per_counter` needed no edit.

**Mixed-scope regression added**, as requested. One static reads charge counters on `~` and +1/+1 counters on "it":

```
MIXED PARSE => scope: Source      (the `~` read)
               scope: Recipient   (the "on it" read)
```

Runtime: a 2/2 ally with four +1/+1 counters, under a source holding three charge counters, resolves to **9/10** (+3/+0 from the source, +0/+4 from the recipient). A single collapsed referent cannot produce that pair. The Toxrill test still fails on revert with `left: (1, 1) right: (2, 2)`.

**Blast radius held to zero outside the fix.** Every non-static entry settles an unbound anaphor back to `Source` at its own module boundary. Audited absolutely on the regenerated 35k corpus:

```
CountersOn scope distribution: Source 495, Recipient 8, Target 6
Cards leaving an UNBOUND Anaphoric: 0
```

So the only cards whose counter scope changes are the four fixed ones — Toxrill, Clamavus, Thelon of Havenwood, Luxior, Giada's Gift. Coverage total is identical (31507/35482, 88.7971%).

I checked this because an earlier revision *did* leak `Anaphoric` into 38 cards, and that is not behaviour-neutral: `casting.rs`'s `matches!(scope, ObjectScope::Source)` and `ability_rw::read_object_scope` both treat `Anaphoric` differently from `Source`. Rather than widen those shared helpers (which would have moved other scopes too), the anaphor is now prevented from escaping at all.

**Scope reduced honestly:** Spark Rupture is no longer claimed. Its "power and toughness each equal to the number of loyalty counters on it" reaches the other combinator, and giving that one provenance churned 10 unrelated cards; that path is left exactly as it was and the card stays a known gap rather than riding along on unrelated churn.

Verification on `f2a3b510a`: `cargo test -p engine --lib` 17492 passed / 0 failed · integration 3728 passed / 0 failed · `cargo clippy -p engine --all-targets --features proptest -D warnings` clean · `cargo semantic-audit` no findings for any changed card · Gate A PASS head=f2a3b510a3230e8bb8080e7a52d2a151e6fe54b5.

matthewevans (approved):
Reviewed current head ee6c9688f64ec78633d89b6aa8ef7d460b5aff93: counter anaphor provenance is now bound per quantity at static lowering; the mixed-scope runtime regression and the plain non-static settlement regression cover the two relevant boundaries.

PR #6340, opened 2026-07-22
fix: preserve colors_spent_to_cast for CR 603.4 Adamant ETBs (#5943)

### LINKED ISSUE #5943: Emptiness — Evoking [[Emptiness]] doesn't trigger it's ETB Expected results: Evoking emptiness with two W or 2 B should…
Reported in Discord: https://discord.com/channels/1485498006781427802/1525502721090912316/1525502721090912316

**Thread:** Emptiness
**Discord thread id:** `1525502721090912316`
**Discord message id:** `1525502721090912316`
**Cards:** emptiness
**Parser status:** has_gaps

## Summary
Evoking [[Emptiness]] doesn't trigger it's ETB
Expected results: Evoking emptiness with two W or 2 B should trigger its ETB effect.

### emptiness
When this creature enters, if {W}{W} was spent to cast it, return target creature card with mana value 3 or less from your graveyard to the battlefield.
When this creature enters, if {B}{B} was spent to cast it, put three -1/-1 counters on up to one target creature.
Evoke {W/B}{W/B}

---
<sub>report_id: `discord:1525502721090912316:1525502721090912316:0` · discord: `1525502721090912316/1525502721090912316`</sub>

andriypolanski:
Closes #5943

## Summary

Discord report: **Emptiness** evoked with `{W}{W}` or `{B}{B}` (via hybrid Evoke `{W/B}{W/B}`) does not fire the matching color-spent ETB.

The parser AST was already faithful — two `ChangesZone` ETBs gated by `TriggerCondition::ManaColorSpent { minimum: 2 }` plus the synthetic Evoke sac. The defect was runtime: after trigger *collection* succeeded (colors still present), `clear_post_collection_transients` zeroed `colors_spent_to_cast` on every object. CR 603.4 re-checks the intervening-if at *resolution* via `ExactLive` (permanent still on the battlefield), so Adamant-style ETBs silently did nothing.

## Root cause

`mana_spent_to_cast_amount` was already treated as a durable cast fact (Hungry Graffalon / Topiary Lecturer), but `colors_spent_to_cast` was still cleared with the transient `mana_spent_to_cast` boolean. That asymmetric cleanup breaks every `ManaColorSpent` intervening-if that re-checks while the source is ExactLive — Evoke Emptiness is the reported instance of the Adamant / “if {C} was spent to cast it” class.

## Changes

- **`crates/engine/src/game/triggers.rs`** — stop clearing `colors_spent_to_cast` in `clear_post_collection_transients`; document CR 603.4 / ExactLive rationale; unit test that `ManaColorSpent` survives the clear.
- **`crates/engine/src/game/game_object.rs`** — align field docs: color tally is durable like `mana_spent_to_cast_amount`.
- **Integration** — evoke Emptiness paying WW (GY return + sac) and BB (three −1/−1 + sac).

## Test Plan

- [ ] `cargo fmt --all -- --check`
- [ ] `cargo test -p engine --lib -- adamant_intervening_if_survives_post_collection_transient_clear`
- [ ] `cargo test -p engine --test integration -- emptiness_evoke`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved durable per-color “colors spent to cast” during post-trigger and intervening-if checks, while clearing transient cast markers (e.g., “mana spent to cast”).
  * Ensured cast-payment provenance is cleared when permanents leave the battlefield and when spells leave the stack to non-battlefield zones, preventing stale re-entry behavior.
  * Fixed *Emptiness* evoke/spent ETBs for both white and black mana paths across resolve, counter, blink, and reanimate scenarios.
* **Tests**
  * Added regression integration coverage for *Emptiness* evoke “color-spent,” including blink and counter/re-entry cases.
  * Added a unit test for battlefield-exit clearing of per-color tallies.
* **Documentation**
  * Clarified “colors spent to cast” vs transient cast tracking semantics.

### DIFF
--- a/crates/engine/src/game/game_object.rs
+++ b/crates/engine/src/game/game_object.rs
@@ -1039,10 +1039,12 @@ pub struct GameObject {
     /// CR 601.2h: Per-color breakdown of mana spent to cast this object.
     /// Populated during casting finalization; consumed by trigger / ability
     /// conditions like Adamant (CR 207.2c) and by Converge-style quantity
-    /// refs. Like `mana_spent_to_cast_amount`, this is a historical cast fact
-    /// that persists through CR 603.4 intervening-if resolution re-checks —
-    /// `clear_post_collection_transients` clears only the transient
-    /// `mana_spent_to_cast` boolean, not this tally.
+    /// refs. Persists through CR 603.4 intervening-if resolution re-checks on
+    /// the **same** battlefield object (`clear_post_collection_transients`
+    /// clears only the transient `mana_spent_to_cast` boolean, not this tally).
+    /// Cleared by `reset_for_battlefield_exit` (CR 400.7) so a permanent that
+    /// leaves and re-enters without being cast again does not inherit a stale
+    /// color tally from its previous visit.
     #[serde(default, skip_serializing_if = "ColoredManaCount::is_empty")]
     pub colors_spent_to_cast: ColoredManaCount,
 
@@ -2322,6 +2324,13 @@ impl GameObject {
         // is a new object on any re-entry — clear the stale cast provenance.
         self.cast_from_zone = None;
         self.cast_controller = None;
+        // CR 601.2h + CR 400.7: The per-color cast tally is a historical fact
+        // for the object that actually resolved onto the battlefield from a
+        // cast, kept alive through CR 603.4 intervening-if re-checks on that
+        // same visit. Once the permanent leaves, any re-entry is a new object
+        // with no memory of how it was previously cast — clear alongside the
+        // other cast-entry provenance above (Emptiness / Adamant class).
+        self.colors_spent_to_cast = ColoredManaCount::default();
         // CR 611.2f: the cast-time keyword snapshot is bound to the same casting
         // event as `cast_from_zone`; clear it on the same zone-change boundary.
         self.cast_spell_keywords.clear();
@@ -2641,6 +2650,34 @@ mod tests {
     };
     use crate::types::counter::parse_counter_type;
     use crate::types::triggers::TriggerMode;
+
+    #[test]
+    fn reset_for_battlefield_exit_clears_colors_spent_to_cast() {
+        use crate::types::mana::ManaColor;
+
+        let mut obj = GameObject::new(
+            ObjectId(1),
+            CardId(1),
+            PlayerId(0),
+            "Emptiness".to_string(),
+            Zone::Battlefield,
+        );
+        obj.colors_spent_to_cast.add(ManaColor::White, 2);
+        obj.cast_from_zone = Some(Zone::Hand);
+
+        obj.reset_for_battlefield_exit();
+
+        assert_eq!(
+            obj.colors_spent_to_cast.get(ManaColor::White),
+            0,
+            "CR 400.7: a permanent that leaves the battlefield must not carry \
+             a stale cast-color tally into a later zone"
+        );
+        assert_eq!(
+            obj.cast_from_zone, None,
+            "cast provenance fields clear on the same exit boundary"
+        );
+    }
 
     #[test]
     fn game_object_has_all_rules_relevant_fields() {
--- a/crates/engine/src/game/game_object.rs
+++ b/crates/engine/src/game/game_object.rs
@@ -1042,9 +1042,9 @@ pub struct GameObject {
     /// refs. Persists through CR 603.4 intervening-if resolution re-checks on
     /// the **same** battlefield object (`clear_post_collection_transients`
     /// clears only the transient `mana_spent_to_cast` boolean, not this tally).
-    /// Cleared by `reset_for_battlefield_exit` (CR 400.7) so a permanent that
-    /// leaves and re-enters without being cast again does not inherit a stale
-    /// color tally from its previous visit.
+    /// Cleared by `reset_for_battlefield_exit` and when a spell leaves the stack
+    /// for any zone other than the battlefield (CR 400.7; CR 400.7d preserves
+    /// the tally only for the permanent the spell becomes).
     #[serde(default, skip_serializing_if = "ColoredManaCount::is_empty")]
     pub colors_spent_to_cast: ColoredManaCount,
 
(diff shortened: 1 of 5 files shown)


matthewevans (changes requested):
**Changes requested — the current-head fix retains cast colors across a new-object boundary.**

## 🔴 Blocker

`clear_post_collection_transients` now deliberately preserves `colors_spent_to_cast` so the original ETB's intervening-if can re-check it. That is the right lifetime for the original object, but `GameObject::reset_for_battlefield_exit` does not clear the tally (`crates/engine/src/game/game_object.rs:2291`), while the zone-move path calls that reset for every battlefield exit (`crates/engine/src/game/zones.rs:350-351`). A creature cast with `{W}{W}`, then blinked or returned without being cast again, can therefore enter with the old white tally and incorrectly satisfy a later `ManaColorSpent` ETB.

Please clear the cast-color tally at the battlefield-exit/new-object boundary (alongside the other cast-entry provenance), while retaining it through the original trigger-resolution window. Add a registered runtime regression that casts with the qualifying color, leaves the battlefield, re-enters without a new cast, and proves the color-spent condition does not fire.

The two existing Emptiness tests cover direct cast-to-ETB behavior only; they do not exercise the required reset boundary. CI/CodeRabbit and the engine parse-diff are also not yet settled for this head, so they remain independent merge gates.

**Recommendation: update this PR with the exit-boundary reset and regression test, then request re-review.**

andriypolanski:
Could you please let me know how to remove Contribution trust flagged mark?

matthewevans (changes requested):
**Changes requested — the battlefield-exit repair is correct, but cast-color provenance still leaks across a stack-to-nonbattlefield new-object boundary.**

## 🔴 Blocker

**[HIGH]** `colors_spent_to_cast` is cleared only on a battlefield exit, not when a cast spell leaves the stack without becoming that permanent. Evidence: `crates/engine/src/game/triggers.rs:4637-4641` preserves the tally in every zone while clearing only `cast_from_zone` outside battlefield/stack; `crates/engine/src/game/zones.rs:339-352` invokes the new reset only for `from == Zone::Battlefield`; and `game_object.rs:2327-2333` therefore cannot clear stack-to-graveyard/exile survivors. CR 400.7 says an object that changes zones is a new object with no memory of its prior existence; CR 400.7d only preserves cast-cost information for the permanent the spell becomes. Why it matters: a `{W}{W}` Emptiness countered to a graveyard and later returned without being cast retains the old tally and can incorrectly pass its white-spent ETB condition. Suggested fix: give cast-payment provenance one lifecycle authority that preserves it only for the stack spell and the permanent it becomes, clears it on stack-to-nonbattlefield transitions as well as battlefield exits, and add a countered-then-reanimated runtime regression.

## 🟡 Non-blocking

**[MED]** The blink regression reaches the raw delivery helper rather than a replacement-aware zone-change production entry. Evidence: `crates/engine/tests/integration/issue_5943_emptiness_evoke_color_spent.rs:146-163` calls `zones::move_to_zone` and `process_triggers` directly; the unresolved CodeRabbit thread identifies the same gap. Why it matters: the test proves the reset helper but not its behavior through a real proposed zone-change/replacement path. Suggested fix: drive the blink through an existing production effect or `ZoneMoveRequest` pipeline while retaining the post-reentry no-trigger assertion.

**Recommendation: extend the provenance lifecycle and its production-path regression, then request re-review.**

matthewevans (changes requested):
**Changes requested — the current head does not address the previously reported cast-provenance boundary.**

## 🔴 Blocker

**[HIGH]** `colors_spent_to_cast` still survives a spell leaving the stack without becoming its permanent. Evidence: `crates/engine/src/game/triggers.rs:4637-4641` preserves the tally for every zone, while `crates/engine/src/game/game_object.rs:2327-2333` clears it only from `reset_for_battlefield_exit`; the only current-head delta is the test's explicit normal-cast choice. Why it matters: a `{W}{W}` Emptiness countered to a graveyard and later returned can inherit its prior cast colors, contrary to CR 400.7; CR 400.7d's exception is only for the permanent the spell becomes. Suggested fix: make cast-payment provenance clear on stack-to-nonbattlefield transitions as well as battlefield exits, and cover countered-then-reanimated behavior.

**[HIGH]** The blink regression still calls raw `zones::move_to_zone` and `process_triggers` directly. Evidence: `crates/engine/tests/integration/issue_5943_emptiness_evoke_color_spent.rs:146-163`; the current CodeRabbit finding identifies the same path. Why it matters: replacements and normal zone-change trigger collection are not exercised. Suggested fix: drive exile and return through the replacement-aware production zone-change pipeline.

**Recommendation: implement both lifecycle and production-path regressions, then request re-review.**

matthewevans (changes requested):
## Blocking findings

1. `colors_spent_to_cast` is still only cleared by `reset_for_battlefield_exit`, which is called when `from == Zone::Battlefield`. A spell carries the tally while on the stack, so a countered spell moved Stack → graveyard/exile and later put onto the battlefield can still satisfy the Adamant ETB without being cast for that entry. CR 400.7 applies to the zone change, not only battlefield exits. Clear cast-bound state at the stack-to-nonbattlefield boundary too, and add a production-pipeline regression that counters the originally paid spell and reanimates/returns it; the current blink test only covers Battlefield → Exile → Battlefield.

2. The integration regression invokes `engine::game::zones::move_to_zone` directly for both blink legs. That validates the zone helper, but not the spell/effect resolution path the card behavior depends on. Please drive the blink through a real resolving effect/ability so the test would catch a future pipeline bypass.

The current head contains no contributor commit after the prior requested-changes review; this re-review is for the merge-main head update. The existing blocks remain unresolved.

matthewevans:
**Closing as superseded by #6352.**

Both PRs close #5943. The current head here fixes the original ETB and zone-exit cases, but it still clones `colors_spent_to_cast` into spell copies: `copy_spell.rs` clones the source `GameObject` and clears `cast_from_zone` without clearing the color-payment tally. A copied qualifying spell can therefore satisfy `ManaColorSpent`, contrary to CR 707.10.

#6352 covers the same reported defect and resets cast-payment stamps at the spell-copy birth sites as well as the zone-identity boundaries, with regressions for those paths. Keeping one implementation avoids two competing lifecycle changes for the same issue.

PR #6346, opened 2026-07-22
fix(engine): delayed ParentTarget snapshot must not bind the trigger source when zero targets were chosen (Depthshaker Titan)

### LINKED ISSUE #5901: Depthshaker Titan sacrifices itself without cause — [[Depthshaker Titan]] only calls to sacrifice the noncreature artif…
Reported in Discord: https://discord.com/channels/1485498006781427802/1526813203823329311/1526813203823329311

**Thread:** Depthshaker Titan sacrifices itself without cause
**Discord thread id:** `1526813203823329311`
**Discord message id:** `1526813203823329311`
**Cards:** depthshaker titan

## Summary
[[Depthshaker Titan]] only calls to sacrifice the noncreature artifacts chosen to become creatures, but it's sacrificing itself as well.

## Reported text

[[Depthshaker Titan]] only calls to sacrifice the noncreature artifacts chosen to become creatures, but it's sacrificing itself as well.

## Attachments

- [game-state-turn-31-2026-07-15T04-49-38-028Z.zip](https://cdn.discordapp.com/attachments/1526813203823329311/1526813204007751680/game-state-turn-31-2026-07-15T04-49-38-028Z.zip?ex=6a59b49a&is=6a58631a&hm=763518e20c3c35b7ec39d5379e6928b827ac16457f688ce2790157eebe9851cf&)

### depthshaker titan
When this creature enters, any number of target noncreature artifacts you control become 3/3 artifact creatures. Sacrifice them at the beginning of the next end step.
Each artifact creature you control has melee, trample, and haste. (Whenever a creature with melee attacks, it gets +1/+1 until end of turn for each opponent you attacked this combat.)

---
<sub>report_id: `discord:1526813203823329311:1526813203823329311:0` · discord: `1526813203823329311/1526813203823329311`</sub>

rsnetworkinginc:
Tier: Standard
Model: claude-fable-5
Thinking: high

Closes #5901

## What was broken

Depthshaker Titan — "When this creature enters, **any number of target** noncreature artifacts you control become 3/3 artifact creatures. **Sacrifice them** at the beginning of the next end step." — sacrificed **itself** at the next end step when the controller chose **zero** targets for the ETB trigger (a legal choice for an "any number of target …" slot, CR 115.1d).

## Root cause

The delayed trigger's inner `Sacrifice { target: ParentTarget }` snapshots its subject at creation time via `parent_target_snapshot` (`crates/engine/src/game/effects/delayed_trigger.rs`). The referent ladder was:

1. root-chain chosen targets (`parent_chain_targets_from_root`),
2. the node's own propagated `targets`,
3. the creation event's source object (`TriggeringSource`).

With zero chosen targets, tiers 1–2 are empty **because the chosen set is empty**, but the ladder couldn't tell that apart from "the parent has no target slots at all" — so it fell through to tier 3 and bound the ETB event's source, i.e. the Titan itself, as "them". At the next end step the delayed trigger dutifully sacrificed the Titan.

## Fix

`parent_target_snapshot` now returns the empty set instead of consulting the event-context fallback when the resolving root chain **declared a chooseable target slot** (`multi_target` — "any number of target …" — or `optional_targeting` — "up to one target …"). Reaching the fallback with such a declaration means the player legally chose zero targets, and the `ParentTarget` anaphor's referent is that empty chosen set (CR 608.2c), not the trigger's own source (CR 603.7c).

The event-source fallback is preserved unchanged for slotless parents, where it is load-bearing (a dies/LTB trigger's "exile **it** at the beginning of the next end step" — "it" genuinely names the event source).

Supporting refactor: the root-entry lookup previously inlined in `parent_chain_targets_from_root` (`game/targeting.rs`) is extracted into `resolving_root_ability` so the new predicate reads the same root the target flattening already uses — one authority, no duplicated stack walk.

## Tests

`crates/engine/tests/integration/issue_5901_depthshaker_titan.rs`, driving the real cast → ETB trigger targeting → delayed-trigger pipeline end to end:

- `depthshaker_titan_zero_targets_does_not_sacrifice_itself` — the reported bug. **Fails pre-fix** (Titan in graveyard at the end step, verified via stash) and passes post-fix; also asserts no sacrifice choice prompt surfaces and a bystander artifact is untouched.
- `depthshaker_titan_sacrifices_only_chosen_artifacts` — over-fix guard for the already-correct path: two chosen artifacts are animated and both ARE sacrificed at the next end step, while the Titan and an unchosen artifact survive. Passes pre- and post-fix.

## Verification

- `cargo fmt --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- full engine lib suite: green
- full integration suite: green

## Gate A

`bash scripts/check-parser-combinators.sh` → **Gate A PASS head=6b9e1295074bdaa02fb906477eccd338b9347202 base=21fc5f3b476021ea78e38b712cbffe0ae417a5d9** (Gate G PASS as well — no parser files are touched by this diff). `scripts/check-engine-authorities.sh` → Gate B PASS (37 classified hits), Gate D PASS (28 classified hits), draw-replacement producers PASS.

## Anchored on

- `crates/engine/src/game/effects/delayed_trigger.rs` — `parent_target_snapshot`'s own tier documentation (the CR 608.2c phase#4767 note): the existing ladder already distinguishes "root chain exposes no concrete slot because the target was runtime-injected" (Animate Dead) from the declared-slot case; this fix adds the missing third distinction — "slots declared, zero chosen" — at the same seam.
- `crates/engine/src/game/targeting.rs` — `parent_chain_targets_from_root`, the single root-entry lookup authority; the new `resolving_root_ability` is extracted from it (not duplicated) so the predicate reads the same root the target flattening reads.
- `crates/engine/src/game/targeting.rs` — `resolved_targets`' documented `None`/`ParentTarget` → source fallback (Rancor / Spirit Loop): the reason the event-source tier must be *preserved* for slotless parents rather than removed outright.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected delayed effects with zero selected targets so they no longer incorrectly affect the effect’s source.
  * Improved target resolution across nested abilities and optional or multi-target effects.
  * Fixed Depthshaker Titan interactions so only selected artifacts are sacrificed, while zero-target resolutions leave all permanents intact.

* **Tests**
  * Added regression coverage for selected-target and zero-target Depthshaker Titan scenarios.

### DIFF
--- a/crates/engine/src/game/effects/delayed_trigger.rs
+++ b/crates/engine/src/game/effects/delayed_trigger.rs
@@ -245,6 +245,23 @@ fn parent_target_snapshot(state: &GameState, ability: &ResolvedAbility) -> Vec<T
         return ability.targets.clone();
     }
 
+    // CR 601.2c + CR 608.2c (issue #5901): When the resolving root chain
+    // DECLARED a chooseable target slot — a `multi_target` bound ("any number
+    // of target noncreature artifacts", Depthshaker Titan) or an optional
+    // "up to one target" slot — reaching this point means the player legally
+    // chose ZERO targets: the two tiers above are empty precisely because the
+    // chosen set is empty. The ParentTarget anaphor ("them"/"it") refers to
+    // that empty chosen set, so the delayed trigger has no subject. Falling
+    // through to the triggering-source fallback instead bound the trigger's
+    // own event source — the Titan sacrificed ITSELF at the next end step.
+    // The fallback below remains for slotless parents (a dies/LTB trigger's
+    // "exile it at end of turn", where "it" genuinely names the event source).
+    if chain_declares_chooseable_target_slots(crate::game::targeting::resolving_root_ability(
+        state, ability,
+    )) {
+        return Vec::new();
+    }
+
     crate::game::targeting::resolve_event_context_target(
         state,
         &TargetFilter::TriggeringSource,
@@ -252,6 +269,25 @@ fn parent_target_snapshot(state: &GameState, ability: &ResolvedAbility) -> Vec<T
     )
     .map(|target| vec![target])
     .unwrap_or_default()
+}
+
+/// True when any link of the chain declares a target slot whose selection may
+/// legally be empty: a `multi_target` bound (CR 115.1d "any number of target
+/// ...") or `optional_targeting` (CR 115.1d "up to one target ..."). Used by
+/// [`parent_target_snapshot`] to distinguish "slots were declared but zero
+/// were chosen" (referent = empty set) from "no slots exist at all" (referent
+/// = the creation event's source object).
+fn chain_declares_chooseable_target_slots(ability: &ResolvedAbility) -> bool {
+    ability.multi_target.is_some()
+        || ability.optional_targeting
+        || ability
+            .sub_ability
+            .as_deref()
+            .is_some_and(chain_declares_chooseable_target_slots)
+        || ability
+            .else_ability
+            .as_deref()
+            .is_some_and(chain_declares_chooseable_target_slots)
 }
 
 fn triggering_source_destination_zone(state: &GameState) -> Option<Zone> {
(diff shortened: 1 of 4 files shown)


matthewevans (approved):
Approved after maintainer fixup: the end-step helper now fails on a non-settling loop, and the zero-target CR citations now correctly use CR 115.6 with CR 603.3d for triggered-ability target selection.

PR #6526, opened 2026-07-23
fix(engine): Ovika parses 'mana value of that spell' of-form, not just possessive

### LINKED ISSUE #1718: Ovika, Enigma Goliath does not create Goblin tokens when casting noncreature spells
## Bug description
**Ovika, Enigma Goliath** does not appear to trigger when its controller casts noncreature spells. Because of this, the expected Goblin tokens are not being created.

## Current behavior
- Ovika is on the battlefield.
- The Ovika player casts a noncreature spell.
- Ovika's ability does not trigger or does not resolve properly.
- No Goblin tokens are created.

## Expected behavior
Whenever Ovika's controller casts a noncreature spell, Ovika should trigger and create X 1/1 red Phyrexian Goblin creature tokens, where X is that spell's mana value. Those tokens should gain haste until end of turn.

Expected card text behavior:
> Whenever you cast a noncreature spell, create X 1/1 red Phyrexian Goblin creature tokens, where X is the mana value of that spell. They gain haste until end of turn.

## Notes
This may be an issue with Ovika's card-specific triggered ability or with noncreature-spell cast triggers in general.

rsnetworkinginc:
fix(parser): bind "the mana value of that spell" of-form (Ovika, Enigma Goliath)

Closes #1718

## Bug

**Ovika, Enigma Goliath** never creates its Goblin tokens. Its trigger —
"Whenever you cast a noncreature spell, create X 1/1 red Phyrexian Goblin
creature tokens, where X is the mana value of that spell. They gain haste until
end of turn." — fires correctly, but the token-creation effect silently lowers
to `Effect::Unimplemented`, so zero tokens are produced (the exact reported
symptom).

## Root cause

`parse_object_mana_value_ref` (`crates/engine/src/parser/oracle_nom/quantity.rs:2945`)
is the shared recognizer for object mana-value quantity references. It handled
the **possessive** front-form `"that spell's mana value"` → `ObjectManaValue {
scope: EventSource }` (via `parse_object_possessive_scope`, line 4535), but the
**prepositional of-form** `"the mana value of that spell"` — Ovika's actual
printed wording — was only accepted when it named a `target` object (yielding
`TargetObjectManaValue`). For a non-target anaphor the `?` on
`parse_target_with_syntax_target_keyword` propagated an error out of the whole
function, so the quantity failed to bind.

Consequently the token clause's `where X is …` binder
(`token.rs` → `parse_where_x_quantity_expression` → `parse_cda_quantity`) got
`None`, and per its own honest-failure contract the entire `create X … tokens`
clause dropped to `Effect::Unimplemented` rather than fabricating a dead
count-0 placeholder.

Verified directly before/after the fix:
`parse_cda_quantity("the mana value of that spell")` returned `None` on `main`
and now returns `ObjectManaValue { scope: EventSource }`; the full Ovika parse's
trigger effect went from `Unimplemented` to
`Token { name: "Phyrexian Goblin", 1/1, Red, count: ObjectManaValue { EventSource } }`.

## Fix

Accept the non-target prepositional of-form by mirroring the possessive scope
mapping. `parse_object_mana_value_ref` now consumes an optional leading `"the "`
(like the sibling `parse_cost_paid_object_prepositional_ref`) and, when the
targeted path does not match, falls back to a new
`parse_object_prepositional_anaphor_scope` combinator
(`quantity.rs:3017`): `that spell` / `the triggering spell` → `EventSource`,
`that creature` / `that permanent` / `that planeswalker` → `Target`, `this …` /
`~` → `Source`. The targeted `TargetObjectManaValue` path is still tried first,
so `"mana value of target creature"` is unchanged. No runtime change was needed
— `ObjectManaValue { EventSource }` already resolves to the triggering spell on a
`SpellCast` trigger (same scope the possessive form produced).

## Anchored on

- `parse_object_possessive_scope` (`crates/engine/src/parser/oracle_nom/quantity.rs:4535`)
  — the possessive front-form scope table (`"that spell's"` → `EventSource`,
  `"that creature's"` → `Target`, …). The new of-form combinator reproduces this
  exact mapping so both surface phrasings bind identically.
- `parse_cost_paid_object_prepositional_ref` (`quantity.rs:3094`) — the existing
  sibling that reads `"[the] mana value of the <participle> <noun>"` as the
  prepositional mirror of a possessive cost-paid ref; the optional-`"the "`
  handling and "prepositional mirrors possessive" pattern follow it.
- `parse_object_color_of_scope` (`quantity.rs:4556`) — an existing of-form
  anaphor→`ObjectScope` combinator (for color references) with the same
  `that spell → EventSource` / `that creature → Target` mapping, confirming this
  anaphor-scope shape is the idiomatic building block here.

## Regression tests (RED before, GREEN after)

- Runtime, real cast pipeline —
  `crates/engine/src/game/casting_tests.rs:36295`
  (`mod ovika_noncreature_spell_token_trigger`), mirroring the sibling Namor
  cast-trigger tests:
  - `casting_mana_value_three_spell_creates_three_goblins` — Ovika on the
    battlefield, cast a mana-value-3 noncreature spell, resolve the stack:
    exactly **3** tokens, each a red **Goblin** that has **haste**. Pre-fix:
    0 tokens.
  - `token_count_tracks_spell_mana_value` — control: a mana-value-2 spell makes
    exactly 2 tokens, proving the count reads the spell's mana value rather than
    a fixed number. Pre-fix: 0 tokens.
- Parser unit — `crates/engine/src/parser/oracle_nom/quantity.rs:10730`
  (`mana_value_of_form_mirrors_possessive_scope`): each of-form phrase binds the
  same `ObjectManaValue` scope as its possessive counterpart, plus a negative
  control that `"mana value of target creature"` still routes to
  `TargetObjectManaValue`.

RED confirmed by stashing only the `quantity.rs` production change and re-running
the runtime tests on the unmodified base: both failed with `got 0` tokens.

## Validation (full CI-parity gate, all green)

- `cargo clippy --workspace --all-targets -- -D warnings` — clean (incl.
  `engine-wasm` / `draft-wasm`)
- `cargo test -p engine` — lib **17568 passed / 0 failed / 6 ignored**;
  integration **3835 passed / 0 failed / 2 ignored**
- `./scripts/check-parser-combinators.sh` — `Gate G PASS` +
  `Gate A PASS head=b22369f8a567c4adf62e92716fcfa873beccc146 base=70e4f2d77e0fd8de000c97bc2718924cb2505515`
- `./scripts/check-legacy-quantity-callsites.sh` — exit 0
- Rebases clean on `origin/main` (`70e4f2d`, up to date).

## Gate A

```
Gate A PASS head=b22369f8a567c4adf62e92716fcfa873beccc146 base=70e4f2d77e0fd8de000c97bc2718924cb2505515
```

Model: Claude Opus 4.8
Tier: scoped bugfix (single parser combinator; new enum-free `alt`-composed combinator, no runtime change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of “mana value of …” / “converted mana cost of …” (including optional “the”) so referenced object scope is resolved correctly.
  * Confirmed that “mana value of target …” remains treated as the targeted form.

* **Tests**
  * Added coverage for Ovika’s cast-trigger, verifying noncreature spells generate the correct number of 1/1 red Phyrexian Goblin tokens with haste based on mana value.
  * Added regression coverage for “mana value of …” phrasing and added checks for Pure Reflection’s token stats derived from mana value.


matthewevans (changes requested):
## Blocking finding

`parse_object_prepositional_anaphor_scope` duplicates the existing `parse_object_color_of_scope` object-scope grammar while covering fewer sibling forms (`it`, `the enchanted creature`, and `the equipped creature` are missing). The new function’s “mirrors” claim is therefore not true, and future object-property grammar will drift.

Please reuse/parameterize the existing prepositional object-scope parser for the mana-value fallback (after the targeted-of-form path), retaining its recipient and target sibling coverage. The Ovika runtime test is a good production-path proof for the target case; Scryfall’s current Oracle text confirms the intended “mana value of that spell” clause.

CodeRabbit’s request to assert every token characteristic is non-blocking test-strengthening feedback; this architecture issue is the blocking item.

rsnetworkinginc:
Addressed the blocking architecture finding: the duplicate object-scope table is gone.

**What changed**
- Removed `parse_object_prepositional_anaphor_scope` entirely.
- The mana-value "of"-form fallback (tried *after* the targeted `parse_target_with_syntax_target_keyword` path, as requested) now binds its object through the existing shared prepositional object-scope parser — the one `parse_color_of_object_for_each` and `parse_object_typeline_scope` already use. So the mana-value axis now inherits its **full** sibling coverage, including the three forms the duplicate table was missing: `it`, `the enchanted creature`, `the equipped creature` (all `ObjectScope::Recipient`, and `resolve_object_mana_value` already handles `Recipient` via `object_for_scope`).
- Renamed that shared parser `parse_object_color_of_scope` -> `parse_object_prepositional_scope` and documented it as the property-agnostic "of"-form sibling of `parse_object_possessive_scope`. The old name was already inaccurate for the typeline caller; keeping it while adding a third property would have invited exactly the drift you flagged. Call sites updated (5), no behavior change for the existing colors/typeline callers.

**Tests**
- `mana_value_of_form_mirrors_possessive_scope` extended with the three inherited recipient forms, so a future regression that narrows the shared table fails here.
- Negative control unchanged: `mana value of target creature` still routes to `TargetObjectManaValue`.
- Ovika runtime test (production path, 3 tokens for a mana-value-3 spell, plus the mana-value-2 control) unchanged.

**Full local parity on this head (isolated CARGO_TARGET_DIR, rebased onto `c754a087a`)**
- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --all-targets --features engine/proptest -- -D warnings`: clean
- `cargo test -p engine --lib`: 17593 passed / 0 failed
- `cargo test -p engine --test integration`: 3869 passed / 0 failed

matthewevans (commented):
**Held on current head `9e42854b`.** The prior duplicated object-scope grammar blocker is resolved: the mana-value of-form now delegates to the shared prepositional scope parser, and the Ovika regression drives the cast/trigger pipeline at two mana values. Required Rust verification is still in progress and Tauri currently reports a failed job, so this is not ready for approval or enqueue.

matthewevans:
**One contributor-side blocker on `9e42854b`: the new test trips the engine single-authority gate. The `Tauri compile check` red on this head is ours, not yours — ignore it.**

## 🔴 Blocker

- `crates/engine/src/game/casting_tests.rs` (added line `obj.keywords.contains(&Keyword::Haste),`) — this is a raw keyword query and the **Engine authority gate** (`scripts/check-engine-authorities.sh`, section A) reds on it. Verbatim CI output from run [30027016689](https://github.com/phase-rs/phase/actions/runs/30027016689):

  ```
  ERROR: New engine code bypasses a single-authority helper.

  (A) Raw keyword query — use the authorities in game/keywords.rs:
      obj.keywords.contains(&kw)        ->  keywords::has_keyword(obj, &kw)

  Forbidden in added lines (diff vs 6ae8737cdab0fa1ed291cad0f0808473a90f4cf8):

    crates/engine/src/game/casting_tests.rs:
                      obj.keywords.contains(&Keyword::Haste),
  ```

  **Fix:** route the assertion through the authority — `keywords::has_keyword(obj, &Keyword::Haste)` — rather than reading the field. A raw `.keywords.contains()` silently misses off-zone keyword grants, which is exactly the class of bug the gate exists to prevent. If the read is genuinely structural, the gate's escape hatch is a `// allow-raw-authority: <one-line reason>` annotation on the line, but for a battlefield-object assertion the helper is the correct call.

  This is the only failure in `Rust lint (fmt, clippy, parser gate)` — Gates A, B, D, G and the parser-combinator gate all pass on this head.

## 🟡 Non-blocking — not your defect

- **`Tauri compile check` is failing on `main` itself, not because of this PR.** Evidence: the release commit `21a53d50cb` ("release: v0.35.2") bumped `client/src-tauri/Cargo.toml` to `0.35.2` without regenerating `client/src-tauri/Cargo.lock`, which still records `phase-tauri 0.35.1`. The job builds with `--locked`, so cargo refuses:

  ```
  ```

  The same failure is currently red on `main`'s own CI run and on unrelated PRs. **Do not try to fix this** — a maintainer is handling the lockfile. Once it lands, rerun CI here.

## ✅ Clean

- The parser change in `crates/engine/src/parser/oracle_nom/quantity.rs` is scoped to the shared object-scope combinator rather than card-special-cased, and `Rust tests (shard 1/2 + 2/2)`, `Card data (generate, validate, coverage)`, `WASM compile check`, and `Frontend (lint, type-check, test)` are all green on this head.
- The `<!-- coverage-parse-diff -->` artifact is present and current for `9e42854b`.

**Recommendation: push one commit swapping the raw keyword read for `keywords::has_keyword`. That is the only thing standing between this head and a merge-queue enqueue — the Tauri red will clear on its own once the maintainer-side lockfile fix lands.**

rsnetworkinginc:
Pushed a one-line fix for the failing Engine authority gate: the new Ovika token-creation test used a raw `keywords.contains(&Keyword::Haste)`, which the gate flags. It asserts the literal keyword set stamped on the freshly created token (not an effective-keyword query), so it now carries the documented `// allow-raw-authority:` annotation. Verified locally with `./scripts/check-engine-authorities.sh origin/main` (Gate B/D PASS, no bypass error) and `cargo fmt --all -- --check` clean.

The remaining `Tauri compile check` failure is an unrelated `client/src-tauri/Cargo.lock --locked` staleness in the Tauri toolchain job, not touched by this parser-only change.

matthewevans:
**The authority-gate blocker is confirmed fixed on `4b83ea28e`, and both remaining CI reds are ours. One test-adequacy ask before enqueue.**

## ✅ The blocker is resolved — and CI proves it, not just my reading

Your delta `9e42854b..4b83ea28e` is exactly the claimed one-liner: one file, one line, comment-only.

```rust
obj.keywords.contains(&Keyword::Haste), // allow-raw-authority: asserts the literal
// keyword set stamped on the freshly created token, not an effective-keyword query
```

The mechanism works for a trailing same-line comment — `scripts/check-engine-authorities.sh:102` pre-filters added lines through `grep -Ev 'allow-raw-authority'` *before* the forbidden-pattern grep, and `:78` is the same-line escape inside `filter_allow_annotation`. But the decisive evidence is empirical: the gate runs as the `Engine authority gate` step (`.github/workflows/ci.yml:75-86`) inside the job that surfaces as **`Rust lint (fmt, clippy, parser gate)`, which is `SUCCESS` on this head.** And the justification in the annotation is the right one — you're asserting the literal stamped keyword set on a freshly created token, which is exactly the case the escape hatch exists for, not a smuggled effective-keyword query.

## 📋 Both CI reds are ours — nothing for you to do

| Check | Cause |
|---|---|
| `Tauri compile check` | **Root.** `error: cannot update the lock file client/src-tauri/Cargo.lock because --locked was passed to prevent this` — the `v0.35.2` release commit bumped `client/src-tauri/Cargo.toml` without regenerating its lock. |
| `Rust (fmt, clippy, test, coverage-gate)` | **Derivative.** Pure fan-in: `RUST_LINT=success RUST_TEST=success CARD_DATA=success DRAFT_POOLS=skipped WASM=success TAURI=failure`. |

Everything else on this head is green — both Rust test shards, Card data, WASM, Frontend, Lobby worker.

Attributed by ancestry rather than timestamps, with the probe run first so the test isn't vacuously exonerating: `merge-base(origin/main, head)` = `c754a087a2`, and `--is-ancestor c754a087a2 4b83ea28e` → **ancestor** (test discriminates). Then `cb7d5a25fa` (#6568, the lock sync) → **not an ancestor**, and `6a3d080aee` (#6574, which prevents recurrence) → also not on the branch. Your head is internally consistent (`Cargo.toml` 0.35.0 / lock `phase-tauri 0.35.0`); the breakage exists only in the ephemeral merge ref computed against a `main` that had the bump but not yet the fix. Your diff is two files, neither of them `Cargo.toml` nor `Cargo.lock`.

**We'll bring the branch current** — it's `BEHIND` anyway, and current `main` carries both fixes, so one update clears both reds.

## ✅ Architecture fix — re-verified independently

I didn't rely on my earlier confirmation. `parse_object_color_of_scope` → `parse_object_prepositional_scope` (`quantity.rs:4546`) is a genuine single property-agnostic of-form table, not a duplicated grammar:

- Zero stale references to the old name at head (grep exits 1).
- Four call sites share it: mana value `:3002`, typeline `:4368`, color-for-each `:4412`, number-of-colors tail `:4444`.
- It's a true of-form mirror of `parse_object_possessive_scope` (`:4520-4535`) — same referent set, of-form spellings.
- The `opt(tag("the "))` at `:2957` is correctly scoped, and your code comment's claim holds: the possessive fallback at `:3006` re-parses from the original `input`, so article consumption only affects the of-form branch.
- No shadowing regression — `:3002` propagates `Err` via `?` for unmatched scopes exactly as the old `parse_target_with_syntax_target_keyword(rest)?` did, so the caller `alt()` at `:862` still falls through to the cost-paid-object prepositional branch (Morbid Curiosity, *"the mana value of the sacrificed permanent"*). The new code returns `Ok` in strictly more cases, never fewer.

## ✅ The parse-diff strengthens the PR — mention it in the description

Sticky is current (updated `00:23:39Z`, after your `22:43Z` push), baseline `21a53d50cb`: **2 cards / 3 signatures**, `create` placeholder removed and a real `Token` signature added on both.

The PR claims Ovika only, but a second card moved — **Pure Reflection**. I verified its Oracle text against Scryfall rather than assuming: *"Whenever a player casts a creature spell, destroy all Reflections. Then that player creates an X/X white Reflection creature token, **where X is the mana value of that spell**."* Identical of-form construction. That's in-class coverage, and it's the "build for the class, not the card" evidence — worth one line in the description so the 2-card delta is self-explaining.

## 🟡 One ask before enqueue

**Pin the token's P/T** — `casting_tests.rs` (patch lines 64–82). The comment above the loop promises *"Every created token is a red Phyrexian Goblin, not a generic token"*, but only `Goblin` and `Haste` are asserted; 1/1 P/T, red, and `Phyrexian` are not.

This matters more here than the usual "assert what the comment claims" nit, and the parse-diff is what sharpens it: your seam binds *event source's mana value*, and the two cards in the diff bind it to **different axes** — Ovika to the token **count**, Pure Reflection to the token **P/T**. With count pinned at 3 but P/T unpinned, a count↔P/T axis confusion (three 3/3 tokens instead of three 1/1s) passes green. Adding `power == 1 && toughness == 1` inside the existing loop is the discriminating assertion; red in `colors` and `Phyrexian` in `subtypes` come along for free.

This confirms CodeRabbit's inline finding at `casting_tests.rs:36362` (🟠 Major, still open) — I'm not returning a verdict softer than it.

**Nit:** `CR 601.2f` on the mana-pool funding comment (patch line 45) is slightly off — 601.2f is total-cost *determination*; funding and payment are 601.2g/h. Test comment only.

## 📋 Process note

My `CHANGES_REQUESTED` from `2026-07-23T01:09:12Z` is still the standing review, so it has to be superseded by an approve before the queue will take this.

---

**Recommendation: add the P/T assertion, and we'll update your branch and approve.** The CR annotations all grep-verify and describe their code — CR 202.3 (`:1359`), 603.2 (`:2561`), 608.2c (`:2793`), 115.1 (`:838`), 107.4 (`:494`). The shared-grammar refactor is the right shape and it demonstrably generalized past the card you were fixing.

matthewevans:
**Your round-4 blocker is resolved and I'm clearing it. The red is ours, not yours — don't touch it.** Reviewed at head `4b83ea28e`.

## The Tauri failure is our lockfile skew

The `Rust (fmt, clippy, test, coverage-gate)` aggregator is pure fan-in, so its red names only *that* something failed. Opening the child job:

```
RUST_LINT_RESULT: success   RUST_TEST_RESULT: success   CARD_DATA_RESULT: success
DRAFT_POOLS_RESULT: skipped WASM_RESULT: success        TAURI_RESULT: failure
```

Only Tauri, and the real error is:

```
error: cannot update the lock file … client/src-tauri/Cargo.lock
```

`21a53d50cb` (release: v0.35.2) bumped `client/src-tauri/Cargo.toml` without regenerating the lock. We fixed it on `main` in `cb7d5a25fa` (#6568) at 22:56:02Z; your CI run was created at 22:43:22Z, **13 minutes before our fix landed**. Confirmed by ancestry:

```
git merge-base --is-ancestor 21a53d50cb 4b83ea28e  -> not ancestor
git merge-base --is-ancestor cb7d5a25fa 4b83ea28e  -> not ancestor
git merge-base --is-ancestor origin/main 4b83ea28e -> not ancestor   # non-vacuity probe, live
git rev-list --left-right --count origin/main...4b83ea28e -> 24  2
```

Your diff touches exactly two files, both engine (`oracle_nom/quantity.rs`, `casting_tests.rs`) — no manifest, no lockfile, nothing Tauri. **We'll bring the branch current; you don't need to do anything about that red.**

## Clearing the round-4 blocker — and correcting myself on it

The keyword-assertion blocker is genuinely resolved. Your head commit annotates it with the gate's documented escape hatch:

```rust
obj.keywords.contains(&Keyword::Haste), // allow-raw-authority: asserts the literal keyword set stamped on the freshly created token, not an effective-keyword query
```

and `scripts/check-engine-authorities.sh:102` filters added lines through `grep -Ev 'allow-raw-authority'`. CI proves it — `Rust lint (fmt, clippy, parser gate)` passes on this exact head.

I also want to correct the rationale I gave you in round 4. I said a raw `.keywords.contains()` "silently misses off-zone keyword grants," implying the helper would be safer. That doesn't actually separate the two options: `keywords::has_keyword` is *itself* a raw field read (`game/keywords.rs:31-36` is `obj.keywords.iter().any(...)`, self-annotated `allow-raw-authority: this IS the object-scoped authority`), and its own doc comment carries the identical battlefield-only caveat, deferring off-zone cases to `object_has_effective_keyword_kind`. Your tokens are on the battlefield, so the hatch and the helper are semantically equivalent here. Your reasoning was right and my objection was weaker than I presented it. A helper swap would have been house style, not correctness.

## What's genuinely good here

Round 3's duplicate-grammar blocker is properly fixed rather than worked around. The of-form now delegates to the shared `parse_object_prepositional_scope` (`quantity.rs:4546`), the rename is complete — zero surviving references to the old `parse_object_color_of_scope` anywhere in `crates/` — and all four call sites point at the shared table. No new enum variant, no bool, no verbatim Oracle string match, combinators throughout.

The `opt(tag("the "))` addition can't regress the possessive path: it's bound to a separate `of_form_input` at `quantity.rs:2954` and the possessive fallback at `:3006` re-parses from the original `input`. Targeted-form ordering is preserved with an explicit negative control that `"mana value of target creature"` stays `TargetObjectManaValue`.

I verified Ovika's Oracle text from Scryfall rather than your description, and confirmed against the shipped `card-data.json` that the pre-fix state was an honest `Unimplemented { name: "create", … }`. So this converts honest-red to live rather than papering over a silent misparse. All six CR citations grep-verify (202.3, 115.1, 608.2c, 701.20b, 105.1, 601.2f).

**The parse-diff measures 2 cards, not 1** — Pure Reflection also gains a `Token` signature, because its text carries the identical "where X is the mana value of that spell" of-form. That's a class win, and worth naming in the PR body.

## 🟡 One open item before I enqueue

**The token assertions verify two of the five characteristics the test claims.** `casting_tests.rs:36344-36362` asserts the `Goblin` subtype and `Haste`, while the comment directly above claims "Every created token is a red Phyrexian Goblin". No color assertion, no `Phyrexian` subtype, no 1/1 P/T. This is CodeRabbit's open 🟠 finding, and I re-read the code at this head — it's unchanged and still valid. It matters more than usual here because the whole `Token` effect is newly live in this PR, so nothing else in the suite pins those characteristics.

**[LOW]** Both runtime fixtures use generic-only mana costs (`ManaCost::generic(mv)` funded with `ManaType::Colorless`), so a mana-value computation that ignored colored pips would still pass — and `ObjectManaValue` is the value under test. Making one case `{1}{R}` would exercise generic+colored summation.

That's a ~6-line test-only commit. Push it and I'll enqueue.

matthewevans (changes requested):
**Changes requested — the parser seam is sound, but the runtime proof misses a changed card class.**

## 🔴 Blocker

**[MED]** Cover mana-value-derived token P/T through the production path. Evidence: `crates/engine/src/game/casting_tests.rs:36344-36362` asserts Ovika's token count/Goblin/Haste but not its power or toughness; the current parse diff also changes Pure Reflection's mana-value-derived token P/T. A wrong `ObjectManaValue { EventSource }` binding can therefore pass. Add a Pure Reflection runtime test that asserts the created token's P/T and strengthen Ovika's characteristic assertion.

**Recommendation: request changes; add the discriminating runtime coverage and re-run the current parse-diff.**

rsnetworkinginc:
Pushed 1d5376e addressing the remaining review items:

- **Ovika characteristic assertions**: the per-token loop now also pins power/toughness to exactly 1/1, red in colors, and the Phyrexian subtype, matching the comment's claim. With count pinned at 3 and P/T at 1/1, a count-vs-P/T axis confusion can no longer pass.
- **Pure Reflection runtime module** (`pure_reflection_mana_value_token_pt`): drives the production cast/trigger path with Pure Reflection on the battlefield and asserts the created Reflection token's P/T equal the triggering creature spell's mana value (plus white color, Reflection subtype, and exactly one token). Oracle text taken verbatim from the card.
- **Generic+colored cost**: the second Pure Reflection case casts a `{1}{R}` spell funded with one red + one colorless unit and expects a 2/2 — a computation that ignored colored pips would produce a 1/1 and fail.
- **Citation fix**: the mana-pool funding comment now cites CR 601.2g-h (601.2f is total-cost determination).

Verification: `cargo test -p engine --lib` — `casting_mana_value_three_spell_creates_three_goblins`, `token_count_tracks_spell_mana_value`, `reflection_token_pt_binds_spell_mana_value`, and `reflection_token_pt_sums_generic_and_colored_pips` all pass. Discrimination checked by temporarily asserting wrong values: Ovika tokens asserted as 3/3 fails (actual 1/1), and both Reflection cases asserted as 1/1 fail (actual 3/3 and 2/2 respectively), then restored. `cargo fmt --check` and `scripts/check-engine-authorities.sh` against the merge-base are clean; no new raw keyword queries were added.

matthewevans:
Maintainer acknowledgement: rechecked current head 1d5376e212c7a31b3ec01aa29ece994d7df5e00b after the latest follow-up. The reviewed two-line CR 107.3 citation correction is preserved in a local maintainer fixup, but is blocked from publication because the authorized fork-push GH_TOKEN is unavailable. No further author action is requested.

PR #6530, opened 2026-07-23
Fix desktop card art data loading

matthewevans:
## Summary\n- allow the remote shell to fetch shared R2 card/image data\n- lock that CSP requirement with a focused Tauri config test\n\n## Verification\n- cargo fmt --all\n- jq CSP assertion\n- git diff --check\n\nThe running desktop shell showed the active deck but an empty art header: its CSP allowed Scryfall image URLs but blocked the R2 Scryfall-data fetch needed to resolve them.

### DIFF
--- a/client/src-tauri/src/lib.rs
+++ b/client/src-tauri/src/lib.rs
@@ -102,4 +102,16 @@ mod tests {
             "must stay false so run()'s setup hook is the only window creator"
         );
     }
+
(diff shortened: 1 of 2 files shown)


matthewevans:
Closing: desktop deck tiles render the same data and Scryfall image path, so the CSP change does not address the dashboard-only missing-art defect.

PR #6538, opened 2026-07-23
fix(engine): unearth's leave-battlefield exile replacement must survive layer resets (Rotting Rats + Sacrifice)

### LINKED ISSUE #5976: Unearth bug — If I return [[Rotting Rats]] to the battlefield using its unearth ability and then sacrifice it to [[Sacr…
Reported in Discord: https://discord.com/channels/1485498006781427802/1525071374354223114/1525071374354223114

**Thread:** Unearth bug
**Discord thread id:** `1525071374354223114`
**Discord message id:** `1525071374354223114`
**Cards:** rotting rats, sacrifice
**Parser status:** has_gaps

## Summary
If I return [[Rotting Rats]] to the battlefield using its unearth ability and then sacrifice it to [[Sacrifice]], the rats die, but they should be exi

### rotting rats
When this creature enters, each player discards a card.
Unearth {1}{B} ({1}{B}: Return this card from your graveyard to the battlefield. It gains haste. Exile it at the beginning of the next end step or if it would leave the battlefield. Unearth only as a sorcery.)
### sacrifice
As an additional cost to cast this spell, sacrifice a creature.
Add an amount of {B} equal to the sacrificed creature's mana value.

---
<sub>report_id: `discord:1525071374354223114:1525071374354223114:0` · discord: `1525071374354223114/1525071374354223114`</sub>

michiot05:
Tier: Frontier
Model: claude-opus-4-8

Closes #5976.

## Summary

**Rotting Rats** returned via **Unearth** and then sacrificed to **Sacrifice** went to the graveyard instead of exile. Unearth's rider — *"Exile it at the beginning of the next end step **or if it would leave the battlefield**"* (CR 702.84a) — has two halves, and only the end-step delayed trigger worked; the "if it would leave the battlefield, exile it instead" replacement did nothing on a sacrifice, destroy, or bounce.

Root cause is **not** the sacrifice path (that correctly routes through `sacrifice_permanent` → `replace_event`, and the matcher accepts the rider). The rider was installed onto the returned creature's **live** `replacement_definitions` only. `seed_live_characteristics_from_base` reseeds live characteristics from base on **every layer pass** (CR 613.1) — and unearth's return grants haste, forcing an immediate layer evaluation — so the rider was wiped before any real leave event ever reached it. The end-step half survived only because it is a state-hosted delayed trigger, not an object-hosted replacement.

Class-level fix, not a Rotting-Rats special case: a typed `RestrictionExpiry::UntilHostLeavesPlay`, stamped on the rider at both build sites (the unearth synthesizer and the shared `try_parse_leave_battlefield_exile_replacement` parser builder that also handles the Whip-of-Erebos rider class). The stamp threads through the three existing runtime-rider seams that the turn-bound die-exile rider (Torch the Tower) already established:

1. **Base-install** (`add_target_replacement.rs`) — the rider persists in `base_replacement_definitions`, so it survives every CR 613.1 reseed.
2. **Non-copiable** (`printed_cards.rs`, CR 707.2) — token/merge/augment copies never inherit it (mana isn't copied; a copy isn't the cast object).
3. **Host-leave prune** (`layers.rs`, CR 400.7) — on the host's battlefield exit (including the exile redirect itself) the rider is dropped from base+live, so a re-entered same-`ObjectId` object is a new object and is not wrongly re-exiled.

Because the rider is bound to the object (`valid_card: SelfRef`), it survives control theft — a stolen unearthed creature sacrificed by its new controller is still exiled.

Method: /engine-implementer — plan → /review-engine-plan (1 round; findings integrated: corrected class inventory, verbatim-card-text discipline, prune reach-guard, variant-doc scoping) → implement → verification pass (one test-harness foot-gun fixed, no production defect) → /review-impl (CLEAN, zero findings) → commit → final read-only /review-impl: **PASS head=a65dd1a2229f2c231fab50579e346df2901e2239**. Each step in a fresh agent context.

## Gate A

`./scripts/check-parser-combinators.sh` → **Gate A PASS head=a65dd1a2229f2c231fab50579e346df2901e2239** (Gate G PASS). The parser change is a single `.expiry(...)` stamp on an already-built `ReplacementDefinition` — no string probes, no nom-chain change.

## Anchored on

- `crates/engine/src/database/unearth.rs` (`leaves_battlefield_exile_step`) — the synthesized unearth rider, now stamped; and `crates/engine/src/game/layers.rs:1692` (`seed_live_characteristics_from_base`) — the CR 613.1 reseed that wiped the live-only rider (the root cause).
- The **turn-bound die-exile rider** precedent (Torch the Tower, `EndOfTurn` expiry) — the exact three-seam durable-runtime-rider pattern this change mirrors: `is_runtime_target_die_exile_replacement` (`printed_cards.rs`) + `install_to_base` (`add_target_replacement.rs`) + the host-left prune (`layers.rs::prune_controller_controls_source_on_leave`). The new predicate `is_runtime_host_lifetime_replacement` is the object-lifetime sibling.
- `crates/engine/src/types/ability.rs` — `Duration::UntilHostLeavesPlay` (the existing object-lifetime vocabulary) established that host-bound and turn-bound lifetimes coexist as distinct categorical members; the new `RestrictionExpiry` variant is the replacement-expiry counterpart.

## Verification

- `cargo fmt --all` clean; `cargo clippy -p engine --tests` **0 warnings**.
- **New integration file** (`crates/engine/tests/integration/issue_5976_unearth_leave_redirect.rs`, registered in `main.rs`) — **8 rows**, each driving the real pipeline (unearth activate/resolve → a genuine layer pass → real leave event → zone assert, with `PermanentSacrificed`/damage/priority reach guards): the reported additional-cost sacrifice (verbatim **Sacrifice** card) → exile; ability-cost sacrifice outlet → exile; destroy → exile; bounce → exile (any leave event, CR 702.84a); non-unearthed sibling → graveyard; host-exit prune with no revival on re-entry; control-theft object-binding; and the real parser-rider card **Gruesome Encore** (verbatim) → bounce → exile.
- **Unit tests**: unearth base-install + reseed-survival; the two `printed_cards` predicate tests (non-copiable exclusion; and the **Personal Decoy** printed-static shape — no expiry stamp — staying copiable, the negative that proves the detector keys on the stamp, not the redirect shape); parser expiry-stamp assertion. All green.
- **Regression sweep**: `cargo test -p engine --lib -- replacement layer copiable torch_the_tower ninjutsu die_exile` → **2398 passed, 0 failed** (Torch-the-Tower `EndOfTurn` die-exile, copy/merge/augment, layers, ninjutsu all intact).
- **Fails before, passes after** (tests written first): with the base-install stamp reverted, R1/R3/R4/R7/R8/R9 flip to `Graveyard`/`Hand` (`left: Graveyard, right: Exile`) while the non-unearthed sibling R5 stays green; restored → all pass.
- **CR annotations**: every number grep-verified against `docs/MagicCompRules.txt` — CR 702.84a, 400.7, 613.1, 614.1a, 707.2, 611.2b.

## Claimed parse impact

Regenerated `card-data.json` and isolated the change from the pre-existing `Duration::UntilHostLeavesPlay` namesake (288 pre-existing `duration` occurrences on the unrelated "exile until ~ leaves" family, untouched): **exactly 64 cards** gain an additive `expiry: UntilHostLeavesPlay` field on their existing leave-battlefield exile replacement — the complete class (~58 unearth cards: Rotting Rats, Hellspark Elemental, Fatestitcher, the Scrapwork/Dregscape cycle, Sedraxis Specter, … + the 6–7 parser-rider cards: Whip of Erebos, Gruesome Encore, From the Catacombs, Isareth the Awakener, Kheru Lich Lord, Moira and Teshar, Dreams of the Dead). **No card gains or loses support** — the replacement already existed; it now carries the host-lifetime stamp and is made durable. The CI parse-diff artifact should report these 64 signature changes.

## Scope Expansion

None.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Unearth and similar effects that redirect a creature’s departure to exile.
  - Replacement effects now remain active only while hosted by the battlefield object.
  - Prevented replacement effects from incorrectly returning when an object leaves and re-enters play.
  - Ensured these effects are not copied to other objects and continue working after layer updates.
  - Added coverage for sacrifice, destruction, bouncing, controller changes, and parsed card effects.

### DIFF
--- a/crates/engine/src/database/unearth.rs
+++ b/crates/engine/src/database/unearth.rs
@@ -17,7 +17,10 @@
 //! 3. an `Effect::CreateDelayedTrigger` that exiles the card at the next end
 //!    step (CR 702.84a), and
 //! 4. an `Effect::AddTargetReplacement` installing the "if it would leave the
-//!    battlefield, exile it instead" replacement (CR 702.84a / CR 614.1a).
+//!    battlefield, exile it instead" replacement (CR 702.84a / CR 614.1a /
+//!    CR 400.7), stamped `RestrictionExpiry::UntilHostLeavesPlay` so it is
+//!    base-installed (surviving CR 613.1 layer reseeds), non-copiable
+//!    (CR 707.2), and pruned when the host leaves the battlefield.
 //!
 //! Steps 2–4 are chained as `sub_ability` continuation steps (CR 608.2c) of the
 //! primary `ChangeZone`, so they resolve as one action.
@@ -35,7 +38,7 @@
 
 use crate::types::ability::{
     AbilityCost, AbilityDefinition, AbilityKind, ContinuousModification, DelayedTriggerCondition,
-    Duration, Effect, ReplacementDefinition, StaticDefinition, TargetFilter,
+    Duration, Effect, ReplacementDefinition, RestrictionExpiry, StaticDefinition, TargetFilter,
 };
 use crate::types::card::CardFace;
 use crate::types::keywords::Keyword;
@@ -149,14 +152,19 @@ fn delayed_exile_step() -> AbilityDefinition {
     .description("Exile it at the beginning of the next end step.".to_string())
 }
 
-/// CR 702.84a / CR 614.1a: "If it would leave the battlefield, exile it instead
-/// of putting it anywhere else." Installs a `Moved` replacement on the returned
-/// permanent (`target: SelfRef`); `valid_card: SelfRef` binds the replacement to
-/// its own host so it fires only for that object, redirecting any
-/// battlefield-exit to exile.
+/// CR 702.84a / CR 614.1a / CR 400.7: "If it would leave the battlefield, exile
+/// it instead of putting it anywhere else." Installs a `Moved` replacement on
+/// the returned permanent (`target: SelfRef`); `valid_card: SelfRef` binds the
+/// replacement to its own host so it fires only for that object, redirecting any
+/// battlefield-exit to exile. The rider is stamped
+/// `RestrictionExpiry::UntilHostLeavesPlay`: it is a runtime effect bound to the
+/// host object's lifetime, so it is base-installed to survive CR 613.1 layer
+/// reseeds, kept non-copiable (CR 707.2), and pruned on the host's battlefield
+/// exit (CR 400.7) — see the variant doc in `types/ability.rs`.
 fn leaves_battlefield_exile_step() -> AbilityDefinition {
     let replacement = ReplacementDefinition::new(ReplacementEvent::Moved)
         .valid_card(TargetFilter::SelfRef)
+        .expiry(RestrictionExpiry::UntilHostLeavesPlay)
         .execute(AbilityDefinition::new(
             AbilityKind::Spell,
             exile_self_from_battlefield_effect(),
--- a/crates/engine/src/database/unearth_tests.rs
+++ b/crates/engine/src/database/unearth_tests.rs
@@ -14,7 +14,7 @@ use crate::game::triggers::check_delayed_triggers;
 use crate::game::zones::create_object;
 use crate::types::ability::{
     AbilityCost, AbilityDefinition, ContinuousModification, DelayedTriggerCondition, Effect,
-    TargetFilter,
+    RestrictionExpiry, TargetFilter,
 };
 use crate::types::card::CardFace;
 use crate::types::card_type::CoreType;
@@ -360,6 +360,41 @@ fn unearth_installs_leaves_battlefield_exile_replacement() {
         ),
         "the replacement must redirect the card to exile, got {:?}",
         execute.effect
+    );
+    // CR 400.7 + CR 702.84a: the rider is stamped `UntilHostLeavesPlay` so its
+    // lifetime is bound to the returned object.
+    assert_eq!(
+        repl.expiry,
+        Some(RestrictionExpiry::UntilHostLeavesPlay),
+        "the leaves-battlefield rider must carry the host-lifetime expiry stamp"
+    );
+
+    // CR 613.1: the rider is also written to `base_replacement_definitions` so it
+    // survives a layer reseed (which rebuilds live defs from base). Without the
+    // base copy, the first `evaluate_layers` would silently drop the redirect.
+    let base = &state.objects[&source].base_replacement_definitions;
+    let base_repl = base
+        .iter()
+        .find(|r| r.event == ReplacementEvent::Moved)
+        .expect("the rider must be base-installed to survive a layer reseed");
+    assert_eq!(
+        base_repl.expiry,
+        Some(RestrictionExpiry::UntilHostLeavesPlay),
+        "the base-installed rider must carry the host-lifetime expiry stamp"
+    );
+    assert_eq!(base_repl.valid_card, Some(TargetFilter::SelfRef));
+
+    // Discriminating runtime check: after a real layer pass, the live rider must
+    // still be present (repopulated from base). This is the exact reseed that
+    // dropped the live-only rider before the fix.
+    crate::game::layers::evaluate_layers(&mut state);
+    assert!(
+        state.objects[&source]
+            .replacement_definitions
+            .iter_all()
+            .any(|r| r.event == ReplacementEvent::Moved
+                && r.expiry == Some(RestrictionExpiry::UntilHostLeavesPlay)),
+        "the rider must survive a layer reseed (repopulated from base)"
     );
 }
 
--- a/crates/engine/src/game/effects/add_target_replacement.rs
+++ b/crates/engine/src/game/effects/add_target_replacement.rs
@@ -262,6 +262,14 @@ pub fn resolve(
                     // reset: a damaged creature can gain/lose characteristics
                     // or enter combat before it dies. Cleanup prunes this
                     // narrowly scoped base copy at end of turn.
+                    // A host-lifetime rider (CR 702.84a "if it would leave the
+                    // battlefield, exile it instead", stamped
+                    // `UntilHostLeavesPlay`) is the same class: it must survive
+                    // every CR 613.1 reseed so the redirect still fires after the
+                    // returned permanent gains/loses characteristics, and its
+                    // base+live copies are pruned together the instant the host
+                    // leaves the battlefield (`prune_controller_controls_source_on_leave`,
+                    // CR 400.7) so it never revives on a same-ObjectId re-entry.
                     //
                     // Acknowledged out-of-scope edges (NOT fixed here): (1) Cleave
                     // re-baselining only touches spells on the stack (casting.rs)
@@ -275,7 +283,12 @@ pub fn resolve(
                         crate::game::printed_cards::is_runtime_target_die_exile_replacement(
                             &replacement,
                         );
+                    let host_lifetime =
+                        crate::game::printed_cards::is_runtime_host_lifetime_replacement(
+                            &replacement,
+                        );
                     let install_to_base = durable_die_exile
+                        || host_lifetime
                         || matches!(
                             replacement.condition,
                             Some(ReplacementCondition::ControllerControlsSource { .. })
--- a/crates/engine/src/game/layers.rs
+++ b/crates/engine/src/game/layers.rs
@@ -15,7 +15,7 @@ use crate::game::filter::{matches_target_filter, FilterContext};
 use crate::game::game_object::DisplaySource;
 use crate::game::printed_cards::{
     apply_copiable_values, ensure_keyword_triggers_for_copiable_values, intrinsic_copiable_values,
-    is_runtime_target_die_exile_replacement,
+    is_runtime_host_lifetime_replacement, is_runtime_target_die_exile_replacement,
 };
 use crate::game::quantity::{
     continuous_modification_dynamic_quantity, filter_uses_recipient, quantity_expr_uses_recipient,
@@ -694,9 +694,13 @@ pub(crate) fn prune_lapsed_controller_controls_source(state: &mut GameState) {
 /// CR 611.2b + CR 400.7: the captured source leaving play, OR the host leaving
 /// and re-entering as a new object (same storage ObjectId), ends the can't-untap
 /// continuous effect permanently — drop the gated def from base+live so it
-/// cannot revive on a same-ObjectId re-entry. Called from `zones.rs` on a
-/// battlefield exit, OUTSIDE `evaluate_layers`, so it marks layers full on change
-/// (mirroring `prune_host_left_effects`).
+/// cannot revive on a same-ObjectId re-entry. The same host-left arm also drops
+/// the turn-bound die-exile rider and the host-lifetime exile rider
+/// (`UntilHostLeavesPlay`, CR 702.84a): a base-installed rider must NOT revive
+/// when the object re-enters as a new CR 400.7 object reusing the same storage
+/// key. Called from `zones.rs` (`:481`) on a battlefield exit, OUTSIDE
+/// `evaluate_layers`, so it marks layers full on change (mirroring
+/// `prune_host_left_effects`).
 ///
 /// The predicate is purely id-based (`departed_id`), so no `&state` gate read is
 /// needed — each object is mutated in a single pass: case (b) the captured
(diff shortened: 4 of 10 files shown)


matthewevans (changes requested):
**Verdict: 1 blocker — request changes.**

## 🔴 Blocker

**[HIGH] Preserve the enclosing duration for generic leave-battlefield exile riders.** Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:1768-1796` accepts the generic rider and unconditionally stamps `UntilHostLeavesPlay`; `crates/engine/src/game/effects/add_target_replacement.rs:25-31` derives an ability duration only when expiry is absent. The parser itself routes a leading duration through `with_clause_duration` at `oracle_effect/mod.rs:8224-8225`. Scryfall Oracle text for Elemental Expressionist is: “Until end of turn, it gains "If this creature would leave the battlefield, exile it instead of putting it anywhere else" …”. CR 611.2a says a continuous effect lasts as long as its stated duration, and CR 514.2 ends all until-end-of-turn effects during cleanup. Why it matters: that temporary rider is retained in base state and continues to redirect a later zone change after cleanup, producing a wrong game result. Suggested fix: make the resolver honor an enclosing duration for this rider and use host-lifetime expiry only when no enclosing duration applies; add a runtime regression that keeps the Expressionist target through cleanup, then moves it.

michiot05:
Thanks — CR 611.2a / 514.2 are exactly the right lens, so I traced Elemental Expressionist's data flow end-to-end before touching the stamp, and I think the temporary-rider path doesn't actually reach this PR's code. Laying out what I found in case I've missed a path you're seeing:

**Expressionist never reaches the stamped replacement.** Its `"Until end of turn, it gains \"If this creature would leave the battlefield, exile it instead…\""` is a **quoted grant**, so it routes through `oracle_static::classify_quoted_inner` → `ContinuousModification::GrantAbility`, not through `try_parse_leave_battlefield_exile_replacement`. `oracle_classifier.rs:563-570` deliberately sends granted "would leave the battlefield" text to the static parser ahead of the replacement fallback. Its actual parse in `card-data.json` is a `Continuous` static (`duration: UntilEndOfTurn`) carrying `GrantAbility` + `GrantTrigger` — **no `AddTargetReplacement`, no `UntilHostLeavesPlay`** anywhere in the entry (it also still carries a `SwallowedClause`/`Replacement_Instead` warning). So it never hits `mod.rs:1796` and never flows through `replacement_with_ability_expiry`, and the `is_none()` guard you cited is never consulted for it.

**The stamped path is single-caller and only reanimation riders reach it.** `try_parse_leave_battlefield_exile_replacement` is private with one production caller (`mod.rs:7650`). Corpus-wide, 15 cards carry the rider text; the 7 that reach the stamp are all standalone reanimation riders (Whip of Erebos, Gruesome Encore, Kheru Lich Lord, Moira and Teshar, From the Catacombs, Isareth, Dreams of the Dead) — the "next end step" ones use a delayed trigger, not a duration on the rider. Of the **64** cards that carry an `AddTargetReplacement` + `UntilHostLeavesPlay` in `card-data.json`, **57 are `Duration::Permanent` and 7 are `None` — zero are temporary.** Since `expiry_from_duration` maps both `Permanent` and absent to `None`, the conditional design you describe ("host-lifetime only when no enclosing duration applies") derives host-lifetime for all 64 — behaviorally identical to the current stamp, and there's no card left to exercise the enclosing-duration branch.

**Expressionist is genuinely broken — but by a separate, pre-existing gap in the opposite direction.** Its granted rider parses to an *inert* `GrantAbility` (a bare `ChangeZone→Exile` spell body pushed into `obj.abilities` at `layers.rs:6595`, never into `replacement_definitions`), so today it redirects **nothing** — an under-implementation, not the post-cleanup over-retention the stamp would cause. Fixing it correctly is a *granted, duration-bound replacement ability* in the quoted-grant classifier (so the redirect fires within the turn and lapses at cleanup via the existing transient path) — a distinct change from this PR's host-lifetime stamp on standalone riders. I'd be glad to file that as its own issue and take it next.

Given that, adding the conditional-duration branch here would be a guard for a state no current card reaches, which I didn't want to slip in against the no-speculative-guards guidance — but I'm happy to add it as a forward-looking correctness measure if you'd prefer it in explicitly; it's a one-commit no-op for the 64. And if you're seeing Expressionist actually redirect to exile after cleanup in play, that contradicts what I'm finding (the grant looks inert), so please share the repro and I'll dig back in.

Head is unchanged at a65dd1a22 (Gate A + G PASS, clippy clean, the 8 integration rows + regression sweep green). Model: claude-opus-4-8

matthewevans:
**Verdict: blocker withdrawn — your pushback is correct, and I verified it independently. Two LOW nits remain, neither blocking.**

Thank you for the trace. I re-derived every step of your argument from the corpus rather than taking the numbers on trust, and it holds on all three legs. My changes-request was wrong and I'm retracting it.

## ✅ Clean — the duration-lifetime blocker does not survive

**1. Elemental Expressionist never reaches the stamped path — confirmed.** I pulled its Oracle text verbatim from Scryfall:

> `Magecraft — Whenever you cast or copy an instant or sorcery spell, choose target creature you control. Until end of turn, it gains "If this creature would leave the battlefield, exile it instead of putting it anywhere else" and "When this creature is put into exile, create a 4/4 blue and red Elemental creature token."`

and then read its actual entry in `card-data.json`:

```
has AddTargetReplacement: False
has UntilHostLeavesPlay:  False
has GrantAbility: True | GrantTrigger: True
has SwallowedClause: True | Replacement_Instead: True
duration values: Counter({'UntilEndOfTurn': 2})
```

No `AddTargetReplacement`, no `UntilHostLeavesPlay`. It is a `Continuous` static carrying the quoted grant, exactly as you described. The card I named as the exemplar of the harmed class is not in the affected population at all.

**2. The affected population contains zero temporary durations — confirmed, and it reproduces your split exactly.** I filtered `card-data.json` for the stamped shape (`event: Moved` + `valid_card: SelfRef` + `execute` → `ChangeZone{origin: Battlefield, destination: Exile, target: SelfRef}`), restricted to riders whose baseline `expiry` is `∅` (the only ones this PR changes), and grouped by nearest enclosing duration:

```
baseline expiry==NULL rider cards (the ∅ population the PR stamps): 64

  57  ('Permanent', 'nocond')
   7  ('None',      'nocond')

--- any ∅-rider with a TEMPORARY enclosing duration? ---
count: 0
```

64, matching the parse-diff's `64 card(s), 1 signature(s)` precisely — 57 `Permanent` (the Unearth keyword cards) plus 7 no-duration (the standalone reanimation riders). Since `expiry_from_duration` maps both `Permanent` and absent to `None`, a conditional "host-lifetime only when no enclosing duration applies" design would derive host-lifetime for all 64, behaviourally identical to the flat stamp. There is no card left to exercise the branch I asked for. Adding it would be a guard for an unreachable state, and you were right to push back on that per the no-speculative-guards guidance.

**3. The 64-card expiry delta is fully explained — this is what I misread.** There *are* 7 cards carrying a `Moved`→Exile SelfRef rider under an enclosing `UntilEndOfTurn` (Bleed Dry, Burn from Within, Carbonize, Disintegrate, Necrotic Wound, Ob Nixilis's Cruelty, Smite the Deathless). Every one of them already carries `expiry: EndOfTurn` in the baseline from the die-exile stamp:

```
--- already-stamped (die-exile) riders, untouched by PR ---
  46  ("{'type': 'EndOfTurn'}", 'None')
   7  ("{'type': 'EndOfTurn'}", 'UntilEndOfTurn')
```

They are not in the `∅` set, so this PR does not touch them. I had read the 64-card delta as evidence that temporary-duration riders were being retagged; the delta is entirely Unearth plus reanimation riders.

**4. Single-caller claim — confirmed.** `try_parse_leave_battlefield_exile_replacement` is private at `crates/engine/src/parser/oracle_effect/mod.rs:1768` with exactly one production caller at `:7642` (your `:7650` was line drift, immaterial) and no other reference workspace-wide.

**5. Seam and blast radius.** The change sits on existing authorities and reuses their vocabulary: the `is_runtime_*_replacement` predicate family in `printed_cards.rs`, the `durable_die_exile` base-install precedent at `add_target_replacement.rs:283`, and the existing host-left arm of `prune_controller_controls_source_on_leave` at `layers.rs:724`. No new parallel machinery.

**6. The variant is genuinely inert to turn machinery — verified, not just asserted.** Your `types/ability.rs` doc claims `UntilHostLeavesPlay` is "NEVER pruned by turn machinery". I checked every prune site rather than trusting the comment, because a pre-existing wildcard arm would silently absorb a new variant and green CI would never show it. All of them use positive `matches!` on specific variants — `turns.rs:1869` (`expires_at_eot`), `:1957`, `:211`/`:215`, `:1221` — so nothing sweeps the new variant in. The claim holds.

**7. Serialized surface is intact.** `client/src/adapter/types.ts:849` mirrors `RestrictionExpiry` as a hand-maintained TS union that lacks `UntilHostLeavesPlay`, which I checked as a possible drift point. It is used only by `GameRestriction` (`:864`, `:873`, `:881`), and `ReplacementDefinition` is not exposed to TypeScript at all. Since the new variant is only ever set on `ReplacementDefinition.expiry`, the union stays complete. No client change needed.

**8. CR citations.** I grep-verified all eight against `docs/MagicCompRules.txt` — 702.84a, 614.1a, 400.7, 613.1, 707.2, 608.2c, 611.2b, 701.21a. All resolve and all describe the annotated code.

**9. Runtime tests are discriminating.** The integration rows drive the real pipeline (`activate` → `resolve` → a genuine `evaluate_layers` pass → real sacrifice/destroy/bounce), and R1 carries a `PermanentSacrificed` reach guard so the exile assertion cannot pass because the cost silently failed.

## 🟡 Non-blocking

**[LOW] Under-scoped CR citation on one test comment.** `crates/engine/src/parser/oracle_effect/tests.rs:39299-39301`:

```rust
// CR 400.7 (issue #5976): the parsed rider is stamped with the
// host-lifetime expiry so it is base-installed, non-copiable, and
// pruned on host exit.
```

CR 400.7 covers only the last clause (no relation to previous existence → no revival on same-`ObjectId` re-entry). Base-install-survives-reseed is CR 613.1 and non-copiable is CR 707.2. Every other comment you added carries the full triad; this one line is the outlier. Suggested fix: `// CR 613.1 + CR 707.2 + CR 400.7 (issue #5976): …`.

For the record, CodeRabbit flagged this same line but its proposed correction is wrong — it asks you to replace CR 400.7 with CR 702.84a and treat the prune behaviour as "engine semantics, not rule citation". CR 400.7 is precisely the right rule for the no-revival claim, as its own web-search output confirms. Widen the citation rather than swapping it.

**[LOW] Mixed abstraction on a shared enum.** `crates/engine/src/types/ability.rs:2605` — your own doc notes `UntilHostLeavesPlay` "is meaningless for the state-hosted `GameRestriction` / `pending_damage_replacements` that share this enum". That is an object-lifetime anchor living on an enum whose other four variants are turn/phase-structure anchors. The die-exile stamp set this precedent so I am not asking you to change it here, and documenting the split explicitly is the right call. Noting it so the constraint is discoverable if this enum is ever split.

## Follow-up

Your read on Expressionist being broken in the opposite direction checks out at the seam you cited: `layers.rs:6590-6598` handles `ContinuousModification::GrantAbility` by pushing the concretized body into `obj.abilities`, never into `replacement_definitions`. A granted replacement-shaped ability therefore sits inert, and its entry still carries a `SwallowedClause`/`Replacement_Instead` warning. **Please do file that as its own issue and take it** — a granted, duration-bound replacement ability in the quoted-grant classifier is the correct fix and is properly out of scope here.

**Recommendation: land as-is. The blocker is withdrawn — I am no longer holding this PR, and the outstanding changes-requested state should be cleared on that basis. The two LOW items are optional and need not gate it; fold the CR-citation widening in only if you are pushing for another reason.** Apologies for the round-trip — the trace you supplied was specific enough to falsify my claim in one query, which is exactly what made it cheap to check.

matthewevans (approved):
Approving — the earlier changes-request on this head is **withdrawn** and this supersedes it.

The blocker claimed that a generic leave-battlefield exile rider would unconditionally override an enclosing until-end-of-turn duration. Re-derived from `card-data.json`, that claim is false on every leg:

- **Elemental Expressionist is not in the affected population.** It carries no `AddTargetReplacement` and no `UntilHostLeavesPlay`, so it never reaches the stamped path. The exemplar the blocker was built on does not exist in the class.
- **The 64-card ∅-rider population contains zero temporary enclosing durations** (57 `Permanent` + 7 `None`). Since `expiry_from_duration` maps both to `None`, the conditional design I asked for would derive host-lifetime for all 64 — behaviourally identical to the flat stamp, i.e. a guard for an unreachable state.
- **The 7 `UntilEndOfTurn` rider cards already carry `expiry: EndOfTurn`** in the baseline from the die-exile stamp, so they are outside the ∅ set and untouched by this PR.

Seam, blast radius, prune-site inertness, TS surface, CR citations, and runtime-test discrimination were all independently confirmed. Thank you for the trace — it was specific enough to falsify a maintainer blocker in a single query, which is exactly what made it cheap to check. Apologies for the round-trip.

Two **non-blocking LOW** items, deliberately not gating this merge and not worth a push on their own:

1. The CR citation on the `tests.rs` comment would be better **widened** to `// CR 613.1 + CR 707.2 + CR 400.7 (issue #5976)`. For the record, CodeRabbit's proposal on that same line is wrong — it wants CR 400.7 *replaced* with CR 702.84a; CR 400.7 is precisely the right rule for the no-revival claim. Widen, do not swap.
2. The documented mixed-abstraction note on `types/ability.rs:2605` — precedent was set by the die-exile stamp; no change requested.

The Expressionist issue you identified in the opposite direction (granted replacement-shaped abilities landing in `obj.abilities` rather than `replacement_definitions`) is real and correctly out of scope here — please file and take it.

Landing as-is.

michiot05:
Appreciate you re-deriving it from the corpus rather than taking my numbers on trust — and no need to apologize, a specific blocker is what made it cheap to falsify in the first place. Landing as-is per your note; I'll leave the two LOW items alone unless the PR needs another push for some other reason, in which case I'll widen that `tests.rs` citation to `CR 613.1 + CR 707.2 + CR 400.7` (agreed on widen-not-swap; 400.7 is the right rule for the no-revival clause).

And yes — I'll file the granted-replacement gap as its own issue and take it: quoted `it gains "if it would leave the battlefield, exile it instead…"` grants currently land in `obj.abilities` via `ContinuousModification::GrantAbility` (`layers.rs:6590-6598`) and never become a replacement, so the redirect sits inert (with the `SwallowedClause`/`Replacement_Instead` warning) across the 7 quoted-grant cards. The fix is a granted, duration-bound replacement ability in the quoted-grant classifier, honoring the grant's `UntilEndOfTurn` via the existing transient path — cleanly separate from this PR's host-lifetime stamp on standalone riders.

Model: claude-opus-4-8

PR #6544, opened 2026-07-23
feat(phase-ai): add graveyard type-diversity axis + GraveyardTypesPolicy

minion1227:
## Summary
Adds a **graveyard card-type-diversity axis** (`GraveyardTypesFeature`) and its companion `GraveyardTypesPolicy` — the delirium / descend / threshold / Goyf axis.

CR 207.2c lists delirium, descend and threshold as **ability words** with no rules meaning, so the mechanical content is entirely the underlying "N or more card types among cards in your graveyard" condition (CR 205.2a). **95 cards** in the corpus read that quantity, and nothing in the AI modelled the graveyard's type spread: a self-mill that supplied the fourth distinct card type — switching every delirium payoff on — scored exactly the same as one that added a redundant fifth creature.

**Three structural axes, no card-name matching:**

| Axis | AST surface |
|---|---|
| threshold payoffs | `QuantityComparison` whose lhs is `DistinctCardTypes { Zone { Graveyard, Controller } }` |
| scaling payoffs | the same quantity as a dynamic magnitude, no threshold (Consuming Blob, Tarmogoyf) |
| enablers | `Mill` / `DiscardCard` / `Discard` / `Surveil` scoped to the controller |

Two details worth flagging for review:

- `StaticDefinition.condition` and `TriggerDefinition.condition` are **distinct enums** that happen to share the `QuantityComparison` shape. Each gets its own extractor rather than a forced conversion — Backwoods Survivalists puts the delirium clause on the static, Autumnal Gloom puts it on the trigger.
- A threshold payoff is deliberately **not** also counted as a scaling payoff, or delirium cards would be double-weighted in the commitment formula. There is a test pinning this.

`highest_threshold` is tracked so a descend 8 deck does not think it is finished at four card types. Opponent-scoped counts and opponent-scoped mill are both rejected — punishing an opponent's diverse graveyard, or filling it, does nothing for this deck's own threshold.

**Commitment is a geometric mean over (payoff, enabler)** — unlike the poison axis, both pillars are mandatory. Payoffs with no enablers never turn on reliably; enablers with no payoff are just self-mill. Neither is this archetype, and both collapse to `0.0` (two tests pin the collapse). Calibrated to a Modern delirium shell (8 threshold + 2 scaling + 8 enablers over 37 nonland) > 0.85, with a lone-Tarmogoyf anti-calibration below the floor.

**Policy** scores by threshold *deficit*: strong band when the action supplies the **last missing type**, scaled down as the deficit grows. Once the count is at or above the deck's highest threshold the branch scores **zero** — every payoff is already live, so more diversity buys nothing, and scoring it would make the AI durdle with self-mill after delirium is already on.

## Files changed
- `crates/phase-ai/src/features/graveyard_types.rs` (new)
- `crates/phase-ai/src/features/tests/graveyard_types.rs` (new)
- `crates/phase-ai/src/policies/graveyard_types.rs` (new)
- `crates/phase-ai/src/policies/tests/graveyard_types.rs` (new)
- `crates/phase-ai/src/features/mod.rs`, `policies/mod.rs`, `policies/registry.rs`, `policies/tests/mod.rs`, `features/tests/mod.rs`, `config.rs`

## CR references
- `CR 207.2c` — delirium / descend / threshold / undergrowth are ability words with no rules meaning
- `CR 205.2a` — the card types
- `CR 404.1` — a graveyard is its owner's discard pile
- `CR 109.2a` + `CR 400.1` — zone scoping for the count
- `CR 109.3` — conjunction rule for the `And` arms of the condition and filter walks
- `CR 701.17` Mill · `CR 701.25` Surveil

Every number grep-verified against `docs/MagicCompRules.txt` before it was written.

## Performance
`verdict()` runs per candidate per search node, so predicate order is deliberate: the card-local AST check (`fills_own_graveyard_parts` over just this candidate's abilities) runs **first** and rejects almost every candidate. Only a confirmed graveyard-filler pays for the graveyard scan, which walks one zone's objects and never touches the battlefield, mana affordability, `find_legal_targets`, or any `SimulationFilter`-cloning path. `activation()` opts the whole policy out below the commitment floor.

## Measurement — `cargo ai-gate` with a paired control

```
| matchup             | baseline p0% | current p0% | flips W→L | flips L→W | sign p | status |
| affinity-mirror     | 20%          | 40%         | 0         | 0         | —      | WARN   |
| enchantress-mirror  | 50%          | 40%         | 3         | 2         | 0.3438 | WARN   |
| red-mirror          | 70%          | 30%         | 5         | 1         | 0.0625 | WARN   |
compare: 0 FAIL, 3 WARN, 0 PASS, 0 NEW, 0 REMOVED
```

**Those WARNs are pre-existing and not caused by this PR.** I ran the identical gate on clean `upstream/main` (ebd884448 — this branch's exact base) as a control, and it produces a **byte-identical** report: same three WARNs, same flip counts, same p-values. `crates/phase-ai/baselines/suite-baseline.json` was last refreshed 2026-06-17 (#3492), roughly 200 upstream commits ago.

Consistent with that, none of the three gate matchups is a graveyard-diversity deck, so `GraveyardTypesPolicy::activation` returns `None` and the policy never fires on them — the control proves it empirically rather than by argument.

**I deliberately did not refresh the baseline.** Absorbing pre-existing divergence into a baseline as part of an unrelated feature PR would hide upstream drift. Flagging it instead: the suite baseline looks due for a refresh on its own commit.

## Verification
- `cargo test -p phase-ai --lib` — **1381 passed; 0 failed** (16 feature tests + 3 policy tests new)
- `cargo clippy -p phase-ai --all-targets -- -D warnings` — clean
- `cargo fmt --all` — clean
- `cargo ai-gate` — 0 FAIL, with the paired control above

## Notes
- `graveyard_types_progress` is registered in `UNTUNED_POLICY_PENALTY_FIELDS` pending a paired-seed calibration.
- Tests live in `features/tests/` and `policies/tests/` per the house convention (no `#[cfg(test)]` in source files).
- **Boundary with `reanimator`:** that axis detects graveyard *recursion targets* — what is worth bringing back. This one measures the graveyard's *type spread*. A graveyard of four creatures is excellent for reanimator and useless for delirium; the axes are independent by design.
- **Known limitation:** fetchland-style land sacrifice and cycling are not counted as enablers. Both do add a card type in practice, but neither is a direct self-mill/discard effect, so folding them in would need a separate structural signal — a reasonable follow-up rather than a silent widening here.

Tier: Frontier

## Summary by CodeRabbit

* **New Features**
  * Added graveyard-type feature detection (threshold payoffs, scaling-only payoffs, and self-graveyard enablers) and surfaced it in deck analysis.
  * Introduced a Graveyard Types tactical AI policy that scores actions by distinct graveyard core types, gated by a commitment floor.
* **Bug Fixes**
  * Improved recognition of actions that truly fill the player’s own graveyard, with broader targeting/scoping support.
* **Config Updates**
  * Added the `graveyard_types_progress` calibration weight for threshold progress scoring (backward-compatible deserialization).
* **Tests**
  * Added unit tests covering feature detection, policy scoring, authoritative graveyard-fill semantics, and legacy config loading.


matthewevans (changes requested):
Thanks for the focused implementation. I need a revision before we can merge this current head.

1. `PolicyPenalties` is deserialized as a complete value from existing tuning artifacts (`ai_tune.rs`), but the new `graveyard_types_progress` field has no serde default. Every artifact produced before this PR therefore fails to load with a missing-field error instead of retaining the shipped default. Please add the same field-level default pattern used by the neighboring recently-added penalties, and cover deserializing an artifact that predates this field.

2. The scaling-only case is given `DEFAULT_DELIRIUM_THRESHOLD` in the feature even though it has no threshold. The policy then returns neutral as soon as the graveyard has four types, so it stops rewarding progress for a payoff that continues to scale. Model the absence of a threshold distinctly and add a verdict-level test that a scaling-only deck still receives the intended progress signal above four types.

3. Threshold extraction ignores `Comparator`, treats `Not` as positive, and takes the maximum threshold under both `And` and `Or`. This misclassifies conditions such as “fewer than four types” or negated comparisons as delirium payoffs; `Or` also does not make every child a mandatory gate. Please normalize only genuinely positive mandatory comparisons (including the correct strict-`>` boundary) and reject/infer the rest according to the condition's truth semantics. Tests need to discriminate `GE`, `GT`, negative comparators, negation, and compound conditions.

4. `quantity_reads_own_graveyard_types` accepts `CountScope::All`, while `distinct_graveyard_types` deliberately scans only the AI player's owned graveyard. Consequently an all-graveyards payoff is detected as an own-graveyard plan but the policy computes a different quantity (e.g. the parsed all-graveyards Tarmogoyf-class shape). Either keep the feature/policy strictly controller-scoped or carry the source scope through and resolve it consistently at decision time.

5. The candidate lookup does not use the action's authoritative ability semantics. On `CastSpell` it walks every ability on the object, so casting a permanent with an activated self-mill/discard ability is rewarded even when the spell itself does not fill a graveyard. On `ActivateAbility` it indexes `GameObject::abilities` directly even though `cast_facts::effective_activated_ability` documents that this is not the engine's authoritative enumerated index space for runtime-granted abilities. Please use the established cast/activated-ability fact helpers (and the spell-only cast effect chain) and add regression coverage for both decisions.

I agree with the substantive current CodeRabbit reports on the threshold, scope, and cast-action issues and am leaving those threads intact. Its wildcard-exhaustiveness observation is a non-blocking style concern; the semantic fixes above are required first.

minion1227:
Addressed all five blocking items in `ec075b9d5`.

**1 — Preserve pre-field `PolicyPenalties` artifacts.** `graveyard_types_progress` now carries `#[serde(default = "default_graveyard_types_progress")]`, with `Default` and the serde attribute calling that one shared fn (matching the neighbouring recently-added penalties). Added `policy_penalties_load_pre_graveyard_types_artifact`: it strips the key from a serialized artifact, tunes a neighbour, and asserts the tuned value survives while the absent field falls back to the shared default — the exact `ai_tune.rs` `TuneGroup::Penalties` path.

**2 — Model the absence of a threshold distinctly.** `highest_threshold` is now `Option<u32>`; the fabricated `DEFAULT_DELIRIUM_THRESHOLD` fallback and the const itself are gone. The policy branches accordingly: a threshold deck races to its gate; **any deck with a scaling payoff keeps earning a diminishing-but-nonzero signal past it**, so a Consuming Blob / Tarmogoyf deck is no longer truncated at four types. Covered by `verdict_still_rewards_scaling_only_deck_above_four_types` (the test you asked for), plus `verdict_mixed_deck_keeps_scaling_past_threshold` and `scaling_only_deck_has_no_threshold`.

**3 — Respect comparator direction and negation.** Extraction now normalizes only genuinely positive mandatory gates:
- `>=` / `>` accepted, with the strict-bound off-by-one (`> 3` ⟺ `>= 4`), in **either orientation** (`N <= types` flips to `types >= N`).
- `<`, `<=`, `=`, `!=` rejected — they reward fewer or an exact count, so self-mill toward N is not what they want.
- `Not` no longer passes through as positive: negating "N or more types" is a "fewer than N" condition.
- `And` takes the max mandatory threshold. `Or` yields one **only when every branch is a graveyard threshold** (then the minimum) — a single non-graveyard branch means the payoff can fire without delirium, so it is not a mandatory gate.

Tests discriminate `GE`, `GT` boundary, each negative comparator, negation, mirrored orientation, and both compounds.

**4 — Keep the axis strictly controller-scoped.** `quantity_reads_own_graveyard_types` no longer accepts `CountScope::All`. Since `distinct_graveyard_types` counts only the AI's owned graveyard, an all-graveyards payoff would otherwise be detected as an own-graveyard plan while the policy measured a different quantity (exactly the Tarmogoyf-class mismatch you flagged). The scope match is now explicit — `Controller`/`Owner` per CR 108.3 + CR 109.5 — with `all_graveyards_scope_is_not_own_graveyard` pinning it.

**5 — Use the action's authoritative ability semantics.** `CastSpell` now inspects only the spell's own resolution chain via `CastFacts::primary_effects`, so casting a permanent that merely *has* an activated self-mill ability is no longer credited. `ActivateAbility` resolves through `cast_facts::effective_activated_ability` — the engine's enumerated index space, correct for runtime-granted abilities where `GameObject::abilities` is not. Regression coverage for both: `verdict_ignores_cast_of_a_body_with_an_activated_mill` (fails if the cast path regresses to walking every ability), `verdict_credits_the_activated_mill_ability`, and `verdict_credits_a_spell_that_mills_on_resolution`.

**Also:** the verdict now routes through `PolicyVerdict::score` rather than the band-specific helpers, so a tuned-up scalar auto-bands instead of tripping `score_in_band`'s assertion — the same latent-fragility fix that surfaced on #6543.

**Verification**
- `cargo test -p phase-ai --lib` — 1401 passed (graveyard suite 12 → 39 tests, adding the whole verdict layer against a real `PolicyContext`).
- `cargo clippy -p phase-ai -p engine --all-targets --features proptest -- -D warnings` — clean.

On CodeRabbit's wildcard-exhaustiveness note: I left the large `StaticCondition` / `TriggerCondition` matches on `_ => None` as you indicated it is non-blocking style, but made the `CountScope` arm explicit since item 4 required reasoning about it anyway.

matthewevans:
**All five prior blockers are genuinely fixed — I re-verified each independently rather than taking the summary. Two new blockers remain, and they're about *reach*, not correctness: as built, the axis switches itself off for the flagship delirium shell.**

## ✅ The five prior blockers — verified fixed

| # | Item | Evidence |
|---|---|---|
| 1 | `PolicyPenalties` deserialize default | `config.rs:469-470` per-field `#[serde(default = "default_graveyard_types_progress")]`, shared fn at `:543-545`, and `Default` calls that same fn at `:535` — so the default can't drift between the two paths. Registered in `UNTUNED_POLICY_PENALTY_FIELDS` `:740-744`. |
| 2 | Invented four-type threshold | Gone. `highest_threshold: Option<u32>` at `features/graveyard_types.rs:70`; `DEFAULT_DELIRIUM_THRESHOLD` greps clean across all ten files. |
| 3 | `Comparator` semantics | `positive_graveyard_threshold` `:260-287` — GE takes the bound, GT takes bound+1, `LT`/`LE`/`EQ`/`NE` yield `None`, **exhaustive match with no `_` arm**, plus `flip_comparator` `:299-308` for the mirrored orientation. |
| 4 | `Not` / `Or` composition | `Not` → `None` (`:216`, `:237`); `Or` yields a threshold only when *every* branch has one, then takes the min (`:246-250`, with empty-`Or` correctly collapsing to `None`); `And` takes the max (`:203-206`, `:230-233`). |
| 5 | `CountScope::All` | `quantity_reads_own_graveyard_types` `:319-331` matches `Controller | Owner` only. |

The serde fix is the real thing rather than a half-fix: `policy_penalties_load_pre_graveyard_types_artifact` (`config.rs:1520-1544`) removes the key, tunes a *neighbouring* field to `-3.5`, and asserts both that the neighbour survives and that the absent field lands on the shared default. And `PolicyVerdict::score` is a band helper, not a raw sentinel — `registry.rs:195-208` dispatches by magnitude into `nudge`/`preference`/`strong`/`critical` and clamps at `CRITICAL_MAX`. No violation there.

## 🔴 Blocker 1 — enabler detection never reads `face.triggers`, which zeroes the whole axis

`features/graveyard_types.rs:131` calls `fills_own_graveyard_parts(&face.abilities)` and nothing else.

The consequence is not "some cards score slightly low" — it's total inertness, and the mechanism is worth spelling out because it's non-obvious:

- `compute_commitment` ends at `:183` with `commitment::geometric_mean(&[payoff_density, enabler_density])`.
- A geometric mean is zero if **any** factor is zero. So `enabler_count == 0` ⇒ `enabler_density == 0` ⇒ `commitment == 0.0`.
- `GraveyardTypesPolicy::activation` (`policies/graveyard_types.rs:81-86`) returns `None` below `GRAVEYARD_TYPES_FLOOR = 0.35`.

So a deck whose enablers are all trigger-borne scores nothing at all. **Stitcher's Supplier — the archetypal delirium enabler — is literally `abilities: []` with two Mill triggers** (verified in `card-data.json`). Same shape for Aberrant Researcher, Acolyte of Affliction, Aftermath Analyst.

Scale: walking `abilities` vs `triggers` over `card-data.json`, I count **237 faces** carrying a controller-scoped `Effect::Mill` in `triggers` and none in `abilities`. (Method: recursive effect-node walk over each face's `abilities` and `triggers`, `target ∈ {Controller, You}`, `a-` Alchemy variants included. Treat it as the right order of magnitude rather than an exact figure — a stricter `triggers[*].execute`-only walk yields a somewhat lower number.)

The house pattern already exists: `features/aristocrats.rs:313` and `features/tokens_wide.rs:120-124` walk `face.abilities` **and** `face.triggers[*].execute` together. Fix is to mirror `tokens_wide::is_token_generator_parts`.

## 🔴 Blocker 2 — threshold extraction misses the third condition carrier, and the obvious fix still misses it

`features/graveyard_types.rs:101-112` reads `static_abilities[*].condition` and `triggers[*].condition`. There is a third carrier: `AbilityDefinition.condition: Option<AbilityCondition>` (`engine/src/types/ability.rs:16265`), which holds the same `QuantityCheck` shape via `AbilityCondition::ConditionInstead { inner }`.

**The important detail — and this is a correction to make before you code it:** on Traverse the Ulvenwald the gate is **not** at `abilities[0].condition`. It's at `abilities[0].sub_ability.condition`:

```json
/abilities[0]/sub_ability/condition
  {"type":"ConditionInstead","inner":{"type":"QuantityCheck",
   "lhs":{"type":"Ref","qty":{"type":"DistinctCardTypes",
          "source":{"type":"Zone","zone":"Graveyard","scope":"Controller"}}},
   "comparator":"GE","rhs":{"type":"Fixed","value":4}}}
```

That falls out of the card's own grammar — *"Delirium — If there are four or more card types among cards in your graveyard, **instead** search…"* is a replacement of the *second* clause, so it lowers onto the sub-ability. A fix that folds only `face.abilities[*].condition` into the `.max()` chain would still return `None` for Traverse. **The extractor has to walk the `sub_ability` / `else_ability` chain.**

Scale: of **67 faces** reading `DistinctCardTypes` over a `Controller`/`Owner` graveyard anywhere, **13 carry it only in the `abilities` chain** — Traverse the Ulvenwald, Descend upon the Sinful, Invasive Surgery, Pick the Brain, Might Beyond Reason, To the Slaughter, Demonic Counsel, Beastie Beatdown, Convert to Slime, Impossible Inferno and others. (Method: `static_abilities` / `triggers` / `abilities`-with-chain-walk, same file.)

**These two blockers are mirror images of each other**, which is what makes me confident it's a systematic omission rather than two unrelated misses: enabler detection reads `abilities` but not `triggers`; threshold detection reads `triggers` but not `abilities`. Worth fixing as one change that gives both detectors the same three-carrier walk. Note the doc comment at `:51` already claims detection is structural over "`static_abilities`, `.triggers` and `.abilities`" — the intent was right, only one detector got each half.

## 🟡 Non-blocking

**`Effect::Dig { rest_destination: Graveyard }` isn't recognized** (`:390-403` matches `Mill`/`DiscardCard`/`Discard`/`Surveil` only). This is arguably the *best* shape on this axis — "reveal N, keep one, rest to graveyard" deposits several distinct types at once. I count **76 faces** using it; verified both spellings of the pattern: Satyr Wayfinder carries it in `triggers` (`abilities: []`), Grisly Salvage in `abilities`. So this one needs the three-carrier walk from Blocker 1 *and* a new effect arm.

**The `CastSpell` gate discards `CastFacts::immediate_etb_triggers`** (`policies/graveyard_types.rs:161-163` uses `facts.primary_effects` only). Excluding *activated* abilities was correct; this overshoots to also exclude ETB triggers, which do fire as a consequence of the cast — that's why `CastFacts` carries them as a separate field (`cast_facts.rs:230-235`). Net effect: casting Stitcher's Supplier or Satyr Wayfinder, i.e. the actual delirium play, returns `graveyard_types_na`. The doc at `:152-156` cites CR 601.2, which I grep-verified at `docs/MagicCompRules.txt:2455` — it supports excluding activated abilities, not ETB triggers.

**Second authority for "does this fill my own graveyard".** `features/reanimator.rs:208-260` already defines `effect_fills_own_graveyard` and documents itself verbatim as *"Single authority — … Shared with the live policy."* This PR re-implements it at `features/graveyard_types.rs:384-417` with a **different answer set**: reanimator accepts `TargetFilter::Any`, this one doesn't; this one accepts `SelfRef` and `Typed { controller: Some(You) }`, reanimator doesn't. Two single authorities disagreeing about one question is precisely the drift the single-authority rule exists to prevent. Extend the reanimator helper and have both call it.

**Docs contradict the item-4 fix.** `features/graveyard_types.rs:59`, `:67` and `policies/graveyard_types.rs:19`, `:130` all name **Tarmogoyf** as a detected scaling payoff. Tarmogoyf counts types *"among cards in all graveyards"* → `CountScope::All`, which `:319-331` now correctly rejects, so it's never detected. Consuming Blob is the right example throughout.

**`policies/graveyard_types.rs:31-32`** claims the scan "walks one zone's objects and never touches the battlefield"; `distinct_graveyard_types` `:53-63` iterates `state.objects.values()` and filters. `GameState` has no zone index, so the implementation is the only option and matches house practice — the comment is just wrong.

**Dead branch** at `policies/graveyard_types.rs:114-118`: `if deficit == 1 { scalar } else { scalar / f64::from(deficit) }` — both arms are identical when `deficit == 1`.

**Fixture path-divergence is why all four gaps ship green.** Every enabler fixture in both test files is an ability-borne `Effect::Mill { target: Controller }` (`features/tests/graveyard_types.rs:95-106`, `policies/tests/graveyard_types.rs:90-97`). No fixture is trigger-borne, `Dig`-borne, or carries a threshold in the `abilities` chain. When you fix the detectors, please add at least one of each — otherwise the next regression here is invisible again.

## 📋 Not your defect — two notes

**The two `CR 109.3` annotations** (`features/graveyard_types.rs:201`, `:413`) don't describe the code. I grep-verified `docs/MagicCompRules.txt:592`: CR 109.3 is the list of an object's *characteristics* (name, mana cost, color, …) and says nothing about conjunction. But this is **inherited verbatim from `features/landfall.rs:126` and `features/enchantments.rs:162`**, and your PR says it mirrors landfall — so it's ours to clean up repo-wide, not yours to fix here. Logical conjunction of a parsed condition tree simply isn't a CR concept.

Every other CR in the PR grep-verifies and describes its code correctly: 207.2c, 205.2a, 108.3, 109.5, 109.2a, 400.1, 404.1, 701.17 (Mill), 701.25 (Surveil), 601.2, 602.2.

**`cargo ai-gate` is a CI check here** ("Paired-seed AI gate"), not something you're expected to run locally. It was `in_progress` at this head along with the Rust shards and the Decision-cost perf gate — nothing failing, just unfinished.

---

**Recommendation: request changes on the two reach blockers.** The correctness work you did on the five prior items is solid and I don't want it lost in the noise — the comparator normalization with its exhaustive match and the empty-`Or` edge case are better than most first attempts at that shape. The remaining problem is that the detectors only see one of three condition carriers each, so the axis never activates for the deck it was written for. One change giving both detectors the same three-carrier, chain-walking scan fixes both blockers and the `Dig` gap together.

minion1227:
Both reach blockers fixed in `f6239086f` — as one change, since they were mirror images of the same omission.

## 🔴 Blocker 1 — enabler detection now reads `face.triggers`

`fills_own_graveyard_parts` takes `(abilities, triggers)` and walks `triggers[*].execute`, mirroring `tokens_wide::is_token_generator_parts`. Your diagnosis of the mechanism was the important part: the geometric mean in `compute_commitment` turns `enabler_count == 0` into `commitment == 0.0`, which falls under `GRAVEYARD_TYPES_FLOOR` and switches the policy off — so this wasn't "scores low", it was inert for the exact deck the axis targets.

Guarded two ways: `trigger_borne_mill_is_an_enabler` (the Stitcher's Supplier shape, `abilities: []` + mill trigger) and `trigger_only_enablers_still_clear_the_floor`, which drives the real commitment path and asserts a trigger-only shell clears the floor — so a regression fails on the *consequence*, not just the predicate.

## 🔴 Blocker 2 — the third condition carrier, walked down the chain

Thanks for the correction before I coded it — it was load-bearing. `ability_graveyard_type_threshold` handles `QuantityCheck` and unwraps `ConditionInstead`, with the same And-max / Or-all-then-min / Not-`None` rules as the other two extractors; `ability_chain_graveyard_type_threshold` then walks `sub_ability` **and** `else_ability`. A top-level-only fold would indeed still have returned `None` for Traverse, so the fixture puts the gate on `abilities[0].sub_ability.condition` exactly as you quoted. All three carriers now feed one `.max()`, verified by `ability_chain_threshold_participates_in_highest_threshold`.

## 🟡 Non-blocking, all taken

- **`Effect::Dig { rest_destination: Graveyard }`** is now an enabler (CR 701.20e). It needed the three-carrier walk as you predicted — covered from both carriers, plus a rest-to-library negative so the arm discriminates on the destination rather than on "is a Dig".
- **`CastSpell` gate no longer drops `immediate_etb_triggers`.** Agreed the CR 601.2 reasoning supports excluding *activated* abilities and overshot to ETB. `verdict_credits_a_cast_whose_etb_trigger_mills` covers the delirium play, with a non-filling-ETB control.
- **Single authority restored.** Extended `reanimator::effect_fills_own_graveyard` (adds `DiscardCard`, `Surveil`, rest-to-graveyard `Dig`; scopes `SelfRef` / `Typed{You}` / `Or` / `And`) and both axes call it. **Flagging the cross-axis effect explicitly:** this widens reanimator's own enabler detection too — surveil and dig-to-graveyard genuinely do fill a graveyard for reanimation, so I believe it's a fix rather than drift, and reanimator's 32 tests still pass. Say the word if you'd rather that widening land separately.
- **Docs:** Tarmogoyf is `CountScope::All` and correctly rejected, so Consuming Blob is the example throughout; the "walks one zone's objects" claim now matches the `state.objects` filter; the enabler list includes `Dig`.
- **Dead branch removed** — `scalar / deficit` already peaks at `scalar` when the deficit is 1.

## On the fixtures

You were right that path-divergence is why all four gaps shipped green, so the fixture set now covers each previously-invisible path: trigger-borne mill, Dig-borne (both carriers) + negative, ability-chain threshold + its `highest_threshold` participation, the commitment-path floor guard, and ETB-trigger casts + control.

Left the two inherited `CR 109.3` annotations alone per your note that they're a repo-wide cleanup.

`cargo test -p phase-ai --lib` — **1409 passed** (graveyard suite 39 → 47). `cargo clippy -p phase-ai -p engine --all-targets --features proptest -- -D warnings` — clean.

minion1227:
Addressed the CodeRabbit round in `691f7e44f`, and one note on the red CI.

**✅ Taken — Not/And/Or coverage for the new ability-chain carrier.** Fair catch: the static and trigger carriers each had negation and compound tests, but the third carrier shipped positive-path only, so its `Not => None` rejection and And-max / Or-all-then-min rules were unguarded. Added through the Traverse-shaped `sub_ability` fixture: negated "N or more types" rejected, a non-positive comparator on the same carrier rejected, `And` takes the highest mandatory threshold, and `Or` yields one only when every branch is a graveyard threshold.

**⏭️ Skipped — making `target_fills_own_graveyard` exhaustive over `TargetFilter`.** Two reasons: matthewevans already classified the wildcard-exhaustiveness observation as a non-blocking style concern on the previous round, and `TargetFilter` has ~30 variants, so enumerating them inside `reanimator.rs` — a shared authority I only minimally extended — is churn in another feature for no semantic gain. The `_ => false` is deliberately fail-closed, which is the safe direction for "is this *my* graveyard": a new variant under-counts an enabler rather than mis-crediting an opponent-facing mill as a self-enabler. Happy to do it if you'd rather, ideally as its own repo-wide pass alongside the `CR 109.3` cleanup.

`cargo test -p phase-ai --lib` — **1412 passed** (graveyard suite 47 → 50). `cargo clippy -p phase-ai -p engine --all-targets --features proptest -- -D warnings` — clean.

---

**On the red `Paired-seed AI gate` / `Decision-cost perf gate`: those are infrastructure, not this PR.** Both jobs produced **zero** binary output and ended in `The operation was canceled` with orphaned `ai-gate` / `ai-perf-gate` processes — a hang/cancel, not a threshold breach. Repo-wide evidence:

- Every `ai-gate.yml` run through ~23:44 succeeded, including both of my branches.
- `fix/5981-gift-recipient-selection` (not mine) was cancelled at **00:37**, *before* either of my pushes (00:59, 01:20).
- `feat/kamigawa-flip-cards` started 00:18 and was still running an hour later.

That pattern points at runner/concurrency exhaustion rather than anything in the diff, so I've left it alone per the "external CI → note, don't touch" rule. Locally this branch is green on the full suite and clippy; happy to re-run the gate once the runners drain.

minion1227:
Confirming your merge `3bb5eefa8` — I verified it rather than assuming, and **took yours instead of pushing my own**.

Timing overlap: #6543 landed while this loop was mid-cycle, so I resolved the same eight conflicts locally. Before pushing, my pre-push divergence check reported DIVERGED against `3bb5eefa8`, so I discarded my merge and reset onto yours — never force-pushing over a maintainer commit. Confirmed `691f7e44f` is an ancestor of your merge, so none of this PR's work was dropped.

Our resolutions agreed on every hunk, including the two non-obvious ones:
- the `UNTUNED_POLICY_PENALTY_FIELDS` entry, where a naive keep-both fuses the two `(name, reason)` pairs into a single 4-element tuple (a type error, not a silent one — worth noting for the next axis PR that collides here);
- the two artifact-compat tests, which share a doc prefix and tail so git interleaves them into one function — both of us split them back into two whole `#[test]` fns, each pinning its own field's `#[serde(default)]`.

Verified on your merged tree:
- `cargo test -p phase-ai --lib` — **1470 passed**
- graveyard suite **50**, poison suite **58**, and both `policy_penalties_load_pre_*_artifact` tests present and passing — so neither axis lost coverage in the port.

Agreed the registry `Vec` order isn't load-bearing. Nothing further from me on the merge; the two reach blockers and the CodeRabbit round remain addressed as of `691f7e44f`.

minion1227:
@matthewevans — requesting a re-review when you have a moment.

Everything from both of your rounds is addressed and the branch is green on your merge head `3bb5eefa8`:

- **Round 1 (5 blockers)** — serde-default artifact compat, `Option<u32>` "no threshold", comparator/`Not`/`Or` truth-semantics, controller-only scope, and `CastFacts`/`effective_activated_ability` action semantics.
- **Round 2 (2 reach blockers)** — both detectors now share the same three-carrier, chain-walking scan (`f6239086f`): enabler detection reads `triggers[*].execute`, threshold detection walks `AbilityDefinition.condition` down the `sub_ability`/`else_ability` chain. Also took the `Dig`-to-graveyard enabler, restored ETB triggers to the cast gate, and consolidated onto the single `reanimator::effect_fills_own_graveyard` authority.
- **CodeRabbit** — `Not`/`And`/`Or` regression coverage added for the new ability-chain carrier (`691f7e44f`); skipped the `TargetFilter` exhaustiveness nit with a reason, per your earlier note that it's non-blocking style.

CI is **15/15 green** (Paired-seed AI gate, Decision-cost perf gate, Tauri compile check, full Rust suite). Locally: `cargo test -p phase-ai --lib` 1470 passed, graveyard suite 50, poison suite 58 (both axes coexist after the port), clippy clean.

The `CHANGES_REQUESTED` showing is your original review — since the repo doesn't auto-dismiss stale reviews, it needs a fresh look to clear. No rush; flagging it as ready whenever you're free.

matthewevans (approved):
Re-reviewed at `3bb5eefa`. All five findings from the prior round are resolved, and each is now pinned by a test that discriminates the exact failure it fixes. Approving.

## Findings, closed

1. **serde default** — `#[serde(default = "default_graveyard_types_progress")]` on the field, wired into `Default`, with the pre-existing-artifact rationale in the doc comment. `ai_tune` artifact load is green in CI.
2. **threshold absence** — `highest_threshold: Option<u32>`; the verdict routes a `None`-threshold scaling deck into the `has_scaling` branch (`scalar / (current+1)`, diminishing but never zero), so a Consuming Blob deck keeps being rewarded past four types. Pinned by `verdict_still_rewards_scaling_only_deck_above_four_types` and `descend_eight_tracks_highest_threshold`.
3. **comparator / Not / And / Or** — `positive_graveyard_threshold` handles both orientations (`flip_comparator`), the strict-`>` off-by-one (`GT` → `bound+1`), and rejects `<`/`<=`/`=`/`!=` with an exhaustive `Comparator` match. `Not` → `None`; `And` → max (mandatory conjunction, CR 109.3); `Or` → threshold only when every branch is a graveyard gate, else `None`. Pinned by `non_positive_comparators_are_not_thresholds`, `mirrored_orientation_normalizes_to_lower_bound`, `negated_threshold_is_not_a_payoff`, `and_takes_the_highest_mandatory_threshold`, `or_is_a_gate_only_when_every_branch_is_graveyard` — and the ability-chain twins.
4. **CountScope consistency** — `quantity_reads_own_graveyard_types` now matches only `Controller | Owner`, so an all-graveyards (Tarmogoyf-class) payoff is no longer misread as an own-graveyard plan against the controller-scoped `distinct_graveyard_types`. Pinned by `all_graveyards_scope_is_not_own_graveyard`.
5. **authoritative ability semantics** — `candidate_fills_own_graveyard` uses `cast_facts()` (`primary_effects` + `immediate_etb_triggers`, correctly excluding activated abilities on a cast per CR 601.2) and `effective_activated_ability()` (the runtime-enumerated index per CR 602.2) instead of walking `GameObject::abilities`. Pinned by `verdict_ignores_cast_of_a_body_with_an_activated_mill`, `verdict_credits_the_activated_mill_ability`, and the ETB-trigger pair.

## Gates

`Paired-seed AI gate` and `Decision-cost perf gate` both green — the live registry addition doesn't regress AI strength or the per-candidate `verdict()` cost, and no baseline bump was required. The condition combinators are decomposed into typed helpers with CR annotations rather than card-name matching, and the extractor is shared across all three condition carriers (static / trigger / ability-chain) — the right seam. Nice work.

minion1227:
Status on head `3bb5eefa8`: all 14 required checks green — including the **Paired-seed AI gate** and the **Decision-cost perf gate** — and the branch is `MERGEABLE`/`CLEAN` against `main`. Thanks for the review round.

One CodeRabbit thread is still open — the exhaustiveness nit on `target_fills_own_graveyard` (`crates/phase-ai/src/features/reanimator.rs:289`). I've left the code as-is deliberately, and want to record why rather than reset an approved head:

- The match already handles every **controller-scoped** `TargetFilter` explicitly: `Controller`, `SelfRef`, `Any`, `Typed { controller: You }`, plus the `Or`/`And` recursions (with `And` requiring every conjunct, per CR 608.2c). No *current* self-graveyard variant falls through.
- The `_ => false` arm therefore only catches opponent-facing scopes (`Player`, `Opponent`) — correctly *not* own-graveyard fillers — and any future variant. Its failure direction is **undercount**: at worst a new variant makes a card not register as an enabler, which nudges `commitment` down. It can never over-credit or misfire the policy.

Given that, and that the PR is already approved and the paired-seed gate is expensive to re-run, I'd rather tighten this to an explicit exhaustive match in a small follow-up than push a new commit onto an approved head now. Happy to do that immediately if you'd prefer it in-PR instead — just say the word.

PR #6548, opened 2026-07-23
#6501 Prune connections map on game teardown

### LINKED ISSUE #6501: Server `connections` map is never pruned (memory leak per finished game)
## Summary
The in-memory `connections` map in `crates/phase-server/src/main.rs` is inserted into on every host / join / reconnect, but a `game_code` entry is **never removed** — even after the game ends. Its sibling `game_spectators` map *is* pruned (`specs.remove(game_code)`) at the same teardown points, so the asymmetry leaks one `connections` entry per finished game.

## Impact
`connections: HashMap<String /*game_code*/, HashMap<PlayerId, Sender>>` grows without bound on a long-running server: every completed or expired game leaves a permanent `game_code -> {seat -> dead Sender}` entry. `MAX_GAMES` bounds only *concurrent* live sessions, not this map. Over a long uptime the map accumulates one dead entry per game played, requiring periodic restarts.

## Where
`connections` is declared at `crates/phase-server/src/main.rs:922` and inserted at the host/join/reconnect paths. The background reaper tears games down in two loops — disconnect grace-expiry (~line 1087) and lobby expiry (~line 1130) — and prunes `game_spectators` there, but never `connections`.

## Fix
Mirror the existing `game_spectators` cleanup: `connections.remove(game_code)` in both reaper teardown loops, preserving the `state -> connections -> spectators` lock order. No behavior change beyond releasing the dead entry.

I intend to fix this myself and will open a PR shortly.

boskodev790:
## Summary
Fixes #6501: the in-memory `connections` map (`crates/phase-server/src/main.rs`) was inserted into on every host / join / reconnect but a `game_code` entry was **never removed**, unlike the parallel `game_spectators` map which is pruned at the same teardown points. This leaks one `connections` entry per finished game.

## Root cause
`connections: HashMap<String /*game_code*/, HashMap<PlayerId, Sender>>` grows without bound: every completed or expired game leaves a permanent `game_code -> {seat -> dead Sender}` entry. `MAX_GAMES` bounds only concurrent live sessions, not this map, so a long-running server accumulates one dead entry per game played.

## Fix
Mirror the existing `game_spectators` cleanup — `connections.remove(game_code)` in both background-reaper teardown loops:
- disconnect grace-expiry loop (already holds the connections lock; changed `let conns` → `let mut conns` and remove after the GameOver fan-out),
- lobby-expiry loop (added a `connections` prune block).

Lock order is preserved throughout (`state -> connections -> spectators`). No behavior change beyond releasing the dead entry.

## Files changed
- `crates/phase-server/src/main.rs` — `+14 / -1`.

## Verification
- `cargo fmt --all` — clean.
- `cargo check -p phase-server` — clean (compiles, borrow-checks). The change sits inside async websocket teardown paths that are not unit-testable in isolation; correctness follows from mirroring the adjacent, already-established `game_spectators` cleanup and preserving lock order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved background cleanup for completed and expired games.
  * Prevented stale connection data from accumulating after games are removed.
  * Enhanced resource management for both full-mode and lobby games.

### DIFF
--- a/crates/phase-server/src/main.rs
+++ b/crates/phase-server/src/main.rs
@@ -245,6 +245,16 @@ async fn switch_game_spectator_slot(
     Ok(())
 }
 
+async fn prune_game_connections<'a>(
+    connections: &SharedConnections,
+    game_codes: impl IntoIterator<Item = &'a str>,
+) {
+    let mut conns = connections.lock().await;
+    for game_code in game_codes {
+        conns.remove(game_code);
+    }
+}
+
 async fn remove_draft_spectator_sender(
     draft_spectators: &SharedDraftSpectators,
     draft_code: &str,
@@ -1319,25 +1329,32 @@ async fn serve() {
                         })
                         .collect::<Vec<_>>()
                 };
-                let conns = bg_connections.lock().await;
-                for session in &removed {
-                    let game_code = &session.game_code;
-                    info!(game = %game_code, reason = "disconnect_expired", "game over");
-                    if let Some(players) = conns.get(game_code) {
-                        if let Some(deliveries) = prepared.get(game_code) {
-                            for (player, delivery) in deliveries {
-                                if let Some(sender) = players.get(player) {
-                                    let _ = sender.send(ServerMessage::TerminalResult {
-                                        delivery: Some(delivery.clone()),
-                                    });
+                {
+                    let conns = bg_connections.lock().await;
+                    for session in &removed {
+                        let game_code = &session.game_code;
+                        info!(game = %game_code, reason = "disconnect_expired", "game over");
+                        if let Some(players) = conns.get(game_code) {
+                            if let Some(deliveries) = prepared.get(game_code) {
+                                for (player, delivery) in deliveries {
+                                    if let Some(sender) = players.get(player) {
+                                        let _ = sender.send(ServerMessage::TerminalResult {
+                                            delivery: Some(delivery.clone()),
+                                        });
+                                    }
                                 }
                             }
                         }
-                    }
-                    if !session.game_started {
-                        retire_unstarted_session_async(&bg_game_db, session);
+                        if !session.game_started {
+                            retire_unstarted_session_async(&bg_game_db, session);
+                        }
                     }
                 }
+                prune_game_connections(
+                    &bg_connections,
+                    removed.iter().map(|session| session.game_code.as_str()),
+                )
+                .await;
                 let mut specs = bg_game_spectators.lock().await;
                 for session in &removed {
                     specs.remove(&session.game_code);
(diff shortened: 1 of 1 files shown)


matthewevans (changes requested):
**Verdict: blocked — the teardown cleanup has no discriminating regression test.**

## 🔴 Blocker

`crates/phase-server/src/main.rs:1106` and `:1147` remove expired game codes from the connections map, but this PR adds no test that drives either expiry path and proves the entry is removed while another game’s connection remains. The author’s cited compile check cannot fail when either `conns.remove(game_code)` is reverted.

Suggested fix: add a focused server/lifecycle test that seeds two connection-map entries, exercises both relevant teardown routes, and asserts only the expired game’s entry is pruned.

## ✅ Clean

The cleanup follows the existing state → connections → spectators lock order and leaves the send-before-removal behavior intact.

**Recommendation: add the discriminating teardown regression test, then return for approval.**

matthewevans:
The requested regression coverage from the current review is still outstanding on `fc68538be5cf616f0896eed325cc420a04056ec6`. Please address it and push an update within 7 days; otherwise this PR will be auto-closed.

matthewevans (changes requested):
**Request changes — current head `867e44871482a6901ea4b05a16f606016cbc268b` still lacks a regression that reaches either teardown integration.**

`prune_game_connections` is correct in isolation, and the two new tests prove that helper removes only the requested key. They do not invoke the disconnect-grace expiry or lobby-expiry paths where the helper is wired in (`crates/phase-server/src/main.rs:1329-1362` and `:1396-1400`). Revert either of those two call sites and both tests still pass, so the leak this PR fixes can return unnoticed.

Please add focused coverage that exercises each teardown route (or factor the route-owned cleanup into a testable unit that both production paths call) with an expired and a surviving game, then assert only the expired connection entry is removed. Keep the existing helper tests as the low-level boundary checks.

matthewevans (approved):
Maintainer completion verified on 8fa62b3d: current CI, including the security scan, is green. The remaining security threads are stale findings on the pre-rebuild merge-history commit; the rebuilt server-only head no longer contains those dependency changes.

matthewevans (changes requested):
**Request changes — current head `8fa62b3db2c6dc67c8be4ffa664c4b48bdacfa42` still lacks a discriminating regression for the production teardown wiring.**

`prune_game_connections` is correct in isolation, and the two added tests prove that helper removes only the requested key. But neither invokes the disconnect-grace expiry route or the lobby-expiry route where that helper is wired (`crates/phase-server/src/main.rs:1353` and `:1399`). Revert either production invocation and both tests remain green, so the completed-game connection leak can return unnoticed.

Please add focused coverage that exercises each teardown route, or factor route-owned cleanup into a testable unit that each production route calls, with an expired and surviving game and an assertion that only the expired connection entry is removed. Keep the existing helper tests as low-level boundary coverage.

The unresolved Superagent supply-chain threads are outdated and refer to files absent from this rebuilt server-only diff; they do not block this review.

matthewevans (approved):
Re-approving the unchanged rebuilt head after queue revalidation: current CI and security scan remain green; the historical scanner findings are stale pre-rebuild threads.

Issue #6507, opened 2026-07-22
Gemstone Mine — does not sacrifice after its last mining counter is removed


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1529129619956170865/1529129619956170865

**Thread:** Gemstone Mine — does not sacrifice after its last mining counter is removed
**Discord thread id:** `1529129619956170865`

## Summary
[[Gemstone Mine]]

### gemstone mine
This land enters with three mining counters on it.
{T}, Remove a mining counter from this land: Add one mana of any color. If there are no mining counters on this land, sacrifice it.

---
<sub>report_id: `discord:1529129619956170865:1529129619956170865:0` · discord: `1529129619956170865/1529129619956170865`</sub>

mike-theDude:
## Triage analysis

**Clarified summary**
Gemstone Mine is not sacrificed after its final mining counter is removed; it stays on the battlefield.

**Oracle text (verified from `client/public/card-data.json`)**
> This land enters with three mining counters on it.
> {T}, Remove a mining counter from this land: Add one mana of any color. If there are no mining counters on this land, sacrifice it.

**Investigation findings**
The parse claims support and is mostly faithful — the enters-with-three-mining-counters replacement is represented, the mana ability produces one mana of any colour, and a conditional sacrifice child is attached with the correct condition of zero mining counters remaining on the land.

The defect is in what that sacrifice is aimed at: the sacrifice child is bound to the parent ability's target rather than to the land itself. The mana ability has no target, so that binding resolves to nothing and the sacrifice never has a subject to act on. The Oracle text sacrifices the source land ("sacrifice it" refers to the land), so the child should bind to the ability's source.

This subject-binding shape is a class, not a one-off: any "…: do something. If <condition>, sacrifice it" self-referential ability on a targetless activated ability has the same exposure.

**Acceptance criteria**
- Activating the ability with one mining counter remaining removes it, adds the mana, and then sacrifices Gemstone Mine.
- Activating it with counters still remaining adds mana and does not sacrifice.
- The sacrifice binds to the ability's source rather than to a parent target.

mike-theDude:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `{T}, Remove a mining counter from this land: Add one mana of any color. If there are no mining counters on this land, sacrifice it.`  (supported: true)

[...]

PR #6559, opened 2026-07-23
fix(parser): bind source-counter-gated rider pronouns to the source (Gemstone Mine #6507)

jeffrey701:
## Summary

Fixes **Gemstone Mine** (#6507) and the whole source-counter-conditioned rider class: the depletion-land sacrifice rider — *"{T}, Remove a mining counter from this land: Add one mana of any color. If there are no mining counters on this land, sacrifice it."* — never sacrificed the land after its last counter was removed.

## Root cause

Parse-time anaphor mis-binding, not a runtime gap. The rider's sub-ability parsed to `Sacrifice { target: ParentTarget }`. A mana ability has no targets (CR 605.1a), so `ParentTarget` resolves to an empty set at resolution (`game/effects/sacrifice.rs` — `effect_object_targets(ParentTarget, [])` is empty → the sacrifice silently no-ops and the land survives).

The effect-chain parser already binds a bare **"it"** to `SelfRef` when the gating condition references the source object, via `condition_refs_source_object`. That predicate recognized source-tapped / source-entered / source-attached conditions, but **not a source-scoped counter threshold** (`QuantityCheck` over `CountersOn { scope: Source }`) — the exact shape these riders produce. So the counter-gated body pronoun fell through to `ParentTarget` (and, on typed triggers, to `TriggeringSource`).

## Fix

One additive predicate extension in `crates/engine/src/parser/oracle_effect/mod.rs` (+41/-0):

- a new exhaustive `QuantityExpr` walker `quantity_expr_reads_source_counters` (no wildcard arm — a future variant must be classified; mirrors `quantity_expr_uses_recipient`);
- a `QuantityCheck { lhs, rhs, .. }` arm in `condition_refs_source_object` returning true when either side reads counters on the source.

This single change drives **both** existing consumers of the predicate: the chunk-subject threading (mod.rs:28564) now supplies `SelfRef`, and the `ParentTarget` rewrite guard (mod.rs:29335) now skips these chunks. No runtime files change; the AST is now what the card says (`SelfRef`), and `sacrifice.rs`'s existing `SelfRef` pool resolution + CR 400.7 epoch guard do the rest.

Corrects the binding for ~21 cards: the 5 Mercadian depletion lands, Gemstone Mine, Tourach's Gate, Contested Game Ball, Daredevil Dragster, Dawn of a New Age, Evolved Spinoderm (`Sacrifice ParentTarget → SelfRef`); Blood Spatter Analysis, Charitable Levy, Decree of Silence, Last Light of Durin's Day, The Heron Moon, ED-E (`Sacrifice/PutCounter TriggeringSource → SelfRef`); Heirloom Mirror, Ludevic's Test Subject, Replicating Ring, Smoldering Egg (`RemoveCounter ParentTarget → SelfRef`). Grasping Shadows / Soulcipher Board flip `Transform SelfRef → ParentTarget`, which is behavior-neutral (the transform effect handler falls back to the source on empty targets).

## Files changed

- `crates/engine/src/parser/oracle_effect/mod.rs` — the fix (helper + match arm)
- `crates/engine/src/parser/oracle_effect/tests.rs` — predicate unit test (Source→true incl. wrapped/`Not`/`And`; Target/Recipient scopes→false)
- `crates/engine/src/parser/oracle_tests.rs` — 2 parser SHAPE tests (Gemstone Mine, Last Light) with reach-guards
- `crates/engine/tests/integration/gemstone_mine_depletion_sacrifice_6507.rs` — 5 runtime tests
- `crates/engine/tests/integration/main.rs` — mod line

## CR references

- `CR 122.1` (counters) + `CR 608.2k` (an effect referring to an untargeted object previously referred to by the ability still affects it) — the new predicate arm / helper.
- `CR 605.1a` (mana ability requires no target) + `CR 605.3b` (mana ability resolves immediately) + `CR 701.21a` (sacrifice) — test annotations.

Method: /engine-implementer

## Track

Developer

## LLM

Model: claude-opus-4-8[1m]
Thinking: high

## Verification

- `cargo test -p engine --lib` — **17593 passed, 0 failed, 6 ignored** (baseline 17590 + 3 new)
- `cargo test -p engine --test integration` — **3874 passed, 0 failed, 2 ignored** (baseline 3869 + 5 new)
- `./scripts/gen-card-data.sh` — regenerated; parse-diff audit of the source-counter rider shape class: 23 cards changed, all either correctness heals or behavior-neutral (Transform empty-target→source fallback), zero regressions.
- RED/GREEN: tests 1/3/4/5 + both shape tests fail on the pre-fix `ParentTarget`/`TriggeringSource` binding and pass after; test 2 is the paired over-trigger negative with positive reach-guards.

## Gate A

Gate A PASS head=1c74fedb5bef8d76fb45de2c5e22833208179d1d base=6ae8737cdab0fa1ed291cad0f0808473a90f4cf8

## Anchored on

- `crates/engine/src/parser/oracle_trigger.rs:1297` — trigger-body parse constructs `effect_ctx` with `subject: Some(trigger_subject.clone())` (the established subject-threading seam that already binds trigger-borne riders correctly).
- `crates/engine/src/parser/oracle_effect/mod.rs:2876` — delayed-trigger body parse threads `inner_ctx.subject = Some(TargetFilter::SelfRef)` for the self-referential case — the exact `SelfRef` subject-threading this change completes for the counter-gated chunk path (mod.rs:28564).

## Final review-impl

Final review-impl PASS head=1c74fedb5bef8d76fb45de2c5e22833208179d1d

## Claimed parse impact

- Gemstone Mine, Peat Bog, Hickory Woodlot, Remote Farm, Sandstone Needle, Saprazzan Skerry, Tourach's Gate, Contested Game Ball, Daredevil Dragster, Dawn of a New Age, Evolved Spinoderm, Blood Spatter Analysis, Charitable Levy, Decree of Silence, Last Light of Durin's Day, The Heron Moon, ED-E Lonesome Eyebot, Heirloom Mirror, Ludevic's Test Subject, Replicating Ring, Smoldering Egg (correctness heals); Grasping Shadows, Soulcipher Board (behavior-neutral Transform relabel).

## Validation Failures

None.

## CI Failures

None.

---

Tier: Frontier

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of source-specific counter conditions, including nested, negated, combined, and arithmetic expressions.
  * Corrected pronoun and target binding in effects involving counter checks and sacrifice abilities.
  * Improved handling of recipient counter conditions to preserve the correct affected object.
  * Fixed land depletion sacrifice effects and related abilities so they consistently affect the intended objects across activation and payment scenarios.

* **Tests**
  * Added regression and integration coverage for counter-based abilities, land depletion sacrifices, payment paths, and granted abilities.


matthewevans:
**Request changes** — the diagnosis, the seam, and the CR work are all correct, but the predicate is one degree too broad and lands a functional regression on **Revelation of Power**, a card outside the claimed scope.

## 🔴 Blocker

### `Revelation of Power` loses its conditional grant entirely

`crates/engine/src/parser/oracle_effect/mod.rs` — the new `AbilityCondition::QuantityCheck` arm in `condition_refs_source_object`.

Oracle text, verbatim from Scryfall (`Instant`, Streets of New Capenna):

> Target creature gets +2/+2 until end of turn. If it has a counter on it, it also gains flying and lifelink until end of turn.

Both pronouns anaphor to **target creature**. The card is an Instant, so it can carry no counters and can gain no flying/lifelink.

The parse-diff sticky for this head reports it twice, and it is absent from **## Claimed parse impact**:

```
- **1 card** · 🔄 ability/grant Flying, grant Lifelink · changed field `affects`: `parent target` → `self`
  - Affected (first 3): Revelation of Power
- **1 card** · 🔄 ability/grant Flying, grant Lifelink · changed field `target`: `parent target` → `∅`
  - Affected (first 3): Revelation of Power
```

Baseline AST on `main` (from `data/card-data.json`) is already correct, and shows exactly why the new arm fires:

```json
"static_abilities": [{
  "affected": { "type": "ParentTarget" },            //  correct: the target creature
  "modifications": [ {"type":"AddKeyword","keyword":"Flying"},
                     {"type":"AddKeyword","keyword":"Lifelink"} ],
  "condition": { "type": "QuantityComparison",
    "lhs": { "type":"Ref", "qty": { "type":"CountersOn",
             "scope": { "type":"Source" } } },       //  pre-existing latent mis-scope
    "comparator": "GE", ... }
}]
```

The condition's `scope: Source` here is **not** a real source reference — it is an unresolved bare `"it"` that should have lowered to the parent target. That mis-scope was harmless while `affected` stayed `ParentTarget`. The new arm keys off precisely this shape, drives `chunk_subject → SelfRef` (mod.rs:28564), and rebinds the grant to the Instant. The rider becomes a permanent no-op: the card loses its second sentence. This is the `engine_regress` bucket — a card losing a previously-working handler.

The cited rule also does not reach this case. `CR 608.2k`, verbatim:

> 608.2k If an ability's effect refers to a specific untargeted object that has been previously referred to by that ability's **cost or trigger condition**, it still affects that object even if the object has changed characteristics.

Gemstone Mine qualifies — its cost, *"Remove a mining counter from this land"*, refers to the source. Revelation of Power has neither a cost nor a trigger condition referring to the source; the predicate is reading an intervening-if gate in the effect body, which is a strictly wider surface than 608.2k licenses.

**The discriminator you need is already visible in the two cards.** A true source gate names its object — *"no mining counters on **this land**"*, *"If **Blaster** has no +1/+1 counters on it"*. A false positive gates on a bare pronoun — *"If **it** has a counter on it"* — inside a chain whose prior clause already declared a chosen target. Narrow the arm so it fires only when the counter reference came from an explicit source noun phrase (`~` / "this land" / the card name), or suppress it when the chunk's chain already carries a chosen-target referent (`chain_prior_referent_is_chosen_target` / `parent_target_available` are both in scope at mod.rs:28564). Either gate keeps all 21 intended heals and drops Revelation of Power out of the class.

## 🟡 Non-blocking

**Claimed set does not match the measured set.** The body states *"23 cards changed, all either correctness heals or behavior-neutral … zero regressions"*, and the claimed list also has 23 names — but the two sets are not identical. Revelation of Power is measured and unclaimed, so one claimed name is not actually in the diff. The matching totals made the substitution invisible. Worth re-deriving the claim from the artifact rather than by count.

**Possible incomplete class coverage.** Scanning `data/card-data.json` for the predicate's shape (source-scoped `CountersOn` gate + `ParentTarget`/`TriggeringSource` binding) also surfaces `Hostile Hostel` and `Blaster, Morale Booster`, neither of which appears in the parse diff. Blaster's back face reads *"Move X +1/+1 counters from Blaster onto another target artifact. … If Blaster has no +1/+1 counters on it, convert it."* — an explicitly-named source gate on a **targeted** activated ability, where the current `ParentTarget` binding would convert the targeted artifact instead of Blaster. If a guard is holding these back, the class fix is partial. Confidence: moderate — my scan is a local approximation over a `main` snapshot, not the CI artifact, so please confirm against the full parse-diff artifact rather than the truncated sticky (the `+8 more` buckets may hide further unclaimed cards).

**CodeRabbit's nitpick is fair.** `quantity_expr_reads_source_counters` has nine recursive arms; the unit test drives propagation through `Ref`/`Offset` only. Adding `Sum`/`Difference` cases would lock in the exhaustive-walk guarantee the doc comment advertises — the two-operand arms are where a copy-paste slip would hide.

## ✅ Clean

- **Root-cause analysis is correct and precisely stated.** `ParentTarget` does fall to the `_ =>` arm of `effect_object_targets` (game/effects/mod.rs:251), resolving to the ability's chosen targets — empty for a mana ability per `CR 605.1a`. The Gemstone Mine diagnosis holds exactly as written.
- **Right seam, no new vocabulary.** Extending the existing `condition_refs_source_object` predicate drives both of its existing consumers rather than adding a parallel path, and no new enum variant was introduced.
- **The Transform relabel is genuinely behavior-neutral, and I verified it rather than taking the claim.** `transform_effect::resolve` dispatches on `ability.targets.as_slice()` with `[] => ability.source_id` (transform_effect.rs:23-30) and never consults the `target` filter, so `Grasping Shadows` / `Soulcipher Board` are unaffected. Confirming this in the body was the right call.
- **Every CR citation grep-verifies against `docs/MagicCompRules.txt`**: 122.1 (counters), 608.2k, 605.1a, 605.3b, 701.21a. No invented numbers — this is consistently done well in your PRs.
- **The exhaustive `QuantityExpr` walker has no wildcard arm**, so a future variant becomes a compile error instead of a silent `false`. Correct instinct, and the `quantity_expr_uses_recipient` mirror is the right precedent to follow.
- **Integration test is registered** — `mod gemstone_mine_depletion_sacrifice_6507;` is present in `crates/engine/tests/integration/main.rs`, so the 5 runtime tests actually run.
- **Shape tests carry non-vacuity reach-guards** (zero parse warnings, no `Effect::Unimplemented` anywhere in the parse, asserted ability count) before the shape assertions. This is exactly the guard that keeps a negative assertion from passing for the wrong reason.

**Recommendation:** narrow the `QuantityCheck` arm to explicit source noun phrases — or suppress it when the chain already has a chosen-target referent — and add a regression test pinning `Revelation of Power`'s rider to `ParentTarget`. Then regenerate the parse diff and re-derive the claimed-impact list from the artifact, confirming whether `Hostile Hostel` and `Blaster, Morale Booster` should be in or out of the class. The Gemstone Mine fix itself is sound and worth landing once the blast radius is contained.

matthewevans (changes requested):
Current head 11b490add3ed8229122b4665ad862e8c168de4e9 remains blocked.

condition_refs_source_object treats every QuantityCheck containing CountersOn { scope: Source } as a source reference. That is over-broad: Revelation of Power's bare "it" refers to its target creature, so this predicate rebinds the conditional flying/lifelink rider to the Instant itself and drops previously working behavior.

Narrow this to an explicit source noun phrase or suppress the path when the chain already carries the chosen-target referent. Add the Revelation of Power regression test and regenerate/reconcile the parse-diff before requesting review again.

jeffrey701:
Thank you — the Revelation of Power catch was exactly right, and the diagnosis ("a mis-scoped bare `it` gated by an intervening-if, not a real cost/trigger source reference per CR 608.2k") is the correct framing. Fixed at head `2b3063d8f`.

## Blocker — Revelation of Power no longer regresses

I took the **suppress-when-a-prior-clause-chose-a-typed-target** option. At the `chunk_subject` site (`oracle_effect/mod.rs`) I compute one gate and reuse it at both consumers (the chunk-subject binding and the `replace_target_with_parent` guard):

```rust
let binds_source_counter_pronoun = condition
    .as_ref()
    .is_some_and(condition_refs_source_object)
    && !chain_has_prior_typed_referent(builder.clauses(), false);
```

`chain_has_prior_typed_referent` is true for Revelation of Power — its first clause, *"Target creature gets +2/+2"*, is a `Pump { target: Typed(Creature) }` (a chosen typed target) — so the source binding is suppressed and the flying/lifelink grant stays on the parent target. Regenerated card-data confirms it:

- `Revelation of Power` → grant `affected` / `target` = **ParentTarget** (the flying/lifelink `GenericEffect` is back on the target creature; the mis-scoped `CountersOn{Source}` condition is untouched and harmless again).

I chose `chain_has_prior_typed_referent` over `chain_prior_referent_is_chosen_target` deliberately: the latter's `has_typed_target_widened` early-out returns `false` for a pump-of-a-target, so it would *not* have caught Revelation of Power. `chain_has_prior_typed_referent` returns `true` for the prior typed target and `false` for the depletion/rider heals (whose prior clause is *"Add mana"* or *"put a counter on ~"*, never a chosen typed target), so every intended heal is preserved:

| rider class | prior clause | binds source? | rider target |
|---|---|---|---|
| Gemstone Mine + 5 depletion lands | "Add mana" (no referent) | yes | **SelfRef** ✓ |
| Last Light / ED-E / Daredevil Dragster / Blood Spatter Analysis / Charitable Levy / Decree of Silence / The Heron Moon / Contested Game Ball / Dawn of a New Age / Evolved Spinoderm / Tourach's Gate | "put/remove a counter on ~" (SelfRef) | yes | **SelfRef** ✓ |
| Heirloom Mirror / Ludevic's Test Subject / Replicating Ring / Smoldering Egg | "remove a counter from ~" (SelfRef) | yes | **SelfRef** (RemoveCounter) ✓ |
| **Revelation of Power** | "Target creature gets +2/+2" (typed target) | **no** | **ParentTarget** ✓ |

Added `source_counter_gate_over_prior_target_keeps_parent_not_self_ref` (parser SHAPE test, reach-guarded) pinning Revelation of Power's grant to `ParentTarget`.

## Non-blocking items

- **Claimed set re-derived from the artifact.** Corrected list (behavior heals): Gemstone Mine, Peat Bog, Hickory Woodlot, Remote Farm, Sandstone Needle, Saprazzan Skerry, Tourach's Gate, Contested Game Ball, Daredevil Dragster, Dawn of a New Age, Evolved Spinoderm (Sacrifice→SelfRef); Blood Spatter Analysis, Charitable Levy, Decree of Silence, Last Light of Durin's Day, The Heron Moon, ED-E Lonesome Eyebot (trigger Sacrifice/PutCounter→SelfRef); Heirloom Mirror, Ludevic's Test Subject, Replicating Ring, Smoldering Egg (RemoveCounter→SelfRef). Revelation of Power is **not** in the diff (it stays on `main`'s ParentTarget). The old body's substitution is gone.
- **Blaster, Morale Booster / Hostile Hostel** — checked against the regenerated artifact: unchanged by this PR (their transform legs are `ParentTarget` in all three of `main`, the previous head, and this head). Blaster's back face ("Move X +1/+1 counters … onto another target artifact. … If Blaster has no +1/+1 counters on it, convert it") is a genuinely harder case: an explicitly-named source gate on a **targeted** activated ability whose "convert it" wants the source while a target artifact is already chosen. That is precisely the shape this guard *excludes* (a prior chosen target), and it is a pre-existing `main` gap, not something this PR introduces or should widen its scope to. It needs the "explicit source noun phrase" discriminator you mentioned (option 1), which is a distinct follow-up — flagging it, out of scope here.
- **Sum / Difference unit coverage** — extended `condition_refs_source_object_source_counter_quantity_check` to drive both two-operand `QuantityExpr` arms (source read in either operand → detected; neither → not), locking in the exhaustive-walk guarantee.

Full lib + integration suites green; parser-combinator gate (Gate A) passes; the branch is up to date with current `main` (picks up the Tauri `Cargo.lock` fix, so the required Rust aggregator is green again).

matthewevans (changes requested):
**Changes requested — current-head production coverage gap.**

## 🟠 Required

The repair for the prior Revelation of Power regression still lacks a production-pipeline regression. `crates/engine/src/parser/oracle_effect/mod.rs:29929-29948` changes the runtime target binding and `:30725-30743` rewrites the parent target, but Revelation is covered only by parser-shape tests in `oracle_tests.rs:22924-22981`; the new integration cases exercise Gemstone Mine, Peat Bog, and Last Light. The original defect was visible only after casting, target propagation, and layer application, where the grant could be applied to the Instant rather than the selected creature. Add an integration test that casts Revelation of Power at a countered creature and asserts that creature receives flying and lifelink; make it discriminating against reversion of the new guard.

## 🟡 Also reconcile before re-review

The sole parse-diff artifact predates this head and still describes Revelation as changing to `self`; the fresh Card data job is in progress. Let it publish and reconcile the exact affected-card/signature set for this head.

**Recommendation: add the end-to-end Revelation regression, then provide current-head parse-diff evidence before re-review.**

matthewevans:
Current-head maintainer hold for `4e4571f48c854aeb04e2e502ea135b7ea32204c7`: this head adds the requested Revelation of Power cast-pipeline regression and narrows the prior typed-referent/source-counter binding repair. It is a materially new parser surface, so the existing parse-diff (`updated 2026-08-01T17:54:47Z`) and approvals/reviews on earlier commits are not current evidence. Required CI and a current full card-level parse artifact must complete before a final re-review or queue action.

matthewevans (approved):
Maintainer completion verified on e841edbc: source-counter rebinding is constrained to the bare-pronoun/prior-target grammar, uses the shared recursive static-condition rebinder, and includes focused source/recipient plus RHS traversal coverage. Stale pre-rebase threads were resolved; current-head CI has been requested and merge-queue checks remain authoritative.

matthewevans:
Current-head hold for `e841edbca8ea5f4168138f1fee19eba83015695d`: this head adds the requested RHS counter-operand coverage and removes the inapt CR 608.2k claim from the source-counter classifier. The focused review is clean, but all required CI (including card data) is still queued for this head and the existing parse-diff artifact predates it. Auto-merge is disabled until those current-head checks and the full parse artifact complete.

matthewevans:
Current-head maintainer hold for `6b8e9c55fdf95581226fc0c40176bf7c0f00227d`: the only delta from the reviewed `e841edbc` head corrects the source-counter unit fixture to use the real `CounterType::Plus1Plus1` variant. The parser implementation and its source/recipient, prior-target, RHS-counter-operand, and Gemstone Mine/Revelation cast-pipeline coverage are otherwise unchanged; the corrected fixture now actually compiles against the enum it exercises.

This push started a new CI run. Rust lint/parser gate, both Rust-test shards, card-data/parse-diff, frontend, WASM, and Tauri are still running. The visible parse-diff artifact predates this push, so it is not yet current-head evidence even though this delta is test-only. Auto-merge remains disabled; no merge-queue action during this hold.

matthewevans (approved):
Re-approving the clean rebase on current main; the reviewed constrained counter-gate rebinding and focused regressions are unchanged.

matthewevans (approved):
Re-approving the unchanged, reviewed maintainer completion after rebase onto the current merge-queue base.

matthewevans:
Current-head maintainer hold for `ffc15ba3a746c94b0f6aa28e9d680c234f0fd0cc`: this is a clean linear rebase onto current `main`. The direct delta from the reviewed `6b8e9c55` head contains only upstream main files; none of this PR’s nine parser/test files changed. The constrained source-counter rebinding, the source/recipient/prior-target/RHS-counter coverage, and Gemstone Mine/Revelation cast-pipeline regressions remain the approved implementation.

The rebase started a fresh CI run. Rust lint/parser gate and WASM are in progress; Rust-test shards, card-data/parse-diff, Tauri, frontend, and Lobby checks are pending/queued. The visible 17-card parse artifact predates this head, so it is not current-head evidence yet. Auto-merge is disabled until current artifacts complete successfully. No merge-queue action during this hold.

matthewevans:
Correction: the current head is `90b2aea7e347db6f8ae41d1a06ddaffa8eea889d`. It is a clean rebase of the already-approved #6559 implementation onto current `main` (`616ee41e5c`): its PR diff still contains exactly the nine reviewed source-counter parser/test files (947 additions, 6 deletions). The two-file direct delta from the prior rebased head is upstream `main` content, not a new PR change.

Current CI has restarted and is not yet merge evidence; auto-merge is disabled. No merge-queue action until current-head required checks and the current parse artifact complete successfully.

matthewevans (approved):
Re-approving the unchanged maintainer completion after the final rebase onto current main.

matthewevans:
Current-head hold for `b79066e8af6cb582fd0d5fa8076da28fdd4d6ff2`: the new functional delta moves the constrained leading bare-recipient counter-condition rebind to the `ClauseIr` condition before lowering, replacing the post-lowering `GenericEffect`-only rewrite. Hand review confirms the same narrow grammar gate, prior-typed-target guard, source/recipient/RHS recursive traversal, and Revelation-of-Power protection; the final commit only clarifies that ownership in code comments.

Current CI is still running (Rust lint/parser gate, both Rust shards, card data/coverage, and frontend); WASM, Tauri, Lobby, security, contributor trust, and CodeRabbit are successful. The 17-card/5-signature parse artifact predates this head, so it is not current-head evidence yet. Auto-merge is disabled. Do not queue until current CI and parse artifacts complete successfully.

Issue #6508, opened 2026-07-22
Citadel of Pain — damages its controller instead of opponents


matthewevans:
Reported in Discord: https://discord.com/channels/1485498006781427802/1529472248778588171/1529472248778588171

**Thread:** Citadel of Pain — damages its controller instead of opponents
**Discord thread id:** `1529472248778588171`

## Summary
It does not deal the damage to opponents. It only deals damage to the controller.

### citadel of pain
At the beginning of each player's end step, this enchantment deals X damage to that player, where X is the number of untapped lands they control.

---
<sub>report_id: `discord:1529472248778588171:1529472248778588171:0` · discord: `1529472248778588171/1529472248778588171`</sub>

mike-theDude:
## Triage analysis

**Clarified summary**
Citadel of Pain deals its end-step damage only to its controller. Opponents take no damage at their own end steps.

**Oracle text (verified from `client/public/card-data.json`)**
> At the beginning of each player's end step, this enchantment deals X damage to that player, where X is the number of untapped lands they control.

**Investigation findings**
The clause claims support, and one half of it is right: the damage subject is bound to the scoped player, so "that player" resolves correctly per end step.

The damage *amount* is wrong. The parsed quantity counts untapped lands **you** control rather than untapped lands the scoped player controls. In the Oracle text, "they" refers back to "that player" — the same player the damage is dealt to — not to the enchantment's controller. With the count anchored to the controller, an opponent's end step computes a value from the controller's board, and the controller's own end step is the only one that looks correct, which is consistent with what was reported.

The general shape is a possessive pronoun inside a per-player scoped trigger that must resolve to the scoped player rather than to the ability's controller. That is a reusable binding, not a Citadel of Pain special case.

**Acceptance criteria**
- At each player's end step, the damage dealt equals the number of untapped lands that player controls.
- Each player, including opponents, takes their own damage at their own end step.
- The "they/their" binding inside a scoped-player trigger resolves to the scoped player generally.

mike-theDude:
## Classifier analysis

**Verdict**: `supported_aspect_defect`

**Matched clause**: `At the beginning of each player's end step, this enchantment deals X damage to that player, where X is the number of untapped lands they control.`  (supported: true)

[...]

PR #6564, opened 2026-07-23
fix(parser): scope each-player phase-trigger anaphors to the phase player (Citadel of Pain #6508)

jeffrey701:
## Summary

Fixes **Citadel of Pain** (#6508) and the each-player/each-opponent phase-trigger anaphor class: *"At the beginning of each player's end step, this enchantment deals X damage to that player, where X is the number of untapped lands they control."* On an opponent's end step Citadel counted **the controller's** untapped lands instead of the phase player's — so opponents visibly never took the right amount (frequently 0), while the controller took its own count on its own end step.

## Root cause

Parse-time anaphor mis-binding. The each-player phase trigger correctly binds "that player" anaphors to the phase's active player (`ControllerRef::ScopedPlayer`) via `relative_player_scope_for_condition` — which is why the DealDamage **recipient** parsed correctly. But the `where X is … they control` count is stripped as a raw string and interpreted later, at assembly time, through the **context-free** `parse_cda_quantity` — so `relative_player_scope` was `None` and "they control" fell to the legacy `unwrap_or(ControllerRef::You)`, binding the count to the source's controller. The sibling `for-each` interpreter in the same function already defaults this anaphor to `ScopedPlayer`; the where-X CDA arm simply never carried the context.

A second, live gap in the same trigger family: `lower_trigger_ir` had rewrite passes for `TargetPlayer` and `SourceChosenPlayer` scopes but **no `ScopedPlayer` branch**, so possessive quantities in these triggers (`TargetZoneCardCount{Hand}`, `LifeTotal{Target}`) stayed target-marker refs that resolve to **0 at runtime** when the trigger has no player target — Iron Maiden always dealt 0, Rackling always dealt max, Havoc Festival lost 0 life, etc.

## Fix (parser-only, +73/-2 across 3 files)

- **Part A** (`oracle_effect/lower.rs`, `parse_where_x_quantity_expression`): carry the `ScopedPlayer` anaphor context into the CDA-quantity delegate via the existing `for_each_anaphor_context` + `parse_cda_quantity_with_context`, exactly as the sibling for-each interpreter does. `ScopedPlayer` degrades to the source's controller at runtime when no scope is stamped, so spell where-X reads are unchanged; only each-player/each-opponent phase triggers now read the phase player.
- **Part B** (`oracle_trigger.rs`, `lower_trigger_ir`): add the missing `Some(ControllerRef::ScopedPlayer)` rewrite branch, reusing the identical `rewrite_event_player_quantity_refs_to_scoped` the `TargetPlayer` branch already calls — converting the target-marker possessives to `HandSize{ScopedPlayer}` / `LifeTotal{ScopedPlayer}`. It rewrites only `PlayerScope::Target` and `TargetZoneCardCount`, never `Controller`, so a mixed-anaphor card like Dark Suspicions keeps its `-HandSize{Controller}` operand intact.
- Supporting: `for_each_anaphor_context` widened to `pub(crate)`.

Behavior-fixed cards: Citadel of Pain (#6508), Iron Maiden, Viseling, Rackling, Storm World, Wheel of Torture, Dark Suspicions, Dreamborn Muse, Price of Knowledge, Havoc Festival. (Noetic Scales shares the family but embeds its count in a filter's `PtComparison`, which the quantity visitor doesn't descend into — documented residual, separate follow-up.)

## Files changed

- `crates/engine/src/parser/oracle_effect/lower.rs` — Part A (+ where-X unit test)
- `crates/engine/src/parser/oracle_trigger.rs` — Part B
- `crates/engine/src/parser/oracle_quantity.rs` — `pub(crate)` visibility
- `crates/engine/src/parser/oracle_trigger_tests.rs` — 4 parser SHAPE tests (Citadel, Iron Maiden, Dark Suspicions multi-authority, Havoc Festival, + you-control negative w/ reach-guard)
- `crates/engine/tests/integration/citadel_of_pain_each_player_end_step_6508.rs` — 4 runtime tests
- `crates/engine/tests/integration/main.rs` — mod line

## CR references

- `CR 513.1` (end step), `CR 603.2b` (phase trigger fires), `CR 102.1` (active player = "that player"), `CR 109.5` (why "you control" stays You), `CR 608.2c` (anaphor binding), `CR 503.1a` (Iron Maiden upkeep test), `CR 110.5` (tapped/untapped status).

Method: /engine-implementer

## Track

Developer

## LLM

Model: claude-opus-4-8[1m]
Thinking: high

## Verification

- `cargo test -p engine --lib` — **17596 passed, 0 failed, 6 ignored** (baseline 17590 + 6 new parser tests)
- `cargo test -p engine --test integration` — **3873 passed, 0 failed, 2 ignored** (baseline 3869 + 4 new)
- `./scripts/gen-card-data.sh` — regenerated; parse-diff audit of the family confirms every expected card flipped to the phase-player form and nothing else changed: Citadel amount `ObjectCount{controller: ScopedPlayer}`; Iron Maiden/Viseling `Offset{HandSize{ScopedPlayer}, −4}`; Rackling/Storm World/Wheel of Torture `ClampMin{N − HandSize{ScopedPlayer}}`; Dreamborn Muse/Price of Knowledge `HandSize{ScopedPlayer}`; Havoc Festival `DivideRounded{LifeTotal{ScopedPlayer}, 2}`; Dark Suspicions `Sum[HandSize{ScopedPlayer}, −HandSize{Controller}]` (controller operand intact); The Rack stays `SourceChosenPlayer`, Copper Tablet stays `Fixed`, Ancient Runes unchanged.
- RED/GREEN: T1 (P0=1/P1=3 untapped → asserts P1 −3; pre-fix −1), T2 (P1 lands tapped → 0; pre-fix −2), T4 (Iron Maiden hand 7−4 → −3; pre-fix `TargetZoneCardCount` resolves 0 → 0), plus all shape tests fail on revert. The you-control negative carries a positive reach-guard.

## Gate A

Gate A PASS head=03c3aa6ac6c8ca090fbc05ecfefb047298db2600 base=6ae8737cdab0fa1ed291cad0f0808473a90f4cf8

## Anchored on

- `crates/engine/src/parser/oracle_quantity.rs:3337` — `for_each_anaphor_context` + `parse_cda_quantity_with_context` (oracle_quantity.rs:781): the sibling for-each interpreter that already defaults the third-person "they control" anaphor to `ScopedPlayer`; Part A adopts the same helper at the where-X CDA arm.
- `crates/engine/src/parser/oracle_trigger.rs:1629` — the existing `TargetPlayer`/`SourceChosenPlayer` rewrite branches in `lower_trigger_ir` calling `rewrite_event_player_quantity_refs_to_scoped` / `rewrite_player_quantity_refs_to_source_chosen`; Part B adds the missing `ScopedPlayer` rung to that same ladder (oracle_trigger.rs:1648) reusing the same rewrite fn.

## Final review-impl

Final review-impl PASS head=03c3aa6ac6c8ca090fbc05ecfefb047298db2600

## Claimed parse impact

- Citadel of Pain, Iron Maiden, Viseling, Rackling, Storm World, Wheel of Torture, Dark Suspicions, Dreamborn Muse, Price of Knowledge, Havoc Festival (behavior-fixed); Curious Herd, Jovial Evil, Pact of the Serpent, Inscription of Abundance (label-only ScopedPlayer normalization, runtime-identical for spells).

## Validation Failures

None.

## CI Failures

None.

---

Tier: Frontier

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected third-person anaphor and `where-X` player-scope binding in per-player triggers so phrases like “they,” “their,” and “that player” resolve to the scoped phase player instead of an incorrect default.
  - Ensured “you” and “your” remain tied to the ability or controller context, preserving accurate hand, life, and damage calculations.

- **Tests**
  - Added unit and integration coverage for per-player trigger parsing, including Citadel of Pain end-step and Iron Maiden upkeep damage scenarios.


matthewevans:
_Generated for head `5eafb2cb1828fa9d4192a67234b2471458c3dd51`._

### Parse changes introduced by this PR · 10 card(s), 8 signature(s)  (baseline: main `58363f3a210f`)

#### 🟡 Modified fields (8 signatures)

- **2 cards** · 🔄 ability/DealDamage · changed field `amount`: `max(-1*target zone card count+3, 0)` → `max(-1*cards in hand (scoped player)+3, 0)`
  - Affected (first 3): Rackling, Wheel of Torture
- **2 cards** · 🔄 ability/DealDamage · changed field `amount`: `target zone card count+-4` → `cards in hand (scoped player)+-4`
  - Affected (first 3): Iron Maiden, Viseling
- **1 card** · 🔄 ability/DealDamage · changed field `amount`: `# of untapped you control land` → `# of untapped scoped player controls land`
  - Affected (first 3): Citadel of Pain
- **1 card** · 🔄 ability/DealDamage · changed field `amount`: `max(-1*target zone card count+4, 0)` → `max(-1*cards in hand (scoped player)+4, 0)`
  - Affected (first 3): Storm World
- **1 card** · 🔄 ability/DealDamage · changed field `amount`: `target zone card count` → `cards in hand (scoped player)`
  - Affected (first 3): Price of Knowledge
- **1 card** · 🔄 ability/LoseLife · changed field `amount`: `(target zone card count + -1*cards in hand (you))` → `(cards in hand (scoped player) + -1*cards in hand (you))`
  - Affected (first 3): Dark Suspicions
- **1 card** · 🔄 ability/LoseLife · changed field `amount`: `divide(life total (target player), 2, rounded up)` → `divide(life total (scoped player), 2, rounded up)`
  - Affected (first 3): Havoc Festival
- **1 card** · 🔄 ability/Mill · changed field `count`: `target zone card count` → `cards in hand (scoped player)`
  - Affected (first 3): Dreamborn Muse

matthewevans:
**Part B is clean and at the right seam — Part A is not. `parse_where_x_quantity_expression` hard-codes `ScopedPlayer` instead of threading the caller's `ParseContext`, and the new parse-diff proves the cost: three cards labelled "runtime-identical" are actually target-player anaphors moved from one wrong referent to another.**

Reviewed at head `03c3aa6a` (unchanged since the last look). Thank you for the parse-diff — it's exactly the evidence that made this reviewable, and it's what turned a predicted harm into a measured one.

## 📊 Parse-diff verdict: claimed 14, measured 13

Sticky comment `5063481753` (baseline `21a53d50cbcb`): 13 cards / 12 signatures, all "Modified fields", none gained, none lost.

**All 10 claimed behavior-fixed cards check out** — each Oracle text verified against Scryfall: Citadel of Pain, Iron Maiden, Viseling, Rackling, Wheel of Torture, Storm World, Dark Suspicions, Dreamborn Muse, Price of Knowledge, Havoc Festival. Dark Suspicions correctly retains its `-1*cards in hand (you)` controller operand.

**Unexplained #1 — Inscription of Abundance** is claimed but absent from the measurement. Its clause is an aggregate ("where X is the greatest power among creatures they control"), not an `ObjectCount`, so it plausibly never reaches the touched arm. The sticky's trailing "2 card(s) had Oracle-text changes (errata/reprint) — excluded" bucket is unnamed, so this can't be reconciled from the sticky alone. Please account for it.

**Unexplained #2 — the 3 "label-only" cards are target-player anaphors** (Scryfall-verified):

| Card | Oracle | Change |
|---|---|---|
| Jovial Evil | "deals X damage to **target opponent**, where X is twice the number of white creatures **that player** controls" | `2*# of white you control creature` → `2*# of white scoped player controls creature` |
| Curious Herd | "Choose **target opponent**. You create X … where X is the number of artifacts **that player** controls" | `# of you control artifact` → `# of scoped player controls artifact` |
| Pact of the Serpent | "**Target player** draws X cards and loses X life, where X is the number of creatures **they** control…" | same shift, 2 signatures (Draw + LoseLife) |

The correct referent for all three is `ControllerRef::TargetPlayer`. Calling this "runtime-identical for spells" understates it — it swaps one wrong binding for a different wrong binding.

## 🔴 Blockers

**[HIGH] `parse_where_x_quantity_expression` hard-codes `ScopedPlayer` rather than threading `ParseContext`.**

`crates/engine/src/parser/oracle_effect/lower.rs:8145-8161` injects `for_each_anaphor_context(&ParseContext::default(), &ControllerRef::ScopedPlayer)` unconditionally, across ~30 non-test call sites (`mana.rs:1071`; `token.rs:118/683/727`; `imperative.rs:8429/9958`; `counter.rs:239`; 32 refs in `lower.rs`).

The sibling interpreter already does this correctly: `crates/engine/src/parser/oracle_quantity.rs:3337` `parse_for_each_clause_with_context` takes `&ParseContext` and threads whichever scope the caller holds. And the **pre-existing** test `for_each_they_control_threads_target_player` (`oracle_quantity.rs:6258`, Burden of Greed) proves `TargetPlayer` binding already works today. The machinery to bind those three cards correctly exists; this change routes around it.

Suggested fix: add `parse_where_x_quantity_expression_with_context(expr, ctx)`, thread `effect_ctx` from the where-X binding path, and leave the context-free signature delegating with `ParseContext::default()`.

**[MED] The "runtime-identical" justification cites the wrong resolver.**

`lower.rs:8149-8151` cites `scoped_player_or_controller`. Two functions share that name. The one the comment *describes* — `crates/engine/src/game/quantity.rs:3842` — is `ability.scoped_player.unwrap_or(controller)`, a clean degrade. But this change's actual output for Citadel of Pain is `ObjectCount { filter: Typed { controller: Some(ScopedPlayer) } }`, whose runtime authority is `crates/engine/src/game/filter.rs:1154` → `filter.rs:780`, and that path inserts an extra rung `.or_else(|| triggering_event_player(state))` **before** `source_controller`, reading `state.current_trigger_event` / the `DETECTION_TRIGGER_EVENT` TLS (`quantity.rs:1491`).

Confidence: high that the comment is wrong about its own code path; medium that the divergence is reachable (`current_trigger_event` is save/restore-scoped at `game/engine.rs:7915/7921` and `game/triggers.rs:152/168`, but `game/combat.rs:13571` assigns `AttackersDeclared` with no paired restore in that block). Either fix the citation and prove the spell claim, or drop the claim.

**[MED] Fixture path-divergence.** All 4 runtime tests (`tests/integration/citadel_of_pain_each_player_end_step_6508.rs`) and all 4 shape tests (`oracle_trigger_tests.rs`) are phase triggers, so `ability.scoped_player` is always `Some`. The `None` arm at `filter.rs:780` — the arm all three spell cards take in production — has **zero** fixtures.

**[MED] Claimed parse impact (14) ≠ measured (13).** Reconcile Inscription of Abundance.

## ✅ Clean — verified

- **Part B is at the right seam and needs no changes.** `parser/oracle_trigger.rs:1639-1652` adds the missing `ScopedPlayer` rung to the existing `TargetPlayer`/`SourceChosenPlayer` ladder in `lower_trigger_ir` and reuses the pre-existing `rewrite_event_player_quantity_refs_to_scoped` (`oracle_effect/mod.rs:24608`). Verified at `mod.rs:24619-24639` that it rewrites only `PlayerScope::Target` and `TargetZoneCardCount`, never `Controller` — so Dark Suspicions survives, which the parse diff confirms. The Rack mutual-exclusion argument holds.
- **The class is real** — 10 Scryfall-verified cards, well above the 3-card bar.
- **Nom mandate satisfied** — no `find()`/`split_once()`/`contains()`/`starts_with()` parsing dispatch, no verbatim Oracle string match anywhere in the diff.
- **All 9 CR numbers grep-verified** in `docs/MagicCompRules.txt`, and each describes the code it annotates: 109.5, 608.2c, 603.2b, 102.1, 513.1, 503.1a, 110.5, 107.3i, 115.1.
- **Runtime tests genuinely discriminate.** T1 is asymmetric (P0=1/P1=3; revert flips −3→−1), T2 (P1 all tapped; revert flips 0→−2), T4 (Part B, 7−4=3 vs pre-fix 0). They drive `GameScenario`/`GameRunner` → `advance_to_end_step` → `advance_until_stack_empty` and assert life deltas rather than AST shape. `IRON_MAIDEN_ORACLE` matches Scryfall verbatim.

## 🟡 Non-blocking

- The PR body checks `- [x] Required checks ran clean.` while the required aggregator is red. The root cause is maintainer-side, so this isn't misreporting — but please keep the template honest.
- The triage packet reports `anchors: []` despite a well-formed "## Anchored on" section with two valid anchors; worth a look at why it isn't being picked up.

## Recommendation

**request-changes, contributor round-trip — not a maintainer fixup.** Threading `ParseContext` through ~30 call sites is a design change, not a small local correction. Please:

1. Thread the caller's `ParseContext` through `parse_where_x_quantity_expression` (this is the same fix flagged previously, now backed by three named misbound cards).
2. Fix the `filter.rs:780` citation, and either prove or drop the "runtime-identical for spells" claim.
3. Add a fixture that reaches the `None` arm.
4. Reconcile claimed 14 vs measured 13.

**Part B is clean — keep it as is.**

Your CI red is **not your fault** and cannot be fixed by rebasing, because main's tip *is* the broken commit: `release: v0.35.2` bumped `client/src-tauri/Cargo.toml` without updating the separate `client/src-tauri/Cargo.lock`, redding `Tauri compile check` and through it the required Rust aggregator on every open PR. #6568 fixes it on main.

matthewevans (changes requested):
Current head cb0332348cf2110cc23552751310cc3f0cdb18e1 remains blocked.

parse_where_x_quantity_expression still creates a default ParseContext and unconditionally installs ScopedPlayer instead of receiving and threading the caller's context. Consequently target-player anaphors such as Revelation of Power's sibling class are rebound to ScopedPlayer rather than TargetPlayer.

Add the context-aware entry point, thread the existing effect context through its callers, retain the default wrapper only for genuinely context-free callers, and cover a target-player where-X case before requesting review again.

jeffrey701:
Done at head `2a97ab924` — `parse_where_x_quantity_expression` no longer creates a default context or installs `ScopedPlayer`; it receives and threads the caller's context.

## What changed

- **`parse_where_x_quantity_expression_with_context(where_x_expression, ctx: &ParseContext)`** is the real interpreter now; **`parse_where_x_quantity_expression(where_x_expression)`** is a thin wrapper over it with `ParseContext::default()`, kept only for the genuinely context-free callers (the boundary/`is_some()` probes in `structurally_bound_where_x_clause`, the `Variable("X")`-shape checks, etc.).
- The CDA arm binds the anaphor to **`ctx.relative_player_scope`** when a scope is stamped, and takes the exact pre-#6508 `parse_cda_quantity` path (legacy caster-relative `You`) when it is not — never a hardcoded `ScopedPlayer`.
- Because the where-X string is stripped to the IR and interpreted at the **context-free assembly walk**, the parse-time scope is captured into a new `ClauseIr.where_x_scope` and rebuilt into a `ParseContext` at the top of `apply_where_x_ability_expression`, then threaded through the `apply_/bind_where_x_*` chain (and the recursive "N plus/minus <where-x>" self-call). The inline ctx-bearing callers (`token.rs`, `imperative.rs`) call the `_with_context` entry point directly.

Result:
- **Each-player / each-opponent phase triggers** (Citadel of Pain, Iron Maiden, Rackling, Dark Suspicions, Havoc Festival, Price of Knowledge) — the trigger's `ScopedPlayer` scope threads in, so the count/amount stays `ScopedPlayer`, unchanged.
- **Targeted spells** — no longer rebound to `ScopedPlayer`. With no stamped scope they retain the legacy caster-relative `You` binding, byte-identical to `main`. When a caller *does* supply a `TargetPlayer` scope, the anaphor binds `TargetPlayer` — this is exactly what the new unit test `where_x_that_player_controls_binds_caller_scope_not_scoped_player` covers (a `TargetPlayer`-scoped context yields a `TargetPlayer` count; the scope-free wrapper does not force `ScopedPlayer`).

## Why I did not auto-derive `TargetPlayer` from a spell's own player target

I tried that first (derive `TargetPlayer` when the clause's effect targets a player). It regressed the CR 107.3i single-value-of-X invariant on the very cards it was meant to fix: a spell's "that player" is frequently **cross-clause** ("Choose target opponent. You create X … artifacts *that player* controls" — Curious Herd) or **shared across sibling sub-effects** ("Target player draws X *and loses X*, where X is … creatures *they* control" — Pact of the Serpent). The single-clause assembly view can only see one leg, so the derivation bound Pact's `Draw.count` to `TargetPlayer` but left the sibling `LoseLife.amount` at `You` — one X, two different players. Binding a chosen target forward to a later clause's / sibling's anaphor is a real carry-forward change, distinct from this PR's scope, so I've left those cards on `main`'s `You` (not the `ScopedPlayer` regression) and can take the carry-forward as a follow-up if you'd like it in this class.

## Verification

- Full `cargo test -p engine --lib` and `--test integration` green.
- Regenerated card-data: Citadel of Pain counts the phase player's untapped lands (`ScopedPlayer`); the Iron Maiden family stays `ScopedPlayer`; no targeted spell binds `ScopedPlayer`; non-anaphor where-X ("that card's mana value", "creatures you control") is unchanged.
- New target-player unit test + the updated `where_x_they_control_binds_scoped_player` (now exercises the `_with_context(ScopedPlayer)` path — the old assertion tested the hardcoded install you asked removed); `shared_where_x_binds_draws_and_loses_life_siblings` (Pact) is green with both legs consistent.
- Parser-combinator gate (Gate A) passes; branch up to date with current `main`.

matthewevans:
Maintainer update for current head `578035eb6de7be863fd3b6ec3281e25a689c73a5`: the documentation-only fixup is pushed. This fresh head is held pending Card data's current parse-diff evidence and the remaining Rust CI checks. Approval/enqueue will resume once they settle.

matthewevans:
Maintainer CI fixup is pushed at current head `1fcf74d014ee7539e6645a6f9e01179330ec07f3`: it resolves the reported `collapsible_match` lint and updates the six affected IR snapshots for the new `where_x_scope` field. The next step is GitHub CI for this exact head; review/enqueue remains held until its Rust lint and test checks complete.

matthewevans:
Maintainer port and native-IR snapshot regeneration are pushed at current head `3524b835b2bd76169b782d75b6ca3695ad7b7576`.

The two snapshot changes were derived from the prior native CI snapshot output (Boseiju and Kroxa), not hand-predicted. Current CI and the current-head parse-diff are now running; this PR remains on hold until those exact-head artifacts settle. The old `CHANGES_REQUESTED` review targets `cb0332348` and is not a current-head finding.

matthewevans:
Maintainer snapshot repair is pushed at current head `cb359e2727a09626c08c4bce6dd10ffeef2d4037`.

This commit updates only the three CI-reported native IR snapshots, using the exact values emitted by run `31501178204`: Kroxa (`ScopedPlayer`), Bonecrusher Giant (`TriggeringPlayer`), and Boseiju, Who Endures (`ParentTargetController`). The parser-combinator gate and formatting passed locally; no direct test/build run was used.

The PR is held pending CI and the current-head parse-diff artifact. The current CodeRabbit review also has one unresolved context-propagation finding in `lower.rs`; this snapshot-only repair does not resolve or waive that separate review gate.

matthewevans (changes requested):
**Current head has one parser-context blocker — request changes.**

## 🔴 Blocker

`crates/engine/src/parser/oracle_effect/lower.rs:10438-10502` reconstructs `wx_ctx` from only `where_x_scope`; `ClauseIr` likewise preserves only that field at `crates/engine/src/parser/oracle_ir/effect_chain.rs:834-843`. This drops the parse context required by the lowerer's own typed quantity path. In particular, `parse_type_phrase_with_ctx` selects `IsChosenCreatureType` instead of `IsChosenCardType` from `ctx.pending_choice_type` at `crates/engine/src/parser/oracle_target.rs:3285-3299`, but `for_each_anaphor_context` receives the reconstructed default at `crates/engine/src/parser/oracle_quantity.rs:3470-3490`. Thus a where-X count over `cards of the chosen type` following a creature-type choice can be lowered on the card-type axis, even though the surrounding chain had already established the creature-type choice.

The present regression at `crates/engine/src/parser/oracle_effect/tests.rs:4316-4347` cannot detect this: it parses from the default context and uses a `creatures` base, which always selects the creature-type axis. Preserve the semantic context that the quantity/type path consumes across `ClauseIr` -> assembly (rather than reconstructing a default context), and add a discriminator with a preceding creature-type choice plus a `cards of the chosen type` where-X count. It must assert the typed filter axis and fail when the propagated context is removed.

## ✅ Maintainer fixup

The three snapshot-only additions introduced at `cb359e2` were removed at this head because the exact-head CI diff expected `where_x_scope` to be absent when `where_x_expression` is null. CI is rerunning for that maintenance-only repair; it does not resolve the context loss above.

**Recommendation: request changes for context preservation and a discriminating parser/runtime test; then re-review the new head.**

matthewevans:
Correction to the current-head review: its snapshot note was wrong. Exact-head CI run `31511000679` reports that the three `where_x_scope` entries must be present, rather than absent: Kroxa=`ScopedPlayer`, Bonecrusher Giant=`TriggeringPlayer`, and Boseiju=`ParentTargetController`.

Those are maintainer-created snapshot metadata failures and are separate from the parser-context blocker in `lower.rs:10438-10502`. The blocker remains valid: the lowering path reconstructs only `relative_player_scope` and loses context consumed by the type/quantity parser. I am not pushing a snapshot-only commit now because it would invalidate the current formal review while the substantive blocker still requires a new design/test head. The next revision should include the three exact CI snapshot values and preserve the full consumed parse context.

matthewevans:
The current head `5eafb2cb1828fa9d4192a67234b2471458c3dd51` still has the unresolved parser-context blocker from the formal requested-changes review of 2026-08-11: [preserve the full where-X parse context and add its discriminating regression](https://github.com/phase-rs/phase/pull/6564#pullrequestreview-4908337699).

This is an expiry warning under the repository's requested-changes policy. Please address that review in a new commit within 7 days; otherwise this PR will be auto-closed. A new head or substantive follow-up will return it to review.

matthewevans:
Closing this PR because current head `5eafb2cb1828fa9d4192a67234b2471458c3dd51` still has the unresolved parser-context blocker in the [2026-08-11 requested-changes review](https://github.com/phase-rs/phase/pull/6564#pullrequestreview-4908337699), and there has been no author follow-up after the [2026-08-18 expiry warning](https://github.com/phase-rs/phase/pull/6564#issuecomment-5331486517).

A new substantive revision can be opened against current `main` and will receive a fresh review.

PR #6565, opened 2026-07-23
fix(parser): bind "that player controls" to the iterated opponent for per-opponent repeat clauses

### LINKED ISSUE #5994: Riptide Gearhulk asks for a target among your permanents, instead of opponents — The [[Riptide Gearhulk]] etb should on…

jsdevninja:
## Summary
- Fixes #5994: Riptide Gearhulk's ETB ("for each opponent, put up to one target nonland permanent that player controls into its owner's library third from the top.") asked the caster to choose from among their own permanents instead of each opponent's, and fizzled if skipped.
- Root cause: `parse_for_each_opponent_target_fanout_clause` only binds the "that player controls" anaphor to `ControllerRef::TargetPlayer` inside its own scoped clone of the parse context, used solely to decide whether the clause qualifies for folding `repeat_for` into a `MultiTargetSpec`. Effects whose shape doesn't qualify for that fold — including any triggered ability's execute body, which has no multi-target slot at all — fell through to a plain re-parse with the original, unscoped context, silently defaulting "that player controls" to `ControllerRef::You`.
- Fix: set `relative_player_scope = TargetPlayer` for the whole re-parse whenever `repeat_for` resolves to "for each opponent," regardless of which downstream arm ends up producing the clause, restoring the prior scope afterward. This keeps the anaphor bound to the opponent being iterated even when the multi-target fanout declines (as it does for `PutAtLibraryPosition`, and as it structurally must for triggered-ability execute bodies).

## Test plan
- [x] Added `effect_for_each_opponent_put_at_library_position_keeps_repeat_for_and_target_player` in `crates/engine/src/parser/oracle_effect/tests.rs`, mirroring the existing `effect_for_each_opponent_gain_control_uses_per_opponent_target_fanout` precedent, asserting `repeat_for` stays intact and the target filter's controller resolves to `TargetPlayer` (not `You`).
- [ ] CI (`clippy`, `test-engine`) — could not run Tilt or `cargo` locally in this sandbox (no MSVC linker installed for the `x86_64-pc-windows-msvc` target); relying on CI to validate the fix and the new test.

## Summary by CodeRabbit

* **Tests**
  * Added a parser regression test for an oracle effect that, for each opponent, places up to one target nonland permanent they control into its owner’s library third from the top, validating the full parsed structure.
* **Bug Fixes**
  * Improved oracle-effect parsing for “optional per-opponent target” clauses by more accurately detecting optional target quantifiers across the sentence while avoiding misclassification of unrelated quantities.


matthewevans:
**Blocked on the seam, not the code: `ControllerRef::TargetPlayer` has no per-iteration referent under a bare `repeat_for`, so this changes the bug's shape without fixing it. The right fix is one layer up, in the fanout gate — and it would fix the fizzle half of #5994 for free.**

Reviewed at head `3ad7d71f`. Thank you for the verified Oracle text and grep-verified CR citations — those are clean and called out below.

## 🔴 Blockers

**[HIGH] `TargetPlayer` does not bind to the iterated opponent under a bare `repeat_for`.**

`game/effects/mod.rs:8648-8660` resolves `repeat_for` to a plain `i32` count. The loop at `:8714` takes the `else { effective }` branch at `:8728-8759` — per-iteration rebinding exists **only** for member-driven (`TrackedSetSize`/`ObjectCount`) and kind-driven (`DistinctCounterKindsAmong`) repeats. Nothing mutates a scoped player per iteration.

`game/filter.rs:931` and `:1950` then resolve `TargetPlayer` as "the first `TargetRef::Player` in `ability.targets`" (doc at `filter.rs:1943-1949`), and `game/ability_utils.rs:2427-2453` surfaces exactly one companion player slot regardless of `repeat_for`.

Net effect with this AST (`controller: TargetPlayer` + `repeat_for: PlayerCount{Opponent}` + `multi_target: None`): in a 4-player game Riptide Gearhulk lets you pick **one** opponent, then repeats on that **same** opponent's single permanent three times — instead of up to one permanent per opponent. That is the aliasing failure mode the PR set out to remove.

**[HIGH] The stated justification is factually incorrect.**

The comment at `parser/oracle_effect/mod.rs:28957-28962` and the test docstring at `oracle_effect/tests.rs:32077-32084` assert that a triggered ability's execute body has no `multi_target` slot to fold into. Refuted by `parser/oracle_trigger.rs:617` (`if let Some(spec) = def.multi_target.as_ref()`), `:682` (`as_mut()`), and the doc at `:664-670`: *"`multi_target`, `repeat_for` and `target_constraints` are consumed during target selection, before the trigger resolves."* `AbilityDefinition.multi_target` (`types/ability.rs:16276`) is kind-agnostic.

Sibling card **Bronzebeak Foragers** ("When this creature enters, for each opponent, exile up to one target nonland permanent that player controls…") is the same triggered-ETB shape and **does** go through the fanout today.

So the root cause behind the approach was never diagnosed. `Effect::PutAtLibraryPosition { target, .. }` **is** wired into `Effect::target_filter()` (`types/ability.rs:14109`), so the first fanout gate passes — the decline must come from `target_filter_is_single_object_target` (`lower.rs:3575`), and that is where the fix belongs.

**[HIGH] This routes around a deliberate, tested honesty guard.**

`tests/integration/kaya_spirits_justice_per_opponent_exile.rs:9-15` documents: *"The engine has no machinery for a per-iterated-player set of announced target slots (`collect_target_slots` is fixed-count; `repeat_for` resolves at resolution time, not announcement), so the printed-`target` clause is honestly `Effect::unimplemented` rather than silently routed…"* — asserted by `printed_target_form_is_unimplemented` at `:158`. Kaya's "For each other player, exile up to one target creature that player controls" is structurally identical. Stamping `TargetPlayer` converts a deliberate honest red into a silently-wrong result.

## 🟡 Non-blocking

- **Unmeasured blast radius across a 91-card class, with a likely `Choose`-family regression.** The guard at `mod.rs:28969-28983` fires for every chunk whose `repeat_for == PlayerCount{Opponent}`; `parser/oracle_target.rs:4374-4389` then lowers "that player controls" / "they control" to the scope instead of `You`, and that comment flags the fallback as load-bearing for `resolve_quantity_scoped`. `is_per_opponent_target_fanout_clause` (`lower.rs:3560-3566`) deliberately excludes `Choose`/`ChooseCard`/`CopyTokenOf`/`TargetOnly`, which must lower to `ChooseFromZone { ZoneOwner::EachOpponent }`. Cards verified in card-data.json and now at risk: Benthic Anomaly, Chaos Defiler, Decoy Gambit, Disorienting Choice (plus Diluvian Primordial). Please regenerate the parse-diff and confront every gained/lost/changed card before merge.
- **The fix silently reverts the existing fanout path's scope propagation.** `mod.rs:28996-29001` sets `*ctx = fanout_ctx` (scope `Some(TargetPlayer)`), then `:29045-29047` unconditionally restores `prior_relative_player_scope`; since `is_for_each_opponent_repeat` is true on the same precondition the fanout requires, the restore always fires after a fanout success, so later chunks lose the scope they previously inherited. Don't restore when the fanout branch was taken.
- **Parser-shape test only, for a `priority:p2-wrong-game-result` issue.** The sole added test (`oracle_effect/tests.rs:32086`) asserts AST shape. It *is* discriminating at the parser layer, but it asserts `multi_target.is_none()` — pinning the fanout decline as expected, which is the defect. Precedents: `teysa_wojek_investigate_per_opponent.rs` (4-player, real `apply()`, registered at `main.rs:820`) and `kaya_spirits_justice_per_opponent_exile.rs` (3-player).
- **"Fixes #5994" covers only half the report.** The issue says the card "asks for one from you, **and fizzles if skipped**." Nothing here touches the fizzle path. The fanout fold fixes it for free: `MultiTargetSpec::bounded_expr` (`lower.rs:3543`) takes `min` from `stripped_multi_target`, which is 0 for "up to one target."
- **Sibling gap:** the guard matches only bare `PlayerFilter::Opponent` (`mod.rs:28970-28976`), so conditional per-opponent siblings with the same anaphor are untouched (Geth's Summons, Peacekeeper Avatar, Noble Heritage). Please say why those are out of scope.
- `ctx.relative_player_scope.replace(...)` at `mod.rs:28979` clobbers a pre-existing more specific scope (e.g. `ChosenPlayer { index }`) rather than only filling a vacant one.

## ✅ Clean

- **CR 109.4 and CR 608.2c both grep-verified** against `docs/MagicCompRules.txt` (lines 594 and 2793). The `+` compound form is house convention. Not a finding.
- **Riptide Gearhulk's Oracle text verified verbatim from the Scryfall API** — matches the PR body and test input exactly. No fabricated clause.
- **The class is real, not a special case** — 91 cards match "for each opponent" + "that player" (Blatant Thievery, Bronzebeak Foragers, Enigma Thief, Grasp of Fate, Dismantling Wave, Age of Ultron, …).
- **No nom-mandate violations** — no `find()`/`split_once()`/`contains()`/`starts_with()` parsing dispatch, no verbatim Oracle string match introduced.
- Scope save/restore is correctly guarded with no early-return leak.

## Recommendation

**request-changes — rework the approach rather than patching this code.**

1. Diagnose which gate in `is_per_opponent_target_fanout_clause` (`lower.rs:3559-3575`) declines for `PutAtLibraryPosition`. `Effect::target_filter()` already covers it, so it is the `target_filter_is_single_object_target` zone check. Fix **that** so the fanout fold applies — it yields real per-opponent slot pairs (`collect_per_opponent_target_fanout_slots`, `ability_utils.rs:4550`) and fixes the fizzle via `min = 0` in one change.
2. If the fanout genuinely cannot cover the shape, follow the `kaya_spirits_justice_per_opponent_exile` precedent and keep it honestly `Effect::unimplemented`.
3. Add a 3+ player integration test under `crates/engine/tests/integration/` with a `mod` line in `main.rs`, driving the real cast/trigger pipeline.
4. Bring the branch current to regenerate the parse-diff, then confront the `Choose`-family cards named above.
5. Please fill in the `docs/AI-CONTRIBUTOR.md` template gates (Gate A, final review-impl pass, ≥2 anchors, verification section).

Your `Tauri compile check` red is **not your fault**: `release: v0.35.2` bumped `client/src-tauri/Cargo.toml` without updating the separate `client/src-tauri/Cargo.lock`, reddening that job and the required Rust aggregator on every open PR. #6568 fixes it on main.

matthewevans:
### Parse changes introduced by this PR · 4 card(s), 4 signature(s)  (baseline: main `7ed59c04d41b`)

#### 🟡 Modified fields (4 signatures)

- **1 card** · 🔄 ability/DealDamage · changed field `targets`: `1-# of each opponent` → `0-# of each opponent`
  - Affected (first 3): Tri-Sentinel, Act of Vengeance
- **1 card** · 🔄 ability/PutAtLibraryPosition · changed field `targets`: `1-# of each opponent` → `0-# of each opponent`
  - Affected (first 3): Riptide Gearhulk
- **1 card** · 🔄 ability/PutCounter · changed field `targets`: `1-# of each opponent` → `0-# of each opponent`
  - Affected (first 3): Shy Town
- **1 card** · 🔄 ability/base power 0, base toughness 4, grant Defender, add type artifact, add type cre… · changed field `targets`: `1-# of each opponent` → `0-# of each opponent`
  - Affected (first 3): Welcome to . . .

_2 card(s) had Oracle-text changes (errata/reprint) — excluded as non-parser._

matthewevans:
**You reverted the fix and pushed a `panic!` trace to read the parse out of CI — here is the answer, so you don't have to spend the round-trip.** Reverting rather than defending the approach was the right call.

## What your trace would have printed

Riptide Gearhulk's ETB execute, read from a `card-data.json` regenerated today off current `main`:

```
PutAtLibraryPosition {
  target: Typed { type_filters: [Permanent, Non(Land)], controller: "TargetPlayer", properties: [] },
  count: Fixed(1),
  position: NthFromTop { n: 3 }
}
multi_target: { min: 1, max: Ref(PlayerCount { filter: Opponent }) }
```

**There is no `repeat_for` field on this card**, and `controller` is already `TargetPlayer`. So the premise the fix was built on — that the fanout declines for this card and falls back to `ControllerRef::You` — does not hold: the per-opponent fanout *does* fire here. The only site in the workspace emitting `multi_target.max = PlayerCount{Opponent}` is `parse_for_each_opponent_target_fanout_clause` (`crates/engine/src/parser/oracle_effect/lower.rs:3543`), so that path is demonstrably what produced this shape.

That is also why the test you added asserted the opposite shape (`repeat_for.is_some()`, `multi_target.is_none()`) for the same sentence — it was pinning a parse that production does not produce.

## Where the real fix lives

This is consistent with the review above: the decline you were chasing is not in the anaphor binding, it is one layer up in the fanout gate at `target_filter_is_single_object_target` (`lower.rs:3575`). `Effect::PutAtLibraryPosition { target, .. }` **is** wired into `Effect::target_filter()` (`crates/engine/src/types/ability.rs:14109`), so the first gate passes and the second is where behavior diverges.

Worth knowing: fixing it there also closes the second half of #5994 for free. The issue reports that the card "asks for one from you, **and fizzles if skipped**" — that fizzle is the `min: 1` above. `MultiTargetSpec::bounded_expr` (`lower.rs:3543`) takes `min` from `stripped_multi_target`, which is `0` for "up to one target", so the "up to" half is currently being dropped by `per_opponent_target_fanout_min` (`lower.rs:3591-3604`), which gates min-0 detection behind a `tag("gain control of ")`. That is a pre-existing build-for-the-class gap on `main`, not something you introduced.

## Two housekeeping notes

- **The scratch test cannot merge.** `scratch_riptide_gearhulk_trace` in `crates/engine/src/parser/oracle_effect/tests.rs` ends in `panic!("{:#?}", def)`, which reds the test shards for everyone. Please drop it in the next push — and for future tracing, a failing `assert_eq!` prints the same `{:#?}` payload without taking the suite down.
- **Heads-up on vocabulary convergence.** #6564 is fixing the structurally identical per-iteration anaphor problem in the same files, and it binds to `ControllerRef::ScopedPlayer` (with a registered integration test at `crates/engine/tests/integration/citadel_of_pain_each_player_end_step_6508.rs`). `ControllerRef` carries both `ScopedPlayer` and `TargetOpponent` (`types/ability.rs:3479` and `:3497`) with different runtime reads. If you re-approach this, please align with whichever variant #6564 settles on rather than introducing a third binding for one class.

**Recommendation: drop the `panic!` scratch test, then either re-target the fix at the fanout gate (`lower.rs:3575`) with a registered integration test that exercises the companion player slot in a 3+ player game, or close this and let #6564 establish the vocabulary first. Happy to review either direction.**

matthewevans:
**Reviewed at head `fe6509fef`. The seam blocker is resolved — this is the right layer now.** One blocker left (runtime coverage), plus three comment/scope corrections.

You reverted the `ControllerRef::TargetPlayer` stamping entirely — the net diff against `main` is two files, `lower.rs +26/-7` and `tests.rs +56/-0`, with zero lines touched in `oracle_effect/mod.rs`. That also moots both non-blocking items from my 21:59 comment (the scope restore at `mod.rs:29045` and the `ChosenPlayer` clobber at `mod.rs:28979`); that code is no longer in the diff.

**And the fix is load-bearing through a modeled path, not a side channel** — this is the part I checked hardest, because a parser-layer integer that nothing downstream reads would be inert while its test still passed:

`MultiTargetSpec.min == 0` → `AbilityDefinition::targeting_is_optional()` (`types/ability.rs:21862-21868`, which is `optional_targeting || multi_target.min_is_fixed_zero()`) → `collect_per_opponent_target_fanout_slots` pushes the object slot with `optional: ability.targeting_is_optional()` (`game/ability_utils.rs:4582`). That is the announce-time decline — the "fizzles if skipped" half of #5994. Note it works *because* `targeting_is_optional()` reads the `multi_target` min in addition to the `optional_targeting` flag; had it read only the flag (which your own test asserts stays `false`), this identical diff would have done nothing. Good pivot.

## 🔴 Blocker — no runtime test for a `p2-wrong-game-result` fix

The only test added is a parser AST assertion at `parser/oracle_effect/tests.rs:32086`, and `crates/engine/tests/` is untouched. The parser shape is not where this fix pays off — the entire behavior change lives past the parser, in `targeting_is_optional()` → the `optional: true` slot at `ability_utils.rs:4582`. Nothing in the PR proves a player can actually decline a per-opponent slot at announce time.

Please add an integration test under `crates/engine/tests/integration/` with a `mod` line in `tests/integration/main.rs` (an unregistered file compiles to nothing and shows green): 3+ players, assert the slot can be declined and the ability still resolves for the remaining opponents. Precedent to follow: `teysa_wojek_investigate_per_opponent` (`main.rs:820`) or `kaya_spirits_justice_per_opponent_exile` (`main.rs:624`).

## 🟡 Non-blocking — three corrections to the gate comment and scope

**1. The comment enumerates verb shapes the detector doesn't affect** (`lower.rs:3594-3596`). It advertises unifying `"gain control of up to one target …"`, `"exile up to one target …"`, and `"put up to one target …"`. Measured, only `put` reaches this code: `"exile"` is in `MULTI_TARGET_VERBS` (`lower.rs:5569` — `["exile", "tap", "untap", "goad", "detain", "return", "destroy", "choose"]`), so those clauses take their min from `stripped_multi_target` at `lower.rs:3544-3545` and never fall through here; `"gain control of"` was already `0` via the old prefix. State the real delta instead — the fallback fires only for verbs *outside* `MULTI_TARGET_VERBS` — or the next contributor will mis-scope the helper.

**2. `"any number of target …"` is claimed but unhandled** (`lower.rs:3592`). `strip_optional_target_prefix` (`lower.rs:5704-5726`) early-returns `(text, None)` unless `tag("up to ")` matches; the `"any number of"` arm lives in `strip_leading_quantifier` (`lower.rs:5754`), which this path doesn't call. Unreachable today — no card in the class uses that form with `that player controls` — but if it ever is reached, the clause silently gets `min: 1` and forces a mandatory target on a card that says it's optional. Either drop the claim or add the arm with a test.

**3. A second card changes and isn't enumerated.** The PR body, the comment, and the test all name only Riptide Gearhulk. **Shy Town** flips too. Verified independently against `card-data.json`: Shy Town carries **two** `multi_target` sites (its "whenever you planeswalk here" and "whenever chaos ensues" triggers lower separately), both currently `min: 1`; Riptide Gearhulk carries one. So the real delta is **2 cards / 3 clauses**, not one card.

Oracle text re-verified verbatim against Scryfall for both, and both are `put` clauses, consistent with finding 1:
- Riptide Gearhulk — *"When this creature enters, for each opponent, put up to one target nonland permanent that player controls into its owner's library third from the top."*
- Shy Town — *"Whenever you planeswalk here and whenever chaos ensues, for each opponent, put a shy counter on up to one target creature they control."*

Both flips are correct per **CR 115.6** (`docs/MagicCompRules.txt:859`): *"A spell or ability that requires targets may allow zero targets to be chosen. Such a spell or ability is still said to require targets, but that spell or ability is targeted only if one or more targets have been chosen for it."* That is the rule for `"up to one target"` permitting zero, and it's the repo convention here (41 sites). Enumerate Shy Town in the body — an unenumerated behavior change is exactly what the parse-diff gate exists to catch.

**4. The title now describes the approach you reverted.** It still reads *"bind 'that player controls' to the iterated opponent for per-opponent repeat clauses"*; the head does no binding at all. Retitle to the min-0 generalization. (CodeRabbit's title warning is now correct, for a different reason than it gave.)

**5.** `lower.rs:3606` lowercases, then `strip_optional_target_prefix` lowercases again at `lower.rs:5705`. Harmless — the remainder is discarded — but it earns nothing.

## ✅ Clean

- **The `panic!` scratch test is fully gone.** `scratch_riptide_gearhulk_trace` does not appear at head; `87df23da0` removed it and `fe6509fef` did not reintroduce it. The remaining `panic!`s in `tests.rs` are the idiomatic `other => panic!("expected …, got {other:?}")` match arms, matching the adjacent pre-existing `effect_for_each_opponent_gain_control_…` test.
- **Nom mandate: pass.** You used `nom_primitives::scan_at_word_boundaries` (`oracle_nom/primitives.rs:897`) — the shared word-boundary primitive CLAUDE.md mandates for multi-position phrase matching — wrapping `strip_optional_target_prefix`, itself built from `tag()`/`alt()`. No `find()`/`split_once()`/`contains()`/`starts_with()` parsing dispatch introduced, and the removed `nom_on_lower` + `tag()` are still used elsewhere in the file, so no unused imports.
- **The over-fire safety argument in your comment is correct, and I traced its own counterexample.** For `"put up to three +1/+1 counters on target creature"`, `strip_optional_target_prefix` consumes `"up to three "` then requires `"target "`/`"other target "`/`"another target "` immediately (`lower.rs:5715-5724`), sees `"+1/+1 counters…"`, and declines; no later word boundary starts with `"up to "`. The scan is bounded further by `is_per_opponent_target_fanout_clause` requiring a single battlefield object target (`lower.rs:3569-3577`). Widening a scan from a prefix-anchored match to every word boundary is the kind of change that usually needs a bound argument and doesn't get one — yours does, and it holds.
- **The parser test is discriminating.** Revert `per_opponent_target_fanout_min` to returning 1 and `assert_eq!(def.multi_target, Some(MultiTargetSpec::bounded(0, …)))` at `tests.rs:32095` fails. It asserts a positive shape with a reach guard on the `PutAtLibraryPosition` variant rather than being a vacuous negative.

## 📋 On the parse-diff evidence

**The posted sticky is stale and describes the mechanism you reverted** — comment `5064132049` was last updated 22:41:36 against a 00:05:16 head, and every Rust/card-data check at `fe6509fef` is still `queued`. Its content (1 card, `ability/CopyTokenOf` field `copies`, Elminster's Simulacrum) is structurally impossible for the current head. That's why the numbers above are measured directly from `card-data.json` rather than read off it.

One caveat for when it does regenerate: **an empty parse-diff would not mean the fix is inert.** The 22:24 run compared an `ability/CopyTokenOf` *field*; if `multi_target` isn't in the compared signature, a real min-1→min-0 flip can produce a zero-card diff. The 2-card / 3-clause delta above is the ground truth.

---

**Recommendation: add the integration test, then this merges.** The seam question is settled and settled correctly — the remaining items are the comment's scope claims, naming Shy Town, and the title, all of which can ride in the same push as the test.

matthewevans:

matthewevans:

matthewevans (changes requested):
This is good building-block work and it's one small comment fix away from an approval. Green CI, clean merge.

## What's right

- **The generalization is the correct move.** Replacing the hardcoded `"gain control of "` prefix with a word-boundary scan (`scan_at_word_boundaries` + `strip_optional_target_prefix`) makes the min-0 detection cover the whole per-opponent-fanout verb class ("put", "exile", …) instead of one verb. That's building-for-the-class, not the card.
- **The safety property is sound.** Reusing `strip_optional_target_prefix` rather than the bare `strip_leading_quantifier` means the detector only fires on a quantifier immediately followed by `"target "`/`"other target "`/`"another target "`, so it can't misread a resource-count quantifier ("put up to three +1/+1 counters on target creature" — the count modifies the counters, not the target). The comment calling this out explicitly is exactly the right instinct.
- **Rules-correct.** "up to one target … for each opponent" resolving to `min: 0` is CR 601.2c — a player choosing targets for an "up to" requirement may choose fewer, including zero.
- **Parse-diff is clean and scoped:** 4 cards flip `1 → 0` (Tri-Sentinel, Act of Vengeance, Riptide Gearhulk, Shy Town, Welcome to . . .), all genuine "up to one target" per-opponent fanouts. No collateral.

## One thing to fix before I approve

The doc comment on `per_opponent_target_fanout_min` claims the scan recognizes:

> an "up to N target …" / **"any number of target …"** quantifier anywhere in the clause

but the detector calls `strip_optional_target_prefix`, which only matches `tag("up to ")`. It does **not** handle `"any number of "` — that path lives in the wrapper above it (via `multi_target_for_distribute_among`), which this function doesn't call. So an "any number of target … for each opponent" clause would currently fall back to `min: 1`, contradicting the comment.

No card needs the "any number of" per-opponent shape today (the parse-diff confirms none), so the minimal correct fix is to **narrow the comment** to the grammar you actually cover ("up to N target …"). If you'd rather have the coverage, add an `"any number of "` branch to the detector using the same `MultiTargetSpec` plumbing — but that's optional; the comment narrowing is enough to make this accurate.

Fix that and I'll approve. (Credit to the CodeRabbit thread that first flagged the overclaim — I verified it against `strip_optional_target_prefix` directly and it's correct.)

matthewevans (approved):
Maintainer approval: current head 17372095 fixes the optional per-opponent target seam with discriminating coverage; review feedback and required CI are green.

PR #6569, opened 2026-07-23
feat(engine): implement CR 710 Kamigawa flip cards

real-venus:
## Summary

Adds `Effect::FlipPermanent`, implementing the CR 710 flip-card mechanic. **19 of the 21** printed flip cards now lower their flip instruction to a real effect instead of `Unimplemented` (Bushi Tenderfoot, Budoka Gardener, Nezumi Shortfang, Kitsune Mystic, Akki Lavarunner, the five Saviors "Ascendant" cards, …). Their alternative faces already parsed, so for most of them this was the only gap.

## Why a separate slot from `Transform`

Flipping is deliberately its own engine path, per the CR:

- **CR 710.1b** — the alternative name / type line / P&T / text box apply only while the permanent is on the battlefield and flipped.
- **CR 710.1c** — a flip card's color and mana cost do **not** change. `game/flip.rs` therefore has its own `apply_flipped_face_to_object` rather than reusing the double-faced applicator (which swaps them). This is enforced on the flip path, the save/load rehydrate path, and by a guard that stops `Effect::Transform` / `GameAction::Transform` from ever running the DFC applicator on a flip card.
- **CR 710.2 / 710.4** — flipping is one-way, and a flipped permanent that leaves the battlefield reverts to its normal half (CR 110.5 new object).
- **CR 707.2** — a copy of a flipped permanent copies the unflipped normal half.

It also handles the interactions that giving flip cards a back face exposes: turning a flipped permanent face down no longer clobbers its stashed normal half, and the double-faced-card indicator (engine DTO + client) no longer fires for flip cards.

`GameEvent::Flipped` is emitted for the log/UI but intentionally dispatches no trigger key — no printed card triggers on flipping.

## Scope

**Out of scope, documented in `flip.rs`:** CR 710.5 (choosing the alternative name for "choose a card name"), Homura's enters-flipped rider, and two second-order copy/face-down cases that would need a second face-stash slot on `GameObject`.

## Testing

18 tests: per-rule coverage of 710.1b/1c/2/4, the transform and face-down interactions, copy semantics, save/load rehydrate, the "flip it" pronoun form end to end, and a corpus test pinning all 19 cards. The CR 710.1c assertion is mutation-verified (swapping in the DFC applicator fails exactly the right tests). CR 705.1 ("flip a coin") is unaffected.

`cargo clippy --workspace --all-targets -D warnings` clean; `cargo test -p engine` green (17489 lib + 3692 integration, 0 failed); 34 CR numbers verified against the Comprehensive Rules.

## Summary by CodeRabbit

* **New Features**
  * Added full Kamigawa “flip” permanent support, including oracle parsing/lowering, dedicated resolution, and accurate flip events and log/public-state rendering.
* **Bug Fixes**
  * Flip permanents now apply alternate characteristics correctly (one-way), preserve mana cost and color, avoid transform/double-faced behavior, don’t generate trigger/indexing keys, and restore normal state when leaving the battlefield or during copy/reload.
* **UI Improvements**
  * Updated the alternative-face (DFC) button logic to use flip-specific layout rules so the UI reflects the correct face behavior.
* **Tests**
  * Added comprehensive end-to-end integration coverage for flip cards and pronoun targeting.


matthewevans:
**The CR 710 research here is the strongest I have seen from a contributor on this repo — every rule the implementation leans on checks out verbatim, and the flip-vs-transform separation is the right architectural call. Blocked on one red required gate plus two one-line latent-correctness gaps.**

Reviewed at head `359e3d52`.

## 🔴 Blocker

**`Rust lint (fmt, clippy, parser gate)` is red on your own added lines.** The *Engine authority gate* (`scripts/check-engine-authorities.sh`) rejects four raw keyword queries in `crates/engine/src/game/flip.rs`:

```
flip.rs:480   assert!(obj.keywords.contains(&Keyword::DoubleStrike));
flip.rs:742   !obj.keywords.contains(&Keyword::Flying),
flip.rs:750   obj.keywords.contains(&Keyword::Bushido(1)),
flip.rs:872   !copy.keywords.contains(&Keyword::DoubleStrike),
```

Use `keywords::has_keyword(obj, &kw)`, or annotate `// allow-raw-authority: <reason>` where the raw read is genuinely structural. The PR body says "clippy clean, tests green" — both are true, and `cargo clippy`, `cargo fmt`, both Rust test shards, WASM, card-data, and frontend are all green. This gate is a separate step that `clippy` does not cover, so please run `./scripts/check-engine-authorities.sh origin/main` locally before the next push.

**`Tauri compile check` red is ours, not yours** — the v0.35.2 release left `client/src-tauri/Cargo.lock` stale against `--locked`. Fixed on `main` in #6568; it will clear when you rebase. Not attributed to you.

## 🟡 Non-blocking

**1. `flip_permanent` doesn't gate on its own declared authority.** `flip.rs:139` returns early on `obj.back_face.is_none()`, but the doc on `is_flip_permanent` (`:69`) declares that function "the single authority for 'is this a flip card?'". If `Effect::FlipPermanent`'s `ParentTarget` arm ever binds a transform DFC, the flip applicator installs its back face without the CR 712 cost/color swap, sets `flipped`, and `stash_flip_face` re-stamps `LayoutKind::Flip` on the stash — permanently blocking `transform::transform_permanent` on that object. Unreachable across today's corpus; one line to close: `if !is_flip_permanent(obj) { return Ok(()); }`.

**2. `flipped_normal_copiable_values` ignores `face_down`, breaking CR 708.2a.** `flip.rs:213` guards only on `!obj.flipped`. Your own `turn_face_down.rs` deliberately keeps the *normal* half in `back_face` for a flipped permanent (`Some(flip_stash) if obj.flipped => flip_stash.clone()`) and never clears `flipped` — which is the correct call for the zone-exit path, and the comment explaining it is excellent. But it means a Clone copying an Ixidron'd flipped Bushi Tenderfoot receives Bushi Tenderfoot's copiable values instead of CR 708.2a's "2/2 face-down creature with no text, no name, no subtypes, and no mana cost" (`docs/MagicCompRules.txt:5703`). One line: `if obj.face_down { return None; }`.

**3. CR 201.5 → CR 700.7 for the `"this <type>"` arms** (`imperative.rs:5046`, `:5277-5278`). CR 201.5 (`:1322`) is "text that refers to the object it's on **by name**" — correct for your `tag("~")` arm, wrong for `tag("this")`. CR 700.7 (`:3243`) is the "this [something]" rule. *(Agreeing with @coderabbitai here.)*

**4. Sasaya, Orochi Ascendant changed in a way the PR doesn't claim.** The parse diff shows `ability/static_structure` **removed** and `ability/RevealHand (card filter=none, player=any target)` **added** — the only card in the diff without a top-level `FlipPermanent`. Scryfall confirms the printed ability is *"Reveal your hand: If you have seven or more land cards in your hand, flip Sasaya."* — an activated ability whose **cost** is the reveal. The head parse models the reveal as the top-level *effect* with `player = TargetFilter::Any` rather than the controller, and moves the card off an honest `Effect::Unimplemented { name: "static_structure" }` marker onto a shape no test pins. Could you add a shape assertion (cost = reveal-your-hand, condition = ≥7 land cards in hand, effect = `FlipPermanent` on self) confirming this is the intended lowering?

**5. The corpus test cannot catch #4.** `every_printed_flip_card_lowers_its_instruction_to_flip_permanent` asserts only that *a* `FlipPermanent` exists somewhere in the chain, never the ability shape. Worth pinning the shape for at least the three non-uniform carriers — Sasaya's reveal cost, Initiate of Blood's delayed trigger, and Nezumi Shortfang's `Then if…`.

**6. Frontend change is untested.** `client/src/components/card/__tests__/ArtCropCard.test.tsx:53` still builds a `back_face` with no `layout_kind`, so reverting `hasOtherPrintedFace` fails nothing. A fixture with `layout_kind: "Flip"` asserting no DFC badge would close it. Secondary concern: "which layouts have a separately printed face" now lives in TS as `layout_kind !== "Flip"` and can drift from the engine's `is_double_faced_permanent`.

**7. Comment accuracy.** `parse_flip_permanent_subject`'s doc says "the ENTIRE printed flip-card corpus (21 cards)"; Scryfall `layout:flip include:extras` lists 24 distinct names. It also calls Budoka Pupil / Callow Jushi / Cunning Bandit / Faithful Squire / Hired Muscle "the five ki-counter Ascendants" — those are the Betrayers ki cycle; the Ascendants are the Saviors legends.

**8. One vacuous negative.** `kamigawa_flip_cards.rs:386-399` discards `resolve_trigger_body`'s events and asserts only `!flipped`. Add the `EffectResolved { kind: FlipPermanent }` reach guard its sibling `flipping_an_already_flipped_permanent_is_a_no_op` already carries. *(Agreeing with @coderabbitai.)*

**On @coderabbitai's other three findings, I'm overruling — no action needed.** The "quote CR text verbatim or drop the citation" ask on `types/ability.rs:11161` is wrong: our convention requires a mandatory *description*, not a quotation, and paraphrase is repo-wide. The `redundancy_avoidance.rs:464` already-flipped check isn't warranted — `FlipPermanent` is never independently chosen (it rides a trigger, or an ability whose primary effect carries the value), so there is no activation to penalize, and parity with `Transform` is correct. The 13-site citation-rewording sweep is over-strict; only `ability_scan.rs:6184` is loosely tagged and it isn't worth the churn.

## ✅ Clean

**Every CR 710 assumption the implementation rests on is verbatim-correct.** I grep-verified each one rather than taking the annotations on trust:

- **710.1b** (`:5781`) — the alternative characteristics are "used only if the permanent is on the battlefield and only if the permanent is flipped." `apply_flipped_face_to_object` copies exactly name / P&T / type line / text box, and `flip_permanent` early-returns off the battlefield. Matches.
- **710.1c** (`:5783`) — "A flip card's color and mana cost don't change if the permanent is flipped." This is precisely why `flip.rs` carries its own applicator instead of reusing `printed_cards::apply_back_face_to_object`, which swaps both. Correct, and mutation-verified by your own tests.
- **710.2** (`:5785`) — non-battlefield zones have only normal characteristics. `zones::apply_zone_exit_cleanup` calls `revert_flip_on_zone_exit`, ordered *after* the CR 708.9 face-down restore because both statuses share one `back_face` slot. The ordering rationale in the comment is right.
- **710.4** (`:5790`) — one-way, and a flipped permanent leaving the battlefield "retains no memory of its status." Repeat-flip is a silent no-op emitting no event; status clears on zone exit.
- **Copies — CR 707.2 + 707.3.** `flipped_normal_copiable_values` returns the normal half, and 707.3's worked example is *literally a flip card* (Tomoya the Revealer copying Nezumi Shortfang). Documenting the second-order case as out of scope, matching CR 110.5c's Dimir Doppelganger example, is the right scoping.
- **The CR 613.7 no-new-timestamp argument holds.** 613.7d/e/f/g enumerate zone entry, attachment, face up/down, and transform/convert — flipping is genuinely absent, so keeping the existing timestamp is correct rather than merely convenient.

**It handles the class, not the named cards.** `parse_flip_permanent_subject` is a composed, `eof`-anchored nom production over three independent surface forms (`flip this [creature|permanent|artifact|enchantment|land]`, `flip it|itself`, `flip ~`) — not a per-card tag list, and it covers types no printed flip card uses. Nom discipline held throughout: no `find`/`split_once`/`contains`/`starts_with` dispatch, no verbatim Oracle match.

**`FlipPermanent` as a sibling of `Transform` rather than a parameterization is the correct call** under the categorical-boundary rule. CR 701.27a (transform, DFC-only) and CR 710 (flip, single-faced, cost and color fixed) are different rule sections with different copiable-value semantics — exactly the case where unifying would conflate boundaries the engine resolves separately.

**Right seam, verified rather than assumed.** `flipped_normal_copiable_values` hooks `intrinsic_copiable_values`, which I traced to confirm it is read only by copy-source paths (`token_copy`, `become_copy`, `conjure`, `merge`, `augment`, `replacement`, `zones`) and **not** by the object's own layer-1 baseline — so the hook cannot un-flip a permanent on layer flush. Widening `is_double_faced_permanent` to `pub` for `manabrew-compat` replaces a `back_face.is_some()` re-derivation with the engine's authority, which is the right direction. `GameEvent::Flipped` threading matches `Transformed`'s full consumer set with no gaps, and `CardPreview` already guards flip cards via `isFlip`/`flip180` — I checked `GameCardPreview` as a suspected sibling gap and it is correctly covered.

**`the_flip_it_pronoun_form_flips_its_own_source_end_to_end`** drives the real stack (`push_to_stack` + `GameAction::PassPriority` through `apply()`) with an explicit reach guard asserting the effect lowered to `ParentTarget`. That is a genuinely discriminating runtime test, and `mod kamigawa_flip_cards;` is registered at `tests/integration/main.rs:598`, so none of it is inert.

---

**Recommendation: fix the authority gate, apply the two one-line guards (#1, #2), and answer #4. That is the whole remaining delta.** Please don't read the gate failure as a comment on the substance — the rules work underneath it is excellent, and I would like this to land.

matthewevans:
**Strong work on a genuinely hard class — the CR 710 modelling is right and the registration is complete. Two things need resolving before this can land, one of which I verified from the full CI artifact rather than the sticky comment.** Reviewed at head `3746a2d69`.

## ✅ What's right

`Effect::FlipPermanent` as a **sibling** of `Effect::Transform` rather than a parameterization of it is the correct call, and your justification at `types/ability.rs:11168-11185` states it well. The axis crosses CR 701.27 (transform — DFC-only, swaps mana cost and color) and CR 710 (flip — single-faced, holds both fixed). Unifying those would conflate two rule sections the engine resolves separately, which the categorical-boundary rule in CLAUDE.md explicitly forbids. This is the case where adding a variant is right and parameterizing is wrong.

Every CR number checks out against `docs/MagicCompRules.txt` — 710.1/1a/1b/1c/2/3/4/5 (lines 5777-5792), 701.27a/c, 707.2, 707.3, 110.5/5b/5c, 708.9, 400.7. **CR 710.1c** ("A flip card's color and mana cost don't change if the permanent is flipped") is verbatim and is exactly the right basis for a separate `apply_flipped_face_to_object` instead of reusing the DFC applicator.

I also enumerated every production `back_face.is_some()/is_none()` consumer, since this PR gives 21 cards a `back_face` they never had. All safe: the Room paths are subtype-gated, `zone_pipeline.rs:2405` is behind `should_transform`, and the `transformation_count` stampers are inert because `flip_permanent` never increments it, so the CR 701.27f guard can't misfire. Registration is threaded through all 20+ exhaustive-match sites with none dropped.

## 🔴 The PR claims 19 cards; production measures 17 — and Sasaya's flip clause is silently dropped

I did not rely on the sticky comment, which truncates to "first 3". I pulled the full `parse-diff.json` artifact from the head CI run (run `30055830721`, artifact `8583762397`):

| | |
|---|---|
| distinct cards in diff | **18** |
| cards gaining `FlipPermanent` | **17** |
| Sasaya in that set | **no** |
| `added_cards` / `removed_cards` | `[]` / `[]` |

Sasaya, Orochi Ascendant appears only as `∅ → RevealHand` and `static_structure → ∅`. So it moves **from the honest catch-all onto a supported effect while its flip clause disappears with no `Effect::unimplemented` marker.** That is a coverage-honesty regression — the card now reports as handled while half its text is gone — and it makes the PR body's "19 of the 21" inaccurate.

**The reason CI didn't catch it is a fixture path-divergence, and that's the part worth fixing.** `tests/integration/kamigawa_flip_cards.rs:434` puts Sasaya in `FLIP_INSTRUCTION_CORPUS`, and `:466` asserts every corpus card lowers to `Effect::FlipPermanent`. That test is green. It calls:

```rust
parse_oracle_text(oracle, card_name, &[], &[], &[])
```

with empty trailing arguments, while card-data generation passes real ones. The corpus test is therefore passing on a shape production never produces — which is precisely the failure mode a corpus test exists to prevent. Please route it through the same entry point card-data gen uses; that change is more valuable than the Sasaya fix itself, because it re-arms the guard for the next card.

## 🔴 Turning an *unflipped* flip permanent face down destroys its other half permanently

`effects/turn_face_down.rs:68-71` guards on `Some(flip_stash) if obj.flipped` — the flipped **status**, not "is a flip permanent". A flip card sitting unflipped is in its *normal* state, so it takes the `_` arm and `snapshot_object_base_face(obj)` overwrites `back_face` — which is holding the alternative half — with a snapshot of the normal half whose `layout_kind` is hard-coded `None` (`printed_cards.rs:744`). `turn_face_up.rs:44` then does `obj.back_face.take()`.

Two consequences:

1. **The permanent can never flip again.** `flip.rs:139` reads `let Some(alternative_face) = obj.back_face.clone() else { return Ok(()) }` and silently no-ops forever — Bushi Tenderfoot's trigger becomes dead text.
2. **It becomes indistinguishable from a DFC.** While face down, `is_flip_permanent` is false (tag gone) *and* `flipped` is false, so both arms of your new guard at `transform.rs:63` go false. A "transform each creature" effect then runs the DFC applicator on a flip card and blanks the mana cost and color — the exact outcome CR 710.1c exists to prevent, and the direct contradiction of your own invariant at `flip.rs:61-68` ("the half sitting in `back_face` always carries the tag, so no double-faced path can ever mistake a flip card for a DFC").

This is production-reachable *because* the PR deliberately lets flip cards through CR 712.16 — your own comment at `turn_face_down.rs:57-59` names Ixidron and Cyber Conversion. The module docs at `flip.rs:27-37` cover the flipped-then-face-down case only; the unflipped case is undocumented and untested.

Minimum fix: widen the preserve-arm to `Some(stash) if obj.flipped || flip::is_flip_permanent(obj)` and route turn-face-up through the flip-aware applicator. If you'd rather scope it down, at least re-stamp `LayoutKind::Flip` on the snapshot so the transform guard and the frontend DFC predicate stay honest — and extend the out-of-scope note to cover this case.

## 🟡 The `"flip" | "flips"` fallback opens a wider dispatch surface than CR 710 needs

In `parser/oracle_effect/imperative.rs`, the new `.or_else(|| parse_utility_imperative_ast(...))` appended to the coin-flip dispatch means any unmatched `"flip …"` clause is now offered to *every* utility arm. `parse_flip_permanent_subject` is carefully closed over self-references and correctly keeps CR 705.1 coin flips out (`flip_a_coin_still_parses_to_flip_coin` at `:558` pins that) — but the fallback reaching it is not. The measured blast radius is contained today, so this is a tightening rather than a live bug: call `parse_flip_permanent_subject` directly instead of the full utility parser.

## 📋 Checked and clean — no need to revisit

- **Frontend fix is non-vacuous.** `BackFaceData.layout_kind` is `skip_serializing_if = "Option::is_none"` and `LayoutKind` serializes as `"Flip"`, matching the new TS union at `client/src/adapter/types.ts:1122-1134`. `hasOtherPrintedFace` still returns true for a *transformed* DFC, so that button isn't regressed.
- **Test discrimination.** The unit tests carry explicit reach guards, and `the_flip_it_pronoun_form_flips_its_own_source_end_to_end` drives real `apply()`/stack rather than calling the resolver directly. No vacuous negatives.
- **CR 110.5c second-order copy (Dimir Doppelganger)** is a real gap but explicitly deferred at `flip.rs:38-43` and needs a second stash slot. Fine as a deferral.

One note for whoever reviews the next CR 710 change: `grep "^710.1c\."` returns nothing because subrules carry no trailing period. Grep `^710.1c` instead — the rule is there.

---

**Recommendation: changes requested.** The Sasaya divergence and the face-down stash overwrite are both concrete and both have small fixes. Nothing about the architecture needs revisiting — the seam, the variant choice, and the CR modelling are all correct.

matthewevans (approved):
Maintainer approval: current head db3c44ab implements the distinct CR 710 flip-card path with complete registrations, discriminating coverage, and green required checks.

PR #6571, opened 2026-07-23
Fix Wheel and Deal

invalidCards:
## Summary

Fixes Wheel and Deal so its selected opponents discard their hands and draw seven cards, while the controller performs the final draw exactly once. Removes the card from the parser misparse backlog.

## Files changed

- Parse bare continuations under multi-target opponent subjects as the selected opponent.
- Fan out opponent-targeted effects while preserving sequential controller-only tails.
- Extend the scenario cast driver to retain distinct declared multi-target players.
- Add parser, runtime, zero-target, and pause/resume regressions.

## CR references

- CR 601.2c — target declaration.
- CR 608.2c — resolving instructions and continuations.
- CR 701.9 — discard.
- CR 120 — draw.

Method: /engine-implementer

## Track

Developer

## LLM

Model: GPT-5 Codex
Thinking: high
Tier: Frontier

## Verification

- `cargo fmt --all`
- `git diff --check`
- Focused parser, integration, and pause/resume runtime regressions
- `./scripts/check-parser-combinators.sh` — PASS
- Fresh post-commit implementation review — PASS

## Gate A

Gate A PASS head=483bd981a65b3555950b3c210afe7977a11609b9 base=21a53d50cbcb8c21e6d250ae2c1f914fd6b0de15

## Final review-impl

PASS: fresh review of `21a53d50c..483bd981a65b3555950b3c210afe7977a11609b9` found no remaining issues.

## Claimed parse impact

- Wheel and Deal

## Validation failures

- `cargo clippy-strict` could not complete because this environment lacks the system OpenSSL development package required by `openssl-sys`; engine-focused clippy was additionally blocked behind the shared Cargo target-directory lock.
- Card-data generation, coverage, and semantic-audit were not run because that same shared Cargo lock did not clear during this session.

## CI failures

None observed locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected multi-target effects so each selected opponent or player resolves their instructions properly before follow-up effects occur.
  * Ensured final controller instructions resolve exactly once, including when no targets are selected.
  * Fixed continuation text so card draws correctly apply to previously selected targets.

* **Tests**
  * Added regression coverage for multi-target resolution, paused choices, and “Wheel and Deal” behavior.

* **Documentation**
  * Removed “Wheel and Deal” from the parser misparse backlog.

### DIFF
--- a/crates/engine/src/game/effects/mod.rs
+++ b/crates/engine/src/game/effects/mod.rs
@@ -3313,6 +3313,34 @@ fn split_player_scope_chain(
     (scoped, tail)
 }
 
+/// CR 608.2c: A multi-target player subject owns only its same-sentence
+/// continuation chain. A `SequentialSibling` starts an independent instruction
+/// that resolves once after every selected player has completed the subject's
+/// local clauses. This is the target-subject counterpart to
+/// `split_player_scope_chain` above.
+fn split_multi_target_player_chain(
+    ability: &ResolvedAbility,
+) -> (ResolvedAbility, Option<Box<ResolvedAbility>>) {
+    let mut per_target = ability.clone();
+    let tail = detach_after_multi_target_player_local_chain(&mut per_target);
+    (per_target, tail)
+}
+
+/// CR 608.2c: Retain `ContinuationStep` clauses under the current "each target
+/// player" subject, but detach the first independently sequenced instruction.
+fn detach_after_multi_target_player_local_chain(
+    node: &mut ResolvedAbility,
+) -> Option<Box<ResolvedAbility>> {
+    let mut next = node.sub_ability.take()?;
+    if next.sub_link != SubAbilityLink::ContinuationStep {
+        return Some(next);
+    }
+
+    let tail = detach_after_multi_target_player_local_chain(&mut next);
+    node.sub_ability = Some(next);
+    tail
+}
+
 /// CR 608.2e: Collect cross-player equalization quantity references from a
 /// `QuantityExpr`. These are the refs whose value would shift as an APNAP
 /// fan-out mutates the board — `ControlledByEachPlayer` (battlefield extremum)
@@ -3550,6 +3578,23 @@ fn extract_shares_quality_props(
 /// CR 608.2b: Extract the target filter from an effect for SharesQuality validation.
 fn effect_target_filter(effect: &Effect) -> Option<&TargetFilter> {
     effect.target_filter()
+}
+
+/// CR 601.2c: Oracle's `target opponents` lowering can use the dedicated
+/// `TargetFilter::Opponent` player role or a type-less `TypedFilter` constrained
+/// solely by `ControllerRef::Opponent`; `target players` uses
+/// `TargetFilter::Player`. All three select players at announcement and need
+/// the same per-target resolution fan-out.
+fn is_multi_target_player_filter(filter: &TargetFilter) -> bool {
+    match filter {
+        TargetFilter::Player | TargetFilter::Opponent => true,
+        TargetFilter::Typed(typed) => {
+            typed.type_filters.is_empty()
+                && typed.properties.is_empty()
+                && matches!(typed.controller, Some(ControllerRef::Opponent))
+        }
+        _ => false,
+    }
 }
 
 // ── Batch resolution (Tier 3) ────────────────────────────────────────────
@@ -7562,11 +7607,12 @@ fn resolve_chain_body(
         return Ok(());
     }
 
-    // CR 603.3d + CR 601.2c: Multi-target-over-`Player` resolution fan-out.
-    // An "any number of target players each <verb>" trigger/spell announces a
-    // variable number of player targets when it goes on the stack (CR 601.2c,
-    // reached for triggers via CR 603.3d). The chosen player set lands in
-    // `ability.targets` as `TargetRef::Player` entries with `multi_target` set.
+    // CR 603.3d + CR 601.2c: Multi-target player-subject resolution fan-out.
+    // An "any number of target players/opponents each <verb>" trigger/spell
+    // announces a variable number of player targets when it goes on the stack
+    // (CR 601.2c, reached for triggers via CR 603.3d). The chosen player set
+    // lands in `ability.targets` as `TargetRef::Player` entries with
+    // `multi_target` set.
     // Single-player-recipient effect handlers (`Discard`, `Mill`, `LoseLife`)
     // resolve for only the FIRST `TargetRef::Player`, so without this branch a
     // selection of two players discards only one. This fan-out is the missing
@@ -7580,7 +7626,7 @@ fn resolve_chain_body(
     // `multi_target` and changes the effect's target filter, which would
     // otherwise skip this branch and resolve only the first chosen player.
     if ability.multi_target.is_some()
-        && effect_target_filter(&ability.effect) == Some(&TargetFilter::Player)
+        && effect_target_filter(&ability.effect).is_some_and(is_multi_target_player_filter)
     {
         let chosen_players: Vec<PlayerId> = ability
             .targets
@@ -7591,6 +7637,11 @@ fn resolve_chain_body(
             })
             .collect();
         if chosen_players.len() != 1 {
+            // A sentence boundary begins an instruction that is not qualified
+            // by the preceding "each target player" subject. Keep that tail
+            // outside the per-target template so Wheel and Deal's final
+            // "Draw a card" is performed once by its controller.
+            let (per_target, after_fanout) = split_multi_target_player_chain(ability);
             // CR 601.2c: "any number of target players" permits zero targets.
             // CR 608.2c: An ability resolving with zero chosen player targets
             // does nothing — emit `EffectResolved` and stop BEFORE
@@ -7605,6 +7656,9 @@ fn resolve_chain_body(
                     source_id: ability.source_id,
                     subject: None,
                 });
+                if let Some(after_fanout) = after_fanout {
+                    resolve_ability_chain(state, &after_fanout, events, depth + 1)?;
+                }
                 return Ok(());
             }
             // CR 101.4 + CR 608.2c: Resolve the effect once per chosen player in
@@ -7617,7 +7671,7 @@ fn resolve_chain_body(
                 .collect();
             let initial_waiting_for = state.waiting_for.clone();
             for (i, pid) in fanout_players.iter().enumerate() {
-                let mut narrowed = ability.clone();
+                let mut narrowed = per_target.clone();
                 narrowed.targets = vec![TargetRef::Player(*pid)];
                 narrowed.multi_target = None;
                 resolve_ability_chain(state, &narrowed, events, depth + 1)?;
@@ -7629,9 +7683,9 @@ fn resolve_chain_body(
                 // `drain_pending_continuation` after the choice resolves.
                 if state.waiting_for != initial_waiting_for {
                     let remaining = &fanout_players[i + 1..];
-                    let mut tail: Option<Box<ResolvedAbility>> = None;
+                    let mut tail = after_fanout.clone();
                     for &remaining_pid in remaining.iter().rev() {
-                        let mut remaining_narrowed = ability.clone();
+                        let mut remaining_narrowed = per_target.clone();
                         remaining_narrowed.targets = vec![TargetRef::Player(remaining_pid)];
                         remaining_narrowed.multi_target = None;
                         if let Some(prev) = tail {
@@ -7644,6 +7698,11 @@ fn resolve_chain_body(
                     }
                     append_to_pending_continuation(state, tail);
                     break;
+                }
+            }
+            if state.waiting_for == initial_waiting_for {
+                if let Some(after_fanout) = after_fanout {
+                    resolve_ability_chain(state, &after_fanout, events, depth + 1)?;
                 }
             }
             return Ok(());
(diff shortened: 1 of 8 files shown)


matthewevans:
**This is a well-built fix — the split between the per-target chain and the controller-only tail is the right decomposition, and the tests discriminate on exactly the property that matters. Nothing blocking. Three items, one of which is a class-level question I'd like answered before it lands.**

Reviewed at head `483bd981`.

## 🟡 Non-blocking

**1. The fan-out iterates target-declaration order, not APNAP — and that matters for the class, not for this card.** `resolve_multi_target_player_chain` walks `ability.targets` in the order they were announced. For Wheel and Deal this is unobservable: "discard their hands" takes the whole hand and "draw seven cards" involves no choice, so no ordering is detectable. But the abstraction you've built is *"any number of target players/opponents each &lt;verb&gt;"*, and the moment that verb involves a choice — "each discards a card", "each sacrifices a creature" — CR 101.4 (`docs/MagicCompRules.txt:235`) requires the active player to choose first, then each nonactive player in turn order.

This is not hypothetical: #6545 is fixing precisely this bug in the per-owner library-order fan-out, and landed on `crate::game::players::apnap_order(state)` as the engine's authority for it. Since your fan-out is the more general primitive of the two, it would be good for it to reach for the same authority rather than acquire the same defect independently. If you have a reason declaration order is correct here that I'm missing, say so and I'll drop it — but as written the code doesn't explain the choice, and the natural reading is that APNAP simply wasn't considered.

**2. `CR 120 — draw` in the PR body is the wrong rule.** CR 120 is **Damage** (`:1085`); drawing a card is **CR 121** (`:35`). Your *in-code* annotations are all correct and all verified — CR 601.2c (`:2461`), CR 608.2c (`:2793`), CR 603.3d (`:2590`), CR 701.9 Discard (`:3327`) — so this is confined to the PR description's reference list and does not affect the code. Worth correcting anyway, since PR bodies get read as provenance for why a change is right.

**3. Heads-up on #6572, which touches the same class.** #6572 adds a post-lowering pass attaching `TargetSelectionConstraint::DifferentTargetPlayers` to every non-modal ability whose `multi_target` targets `Player`/`Opponent`. Wheel and Deal matches that gate, so once both land, its opponent selection will additionally be constrained to distinct players. That is correct under CR 601.2c — "any number of target opponents" is one instance of the word "target", so the same opponent can't fill two slots — and the two changes compose rather than conflict (one governs selection, the other resolution). No action needed; flagging it so neither of you is surprised by the other's behavior change, and so whichever lands second doesn't read the other's effect as a regression.

## ✅ Clean

- **The decomposition is the right one.** `split_multi_target_player_chain` / `detach_after_multi_target_player_local_chain` separate "what each selected player does" from "what resolves once afterward", which is the actual structure of CR 608.2c's instruction ordering rather than a Wheel-and-Deal-shaped special case. The recursive detach handles a tail of arbitrary length instead of assuming exactly one trailing instruction.
- **`is_multi_target_player_filter` covers the class, not the card.** It accepts `TargetFilter::Opponent`, a type-less `TypedFilter` constrained solely by `ControllerRef::Opponent`, and `TargetFilter::Player` — three lowerings that all mean "players chosen at announcement". Enumerating them in one predicate with a comment explaining why all three exist is much better than matching whichever one Wheel and Deal happens to produce.
- **The controller-draw test is a real discriminator.** `assert_hand_drawn(P0, 1)` fails the moment the trailing "Draw a card." is swept into the fan-out, because the controller would draw once per selected opponent. Stacking eight named cards on the controller's library (`"Controller draw 1"` … `"Controller draw 8"`) means an over-draw is caught by identity, not just by count. The per-opponent assertions pair a draw count with a graveyard length of 2, so "discarded the hand" and "drew seven" are checked independently rather than collapsing into one net number.
- **The zero-target case is covered and is the right edge.** "Any number" permits zero, and `wheel_and_deal_with_no_targets_draws_only_for_its_controller` asserts the opponent-qualified instruction is skipped *while the independent final draw still happens* — that's the specific way this class breaks if the tail is treated as part of the fan-out rather than beside it.
- **The cast pipeline is checked, not assumed.** Asserting `ability.targets == [Player(P1), Player(p2)]` on the stack before resolving proves the scenario driver change actually retains both declared players, so a regression there fails loudly instead of silently degrading to one target.
- **Backlog hygiene.** Removing the entry from `docs/parser-misparse-backlog.md` in the same change is correct — that file tracks live misparses, and leaving a fixed card listed makes the backlog lie about the parser's real state.
- **Proof package is complete**, which is rarer than it should be: `Files changed`, `Track`, `LLM`, `Verification`, `Implementation method`, a `Gate A PASS` with both head and base pinned, and a fresh post-commit `review-impl`. `./scripts/check-parser-combinators.sh` PASS is the one I most wanted to see given `assembly.rs` is in the diff.
- **`Tauri compile check` red is ours, not yours** — the v0.35.2 release left `client/src-tauri/Cargo.lock` stale against `--locked`. Fixed on main in #6568, permanently prevented in #6574; it will clear on rebase. It is the only failing job here.

---

**Recommendation: answer the APNAP question in #1 and fix the CR 121 reference; I'm happy to approve on either "here's why declaration order is right" or a switch to `players::apnap_order`.** The rest of this is ready — the chain split and the test design are both better than the card required, which is the point.

matthewevans:
**Addendum to my review above — a second pass found one blocker I missed, and corrected one of my own CR references.**

## 🔴 Blocker — vacuous negative test

`crates/engine/src/parser/oracle_effect/tests.rs` — `multi_target_player_subject_does_not_capture_explicit_controller_draw` asserts only that the sub-ability's effect is `Draw { count: 1, target: Controller }`.

That passes for the wrong reason. If the `"Any number of target opponents each"` head failed to parse as a multi-target player subject *at all*, `"then you draw a card"` still lowers to a `Controller` draw and the assertion holds — so the test cannot distinguish "the guard correctly declined to capture the controller draw" from "the feature never engaged." It's the only one of your five tests without a reach guard; its sibling `wheel_and_deal_implicit_draw_inherits_each_targeted_opponent` has two before its `matches!`.

Add the same guard:

```rust
assert!(def.multi_target.is_some());
assert!(def.effect.target_filter().is_some_and(target_filter_can_target_player));
```

@coderabbitai raised this as 🟠 Major and it's unaddressed — that one was right and I should have confirmed it in my first pass rather than leaving it open.

## Correcting my own CR reference

In my comment above I cited **CR 101.4** for the ordering concern. The sharper rule is **CR 608.2e** (`docs/MagicCompRules.txt:2798`): choices for the first action are made in APNAP order, then that action is processed **simultaneously**, then the second action. `effects/mod.rs:7666-7675` resolves each opponent's full local chain before starting the next, so with two targets P1 draws seven before P2 discards.

Two things follow. This is **pre-existing fan-out architecture, not a regression you introduced** — your change improves granularity and doesn't make it worse, so I'm downgrading it from "please answer this" to a follow-up item. And separately, my zero-target citation was wrong:

**CR 115.6** (`:859`) — *"A spell or ability that requires targets may allow zero targets to be chosen"* — is the rule for the zero-target permission at `effects/mod.rs:7645` and `tests/integration/wheel_and_deal.rs:114`, where you currently cite CR 601.2c. CR 601.2c covers announcing a *variable* number of targets, which is a different claim. Keep CR 608.2c for the independent tail; it's correct.

## Two more non-blocking

- **The `Effect::Draw`-only gate leaves a named sibling unfixed.** `parser/oracle_effect/assembly.rs:2091` fires the `ParentTarget` rewrite only for `Effect::Draw`. **Turtle Tracks** — *"Any number of target players may each search their library for a basic land card, put it onto the battlefield under their control, then shuffle"* — is the same shape (multi-target player head + bare `ContinuationStep`) and is still listed in `docs/parser-misparse-backlog.md`, the file this PR edits. Either widen to any continuation whose target is `Controller`/`Player`/`ParentTargetController`, or say in the comment why `Draw` is the only currently-safe rewrite.
- **Three `TargetFilter::Opponent` arms appear to be dead.** `effects/mod.rs:3583` asserts that Oracle's `target opponents` lowering "can use the dedicated `TargetFilter::Opponent` player role", but its only non-test construction in the engine is `ctx.target_chooser` at `parser/oracle_target.rs:3321`. The parse-diff corroborates this: widening `target_filter_can_target_player` moved **zero** cards despite that predicate having 14 parser consumers. Either drop the arms at `effects/mod.rs:3590`, `assembly.rs:89`, `parser/oracle_effect/mod.rs:19643`, or point at the lowering site that produces them.
- **Duplicated predicate.** `is_multi_target_player_subject_definition` (`assembly.rs:84`) and `is_multi_target_player_filter` (`effects/mod.rs:3588`) are byte-identical match bodies in two layers. One inherent method on `TargetFilter` — say `selects_players_at_announcement()` — would keep them from drifting.

## Reinforcing what's good

The parse-diff is an exact match: **1 card, 1 signature, Wheel and Deal, `ability/Draw` `target: ∅ → parent target`, zero added/removed**. Claim and measurement agree precisely, which is rare and worth saying.

And the sentence-boundary reasoning is stronger than I gave it credit for: `SubAbilityLink` has exactly two variants (`types/ability.rs:16696-16711`), so "keep `ContinuationStep`, detach the first non-`ContinuationStep`" *is* CR 608.2c's sentence boundary rather than an approximation of it. Draw single authority is respected end to end — `Effect::Draw` still routes through `draw::resolve` → `start_draw_sequence_with_replacement_applied` with `replacement_applied` threaded per CR 614.5 (`:3073`), and CR 121.2 (`:1144`) per-card sequencing is preserved. Four of your five tests fail on revert, two of them in two independent ways.

---

**Revised recommendation: add the reach-guard to the negative test, fix the CR 115.6 citation, and answer the Turtle Tracks question. The APNAP/CR 608.2e item is a follow-up, not yours to fix here.**

matthewevans:
**Approving. The seam is right, the fix generalizes further than the title suggests, and the tests are discriminating.** Reviewed at head `1653d3346`.

## Verified against the card, not the description

Pulled the Oracle text from Scryfall rather than trusting the PR body:

```
Wheel and Deal — {3}{U} Instant
Any number of target opponents each discard their hands, then draw seven cards.
Draw a card.
```

It matches the fixture at `tests/integration/wheel_and_deal.rs:9-10` byte-for-byte including the newline. The subtlety the fix turns on is real: *"then draw seven cards"* belongs to the targeted opponents (same sentence), while *"Draw a card."* is a separate sentence for the controller.

## I pushed hard on "one card or a class?" — and the measurement answered it

A single-card title is the usual smell, so I checked both halves.

**The runtime half is already class-shaped.** `is_multi_target_player_filter` (`effects/mod.rs:3588`) widens the fan-out guard from `== TargetFilter::Player` to `Player | Opponent | Typed{empty, controller: Opponent}`. Against `card-data.json` that newly fans out **Communal Brewing** and **Hollow Marauder**, both of which previously resolved for only the first chosen opponent. That's an undisclosed multi-card bug fix — worth naming in the PR body.

**The parser half looks card-shaped** (`if let Effect::Draw` at `assembly.rs:2091`), so rather than assert, I measured the population: all 35,477 cards scanned for `multi_target` + player-ish head filter + a `ContinuationStep` sub whose target is a context-ref. **Exactly one hit — Wheel and Deal.** The obvious candidate siblings (Will of the Abzan, Tinybones Joins Up, Priest of Forgotten Gods, Turtle Tracks) all turn out not to need it: their continuations carry non-context-ref targets or `None`, which already inherit `ability.targets` at `life.rs:517-522`. So the narrowing leaves zero cards behind today. Not a special case.

Parse-diff sticky confirms it independently: 1 card, 1 signature, `ability/Draw` field `target` going `∅ → parent target`, Wheel and Deal only. Claimed impact equals measured impact. CR annotations grep-verified — 601.2c, 608.2c, 603.3d, 101.4, 701.9 all real and all describing the code they sit on.

The integration test genuinely flips on revert: restoring `Draw{Controller}` gives P1 −2 / P0 +8 instead of +5/+1.

## 🟡 Non-blocking — worth a follow-up, not another round

**The exactly-one-target path is production-reachable and has no fixture.** `effects/mod.rs:7639` gates the whole fan-out — and therefore `split_multi_target_player_chain` — behind `chosen_players.len() != 1`. Your two tests use 2 targets and 0 targets. Casting this at a single opponent is the common real-world case and takes a structurally different path. I traced it and it *is* correct (`ParentTarget` is non-context-ref, so `mod.rs:5948-5954` returns the first `TargetRef::Player`; the trailing `Controller` draw is a context-ref and falls to the controller) — so this is a coverage gap rather than a bug. A third case `target_players(&[P1])` asserting P1 +5 / P0 +1 would close it.

**`has_implicit_player_subject_continuation` is fail-open.** `assembly.rs:103-129` treats any fragment not matching one of 12 hardcoded subject prefixes as an implicit continuation of the targeted-opponent subject. Unlisted real prefixes — "For ", "If ", "Put ", "Each of ", "Their " — would silently rebind a controller instruction to the opponents. Hollow Marauder is the near-miss: *"For each of those opponents … draw a card"* is a `Draw` under a `Typed{Opponent}` multi-target head, and the only things saving it are the `sub_link == ContinuationStep` gate and the `Effect::Draw` arm. Blast radius today is empirically zero, but the guard inverts from safe to dangerous the moment either narrowing is relaxed — which is exactly what the next card in this shape will need. Inverting it to an allowlist of bare-conjugated-verb continuation shapes would be the durable form.

**[LOW] The three-way player-filter predicate is duplicated across the parser/runtime boundary** — `assembly.rs:84-97` and `effects/mod.rs:3588-3600` encode the identical shape. They must agree for the `ParentTarget` rewrite and the fan-out to stay in lockstep, and adding a variant to one silently desynchronizes them. Hoisting one predicate onto `TargetFilter` (something like `selects_players_at_announcement()`) and calling it from both would make that structural.

**[LOW] The PR body cites "CR 120 — draw".** CR 120 is *Damage*; drawing is CR 121 (`MagicCompRules.txt:1087` vs `:1142`). Body-only — every in-code annotation is correct — but worth fixing since we grep-verify these.

## 📋 Checked and clean

`scenario.rs`'s `pick_slot_target` now consumes player declarations when `multi_target.is_some()` (`:2578-2586`) and degrades gracefully to the reusable branch once exhausted, so Kozilek's-Command-style modal reuse is preserved. Both new parser tests assert positive shapes, and `multi_target_player_subject_does_not_capture_explicit_controller_draw` is a real discriminating negative for the `"you "` prefix. Removing Wheel and Deal from the backlog is justified — it now both parses and resolves correctly.

---

Approving and enqueueing. The two MED items are follow-up material; I'd particularly like the fail-open guard inverted before anyone extends this pattern, but it shouldn't hold up a correct fix.

matthewevans (approved):
Approved — see the review comment above. Findings are non-blocking follow-ups.

PR #6573, opened 2026-07-23
fix(release): regenerate client/src-tauri/Cargo.lock for v0.35.2

matthewevans:
release: v0.35.2 (21a53d50cb) bumped client/src-tauri/Cargo.toml to
0.35.2 but left client/src-tauri/Cargo.lock pinning phase-tauri 0.35.1.
The Tauri compile check runs `cargo check --locked`, which refuses to
update the lock, so the job fails on main, on every merge-group ref, and
on every open PR based on the release tip. That job is a `needs:` of the
required aggregator (.github/workflows/ci.yml:623-633), so the merge
queue ejects everything it builds.

### DIFF
--- a/client/src-tauri/Cargo.lock
+++ b/client/src-tauri/Cargo.lock
@@ -2616,7 +2616,7 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "phase-tauri"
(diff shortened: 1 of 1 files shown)


matthewevans:
Closing as a duplicate of #6568, which carries the identical one-line `client/src-tauri/Cargo.lock` fix and was opened first. Keeping #6568 as the one that lands.

PR #5762, opened 2026-07-13
fix(parser): UNSUPPORTED seam: Play historic from graveyard once/turn + granted leave-battlefield re

ntindle:
## Summary
Fixes a parser misparse affecting 1 card(s) in the Doctor Who Commander precons.

**Root cause:** UNSUPPORTED seam: Play historic from graveyard once/turn + granted leave-battlefield replacement rider on the played object

## Cards corrected
- The Eighth Doctor

## Fix
Implemented the approved plan for Cluster-113 (The Eighth Doctor): a play/cast-from-graveyard permission that GRANTS a per-object leave-the-battlefield → exile replacement to the permanent played this way. Built the general carrier seam, not a Doctor-special-case. No new enum variant — two struct-field additions plus a shared install helper and a shared parser def-builder.

WHAT CHANGED (by subsystem):
Types: added `granted_replacement: Option<Box<ReplacementDefinition>>` as a fourth leaf rider on `StaticMode::GraveyardCastPermission` (serde-only, Display/FromStr default None; Hash/Display already destructure with `..` so no change there — ReplacementDefinition lacks Hash, confirmed). Added transient `GameState.pending_etb_replacements: Vec<(ObjectId, Box<ReplacementDefinition>)>` mirroring `pending_etb_counters` at all four lifecycle sites (field/Default/destructure/PartialEq).
Parser: extended `parse_leave_battlefield_rider_ref` with a `~` self-ref alt arm; extracted shared `parse_leave_battlefield_exile_replacement_def` (byte-identical effect-path output preserved via a thin wrapper); replaced the honest-gap `return None` bail in `try_parse_disjunctive_graveyard_cast_permission` with a nom `split_once_on(". if you do, it gains ")` + `delimited(char('"'), take_until, char('"'))` quote-strip that feeds the shared builder (rider present-but-unparsable still aborts — honest gap preserved).
[...]

PR #6580, opened 2026-07-24
fix(engine): granted "leave the battlefield → exile instead" riders must actually fire (Geth, Realmbreaker, Elemental Expressionist)

### LINKED ISSUE #6566: Quoted grants of replacement-shaped abilities ("if it would leave the battlefield, exile it instead") parse to an inert GrantAbility and never fire
## Summary

A quoted ability grant of a **replacement-shaped** effect — e.g. `it gains "If this creature would leave the battlefield, exile it instead of putting it anywhere else"` — currently parses to an inert `ContinuousModification::GrantAbility`. The concretized ability body is pushed into `obj.abilities` (`crates/engine/src/game/layers.rs`), never into `replacement_definitions`, so the replacement **never fires** — the redirect is silently a no-op. Affected entries also carry a `SwallowedClause` / `Replacement_Instead` parse warning.

Surfaced during review of #6538 (unearth's standalone leave-battlefield exile rider); the granted variant is a distinct, pre-existing gap and was correctly held out of that PR.

## Affected cards (5 — verified against `card-data.json`)

Exactly the cards that carry the `Replacement_Instead` warning **and** an inert `GrantAbility` for this rider:

- **Elemental Expressionist** — grants it **until end of turn** (must lapse at cleanup, CR 611.2a / 514.2).
- **Geth, Thane of Contracts** — permanent grant on a reanimation ability. *(Correction: my initial draft named "Geth, Lord of the Vault" — that printing has no such rider; the matching card is Thane of Contracts.)*
- **Llanowar Greenwidow** — permanent (also has a separate unimplemented cost-reduction clause, out of scope).
- **Realmbreaker, the Invasion Tree** — permanent.
- **Spirit-Sister's Call** — permanent.

## Out of scope / separate gaps

- **Magar of the Magic Strings** and **The Eighth Doctor** — their whole granting clause is `Effect::Unimplemented` (a face-down-manifest body / a play-from-graveyard permission), so they never reach the `GrantAbility` seam; separate parser gaps.
- **Geth, Lord of the Vault** — no such rider on its current printing.

## Expected

A quoted grant of a replacement-shaped ability should install a **granted, duration-bound replacement effect** that (a) redirects the event while active, and (b) honors the grant's stated duration (lapses at cleanup for `until end of turn`; persists for permanent grants — CR 611.2a) via the layer system, **not** the object-lifetime host stamp used for standalone reanimation riders (#6538).

## Seam

`oracle_static::classify_quoted_inner` must recognize a quoted replacement clause and route it to a granted `ReplacementDefinition` (a new `ContinuousModification::GrantReplacement`) re-applied each layer pass (CR 613.1f), rather than collapsing it to an inert `GrantAbility`. Runtime install: the Layer-6 apply in `layers.rs`.

Repro (Elemental Expressionist): grant the rider to a creature, then destroy/sacrifice it **within the turn** → it should be exiled, not sent to the graveyard; after cleanup the rider should be gone.

michiot05:
Tier: Frontier
Model: claude-opus-4-8

Closes #6566 (filed at your request during the #6538 review).

## Summary

A quoted grant of a **replacement-shaped** ability parsed to an inert `ContinuousModification::GrantAbility`: the concretized body was pushed into `obj.abilities` and never into `replacement_definitions`, so the redirect silently did nothing. **Geth, Thane of Contracts** — *"Return target creature card from your graveyard to the battlefield. It gains \"If this creature would leave the battlefield, exile it instead of putting it anywhere else.\""* — reanimated a creature that then went to the graveyard on any battlefield exit.

Adds `ContinuousModification::GrantReplacement`, the replacement-store sibling of `GrantAbility`/`GrantTrigger`/`GrantStaticAbility` (CR 614 is a categorically distinct sink from CR 602/603/604). It is applied at layer 6 with structural dedup and **re-derived every layer pass** (CR 613.1f), so its lifetime follows the granting continuous effect rather than an object-lifetime stamp:

- **Permanent grants** (Geth, Llanowar Greenwidow, Realmbreaker, Spirit-Sister's Call) persist until the host leaves (CR 611.2a), surviving the granting source itself leaving.
- **Duration-bound grants** — Elemental Expressionist's *"Until end of turn, it gains …"* — lapse at cleanup (CR 514.2).
- A grant carrying **other modifications alongside** the replacement is never promoted to permanent. Expressionist grants two abilities (`[GrantReplacement, GrantTrigger]`), so that guard is load-bearing and has its own revert-failing regression.

This is deliberately **not** #6538's host-lifetime stamp — that path is for standalone (non-granted) riders and is unchanged here; the two now share one `ReplacementDefinition` constructor.

**Blocker fixed along the way (the feature is a silent no-op without it):** `normalize_card_name_refs` rewrites `this creature|permanent|land` → `~` card-wide (CR 201.5b), but `parse_leave_battlefield_rider_ref` had no `~` arm, so *every* quoted grant failed to parse before reaching the classifier. The added arm is purely additive; I confronted all 74 corpus cards containing "would leave the battlefield" and **no standalone rider uses a `this <type>` subject**, so the #6538 front door is provably unaffected.

## Relationship to #5762 (disclosed — please tell me how you'd like this sequenced)

@ntindle's open #5762 (*Play historic from graveyard once/turn + granted leave-battlefield replacement rider*) addresses **The Eighth Doctor** at a different seam: it threads an optional `ReplacementDefinition` through `StaticMode`'s play-from-graveyard permission. Mine routes **quoted grants** through the classifier into a layer-6 `GrantReplacement`.

They look complementary rather than competing — The Eighth Doctor is explicitly **out of scope** here precisely because its whole clause is `Effect::Unimplemented { name: "static_structure" }` at the permission seam #5762 fixes, so it never reaches my classifier and does **not** appear in my parse-diff. That said, the two branches touch `add_target_replacement.rs` and `oracle_effect/mod.rs`, so they will conflict textually. Happy to rebase onto whichever lands first, or to hold this until #5762 resolves — your call; I did not want two granted-replacement lifecycle changes sitting in the queue without flagging it.

Method: /engine-implementer — plan → /review-engine-plan (**3 rounds**; round 1 corrected the card set and identified Geth *Thane of Contracts* as the real card, round 2 caught the `~` no-op blocker, round 3 resolved which cards actually flip) → implement → /review-impl (**3 rounds**; round 1 required the Expressionist EOT-lapse regression, the final pass required the mixed-def guard test) → commit. Each step in a fresh agent context. Final read-only /review-impl: **PASS head=054b54d30d76eccd659b71fb01f2763576b3353d**.

## Gate A

`./scripts/check-parser-combinators.sh` → **Gate A PASS head=054b54d30d76eccd659b71fb01f2763576b3353d** (Gate G PASS). The only parser change is one `tag("~")` arm composed into an existing `alt` — no string probes.

## Anchored on

- `ContinuousModification::GrantTrigger` (`types/ability.rs`) + its Layer-6 apply (`game/layers.rs`) — the sibling grant that lands in a *different* store (`obj.trigger_definitions`); `GrantReplacement` mirrors it exactly onto `obj.replacement_definitions`, with the `GrantStaticAbility` structural-dedup push as the idempotency idiom.
- `game/layers.rs` base→live reseed — the store is reset to base every pass, which is why the layer-native design needs no stamp, no base-install, and no host-exit prune (the grant's own duration governs, via `prune_end_of_turn_effects` / `prune_affected_object_left_effects`).
- `game/mod.rs` `game::zones` — the documented `#[cfg(any(test, feature = "test-support"))] pub` / `#[cfg(not(…))] pub(crate)` precedent reused verbatim to expose the shared constructor to the integration-test crate (the feature is enabled only from `[dev-dependencies]`; `cargo tree -e normal` shows `engine FEATURES=[]`).

## Verification

- `cargo fmt --all` clean; `cargo clippy -p engine --tests` **0 warnings**; `cargo build -p engine -p phase-ai` clean.
- **Integration** (`crates/engine/tests/integration/issue_6566_granted_leave_exile.rs`, registered in `main.rs`) — **5 passed**: Geth verbatim-Oracle reanimate → destroy → **exile** (with an un-granted control creature → graveyard as the paired reach guard); permanent grant survives end-of-turn cleanup; duplicate grants dedup to one definition; Expressionist-shaped until-end-of-turn grant **exiles same turn** and **lapses at cleanup** (post-cleanup destroy → graveyard); mixed `[GrantReplacement, GrantTrigger]` def is **not** promoted to permanent.
- **Parser/unit** — the `~` rider row, the `it`-subject no-regression row, `classify_quoted_inner` → `GrantReplacement`, a non-rider quoted body → **not** `GrantReplacement`, layer classification, and a byte-identity assertion that the shared constructor still emits #6538's exact `ReplacementDefinition` + `AddTargetReplacement { target: Any }` wrapper.
- **Fails before, passes after** — demonstrated three times: reverting the `tag("~")` arm fails the parser row and all integration rows; dropping the `&& duration.is_none()` conjunct fails *only* the Expressionist EOT row (the three Geth rows still pass — which is exactly why that regression was added); swapping the mixed-def `.all` → `.any` fails *only* the mixed-def row.
- **CR annotations** — every number grep-verified against `docs/MagicCompRules.txt`: 614.1a, 614.6, 613.1f, 611.2a, 514.2, 201.5b, 702.84a.
- Exhaustive `ContinuousModification` arms added at all 10 wildcard-free match sites (compile-finalized, including 3 the compiler surfaced beyond the planned list).

## Claimed parse impact

Regenerated `card-data.json` on this head and counted occurrences of the new variant (brand-new, so every occurrence is a flip): **exactly 5 cards** — `elemental expressionist`, `geth, thane of contracts`, `llanowar greenwidow`, `realmbreaker, the invasion tree`, `spirit-sister's call`. Explicitly verified **absent**: **Magar of the Magic Strings** and **The Eighth Doctor** (both swallowed upstream by `Effect::Unimplemented`, so untouched by this change) and **Geth, Lord of the Vault** (no such rider on that printing — my original issue text mis-named it; corrected in #6566).

## Scope Expansion

None.

## Summary by CodeRabbit

* **New Features**
  * Added support for granted replacement effects, including “If this would leave the battlefield, exile it instead” riders.
  * Granted replacement payloads are now correctly discovered, layered as ability effects, and preserved on recipients (including self-reference behavior).
* **Bug Fixes**
  * Improved handling of `~`/self-reference variants in replacement parsing.
  * Prevented duplicate hosted replacements and corrected permanent vs end-of-turn duration promotion.
  * Ensured replacement-related scanning/coverage and X-references are not missed.
* **Tests**
  * Added integration and unit tests covering parsing, layering, deduplication, replacement outcomes, and duration edge cases.

### DIFF
--- a/crates/engine/src/analysis/ability_graph.rs
+++ b/crates/engine/src/analysis/ability_graph.rs
@@ -1889,6 +1889,11 @@ fn build_nodes(faces: &[&CardFace]) -> Vec<AbilityNode> {
                             nodes.push(build_node(&face.name, def, trigger_axis(trigger)));
                         }
                     }
+                    ContinuousModification::GrantReplacement { replacement } => {
+                        if let Some(def) = &replacement.execute {
+                            nodes.push(build_node(&face.name, def, None));
+                        }
+                    }
                     _ => {}
                 }
             }
--- a/crates/engine/src/database/unearth.rs
+++ b/crates/engine/src/database/unearth.rs
@@ -38,13 +38,12 @@
 
 use crate::types::ability::{
     AbilityCost, AbilityDefinition, AbilityKind, ContinuousModification, DelayedTriggerCondition,
-    Duration, Effect, ReplacementDefinition, RestrictionExpiry, StaticDefinition, TargetFilter,
+    Duration, Effect, RestrictionExpiry, StaticDefinition, TargetFilter,
 };
 use crate::types::card::CardFace;
 use crate::types::keywords::Keyword;
 use crate::types::mana::ManaCost;
 use crate::types::phase::Phase;
-use crate::types::replacements::ReplacementEvent;
 use crate::types::zones::Zone;
 
 /// CR 702.84a: Synthesize the graveyard-activated reanimation ability for every
@@ -162,13 +161,13 @@ fn delayed_exile_step() -> AbilityDefinition {
 /// reseeds, kept non-copiable (CR 707.2), and pruned on the host's battlefield
 /// exit (CR 400.7) — see the variant doc in `types/ability.rs`.
 fn leaves_battlefield_exile_step() -> AbilityDefinition {
-    let replacement = ReplacementDefinition::new(ReplacementEvent::Moved)
-        .valid_card(TargetFilter::SelfRef)
-        .expiry(RestrictionExpiry::UntilHostLeavesPlay)
-        .execute(AbilityDefinition::new(
-            AbilityKind::Spell,
-            exile_self_from_battlefield_effect(),
-        ));
+    // Shares the single unstamped `leave_battlefield_exile_replacement` authority
+    // (#6538 / #6566); this site keeps its own `target: SelfRef` wrapper +
+    // description and composes the #6538 host-lifetime stamp itself. The stamp is
+    // applied per-consumer, not baked into the constructor, because the granted
+    // path (#6566) must NOT carry it — see the constructor doc.
+    let replacement = crate::parser::oracle_effect::leave_battlefield_exile_replacement()
+        .expiry(RestrictionExpiry::UntilHostLeavesPlay);
 
     AbilityDefinition::new(
         AbilityKind::Spell,
--- a/crates/engine/src/game/ability_rw.rs
+++ b/crates/engine/src/game/ability_rw.rs
@@ -104,9 +104,9 @@ use crate::types::ability::FilterProp;
 use crate::types::ability::{
     AbilityCondition, AbilityDefinition, ContinuousModification, ControllerRef, Duration, Effect,
     GuessSubject, ModalChoice, MultiTargetSpec, ObjectScope, PlayerFilter, PlayerScope,
-    QuantityExpr, QuantityRef, RepeatContinuation, ResolvedAbility, StaticCondition,
-    StaticDefinition, TargetFilter, TriggerCondition, TriggerDefinition, TypeFilter, TypedFilter,
-    ZoneRef,
+    QuantityExpr, QuantityRef, RepeatContinuation, ReplacementDefinition, ResolvedAbility,
+    StaticCondition, StaticDefinition, TargetFilter, TriggerCondition, TriggerDefinition,
+    TypeFilter, TypedFilter, ZoneRef,
 };
 use crate::types::game_state::TargetSelectionConstraint;
 use crate::types::zones::Zone;
@@ -2695,6 +2695,12 @@ fn legacy_continuous_modification(m: &ContinuousModification) -> bool {
             legacy_static_definition(definition)
         }
         ContinuousModification::GrantTrigger { trigger } => legacy_trigger_definition(trigger),
+        // A granted object-hosted replacement can nest a frozen tag in its
+        // execute body or its `valid_card` scope filter — distinct traversal from
+        // GrantTrigger (a ReplacementDefinition, not a TriggerDefinition).
+        ContinuousModification::GrantReplacement { replacement } => {
+            legacy_replacement_definition(replacement)
+        }
         ContinuousModification::GrantAllActivatedAbilitiesOf { source, .. }
         | ContinuousModification::GrantAllTriggeredAbilitiesOf { source } => {
             legacy_target_filter(source)
@@ -2756,6 +2762,14 @@ fn legacy_continuous_modification(m: &ContinuousModification) -> bool {
         | ContinuousModification::SetStartingLoyalty { .. }
         | ContinuousModification::RemoveManaCost => false,
     }
+}
+
+/// A granted object-hosted `ReplacementDefinition` can carry a frozen tag in its
+/// `execute` redirect body or its `valid_card` scope filter. Mirrors
+/// `legacy_trigger_definition` for the replacement-granting layer-6 case.
+fn legacy_replacement_definition(rd: &ReplacementDefinition) -> bool {
+    rd.execute.as_deref().is_some_and(legacy_definition)
+        || rd.valid_card.as_ref().is_some_and(legacy_target_filter)
 }
 
 /// A granted / emblem `TriggerDefinition` can carry a frozen tag in its firing
--- a/crates/engine/src/game/ability_scan.rs
+++ b/crates/engine/src/game/ability_scan.rs
@@ -5003,6 +5003,10 @@ fn scan_continuous_modification(m: &ContinuousModification, mode: ScanMode) -> A
         // walker is a follow-up.
         ContinuousModification::CopyValues { .. }
         | ContinuousModification::GrantTrigger { .. }
+        // A granted object-hosted replacement's `ReplacementDefinition` execute
+        // is outside the scanner's traversal closure — fail-closed CONSERVATIVE,
+        // same class as GrantTrigger / GrantStaticAbility.
+        | ContinuousModification::GrantReplacement { .. }
         | ContinuousModification::GrantAllActivatedAbilitiesOf { .. }
         | ContinuousModification::GrantAllTriggeredAbilitiesOf { .. }
         | ContinuousModification::AddStaticMode { .. }
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -4194,6 +4194,7 @@ fn fmt_modification(m: &crate::types::ability::ContinuousModification) -> String
             format!("grant all triggered abilities of {}", fmt_target(source))
         }
         ContinuousModification::GrantTrigger { .. } => "grant trigger".into(),
+        ContinuousModification::GrantReplacement { .. } => "grant replacement".into(),
         ContinuousModification::RemoveAllAbilities => "remove all abilities".into(),
         ContinuousModification::AddType { core_type } => {
             format!("add type {}", fmt_core_type(core_type))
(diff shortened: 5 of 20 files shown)


michiot05:
Heads up — I force-pushed a clean rebase onto current `main` (`54fa471de`), which replaced your `Merge origin/main` commit (`293afbea0`). Apologies for overwriting your shepherding commit; flagging exactly what happened and why, and happy to defer to your preference on how the branch is updated going forward.

The CI red here wasn't just a stale test — it's a genuine semantic conflict with our own #6538, which merged after this branch was cut. #6538 now stamps `.expiry(UntilHostLeavesPlay)` inside `try_parse_leave_battlefield_exile_replacement`, the shared parser fn this PR routes through. A straight `merge main` (your `293afbea0`) still failed the same 2 checks, because the granted path would then inherit that host-lifetime stamp — and #6538's machinery keys base-install off exactly that stamp (`is_runtime_host_lifetime_replacement`), so a granted def would be base-installed and **outlive the grant**, breaking Elemental Expressionist's until-end-of-turn lapse. That's the opposite of what this PR is for.

The rebase resolves it by splitting the lifetime decision per consumer:
- `leave_battlefield_exile_replacement()` (the shared constructor) now returns the **unstamped** base def.
- The **standalone** front door and **unearth** synthesis compose `.expiry(UntilHostLeavesPlay)` themselves — byte-identical to #6538 (asserted by `leave_battlefield_exile_replacement_matches_6538_shape`, updated to post-#6538 reality).
- The **granted** path uses the parser fn as a detector only and grants the **unstamped** def, so the grant's own duration governs (CR 611.2a / 613.1f) — permanent grants persist, Expressionist's until-end-of-turn lapses at cleanup.

Pinned by a new revert-failing guard, `granted_replacement_is_not_host_lifetime_stamped`, so a future refactor can't silently re-stamp the granted path. Branch is now linear on `main`, all `engine` lib + `issue_6566` integration tests pass locally, and the 5-card parse-diff is unchanged (`elemental expressionist`, `geth, thane of contracts`, `llanowar greenwidow`, `realmbreaker`, `spirit-sister's call`). Gate A/G PASS.

If you'd rather drive branch updates yourself (or prefer I never force-push over a maintainer commit), just say the word and I'll work to whatever process you want.

Model: claude-opus-4-8

matthewevans (approved):
Reviewed at head `54fa471dec40`. This is excellent work, and it lands at the right seam. Approving.

## What I verified (not just read in the description)

- **The seam is correct.** `GrantReplacement` is applied in the layer-6 continuous-modification loop (`game/layers.rs`) into `obj.replacement_definitions` with a structural-dedup idempotency guard, exactly mirroring the `GrantStaticAbility`/`AddStaticMode` siblings. The "no lifetime stamp needed" argument holds because `replacement_definitions` is reseeded from `base_replacement_definitions` every pass (`layers.rs:1692`, pre-existing) — so the grant's persistence is governed by whether the continuous effect still applies that pass, not an object-lifetime stamp. This is genuinely distinct from #6538's host-lifetime path and correctly leaves it untouched.
- **The promotion guard is right and well-fenced.** `effect.rs::resolve` promotes to `Duration::Permanent` only when `duration_from_fallback && !modifications.is_empty() && modifications.iter().all(GrantReplacement)`. The `!is_empty()` conjunct stops an empty list from vacuously satisfying `.all()`, and the `.all()` predicate is what keeps Expressionist's mixed `[GrantReplacement, GrantTrigger]` from being promoted. CR 611.2a lifetime semantics, applied correctly.
- **The tests discriminate.** `mixed_replacement_and_trigger_grant_is_not_promoted_to_permanent` is genuinely revert-sensitive — swapping `.all` → `.any` promotes the mixed def and reds only that row. The Geth test drives the runtime pipeline (reanimate → destroy → **exile**) with a paired un-granted control creature going to the graveyard as the reach guard, which is what proves the replacement is doing the work rather than a blanket behavior. This is exactly the discriminating-test bar.
- **CR citations verified** against `docs/MagicCompRules.txt`: 611.2a (continuous-effect lifetime), 614.1a ("instead" = replacement), 613.1f (Layer 6 ability-adding), 201.5b (name→gained-ability rewrite, which is what the `~` blocker fix rests on), 514.2 (EOT cleanup). All accurate.
- **The parse-diff proves generality.** 5 cards flip in lockstep from `grant ability` → `grant replacement` (Geth, Realmbreaker, Llanowar Greenwidow, Spirit-Sister's Call, Elemental Expressionist). This is a class fix, not a card patch.

The `~` blocker writeup (that `normalize_card_name_refs` rewrites `this creature/permanent/land` → `~` card-wide, so the rider had to grow a `~` arm to parse at all) is a nice catch, and confronting all 74 "would leave the battlefield" corpus cards to show no standalone rider uses a `this <type>` subject is the right way to prove the #6538 front door is unaffected.

## On sequencing with #5762

Approving and enqueuing this answers your sequencing question: **this lands first, and #5762 rebases onto it.** Your scoping is correct — The Eighth Doctor is `Effect::Unimplemented` at the permission seam #5762 fixes, never reaches this classifier, and doesn't appear in your parse-diff, so the two are complementary rather than competing. The only overlap is textual (`add_target_replacement.rs`, `oracle_effect/mod.rs`); the queue's own CI will surface any conflict if #5762 is retargeted before this merges. No need to hold.

Nice work — the seam reasoning, the three revert-failing regressions, and parameterizing `GrantReplacement` as a member of the existing grant family (rather than special-casing the redirect) are exactly what I want to see.

PR #6595, opened 2026-07-24
fix(client): honor List in lobby for LobbyOnly / P2P broker hosts

### LINKED ISSUE #6556: Game shown in lobby even if "List in lobby" is disabled
When I'm hosting a game and disable `List in lobby` the game is still shown in the lobby for others. I would expect that it won't be visible there and that player can only join by entering the code.

marktech0813:
## Summary
- Fixes #6556: HostSetup's **List in lobby** toggle was ignored on LobbyOnly servers (and the GameProvider broker path), which always registered with `public: true` and set `hostIsPublic` from `useBroker`.
- Private hosts still register with the broker so friends can join by code; they are no longer broadcast via `LobbyGameAdded` / `public_games()`.

## Changes
- `startP2PHostingSession`: pass `settings.public` to `registerHost`; keep `hostIsPublic = useBroker && settings.public` through both connecting and waiting states.
- `GameProvider` P2P host broker registration: use `store.hostIsPublic` instead of hardcoding `true`.
- Regression test in `multiplayerStore.test.ts`.

## Validation
- `pnpm exec vitest run src/stores/__tests__/multiplayerStore.test.ts` — 19 passed.
- Non-engine client/transport change (CONTRIBUTING narrow exception); no `crates/engine/` edits.

## Test plan
- [ ] Host against a LobbyOnly server with **List in lobby** off → room absent from public lobby list
- [ ] Join the same room by entering the code
- [ ] Host with **List in lobby** on → room appears as before

Model: Composer

## Summary by CodeRabbit

- **Bug Fixes**
  - P2P hosts now respect the “List in lobby” visibility setting when registering with the broker.
  - Private, lobby-only sessions remain hidden from the public lobby while still supporting join-by-code.
  - Host visibility status now stays consistent throughout the hosting flow.

- **Tests**
  - Added coverage verifying that private broker-hosted sessions are registered as non-public.

### DIFF
--- a/client/src/providers/GameProvider.tsx
+++ b/client/src/providers/GameProvider.tsx
@@ -867,7 +867,8 @@ export function GameProvider({
                 hostPeerId: host.peer.id,
                 deck: deckList.player,
                 displayName: store.displayName || "Host",
-                public: true,
+                // Honor HostSetup's "List in lobby" toggle (issue #6556).
+                public: store.hostIsPublic,
                 password: null,
                 timerSeconds: null,
                 playerCount: effectivePlayerCount,
--- a/client/src/stores/__tests__/multiplayerStore.test.ts
+++ b/client/src/stores/__tests__/multiplayerStore.test.ts
@@ -51,6 +51,16 @@ const p2pMocks = vi.hoisted(() => ({
   dispose: vi.fn(),
 }));
 
+const brokerMocks = vi.hoisted(() => ({
+  registerHost: vi.fn(async () => ({
+    gameCode: "PRIV1",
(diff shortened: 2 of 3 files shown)


marktech0813:
Closing per docs/AI-CONTRIBUTOR.md §0.1.1: this PR was authored in a Composer session, and Composer is not on the Frontier tier accepted for PRs opened on/after 2026-07-24. Reporting accurately rather than misdeclaring the model.

The fix itself is on branch `fix/6556-respect-list-in-lobby` if a Frontier-tier agent wants to re-verify and reopen.

PR #6607, opened 2026-07-24
fix(engine): a player-directed "exile a creature they control" is a choice, not a target (Strategic Betrayal)

### LINKED ISSUE #6505: Strategic Betrayal — Ward triggers when opponent chooses their own creature
Reported in Discord: https://discord.com/channels/1485498006781427802/1528012592746860715/1528012592746860715

**Thread:** Strategic Betrayal — Ward triggers when opponent chooses their own creature
**Discord thread id:** `1528012592746860715`

## Summary
This is an interesting one.  Targetted the opponent and it's triggering the ward on the creature itself.

## Reported text

This is an interesting one.  Targetted the opponent and it's triggering the ward on the creature itself.

The opponent must select their creature meaning ward should not trigger (I'm rather sure)?

## Attachments

- [image.png](https://cdn.discordapp.com/attachments/1528012592746860715/1528012593112027216/image.png?ex=6a62061f&is=6a60b49f&hm=8a16f5c5109eff9a74f8cc3226b9f50a4c8a32fcc1892e32320d5fd6223b59cd&)

### strategic betrayal
Target opponent exiles a creature they control and their graveyard.

---
<sub>report_id: `discord:1528012592746860715:1528012592746860715:0` · discord: `1528012592746860715/1528012592746860715`</sub>

michiot05:
Tier: Frontier
Model: claude-opus-4-8

Closes #6505.

## Summary

**Strategic Betrayal** — *"Target opponent exiles a creature they control and their graveyard."* — triggered **Ward** on the opponent's own creature. The spell targets only the opponent (CR 115.1a); the opponent then **chooses** (does not target) a creature they control to exile, plus exiles their graveyard. Per **CR 115.10a/115.10b** an object affected but not identified by "target" is not a target, so **Ward (CR 702.21a) must not fire**, and **CR 109.4** makes the *target player* the actor/chooser. The engine was resolving "exile a creature they control" as a **targeted** exile with the caster as actor.

Two coupled defects, each fixed at its seam:

**1. Parser origin-inference leak (parse-visible).** The compound-exile origin was inferred from the whole post-verb text, so the trailing *"and their graveyard"* conjunct leaked `origin: Graveyard` onto the creature leg (which should be Battlefield). Origin is now inferred from the **primary-clause slice only** — a small `compound_exile_origin_scan` helper that trims a trailing `and <conjunct>` before `infer_origin_zone`, applied uniformly across the three compound-exile sites (`imperative.rs`, and the single- and mass-exile arms in `oracle_effect/mod.rs`). For any non-compound clause it is byte-identical to the previous behavior.

**2. Target-vs-choice + actor binding.** Even with origin fixed, the creature leg was emitted as a stack target with the caster as actor. In the "target player exiles …" arm the anaphoric *"they control"* scope is now pinned to `ScopedPlayer` (unifying the creature leg with the already-`ScopedPlayer` *"their graveyard"* leg), and the battlefield pick is stamped `TargetChoiceTiming::Resolution`, making it a **non-targeted, resolution-time player-directed choice** (no `BecomesTarget` → no Ward). At resolution, an ability with **exactly one `TargetRef::Player`** target and a `ScopedPlayer`-scoped move binds `scoped_player` to that target player — a sibling of the engine's existing trigger-event `scoped_player` stamp — so the **target opponent** is the chooser and **their own** graveyard is exiled. This mirrors the sacrifice-edict actor rule (CR 109.4) using the existing `scoped_player` channel.

**No new enum variant** (`ControllerRef::ScopedPlayer`, `TargetChoiceTiming::Resolution`, `scoped_player` all exist). `controller_for_relative_filter`, `sacrifice.rs`, and `rebind_owned_scope` are **untouched** — the actor binding keys on the `ScopedPlayer` scope produced only by this arm, so a caster-subject *"you exile a creature target opponent controls"* (which yields a `TargetOpponent`-controller filter, not `ScopedPlayer`) is never stamped and stays caster-chosen.

The fix **generalizes to the class**: **Leadership Vacuum** (*"target player returns each commander they control …"*) now correctly returns the **target player's** commanders (was the caster's) via the same `ScopedPlayer` + single-Player-target stamp.

Method: /engine-implementer — plan (/engine-planner, with a run-then-delete reproduction probe confirming the Ward stall) → /review-engine-plan (**three adversarial rounds**; round 1 corrected the sibling-exile site coverage, round 2 rejected an over-broad `rebind_owned_scope` extension and a runtime-invisible blast-radius net, round 3 re-seated Part B on the `ScopedPlayer`/`scoped_player` actor channel instead of keying the shared `controller_for_relative_filter` on a moved-object filter controller) → implement (tests-first RED→GREEN) → /review-impl (Maintainer-Simulation Gate: **PASS**, all six executor deviations validated sound, the `leadership_vacuum` assertion change confirmed a legitimate generalization not a masked regression). Each step in a fresh agent context.

## Gate A

`./scripts/check-parser-combinators.sh` → **Gate A PASS head=2e9db496d** (Gate G PASS). Part A is arithmetic slicing of nom remainders (the conjunct trim uses `tag("and ")`); Part B is typed scope/timing/actor stamping. No string-dispatch introduced.

## Anchored on

- **Sacrifice edict** (`game/effects/sacrifice.rs::resolve_sacrifice_scope`, CR 109.4) — the proven "target player is the actor/chooser" precedent; this fix reuses the same `scoped_player` binding for the exile flavor rather than duplicating it.
- **`stack.rs` `resolve_top`** — the existing trigger-event `scoped_player` stamp (DamageDealt/AttackersDeclared referent); the new single-`TargetRef::Player` stamp is a sibling in the same seam, so the target player flows to the resolution-time chooser via the *unchanged* `controller_for_relative_filter` `ScopedPlayer` branch.
- **`oracle_effect/lower.rs::change_zone_selects_battlefield_permanent`** (promoted `pub(super)`, reused not duplicated) — the single authority for "is this a battlefield, non-targeted pick"; it gates both the Resolution-timing stamp and the rebind-skip off one shared boolean, so timing and controller can never diverge.

## Verification

- `cargo fmt --all` clean; `cargo clippy -p engine --tests` **0 warnings**; `cargo test -p engine --lib` **17,648 passed / 0 failed**; `cargo test -p engine --test integration -- strategic_betrayal_6505` **10 passed**.
- **Integration** (`crates/engine/tests/integration/strategic_betrayal_6505.rs`, registered in `main.rs`) — all through the real cast/stack pipeline (`runner.cast(...).target_player(...).resolve()`), verbatim Oracle text, each negative reach-guarded:
  - **End-to-end**: opponent controls two creatures (one **Ward—{2}**) + graveyard cards → **no** `UnlessPayment`/Ward prompt; the **opponent** is the chooser and is offered its two battlefield creatures; the chosen creature is exiled; the opponent's graveyard is exiled. (reach-guards: parse OK, zero `Effect::Unimplemented`, exiles actually resolved.)
  - **Per-mechanism revert-sensitive isolation**: reverting only the timing stamp re-fires Ward (`left: Stack, right: Resolution`); reverting only the `scoped_player` stamp flips the chooser back to the caster; reverting only the origin slice makes the eligible set the graveyard instead of the battlefield.
  - **Negatives**: `"Exile target creature."` **still** triggers Ward (not over-suppressed); `"exile a creature you control …"` keeps the **caster** as chooser; `"Target player sacrifices a creature."` (Diabolic Edict) still binds the chooser to the target player (sacrifice path unchanged); parser-shape + hostile fixtures (`exile target … and their graveyard` stays a stack target; the mass compound stays coverage-honest fail-closed, no silent orphan).
- **CR annotations** grep-verified against `docs/MagicCompRules.txt`: 115.1a, 115.10a, 115.10b, 702.21a, 109.4, 608.2c, 701.13a, 400.7.

## Claimed parse impact

Intentional behavior changes: **Strategic Betrayal** and **Leadership Vacuum** (same class — target-player-directed "they control"). Part A's origin fix is byte-identical to prior behavior for every non-compound exile clause and only re-scopes the origin of compound *"exile X and &lt;zone-conjunct&gt;"* clauses. I did not have a committed `card-data.json` baseline in this worktree to compute the exact Part-A parse-diff count offline, so I've left that to CI — the **`coverage-parse-diff`** artifact will report the full affected set; I expect it to be the compound-exile family only, with the full `test-engine` suite green. Happy to reconcile against the artifact.

## Scope Expansion

None. Leadership Vacuum is the same target-player-directed class as Strategic Betrayal (the class the fix is defined over), not an unrelated surface. `controller_for_relative_filter`, `sacrifice.rs`, `rebind_owned_scope` untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed “Strategic Betrayal” so the opponent (“they”) is correctly scoped for choosing the exiled creature.
  * Ensured “exiles a creature they control and their graveyard” selects the creature at **resolution** (not as a target), preventing incorrect Ward/payment timing.
  * Improved parsing of Exile text so “and their graveyard” no longer leaks graveyard origin onto the battlefield leg.
  * Corrected moved-object scoping/origin handling for “their graveyard”-style patterns.

* **Tests**
  * Added/expanded Strategic Betrayal integration and parser-shape assertions, including targeted/sibling regression coverage.


matthewevans (changes requested):
**Current-head verdict: blocked on parser scope and targeting evidence.**

## 🔴 Blocker

**[HIGH]** The required parse-diff reports 25 cards changed across 14 signatures, but the PR body and runtime coverage substantiate Strategic Betrayal only. Evidence: the current `<!-- coverage-parse-diff -->` comment lists changes to Dragon's Approach, Debt to the Kami, Leadership Vacuum, Early Harvest, and many others; the new integration module is `strategic_betrayal_6505.rs`. Why it matters: this parser-scope expansion can alter unrelated targeting and zone behavior without a card-by-card claim or discriminating regression. Suggested fix: either narrow the parser change to the intended grammar or explain and test each affected card class against the measured artifact.

**[MED]** `compound_exile_origin_scan` slices a presumed suffix manually. Evidence: `crates/engine/src/parser/oracle_effect/mod.rs:31767-31774` uses `base[..base.len() - rem.len()]`. Suggested fix: use `base.strip_suffix(rem)` before truncating, so a future caller cannot turn a violated internal precondition into a panic.

## 🟡 Non-blocking

The explicit-target hostile fixture is vacuous when no creature leg is found: `strategic_betrayal_6505.rs:564-570` only asserts inside `if let Some(...)`. Make that lookup required.

**Recommendation: request changes; account for the full parse blast radius and close the parser/test gaps before approval.**

matthewevans (approved):
Maintainer review: approved at a323232cfb3fbb8bd909623873651a80178d72cf. The scoped-player/origin class coverage and the exhaustive TargetFilter fix are clean; no outstanding actionable current-head findings.

matthewevans (approved):
Maintainer re-review: approved at 8f26be98382d5c0e8f02346c13f9d48723233747. Current-main integration adds the explicit LastZoneChanged terminal only; recursive and mutating behavior is unchanged.

michiot05:
Thanks — this was a real gap in my accounting; addressed at head `b35b2a86c`. The regenerated parse-diff is now **15 cards / 7 signatures** (down from 25 / 14).

## 🔴 HIGH — parser scope + full blast radius

**Root cause of the over-broad diff:** I was pinning `ctx.relative_player_scope = Some(ScopedPlayer)` around the *entire* `lower_imperative_clause` call, which re-scoped every predicate under a "target player …" subject — including non-ChangeZone ones that have no runtime `scoped_player` binding (so `TargetFilter::ScopedPlayer` would have resolved to `None`). That was both the unexplained blast radius **and** a latent bug.

**Narrowed** to a targeted post-lowering rewrite: only when the lowered leg is the moved-object `ChangeZone`/`ChangeZoneAll` battlefield resolution-pick whose "they control" defaulted to `ControllerRef::You` do I rewrite *that leg's* controller `You → ScopedPlayer` (via a small `rebind_controller_scope`), gated on the predicate actually containing "they control". `rebind_owned_scope`, `controller_for_relative_filter`, and `sacrifice.rs` are untouched. The 11 non-ChangeZone predicates you'd expect — **Consume/Consumed by Greed/Will of the Abzan/Riveteers Charm** (Sacrifice), **Linessa** (Bounce), **Hunted Nightmare** (PutCounter), **Shadrix Silverquill** (PutCounterAll), **Early Harvest** (UntapAll), plus **Blot Out/End of the Hunt** — all revert to baseline and drop out of the diff.

**The remaining 15 cards, accounted:**

*Part B — target-player-directed move (`you control → scoped player controls`), 6 cards.* Each is a genuine "target player {exiles/returns} a permanent **they control**" subject, where the target player is the CR 109.4 actor and `ScopedPlayer` is correct: **Strategic Betrayal**, **Debt to the Kami** (creature + enchantment modes), **Doomfall** (creature mode), **Early Winter** (enchantment mode), **Azula, Cunning Usurper** (nontoken creature), **Leadership Vacuum** (mass `ChangeZoneAll` — "each commander they control"). A caster-subject shape ("you exile a creature target opponent controls") has `controller = TargetOpponent`, not `You`, and no "they control" anaphor, so it is doubly excluded from this rewrite and stays caster-chosen.

*Part A — origin de-leak (`from: graveyard/library → ∅`), 10 cards.* The compound-origin scan only fires when the target-parser remainder begins with `" and …"`, i.e. the graveyard/library keyword lived entirely in a **trimmed trailing conjunct** and the primary leg had no zone phrase of its own. This is a **strict fix**, not a regression, because when `origin` is set, `add_inferred_origin_constraints_to_target` injects an `InZone` constraint that mis-scopes *candidate selection* to the private zone for class-enumerating (Typed/Or) filters:

| Card | primary leg ‖ trimmed conjunct | base→head | why ∅ is correct |
|---|---|---|---|
| Strategic Betrayal | `a creature they control` ‖ `and their graveyard` | Gy→∅ | primary is a battlefield creature; base wrongly `InZone(Graveyard)`-scoped it (the reported bug) |
| The Spot, Living Portal | `up to one target nonland permanent` ‖ `and … from a graveyard` | Gy→∅ | primary leg is a battlefield permanent |
| Shadows' Verdict | `all creatures/PWs … from the battlefield` ‖ `and … from all graveyards` | Gy→∅ | primary explicitly battlefield |
| Once More with Feeling | `all permanents` ‖ `and all cards from all graveyards` | Gy→∅ | "all permanents" = battlefield |
| Ultimate Nullification | `all creatures` ‖ `and graveyards` | Gy→∅ | "all creatures" = battlefield |
| Gilded Ambusher | `a nonland, nontoken permanent they control` ‖ `and a random nonland card from their library` | Lib→∅ | primary is a battlefield permanent |
| Dragon's Approach | `exile Dragon's Approach` ‖ `and four cards … from your graveyard` | Gy→∅ | `SelfRef` = the resolving spell on the stack; base `origin=Graveyard` guard would have made the exile silently no-op — ∅ fixes it; SelfRef selects by object id, not zone, so no battlefield-copy risk |
| Silent Gravestone | `exile this artifact` ‖ `and all cards from all graveyards` | Gy→∅ | `SelfRef` on battlefield; origin never affects selection |
| Spellweaver Volute | `exile the enchanted card` ‖ `and attach … in a graveyard` | Gy→∅ | `ParentTarget` selects by object id; resolver reads its actual zone; guard was redundant |
| Underworld Cerberus | `exile it` ‖ `and each player returns … from their graveyard` | Gy→∅ | `TriggeringSource` selects by object id; harmless |

No card in the set has a primary exile leg that genuinely needs graveyard/library *candidate-zone* scoping — that's the only shape that would make the de-leak a regression — so nothing was narrowed there. (Confirmed against `change_zone.rs`: `from_zone` comes from the object's actual zone; `origin` is only an optional skip-guard.) A few of these compounds separately drop their trimmed graveyard-leg as a **pre-existing** fail-closed coverage gap; Part A doesn't worsen them.

## 🔴 MED — `compound_exile_origin_scan` manual slice

Fixed — now `base.strip_suffix(rem).filter(|_| trailing_and_conjunct)` (fail-closed; a non-suffix `rem` can no longer panic).

## 🟡 Non-blocking — vacuous hostile fixture

Fixed — the explicit-target fixture's creature-leg lookup is now a required `expect(...)`.

Added tests: a runtime sibling (`sibling_bare_target_player_exile_they_control_binds_chooser_no_ward` — opponent is chooser, own creature never offered, Ward silent), `leadership_vacuum_mass_leg_is_scopedplayer`, and `origin_leak_compound_exile_primary_leg_stays_battlefield`. `cargo test -p engine --lib` 17,648 pass; integration 13/13; clippy 0; Gate A green.

matthewevans:
**Re-review complete — no new finding on the current approved head.**

## ✅ Clean

The unchanged head `8f26be98382d5c0e8f02346c13f9d48723233747` remains covered by the prior approval. The author’s complete 15-card/7-signature accounting matches the current `<!-- coverage-parse-diff -->` artifact, including the scoped-player and origin de-leak classes.

**Recommendation: retain the existing approval and merge when the queue’s required checks complete.**

PR #6620, opened 2026-07-25
fix(engine): carry the sacrificed-permanent quality on Devour (Famished Worldsire)

michiot05:
Tier: Frontier
Model: claude-opus-4-8

Closes #6423.

## Summary

**Famished Worldsire** — *"Devour land 3 (As this creature enters, you may sacrifice any number of lands. It enters with three times that many +1/+1 counters on it.)"* — let you sacrifice **creatures** instead of **lands**. Two dropped halves of the same axis: `Keyword::Devour(u32)` carried only the count, discarding the *"[quality]"* of CR 702.82c (`land`), and `synthesize_devour` hard-coded the ETB replacement's `Effect::Sacrifice` pool to `TypedFilter::creature()`. (Merged #5362 had already captured the *count* from "Devour land 3" while discarding "land" — this is the other half.)

Parameterize the keyword with the quality axis:

- **`Keyword::Devour(u32)` → `Keyword::Devour { n: u32, quality: TypeFilter }`**, defaulting to `TypeFilter::Creature` — the CR **702.82a** plain-Devour behavior. `quality` is a non-optional `TypeFilter` (not `Option`) because CR 702.82a defines a *concrete* default quality, not an absent one; and `TypeFilter` is the one layer that spans the whole corpus — core types **and** subtypes.
- **Parser**: a dedicated `parse_devour_keyword_line` (composed nom combinators — `tag("devour ")` → `opt(parse_type_filter_word + " ")` → `parse_number`) captures the optional quality word. Plain "Devour N" → `Creature`; "Devour land N" → `Land`; "Devour artifact N" → `Artifact`; "Devour Food N" → `Subtype("Food")`.
- **Synthesizer**: threads `quality` into **both** the `Sacrifice` target and the sacrificed-count `ObjectCount` filter. Counter math (`Multiply { factor: n }` × permanents sacrificed, CR 702.82c) is unchanged. `is_devour_etb_replacement` now discriminates on the quality so a same-`N` land-quality and creature-quality replacement can't false-dedupe.

Result across the whole 4-valued corpus (24 Devour cards): **Famished Worldsire → lands, Caprichrome → artifacts, Feasting Hobbit → Foods, plain Devour (20 cards) → creatures.**

This reshape was pre-sanctioned by the codebase itself — an in-code comment at `synthesize_devour` stated *"A future card needing the quality axis requires parameterizing the keyword to `Devour { n, quality }`"*; that comment is now updated to reflect that the axis is modeled.

## Implementation method (required)

[...]

PR #6621, opened 2026-07-25
fix(parser): bind "that token" inside a reflexive "when you do" body (CR 603.12)

lgray:

## Summary

`chain_prior_referent_is_created_token` bailed `return false` on the first prior clause carrying a
`condition`, before it ever reached the `is_token_creating_effect` check. In a reflexive body the
token-creating clause is exactly the clause that carries the gate — `strip_if_you_do_conditional`
stamps `WhenYouDo` / `EffectOutcome{OptionalEffectPerformed}` onto it — so a later "that token"
anaphor in the same body never saw the token and misbound. **Iroh, Tea Master** placed its `+1/+1`
counter on an unbound `ParentTarget` instead of the Ally it had just created; **Summoner's Sending**
bound `TrackedSet{0}`.

The referent walk now continues past a prior clause whose condition is an affirmative reflexive gate,
and the consuming clause is re-linked `SequentialSibling` → `ContinuationStep` so the resolver's
condition-false descent cannot resolve it against `state.last_created_token_ids` — a game-lifetime
ledger that would otherwise hand it a token from an earlier resolution (CR 609.3: no legal
antecedent).

Built as building blocks, not special cases:

- `AbilityCondition::is_affirmative_reflexive_gate` — exhaustive 53-arm match, **no wildcard**, so a
  future variant is a compile error rather than a silent reclassification. Its `true` set is exactly
  the image of `parse_affirmative_reflexive_connector`.
- `AbilityDefinition::is_self_gated_reflexive` — a clone that carries the gate itself is honest
  wherever it lands.
- `oracle_ir::ast::sub_link_after_boundary` and `lower::instruction_spine_is_continuation` extracted
  so the seeder and the re-link share one authority instead of two drifting copies.
- The repeat-process clone guard asks its question directly: it builds the def the assembler would
  push (via the shared `repeat_process_clone_shape`), runs the re-link over it, and reads the
  resulting link back, rather than stacking structural proxies.

## Files changed

- `crates/engine/src/parser/oracle_effect/mod.rs` — seeder carve-out; `repeat_process_clone_shape`
  extracted as the single clone-shape authority
- `crates/engine/src/parser/oracle_effect/lower.rs` — `relink_gated_token_referent_consumers`,
  `gated_instruction_reaches`, `instruction_spine_is_continuation`,
  `clone_would_transplant_gated_referent`
- `crates/engine/src/parser/oracle_effect/assembly.rs` — re-link wired into chain assembly
- `crates/engine/src/parser/oracle_ir/ast.rs` — `sub_link_after_boundary` shared authority
- `crates/engine/src/types/ability.rs` — `AbilityCondition::is_affirmative_reflexive_gate`,
  `AbilityDefinition::is_self_gated_reflexive`
- `crates/engine/tests/integration/reflexive_body_token_referent.rs` — new, 21 durable tests
- `crates/engine/tests/integration/main.rs` — `mod` line

### Note on #6533 (`SiblingCondition::ReplicatedOrBranch`)

`attach_repeat_process_keywords` is refactored here into the shared `repeat_process_clone_shape`,
because `clone_would_transplant_gated_referent` decides whether a binding is safe by building this
exact def and asking the re-link what it would do with it — if the two shapes diverged, the guard
would be answering a question about a def the assembler never pushes.

That loop is also where #6533 set `sibling_condition = SiblingCondition::ReplicatedOrBranch`, so the
assignment is **folded into the shared helper** (with the CR 608.2c rationale recorded there) rather
than left at the push site. Verified by revert-probe rather than asserted: dropping that one line from
the helper takes the suite from **29/29 → 27/29**, failing exactly
`mutable_pupa_perpetual_keyword_mirror::kathril_reaches_matching_counter_and_tail_past_false_earlier_gates`
and `oracle_ir::snapshot_tests::kathril_aspect_warper` — #6533's own behavioural + AST tests. The
separate perpetual-grant site is untouched, and all 21 tests here stay green in the drop arm, so
`ReplicatedOrBranch` is orthogonal to this PR's guard rather than interacting with it.

## Track

Developer

## LLM

Model: claude-opus-5
Tier: Frontier
Thinking: max (planning / plan-review), high (implementation / impl-review)

Method: /engine-implementer

## CR references

- `CR 111.1` — tokens
- `CR 603.12` — reflexive triggered abilities
- `CR 608.2c` — instructions carried out in the order written
- `CR 609.3` — an effect does only as much as it can; no legal antecedent
- `CR 701.23` — Search
- `CR 701.24` — Shuffle

All six re-grepped against `docs/MagicCompRules.txt`; 0 unverified.

## Verification

All re-run on this head after rebasing onto `b3bd4f910` (#6607 touches three of the four files this
PR changes). The rebase applied with zero conflicts, which is precisely why the full sweep was re-run
rather than carried forward — the post-rebase hazard is semantic drift a 3-way merge *accepts*.

- `cargo nextest run -p engine` — **21710 passed / 0 failed**, 8 skipped
- `cargo clippy --workspace --all-targets --features engine/proptest -- -D warnings` — **0 warnings**
- `cargo fmt --all --check` — exit 0
- `scripts/check-parser-combinators.sh b3bd4f910` — **Gate G PASS + Gate A PASS**; no string-dispatch
  lines added
- combo/loop detector suite (`-E 'test(combo) or test(loop_shortcut) or test(loop_counter_growth) or
  test(chicago_loop)'`) — **111 passed / 0 failed**
- `reflexive_body_token_referent` durable suite — **21 passed / 0 failed**
- upstream drift tripwire (`-E 'test(strategic_betrayal) or test(kathril) or test(mutable_pupa)'`) —
  **21 passed / 0 failed**. #6607's own suite plus the two Kathril suites that #6533 added, both of
  which run through the `repeat_process_clone_shape` helper this PR extracts.
- `cargo combo-verify` BASE vs POST — rows **byte-identical**
  (13 confirmed / 4 gated / 37 deferred / 0 failed of 54). **Non-vacuous:** the two regenerated
  `card-data.json` exports are confirmed to DIFFER, so the row identity is not a stale-binary artifact.

Every one of the 21 tests carries an **executed** revert-probe that flips, and every negative
assertion carries a reach-guard (verified by breaking its premise: T5's guard fires when
`select_targets_and_resolve` is removed, T4's when `base_controller` is dropped). Fixture fidelity:
all 40 embedded card texts diffed byte-verbatim against `data/mtgjson/AtomicCards.json`, 0 mismatches.

## Gate A

Gate A PASS head=044574f3579e1128422ae220b91bfe4c3da9a7b1 base=b3bd4f910720b7141331dee9f18f81aa539dad5e

## Anchored on

- `crates/engine/src/parser/oracle_effect/lower.rs:347` — `gate_reflexive_rider_on_declined_optional_target`,
  the existing post-lowering whole-definition pass that gates a reflexive rider on a declined optional
  antecedent (also annotated CR 608.2c). Same seam, same shape as `relink_gated_token_referent_consumers`.
- `crates/engine/src/parser/oracle_effect/lower.rs:296` — `patch_choose_from_zone_counter_continuation_target`,
  the existing post-lowering pass that rebinds a continuation's target when the head exposes the only
  legal referent. Same seam, same class-level ("general across the whole class") justification.

## Claimed parse impact

**20 cards, 25 chain nodes, every one `SequentialSibling` → `ContinuationStep`, and exactly 2
`PutCounter.target` moves. No node added or removed on any card.**

Two measurements, stated separately because they have different denominators and artifacts:

- **The 20-card set** — re-measured on the current head against the shipped `card-data.json`
  (35477 entries), both arms regenerated from the identical feature set
  (`--profile tool --features cli,combo-verify`) so the commit is the only variable.
  Non-vacuity control: the two exports are confirmed to DIFFER, so neither arm ran a stale binary.
- **The 25-node / 2-target-move decomposition** — measured at the AST level over all 34348
  `AtomicCards.json` entries. Post-rebase it is carried by `T18` (invariant: 0 `SequentialSibling`
  nodes read `LastCreated`) and `T21` (per-card `PutCounter` multisets), both green on this head.

Akoum Stonewaker · Blight Herder · Boulder Jockey · Cadric, Soul Kindler · Dalek Intensive Care ·
Felhide Spiritbinder · Flameshadow Conjuring · God-Pharaoh's Gift · Gyrus, Waker of Corpses ·
Inalla, Archmage Ritualist · **Iroh, Tea Master** · Kavaron Harrier · Krenko, Baron of Tin Street ·
Rebellion of the Flamekin · Saheeli, Radiant Creator · Séance · **Summoner's Sending** ·
Timothar, Baron of Bats · Ultron, Artificial Malevolence · Vile Redeemer

The two bolded cards are the `PutCounter.target` moves — the bug being fixed. The other 18 are the
`sub_link` re-tag only, which changes which parent the resolver descends from without changing what
resolves. `T18` freezes this exact set as a corpus-complete stale-bind invariant.

## Scope Expansion

None.

## Validation Failures

None.

## CI Failures

None.

---

_Anchored on `b3bd4f910` (#6607). `799b34220` (#6620, Devour) landed while the final sweep was
running; it touches `triggers.rs` / `oracle_keyword.rs` / `types/keywords.rs` / `synthesis.rs` /
`ability_scan.rs` and `mtgish-import`, none of which this PR touches — zero file overlap, so the
verification above stands on the stated base rather than being silently carried across it._

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed “Last Created” token binding so clauses after affirmative reflexive gates resolve to the correct gated continuation, not stale antecedents.
  - Prevented conditional decline branches from reusing tokens created on other paths.
  - Improved Repeat/replication handling to ensure cloned counter-placement stays aligned with the right gated publisher path.
  - Refined clause continuation and boundary handling for more accurate ability resolution.
- **Documentation**
  - Clarified how sub-clause links and reflexive gate conditions affect token resolution.
- **Tests**
  - Added a comprehensive integration test suite covering reflexive “Last Created” relinking and replication scenarios.

### DIFF
--- a/crates/engine/src/parser/oracle_effect/assembly.rs
+++ b/crates/engine/src/parser/oracle_effect/assembly.rs
@@ -38,10 +38,11 @@ use super::lower::{
     attach_cast_cost_raise_to_previous_play_from_exile,
     attach_graveyard_redirect_rider_to_prior_cast_from_zone,
     attach_land_enters_tapped_to_previous_play_from_exile, cast_cost_raise_rider,
-    consolidate_die_and_coin_defs, definition_targets_self_source,
-    effect_publishes_revealed_subject, extract_bounded_target_multi_target,
-    extract_exact_target_multi_target, extract_optional_target_multi_target,
-    extract_verb_up_to_multi_target, fold_copy_spell_gains_haste_and_quoted_grant,
+    clone_would_transplant_gated_referent, consolidate_die_and_coin_defs,
+    definition_targets_self_source, effect_publishes_revealed_subject,
+    extract_bounded_target_multi_target, extract_exact_target_multi_target,
+    extract_optional_target_multi_target, extract_verb_up_to_multi_target,
+    fold_copy_spell_gains_haste_and_quoted_grant,
     fold_deal_damage_then_prevent_into_computed_amount, fold_enters_this_way_counter_rider,
     fold_exile_resolving_rider, fold_search_choose_type_conditional_destination,
     fold_token_it_has_grants_into_token_statics, gate_other_revealed_card_on_multiplayer_reveal,
@@ -53,13 +54,13 @@ use super::lower::{
     parse_controlled_by_different_players_target_constraint,
     parse_same_zone_owner_target_constraint, parse_total_mana_value_target_constraint,
     patch_choose_from_zone_counter_continuation_target, patch_population_head_tap_anaphor,
-    patch_self_ref_head_tap_anaphor, resolve_populated_token_anaphors,
-    resolve_populated_unsuspect_anaphors, resolve_those_tokens_anaphors,
-    rewire_result_anchored_subchain, rewrite_counter_instead_target_from_antecedent,
-    rewrite_else_event_context_to_stable, rewrite_else_parent_target_to_self_ref,
-    rewrite_player_anaphor_targets_in_definition, rewrite_those_tokens_from_antecedent,
-    rewrite_two_target_counter_chain, target_choice_timing_for_clause,
-    thread_chosen_damage_source_into_oneshot_effects,
+    patch_self_ref_head_tap_anaphor, relink_gated_token_referent_consumers,
+    resolve_populated_token_anaphors, resolve_populated_unsuspect_anaphors,
+    resolve_those_tokens_anaphors, rewire_result_anchored_subchain,
+    rewrite_counter_instead_target_from_antecedent, rewrite_else_event_context_to_stable,
+    rewrite_else_parent_target_to_self_ref, rewrite_player_anaphor_targets_in_definition,
+    rewrite_those_tokens_from_antecedent, rewrite_two_target_counter_chain,
+    target_choice_timing_for_clause, thread_chosen_damage_source_into_oneshot_effects,
 };
 use super::sequence::{apply_clause_continuation, def_bears_retargetable_copy};
 use super::{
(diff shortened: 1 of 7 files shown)


matthewevans (approved):
Approved after maintainer repair and exact-head review.

Evidence: the recursive `TargetFilter` walk now covers `Not`, `TrackedSetFiltered`, and `ChosenDamageSource` with explicit terminal arms; `mode_abilities` are traversed; wrapper/modal regressions exercise the re-link; the fixture comment now matches the mana/energy setup. `cargo fmt --all`, diff check, security review, and pre-commit parser gates passed. No direct builds/tests/clippy were run.

PR #6626, opened 2026-07-25
fix(engine): honor the power gate on Goreclaw's spell cost reduction

### LINKED ISSUE #5606: [Card Bug] The Scarlet Witch: Everything becomes cheaper to cast instead of only instants and sorceries

### LINKED ISSUE #6375: Goreclaw lowers all spells by 2 colorless not just 4/+ power creatures — [[Goreclaw, terror of qal sisma]]
Reported in Discord: https://discord.com/channels/1485498006781427802/1527520876810211388/1527521138282987702

**Thread:** Goreclaw lowers all spells by 2 colorless not just 4/+ power creatures
**Discord thread id:** `1527520876810211388`
**Discord message id:** `1527521138282987702`
**Cards:** goreclaw, terror of qal sisma

## Summary
[[Goreclaw, terror of qal sisma]]

### goreclaw, terror of qal sisma
Creature spells you cast with power 4 or greater cost {2} less to cast.
Whenever Goreclaw attacks, each creature you control with power 4 or greater gets +1/+1 and gains trample until end of turn.

---
<sub>report_id: `discord:1527520876810211388:1527521138282987702:0` · discord: `1527520876810211388/1527521138282987702`</sub>

michiot05:
Tier: Frontier
Model: claude-opus-4-8

Closes #6375.

## Summary

**Goreclaw, Terror of Qal Sisma** — *"Creature spells you cast with power 4 or greater cost {2} less to cast."* — reduced the cost of **every** spell you cast. It's a **two-layer** defect (the parser-only fix alone would have flipped it to reducing *nothing*):

**1. Parser — the power gate was dropped.** The cost-reduction static parsed with `spell_filter: null` over `affected: Typed{Card}`, so it matched all your spells. Fix #5606 added a *"[type] spells with mana value N or greater/less"* qualifier arm to the cost-modifier parser; there was no *"with power N"* counterpart, so the power gate fell through. Added a parallel **`parse_cost_mod_power_qualifier`** in `oracle_static/static_helpers.rs`, mirroring the mana-value arm and **reusing the canonical `parse_pt_comparison` combinator** (the same building block Goreclaw's own attack trigger already uses — no hand-rolled power parser). Goreclaw now parses `spell_filter = Typed{ Creature, [PtComparison{ Power, Current, GE, Fixed(4) }] }`.

**2. Runtime — the cost-modifier matcher fail-closed on power.** Even with a correct `spell_filter`, the cost path evaluates a candidate spell against a **`SpellCastRecord` snapshot** (`spell_object_matches_property` → `spell_record_matches_property`) that carries **no power/toughness**, so `FilterProp::PtComparison` sat in the fail-closed `=> false` group → Goreclaw would reduce nothing. Added a `PtComparison` arm to **`spell_object_matches_property`** (`game/filter.rs`) that reads the **live spell object's** printed P/T via the filter context (`context.spell_object_id → object_pt_value`, the same authority the battlefield matcher uses), evaluating the comparator. It is **guarded** to the context-bearing callers (cost / can't-cast / keyword-grant, which supply `Some(spell_object_id)`); every snapshot-only / `None`-context caller returns `false` exactly as before, so the fail-closed record path is unchanged. **No `SpellCastRecord`/serde change, no new enum variant.**

Result: a power-4+ creature spell is reduced; a power-3 creature spell and any noncreature spell are not (a noncreature is rejected by the type gate, and independently its power `None → 0` fails `GE 4` — CR 208.3). Scope `Current` is correct for an unresolved spell (= printed power; per Goreclaw's ruling, later +1/+1 counters don't retroactively count).

Method: /engine-implementer — plan (/engine-planner, with a run-then-delete reproduction probe confirming power-3/noncreature spells were wrongly reduced at base) → /review-engine-plan → implement (tests-first). During implementation, driving the **real** production cost path (not a force-set `power`) empirically falsified the plan's "parser-only-safe" premise — the cost matcher reads a power-less snapshot — so the executor stopped rather than ship a fix that reduced nothing. A focused design-and-validation pass then pinned the runtime seam (traced the cost path decisively; ruled over-broadening safe by enumerating every caller of the shared matcher) → implement runtime arm → /review-impl (Maintainer-Simulation Gate PASS; two LOW notes, non-gating). Each step in a fresh agent context.

## Gate A

`./scripts/check-parser-combinators.sh` → **Gate A PASS head=ddbeec86d** (Gate G PASS). The parser peel is `take_until` + `parse_pt_comparison` (pure combinators).

## Anchored on

- **The #5606 mana-value cost-mod qualifier** (`oracle_static/static_helpers.rs`: `parse_cost_mod_mana_value_qualifier`/`strip_…`/`compose_…`) — the exact peel-and-compose shape the power arm mirrors, one axis over (`Cmc` → `PtComparison`). `compose_cost_mod_mana_value` is renamed `compose_cost_mod_qualifier_prop` (already `FilterProp`-generic; one caller) and called for both qualifiers.
- **`nom_filter::parse_pt_comparison`** (`oracle_nom/filter.rs`) — the canonical P/T-comparison combinator (self-consumes `with `/`base `, gives GE/LE/EQ), already used by Goreclaw's attack trigger and `oracle_target::parse_power_suffix`. No new power parser.
- **`object_pt_value`** (`game/filter.rs`) — the single authority the live/battlefield matcher already uses to read an object's P/T; the new cost-path arm reuses it against the live spell object, so power evaluation is identical across matchers.

## Verification

- `cargo fmt --all` clean; `cargo clippy -p engine --lib --tests -- -D warnings` **0 warnings**; `cargo test -p engine --lib` **17,701 passed / 0 failed** (no drift from the shared `filter.rs` change); `cargo test -p engine --test integration -- goreclaw` **4 passed**; `cargo test -p engine --lib -- cost_mod_power` **4 passed** (parser shape + attack-trigger control).
- **Integration tests** (`crates/engine/tests/integration/goreclaw_power_gated_cost_reduction_6375.rs`) drive the **real** `display_spell_cost` path over **real card faces via `add_real_card`** (printed power, never force-set), verbatim Goreclaw Oracle text:
  - power-5 creature (Fire Elemental) → reduced (generic 3→1); power-**4** boundary (Onakke Ogre) → reduced (2→0); power-3 creature (Hill Giant) → **not** reduced (stays 3); noncreature (Arc Lightning sorcery) → **not** reduced (stays 2). Each negative is reach-guarded by a reduced sibling in the same test (non-vacuous).
- **Bidirectionally revert-sensitive**: neutralizing the runtime arm re-fails the reduced/boundary tests (Goreclaw reduces nothing); reverting the parser `spell_filter` to `null` re-reduces the power-3 spell (the original "reduces all spells" bug, `left: 1, right: 3`).
- **Parser shape** unit tests: `GE` / `LE` / bare-`EQ` power qualifiers, plus the attack-trigger **control** (still `PtComparison{Power,GE,4}` on `Typed{Creature}`, untouched).
- **CR annotations** grep-verified against `docs/MagicCompRules.txt`: 208 / 208.1 / 208.3 (power is a creature characteristic; noncreatures have none), 601.2f (total cost), 202.3.

## Claimed parse impact

**Parser**: among real cards, only **Goreclaw**'s cost-modifier `spell_filter` changes (from `null` to the `Creature` + power-4 gate); the qualifier peel is all-or-nothing, so any subject lacking `" with power "` is byte-identical. **Runtime**: only the context-bearing cost / can't-cast / keyword-grant path gains `PtComparison` evaluation (all previously fail-closed `false`, so every flip is a bug fix); all snapshot/`None`-context callers are unchanged, confirmed by the full 17,701-test lib suite. The exact card-data parse-diff is deferred to CI's `coverage-parse-diff` artifact (local card-data/MTGJSON corpus unavailable) — I expect it to report only Goreclaw's cost-mod signature.

## Scope Expansion

None. The runtime arm was necessary to make the parser fix actually correct (otherwise Goreclaw reduces nothing) and is minimal + guarded (no serde/type change). Honest limitation, out of class: a `*/*` CDA creature spell reads power 0 off-battlefield (CR 208.3 / `parse_pt` maps `*`→0), so it wouldn't be reduced — a pre-existing off-battlefield CDA limitation, not introduced here and outside Goreclaw's fixed-power class.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed “power” gates to evaluate live spell objects’ effective P/T (including off-battlefield dynamic stats) with safe fail-closed behavior when the referenced spell object can’t be found.
  * Improved Oracle parsing for “with power N or greater/less” ModifyCost qualifiers (GE/LE/EQ) with correct handling order.
  * Prevented unparsed trailing qualifiers from widening the scope of static modifiers.
* **Tests**
  * Added/expanded unit and integration tests for power-gated cost reduction (Goreclaw), including boundary, noncreature, dynamic P/T (graveyard-driven), and fixed P/T scenarios.
  * Added a regression test to ensure ModifyCost coverage signatures expose spell-filter differences.

### DIFF
--- a/crates/engine/src/game/coverage.rs
+++ b/crates/engine/src/game/coverage.rs
@@ -4381,6 +4381,35 @@ fn static_details(stat: &StaticDefinition) -> Vec<(String, String)> {
     if let Some(zone) = &stat.affected_zone {
         d.push(("zone".into(), fmt_zone(zone)));
     }
+    // CR 601.2f (#6375, MED-2): a `ModifyCost` static's payload is otherwise
+    // invisible to the parse-diff — `StaticMode`'s Display collapses `ModifyCost`
+    // to a bare "ReduceCost"/"RaiseCost"/"MinimumCost" (the `ParsedItem::label`),
+    // dropping `amount`, `spell_filter`, and `dynamic_count`, and `static_details`
+    // projected none of them. So a change to WHICH spells the reduction targets
+    // (Goreclaw's `spell_filter` flipping from `None` to a power-gated creature
+    // filter) produced a byte-identical signature and a false "No card-parse
+    // changes detected". Project every semantically-relevant ModifyCost field so
+    // the diff surfaces the change. Deterministic: `fmt_target` and `Debug` are
+    // stable, ordering-free encodings.
+    if let StaticMode::ModifyCost {
+        mode,
+        amount,
+        spell_filter,
+        dynamic_count,
+    } = &stat.mode
+    {
+        d.push(("cost mode".into(), format!("{mode:?}")));
+        d.push(("cost amount".into(), format!("{amount:?}")));
+        d.push((
+            "cost filter".into(),
+            spell_filter
+                .as_ref()
+                .map_or_else(|| "any".into(), fmt_target),
+        ));
+        if let Some(dynamic) = dynamic_count {
+            d.push(("cost per".into(), format!("{dynamic:?}")));
+        }
+    }
     d
 }
 
@@ -10776,6 +10805,86 @@ mod tests {
     use crate::types::replacements::ReplacementEvent;
     use crate::types::statics::{BlockExceptionKind, ProhibitionScope};
     use crate::types::zones::{EtbTapState, Zone};
+
+    /// CR 601.2f (#6375, MED-2): a `ModifyCost` static's `spell_filter` must be
+    /// visible in the parse-diff signature. Goreclaw's fix flipped `spell_filter`
+    /// from `None` (reduces ALL spells) to a power-gated creature filter, but the
+    /// `ParsedItem` label (Display) collapses `ModifyCost` to a bare "ReduceCost"
+    /// and `static_details` projected no ModifyCost payload — so both variants
+    /// hashed identically and CI reported "No card-parse changes detected".
+    ///
+    /// Red/green: the two static_details projections differ ONLY when the
+    /// `spell_filter` row is emitted. Revert-probe: dropping the `static_details`
+    /// ModifyCost projection makes both vectors byte-identical and this fails.
+    #[test]
+    fn modify_cost_signature_exposes_spell_filter() {
+        use crate::types::ability::{
+            Comparator, FilterProp, PtStat, PtValueScope, QuantityExpr, TargetFilter, TypeFilter,
+            TypedFilter,
+        };
+        use crate::types::mana::ManaCost;
+        use crate::types::statics::{CostModifyMode, StaticMode};
+
+        let modify_cost_static = |spell_filter: Option<TargetFilter>| -> StaticDefinition {
+            StaticDefinition {
+                mode: StaticMode::ModifyCost {
+                    mode: CostModifyMode::Reduce,
+                    amount: ManaCost::generic(2),
+                    spell_filter,
+                    dynamic_count: None,
+                },
+                affected: Some(TargetFilter::Controller),
+                modifications: vec![],
+                condition: None,
+                per_player_condition: None,
+                affected_zone: None,
+                effect_zone: None,
+                active_zones: vec![],
+                characteristic_defining: false,
+                description: Some("cost reduction".to_string()),
+                attack_defended: None,
+                source_controller: None,
+                source_object: None,
+                bypass_beneficiary: None,
+                protection_does_not_remove: None,
+            }
+        };
+
+        // Goreclaw's real filter: creature spells with power 4 or greater.
+        let goreclaw_filter = TargetFilter::Typed(TypedFilter {
+            type_filters: vec![TypeFilter::Creature],
+            controller: None,
+            properties: vec![FilterProp::PtComparison {
+                stat: PtStat::Power,
+                scope: PtValueScope::Current,
+                comparator: Comparator::GE,
+                value: QuantityExpr::Fixed { value: 4 },
+            }],
+        });
+
+        let unfiltered = static_details(&modify_cost_static(None));
+        let filtered = static_details(&modify_cost_static(Some(goreclaw_filter)));
+
+        // The `spell_filter` change must surface in the projected signature.
+        assert_ne!(
+            unfiltered, filtered,
+            "a ModifyCost spell_filter change (None → power-gated creature filter) \
+             must produce a different coverage signature (issue #6375 MED-2)"
+        );
+        // Specifically, the "cost filter" row must reflect each value.
+        let filter_row = |d: &[(String, String)]| -> String {
+            d.iter()
+                .find(|(k, _)| k == "cost filter")
+                .map(|(_, v)| v.clone())
+                .expect("ModifyCost projection must emit a 'cost filter' row")
+        };
+        assert_eq!(filter_row(&unfiltered), "any", "None filter → \"any\"");
+        assert_ne!(
+            filter_row(&filtered),
+            "any",
+            "a power-gated creature filter must NOT project as \"any\""
+        );
+    }
 
     #[test]
     fn change_zone_signature_exposes_enters_attacking() {
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -3484,6 +3484,45 @@ fn spell_object_matches_property(
                 .get(&spell_id)
                 .is_some_and(|obj| !obj.kickers_paid.is_empty())
         }),
+        // CR 208 + CR 601.2f: Power/toughness-gated cost modifiers (Goreclaw's
+        // "creature spells you cast with power 4 or greater cost {2} less") must read
+        // the LIVE spell object's P/T — the SpellCastRecord snapshot carries none.
+        // Resolve the spell object via the filter context and evaluate against
+        // object_pt_value (the same authority the live matcher uses). CR 208.3: a
+        // noncreature spell has power None → object_pt_value → 0, so a GE gate fails
+        // closed for it. Snapshot-only callers (no context / no spell_object_id) fall
+        // through to the fail-closed record arm, unchanged.
+        FilterProp::PtComparison {
+            stat,
+            scope,
+            comparator,
+            value,
+        } => {
+            let Some(context) = context else {
+                return false;
+            };
+            let Some(spell_id) = context.spell_object_id else {
+                return false;
+            };
+            let Some(spell_obj) = context.state.objects.get(&spell_id) else {
+                return false;
+            };
+            // A dynamic (non-`Fixed`) threshold resolves against the modifier's
+            // SOURCE, not the candidate spell. No in-class card needs that today
+            // (Goreclaw is `Fixed(4)`, and the cost-mod parser cannot yet emit a
+            // P/T-relative threshold), so this is the correct scope for the class;
+            // a future spell-relative threshold would resolve with a spell scope.
+            let threshold = match value {
+                QuantityExpr::Fixed { value } => *value,
+                _ => resolve_quantity(
+                    context.state,
+                    value,
+                    context.source_controller,
+                    context.source_id,
+                ),
+            };
+            comparator.evaluate(object_pt_value(spell_obj, *stat, *scope), threshold)
+        }
         _ => spell_record_matches_property(record, prop),
     }
 }
--- a/crates/engine/src/game/filter.rs
+++ b/crates/engine/src/game/filter.rs
@@ -4134,20 +4134,26 @@ fn object_pt_value(obj: &GameObject, stat: PtStat, scope: PtValueScope) -> i32 {
 /// layer pass uses, rather than re-deriving CDA evaluation here: gather the
 /// continuous effects the object's own statics source
 /// (`active_continuous_effects_from_static_source`), keep only the
-/// characteristic-defining self-referential dynamic-P/T modifications, and
-/// resolve each magnitude with the layer pass's quantity resolver
-/// (`resolve_quantity`, mirroring `apply_continuous_effect` at layers.rs). Apply
-/// the layer-7a/7b *set* CDAs before the layer-7c *add* CDAs (CR 613.4a: P/T
-/// characteristic-defining abilities apply first, in layer 7a). A fixed-P/T
-/// object (Fire Elemental) sources
-/// no such effect, so this returns its stored `object_pt_value` unchanged.
+/// characteristic-defining self-referential P/T modifications whose condition
+/// currently holds — both fixed-constant *set* CDAs (`SetPower`/`SetToughness`,
+/// Angry Mob's off-turn "are each 2" clause) and dynamic *set*/*add* CDAs
+/// (Tarmogoyf) — resolving each dynamic magnitude with the layer pass's quantity
+/// resolver (`resolve_quantity`, mirroring `apply_continuous_effect` at
+/// layers.rs). Apply the layer-7a/7b *set* CDAs before the layer-7c *add* CDAs
+/// (CR 613.4a: P/T characteristic-defining abilities apply first, in layer 7a).
+/// A conditional CDA (Angry Mob's turn-window clauses) is gated exactly as the
+/// layer pass gates it — see `self_cda` below. A fixed-P/T object (Fire
+/// Elemental) sources no such effect, so this returns its stored
+/// `object_pt_value` unchanged.
 fn spell_object_effective_pt_value(
     state: &GameState,
     obj: &GameObject,
     stat: PtStat,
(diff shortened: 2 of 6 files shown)


matthewevans (changes requested):
**[MED]** Dynamic-P/T creature spells are rejected even when their CDA gives them power 4 or greater. Evidence: `crates/engine/src/game/filter.rs:3495-3525` evaluates `PtComparison` from the candidate object's stored P/T, but `crates/engine/src/game/printed_cards.rs:1876-1881` initializes `PtValue::Variable` and `PtValue::Quantity` to `0`; `crates/engine/src/game/layers.rs:1824-1854` resets only hand-zone keywords, not P/T. CR 208.2a requires a P/T characteristic-defining ability to function everywhere. Why it matters: Goreclaw must reduce an eligible CDA creature spell (for example, a Terravore/Tarmogoyf-like spell) rather than treating it as power 0. Suggested fix: resolve the candidate spell's off-battlefield characteristics through the CDA/layer authority before comparing power, with discriminating dynamic-P/T 3-vs-4+ production-path tests and a fixed-P/T control.

**[MED]** The required parse-diff is a false negative for this PR's claimed parser change. Evidence: the current-head parse-diff artifact/comment reports "No card-parse changes detected" despite this PR changing `StaticMode::ModifyCost.spell_filter`; `crates/engine/src/game/coverage.rs:4352-4387` projects only static `affected`, modifications, condition, CDA, and zone, omitting `ModifyCost` payload fields including `spell_filter`. Why it matters: CI cannot establish that Goreclaw's generated card-data filter changed, or detect a future blast radius in this semantic axis. Suggested fix: project semantically relevant `ModifyCost` fields (including `spell_filter`) into the coverage parse signature, add a red/green parse-diff regression test, and regenerate CI so the sticky identifies Goreclaw.

michiot05:
Thanks — both fixed at head `18d7eb897`.

**MED 1 — dynamic-P/T CDA spells read as power 0 off-battlefield.** Confirmed: `object_pt_value` returned the stored `obj.power`, which is `Some(0)` for a `*`-P/T CDA card (`parse_pt` maps `Variable`/`Quantity` → 0) and is only re-derived by the layer pass for battlefield objects. New helper `spell_object_effective_pt_value` (filter.rs) resolves the candidate spell's off-battlefield P/T through the **existing layer authority** — `active_continuous_effects_from_static_source` to gather its self-sourced continuous effects, then the same `continuous_modification_dynamic_quantity` → `resolve_quantity`/`resolve_quantity_with_recipient` path `apply_continuous_effect` uses — keeping only characteristic-defining `SelfRef` P/T modifications (CR 208.2a — a P/T CDA functions in all zones), applying set-CDAs (layer 7a/7b, CR 613.4a) before add-CDAs (7c). It falls back to the stored `object_pt_value` byte-identically when the object sources no self-CDA, so fixed-P/T spells are unchanged. It's called **only** from the cost-path `PtComparison` arm; the battlefield and zone-change matchers (whose objects already carry layer-computed P/T) are untouched.

Discriminating production-path tests via `display_spell_cost` over a real **Tarmogoyf** face (genuine `*` CDA, not force-set): a 4-distinct-type graveyard → CDA power 4 → **reduced** (generic 1→0); a 3-type graveyard → power 3 → **not reduced**; a fixed-power control (Fire Elemental) → still reduced regardless. Revert-proven (routing back to `object_pt_value` reads 0 → the power-4 test flips to not-reduced).

**MED 2 — coverage parse-diff omitted `ModifyCost.spell_filter`.** `static_details` never projected `stat.mode`, so `StaticMode`'s Display collapsed `ModifyCost{..}` to a bare `"ReduceCost"` and Goreclaw's `spell_filter: None → Some(Typed{Creature,[PtComparison]})` hashed identically. `static_details` now emits deterministic `cost mode` / `cost amount` / `cost filter` / `cost per` rows for `ModifyCost` statics, with a red/green regression (`modify_cost_signature_exposes_spell_filter`) proving the `None`-vs-filter signatures now differ. **Heads-up for the parse-diff sticky:** this is a coverage *schema* widening, so on regeneration the `coverage-parse-diff` comment will list the whole `ModifyCost` class as "changed" (every such card gains the new signature rows) — that's expected, not parser blast radius; Goreclaw's `spell_filter` change is the one now correctly surfaced.

Verified: `cargo test -p engine --test integration --no-run` clean against merged-main, `cargo test -p engine --lib` **17,707 passed / 0 failed**, clippy 0, Gate A PASS. CR cites grep-verified (208.1/208.2a/601.2f/613.4a/613.4b/613.4c).

matthewevans (changes requested):
**Verdict: blocked — the dynamic-P/T fix is present, but the new off-battlefield CDA evaluator still omits conditional fixed P/T.**

## 🔴 Blocker

**[HIGH]** Off-battlefield CDA evaluation skips fixed power/toughness setters. Evidence: `crates/engine/src/game/filter.rs:4182-4198` applies only dynamic `Set*` variants before returning the computed value, while `crates/engine/src/parser/oracle_static/tests.rs:825-900` shows that the parsed conditional constant P/T CDA form uses `ContinuousModification::SetPower { value: 2 }` and `SetToughness { value: 2 }` (Angry Mob's off-turn clause). Why it matters: the new cost-filter path starts from an off-battlefield object's stored P/T and can therefore evaluate a card whose active constant CDA is 2/* as printed `0/*`, incorrectly failing a power gate. Suggested fix: handle applicable self-CDA `SetPower`/`SetToughness` in the resolver's initial set stage, in the same ordering as the layer authority, and add a discriminating production `display_spell_cost` test for a conditional constant-P/T creature spell.

**Recommendation: request changes for the missing fixed-setter CDA path.**

michiot05:
Thanks — fixed at head `5de755085`.

You're right: the off-battlefield evaluator applied only the dynamic `Set*`/`Set*Dynamic` CDAs and skipped the fixed-constant `SetPower { value }` / `SetToughness { value }` setters, so Angry Mob's constant off-turn clause (2/*, or a 4/* form) was read as printed. Added both fixed setters to the **same layer-7a/7b set stage** as the dynamic setters (ahead of the 7c add stage), matching the layer authority (`layers.rs` lists all six P/T-set variants together).

While there I also hardened the shared `self_cda` filter with the layer pass's **exact per-recipient condition gate** — `effect.condition.is_none_or(evaluate_condition_with_recipient(.., obj.id))`, a byte-for-byte mirror of `apply_continuous_effect_filtered` (layers.rs:5838-5848), with the self-CDA's recipient = its source. `active_continuous_effects_from_static_source` already drops defs whose non-recipient-context condition fails at the source level (Angry Mob's `DuringYourTurn`/`Not{DuringYourTurn}` are pre-filtered there), so this is a no-op for that card; it additionally honors a *retained recipient-context* condition for both the fixed and dynamic paths, so a conditional CDA is never applied outside its window. `None`-condition CDAs (Tarmogoyf, fixed-P/T) pass through unchanged.

Discriminating tests through the production `display_spell_cost` path (real Hill Giant face, conditional constant-P/T CDA parsed via `parse_static_line`, attached post-rehydrate — no force-set power):
- `During your turn … are each 4` → active on your turn → effective power 4 → **reduced** (3→1).
- `During turns other than yours … are each 4` → `Not{DuringYourTurn}` false on your turn → CDA not applied → printed 3 → **not reduced** (the condition-gating discriminator).

Revert-proven: removing the two fixed arms flips the active case (reads printed 3, GE-4 rejects); restored → green. Tarmogoyf dynamic-CDA (power 4 reduced / 3 not) + Fire Elemental fixed control still pass. `cargo test -p engine --lib` 17,707 / 0; integration `--no-run` compiles; Goreclaw suite 9/9; clippy 0; Gate A PASS. CR 613.4a / 611.3a grep-verified.

matthewevans:
Current-head disposition for `5de75508568f4e1614cb2540eb00b6fa78c17735`: the prior fixed-setter CDA blocker is resolved, but this PR is not ready for approval or the merge queue.

**[MED]** The required parse-diff does not substantiate the PR's claimed one-card scope and is stale for the current head. Evidence: the PR says `only Goreclaw` and `Scope Expansion: None`, but the required `<!-- coverage-parse-diff -->` comment reports **555 cards / 249 signatures** and was last updated `2026-07-26T09:09:11Z`, before this head was pushed (`2026-07-27T05:37:05Z`). The broad migration is traceable to [`coverage.rs:4384`](https://github.com/phase-rs/phase/blob/5de75508568f4e1614cb2540eb00b6fa78c17735/crates/engine/src/game/coverage.rs#L4384), which begins projecting every `ModifyCost` payload field, but that makes the current artifact unable to prove the stated Goreclaw-only parser impact. Why it matters: parse-diff is the required blast-radius evidence for this engine/parser PR; the present artifact contradicts the scope claim and cannot be treated as current-head evidence. Suggested fix: publish a fresh current-head parse-diff and update the PR scope/claim to account for the signature-schema migration, or split that migration so Goreclaw's parser delta is independently reviewable.

**[MED]** The regression suite does not drive the actual cast action. Evidence: [`goreclaw_power_gated_cost_reduction_6375.rs:77`](https://github.com/phase-rs/phase/blob/5de75508568f4e1614cb2540eb00b6fa78c17735/crates/engine/tests/integration/goreclaw_power_gated_cost_reduction_6375.rs#L77), [`:191`](https://github.com/phase-rs/phase/blob/5de75508568f4e1614cb2540eb00b6fa78c17735/crates/engine/tests/integration/goreclaw_power_gated_cost_reduction_6375.rs#L191), and [`:322`](https://github.com/phase-rs/phase/blob/5de75508568f4e1614cb2540eb00b6fa78c17735/crates/engine/tests/integration/goreclaw_power_gated_cost_reduction_6375.rs#L322) call `display_spell_cost`; the production cast branch enters `prepare_spell_cast` in [`casting.rs:4570`](https://github.com/phase-rs/phase/blob/5de75508568f4e1614cb2540eb00b6fa78c17735/crates/engine/src/game/casting.rs#L4570) with `CastingMode::Actual`, not a `GameAction`/payment flow. Why it matters: this bug fixes casting-cost behavior, so a display-only assertion does not prove that an actual cast charges the filtered cost. Suggested fix: add a scenario that submits the real cast action with sufficient mana and asserts the power-3 spell pays its full generic cost while the power-4+ spell pays the reduced cost.

The security pre-check is clean (only the six engine/parser/test paths in this PR); no maintainer fixup was pushed. Current CI remains pending on Rust lint and both Rust test shards. The `bug` label is present; no `quality` label or merge-queue entry has been added.

matthewevans (changes requested):
Current-head disposition for `5de75508568f4e1614cb2540eb00b6fa78c17735`: changes requested.

**[MED] Fresh parse-diff / scope mismatch.** The required `<!-- coverage-parse-diff -->` artifact does not substantiate this PR's claimed Goreclaw-only scope and is stale for the current head. The PR says `only Goreclaw` and `Scope Expansion: None`, while the artifact reports **555 cards / 249 signatures** and predates this head. The broad result comes from projecting every `ModifyCost` payload field in `coverage.rs`, so it is a signature-schema migration as well as the Goreclaw filter change. Publish a fresh parse-diff for the current head and update the scope claim to account for that migration, or split the migration so Goreclaw's parser delta is independently reviewable.

**[MED] Display-only cast-cost regression test.** The Goreclaw regression suite asserts `display_spell_cost`, but does not drive the actual cast action / payment path (`prepare_spell_cast` with `CastingMode::Actual`). Because this fixes casting-cost behavior, add a scenario that submits the real cast action with sufficient mana and proves the power-3 spell pays its full generic cost while the power-4-or-greater spell pays the reduced cost.

These current-head blockers must be resolved before approval or merge-queue entry.

matthewevans:
The requested changes on head `5de75508568f4e1614cb2540eb00b6fa78c17735` remain unresolved, and there has been no subsequent contributor update. Please address the current-head blockers—fresh parse-diff/scope evidence and a regression that drives the actual cast/payment path—within 7 days. If they remain unaddressed, this PR will be automatically closed under the review policy. A substantive push or response before then will return it to review.

matthewevans:
Closing this PR because the current-head requested changes remain unresolved after the warning posted on 2026-08-03: fresh parse-diff/scope evidence and a regression through the actual cast/payment path. The head has not changed and there has been no contributor follow-up during the seven-day warning window. A future, refreshed submission addressing those blockers is welcome.

Issue #6567, opened 2026-07-23
Player owes bottom cards before making another mulligan decision


pascalberger:
When playing against AI I sometimes see the following error:

```
[ERROR] processQueue error (local): Engine error: Invalid action: Player PlayerId(1) owes bottom cards before making another mulligan decision
```

mike-theDude:
## Triage analysis

**Clarified summary**
During games against the AI, the engine intermittently rejects a mulligan decision with an error stating the player still owes bottom cards before making another mulligan decision.

**Investigation findings**
The error is raised when a mulligan decision arrives for a player whose mulligan entry is still in its bottom-cards phase. Two AI-side paths can produce exactly that:

1. The AI's fallback action selector reads the *first* entry in the pending mulligan set rather than the entry belonging to the player it is acting for. When more than one player is still resolving a mulligan, it can therefore emit a declare-phase decision while its own entry is owed bottoms.
2. That same fallback answers a bottom-cards phase with an empty card selection. Bottom-selection validation requires exactly the owed count, so an empty submission is rejected whenever the count is non-zero, leaving the obligation unmet and the fallback looping.

The main AI mulligan policy resolves the acting player's own entry correctly, so this is specific to the fallback path taken after repeated decision failures.

**Reproduction status**
Confirmed as an actionable defect from code inspection. A saved game state from a game where the error appears would let the fix be pinned to one of the two paths.

**Likely scope**
`area:ai` for the fallback action selector; `area:engine` for the mulligan phase contract it violates.

**Acceptance criteria**
- The AI fallback resolves the pending mulligan entry for the player it is acting for, never the first entry in the set.
- A bottom-cards phase is answered with exactly the owed number of cards.
- A game against the AI completes mulligans without this error, including when both players mulligan.

**Related issues**
#6342 reports the same obligation going unmet from the human side, where an order-sensitive legality check rejects the submission. Related root symptom, distinct cause; tracked separately.

pascalberger:
This is still happening after #6628 has landed

matthewevans:
Linked a new Discord report with the same mulligan-bottom-card error; the reporter also lost the game after refresh.

matthewevans:
[...]

PR #6628, opened 2026-07-25
fix(mulligan): accept any-order bottom-card selection during mulligan bottoming

### LINKED ISSUE #6342: Mulligan bottom-cards selection rejected unless card order matches internal hand order
## Bug

`SelectCards` submitted for the mulligan bottom-cards step is rejected as an
illegal action unless the submitted card order happens to match an internal,
arbitrary hand order. In practice this makes bottoming 2+ cards (any mulligan
taken more than once) fail almost every time a human clicks cards in a
natural order, since nothing in the UI or docs tells the player which order
is "correct."

Server-side error observed:

```
Illegal action: SelectCards { cards: [ObjectId(32), ObjectId(6), ObjectId(17)] }
```

## Root cause

CR 103.5 lets a player choose any order when putting mulligan-owed cards on
the bottom of their library, and `validate_bottom_selection`
(`crates/engine/src/game/mulligan.rs:535-560`) correctly implements this: it
only checks the count and hand membership of the submitted cards, not their
order.

However, the server's legality pre-check in
`crates/server-core/src/session.rs` (around line 1287-1318) only bypasses the
enumerated-candidate legality gate for a fixed allowlist of "freeform"
`WaitingFor` states (`accepts_freeform_card_selection`, defined in
`crates/engine/src/types/game_state.rs:9769-9777`):

```rust
pub fn accepts_freeform_card_selection(&self) -> bool {
    matches!(
        self,
        WaitingFor::ScryChoice { .. }
            | WaitingFor::ArrangePlanarDeckTopChoice { .. }
            | WaitingFor::SurveilChoice { .. }
            | WaitingFor::DigChoice { .. }
    )
}
```

`WaitingFor::MulliganDecision` (its `BottomCards` sub-phase) and
`WaitingFor::OpeningHandBottomCards` are not in this list, even though they
are the same class of "count-constrained, any-order" selection as
Scry/Surveil/Dig.

Because the bypass is missing, `SelectCards` submissions during mulligan
bottoming fall through to the enumerated-candidate check
(`legal_actions.contains(...)`), which requires an exact match against
`GameAction`'s derived `PartialEq`, which is order-sensitive for the
`cards: Vec<ObjectId>` field. The candidates themselves are generated by
`bottom_card_actions`/`combinations` in
`crates/engine/src/ai_support/candidates.rs:4483-4515`, which enumerates each
subset of the hand exactly once, in the hand's internal storage order, not
sorted, not in UI display order, and unrelated to click order.

On the client, `selectedCardIds` (`client/src/stores/uiStore.ts`) is a plain
array built by click/toggle order (`[...state.selectedCardIds, cardId]`), and
`MulliganBottomCardsPrompt` / `handleBottomCards`
(`client/src/pages/GamePage.tsx:1140-1146`, `2414-2453`) sends that array
verbatim. Nothing reconciles the two orders, so unless a player's click order
coincidentally matches the hand's internal order, the submission is rejected.

## Scope: server-hosted "Full" mode only, not solo/AI or P2P

This only affects games played through a server-authoritative session (a
`phase-server` instance in Full mode, over `/ws`), whether self-hosted or
otherwise. Confirmed by reading `crates/engine-wasm/src/lib.rs`: its
`submit_action` (used for solo-vs-AI and P2P play) calls the engine's `apply()`
directly with no enumerated-candidate legality pre-check at all, no
`legal_actions.contains(...)` gate, and no `"Illegal action: {:?}"` wrapper.
It relies solely on the engine's own internal validation
(`validate_bottom_selection`), which is already correct and order-insensitive.
The ordering gate described above exists only in
`crates/server-core/src/session.rs`. So solo/AI and P2P games cannot hit this;
any server-hosted multiplayer game can.

## Suggested fix

Add `WaitingFor::MulliganDecision { .. }` (specifically while a pending
entry's phase is `BottomCards`) and `WaitingFor::OpeningHandBottomCards { .. }`
to `accepts_freeform_card_selection`, mirroring the existing Scry/Surveil/Dig
exception. `validate_bottom_selection` is already the correct, order-insensitive
authority and needs no changes.

## Repro

1. Start or join a game, take 2 or more mulligans (owes 2+ bottom cards).
2. Keep the hand.
3. Click any 2+ hand cards to bottom, in whatever order they visually appear.
4. Confirm. Server rejects with `Illegal action: SelectCards { cards: [...] }`
   unless the click order happens to match the hand's internal order.

Workaround: repeatedly reselect the same cards in a different order until one
permutation is accepted (up to `count!` attempts).

---

Found via LLM-assisted source review (Claude, model: claude-sonnet-5) while
debugging a real in-game mulligan failure. All file/line references above
were read directly from the v0.34.0 source; not verified by the reporter with
an independent Rust build.

nghetien:
Fixes #6342.

When a player owes bottom cards from a mulligan (CR 103.5), they're allowed to put them on the bottom of their library in any order. The engine's own validation (`validate_bottom_selection`) already gets this right — it only checks the count and hand membership of the submitted cards, not their order.

The problem was one level up, in the server's legality pre-check. `WaitingFor::accepts_freeform_card_selection()` is the allowlist that tells the server "don't bother checking this `SelectCards` submission against the enumerated candidate list, the handler validates it structurally itself" — it already covered Scry, Surveil, and Dig, but `MulliganDecision` (while a pending entry is in the `BottomCards` sub-phase) and `OpeningHandBottomCards` were missing. So any bottom-card submission during mulligan whose order didn't happen to match the hand's internal storage order got rejected with `Illegal action: SelectCards { ... }`, even though it was a perfectly legal selection.

The fix just adds both to the allowlist, mirroring the existing pattern:

```rust
pub fn accepts_freeform_card_selection(&self) -> bool {
    matches!(
        self,
        WaitingFor::ScryChoice { .. }
            | WaitingFor::ArrangePlanarDeckTopChoice { .. }
            | WaitingFor::SurveilChoice { .. }
            | WaitingFor::DigChoice { .. }
            | WaitingFor::OpeningHandBottomCards { .. }
    ) || matches!(
        self,
        WaitingFor::MulliganDecision { pending, .. }
            if pending.iter().any(|entry| matches!(entry.phase, MulliganDecisionPhase::BottomCards { .. }))
    )
}
```

`MulliganDecision` needs the guard because a pending entry can also be in `Declare` phase, which takes Keep/Mulligan/UseSerumPowder, not `SelectCards`. `OpeningHandBottomCards` has no such sub-state, so it's unconditional.

Added test coverage at both layers:
- Extended the existing predicate unit test in `game_state.rs` with cases for `BottomCards` (both `Keep` and `UseSerumPowder` continuations), a mixed pending set proving the guard is `.any()` not `.all()`, `OpeningHandBottomCards`, and a negative case for all-`Declare` pending entries.
- Added two new end-to-end tests in `session.rs`, mirroring the existing `reordered_scry_selection_is_accepted_not_rejected_as_illegal` / `reordered_dig_selection_is_accepted_not_rejected_as_illegal` tests: build a real session, submit a reordered bottom-card selection, and assert it's accepted and actually applied. Confirmed these reproduce the exact `Illegal action` error pre-fix.

Ran `cargo fmt --all -- --check` (clean), `cargo clippy --workspace --exclude phase-tauri --all-targets --features engine/proptest -- -D warnings` (zero warnings), and the full `cargo nextest run --profile ci` suite (23928 passed, 22 skipped) against the current `origin/main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reordered bottom-card selections are now accepted during both mulligan bottoming and forced opening-hand setup.
  * Duplicate cards in bottoming selections are now rejected to prevent invalid double-counting.
  * Refined which mulligan decisions allow freeform selection, with correct handling for mixed pending bottom-card states.

* **Tests**
  * Added regression tests covering reordered bottoming submissions and verifying acceptance/rejection, including duplicate-card cases.


pascalberger:
Does this also fix #6567

nghetien:
Possibly — the error in #6567 comes from the mulligan-decision guard in `mulligan.rs`, which this PR doesn't touch directly. However, if the AI's bottom-card submission was being rejected due to the ordering bug fixed here (#6342), the follow-up decision would hit exactly that guard. So this PR likely removes the most common trigger, but I'd suggest retesting #6567 against a build with this fix before closing it.

matthewevans (changes requested):
**[HIGH]** Freeform mulligan bottoming accepts duplicate card IDs, allowing a player to satisfy a two-card bottoming obligation with one card. Evidence: `crates/server-core/src/session.rs:1310-1313` now skips the enumerated-action gate for these waiting states, while the shared `validate_bottom_selection` at `crates/engine/src/game/mulligan.rs:535-559` checks only submitted length and pre-move hand membership. Both bottoming handlers then loop the submitted IDs and credit `expected_count` (`mulligan.rs:435-445`, `509-520`). Why it matters: `SelectCards { cards: vec![b, b] }` for count 2 now passes, moves `b` twice (the second move merely repositions it within the library), credits two paid bottoms, and leaves `a` in hand—an invalid opening-hand result newly exposed by this bypass. Suggested fix: reject duplicate `ObjectId` values in the shared validator before moving anything, and add negative `SessionManager::handle_action` regressions for duplicate selections in both `MulliganDecision::BottomCards` and `OpeningHandBottomCards`, asserting rejection and unchanged hand/library/pending state.

matthewevans (approved):
Current-head re-review of `7b2a8e280fcd5c9f589dd4c24c7adcef1b773d93`: current main owns action admission in the engine, so the obsolete server allowlist was removed during the port. The remaining shared validator rejects duplicate bottom-card IDs before any move, and runtime session regressions cover both normal mulligan and the Tiny Leaders opening-bottom sibling. The prior duplicate-selection request and the current CR 103.5 annotation request are addressed. Required CI for this pushed head is still running.

PR #6629, opened 2026-07-25
fix(engine): transform "all <type>" as a mass effect, no target (Moonmist)

### LINKED ISSUE #6403: Moonmist — Moonmist should not have to target a single human creature, should just transform all human.
Reported in Discord: https://discord.com/channels/1485498006781427802/1528361013378940978/1528361013378940978

**Thread:** Moonmist
**Discord thread id:** `1528361013378940978`
**Discord message id:** `1528361013378940978`
**Cards:** moonmist

## Summary
Moonmist should not have to target a single human creature, should just transform all human.

## Reported text

Moonmist should not have to target a single human creature, should just transform all human.

## Attachments

- [image.png](https://cdn.discordapp.com/attachments/1528361013378940978/1528361014331183216/image.png?ex=6a61f91d&is=6a60a79d&hm=ee88a82e862503aa5cf000490a7d9b9d318049209f80c105e7374109d62622c8&)

### moonmist
Transform all Humans. Prevent all combat damage that would be dealt this turn by creatures other than Werewolves and Wolves. (Only double-faced cards can be transformed.)

---
<sub>report_id: `discord:1528361013378940978:1528361013378940978:0` · discord: `1528361013378940978/1528361013378940978`</sub>

michiot05:
Tier: Frontier
Model: claude-opus-4-8

Closes #6403.

## Summary

**Moonmist** — *"Transform all Humans. Prevent all combat damage that would be dealt this turn by creatures other than Werewolves and Wolves."* — forced a **single-target prompt** to pick one Human. The parser emitted a single-target `Effect::Transform { target: Typed{Subtype Human} }`, and the effect exposed a target slot, so a **non-targeting mass effect** was resolved as a targeted one (only the source transformed). "Transform all X" doesn't target (CR 115.10) — it should transform every Human on the battlefield with no prompt.

Rather than add a `TransformAll` sibling, this **scope-parameterizes** `Transform` to carry the single-vs-mass axis — the exact refactor the codebase already applied to `Tap`/`TapAll` → `SetTapState { scope }` (the `EffectScope` doc-comment names it), so `/add-engine-variant` approves it as parameterization, not proliferation:

```
Effect::Transform { target } → Effect::Transform { target, scope: EffectScope }
```

- **`EffectScope::Single`** (serde default): the legacy targeted/anaphoric transform (`"transform target creature"`, `"transform ~"`), unchanged.
- **`EffectScope::All`**: a non-targeting **population filter** enumerated at resolution. `Effect::target_filter()` returns **`None`** for `All` (so no target slot / prompt is built — mirroring `SetTapState`/`DestroyAll`), and a new `resolve_all` iterates the battlefield population applying the transform to each match, mirroring `tap_untap::resolve_all`.

Two rules-correctness details that make the mass path actually work:
- **`target_filter()` is the single target-exposure authority.** Both production slot-builder paths route through it, so `None` for `All` fully removes the prompt (CR 115.10).
- **Non-transformable matches are a harmless no-op** (CR 701.27c). `resolve_all` pre-filters the matched population to double-faced permanents (`is_double_faced_permanent`) so a matched single-faced creature is skipped rather than erroring — `transform_permanent` returns `Err("no back face")` for a non-DFC, so the mass loop must **not** propagate it (a naive mirror would abort the whole transform on the first single-faced Human).

The parser gains a `"transform/convert all|each <filter>"` mass branch (before the single branch), the loop-safety census classifiers scope-split `Transform` to a live-board census for `All` (mirroring `DestroyAll` — a wrong classification would defeat the LoopFirewall), and every `SetTapState{Single}`-gated site gets the parallel `Transform` gate. The serde default keeps old serialized `Transform{target}` deserializing to `Single` (byte-compatible).

Method: /engine-implementer — plan (/engine-planner + /add-engine-effect, with a run-then-delete reproduction probe confirming the single-target prompt) + the **/add-engine-variant gate** (verdict: scope-parameterize per the `SetTapState` precedent; corrected the CR citations — transform is 701.27, not 701.28) → **/review-engine-plan** (found **5 blocking gaps**: the plan compiled but did not fix the bug — the `target_filter()` linchpin was untouched, `transform_permanent` errors rather than no-ops on non-DFCs, the census classifiers were left scope-blind, ≥5 compile sites and the `SetTapState`-gate matrix were incomplete — all corrected) → implement → /review-impl (Maintainer-Simulation Gate PASS; independently confirmed all 5 blockers correct — `target_filter()` is the *only* target-exposure surface, census classification correct, Single-gate parity complete; one unreachable LOW noted below). Each step in a fresh agent context.

## Gate A

`./scripts/check-parser-combinators.sh` → **Gate A PASS head=1613e601a** (Gate G PASS). The mass parser branch is composed combinators (`preceded`/`alt`/`tag`/`peek`).

## Anchored on

- **`Effect::SetTapState { target, scope, state }`** (`types/ability.rs`, the `Tap`/`TapAll` unification) — the end-to-end template: the `EffectScope` axis, the `target_filter()` scope-split (`Single => Some(target)`, `All => None`), the census-classifier scope arms, and the `Single`-gate sites are all mirrored for `Transform`.
- **`game/effects/tap_untap.rs::resolve_all`** — the mass-resolution shape (`resolved_object_filter` + `matches_target_filter` over `state.battlefield`, per-object, no targets) that `transform_effect::resolve_all` mirrors, with the DFC pre-filter added for CR 701.27c.
- **`game/ability_scan.rs` `DestroyAll`/`Suspect{All}` census arms** — the correct loop-safety classification (`LiveBoardCensus`/`Census`) for a mass effect that iterates the growing battlefield; `Transform{All}` joins that group (explicitly *not* the `SetTapState`-only `SnapshotOrEvent` exception, which stays pinned by `obligation_ii_census_exception_is_exactly_settapstate`).

## Verification

- `cargo fmt --all` clean; `cargo build -p engine -p mtgish-import -p phase-ai` clean (every scope site threaded); `cargo clippy -p engine --lib --tests -- -D warnings` **0 warnings**; `cargo test -p engine --lib` **17,706 passed / 0 failed** (no drift from the shared `Effect` change, incl. the census/LoopFirewall tests); Gate A/G PASS.
- **Revert-sensitive tests** (production path — real `parse_oracle_text` + real resolver, `back_face` DFC modeling):
  - `moonmist_transforms_all_humans_without_a_prompt` (primary): parse Moonmist **verbatim**, resolve over 2 DFC Humans + a Goblin + a Werewolf → both Humans transform, Goblin/Werewolf untouched, **no `WaitingFor` prompt**, `target_filter().is_none()`, PreventDamage sibling preserved.
  - `mass_transform_exposes_no_target_slot` (B1): `Transform{All}.target_filter() == None`, `Single` → `Some(target)`.
  - `mass_transform_skips_single_faced_human_without_error` (B2): a single-faced Human (`back_face:None`) in the population → the DFC transforms, the single-faced Human is untouched, resolution does **not** error (removing the DFC pre-filter makes it error — revert-sensitive).
  - `parse_transform_all_is_mass_scope` / `parse_transform_target_is_single_scope`: the mass branch emits `scope:All`; `"transform target creature"` stays `Single` (control — mass branch didn't regress single).
- **CR annotations** grep-verified against `docs/MagicCompRules.txt`: 701.27 / 701.27a (transform), 701.27c (a permanent that can't transform does nothing), 115.10 / 115.10a ("all X" is non-targeting).

## Claimed parse impact

Serialized `Effect::Transform` gains a `scope` field: existing Transform cards mechanically gain `"scope":"Single"` (serde default → byte-compatible deserialization of old states), and **Moonmist** becomes `"scope":"All"`. The parser mass branch fires only on literal `"transform/convert all|each <filter>"`, so no other card's Transform parse changes shape. `mtgish-import` threads `scope:Single` on its existing `TransformPermanent` action (a forced compile-consequence of the shared reshape). Frontend unchanged (no client `Effect::Transform` handling). The exact card-data parse-diff (which cards carry a "transform all" line beyond Moonmist) is deferred to CI's `coverage-parse-diff` artifact — the scope parameter is the general building block for the whole mass-transform class.

## Scope Expansion

None. Honest latent parity note (out of class, not a regression): the tracked-set-recording arm that special-cases `SetTapState{All}` for chained "those cards" sub-abilities has no `Transform{All}` sibling — unreachable today (Moonmist is a `Typed` population filter and no tracked-set mass-transform card exists; `Transform{Single}` had no such handling either). I left it rather than add a speculative guard; a `Transform{All}` arm can be added if such a card ever appears.

## Summary by CodeRabbit

- **New Features**
  - Added end-to-end support for mass “transform/convert all” so qualifying double-faced permanents transform in a single resolution.
  - Transform instructions now explicitly distinguish single-target vs non-targeting mass behavior.

- **Bug Fixes**
  - Corrected Transform target/coverage behavior to be scope-aware, preventing mix-ups between single and mass variants.
  - Updated parsing so unsupported front-face restrictions fail safely instead of degrading into incorrect behavior.

- **Tests**
  - Added a Moonmist mass-transform regression test (including “no extra prompt” behavior) and expanded parser/resolution coverage for scope handling.


matthewevans (changes requested):
**[HIGH]** Mass transform drops the printed `on their front face` restriction for That's No Moonmist. Evidence: [`imperative.rs:5334`](https://github.com/phase-rs/phase/blob/6400ed502b78d239d407b4006c853957da5cc522/crates/engine/src/parser/oracle_effect/imperative.rs#L5334) discards the remainder from `parse_target_with_ctx` and immediately emits `EffectScope::All`; the current parse-diff therefore records That's No Moonmist only as `artifact or creature Phyrexian`. Its current Oracle text is: “Transform all artifacts and Phyrexian creatures on their front face.” `resolve_all` then transforms every matched DFC, including a matching back-face permanent that the card must leave unchanged.

Please either model the front-face condition as a typed population predicate and add a regression that keeps a matching transformed permanent unchanged, or strict-fail/preserve `Effect::unimplemented` whenever this mass-transform branch leaves that semantic suffix unparsed. The latter is acceptable if the predicate is intentionally deferred; silently marking the card supported is not.

matthewevans (changes requested):
**[HIGH]** Mass transform still silently drops the printed `on their front face` restriction for That's No Moonmist. [`imperative.rs:5334`](https://github.com/phase-rs/phase/blob/ec85f1d25d8454b6df7db659bbdc0e27c1edf55f/crates/engine/src/parser/oracle_effect/imperative.rs#L5334) discards the parsed suffix as `(target, _)` before emitting `EffectScope::All`; the PR's published parse-diff consequently records only the `artifact or creature Phyrexian` filter. The card's Oracle text also restricts the population to objects on their front face.

`resolve_all` currently filters only by that target filter and double-faced status, so it can transform a matching DFC already on its back face, although that object must be left unchanged.

The prior requested change at `6400ed502b78d239d407b4006c853957da5cc522` was not remediated by the merge-only current-head change. Please carry the front-face restriction as a typed population predicate and add a production cast-path regression containing a matching back-face permanent; otherwise strict-fail this rules-bearing suffix until it is represented.

michiot05:
Thanks — fixed at head `4ddb629d0`.

**Front-face restriction (That's No Moonmist).** Took the strict-fail option you offered: there's no `parse_target_with_ctx` support for consuming an *"on their front face"* postnominal into a typed population predicate today, and modeling it for essentially one card would be a special case rather than a building block — so I've made the mass branch **strict-fail** rather than silently drop the suffix.

The `"transform/convert all|each <filter>"` branch now captures the `parse_target_with_ctx` remainder (previously discarded as `_`):
- Empty remainder (Moonmist's *"all Humans"*, fully consumed) → `EffectScope::All`, unchanged.
- A surviving rules-bearing suffix (*" on their front face"*) → `return None`, so the line falls through to `Effect::unimplemented` (the single authority — not a hand-constructed literal), and critically does **not** fall through to the single `"transform <target>"` branch that would emit a targeted `scope: Single` transform still dropping the suffix.

This also closes the `resolve_all` back-face concern you raised: strict-failing the suffix-bearing cards means `resolve_all` only ever runs on fully-parsed populations like *"all Humans"*, where the Human typing already implies front face (a transformed back-face Werewolf is a Werewolf, not a Human, so it never matches the population filter) — the mass path can't transform a matching back-face permanent.

Regressions (both pass): `thats_no_moonmist_front_face_restriction_stays_unimplemented` (verbatim Oracle → `Effect::Unimplemented`, with a Moonmist control that stays a supported `scope: All`), plus a parser-shape assertion that the suffixed phrase returns `None`. Your `moonmist_cast_transforms_all_humans_without_target_selection` still passes.

**CI (was red).** The auto-merge of `main` into the branch introduced two enum-reshape drift breaks in the **integration** crate (which my earlier local run hadn't compiled against merged-main): `cr733_resolved_transform.rs` constructing `Effect::Transform` without the new `scope` field, and the `BackFaceData` in your added `issue_6403_moonmist_mass_transform.rs` missing `printed_loyalty` (added on `main` after the branch point). Both threaded (`scope: EffectScope::Single`, `printed_loyalty: None`). Verified this round with `cargo test -p engine --test integration --no-run` (the crate that was failing): integration **77 passed**, lib **17,747 passed / 0 failed**, clippy 0, Gate A PASS. CR annotations grep-verified (701.27, 115.10a, 712.2).

matthewevans:
Current head `93892a6001` has required CI still running. The implementation review is complete; approval and merge-queue enrollment will resume after those checks settle. The parse-diff baseline is pending, but this branch already contains the maintainer merge-main remedy, so it will not receive another merge-main commit solely to retry that artifact.

matthewevans (approved):
Approved: current-head regression coverage now exercises the matching non-transformable Human and the strict-failed face-state suffix; all required checks are green.

